PR #116, opened 2022-01-10
refactor ML task data model; add create ML task index method

ylwu-amzn:

### Description
1. Refactor ML task data model
2. Add create ML task index method
 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLTask.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLTask.java
@@ -47,7 +47,8 @@ public class MLTask implements ToXContentObject, Writeable {
     public static final String ERROR_FIELD = "error";
     public static final String USER_FIELD = "user";
 
-    private final String taskId;
+    @Setter
+    private String taskId;
     private final String modelId;
     private final MLTaskType taskType;
     private final FunctionName functionName;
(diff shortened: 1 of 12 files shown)


Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/model/MLTaskState.java
```diff
@@ -33,5 +33,6 @@
     CREATED,
     RUNNING,
     COMPLETED,
-    FAILED
+    FAILED,
+    CANCELLED
```
Zhangxunmt, line 37: Have we used this CANCELLED status anywhere, or this is reserved for future API usage?

ylwu-amzn: This is a reserved state

plugin/src/main/java/org/opensearch/ml/model/MLTask.java
```diff
@@ -23,45 +23,161 @@
...
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.commons.authuser.User;
+import org.opensearch.ml.common.dataset.MLInputDataType;
+import org.opensearch.ml.common.parameter.FunctionName;
 
 @Getter
 @EqualsAndHashCode
-public class MLTask implements Writeable {
-    private final String taskId;
+public class MLTask implements ToXContentObject, Writeable {
```
Zhangxunmt, line 34: We can move this class to the ml-common if we find it necessary to do so in the next PRs.

Zhangxunmt (approved):
LGTM. If we find necessary we can move the MLTask to ml-common in the next PR.

PR #117, opened 2022-01-11
add ML Model get API [In progress]

Zhangxunmt:
### Description
This change added the support for Get Model API for ml-common plugin.
 
### Issues Resolved
Add Model-Get API.
 
### Check List
This is the first PR for the change, which is still subject to refine.


common/src/main/java/org/opensearch/ml/common/parameter/MLModelOutput.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.xcontent.XContentBuilder;
+
+import java.io.IOException;
+
+@Getter
+public class MLModelOutput extends MLOutput {
```
ylwu-amzn, line 12: We only need to add new `MLOutput` for new algorithm. For this get model API, we don't introduce new algorithm. So no need to add this class.

common/src/main/java/org/opensearch/ml/common/parameter/MLOutputType.java
```diff
@@ -16,5 +16,6 @@
     TRAINING,
     PREDICTION,
     SAMPLE_ALGO,
-    SAMPLE_CALCULATOR
+    SAMPLE_CALCULATOR,
+    MODEL
```
ylwu-amzn, line 20: No need to add model type

common/src/main/java/org/opensearch/ml/common/parameter/MLModelOutput.java
Zhangxunmt: Correct. I think we still need to include MLModel in the MLModelGetResponse. Based on the current structure, the Response is defined in the ml-common lib, which is a dependency to ml-plugin where the MLModel is defined. So MLModelGetResponse cannot find the MLMode. My thinking is that we can move MLModel into ml-common, so it needs some refactor.

ylwu-amzn: Sure, feel free to move `MLModel` to `commons` package.

common/src/main/java/org/opensearch/ml/common/parameter/MLModel.java
```diff
@@ -0,0 +1,141 @@
...
+        this.content = content;
+        this.user = user;
+    }
+
+    public MLModel(FunctionName algorithm, Model model) {
+        this(model.getName(), algorithm, model.getVersion(), Base64.getEncoder().encodeToString(model.getContent()), null);
+    }
+
+    public MLModel(StreamInput input) throws IOException{
+        name = input.readOptionalString();
```
ylwu-amzn, line 50: Keep the same order when read from `StreamInput` as write out `StreamOutput`. The `writeTo` method write `content` after `name`, so we should read `content` after `name`.

common/src/main/java/org/opensearch/ml/common/parameter/Model.java
```diff
@@ -12,5 +12,5 @@
 public class Model {
     String name;
     int version;
-    byte[] content;
+    public byte[] content;
```
ylwu-amzn, line 15: Why make `content` public?

common/src/main/java/org/opensearch/ml/common/transport/model/MLModelGetAction.java
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.model;
+
+import org.opensearch.action.ActionType;
+
+public class MLModelGetAction extends ActionType<MLModelGetResponse> {
+    public static final MLModelGetAction INSTANCE = new MLModelGetAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/models";
```
ylwu-amzn, line 12: Append "get" to action name? "cluster:admin/opensearch/ml/models/get"
We will have other model actions lile `delete`

plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java
```diff
@@ -0,0 +1,84 @@
...
+                } catch (Exception e) {
+                    log.error("Failed to parse ml model" + r.getId(), e);
+                    actionListener.onFailure(e);
+                }
+            } else {
+                actionListener.onResponse(null);
+            }
+        }, e -> {
+            if (e instanceof IndexNotFoundException) {
+                actionListener.onResponse(null);
```
ylwu-amzn, line 77: Throw `ResourceNotFoundException`?

plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java
```diff
@@ -0,0 +1,84 @@
...
+import org.opensearch.ml.common.transport.model.MLModelGetResponse;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportService;
+
+@Log4j2
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+public class GetModelTransportAction extends HandledTransportAction<ActionRequest, MLModelGetResponse> {
+
+    TransportService transportService;
+    private final Client client;
```
ylwu-amzn, line 38: As line 34 has define filed as private and final, I think we can remove `private final` here

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -48,11 +50,7 @@
 import org.opensearch.ml.engine.algorithms.sample.LocalSampleCalculator;
 import org.opensearch.ml.indices.MLIndicesHandler;
 import org.opensearch.ml.indices.MLInputDatasetHandler;
-import org.opensearch.ml.rest.RestMLExecuteAction;
-import org.opensearch.ml.rest.RestMLPredictionAction;
-import org.opensearch.ml.rest.RestMLTrainAndPredictAction;
-import org.opensearch.ml.rest.RestMLTrainingAction;
-import org.opensearch.ml.rest.RestStatsMLAction;
+import org.opensearch.ml.rest.*;
```
ylwu-amzn, line 53: Avoid wildcard import https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

common/src/main/java/org/opensearch/ml/common/parameter/MLModel.java
Zhangxunmt: Yeah will do in the revision.

common/src/main/java/org/opensearch/ml/common/parameter/Model.java
Zhangxunmt: Some unit tests in the plugin package directly use mode.content and intelliji auto detects them and set it to public. Have updated the UTs and removed the public.

common/src/main/java/org/opensearch/ml/common/transport/model/MLModelGetAction.java
Zhangxunmt: cool will do.

plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java
Zhangxunmt: sure will add them.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
Zhangxunmt: It has to be Intelliji replacing them to * or gradle.spotless task. I find this is a common problem using intellij.

plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java
Zhangxunmt: sure, will remove them.

common/src/main/java/org/opensearch/ml/common/parameter/MLModel.java
```diff
@@ -0,0 +1,141 @@
...
+                    version = parser.intValue();
+                    break;
+                case USER:
+                    user = User.parse(parser);
+                    break;
+                case ALGORITHM:
+                    algorithm = FunctionName.valueOf(parser.text());
+                    break;
+                default:
+                    break;
```
ylwu-amzn, line 125: Add `parser.skipChildren();` before line125 to skip any wrong field name. Refer to AD code https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/AnomalyDetector.java#L552

common/src/main/java/org/opensearch/ml/common/transport/model/MLModelGetRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public MLModelGetRequest(StreamInput in) throws IOException {
+        super(in);
+        this.modelId = in.readOptionalString();
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeOptionalString(this.modelId);
```
ylwu-amzn, line 47: As `modelId` is mandatory , we can just use `out.writeString`, same for line 41

ylwu-amzn:
We'd better put the revision in a new commit after the first round of review, so reviewers can just check the new changes easily.

PR #124, opened 2022-01-19
support train ML model in either sync or async way

ylwu-amzn:

### Description
Support training ML model in either sync or async way.
For example, PPL user should be able to train ML model in sync way; and user can use async way if the model training will be time consuming.

Main changes:
1. Support `async` URL parameter in train API. Add `?async=true` if need to train model in async way, by default will train model in sync way.
2. If train model in sync way, won't persist task in index.
3. For sync way, will return both model id and task id. For async way, will just return task id and user need to poll task to know its state/progress.
4. Use `GetRequest` to get model in predict task runner
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLTask.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLTask.java
@@ -46,10 +46,12 @@ public class MLTask implements ToXContentObject, Writeable {
     public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
     public static final String ERROR_FIELD = "error";
     public static final String USER_FIELD = "user";
+    public static final String IS_ASYNC_TASK_FIELD = "is_async";
 
     @Setter
     private String taskId;
-    private final String modelId;
+    @Setter
+    private String modelId;
     private final MLTaskType taskType;
     private final FunctionName functionName;
     @Setter
@@ -63,6 +65,7 @@ public class MLTask implements ToXContentObject, Writeable {
     @Setter
     private String error;
     private User user; // TODO: support document level access control later
+    private boolean async;
 
     @Builder
     public MLTask(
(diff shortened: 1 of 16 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLTask.java
```diff
@@ -63,6 +65,7 @@
     @Setter
     private String error;
     private User user; // TODO: support document level access control later
+    private boolean async;
 
     @Builder
     public MLTask(
```
Zhangxunmt, line 71: We can consider moving this class to the ml-common in the next PR.

plugin/src/main/java/org/opensearch/ml/task/MLTrainingTaskRunner.java
```diff
@@ -112,17 +113,39 @@ public void createMLTaskAndTrain(MLTrainingTaskRequest request, ActionListener<M
...
+            .async(request.isAsync())
             .build();
 
-        mlTaskManager.createMLTask(mlTask, ActionListener.wrap(r -> {
-            mlTask.setTaskId(r.getId());
+        if (request.isAsync()) {
+            mlTaskManager.createMLTask(mlTask, ActionListener.wrap(r -> {
+                String taskId = r.getId();
+                mlTask.setTaskId(taskId);
+                if (mlTask.isAsync()) {
```
Zhangxunmt, line 123: Is there any user case that the request is Sync, but the mlTask is Async? I am just wondering the difference between the two.

Zhangxunmt (approved):
Approved with a few questions in the comments.

plugin/src/main/java/org/opensearch/ml/model/MLTask.java
ylwu-amzn: Yes, will send out a separate refactoring PR

plugin/src/main/java/org/opensearch/ml/task/MLTrainingTaskRunner.java
ylwu-amzn: Async property of `MLTask` is consistent with request, check line 116 of this class. 
For async task, we will cache task in memory and persist task in index and it will return task id directly.
For sync task, we just cache task in memory. Will train model and return the model id.

PR #142, opened 2022-02-16
add circuit breaker

### LINKED ISSUE #141: Add circuit breaker
Add circuit breaker to protect cluster.

ylwu-amzn:

### Description
Add circuit breaker to protect cluster not overloaded by ML tasks.
 
### Issues Resolved
#141
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/breaker/BreakerName.java
+++ b/common/src/main/java/org/opensearch/ml/common/breaker/BreakerName.java
@@ -1,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.breaker;
+
+public enum BreakerName {
+    MEMORY
+}
--- a/common/src/main/java/org/opensearch/ml/common/breaker/MLCircuitBreakerService.java
+++ b/common/src/main/java/org/opensearch/ml/common/breaker/MLCircuitBreakerService.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.breaker;
+
+import lombok.extern.log4j.Log4j2;
+import org.opensearch.monitor.jvm.JvmService;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * This service registers internal system breakers and provide API for users to register their own breakers.
+ */
+@Log4j2
+public class MLCircuitBreakerService {
+
+    private final ConcurrentMap<BreakerName, CircuitBreaker> breakers = new ConcurrentHashMap<>();
+    private final JvmService jvmService;
+
+    /**
+     * Constructor.
+     *
+     * @param jvmService jvm info
+     */
+    public MLCircuitBreakerService(JvmService jvmService) {
+        this.jvmService = jvmService;
+    }
+
+    public void registerBreaker(BreakerName name, CircuitBreaker breaker) {
+        breakers.putIfAbsent(name, breaker);
+    }
+
+    public void unregisterBreaker(BreakerName name) {
+        if (name == null) {
+            return;
+        }
+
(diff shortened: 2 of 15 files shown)


common/src/main/java/org/opensearch/ml/common/breaker/MLCircuitBreakerService.java
```diff
@@ -0,0 +1,76 @@
...
+        breakers.putIfAbsent(name, breaker);
+    }
+
+    public void unregisterBreaker(BreakerName name) {
+        if (name == null) {
+            return;
+        }
+
+        breakers.remove(name);
+    }
```
Zhangxunmt, line 42: Wondering if a log can be added here just like the log message in registerBreaker().

common/src/main/java/org/opensearch/ml/common/breaker/BreakerName.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.breaker;
+
+public enum BreakerName {
+    MEMORY
```
Zhangxunmt, line 9: May I ask if any breakers based on "CPU", "garbage collection" or "Disk usage" will be added?

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
```diff
@@ -70,5 +79,12 @@ protected void handleMLTaskComplete(MLTask mlTask) {
...
 
-    public abstract void run(Request request, TransportService transportService, ActionListener<Response> listener);
+    public void run(Request request, TransportService transportService, ActionListener<Response> listener) {
+        if (mlCircuitBreakerService.isOpen()) {
+            throw new MLLimitExceededException("Circuit breaker is open");
+        }
+        executeTask(request, transportService, listener);
+    }
+
+    public abstract void executeTask(Request request, TransportService transportService, ActionListener<Response> listener);
```
Zhangxunmt, line 89: This circuit breaker should be only applied to train/execute tasks that require many resources right? I assume the other API tasks wouldn't need it since they only run a light query to the cluster? or maybe the cluster has that circuit breaker implemented already.

Zhangxunmt (approved):
Approved with several question/comments. Some other questions: 1. this would need us to run a load test to exhaust the nodes to test the breakers. 2. what is the strategy to overcome the problem when the service is over loaded? Does the customer have to scale up cluster to resolve it?

common/src/main/java/org/opensearch/ml/common/breaker/BreakerName.java
ylwu-amzn: Yeah, that's possible. We can add more types later.

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
ylwu-amzn: Correct. The train/predict task will consume more resource, so we need to check circuit breaker.

ylwu-amzn:
> Approved with several question/comments. Some other questions:
> 
> 1. this would need us to run a load test to exhaust the nodes to test the breakers.
> 2. what is the strategy to overcome the problem when the service is over loaded? Does the customer have to scale up cluster to resolve it?
> 3. Should we sent memory/CPU/Disks usage metrics to somewhere (Pmet or cloudwatch) to monitor the services? Maybe we just rely on the EC2 metrics.


1. Prefer to test this part in UT. Will add memory threshold cluster setting to make IT test easier later. 
2. If overloaded, customer can scale up cluster or run less train/predict tasks. Will make `MAX_ML_TASK_PER_NODE` as cluster setting as well, so user can tune it to run less less tasks per node.
3. Correct, we just rely on node level metrics.

common/src/main/java/org/opensearch/ml/common/breaker/MLCircuitBreakerService.java
ylwu-amzn: Sure, will add in next PR.

PR #144, opened 2022-02-17
rename tribuo AD algorithm name

ylwu-amzn:

### Description
Rename tribuo AD algorithm name.

 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/parameter/FunctionName.java
+++ b/common/src/main/java/org/opensearch/ml/common/parameter/FunctionName.java
@@ -8,7 +8,7 @@ package org.opensearch.ml.common.parameter;
 public enum FunctionName {
     LINEAR_REGRESSION,
     KMEANS,
-    ANOMALY_DETECTION,
+    AD_LIBSVM,
     SAMPLE_ALGO,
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/opensearch/ml/common/parameter/FunctionName.java
```diff
@@ -8,7 +8,7 @@
 public enum FunctionName {
     LINEAR_REGRESSION,
     KMEANS,
-    ANOMALY_DETECTION,
+    AD_LIBSVM,
```
Zhangxunmt, line 11: I may lack the context - what is the difference between these two names? They represent the same AD algorithm?

Zhangxunmt (approved):
LGTM

common/src/main/java/org/opensearch/ml/common/parameter/FunctionName.java
ylwu-amzn: That's to avoid confusion. Tribuo using libSVM to do anomaly detection. The new name is more accurate. So user can distinguish different AD algorithms.

PR #168, opened 2022-03-08
add security IT

ylwu-amzn:

### Description
Add security IT.
`./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Duser=admin -Dpassword=admin --tests="org.opensearch.ml.rest.SecureMLRestIT"`
 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.rest;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.apache.http.HttpHost;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.rules.ExpectedException;
+import org.opensearch.client.ResponseException;
+import org.opensearch.client.RestClient;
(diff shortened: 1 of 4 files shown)


plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
```diff
@@ -0,0 +1,145 @@
...
+        createRoleMapping(indexSearchAccessRole, new ArrayList<>(Arrays.asList(mlFullAccessUser)));
+
+        ingestIrisData(irisIndex);
+        searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.query(new MatchAllQueryBuilder());
+        searchSourceBuilder.size(1000);
+        searchSourceBuilder.fetchSource(new String[] { "petal_length_in_cm", "petal_width_in_cm" }, null);
+    }
+
+    @After
```
Zhangxunmt, line 94: Just realized this annotation exists. Should we add this to clean up all the data in each other integration tests? That might fix the flaky request count problem.

Zhangxunmt (approved):
Do we want to add one more test for accessing those ml system indices like model and task?

ylwu-amzn:
> 

Yes, will add more test cased in next PRs.

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
ylwu-amzn: For the same cluster, the total request count will not decrease, unless we restart the cluster. The total request is some value cached in memory not in index. So cleanup data won't help

Issue #162, opened 2022-03-04
[BUG] Enhance get task id by checking task id and don't return task id


ylwu-amzn:
If user passed invalid task id, it won't be checked and task id returned in error message which may be injected to cause security flaw.

Suggestion:
1. Use regex to check task id first.
2. Don't include task id in error message, just return general error message like "Can't find task"

Example:
`GET /_plugins/_ml/tasks/wwRRLX8BydmmU1x6I-AIewa1q%3cimg%20src%3da%20onerror%3dalert(1)%3e`

```
{
  "error": {
    "root_cause": [
      {
        "type": "m_l_resource_not_found_exception",
        "reason": "Fail to find task wwRRLX8BydmmU1x6I-AIewa1q<img src=a onerror=alert(1)>m4flx"
      }
    ],
    "type": "m_l_resource_not_found_exception",
    "reason": "Fail to find task wwRRLX8BydmmU1x6I-AIewa1q   >"
  },
  "status": 500
}
```

ylwu-amzn:
Also need to check other APIs like get model.

PR #169, opened 2022-03-08
Do not return task and model id. check id with base64url.

Zhangxunmt:

### Description
This is to fix the error in https://github.com/opensearch-project/ml-commons/issues/162. Do not return model and task id in the APIs. Also check if the input model/task id is a valid Base64url encoded string. https://github.com/elastic/elasticsearch/blob/99f88f15c5febbca2d13b5b5fda27b844153bf1a/server/src/main/java/org/elasticsearch/common/TimeBasedUUIDGenerator.java#L80-L118 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -45,6 +48,8 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
         DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_INDEX, modelId);
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            Base64.getUrlDecoder().decode(modelId);
```
ylwu-amzn, line 51: Can you add comment: the OpenSearch core code to generate ID with `Base64.getUrlEncoder` ?

ylwu-amzn: One general suggestion is don't bind tightly with OpenSearch core code to avoid unnecessary effort if OpenSearch code changes in future.

ylwu-amzn: I think the simple check is ok, and ok if we don't check at all, the search request will always fail if pass invalid mode/task id.

PR #177, opened 2022-03-09
fix no permission to create model/task index bug;add security IT for train/predict API

ylwu-amzn:

### Description
- Fix no permission to create model/task index bug. ML model and task index are system index. Need to stash thread context to create.
- Add security IT for train/predict API
 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -12,8 +12,10 @@ import lombok.extern.log4j.Log4j2;
 
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
+import org.opensearch.action.admin.indices.create.CreateIndexResponse;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.common.xcontent.XContentType;
 
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
@@ -87,19 +89,24 @@ public class MLIndicesHandler {
 
     public void initMLIndexIfAbsent(String indexName, String mapping, ActionListener<Boolean> listener) {
         if (!clusterService.state().metadata().hasIndex(indexName)) {
-            CreateIndexRequest request = new CreateIndexRequest(indexName).mapping("_doc", mapping, XContentType.JSON);
-
-            client.admin().indices().create(request, ActionListener.wrap(r -> {
-                if (r.isAcknowledged()) {
-                    log.info("create index:{}", indexName);
-                    listener.onResponse(true);
-                } else {
-                    listener.onResponse(false);
-                }
-            }, e -> {
-                log.error("Failed to create index " + indexName, e);
+            try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
+                ActionListener<CreateIndexResponse> actionListener = ActionListener.wrap(r -> {
+                    if (r.isAcknowledged()) {
+                        log.info("create index:{}", indexName);
+                        listener.onResponse(true);
+                    } else {
+                        listener.onResponse(false);
+                    }
+                }, e -> {
+                    log.error("Failed to create index " + indexName, e);
+                    listener.onFailure(e);
+                });
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping("_doc", mapping, XContentType.JSON);
+                client.admin().indices().create(request, ActionListener.runBefore(actionListener, () -> threadContext.restore()));
+            } catch (Exception e) {
+                log.error("Failed to init index " + indexName, e);
                 listener.onFailure(e);
-            }));
+            }
         } else {
             log.info("index:{} is already created", indexName);
             listener.onResponse(true);
--- a/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
@@ -260,10 +260,10 @@ public abstract class MLCommonsRestTestCase extends OpenSearchRestTestCase {
     protected void validateStats(
         FunctionName functionName,
         ActionName actionName,
-        int expectedTotalFailureCount,
-        int expectedTotalAlgoFailureCount,
-        int expectedMinumnTotalRequestCount,
-        int expectedTotalAlgoRequestCount
+        int expectedMinimumTotalFailureCount,
+        int expectedMinimumTotalAlgoFailureCount,
+        int expectedMinimumTotalRequestCount,
+        int expectedMinimumTotalAlgoRequestCount
     ) throws IOException {
         Response statsResponse = TestHelper.makeRequest(client(), "GET", "_plugins/_ml/stats", null, "", null);
         HttpEntity entity = statsResponse.getEntity();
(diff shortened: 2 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -87,19 +89,24 @@ public void initMLTaskIndex(ActionListener<Boolean> listener) {
...
+                        listener.onResponse(true);
+                    } else {
+                        listener.onResponse(false);
+                    }
+                }, e -> {
+                    log.error("Failed to create index " + indexName, e);
+                    listener.onFailure(e);
+                });
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping("_doc", mapping, XContentType.JSON);
+                client.admin().indices().create(request, ActionListener.runBefore(actionListener, () -> threadContext.restore()));
```
amitgalitz, line 105: If we are running before creation `threadContext.restore()` doesn't that mean we restore original thread context and user info is restored so we can't create system index or is this fine cause the CreateIndexRequest itself is one line before?

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
```diff
@@ -464,4 +464,67 @@ public void trainAndPredict(
...
+        Consumer<Map<String, Object>> function,
+        boolean async
+    ) throws IOException {
+        MLInputDataset inputData = SearchQueryInputDataset
+            .builder()
+            .indices(ImmutableList.of(indexName))
+            .searchSourceBuilder(searchSourceBuilder)
+            .build();
+        MLInput kmeansInput = MLInput.builder().algorithm(functionName).parameters(params).inputDataset(inputData).build();
+        String endpoint = "/_plugins/_ml/_train/" + functionName.name().toLowerCase(Locale.ROOT);
```
amitgalitz, line 483: nit: Should this be a constant, same for predict endpoint

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
```diff
@@ -142,4 +144,54 @@ public void testTrainAndPredictWithFullAccess() throws IOException {
...
+                assertNull(e);
+            }
+            try {
+                // predict with trained model
+                predict(mlFullAccessClient, FunctionName.KMEANS, modelId, irisIndex, kMeansParams, searchSourceBuilder, predictResult -> {
+                    String predictStatus = (String) predictResult.get("status");
+                    assertEquals(MLTaskState.COMPLETED.name(), predictStatus);
+                    Map<String, Object> predictionResult = (Map<String, Object>) predictResult.get("prediction_result");
+                    ArrayList rows = (ArrayList) predictionResult.get("rows");
+                    assertTrue(rows.size() > 1);
```
amitgalitz, line 171: Can there be any more specific check than result just being existent?

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
ylwu-amzn: The execution order of the code
1. create admin `client.admin().indices().create`
2. restore thread context `threadContext.restore()` 
3. run actionListener

So restore original context happens after index created.

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
ylwu-amzn: Prefer to keep it as is now as we are quite close to release. We can refactor code later. Created an issue to track https://github.com/opensearch-project/ml-commons/issues/178

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
ylwu-amzn: Will be hard to check, the algorithm has some built-in randomness, the result could vary. Prefer to keep the verification simple here to avoid flaky tests. And here our main goal is to test the security part. Should be ok to verify there is result generated which means the user has permission and can run the action successfully.

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
```diff
@@ -194,4 +210,58 @@ public void testTrainModelInAsyncWayWithFullAccess() throws IOException {
...
+                getTask(mlFullAccessClient, taskId, task -> {
+                    String algorithm = (String) task.get("function_name");
+                    assertEquals(FunctionName.KMEANS.name(), algorithm);
+                });
+            } catch (IOException e) {
+                assertNull(e);
+            }
+            try {
+                // Failed to delete task with read only client
+                deleteTask(mlReadOnlyClient, taskId, null);
```
amitgalitz, line 260: Could this be risky test format since if no exception is thrown then it won't go into catch block and test will still pass. same for above delete model test

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
amitgalitz: Okay that makes sense not to do it here and now, maybe in the future some tests can be added similar to AD with some sort of minimum precision in a different place and maybe for other algorithms (will have to approach each algo differently), but also as we know such tests is asking for more flakiness 😆

ylwu-amzn: Yes, we should avoid over-engineering, that may bring unexpected extra load to us.

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
ylwu-amzn: Good catch, will add throwing exception after delete model/task. So if readonly user can delete model/task successfully, we will throw exception and that will make test fail.

amitgalitz:
Build failed on some tests that weren't changed on this PR, could be flaky ones

amitgalitz (approved):

ylwu-amzn:
> 

Yes, seems flaky test, created an issue to track this https://github.com/opensearch-project/ml-commons/issues/179

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
```diff
@@ -464,4 +464,97 @@ public void trainAndPredict(
...
+        String query = String.format(Locale.ROOT, "{\"query\":{\"bool\":{\"filter\":[{\"term\":{\"function_name\":\"%s\"}}]}}}", algoName);
+        searchTasks(client, query, function);
+    }
+
+    public void searchTasks(RestClient client, String query, Consumer<Map<String, Object>> function) throws IOException {
+        Response response = TestHelper.makeRequest(client, "GET", "/_plugins/_ml/tasks/_search", null, query, null);
+        verifyResponse(function, response);
+    }
+
+    private void verifyResponse(Consumer<Map<String, Object>> function, Response response) throws IOException {
```
Zhangxunmt, line 551: This is a very useful function reducing a lot of redundancy.

Zhangxunmt: Nitpick: The name of this Consumer is "function" which leads people to think of it as a Function interface. Maybe we can just name is as consumer?

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
```diff
@@ -142,4 +160,151 @@ public void testTrainAndPredictWithFullAccess() throws IOException {
...
+            String taskId = (String) trainResult.get("task_id");
+            assertNotNull(taskId);
+            String status = (String) trainResult.get("status");
+            assertEquals(MLTaskState.CREATED.name(), status);
+            try {
+                getTask(mlFullAccessClient, taskId, task -> {
+                    String algorithm = (String) task.get("function_name");
+                    assertEquals(FunctionName.KMEANS.name(), algorithm);
+                });
+            } catch (IOException e) {
```
Zhangxunmt, line 207: Is it possible that a different Exception be thrown here? Which would make the test fail.

Zhangxunmt (approved):

plugin/src/test/java/org/opensearch/ml/rest/SecureMLRestIT.java
ylwu-amzn: If the exception is not IOException, we expect the test to fail. So it's ok.

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
ylwu-amzn: will rename in next pr

PR #192, opened 2022-03-10
add more UT and IT for rest actions

Zhangxunmt:

### Description
Do not exclude tests for restful actions and ensure coverage is 80%. Added IT for each Model/Task API. More tests can be added in future PRs if necessary. 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -218,9 +218,7 @@ List<String> jacocoExclusions = [
         'org.opensearch.ml.task.MLTaskCache',
         'org.opensearch.ml.task.MLTaskManager',
         'org.opensearch.ml.task.MLTrainAndPredictTaskRunner',
-        'org.opensearch.ml.rest.RestMLGetModelAction',
-        'org.opensearch.ml.rest.RestMLDeleteModelAction',
-        'org.opensearch.ml.rest.*'
+        'org.opensearch.ml.rest.AbstractMLSearchAction*'
 ]
 
 jacocoTestCoverageVerification {
--- a/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
@@ -25,6 +25,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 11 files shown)


plugin/build.gradle
```diff
@@ -218,9 +218,7 @@ List<String> jacocoExclusions = [
         'org.opensearch.ml.task.MLTaskCache',
         'org.opensearch.ml.task.MLTaskManager',
         'org.opensearch.ml.task.MLTrainAndPredictTaskRunner',
-        'org.opensearch.ml.rest.RestMLGetModelAction',
-        'org.opensearch.ml.rest.RestMLDeleteModelAction',
-        'org.opensearch.ml.rest.*'
+        'org.opensearch.ml.rest.AbstractMLSearchAction.1'
```
ylwu-amzn, line 221: Why add ".1"?

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
```diff
@@ -305,6 +305,16 @@ protected Response ingestModelData() throws IOException {
...
     }
 
+    public void trainAsyncWithSample(Consumer<Map<String, Object>> consumer, boolean async) throws IOException {
+        String endpoint = "/_plugins/_ml/_train/sample_algo";
+        if (async) {
+            endpoint += "?async=true";
+        }
+        Response response = TestHelper
+            .makeRequest(client(), "POST", endpoint, ImmutableMap.of(), TestHelper.toHttpEntity(trainModelDataJson()), null);
+        verifyResponse(consumer, response);
```
ylwu-amzn, line 315: How about we sleep for few seconds to wait for the async task finish? I think 5 seconds should be enough.

plugin/build.gradle
Zhangxunmt: If no this line, the error would be: 
> Rule violated for class org.opensearch.ml.rest.AbstractMLSearchAction.1: branches covered ratio is 0.0, but expected minimum is 0.8
  Rule violated for class org.opensearch.ml.rest.AbstractMLSearchAction.1: lines covered ratio is 0.0, but expected minimum is 0.3

I am checking where is that .1 coming from.

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
Zhangxunmt: Good idea. We should be careful adding more and more ITs. I think we don't want to over test. The more IT we add, the more flaky it would be. I noticed some ITs do interfere each other.

plugin/build.gradle
Zhangxunmt: It comes from the inner anonymous class defined in BaseHandler. There's a $1 class file generated from the build. I have replaced it with * to not confuse people.

ylwu-amzn (approved):
LGTM. Thanks for the change

PR #226, opened 2022-03-16
add more IT for train/predict and train&predict action

ylwu-amzn:

### Description
add more IT for train/predict and train&predict action
 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -10,7 +10,6 @@ plugins {
 }
 
 repositories {
-    mavenLocal()
     mavenCentral()
 }
 
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -10,6 +10,7 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
     mavenCentral()
 }
 
--- a/plugin/src/test/java/org/opensearch/ml/action/prediction/PredictionITTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/prediction/PredictionITTests.java
@@ -5,107 +5,84 @@
 
 package org.opensearch.ml.action.prediction;
 
-import static org.opensearch.ml.utils.IntegTestUtils.DATA_FRAME_INPUT_DATASET;
-import static org.opensearch.ml.utils.IntegTestUtils.TESTING_DATA;
-import static org.opensearch.ml.utils.IntegTestUtils.TESTING_INDEX_NAME;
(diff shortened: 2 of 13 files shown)


ml-algorithms/build.gradle
```diff
@@ -10,6 +10,7 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
```
Zhangxunmt, line 13: Is there something required to pull from the local or just make the process faster with local?

plugin/src/test/java/org/opensearch/ml/action/prediction/PredictionITTests.java
```diff
@@ -5,107 +5,84 @@
...
+import com.google.common.collect.ImmutableList;
 
-    public void testTestingData() throws ExecutionException, InterruptedException {
-        verifyGeneratedTestingData(TESTING_DATA);
-        waitModelAvailable(taskId);
-    }
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 2)
+public class PredictionITTests extends MLCommonsIntegTestCase {
+    private String irisIndexName;
+    private String modelId;
```
Zhangxunmt, line 34: I think these IT test should be in lieu of the UTs? though they don't show up in the jacoco test reports.

Zhangxunmt (approved):
Let's leave a comments in the build.gradle jacocoExclusions saying that the exclusives in the list are tests through IT tests ?

ml-algorithms/build.gradle
ylwu-amzn: Ignore this part, this is for debugging the tribuo bug https://github.com/oracle/tribuo/pull/224. Some tests will fail. Will ignore the flaky part first to pass CI.

plugin/src/test/java/org/opensearch/ml/action/prediction/PredictionITTests.java
ylwu-amzn: Let's keep both UT and IT. Jacoco may not cover the REST API integration test.

ylwu-amzn:
> 

I think we'd better not say that. We should add more test to meet the jacoco coverage bar. It will be more clear and easy to work with community. Otherwise, if community contributor says their code covered by REST IT tests, it's hard to verify.

Zhangxunmt (approved):
LGTM

PR #232, opened 2022-03-18
fix execute request/response serialization/deserialization

ylwu-amzn:

### Description
Fix execute request/response serialization/deserialization;
Update how to add new function doc.
 
 
### Check List
  - [x] All tests pass


common/src/main/java/org/opensearch/ml/common/parameter/AnomalyLocalizationOutput.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.ml.engine.algorithms.anomalylocalization;
+package org.opensearch.ml.common.parameter;
```
Zhangxunmt, line 6: I have been wondering this ml.common.parameter directory is bloating quickly and it already includes so many files. Should we considering using subdirectories like model, task, <algorithmA>, <algorithmB>, etc?

ylwu-amzn: Yes, that's the plan. The refactoring may impact quite a lot places. Let's finish the testing first, then move these classes to different packages.

common/src/main/java/org/opensearch/ml/common/transport/execute/MLExecuteTaskRequest.java
```diff
@@ -31,21 +33,25 @@
 @ToString
 public class MLExecuteTaskRequest extends ActionRequest {
 
+    FunctionName functionName;
```
Zhangxunmt, line 36: Is this FunctionName field the key part that triggered the bug? I see that this triggered most of the required changes in the PR.

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/transport/execute/MLExecuteTaskRequest.java
ylwu-amzn: Yes, we need `functionName` to tell which class to initialize for this algorithm.

PR #279, opened 2022-04-13
support dispatching execute task; don't dispatch ML task again

ylwu-amzn:

### Description
1. support dispatching execute task;
2. don't dispatch ML task again
3. add request id to track the request easily
 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -207,7 +207,8 @@ jacocoTestReport {
 List<String> jacocoExclusions = [
         // TODO: add more unit test to meet the minimal test coverage.
         'org.opensearch.ml.constant.CommonValue',
-        'org.opensearch.ml.plugin.MachineLearningPlugin*'
+        'org.opensearch.ml.plugin.MachineLearningPlugin*',
+        'org.opensearch.ml.task.MLPredictTaskRunner'
 ]
 
 jacocoTestCoverageVerification {
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -207,8 +207,7 @@ jacocoTestReport {
 List<String> jacocoExclusions = [
         // TODO: add more unit test to meet the minimal test coverage.
         'org.opensearch.ml.constant.CommonValue',
-        'org.opensearch.ml.plugin.MachineLearningPlugin*',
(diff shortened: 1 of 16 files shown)


plugin/build.gradle
```diff
@@ -207,7 +207,8 @@ jacocoTestReport {
 List<String> jacocoExclusions = [
         // TODO: add more unit test to meet the minimal test coverage.
         'org.opensearch.ml.constant.CommonValue',
-        'org.opensearch.ml.plugin.MachineLearningPlugin*'
+        'org.opensearch.ml.plugin.MachineLearningPlugin*',
+        'org.opensearch.ml.task.MLPredictTaskRunner'
```
Zhangxunmt, line 211: We might want to refactor this PredictTaskRunner. It has a big body in predict() method which is not test friendly.

Zhangxunmt, line 211: We might want to refactor this PredictTaskRunner. It has a big body in predict() method which is not test friendly.

Zhangxunmt (approved):

plugin/build.gradle
ylwu-amzn: Yes, good point, we can refactor later

Zhangxunmt (approved):

plugin/src/test/java/org/opensearch/ml/task/TaskRunnerTests.java
```diff
@@ -46,6 +49,8 @@ public class TaskRunnerTests extends OpenSearchTestCase {
     MLTaskDispatcher mlTaskDispatcher;
     @Mock
     MLCircuitBreakerService mlCircuitBreakerService;
+    @Mock
```
jackiehanyang, line 52: Curiosity question: this `CuslterService mock has to be used together with extending `OpenSearchTestCase` right?

jackiehanyang (approved):

plugin/src/test/java/org/opensearch/ml/task/TaskRunnerTests.java
jackiehanyang: Answering my own question here: no it doesn't have to

Issue #234, opened 2022-03-18
[BUG] Anomaly localization will throw NullPointerException when index not exist


ylwu-amzn:
**Describe the bug**
Anomaly localization will throw `NullPointerException` when index not exist.

**To Reproduce**
1. Git clone `main` branch to local, then run `./gradlew run`
2. Send this request
```
POST /_plugins/_ml/_execute/anomaly_localization
{
  "index_name": "rca-index",
  "attribute_field_names": [
    "attribute"
  ],
  "aggregations": [
    {
      "sum": {
        "sum": {
          "field": "value"
        }
      }
    }
  ],
  "time_field_name": "timestamp",
  "start_time": 1620630000000,
  "end_time": 1621234800000,
  "min_time_interval": 86400000,
  "num_outputs": 10
}
```

**Expected behavior**
Should return index not found error message rather than `NullPointerException`

ylwu-amzn:
This bug fixed in PR #280

PR #280, opened 2022-04-13
Fix NPE in anomaly localization

jackiehanyang:
Signed-off-by: jackiehanyang <jkhanjob@gmail.com>

### Description
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/234
### Check List
  - [x] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImpl.java
@@ -26,9 +26,9 @@ import org.opensearch.action.search.MultiSearchRequest;
 import org.opensearch.action.search.MultiSearchResponse;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.client.Client;
-import org.opensearch.cluster.ClusterState;
-import org.opensearch.cluster.metadata.Metadata;
+import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.index.IndexNotFoundException;
@@ -78,6 +78,7 @@ public class AnomalyLocalizerImpl implements AnomalyLocalizer, Executable {
     private final Client client;
     private final Settings settings;
     private final ClusterService clusterService;
+    private final IndexNameExpressionResolver indexNameExpressionResolver;
 
     /**
      * Constructor.
@@ -88,10 +89,12 @@ public class AnomalyLocalizerImpl implements AnomalyLocalizer, Executable {
     public AnomalyLocalizerImpl(
             Client client,
             Settings settings,
-            ClusterService clusterService) {
+            ClusterService clusterService,
+            IndexNameExpressionResolver indexNameExpressionResolver) {
         this.client = client;
         this.settings = settings;
         this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
     }
 
     /**
@@ -133,7 +136,7 @@ public class AnomalyLocalizerImpl implements AnomalyLocalizer, Executable {
         AnomalyLocalizationOutput.Result result = new AnomalyLocalizationOutput.Result();
         List<Map.Entry<Long, Long>> intervals = timeBuckets.getAllIntervals();
 
-        if (clusterService.state().metadata().hasIndex(input.getIndexName())) {
+        if (isIndexExist(input.getIndexName(), listener)) {
             for (int i = 0; i < intervals.size(); i++) {
                 double value = getDoubleValue((SingleValue) response.getResponses()[i].getResponse().getAggregations().get(agg.getName()));
 
@@ -149,6 +152,20 @@ public class AnomalyLocalizerImpl implements AnomalyLocalizer, Executable {
             log.info("index: {} does not exist", input.getIndexName());
             listener.onFailure(new IndexNotFoundException("Failed to find index: " + input.getIndexName()));
         }
+    }
+    
+    protected boolean isIndexExist(String indexName, ActionListener<AnomalyLocalizationOutput> listener) {
+        String[] concreteIndices = indexNameExpressionResolver.concreteIndexNames(clusterService.state(),
+                IndicesOptions.lenientExpandOpen(), indexName);
+        if (concreteIndices == null || concreteIndices.length == 0) {
+            return false;
+        }
+        for(String index : concreteIndices) {
+            if (!clusterService.state().metadata().hasIndex(index)) {
+                return false;
+            }
+        }
+        return true;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImplTests.java
```diff
@@ -21,6 +21,7 @@
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.client.Client;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.index.IndexNotFoundException;
```
Zhangxunmt, line 24: I don't see this "IndexNotFoundException" is used in the test. Why add it?

Zhangxunmt, line 24: I don't see this "IndexNotFoundException" is used in the test. Why add it?

Zhangxunmt (approved):

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImplTests.java
```diff
@@ -234,6 +235,15 @@ public void testGetLocalizedResultsForInvalidTimeRange() {
         anomalyLocalizer.getLocalizationResults(input, outputListener);
     }
 
+    @Test(expected = RuntimeException.class)
```
ylwu-amzn, line 238: Why expect `RuntimeException`  rather `IndexNotFoundException` ?
You can use `ExpectedException` to assert the exception type and error message, refer to https://github.com/opensearch-project/ml-commons/blob/main/common/src/test/java/org/opensearch/ml/common/dataframe/ColumnValueTest.java#L17

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImpl.java
```diff
@@ -126,6 +127,10 @@ private void onOverallAggregatesResponse(MultiSearchResponse response, AnomalyLo
         AnomalyLocalizationOutput.Result result = new AnomalyLocalizationOutput.Result();
         List<Map.Entry<Long, Long>> intervals = timeBuckets.getAllIntervals();
         for (int i = 0; i < intervals.size(); i++) {
+            if (response.getResponses()[i].getFailure() instanceof IndexNotFoundException) {
```
ylwu-amzn, line 130: Suggest check if index exists or not like this https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java#L90

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImplTests.java
jackiehanyang: Yeah was originally checking the specific type of exception in the test, then found out that this test class and the test class you mentioned are all checking `Runtime Exception.class`, so continuing using the same approach. 

Will change it to check specific exception type

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImpl.java
```diff
@@ -125,18 +132,23 @@ private void onOverallAggregatesResponse(MultiSearchResponse response, AnomalyLo
...
-            AnomalyLocalizationOutput.Bucket bucket = new AnomalyLocalizationOutput.Bucket();
-            bucket.setStartTime(intervals.get(i).getKey());
-            bucket.setEndTime(intervals.get(i).getValue());
-            bucket.setOverallAggValue(value);
-            result.getBuckets().add(bucket);
-        }
 
-        output.getResults().put(agg.getName(), result);
-        getLocalizedEntities(input, agg, result, output, listener);
+        if (clusterService.state().metadata().hasIndex(input.getIndexName())) {
```
ylwu-amzn, line 136: This seems can't support index pattern like "servicelog*". You can refer to AD code to get concrete indices https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/util/IndexUtils.java#L143
```
        // the original index might be an index expression with wildcards like "log*",
        // so we need to expand the expression to concrete index name
        String[] concreteIndices = indexNameExpressionResolver.concreteIndexNames(state, IndicesOptions.lenientExpandOpen(), indices);
```

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImplTests.java
```diff
@@ -200,8 +233,14 @@ public void setup() {
...
     }
 
+    @AfterClass
+    public static void resetPortCounter() {
+        portGenerator.set(0);
+    }
+
     @Test
     public void testGetLocalizedResultsGivenNoAnomaly() {
+        when(clusterService.state()).thenReturn(testState);
```
Zhangxunmt, line 243: Did you just add this line to resolve the "null" problem?

Zhangxunmt (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImpl.java
```diff
@@ -150,6 +153,20 @@ private void onOverallAggregatesResponse(MultiSearchResponse response, AnomalyLo
...
     }
+    
+    protected boolean isIndexExist(String indexName, ActionListener<AnomalyLocalizationOutput> listener) {
+        String[] concreteIndices = indexNameExpressionResolver.concreteIndexNames(clusterService.state(),
+                IndicesOptions.lenientExpandOpen(), indexName);
+        if (concreteIndices == null || concreteIndices.length == 0) {
+            return false;
+        }
+        for(String index : concreteIndices) {
+            if (!clusterService.state().metadata().hasIndex(index)) {
```
ylwu-amzn, line 164: If "logs*" pattern has 3 concrete index "logs1", "logs2" and "logs3", if "log1" doesn't exist but "log2" and "log3" exist, we should return true as we still have 2 valid index exist. But this code seems will return false unless all three indexes exist.

jackiehanyang: ok, I was confused about this use case. Updated the logic to return true if any valid index exist.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/anomalylocalization/AnomalyLocalizerImplTests.java
jackiehanyang: yes, I was confused before about what should the mocked ClusterService return. I thought it should return a mocked local node, but turns out it's true

ylwu-amzn (approved):

Zhangxunmt (approved):

PR #284, opened 2022-04-18
create model/task index with correct mapping

ylwu-amzn:

### Description
We found security test `testReadOnlyUser_CanSearchModels` failed. Debug locally and found ML model index mapping is wrong: filed `algorithm` was set as `text` type
So when search with this query, will return empty result.
```
POST /_plugins/_ml/models/_search
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "algorithm": "KMEANS"
          }
        }
      ]
    }
  }
}
```

This PR fix this bug by setting mapping in create index request. Will backport this to 1.x .
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -100,7 +100,7 @@ public class MLIndicesHandler {
                     log.error("Failed to create index " + indexName, e);
                     listener.onFailure(e);
                 });
-                CreateIndexRequest request = new CreateIndexRequest(indexName);
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping(mapping);
                 client.admin().indices().create(request, ActionListener.runBefore(actionListener, () -> threadContext.restore()));
             } catch (Exception e) {
                 log.error("Failed to init index " + indexName, e);


Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -100,7 +100,7 @@ public void initMLIndexIfAbsent(String indexName, String mapping, ActionListener
                     log.error("Failed to create index " + indexName, e);
                     listener.onFailure(e);
                 });
-                CreateIndexRequest request = new CreateIndexRequest(indexName);
+                CreateIndexRequest request = new CreateIndexRequest(indexName).mapping(mapping);
```
jackiehanyang, line 103: Can we add a unit test for this? It can be added later if this needs to merge soon in order to catch 2.0

jackiehanyang (approved):

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
ylwu-amzn: yes, this needs to merge ASAP. Will add more test later

PR #346, opened 2022-06-15
dispatch ML task to ML node first

### LINKED ISSUE #79: Support dedicated ml node
We released ml-commons plugin in OpenSearch 1.3. It supports training model and predicting. ML model generally consuming more resources, especially for training process. The community wants to support bigger ML models which might require more resources and special hardware like GPU. 

As OpenSearch doesn’t support ML node, we dispatch ML task to data node only. That means if user wants to train a large model, they need to scale up all data nodes which can be costly. And ML tasks will use shared resources on data nodes which may impact the core searching/indexing function.

Support a dedicated ML node, users don’t need to scale up their data node at all. Instead just configure a new ML node (with different settings, more powerful instance type) and add it to cluster via the YAML file (requires a cluster restart). By doing so, users can separate resource usage better by running ML task on dedicated node which can reduce impact to other critical tasks like search/ingestion.

OpenSearch core will check node role when start node. If role is not built-in roles like `data` role, it will throw exception and node can't start. To support dedicated ML node, we have to remove this limitation in OpenSearch core. That is done with this PR which supports dynamic node role in OpenSearch https://github.com/opensearch-project/OpenSearch/pull/3436. 

With that we can enhance ml-commons code to dispatch task to `ml` nodes first. If no `ml` nodes we can fall back to data nodes.

[Original Proposal](https://github.com/opensearch-project/OpenSearch/issues/2877)

ylwu-amzn:

### Description
We plan to support dynamic role in OpenSearch. Refer to this PR https://github.com/opensearch-project/OpenSearch/pull/3436 and issue https://github.com/opensearch-project/OpenSearch/issues/2877

This PR is to change the task dispatcher to support ML node. Will check if cluster has any node with "ml" role first. If yes, will dispatch ML task to "ml" nodes; otherwise will dispatch to data nodes.

 
### Issues Resolved
close #79
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
+++ b/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
@@ -19,13 +19,11 @@ import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionResponse;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
-import org.opensearch.cluster.node.DiscoveryNodeRole;
 import org.opensearch.cluster.node.DiscoveryNodes;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.common.settings.ClusterSettings;
 import org.opensearch.common.settings.IndexScopedSettings;
-import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.settings.SettingsFilter;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
@@ -115,14 +113,7 @@ public class MachineLearningPlugin extends Plugin implements ActionPlugin {
     private ClusterService clusterService;
     private ThreadPool threadPool;
 
-    public static final Setting<Boolean> IS_ML_NODE_SETTING = Setting.boolSetting("node.ml", false, Setting.Property.NodeScope);
-
-    public static final DiscoveryNodeRole ML_ROLE = new DiscoveryNodeRole("ml", "l") {
-        @Override
-        public Setting<Boolean> legacySetting() {
-            return IS_ML_NODE_SETTING;
-        }
-    };
+    public static final String ML_ROLE_NAME = "ml";
 
     @Override
     public List<ActionHandler<? extends ActionRequest, ? extends ActionResponse>> getActions() {
--- a/plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
+++ b/plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.task;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -23,6 +24,7 @@ import org.opensearch.ml.action.stats.MLStatsNodeResponse;
 import org.opensearch.ml.action.stats.MLStatsNodesAction;
 import org.opensearch.ml.action.stats.MLStatsNodesRequest;
 import org.opensearch.ml.stats.MLNodeLevelStat;
+import org.opensearch.ml.utils.MLNodeUtils;
 
 import com.google.common.collect.ImmutableSet;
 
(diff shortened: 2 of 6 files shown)


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -115,14 +113,7 @@ public class MachineLearningPlugin extends Plugin implements ActionPlugin {
     private ClusterService clusterService;
     private ThreadPool threadPool;
 
-    public static final Setting<Boolean> IS_ML_NODE_SETTING = Setting.boolSetting("node.ml", false, Setting.Property.NodeScope);
```
jackiehanyang, line 118: curiosity question: what is the purpose of `IS_ML_NODE_SETTING` before and why we don't need it now?

jackiehanyang (approved):

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
ylwu-amzn: That part was from prototype when we try to support ML node. But actually it's not being used in our formal release as  OpenSearch core doesn't support ML role and we have no postpone that. Now OpenSearch supported ML role with dynamic role feature. We can add this back but we don't need this prototype/experiment code any more. Just move it to test part.

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
```diff
@@ -107,14 +107,32 @@ public void dispatchTask(ActionListener<DiscoveryNode> listener) {
...
             if (node.isDataNode()) {
                 eligibleDataNodes.add(node);
             }
         }
-        return eligibleDataNodes.toArray(new DiscoveryNode[0]);
+        if (eligibleMLNodes.size() > 0) {
+            DiscoveryNode[] mlNodes = eligibleMLNodes.toArray(new DiscoveryNode[0]);
+            log.debug("Find {} dedicated ML nodes: {}", eligibleMLNodes.size(), Arrays.toString(mlNodes));
+            return mlNodes;
+        } else {
```
Zhangxunmt, line 132: Nitpick: This "else" should be redundant.

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
```diff
@@ -107,14 +107,32 @@ public void dispatchTask(ActionListener<DiscoveryNode> listener) {
...
     }
 
-    private DiscoveryNode[] getEligibleDataNodes() {
+    /**
+     * Get eligible node to run ML task. If there are nodes with ml role, will return all these
+     * ml nodes; otherwise return all data nodes.
+     *
+     * @return array of discovery node
+     */
+    protected DiscoveryNode[] getEligibleNodes() {
```
Zhangxunmt, line 116: Is it only preferable to run ML tasks in ml node? I assume ml-common can run in data node as well. Also is there any logic in the ClusterState.nodes() to evaluate if any ml node is overloaded, etc? I just wonder, in the future, if we want to add more priority based strategy here to prioritize ML node, but still use data node is ML node is heavy loaded, etc.

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
ylwu-amzn: > Is it only preferable to run ML tasks in ml node? I assume ml-common can run in data node as well.

Check the comment "If there are nodes with ml role, will return all these ml nodes; otherwise return all data nodes."

> Also is there any logic in the ClusterState.nodes() to evaluate if any ml node is overloaded, etc?

Yes, we check JVM heap usage and how many ML task running on a node. If exceeds limit, will not dispatch new ML task to that node.

>  I just wonder, in the future, if we want to add more priority based strategy here to prioritize ML node, but still use data node if ML node is heavy loaded, etc.

I think we'd better ask user to scale the cluster by adding more ML node or switch to more powerful node type if ML node is heavy/over loaded. But this is not the one way door, we can always tune the code if cx really needs to run model on data nodes if ML node overloaded.

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
ylwu-amzn: It's code style preference. People have different preference for "No-else-after-return" or not, check https://stackoverflow.com/questions/46875442/unnecessary-else-after-return-no-else-return.

For me, I feel the code is more readable to keep `else` to make the returns have same indentation.

Issue #318, opened 2022-05-11
[FEATURE] Add LogisticRegression algorithm 


ylwu-amzn:
Currently we don't have classification algorithms in `ml-commons`. We can start support classification by adding LogisticRegression.

Zhangxunmt:
What are you proposing?
What problems are you trying to solve?
What is the user experience going to be?
Why should it be built? Any reason not to?
What will it take to execute?
What are remaining open questions?

elfisher:
This looks exciting! Thanks for putting the proposal together @Zhangxunmt! Could we add API examples of how someone would use this (PPL and REST)?

alexahorgan:
Feature Review Feedback 7/21/22:
Add API examples and identify a way to sort the list of algorithms into categories (naming conventions).

Zhangxunmt:
API Example: 
Train a Logistic Regression Model using the API in Ml-Common Plugin. 

POST _plugins/_ml/_train/logistic_regression
{
  "parameters": {
    "target": "class"
  },
  "input_query": {
    "query": {
      "match_all": {}
    },
    "_source": [
      "sepal_length_in_cm",
      "sepal_width_in_cm",
      "petal_length_in_cm",
      "petal_width_in_cm",
      "class"
    ],
    "size": 200
  },
  "input_index": [
    "iris_data"
  ]
}

Predict result for the trained Logistic Regression Model using the API in Ml-Common Plugin.
POST _plugins/_ml/_predict/logistic_regression/bMjQYYIBaEQ9YIwg1Bx3
{
  "parameters": {
    "target": "class"
  },
  "input_data": {
    "column_metas": [
      {
        "name": "sepal_length_in_cm",
        "column_type": "DOUBLE"
      },
      {
        "name": "sepal_width_in_cm",
        "column_type": "DOUBLE"
      },
      {
        "name": "petal_length_in_cm",
        "column_type": "DOUBLE"
      },
      {
        "name": "petal_width_in_cm",
        "column_type": "DOUBLE"
      }
    ],
    "rows": [
      {
        "values": [
          {
            "column_type": "DOUBLE",
            "value": 5.3
          },
          {
            "column_type": "DOUBLE",
            "value": 3.7
          },
          {
            "column_type": "DOUBLE",
            "value": 1.5
          },
          {
            "column_type": "DOUBLE",
            "value": 0.2
          }
        ]
      }
    ]
  }
}

Prediction Result:
{
  "status" : "COMPLETED",
[...]

PR #360, opened 2022-07-08
add logistic regression algorithm

Zhangxunmt:

### Description
Add Logistic Regression algorithm into ml-common. 
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/318

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -45,7 +45,7 @@ dependencies {
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation("com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}")
-    implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson}")
+    implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}")
     implementation group: 'com.google.guava', name: 'guava', version: '31.0.1-jre'
     implementation group: 'com.google.code.gson', name: 'gson', version: '2.9.0'
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'


ml-algorithms/build.gradle
```diff
@@ -20,6 +20,7 @@ dependencies {
     implementation group: 'org.tribuo', name: 'tribuo-clustering-kmeans', version: '4.2.1'
     implementation group: 'org.tribuo', name: 'tribuo-regression-sgd', version: '4.2.1'
     implementation group: 'org.tribuo', name: 'tribuo-anomaly-libsvm', version: '4.2.1'
+    implementation group: 'org.tribuo', name: 'tribuo-classification-experiments', version: '4.2.1'
```
Craigacp, line 23: This is going to pull in a bunch of Tribuo, including the XGBoost native library, which you might not want. If you just want the SGD logistic regression then you can pull in `tribuo-classification-sgd`.

Zhangxunmt: Thanks for the comment. This is a good point. :)
I wanted to use this congregated one to avoid pulling multiple different packages in the first place, but it seems ends up only using "tribuo-classification-sgd".

common/src/main/java/org/opensearch/ml/common/FunctionName.java
```diff
@@ -7,6 +7,7 @@
 
 public enum FunctionName {
     LINEAR_REGRESSION,
+    LOGISTIC_REGRESSION,
```
ylwu-amzn, line 10: We'd better put the new algorithm to last position. It may cause issue for a cluster with mixed version.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/TribuoUtil.java
```diff
@@ -47,6 +49,44 @@ public static Tuple<String[], double[][]> transformDataFrame(DataFrame dataFrame
...
+            }
+            ++i;
+        }
+        featureNames.remove(target);
+        return new Tuple<>(featureNames.toArray(new String[featureNames.size()]), featureValues);
+    }
+
+    public static String[] transformTargetValuesDataFrames(DataFrame dataFrame, String target) {
+        List<String> featureNames = Arrays.stream(dataFrame.columnMetas()).map(ColumnMeta::getName).collect(Collectors.toList());
+        int targetIndex = findFinalTargetIndex(featureNames, target);
```
ylwu-amzn, line 76: Line 75-67 seems duplicate code with line 53-54, maybe add one method `int getColumnIndex(String targetColumn)` in DataFrame ?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
Zhangxunmt: Will do in the revision. These two are both regressions so listed them together in the first place.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/TribuoUtil.java
Zhangxunmt: Yes it's doable.

PR #383, opened 2022-08-02
Add Logistic Regression algorithm

Zhangxunmt:

### Description
Add Tribuo Logistic Regression into ML-Common.
UTs and ITs of logistic regression are included in the PR. Addressed comments from the previous PR.

### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/318
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/dataframe/DefaultDataFrame.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataframe/DefaultDataFrame.java
@@ -132,6 +132,24 @@ public class DefaultDataFrame extends AbstractDataFrame{
     }
 
     @Override
+    public int getColumnIndex(String target) {
+        List<String> featureNames = Arrays.stream(this.columnMetas()).map(ColumnMeta::getName).collect(Collectors.toList());
+
+        int targetIndex = -1;
+        for (int i = 0; i < featureNames.size(); ++i) {
+            if (featureNames.get(i).equals(target)) {
+                targetIndex = i;
+                break;
+            }
+        }
+        if (targetIndex == -1) {
+            throw new IllegalArgumentException("No matched target when generating dataset from data frame.");
+        }
+
+        return targetIndex;
+    }
+
+    @Override
     public Iterator<Row> iterator() {
         return rows.iterator();
     }
--- a/common/src/main/java/org/opensearch/ml/common/dataframe/DefaultDataFrame.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataframe/DefaultDataFrame.java
@@ -133,11 +133,11 @@ public class DefaultDataFrame extends AbstractDataFrame{
 
     @Override
     public int getColumnIndex(String target) {
-        List<String> featureNames = Arrays.stream(this.columnMetas()).map(ColumnMeta::getName).collect(Collectors.toList());
+        List<String> columnNames = Arrays.stream(this.columnMetas()).map(ColumnMeta::getName).collect(Collectors.toList());
 
         int targetIndex = -1;
-        for (int i = 0; i < featureNames.size(); ++i) {
-            if (featureNames.get(i).equals(target)) {
+        for (int i = 0; i < columnNames.size(); ++i) {
+            if (columnNames.get(i).equals(target)) {
                 targetIndex = i;
                 break;
             }
--- a/common/src/main/java/org/opensearch/ml/common/input/parameter/regression/LogisticRegressionParams.java
+++ b/common/src/main/java/org/opensearch/ml/common/input/parameter/regression/LogisticRegressionParams.java
@@ -39,6 +39,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
     public static final String EPSILON_FIELD = "epsilon";
     public static final String EPOCHS_FIELD = "epochs";
     public static final String BATCH_SIZE_FIELD = "batch_size";
+    public static final String LOGGING_INTERVAL_FIELD = "logging_interval";
     public static final String SEED_FIELD = "seed";
     public static final String TARGET_FIELD = "target";
 
@@ -48,6 +49,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
     private Double epsilon;
     private Integer epochs;
     private Integer batchSize;
+    private Integer loggingInterval;
     private Long seed;
     private String target;
 
@@ -59,6 +61,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
         Double epsilon,
         Integer epochs,
         Integer batchSize,
+        Integer loggingInterval,
         Long seed,
         String target
     ) {
@@ -68,6 +71,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
         this.epsilon = epsilon;
         this.epochs = epochs;
         this.batchSize = batchSize;
+        this.loggingInterval = loggingInterval;
         this.seed = seed;
         this.target = target;
     }
@@ -84,6 +88,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
         this.epsilon = in.readOptionalDouble();
         this.epochs = in.readOptionalInt();
         this.batchSize = in.readOptionalInt();
+        this.loggingInterval = in.readOptionalInt();
         this.seed = in.readOptionalLong();
         this.target = in.readOptionalString();
     }
@@ -95,6 +100,7 @@ public class LogisticRegressionParams implements MLAlgoParams {
         Double epsilon = null;
         Integer epochs = null;
         Integer batchSize = null;
+        Integer loggingInterval = null;
         Long seed = null;
         String target = null;
 
@@ -122,6 +128,9 @@ public class LogisticRegressionParams implements MLAlgoParams {
                 case BATCH_SIZE_FIELD:
                     batchSize = parser.intValue(false);
                     break;
+                case LOGGING_INTERVAL_FIELD:
+                    loggingInterval = parser.intValue(false);
+                    break;
                 case SEED_FIELD:
                     seed = parser.longValue(false);
                     break;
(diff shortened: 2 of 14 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/TribuoUtil.java
```diff
@@ -95,6 +134,9 @@ public static <T extends Output<T>> MutableDataset<T> generateDataset(DataFrame
                     // TODO: support anomaly labels to evaluate prediction result
                     example = new ArrayExample<>((T) new Event(defaultEventType), featureNamesValues.v1(), featureNamesValues.v2()[i]);
                     break;
+                case LABEL:
+                    example = new ArrayExample<>((T) new Label(""), featureNamesValues.v1(), featureNamesValues.v2()[i]);
```
Craigacp, line 138: You can use [`UNKNOWN_LABEL`](https://tribuo.org/learn/4.2/javadoc/org/tribuo/classification/LabelFactory.html#UNKNOWN_LABEL) for this if you want. It reduces the amount of allocation, and is fine to be shared among examples at prediction time. Or `outputFactory.getUnknownOutput()` as you already have an output factory reference.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LogisticRegression.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    @Override
+    public Model train(DataFrame dataFrame) {
+        MutableDataset<Label> trainDataset = TribuoUtil.generateDatasetWithTarget(dataFrame, new LabelFactory(),
+                "Logistic regression training data from OpenSearch", TribuoOutputType.LABEL, parameters.getTarget());
+        // Integer epochs = Optional.ofNullable(parameters.getEpochs()).orElse(DEFAULT_EPOCHS);
+        // LinearSGDTrainer(objective=LogMulticlass,optimiser=AdaGrad(initialLearningRate=1.0,epsilon=0.1,initialValue=0.0),epochs=5,minibatchSize=1,seed=12345)
+        Trainer<Label> logisticRegressionTrainer = new LogisticRegressionTrainer();
```
Craigacp, line 119: Our `LogisticRegressionTrainer` is designed to be very simple for onboarding users & demos, I recommend you use `LinearSGDTrainer` and configure it given you're wrapping all the interface.

ylwu-amzn: @Craigacp Thanks for your suggestion. We have integrated `LinearSGDTrainer` to provide linear regression https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LinearRegression.java#L203.  

You mean we should wrap `LinearSGDTrainer` rather than `LogisticRegressionTrainer` to provide logistic regression?

Craigacp: Well, this is slightly unfortunate naming on our part. We have three classes all called `LinearSGDTrainer`, `org.tribuo.classification.sgd.linear.LinearSGDTrainer`, `org.tribuo.multilabel.sgd.linear.LinearSGDTrainer` and `org.tribuo.regression.sgd.linear.LinearSGDTrainer`. They have basically the same interface, but are constrained by the type system to accept the appropriate kind of objective function, and have the right generic types. `LogisticRegressionTrainer` is just a `org.tribuo.classification.sgd.linear.LinearSGDTrainer` subclass with a constructor that has fairly sensible defaults for running quick tests & demos. So for your use case I'd recommend you use `org.tribuo.classification.sgd.linear.LinearSGDTrainer` as then you can expose basically the same set of options as you do for linear regression (aside from the difference in objective functions).

ylwu-amzn: Thanks for the details. @Zhangxunmt I think we should change to `org.tribuo.classification.sgd.linear.LinearSGDTrainer`. 

BTW, @Craigacp , For linear regression `org.tribuo.regression.sgd.linear.LinearSGDTrainer` and this `org.tribuo.classification.sgd.linear.LinearSGDTrainer`, so you have any suggestion for the default parameter value like objective function, epochs, batch size etc?

Craigacp: For the classification/logistic regression the ones we use in the constructor are fine - https://github.com/oracle/tribuo/blob/main/Classification/SGD/src/main/java/org/tribuo/classification/sgd/linear/LogisticRegressionTrainer.java#L44. You could use them for the linear regression too, substituting the `LogMulticlass` objective for `SquaredLoss`. For the linear models a batch size of 1 tends to work just fine, and there isn't really any performance benefit to larger ones given the way the linear algebra system is setup (as it's not multithreaded yet).

Unfortunately we don't have any convergence metrics like validation error built into Tribuo's trainers, and we'd need to modify the trainer API to enable such things.

Zhangxunmt: I was using LinearSGDTrainer with the default parameters in the first place (check the commented out line), then I realized it's almost equivalent to the given LogisticRegressionTrainer since it's basically using LinearSGDTrainer under the hood. We can change it to LinearSGDTrainer, and let customers to configure parameters through the training requests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/TribuoUtil.java
Zhangxunmt: Replaced with outputFactory.getUnknownOutput().

common/src/main/java/org/opensearch/ml/common/dataframe/DefaultDataFrame.java
```diff
@@ -131,6 +131,24 @@ public DataFrame select(int[] columns) {
         return new DefaultDataFrame(newColumnMetas, rows.stream().map(row-> row.select(columns)).collect(Collectors.toList()));
     }
 
+    @Override
+    public int getColumnIndex(String target) {
+        List<String> featureNames = Arrays.stream(this.columnMetas()).map(ColumnMeta::getName).collect(Collectors.toList());
```
ylwu-amzn, line 136: Minor: rename `featureNames` as `columnNames`?

common/src/main/java/org/opensearch/ml/common/input/parameter/regression/LogisticRegressionParams.java
```diff
@@ -0,0 +1,244 @@
...
+                throw new IllegalArgumentException("Wrong objective type");
+            }
+        }
+    }
+
+    public enum OptimizerType {
+        SIMPLE_SGD,
+        LINEAR_DECAY_SGD,
+        SQRT_DECAY_SGD,
+        ADA_GRAD,
```
ylwu-amzn, line 231: @Craigacp help review if all these supported for logistic regression?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LogisticRegression.java
```diff
@@ -0,0 +1,160 @@
...
+        if (parameters.getEpochs() != null && parameters.getEpochs() < 0) {
+            throw new IllegalArgumentException("Epochs should not be negative.");
+        }
+
+        if (parameters.getBatchSize() != null && parameters.getBatchSize() < 0) {
+            throw new IllegalArgumentException("MiniBatchSize should not be negative.");
+        }
+
+        if (parameters.getLoggingInterval() != null && parameters.getLoggingInterval() < 0) {
+            throw new IllegalArgumentException("Logging intervals should not be negative.");
```
ylwu-amzn, line 91: From java doc of `org.tribuo.classification.sgd.linear.LinearSGDTrainer`
> Log the loss after this many iterations. If -1 don't log anything.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LogisticRegression.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    private void createOptimiser() {
+        LogisticRegressionParams.OptimizerType optimizerType = Optional.ofNullable(parameters.getOptimizerType()).orElse(DEFAULT_OPTIMIZER_TYPE);
+        Double learningRate = Optional.ofNullable(parameters.getLearningRate()).orElse(DEFAULT_LEARNING_RATE);
+        Double epsilon = Optional.ofNullable(parameters.getEpsilon()).orElse(DEFAULT_EPSILON);
+
+        switch (optimizerType) {
+            // ToDo: Add more possible optimizer. Tribuo only provides AdaGrad for logistic regression.
```
ylwu-amzn, line 118: @Craigacp can you confirm if this is true? Does `org.tribuo.classification.sgd.linear.LinearSGDTrainer` support other optimizer ?

common/src/main/java/org/opensearch/ml/common/input/parameter/regression/LogisticRegressionParams.java
Craigacp: These are all supported.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LogisticRegression.java
Zhangxunmt: Good catch!

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/regression/LogisticRegression.java
Zhangxunmt: Thanks for the clarification. These optimizers are all supported in the PR now.

ylwu-amzn (approved):

jngz-es (approved):

PR #384, opened 2022-08-02
fix index mapping

ylwu-amzn:

### Description
ML model content stored as text by mistake in index. This PR fixed index mapping and store model content as binary.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -5,6 +5,15 @@
 
 package org.opensearch.ml.indices;
 
+import static org.opensearch.ml.common.CommonValue.META;
+import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
+import static org.opensearch.ml.common.CommonValue.ML_TASK_INDEX;
+import static org.opensearch.ml.common.CommonValue.SCHEMA_VERSION_FIELD;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
 import lombok.experimental.FieldDefaults;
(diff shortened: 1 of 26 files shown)


Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -108,8 +78,67 @@ public void initMLIndexIfAbsent(String indexName, String mapping, ActionListener
...
+                            .putMapping(
+                                new PutMappingRequest().indices(indexName).source(mapping, XContentType.JSON),
+                                ActionListener.wrap(response -> {
+                                    if (response.isAcknowledged()) {
+                                        indexMappingUpdated.get(indexName).set(true);
+                                        listener.onResponse(true);
+                                    } else {
+                                        listener.onFailure(new MLException("Failed to update index: " + indexName));
+                                    }
+                                }, exception -> {
```
jngz-es, line 97: Should we set indexMappingUpdated to false explicitly when getting an exception? I am not sure if there is a chance that we get an exception after setting indexMappingUpdated to true. So I recommend set to false here.

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -108,8 +78,67 @@ public void initMLIndexIfAbsent(String indexName, String mapping, ActionListener
...
+                                    listener.onFailure(exception);
+                                })
+                            );
+                    } else {
+                        // no need to update index if it does not exist or the version is already up-to-date.
+                        indexMappingUpdated.get(indexName).set(true);
+                        listener.onResponse(true);
+                    }
+                }, e -> {
+                    log.error("Failed to update index mapping", e);
```
jngz-es, line 108: Same as above.

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
ylwu-amzn: The `indexMappingUpdated` flag is to make sure we only update the index mapping for first request.  `indexMappingUpdated` was initialized as false, check line 44-45. We will set the flag as true if

1. Index doesn't exist. When create index, it will use latest mapping.
2. Index exists but its mapping is already latest (for example, some other request/node already updated the index mapping), then no need to update it. 
3. Index mapping is old, and we updated its mapping successfully.

For case 1 and 2, they map to line 104-105.
For case3, it maps to line 92.

When we set `indexMappingUpdated` as true. That means the index mapping already updated. No need to set `indexMappingUpdated` as false and update again.

jngz-es: Thanks for the explanation.

jngz-es (approved):
LGTM, thanks.

Issue #410, opened 2022-08-26
Add Profile APIs for ML-Common to return runtime information


Zhangxunmt:
Currently ML-Common APIs do not return any runtime model or ML-Task information. We need to have an API to understand how things go in the runtime when there are multiple model training/prediction Tasks in the memory. 

The profile APIs provide insights for users to know their on-going ML Tasks status before they are stored in the Index.

PR #411, opened 2022-08-26
add task profile APIs

Zhangxunmt:
### Description
Add ML-Common Profile APIs to return runtime information for ML Models/Tasks. 
Will add model profile, per node profile, and more unit tests in the next revision.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/410
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build-tools/repositories.gradle
+++ b/build-tools/repositories.gradle
@@ -7,4 +7,5 @@ repositories {
     mavenLocal()
     maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
     mavenCentral()
+    maven { url "https://d1nvenhzbhpy0q.cloudfront.net/snapshots/lucene/" }
 }


plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileNodeResponse.java
```diff
@@ -0,0 +1,74 @@
...
+        super.writeTo(out);
+        if (mlNodeTasks != null) {
+            out.writeBoolean(true);
+            out.writeMap(mlNodeTasks, StreamOutput::writeString, (o, r) -> r.writeTo(o));
+        } else {
+            out.writeBoolean(false);
+        }
+    }
+
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
```
dhrubo-os, line 58: Can we add javadoc for the public functions?

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileNodeResponse.java
```diff
@@ -0,0 +1,74 @@
...
+     * @throws IOException throws an IO exception if the StreamInput cannot be reML from
+     */
+    public MLTaskProfileNodeResponse(StreamInput in) throws IOException {
+        super(in);
+        if (in.readBoolean()) {
+            this.mlNodeTasks = in.readMap(StreamInput::readString, MLTask::new);
+        }
+    }
+
+    public MLTaskProfileNodeResponse(DiscoveryNode node, Map<String, MLTask> nodeTasks) {
```
dhrubo-os, line 38: Map<String, MLTask> --> is the string for node task id? I hope we aren't expecting readers to have intensive knowledge of this package.

dhrubo-os (changes requested):
Overall I would suggest to add more documentation throughout the PR. As a new reader of this package it was not quite easy for me to understand what's going on.

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileAction.java
```diff
@@ -0,0 +1,22 @@
...
+
+package org.opensearch.ml.action.profile;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.ml.constant.CommonValue;
+
+public class MLTaskProfileAction extends ActionType<MLTaskProfileResponse> {
+    // Internal Action which is not used for public facing RestAPIs.
+    public static final String NAME = CommonValue.ACTION_PREFIX + "tasks/profile";
+    public static final MLTaskProfileAction INSTANCE = new MLTaskProfileAction();
```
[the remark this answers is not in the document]
Zhangxunmt: Check the other Actions. These are normal names.

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileNodeResponse.java
```diff
@@ -0,0 +1,74 @@
...
+    private Map<String, MLTask> mlNodeTasks;
+    /**
+     * Constructor
+     *
+     * @param in StreamInput
+     * @throws IOException throws an IO exception if the StreamInput cannot be reML from
+     */
+    public MLTaskProfileNodeResponse(StreamInput in) throws IOException {
+        super(in);
+        if (in.readBoolean()) {
```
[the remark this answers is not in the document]
Zhangxunmt: This correspond to line 51 out.writeBoolean(true) if mlNodeTasks != null.

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileNodeResponse.java
Zhangxunmt: I added more comments in the next revision. The key is the taskId and the value is the MLTask that we want to retrieve in the memory to the users.

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileAction.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
Zhangxunmt: Yes this class is actually only a hooker that defines the Response and the real action is in the MLTaskProfileTransportAction.

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileAction.java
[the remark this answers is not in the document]
dhrubo-os: what do you mean by normal names? Is this a convention to keep this way? Is is name of the action? ACTION_NAME?

plugin/src/main/java/org/opensearch/ml/action/profile/MLTaskProfileAction.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package org.opensearch.ml.action.profile;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.ml.constant.CommonValue;
+
+public class MLTaskProfileAction extends ActionType<MLTaskProfileResponse> {
+    // Internal Action which is not used for public facing RestAPIs.
+    public static final String NAME = CommonValue.ACTION_PREFIX + "tasks/profile";
```
ylwu-amzn, line 13: How about "tasks/profile" -> "profile" as we are going to profile not just for task, also for model and may add other things in future.

ylwu-amzn:
From the description "Will add model profile, per node profile, and more unit tests in the next revision.", seems this PR only has task profile. Are you planning to add model profile in this PR or next?

plugin/src/main/java/org/opensearch/ml/rest/RestMLTaskProfileAction.java
```diff
@@ -0,0 +1,95 @@
...
+import org.opensearch.rest.BaseRestHandler;
+import org.opensearch.rest.BytesRestResponse;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestStatus;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+
+@Log4j2
+public class RestMLTaskProfileAction extends BaseRestHandler {
```
ylwu-amzn, line 37: Going to create multiple profile APIs, like one for task profile, another one for model? How about we keep one profile API and user can specify the target they want to detect?

Zhangxunmt: I understand your point. We can refactor this one and let user to define the target in the request body.

PR #446, opened 2022-09-20
add profile APIs for model and task

Zhangxunmt:

### Description
Add profile API to return runtime information of cached ML tasks.

User needs to provide the model_ids or task_ids to retrieve the profiles. Otherwise, it returns all the caches tasks in memory. 
Example of usage:
GET _plugins/_ml/profile/model/{model_id}
or
GET _plugins/_ml/profile/task/{task_id}
or 
GET _plugins/_ml/profile           // return all profile for all tasks in training status
or 
GET _plugins/_ml/profile
{
  "model_ids" : [ "model_id1", "model_id2"],
  "task_ids" : [ "task_id1", "task_id2"]
}

Sample Outputs:
{
  "nodes" : {
    "VB8wEWjkSPitn94hOp2c2g" : {
      "91zxzomb23vvkpjxnmzw" : {
        "task_id" : "91zxZoMB23vVkPJxnMzw",
        "model_id" : "-FzxZoMB23vVkPJxncwz",
        "task_type" : "TRAINING",
        "function_name" : "KMEANS",
        "state" : "RUNNING",
        "input_type" : "SEARCH_QUERY",
        "worker_node" : "VB8wEWjkSPitn94hOp2c2g",
        "create_time" : 1663879453935,
        "last_update_time" : 1663879453935,
        "is_async" : true
      },
      "8lzxzomb23vvkpjxhsxc" : {
            "task_id" : "rOOrYoMBmooB0MhyOtML",
            "model_id" : "reOrYoMBmooB0MhyOtMo",
            "task_type" : "TRAINING"
            "function_name" : "Linear Regression"
            "state" : "RUNNING"
            "input_type" : "DATA_FRAME"
            "progress" : ""
            "output_index" : ""
            "worker_node" : "qDqliCEgSk64QQYlGj04cg"
            "create_time" : 1663807732114
            "last_update_time" : 1663807732114
            "error" : ""
            "is_async" : true
      }
    }
  }
}

### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/410
 
### Check List
- [x ] New functionality includes testing.
  - [ x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileAction.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.action.profile;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.ml.constant.CommonValue;
+
+public class MLProfileAction extends ActionType<MLProfileResponse> {
+    // Internal Action which is not used for public facing RestAPIs.
+    public static final String NAME = CommonValue.ACTION_PREFIX + "profile/nodes";
+    public static final MLProfileAction INSTANCE = new MLProfileAction();
+
+    /**
+     * Constructor
+     */
+    private MLProfileAction() {
+        super(NAME, MLProfileResponse::new);
+    }
+}
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
@@ -95,7 +95,7 @@ public class MLProfileTransportAction extends
         Map<String, MLTask> mlLocalTasks = new HashMap<>();
         MLProfileInput mlProfileInput = mlProfileRequest.getMlProfileInput();
         mlTaskManager.getTaskCaches().forEach((key, value) -> {
-            if (!mlProfileInput.emptyTasks() && mlProfileInput.getTaskIds().contains(key)) {
+            if (mlProfileInput.isReturnAllMLTasks() || (!mlProfileInput.emptyTasks() && mlProfileInput.getTaskIds().contains(key))) {
                 log.info("Runtime task profile is found for model {}", value.getMlTask().getModelId());
                 mlLocalTasks.put(key, value.getMlTask());
                 return;
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
@@ -10,6 +10,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import lombok.extern.log4j.Log4j2;
+
 import org.opensearch.action.FailedNodeException;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.nodes.TransportNodesAction;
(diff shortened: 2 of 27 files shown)


plugin/src/main/java/org/opensearch/ml/task/MLTaskManager.java
```diff
@@ -37,6 +38,7 @@
  * MLTaskManager is responsible for managing MLTask.
  */
 @Log4j2
+@Getter
```
ylwu-amzn, line 41: This looks risky to expose `taskCaches` field. As you can see we have several `synchronized` methods. By exposing `taskCaches`, anyone can add/remove tasks from it.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
```diff
@@ -0,0 +1,111 @@
...
+        return new MLProfileNodeResponse(in);
+    }
+
+    @Override
+    protected MLProfileNodeResponse nodeOperation(MLProfileNodeRequest request) {
+        return createMLProfileNodeResponse(request.getMlProfileRequest());
+    }
+
+    private MLProfileNodeResponse createMLProfileNodeResponse(MLProfileRequest mlProfileRequest) {
+        log.info("Calculating ml profile response on node id:{}", clusterService.localNode().getId());
```
ylwu-amzn, line 94: If we get profiles on all nodes of multi-node cluster, do we expect to see this log on every node?

plugin/src/main/java/org/opensearch/ml/profile/MLProfileInput.java
```diff
@@ -0,0 +1,149 @@
...
+        if (nodeIds != null) {
+            builder.field(NODE_IDS, nodeIds);
+        }
+        builder.field(RETURN_ALL, returnAllMLTasks);
+        builder.endObject();
+        return builder;
+    }
+
+    public boolean retrieveProfileOnAllNodes() {
+        return nodeIds.size() == 0;
```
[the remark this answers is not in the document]
Zhangxunmt: This shouldn't be null by any means. All the constructors have initialization. Adding null would make the Unit test branch coverage to drop below 0.8, but there's no way to instantiate a MLProfileInput with a null nodeIds to test that scenario. I will add the null checks anyway and see if there are other ways to increase the coverage.

plugin/src/main/java/org/opensearch/ml/task/MLTaskManager.java
Zhangxunmt: This is a good point. Will remove @Getter add a public method in the MLTaskManager to return cached tasks in arrays.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
Zhangxunmt: Yes that's my intention because this might help us to debug. The NodeResponse should be produced at each node. If any node crushes or stuck somewhere, we can use this log to identify the problem.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileAction.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.action.profile;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.ml.constant.CommonValue;
+
+public class MLProfileAction extends ActionType<MLProfileResponse> {
+    // Internal Action which is not used for public facing RestAPIs.
```
ylwu-amzn, line 12: Minor, seems `MLProfileAction.INSTANCE` is used in `RestMLProfileAction.java`, remove this comment?

ylwu-amzn (approved):

rbhavna (approved):

PR #452, opened 2022-09-27
tune ML model to support custom model

ylwu-amzn:

### Description
Tune ML model to support custom model.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -7,14 +7,21 @@ package org.opensearch.ml.common;
 
 import lombok.Builder;
 import lombok.Getter;
+import lombok.Setter;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
 import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 import org.opensearch.common.xcontent.XContentParser;
 import org.opensearch.commons.authuser.User;
+import org.opensearch.ml.common.model.MLModelConfig;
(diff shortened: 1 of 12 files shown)


common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -28,23 +35,74 @@ public class MLModel implements ToXContentObject {
...
+    public static final String LATEST_UPLOADED_TIME_FIELD = "latest_uploaded_time";
+    public static final String LATEST_LOADED_TIME_FIELD = "latest_loaded_time";
+    public static final String LATEST_UNLOADED_TIME_FIELD = "latest_unloaded_time";
+
+    public static final String MODEL_ID_FIELD = "model_id";
+    public static final String CHUNK_NUMBER_FIELD = "chunk_number";
+    public static final String TOTAL_CHUNKS_FIELD = "total_chunks";
+
     private String name;
     private FunctionName algorithm;
```
Zhangxunmt, line 55: Based on the design, all custom models use "custom" in the algorithm field. We can use this algorithm field to check if model is build-in or custom. Is there any possibility in the future we want to distinguish different custom models by algorithms, so this field may need to include more custom model names? If we reserve this field to allow users to use different custom model algorithms, is it better to add a boolean field "isCustomModel" to help us identify if the model is build-in or custom?

Zhangxunmt (approved):
LGTM. We can modify the MLModel and the Configs whenever we think is necessary later.

common/src/main/java/org/opensearch/ml/common/MLModel.java
ylwu-amzn: Good question. We can differentiate different custom model types with `MLModelTaskType`

dhrubo-os (approved):

PR #462, opened 2022-10-10
update delete model TransportAction to support custom model

Zhangxunmt:
### Description
Delete all chunks in delete Model API for custom models.

DELETE _plugins/_ml/models/"model-ids"
{
  "_index" : ".plugins-ml-model",
  "_id" : "x95LqoMBr6prntZN5AQJ",
  "_version" : 2,
  "result" : "deleted",
  "_shards" : {
    "total" : 2,
    "successful" : 1,
    "failed" : 0
  },
  "_seq_no" : 37,
  "_primary_term" : 7
}
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
@@ -23,6 +23,7 @@ import org.opensearch.client.Client;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.index.query.TermsQueryBuilder;
+import org.opensearch.index.reindex.BulkByScrollResponse;
 import org.opensearch.index.reindex.DeleteByQueryAction;
 import org.opensearch.index.reindex.DeleteByQueryRequest;
 import org.opensearch.ml.common.transport.model.MLModelDeleteAction;
@@ -37,6 +38,10 @@ import com.google.common.annotations.VisibleForTesting;
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
 public class DeleteModelTransportAction extends HandledTransportAction<ActionRequest, DeleteResponse> {
 
+    static final String TIMEOUT_MSG = "Timeout while deleting model of ";
+    static final String BULK_FAILURE_MSG = "Bulk failure while deleting model of ";
+    static final String SEARCH_FAILURE_MSG = "Search failure while deleting model of ";
+    static final String OS_STATUS_EXCEPTION_MESSAGE = "Failed to delete all model chunks ";
     Client client;
 
     @Inject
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -64,4 +72,23 @@ public void onFailure(Exception e) {
...
+        DeleteByQueryRequest deleteModelsRequest = new DeleteByQueryRequest(ML_MODEL_INDEX);
+        deleteModelsRequest.setQuery(new TermsQueryBuilder(MODEL_ID_FIELD, modelId));
+
+        client.execute(DeleteByQueryAction.INSTANCE, deleteModelsRequest, ActionListener.wrap(r -> {
+            if (r.getBulkFailures() == null || r.getBulkFailures().size() == 0) {
+                log.info("All model chunks are deleted for model {}", modelId);
+                actionListener.onResponse(deleteResponse);
+            } else {
+                actionListener
+                    .onFailure(new OpenSearchStatusException("Failed to delete all model chunks", RestStatus.INTERNAL_SERVER_ERROR));
```
jngz-es, line 86: 1. If having bulk failures here, it is more helpful to throw these failures for users to understand why, as delete API is resource related. Usually users use this API when resources are not enough. If failed, they want to know why.
2. Is it possible to partially fail deleting some chunks? If yes, is there a mechanism to clean them, like garbage collection?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -64,4 +77,35 @@ public void onFailure(Exception e) {
...
+        String errorMessage = "";
+        if (response.isTimedOut()) {
+            errorMessage = OS_STATUS_EXCEPTION_MESSAGE + "," + TIMEOUT_MSG + modelId;
+        } else if (!response.getBulkFailures().isEmpty()) {
+            errorMessage = OS_STATUS_EXCEPTION_MESSAGE + "," + BULK_FAILURE_MSG + modelId;
+        } else {
+            errorMessage = OS_STATUS_EXCEPTION_MESSAGE + "," + SEARCH_FAILURE_MSG + modelId;
+        }
+        log.info(response.toString());
+        actionListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.INTERNAL_SERVER_ERROR));
```
dhrubo-os, line 109: What does `Bulk failure` and `search failure` mean?

 If search failure mean: customer is provided a model id and we didn't find that model to delete, then I don't think that error should be treated as `Internal server error 500`.
 
Just curious, if customers see `Bulk failure` what do they suppose to do?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -20,15 +22,26 @@
...
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
 
+import com.google.common.annotations.VisibleForTesting;
+
 @Log4j2
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
 public class DeleteModelTransportAction extends HandledTransportAction<ActionRequest, DeleteResponse> {
 
+    static final String TIMEOUT_MSG = "Timeout while deleting model of ";
```
dhrubo-os, line 41: [nit] Can we add `@VisibleForTesting` ?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
Zhangxunmt: 1. Added a new function to return and log different error messages. AD used the same logic in error handling in this case.
2. This is possible. That's why we will add a clean up job to retry deleting again to make sure no leftovers.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
Zhangxunmt: Check this doc https://www.javadoc.io/doc/org.elasticsearch/elasticsearch/6.0.1/org/elasticsearch/index/reindex/ScrollableHitSource.SearchFailure.html. If no model is found in search, it just doesn't delete and wouldn't return an error. It only delete model chunks that are returned in search results. 

Bulk failure means a list of Failures, Users can use the search API to see which chunks are deleted. If the model metadata is deleted, a cleanup job will remove the rest of the chunks.

jngz-es (approved):

dhrubo-os (approved):

PR #477, opened 2022-10-16
add text embedding model

ylwu-amzn:

### Description
Add text embedding model which will load and run text embedding model on top of DJL.
It supports `PyTorch` and `OnnxRuntime` engine. 

1. User can upload torchscript file or ONNX file (upload function will be in a new PR), 
2. Then load model by calling `TextEmbeddingModel#initModel` method. 
3. Once model loaded, user can inference with `TextEmbeddingModel#predict` method. 
4. If model no longer needed for inference, user can unload model from memory with `TextEmbeddingModel#close` method.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


mingshl (approved):

Zhangxunmt (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/MLEngine.java
```diff
@@ -15,13 +15,64 @@
...
 
+    public static Path DJL_CACHE_PATH;
+    public static Path DJL_MODELS_CACHE_PATH;
+    public static void setDjlCachePath(Path opensearchDataFolder) {
+        DJL_CACHE_PATH = opensearchDataFolder.resolve("djl");
+        DJL_MODELS_CACHE_PATH = DJL_CACHE_PATH.resolve("models_cache");
+    }
+
+    public static Path getUploadModelPath(String modelId, String modelName, Integer version) {
+        return getUploadModelPath(modelId).resolve(version + "").resolve(modelName);
```
frankfliu, line 34: Can version be null? if not it's better use int

Avoid using `(version + "")`, use version.toString() or `String.valueOf(version)` if version is primitive type.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/HuggingfaceTextEmbeddingServingTranslator.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Output processOutput(TranslatorContext ctx, NDList list) throws Exception {
+        float[] ret = translator.processOutput(ctx, list);
+        List<ModelTensor> outputs = new ArrayList<>();
+        Number[] data = new Float[ret.length];
+        for (int i=0;i<ret.length; i++) {
```
frankfliu, line 58:
```suggestion
        for (int i = 0; i < ret.length; i++) {
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/HuggingfaceTextEmbeddingServingTranslator.java
```diff
@@ -0,0 +1,70 @@
...
+            data[i] = ret[i];
+        }
+        long[] shape = new long[]{1, ret.length};
+        outputs.add(new ModelTensor(SENTENCE_EMBEDDING, data, shape, MLResultDataType.FLOAT32, null));
```
frankfliu, line 62:
```suggestion
        ModelTensor tensor = new ModelTensor(SENTENCE_EMBEDDING, data, shape, MLResultDataType.FLOAT32, null);
        List<ModelTensor> outputs = Collections.singletonList(tensor);
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,239 @@
...
+                            if (findModelFile) {
+                                throw new IllegalArgumentException("found multiple models");
+                            }
+                            findModelFile = true;
+                            String suffix = name.substring(name.lastIndexOf("."));
+                            file.renameTo(new File(modelPath.resolve(modelName + suffix).toUri()));
+                        }
+                    }
+                    Map<String, Object> arguments = new HashMap<>();
+                    arguments.put("engine", engine);
```
frankfliu, line 160: You don't need use arguments, you can just use:

```
Criteria.builder().optEngine(engine)
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,239 @@
...
+                    System.setProperty("DJL_CACHE_DIR", DJL_CACHE_PATH.toAbsolutePath().toString());
+                    System.setProperty("java.library.path", DJL_CACHE_PATH.toAbsolutePath().toString());
+                    Thread.currentThread().setContextClassLoader(ai.djl.Model.class.getClassLoader());
+                    Path modelPath = getModelCachePath(modelId, modelName, version);
+                    File pathFile = new File(modelPath.toUri());
+                    if (pathFile.exists()) {
+                        FileUtils.deleteDirectory(pathFile);
+                    }
+                    ZipUtils.unzip(new FileInputStream(modelZipFile), modelPath);
+                    boolean findModelFile = false;
```
frankfliu, line 147: Why you need check file here?
You already know which engine to use. it doesn't hurt if the zip file contains multiple models, why need add this restriction?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,239 @@
...
+                    Map<String, Object> arguments = new HashMap<>();
+                    arguments.put("engine", engine);
+                    Criteria.Builder<Input, Output> criteriaBuilder = Criteria.builder()
+                            .setTypes(Input.class, Output.class)
+                            .optApplication(Application.UNDEFINED)
+                            .optArguments(arguments)
+                            .optModelPath(modelPath);
+                    TextEmbeddingModelConfig textEmbeddingModelConfig = (TextEmbeddingModelConfig) modelConfig;
+                    TextEmbeddingModelConfig.FrameworkType transformersType = textEmbeddingModelConfig.getFrameworkType();
+                    if (ONNX_ENGINE.equals(engine)) { //ONNX
```
frankfliu, line 168: Why use two different type of model. They are actually the same, just with one more layer.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,239 @@
...
+                        criteriaBuilder.optTranslator(new ONNXSentenceTransformerTextEmbeddingTranslator());
+                    } else { // pytorch
+                        if (transformersType == SENTENCE_TRANSFORMERS) {
+                            criteriaBuilder.optTranslator(new SentenceTransformerTextEmbeddingTranslator());
+                        } else {
+                            criteriaBuilder.optTranslatorFactory(new HuggingfaceTextEmbeddingTranslatorFactory());
```
frankfliu, line 174: Is that possbile add a `serving.properties` in the zip file, so you don't have to write this ugly if else here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/FileUtils.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.opensearch.ml.engine.utils;
+
+import lombok.extern.log4j.Log4j2;
+
+import java.io.File;
+import java.nio.file.Path;
+
+@Log4j2
+public class FileUtils {
```
frankfliu, line 14: You can consider use 'ai.djl.util.Utils.deleteQuitely()`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: Good point, will fix

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: Will the engine throw error if have multiple model files? Or the engine will just load the first model file and ignore others?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: Yes, agree, ideally they should be the same, we still met some issue for ONNX result. Seems its result can't match 
 Python model. This ONNX translator is for debugging that part. Once that testing done, we may merge it with other translator.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
frankfliu: You specify the engine in the criteria, so it will only load the model with the engine you specified. All other files will be ignored.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: That's a good point. We can do that. We add transformer type in model for UI consideration, so user can review these model serving properties easily on model list page.

Put these serving properties in config file looks good for our future pre-built model or any model from model zoo. For current phase, we only support custom model. I think it's ok to leave it as is.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
frankfliu: It looks for MODEL_NAME.pt file in the model directory, if modelName is not specified in the criteria, it assume the model has the same name as the directory.

ylwu-amzn: Thanks, that logic seems not so user friendly as they have to make sure model name matches their `.pt` file name. We add rename logic to make things easier: `file.renameTo(new File(modelPath.resolve(modelName + suffix).toUri()));` . So user can use `.pt` file with any name, we will rename it internally to `<model_name>.pt` when load model. That means we have to make sure there is only one `.pt` file in the zip, otherwise we don't know which `.pt` file we should rename.

frankfliu (approved):

mingshl (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.opensearch.ml.engine.MLEngine.getModelCachePath;
+import static org.opensearch.ml.engine.MLEngine.getUploadModelPath;
+import static org.opensearch.ml.engine.utils.FileUtils.deleteFileQuietly;
+
+/**
+ * A helper class which provides common function for models.
+ */
+@Log4j2
+public class ModelHelper {
+    public static final String PT = ".pt";
```
dhrubo-os, line 19: may be: PYTORCH_FILE_EXTENSION and ONNX_FILE_EXTENSION

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,241 @@
...
+                        }
+                    }
+                    Criteria<Input, Output> criteria = criteriaBuilder.build();
+                    ZooModel<Input, Output> model = criteria.loadModel();
+                    Predictor<Input, Output> predictor = model.newPredictor();
+                    this.predictor = predictor;
+                    this.model = model;
+                    return null;
+                } catch (Exception e) {
+                    String errorMessage = "Failed to load model " + modelName + ", version: " + version;
```
dhrubo-os, line 186: printing model id might be easier for tracking down the issue?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
```diff
@@ -0,0 +1,241 @@
...
+                    log.error(errorMessage, e);
+                    close();
+                    throw new MLException(errorMessage, e);
+                } finally {
+                    deleteFileQuietly(getLoadModelPath(modelId));
```
dhrubo-os, line 191: So are we deleting the model file even if we load the model successfully?

dhrubo-os (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: Yes, the load model path is just temporary folder.

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
ylwu-amzn: Sure, will address this in next PR

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModel.java
ylwu-amzn: Sure, will address this in next PR

PR #484, opened 2022-10-19
Support the generic ml ppl command.

jngz-es:

### Description
Add a new method run() to support the generic ml ppl command.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/448
 
### Check List
  - [ ] All tests pass


client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -169,4 +229,77 @@ private void validateMLInput(MLInput mlInput, boolean requireInput) {
...
+                        BATCH_RCF : FIT_RCF;
+            default:
+                throw new IllegalArgumentException(
+                        String.format("unsupported algorithm: %s.", algo));
+        }
+    }
+
+    private MLAlgoParams convertArgumentToMLParameter(Map<String, Object> arguments,
+                                                        FunctionName func) {
+        switch (func) {
```
ylwu-amzn, line 255: For adding new algorithms, we need to change this client part as well. How about moving this part inside plugin or ml-algorithm , so we keep client part stable.

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -87,6 +89,26 @@ default ActionFuture<MLOutput> train(MLInput mlInput, boolean asyncTask) {
      */
     void train(MLInput mlInput, boolean asyncTask, ActionListener<MLOutput> listener);
 
+    /**
+     * Execute train/predict/trainandpredict.
+     * @param mlInput ML input
+     * @param args algorithm parameters
+     * @return ActionFuture of MLOutput
+     */
+    default ActionFuture<MLOutput> execute(MLInput mlInput, Map<String, Object> args) {
```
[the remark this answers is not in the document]
jngz-es: Renamed to run.

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
jngz-es: Good suggestion, moved to common part as client only depends on common part.

common/src/main/java/org/opensearch/ml/common/input/Constants.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.input;
+
+public class Constants {
```
ylwu-amzn, line 8: Seems this class is duplicate of MLConstants. Move to `common` package?

jngz-es: It is not duplicate. I separated the constants to two parts. The constant like "train" should be only used by the client. Anyway, I can put them together into the commom.

ylwu-amzn (approved):

PR #490, opened 2022-10-20
Add UT for TransportLoadModelAction

zane-neo:
Signed-off-by: Zan Niu <zaniu@amazon.com>

### Description
 
### Issues Resolved
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
@@ -54,6 +54,7 @@ import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 
 @Log4j2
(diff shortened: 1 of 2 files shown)


ylwu-amzn:
hi, @zane-neo Thanks for helping improve code coverage. Seems the code style check failed in Github CI. Suggest run `./gradlew spotlessApply` and publish a new commit.

plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
```diff
@@ -106,130 +107,121 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<LoadMo
...
-                    eligibleNodes.add(node);
-                }
-            }
-            if (nodeIds.size() == 0) {
-                listener.onFailure(new MLResourceNotFoundException("no eligible node found"));
-                return;
             }
+        } else {
+            nodeIds.addAll(allEligibleNodeIds);
+            eligibleNodes.addAll(Arrays.asList(allEligibleNodes));
```
ylwu-amzn, line 129: Thanks for tuning this part. Not sure if this code will perform better than the original one. I guess they are almost the same. But prefer to learn any result if you have done some experiments. Here the allEligibleNodes will be a relatively small number like 100s.

plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
```diff
@@ -106,130 +107,121 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<LoadMo
...
-                        .createTime(Instant.now())
-                        .lastUpdateTime(Instant.now())
-                        .state(MLTaskState.CREATED)
-                        .workerNode(workerNodes)
-                        .build();
-                    mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
-                        String taskId = response.getId();
-                        mlTask.setTaskId(taskId);
-
-                        try {
```
ylwu-amzn, line 163: I think we should keep this try block, this was to catch exception of different parts, will print different exception message and also remove task `mlTaskManager.remove(taskId);`

zane-neo: Added this try catch back.

plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
zane-neo: Checked the code, `Arrays.asList` construct a new object without any iteration, `addAll` method has similar logic with `add` so there's no extra time consuming. Also did a test with 120 records, this approach has faster completion time.

plugin/src/main/java/org/opensearch/ml/action/load/TransportLoadModelAction.java
```diff
@@ -106,134 +107,138 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<LoadMo
...
+                    .workerNode(workerNodes)
+                    .build();
+                mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
+                    String taskId = response.getId();
+                    mlTask.setTaskId(taskId);
+                    try {
+                        mlTaskManager.add(mlTask, nodeIds);
+                        listener.onResponse(new LoadModelResponse(taskId, MLTaskState.CREATED.name()));
+                        threadPool
+                            .executor(TASK_THREAD_POOL)
```
ylwu-amzn, line 164: Tuned thread pool, so there will be conflict for this part

ylwu-amzn (dismissed):

Zhangxunmt (approved):
This TransportLoadModelAction is still in the jacocoExclusions list. Did you check how much percentage for the coverage now in this class?

ylwu-amzn (approved):

Zhangxunmt (approved):

PR #491, opened 2022-10-21
tune interface to support predicting loaded built-in algorithms; remove stale task in cache

ylwu-amzn:

### Description
tune interface to support predicting loaded built-in algorithms;
remove stale task in cache
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
@@ -86,13 +86,13 @@ public class ModelHelper {
                 String fileName = ((ZipEntry) zipEntries.nextElement()).getName();
                 if (fileName.endsWith(PYTORCH_FILE_EXTENSION) || fileName.endsWith(ONNX_FILE_EXTENSION)) {
                     if (hasModelFile) {
-                        throw new IllegalArgumentException("Find multiple model files");
+                        throw new IllegalArgumentException("Find multiple model files, but expected only one");
                     }
                     hasModelFile = true;
                 }
                 if (fileName.equals(TOKENIZER_FILE_NAME)) {
                     if (hasTokenizerFile) {
-                        throw new IllegalArgumentException("Can't find tokenizer files");
+                        throw new IllegalArgumentException("Find multiple tokenizer files");
                     }
                     hasTokenizerFile = true;
                 }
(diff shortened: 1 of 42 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
```diff
@@ -86,13 +86,13 @@ private void verifyModelZipFile(String modelZipFilePath) throws IOException {
                 String fileName = ((ZipEntry) zipEntries.nextElement()).getName();
                 if (fileName.endsWith(PYTORCH_FILE_EXTENSION) || fileName.endsWith(ONNX_FILE_EXTENSION)) {
                     if (hasModelFile) {
-                        throw new IllegalArgumentException("Find multiple model files");
+                        throw new IllegalArgumentException("Find multiple model files, but expected only one");
                     }
                     hasModelFile = true;
                 }
                 if (fileName.equals(TOKENIZER_FILE_NAME)) {
```
dhrubo-os, line 93: how can we have two files with the same name `tokenizer.json`?

plugin/src/main/java/org/opensearch/ml/task/MLTaskManager.java
```diff
@@ -384,4 +389,39 @@ public boolean containsModel(String modelId) {
...
+            MLTask mlTask = entry.getValue().getMlTask();
+            if (mlTask.getTaskType() == MLTaskType.LOAD_MODEL && mlTask.getState() != MLTaskState.CREATED) {
+                runningLoadModelTaskIds.add(entry.getKey());
+            }
+        }
+        return runningLoadModelTaskIds.toArray(new String[0]);
+    }
+
+    public void syncRunningLoadModelTasks(Map<String, Set<String>> runningLoadModelTasks) {
+        Instant ttlEndTime = Instant.now().minus(120, ChronoUnit.SECONDS);
```
Zhangxunmt, line 405: Is this 2 mins too short? If possible, why not use a longer time? It does no harm to keep a task for 10 mins?

plugin/src/main/java/org/opensearch/ml/action/syncup/TransportSyncUpOnNodeAction.java
```diff
@@ -141,9 +144,13 @@ private MLSyncUpNodeResponse createSyncUpNodeResponse(MLSyncUpNodesRequest loadM
             mlModelManager.syncModelRouting(modelRoutingTable);
```
Zhangxunmt, line 144: Is it useful to add a log info here showing an import action is being taken like this "clear Routing Table"? Same for remove worker nodes, add work nodes, etc?

Zhangxunmt (approved):

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/task/MLTaskManager.java
ylwu-amzn: Yes, good point. I used 2 minutes for quick testing. I think we do need to increase to a longer TTL. 10 mins looks good. Will change this part in next PR.

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
ylwu-amzn: Good point, planned to check nested subfolder, should be `endsWith`. Now the code doesn't support nested subfolders in model zip. So we can remove this check now. Will fix this in next PR.

PR #500, opened 2022-10-25
add integ tests for new APIs: upload/load/unload

ylwu-amzn:

### Description
add integ tests for new APIs: upload/load/unload
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


rbhavna (approved):

common/src/main/java/org/opensearch/ml/common/breaker/DiskCircuitBreaker.java
```diff
@@ -33,6 +32,12 @@ public DiskCircuitBreaker(long threshold, String diskDir) {
...
     @Override
     public boolean isOpen() {
-        return (new File(diskDir).getFreeSpace()/1024/1024/1024) < getThreshold();  // in GB
+        try {
+            return AccessController.doPrivileged((PrivilegedExceptionAction<Boolean>) () -> {
+                return (new File(diskDir).getFreeSpace()/1024/1024/1024) < getThreshold();  // in GB
+            });
+        } catch (PrivilegedActionException e) {
+            throw new MLException("Failed to run disk circuit breaker");
+        }
```
Zhangxunmt, line 41: https://github.com/opensearch-project/ml-commons/blob/2.x/plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java#L88. I think the access control added here can be removed. The try catch looks redundant here and needs unnecessary tests coverage.

plugin/src/test/java/org/opensearch/ml/action/custom_model/CustomModelITTests.java
```diff
@@ -0,0 +1,139 @@
...
+            SearchResponse response = searchModelChunks(modelId.get());
+            AtomicBoolean modelChunksReady = new AtomicBoolean(false);
+            if (response != null) {
+                long totalHits = response.getHits().getTotalHits().value;
+                if (totalHits == 9) {
+                    modelChunksReady.set(true);
+                }
+            }
+            return modelChunksReady.get();
+        }, 20, TimeUnit.SECONDS);
```
Zhangxunmt, line 97: Does this 20 seconds adds to the time needed to run integration tests? I see there are multiple 20 seconds waits here, so this means the tests would take very long?

Zhangxunmt (approved):

plugin/src/test/java/org/opensearch/ml/action/custom_model/CustomModelITTests.java
ylwu-amzn: If you check the Github CI below, the build and test steps now takes about 5 minutes, latest 2.x Github [CI flow](https://github.com/opensearch-project/ml-commons/actions/runs/3322861550/jobs/5492471950) shows 4 minutes 10 seconds,

common/src/main/java/org/opensearch/ml/common/breaker/DiskCircuitBreaker.java
ylwu-amzn: This is to avoid catching `PrivilegedActionException` everywhere using this method. We can save some effort by removing this try catch for this method, but will add more efforts for everywhere using this method. If more and more places using this method, we need to catch `PrivilegedActionException` for all of these places and that seems much more effort to do test coverage.

Zhangxunmt: We are talking about the same thing. Check the [PR](https://github.com/opensearch-project/ml-commons/pull/503). I meant to remove the try catch in the TaskRunner because the access control is added in the circuit breaker.

ylwu-amzn: Sorry, misunderstood. Good catch.

PR #501, opened 2022-10-26
add access controller for CB in upload and load API

Zhangxunmt:

### Description
Add access controller for all circuit breakers in upload and load APIs. Currently the permissions are not loaded to check circuit breakers in these APIs.

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -36,6 +36,8 @@ import static org.opensearch.ml.utils.MLNodeUtils.createXContentParserFromRegist
 
 import java.io.File;
 import java.nio.file.Path;
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Base64;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -294,9 +296,20 @@ public String checkAndAddRunningTask(MLTask mlTask, Integer runningTaskLimit) {
         if (error != null) {
             return error;
         }
-        if (mlCircuitBreakerService.isOpen()) {
-            mlStats.getStat(MLNodeLevelStat.ML_NODE_TOTAL_CIRCUIT_BREAKER_TRIGGER_COUNT).increment();
+        try {
+            AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                if (mlCircuitBreakerService.isOpen()) {
```
ylwu-amzn, line 301: Can we move `AccessController.doPrivileged(` to inside of the `isOpen` method? So we don't need to add this for everywhere using `isOpen` method

Zhangxunmt: Yes I think it's doable. This is a quick change complaint with all APIs. Moving to CB level would need revisits to AccessController in other APIs. But overall the change shouldn't be big.

PR #503, opened 2022-10-26
add custom model size limit

Zhangxunmt:

### Description
Add 4GB limit for custom model size. 
Remove redundant access control in taskRunner.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -201,6 +201,9 @@ public class MLModelManager {
                         log.info("create new model meta doc {} for upload task {}", modelId, taskId);
                         modelHelper.downloadAndSplit(modelId, modelName, version, uploadInput.getUrl(), ActionListener.wrap(result -> {
                             Long modelSizeInBytes = (Long) result.get(MODEL_SIZE_IN_BYTES);
+                            if (modelSizeInBytes / 1024 / 1024 / 1024 >= 4L) {
+                                throw new MLException("Failed to save oversize model");
+                            }
                             List<String> chunkFiles = (List<String>) result.get(CHUNK_FILES);
                             String hashValue = (String) result.get(MODEL_FILE_HASH);
                             Semaphore semaphore = new Semaphore(1);
--- a/plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
+++ b/plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
@@ -5,8 +5,6 @@
 
 package org.opensearch.ml.task;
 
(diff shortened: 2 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -201,6 +201,11 @@ public void uploadMLModel(MLUploadInput uploadInput, MLTask mlTask) {
                         log.info("create new model meta doc {} for upload task {}", modelId, taskId);
                         modelHelper.downloadAndSplit(modelId, modelName, version, uploadInput.getUrl(), ActionListener.wrap(result -> {
                             Long modelSizeInBytes = (Long) result.get(MODEL_SIZE_IN_BYTES);
+                            if (modelSizeInBytes/1024/1024/1024 > 4L) {
```
ylwu-amzn, line 204: Will 4.6 GB be truncated as 4GB in this way?

Zhangxunmt: That depends on how accurate we want to limit it. I think >= 4L will resolve your concern. Anything equals to 4 after truncation will be treated as oversize.

ylwu-amzn: If a file size is 4.5 GB, this line will truncate it to 4GB,  it won't throw exception because the truncated file size is not greater than the limit `4GB` ?

Zhangxunmt: I already change to >= 4GB. So 4.5GB will be truncated to 4, which makes >= 4GB is true and throw an exception.

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -201,6 +201,9 @@ public void uploadMLModel(MLUploadInput uploadInput, MLTask mlTask) {
                         log.info("create new model meta doc {} for upload task {}", modelId, taskId);
                         modelHelper.downloadAndSplit(modelId, modelName, version, uploadInput.getUrl(), ActionListener.wrap(result -> {
                             Long modelSizeInBytes = (Long) result.get(MODEL_SIZE_IN_BYTES);
+                            if (modelSizeInBytes / 1024 / 1024 / 1024 >= 4L) {
+                                throw new MLException("Failed to save oversize model");
```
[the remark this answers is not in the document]
Zhangxunmt: sure will change the message to "Model file size exceeds limit of 4GB" in the next PR, so they know what is the limit not to exceed.

b4sjoo (approved):

PR #518, opened 2022-11-03
add trusted URL regex

ylwu-amzn:

### Description
Address security review comments by adding trusted URL regex, so user can set their own trusted URL regex to avoid uploading model from untrusted source.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
@@ -65,13 +65,11 @@ public class RestMLPredictionAction extends BaseRestHandler {
     public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
         String algorithm = request.param(PARAMETER_ALGORITHM);
         String modelId = getParameterId(request, PARAMETER_MODEL_ID);
-        if (modelId == null) {
-            throw new IllegalArgumentException("model id is null");
-        }
 
         if (algorithm != null) {
+            MLPredictionTaskRequest mlPredictionTaskRequest = getRequest(modelId, algorithm, request);
             return channel -> client
-                .execute(MLPredictionTaskAction.INSTANCE, getRequest(modelId, algorithm, request), new RestToXContentListener<>(channel));
+                .execute(MLPredictionTaskAction.INSTANCE, mlPredictionTaskRequest, new RestToXContentListener<>(channel));
         }
 
         return channel -> {
--- a/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
+++ b/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
@@ -24,7 +24,7 @@ public final class MLCommonsSettings {
     public static final Setting<Integer> ML_COMMONS_SYNC_UP_JOB_INTERVAL_IN_SECONDS = Setting
         .intSetting(
             "plugins.ml_commons.sync_up_job_interval_in_seconds",
-            5,
+            10,
             0,
             86400,
             Setting.Property.NodeScope,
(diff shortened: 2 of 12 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -34,7 +34,7 @@ private MLCommonsSettings() {}
     public static final Setting<Long> ML_COMMONS_MONITORING_REQUEST_COUNT = Setting
         .longSetting(
             "plugins.ml_commons.monitoring_request_count",
-            10000,
+            100,
```
Zhangxunmt, line 37: Is this reduction from 10000 to 100 due to the tuning after performance tests result?

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -43,4 +43,12 @@ private MLCommonsSettings() {}
 
     public static final Setting<Integer> ML_COMMONS_MAX_UPLOAD_TASKS_PER_NODE = Setting
         .intSetting("plugins.ml_commons.max_upload_model_tasks_per_node", 10, 0, 10, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
+    public static final Setting<String> ML_COMMONS_TRUSTED_URL_REGEX = Setting
+        .simpleString(
+            "plugins.ml_commons.trusted_url_regex",
+            "^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]",
```
dhrubo-os, line 50: does this regex cover http & https both?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: This part is mainly for testing/monitoring purpose, big value will use more memory. So tune the default value to a smaller number. User can tune it if they want to monitor more requests.

dhrubo-os (approved):
should we have tests for other ways: file, ftp?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: Yes, I tested manually , I will add more test cases in next PR.

jngz-es: Yes, the question mark means 0 - 1.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -65,13 +65,11 @@ public List<Route> routes() {
...
         String modelId = getParameterId(request, PARAMETER_MODEL_ID);
-        if (modelId == null) {
-            throw new IllegalArgumentException("model id is null");
-        }
 
         if (algorithm != null) {
+            MLPredictionTaskRequest mlPredictionTaskRequest = getRequest(modelId, algorithm, request);
             return channel -> client
-                .execute(MLPredictionTaskAction.INSTANCE, getRequest(modelId, algorithm, request), new RestToXContentListener<>(channel));
+                .execute(MLPredictionTaskAction.INSTANCE, mlPredictionTaskRequest, new RestToXContentListener<>(channel));
```
Zhangxunmt, line 72: Is this because the JVM complier is not able to dynamic resolve getRequest() before executing  client.execute()? If that's the case, we need to use another local variable for line 83. https://github.com/opensearch-project/ml-commons/blob/2.x/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java#L83

ylwu-amzn: just for easy testing

PR #632, opened 2022-12-07
Integrate with sageMaker prototype

Zhangxunmt:

### Description
This PR is the prototype to send request to sageMaker model inference endpoint. Currently for now it is primarily for discussion use. DO NOT merge until it's finalized. 

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/.whitesource
+++ b/.whitesource
@@ -1,0 +1,22 @@
+{
+  "scanSettings": {
+    "configMode": "AUTO",
+    "configExternalURL": "",
+    "projectToken": "",
+    "baseBranches": []
+  },
+  "checkRunSettings": {
+    "vulnerableCheckRunConclusionLevel": "failure",
+    "displayMode": "diff",
+    "useMendCheckNames": true
+  },
+  "issueSettings": {
+    "minSeverityLevel": "LOW",
+    "issueType": "DEPENDENCY"
+  },
+  "remediateSettings": {
+    "workflowRules": {
+      "enabled": true
+    }
+  }
+}


common/src/main/java/org/opensearch/ml/common/transport/remote/MLRemoteModelInput.java
```diff
@@ -0,0 +1,177 @@
...
+    public static final String DESCRIPTION_FIELD = "description";
+    public static final String VERSION_FIELD = "version";
+    public static final String ENDPOINT = "endpoint";
+    public static final String ACCESS_KEY_ID = "access_key_id";
+    public static final String SECRET_ACCESS_KEY = "secret_access_key";
+
+    private String remoteModelName;
+    private String description;
+    private String version;
+    private String endpoint;
```
wujunshen, line 31: from line 31-33,these 3 variable in input class,If we used those in input parameters, whatever in rest or tcp transfer case,is it will trigger security problem? I remember that things like id&key can't be passed in and out as plain text in the parameters, and it's not correct if their values are written in the code(will it be written in yml file?)

plugin/build.gradle
```diff
@@ -51,6 +51,14 @@ dependencies {
...
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.19.0"
+    // implementation group: 'software.amazon.awssdk', name: 'sagemaker', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'sagemakerruntime', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'auth', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'sdk-core', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'annotations', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'aws-core', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'regions', version: '2.18.21'
+    implementation group: 'software.amazon.awssdk', name: 'utils', version: '2.18.21'
     testImplementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
```
wujunshen, line 62: It is true that in this scenario we can use our own company sdk package. But if we want to access other third parties, we should not rely on their sdk packages directly in the ml-common code. This would involve the question of whether the code is allowed to be used between companies, has this been considered before?

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -185,6 +198,83 @@ protected void executeTask(MLPredictionTaskRequest request, ActionListener<MLTas
         }
     }
 
+    private void remoteInference(String modelId, MLTask mlTask, MLInput mlInput, ActionListener<MLTaskResponse> listener) {
```
wujunshen, line 201: I think this private method can be abstracted into a template class, which can be directly reused by other third-party applications performing remote inference operations.

Issue #577, opened 2022-11-28
[FEATURE] Auto reload model when cluster rebooted/node rejoin


ylwu-amzn:
**Is your feature request related to a problem?**
Refer to https://forum.opensearch.org/t/feedback-machine-learning-model-serving-framework-experimental-release/11439/17

> I have noticed that models are not automatically loaded when the OpenSearch cluster is rebooted. Is there a way to do that? It would be impractical to manually reload models each time the cluster is rebooted, especially in a containerized setup.

**What solution would you like?**
We track each model's state in model index. When node restarted/rejoin cluster, check model state and reload model on corresponding ML node.

**What alternatives have you considered?**
NO

**Do you have any additional context?**
N/A

PR #639, opened 2022-12-19
the dev of  [FEATURE]Auto reload model when cluster rebooted/node rejoin

wujunshen:
### Description
the new feature: 
 
### Issues Resolved

### Check List
  - [x] All tests pass


ylwu-amzn:
Thanks for adding this feature. Can you add more details in description?
BTW, please make sure `./gradlew build` can pass locally before publishing PR.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -345,6 +368,13 @@ public Collection<Object> createComponents(
             nodeHelper
         );
 
+        try {
+            CompletableFuture<Void> completableFuture = CompletableFuture.runAsync(() -> mlModelAndNodeManager.autoReLoadModel());
+            completableFuture.get();
```
zane-neo, line 373: Do not use blocking approach in our code, always use callbacks to not blocking the main thread.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -103,18 +103,18 @@
 import com.google.common.io.Files;
 
 /**
- * Manager class for ML models. It contains ML model related operations like upload, load model etc.
+ * Manager class for ML models. It contains ML model related operations like upload, load model
```
zane-neo, line 106: Why this file is modified?

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        this.mlIndicesHandler = mlIndicesHandler;
+        this.mlTaskManager = mlTaskManager;
+        this.mlEngine = mlEngine;
+
+        enableAutoReLoadModel = ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE.get(settings);
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE, it -> enableAutoReLoadModel = it);
+    }
+
```
zane-neo, line 122: Using a new thread and pass necessary variables to this thread to execute autoReload action, after the retry logic, the thread should be destroyed.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
wujunshen: my bad,the java format of this file is changed accidentally

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        // According to the node id to query the number of retries, if more than 2 (the maximum number of retries), we do not need to retry,
+        // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+        Integer reTryTimes = 0;
+        if (isExistedIndex(ML_MODEL_RELOAD_INDEX)) {
+            reTryTimes = getReTryTimes(localNodeId);
+        }
+        if (reTryTimes > ML_MODEL_RELOAD_MAX_RETRY_TIMES) {
+            throw new RuntimeException("have exceeded max retry times, always failure");
```
zane-neo, line 145: No need to throw exception, instead we should log the event for user's convenience.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        // auto reload all models of this local node, if it fails, reTryTimes+1, if it succeeds, reTryTimes is cleared to 0
+        autoReLoadModelByNodeId(localNodeId);
+    }
+
+    @VisibleForTesting
+    void autoReLoadModelByNodeId(String nodeId) {
+        SearchRequest searchRequest = new SearchRequest(ML_TASK_INDEX);
+        SearchResponse response = client.execute(SearchAction.INSTANCE, searchRequest).actionGet();
```
zane-neo, line 155: Change to asynchronous approach.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+
+                if (workerNode.contains(nodeId) && mlTaskType == MLTaskType.LOAD_MODEL && mlTaskState == MLTaskState.COMPLETED) {
+                    autoReLoadModelByNodeAndModelId(nodeId, mlTask.getModelId());
+                }
+            } catch (Exception e) {
+                reTryTimes++;
+                // Store the latest value of the reTryTimes and node id under the index ".plugins-ml-model-reload"
+                saveLatestReTryTimes(nodeId, reTryTimes);
+                log.error("ML Task id {} failed to parse, the reason is: {}", hit.getId(), e);
+                throw new RuntimeException(e);
```
zane-neo, line 187: No need to throw exception and `saveLatestRetryTimes` can be delete from here as well.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /**
+     *
+     * @param nodeId
+     * @param modelId
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeAndModelId(String nodeId, String modelId) {
+        MLLoadModelRequest mlLoadModelRequest = new MLLoadModelRequest(modelId, new String[] { nodeId }, false, false);
```
zane-neo, line 202: We should add nodeIds here, since this function `client.execute(MLLoadModelAction.INSTANCE, ...)` will collect all ml nodes by default, this can cause unnecessary network calls.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        LoadModelResponse loadModelResponse = client.execute(MLLoadModelAction.INSTANCE, mlLoadModelRequest).actionGet();
+
+        if (loadModelResponse == null) {
+            throw new RuntimeException("Can't auto reload model " + modelId + " by node " + nodeId);
```
zane-neo, line 206: No need throw exception

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        if (response == null) {
+            throw new RuntimeException("can't get retry times by " + nodeId);
+        }
+        SearchHits searchHits = response.getHits();
+        if (searchHits == null) {
+            throw new RuntimeException("can't get retry times by " + nodeId);
+        }
+        SearchHit[] hits = searchHits.getHits();
+        if (CollectionUtils.isEmpty(hits)) {
+            throw new RuntimeException("can't get retry times by " + nodeId);
```
zane-neo, line 228: If no record exist, we should treat the retry time as 0 instead of throw exception

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        for (SearchHit hit : hits) {
+            Map<String, Object> sourceAsMap = hit.getSourceAsMap();
+            return (Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD);
```
zane-neo, line 233: Do we need a force converting on this? Can we use an object and parse the result to object for better reading experience.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        String localNodeId = clusterService.localNode().getId();
+        DiscoveryNode node = nodeHelper.getNode(localNodeId);
+        if (!MLNodeUtils.isMLNode(node)) {
+            return;
+        }
+
+        // According to the node id to query the number of retries, if more than 2 (the maximum number of retries), we do not need to retry,
+        // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+        Integer reTryTimes = 0;
+        if (isExistedIndex(ML_MODEL_RELOAD_INDEX)) {
```
zane-neo, line 141: Do we need a specific check of existing? Since getRetryTimes also query on this index, we can use the http code like `404` to figure out if the index exists or not.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        return exists.isExists();
+    }
+
+    @VisibleForTesting
+    void saveLatestReTryTimes(String nodeId, Integer reTryTimes) {
+        Map<String, Object> content = new HashMap<>();
+        content.put(NODE_ID_FIELD, nodeId);
+        content.put(MODEL_LOAD_RETRY_TIMES_FIELD, reTryTimes);
+
+        if (isExistedIndex(ML_MODEL_RELOAD_INDEX)) {
```
zane-neo, line 253: No need to check existence of the index, adding a new document to an un-exist index could trigger the auto-creation of the index.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+            UpdateResponse updateResponse = client.execute(UpdateAction.INSTANCE, updateRequest).actionGet(5000);
+
+            if (updateResponse.status() == RestStatus.OK) {
+                log.debug("node id:{} update retry times successfully", nodeId);
+            } else {
+                throw new RuntimeException("can't update retry times by " + nodeId);
+            }
+        } else {
+            // create index and store data
+            CreateIndexRequest createIndexRequest = new CreateIndexRequest(ML_MODEL_RELOAD_INDEX);
```
zane-neo, line 270: No need to create index explicitly

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
+        String localNodeId = clusterService.localNode().getId();
+        DiscoveryNode node = nodeHelper.getNode(localNodeId);
+        if (!MLNodeUtils.isMLNode(node)) {
+            return;
+        }
+
+        // According to the node id to query the number of retries, if more than 2 (the maximum number of retries), we do not need to retry,
+        // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+        Integer reTryTimes = 0;
```
zane-neo, line 140: why we use object instead of primitive value?

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @VisibleForTesting
+    void saveLatestReTryTimes(String nodeId, Integer reTryTimes) {
+        Map<String, Object> content = new HashMap<>();
+        content.put(NODE_ID_FIELD, nodeId);
+        content.put(MODEL_LOAD_RETRY_TIMES_FIELD, reTryTimes);
+
+        if (isExistedIndex(ML_MODEL_RELOAD_INDEX)) {
+            // update data under the index
+            UpdateRequest updateRequest = new UpdateRequest();
```
zane-neo, line 255: We can try always using IndexRequest with docId (if exist), this is a upsert operation.

zane-neo:
Please review the Exceptions thrown, we should not throw any exception to outside of the auto reload logic.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -26,129 +26,145 @@ public class CommonValue {
...
+    public static final String MODEL_LOAD_RETRY_TIMES_FIELD = "retry_times";
+
+    public static final String ML_MODEL_RELOAD_INDEX_MAPPING = "{\n"
+        + "    \"properties\": {\n"
+        + "      \""
+        + NODE_ID_FIELD
+        + "\" : {\"type\": \"keyword\"},\n"
+        + "      \""
+        + MODEL_LOAD_RETRY_TIMES_FIELD
+        + "\" : {\"type\": \"long\"}\n"
```
zane-neo, line 56: we do we use long type?

zane-neo: How do we differentiate different model's reload times?

plugin/build.gradle
```diff
@@ -52,6 +51,7 @@ dependencies {
     implementation group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.19.0"
+    implementation "com.github.rholder:guava-retrying:2.0.0"
```
zane-neo, line 54: What this for?

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
```diff
@@ -0,0 +1,294 @@
...
+        if (searchHits == null) {
+            return;
+        }
+        SearchHit[] hits = searchHits.getHits();
+        if (CollectionUtils.isEmpty(hits)) {
+            return;
+        }
+
+        Integer reTryTimes = 0;
+        for (SearchHit hit : hits) {
```
zane-neo, line 170: Instead try each record in the result, it's better to aggregate the result first to figure out which models should be reloaded and then arrange several parallel tasks to accomplish this.

plugin/build.gradle
wujunshen: have rollback

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: u can see ` new String[] { nodeId }` the string array is just has 1 element,it is our node that want to do reload operation,so it will not cause the effort of  model reloading by other node ids

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
wujunshen: move in ModelAutoReloader class

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: I think we can remove exception code,but the saveLatestRetryTimes method should be here.
otherwise,we don't know how many times we have failed to auto reload model.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: u can see the code above:  
```
 SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
        searchSourceBuilder.fetchSource(new String[] { MODEL_LOAD_RETRY_TIMES_FIELD }, null);
```
I have let the source returned 1 field,it is integer type.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: we need when the first time to reload model. at that time, the `ML_MODEL_RELOAD_INDEX` didn't exist,so wee haven't saved reTryTimes before that time.

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: if return object is Objected, can be casted easily

common/src/main/java/org/opensearch/ml/common/CommonValue.java
wujunshen: it should be "integer" type, I will modify it.

wujunshen: > How do we differentiate different model's reload times?

we don't differentiate different model. If a node halt down with unknown reason, it will reload all the model under this stopped node. so what we should do is to reload all the model under broaken node

plugin/src/main/java/org/opensearch/ml/model/MLModelAndNodeManager.java
wujunshen: see the comment above.

build.gradle
```diff
@@ -72,7 +72,7 @@ ext {
 }
 
 dependencies {
-    implementation 'junit:junit:${versions.junit}'
+    implementation 'junit:junit:4.13.1'
```
zane-neo, line 75: We shouldn't modify this.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -26,129 +26,135 @@ public class CommonValue {
     public static String WARM_BOX_TYPE = "warm";
 
     public static final String ML_MODEL_INDEX = ".plugins-ml-model";
+    public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
```
zane-neo, line 29: Where is the schema of this index?

ml-algorithms/build.gradle
```diff
@@ -41,7 +41,7 @@ dependencies {
 }
 
 configurations.all {
-    resolutionStrategy.force 'com.google.protobuf:protobuf-java:3.21.9'
+    resolutionStrategy.force 'com.google.protobuf:protobuf-java:3.21.7'
```
zane-neo, line 44: Why protobuf version is downgraded?

plugin/build.gradle
```diff
@@ -240,7 +238,11 @@ List<String> jacocoExclusions = [
         'org.opensearch.ml.profile.MLModelProfile',
         'org.opensearch.ml.profile.MLPredictRequestStats',
         'org.opensearch.ml.action.load.TransportLoadModelAction',
+        'org.opensearch.ml.action.load.TransportLoadModelOnNodeAction',
         'org.opensearch.ml.model.MLModelManager',
+        'org.opensearch.ml.action.unload.TransportUnloadModelAction',
+        'org.opensearch.ml.action.forward.TransportForwardAction',
+        'org.opensearch.ml.action.syncup.TransportSyncUpOnNodeAction',
```
zane-neo, line 245: Why adding these files to exclusions?

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -470,8 +481,25 @@ public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
             ML_THREAD_POOL_PREFIX + PREDICT_THREAD_POOL,
             false
         );
+        FixedExecutorBuilder reloadModelThreadPool = new FixedExecutorBuilder(
```
zane-neo, line 484: We don't need to create a thread pool to accomplish this, we can reuse the general thread pool in opensearch or even a simple Thread would be sufficient.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+     * @param nodeId node id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeId(String nodeId) {
+        if (!isExistedIndex(ML_TASK_INDEX)) {
+            return;
+        }
+
+        SearchRequest searchRequest = new SearchRequest(ML_TASK_INDEX);
+        SearchResponse response = client.execute(SearchAction.INSTANCE, searchRequest).actionGet();
```
zane-neo, line 160: How many records we can get by default with this invocation?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+
```
zane-neo, line 161: Three checks can be merged into one if check.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+     */
+    public void autoReLoadModel() {
+        log.debug("enableAutoReLoadModel {} ", enableAutoReLoadModel);
+
+        // if we don't need to reload automatically, just return without doing anything
+        if (!enableAutoReLoadModel) {
+            return;
+        }
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
zane-neo, line 117: Why do we need this here?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+        if (searchHits == null) {
+            return;
+        }
+        SearchHit[] hits = searchHits.getHits();
+        if (CollectionUtils.isEmpty(hits)) {
+            return;
+        }
+
+        Integer reTryTimes = 0;
+        for (SearchHit hit : hits) {
```
zane-neo, line 175: We shouldn't retry for every task, we should only retry the latest success one.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     *  auto reload 1 model under the node id
+     * @param nodeId node id
+     * @param modelId model id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeAndModelId(String nodeId, String modelId) {
+        MLLoadModelRequest mlLoadModelRequest = new MLLoadModelRequest(modelId, new String[] { nodeId }, false, false);
+        client.execute(MLLoadModelAction.INSTANCE, mlLoadModelRequest).actionGet();
```
zane-neo, line 207: Can we just ignore the response or should we check the load result?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * get retry times from the index ".plugins-ml-model-reload" by 1 ml node
+     * @param nodeId the filter condition to query
+     * @return retry times
+     */
+    @VisibleForTesting
+    Integer getReTryTimes(String nodeId) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.fetchSource(new String[] { MODEL_LOAD_RETRY_TIMES_FIELD }, null);
+        QueryBuilder queryBuilder = QueryBuilders.termQuery("_id", nodeId);
```
zane-neo, line 219: If nodeId is the document id, I suggest to use GetRequest with id to query rather than term query.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+            SearchHits searchHits = response.getHits();
+            if (searchHits == null) {
+                throw new RuntimeException("can't get retry times by " + nodeId);
+            }
+            SearchHit[] hits = searchHits.getHits();
+            if (CollectionUtils.isEmpty(hits)) {
+                return 0;
+            }
+
+            for (SearchHit hit : hits) {
```
zane-neo, line 238: Check the hits size and if not expected then throw exception rather than throw at the bottom of the method.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /**
+     * auto reload all the models under the node id
+     * the node must be a ml node
+     * @param nodeId node id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeId(String nodeId) {
+        if (!isExistedIndex(ML_TASK_INDEX)) {
```
zane-neo, line 155: Is the index created automatically by opensearch?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
wujunshen: the method of creating index and saving data can be seen in  MLModelAutoReLoader#saveLatestReTryTimes

build.gradle
wujunshen: it will report error when run "./gradlew build"

plugin/build.gradle
wujunshen: these are  original content,

ml-algorithms/build.gradle
wujunshen: in github, When I commit code, there will be the dependency check and I changed it according to the tip

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
wujunshen: yes, I have removed it

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: not all, so there will be modified, I will changed code

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I removed it, too

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: ` if (!isExistedIndex(ML_TASK_INDEX)) {
            return;
        }`
this check shouldn't be removed, because I found the code will throw the  IndexNotFoundException if I didn't check whether ML_TASK_INDEX exised.
And other checks have been merged into 1 by me

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: in the iteration, I add the condition
` if (workerNode.contains(nodeId) && mlTaskType == MLTaskType.LOAD_MODEL && mlTaskState == MLTaskState.COMPLETED) {
                    autoReLoadModelByNodeAndModelId(nodeId, mlTask.getModelId());
                }`
if the condition didn't be meet, it will not do retry operation.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: it is a reused method, I call the model load method, if the process of model reload throws the exception, this method also will throw the exception.So we don't make a check by myself.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: yes, u r right~I will changed it to `idsQuery` method

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: because I will use `idsQuery`, so the result will return a distinct result, the size of hits will be 0 or 1,and before this code,I check the situation of 0,so in this code,the size of hits must be 1

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: At the begin, I think it should be created automatically,but when I wrote the unit test,
I found the code will throw the `IndexNotFoundException` if I didn't check whether `ML_TASK_INDEX` existed.So I add this check.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I found 3 field "task_type","state" and "worker_node" are all `keyword` type.
So I think I can use `termQuery` to make search condition `task_type=LOAD_MODEL and state =COMPLETED` but field of "worker_node" is a string separated by comma. So it will be hard to use `termQuery`, this code I will make a test to see how to make an accurate query and to ensure smaller numbers for query records

build.gradle
zane-neo: Did you rebase the code on main branch?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * save retry times
+     * @param nodeId node id
+     * @param reTryTimes actual retry times
+     */
+    @VisibleForTesting
+    void saveLatestReTryTimes(String nodeId, Integer reTryTimes) {
+        Map<String, Object> content = new HashMap<>();
+        content.put(NODE_ID_FIELD, nodeId);
+        content.put(MODEL_LOAD_RETRY_TIMES_FIELD, reTryTimes);
```
zane-neo, line 238: We need model_id here based on the discussion.

plugin/build.gradle
zane-neo: you should rebase the main branch first to avoid the confused change shown in your PR.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: UT doesn't show the fact of the cluster behavior, please confirm this. If this is created automatically it's fine to add this check otherwise the auto reload logic will never be reachable.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: Even if with this check, we shouldn't auto reload for every document that meets the criteria, think of user did two deployment, and the first time deployment is not the correct status anymore, we should always focus on the latest successful deployment.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: Add assertion here or using `hits.get(0)` instead of for loop and leaking condition to the end of the method.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.fetchSource(new String[] { MODEL_LOAD_RETRY_TIMES_FIELD }, null);
+        QueryBuilder queryBuilder = QueryBuilders.idsQuery().addIds(nodeId);
+        searchSourceBuilder.query(queryBuilder);
+        SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(ML_MODEL_RELOAD_INDEX);
+
+        try {
+            SearchResponse response = client.execute(SearchAction.INSTANCE, searchRequest).actionGet(5000);
```
zane-neo, line 199: Avoid using synchronous blocking wait programming paradigm, using listener approach to ensure the coding format is unified.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -26,129 +26,134 @@ public class CommonValue {
...
-            + "\": {\n"
-            + "        \"type\": \"nested\",\n"
-            + "        \"properties\": {\n"
-            + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
-            + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-            + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-            + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
-            + "        }\n"
-            + "      }\n";
+        + CommonValue.USER
```
zane-neo, line 35: Format your code before commit to avoid change the format, this isn't your change.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,253 @@
...
+
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
+        String localNodeId = clusterService.localNode().getId();
+        if (!MLNodeUtils.isMLNode(nodeHelper.getNode(localNodeId))) {
+            return;
+        }
+
+        // According to the node id to get retry times, if more than the maxi retry times, don't need to retry
+        // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+        Integer reTryTimes = getReTryTimes(localNodeId);
```
zane-neo, line 111: Basically we use encapsulated type in class fields, and in methods we usually use primitive type to avoid null checking logic. The retryTimes is totally controlled by ourself, so it makes sense to use primitive type instead of encapsulated type.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,253 @@
...
+        if (reTryTimes > ML_MODEL_RELOAD_MAX_RETRY_TIMES) {
+            log.debug("have exceeded max retry times, always failure");
+        }
+
+        // auto reload all models of this local node, if it fails, reTryTimes+1, if it succeeds, reTryTimes is cleared to 0
+        CompletableFuture<Void> completableFuture = CompletableFuture.runAsync(() -> autoReLoadModelByNodeId(localNodeId));
+
+        try {
+            completableFuture.get();
```
zane-neo, line 120: Blocking approach could do harm to the whole cluster performance, avoid this.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -345,6 +352,8 @@ public Collection<Object> createComponents(
             nodeHelper
         );
 
+        mlModelAutoReLoader.autoReLoadModel();
```
zane-neo, line 355: Underlying code is blocking the method's return, this shouldn't happen.

ml-algorithms/build.gradle
zane-neo: I also have a PR to ml-commons, and I didn't change this, no error raised: https://github.com/opensearch-project/ml-commons/pull/645

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: completableFuture has been called by runAsync,so it must be an async method

ml-algorithms/build.gradle
wujunshen: I don't know what happen,but I merge the code from main branch. It didn't pass the check ,so I changed it.

build.gradle
wujunshen: yes, of course, my code is forked from the latest code under the main branch

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: if you remove the check, it will report error

ylwu-amzn:
Just realize your PR is based on main branch. Please cut PR to 2.x branch. main branch is using 3.0 which may have breaking change. When PR merged to 2.x, you should forward port it to main branch

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: cool, it is ok. I will changed it.

build.gradle
wujunshen: rollback

ml-algorithms/build.gradle
wujunshen: rollback

plugin/build.gradle
wujunshen: rollback

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I am start to understand your means. I just to get 1 doc from index `ML_TASK_INDEX`, is right?

PR #650, opened 2022-12-28
the dev of [FEATURE]Auto reload model when cluster rebooted/node rejoin

wujunshen:

### Description
the new feature: 
 
### Issues Resolved
 
### Check List
  - [x] All tests pass


common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -29,132 +26,134 @@ public class CommonValue {
...
-            + "      \""
-            + MLTask.ERROR_FIELD
-            + "\": {\"type\": \"text\"},\n"
-            + "      \""
-            + MLTask.IS_ASYNC_TASK_FIELD
-            + "\" : {\"type\" : \"boolean\"}, \n"
-            + USER_FIELD_MAPPING
-            + "    }\n"
-            + "}";
+        + "    \"_meta\": {\"schema_version\": "
```
zane-neo, line 116: Reformat this file.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -345,6 +352,8 @@ public Collection<Object> createComponents(
             nodeHelper
         );
 
+        mlModelAutoReLoader.autoReLoadModel();
```
zane-neo, line 355: Do this in the MLModelAutoReloader constructor.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        String localNodeId = clusterService.localNode().getId();
+        if (!MLNodeUtils.isMLNode(nodeHelper.getNode(localNodeId))) {
+            return;
+        }
+
+        // According to the node id to get retry times, if more than the maxi retry times, don't need to retry
+        // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+        int reTryTimes = getReTryTimes(localNodeId);
+        if (reTryTimes > ML_MODEL_RELOAD_MAX_RETRY_TIMES) {
+            log.debug("have exceeded max retry times, always failure");
```
zane-neo, line 114: Add return here.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        // auto reload all models of this local node, if it fails, reTryTimes+1, if it succeeds, reTryTimes is cleared to 0
+        CompletableFuture.runAsync(() -> autoReLoadModelByNodeId(localNodeId));
```
zane-neo, line 118: Use generic thread pool, better not introduce new thread pools if it's really necessary.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeId(String nodeId) {
+        if (!isExistedIndex(ML_TASK_INDEX)) {
+            return;
+        }
+
+        SearchResponse response = queryTask(nodeId);
+
+        SearchHit[] hits = response.getHits().getHits();
```
zane-neo, line 134: Response could be null or not? As well as response.getHits().

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        if (CollectionUtils.isEmpty(hits)) {
+            return;
+        }
+
+        if (hits.length != 1) {
+            log.error("Can't get the latest successful deployment in node id {}", nodeId);
```
zane-neo, line 140: add return here.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().from(0).size(1);
+
+        QueryBuilder queryBuilder = QueryBuilders
+            .boolQuery()
+            .must(QueryBuilders.matchPhraseQuery("task_type", "LOAD_MODEL"))
+            .must(QueryBuilders.matchPhraseQuery("state", "COMPLETED"))
+            .must(QueryBuilders.matchPhraseQuery("worker_node", nodeId));
+        searchSourceBuilder.query(queryBuilder);
+
+        SortBuilder sortBuilderOrder = new FieldSortBuilder("last_update_time").order(SortOrder.DESC);
```
zane-neo, line 177: When a task will be updated?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        searchSourceBuilder.sort(sortBuilderOrder);
+
+        SearchRequest searchRequest = new SearchRequest(ML_TASK_INDEX);
+        searchRequest.source(searchSourceBuilder).indices(ML_TASK_INDEX);
```
zane-neo, line 181: why adding same index two times?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        try {
+            return client.execute(SearchAction.INSTANCE, searchRequest).actionGet(5000);
```
zane-neo, line 183: Avoid blocking programming paradigm.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        } catch (IndexNotFoundException e) {
+            throw new IndexNotFoundException("index " + ML_TASK_INDEX + " not found");
```
zane-neo, line 185: Add logging here.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     *  auto reload 1 model under the node id
+     * @param nodeId node id
+     * @param modelId model id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeAndModelId(String nodeId, String modelId) {
+        MLLoadModelRequest mlLoadModelRequest = new MLLoadModelRequest(modelId, new String[] { nodeId }, false, false);
```
zane-neo, line 196: Use all nodeIds fetched from task index, otherwise this can break other functionalities.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        client.execute(MLLoadModelAction.INSTANCE, mlLoadModelRequest).actionGet(5000);
```
zane-neo, line 197: The result is ignored, what if no exception thrown but the load still failed?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+        int reTryTimes = 0;
+        try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, hits[0].getSourceRef())) {
+            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+            MLTask mlTask = MLTask.parse(parser);
+
+            autoReLoadModelByNodeAndModelId(nodeId, mlTask.getModelId());
+        } catch (RuntimeException | IOException e) {
+            reTryTimes++;
+            // Store the latest value of the reTryTimes and node id under the index ".plugins-ml-model-reload"
+            saveLatestReTryTimes(nodeId, reTryTimes);
```
zane-neo, line 152: Add model id here.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+            log.error("Can't auto reload model in node id {} ,has try {} times\nThe reason is:{}", nodeId, reTryTimes, e.getMessage());
+        }
+
+        // Store the latest value of the reTryTimes and node id under the index ".plugins-ml-model-reload"
+        saveLatestReTryTimes(nodeId, reTryTimes);
```
zane-neo, line 157: What if the reload successfully? Do we still need to save retry times or clear it up.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,273 @@
...
+     */
+    @VisibleForTesting
+    int getReTryTimes(String nodeId) {
+        if (!isExistedIndex(ML_MODEL_RELOAD_INDEX)) {
+            return 0;
+        }
+
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.fetchSource(new String[] { MODEL_LOAD_RETRY_TIMES_FIELD }, null);
+        QueryBuilder queryBuilder = QueryBuilders.idsQuery().addIds(nodeId);
```
zane-neo, line 213: We shouldn't use nodeId, a possible solution is to use nodeId + modelId as the document id.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: Since we'll support node level auto reload only, resolving this.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: node level auto reload no need model id, resolving this.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
wujunshen: it has been the result of reformat. I use spotlessJavaApply task

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I don't understand that why we use thread pool. just use the feature of CompletableFuture, we shouldn't care about the implement logic inside it.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: added it

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: has fixed

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: mark,let me think for a while

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: in this situation, we should clear `retrytimes` to zero when we reload successfully,so we use  `saveLatestReTryTimes(nodeId, reTryTimes);`to save 0 into it

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: in this situation, the method of model load is just invoked by me. It should assure that will throw exception when loading failed by itself

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: but our aim is to reload model on 1 node halt down with unknown reason. so the other nodes may be run model normally. use all nodes to load, will corrupt the status of models on the other nodes.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: It is not a block. u can see many similar write-ups in the ml-commons

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: This method will be invoked twice if entered the catch block, remove the invocation in catch block.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: I didn't see `actionGet` been invoked anywhere except the test classes, please show me an example.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: When logging exception, log the exception itself like `log.error("xxxxxx", e)`

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: This won't corrupt the status of other node's models, and if not doing this, this will break the profile api functionality, because profile api reads ml task index to get the expected deployment plan.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,271 @@
...
+        searchSourceBuilder.query(queryBuilder);
+        SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(ML_MODEL_RELOAD_INDEX);
+
+        try {
+            SearchResponse response = client.execute(SearchAction.INSTANCE, searchRequest).actionGet(5000);
+
+            SearchHit[] hits = response.getHits().getHits();
```
zane-neo, line 222: null check?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,271 @@
...
+            if (CollectionUtils.isEmpty(hits)) {
+                return 0;
+            }
+
+            Map<String, Object> sourceAsMap = hits[0].getSourceAsMap();
+            return (Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD);
+        } catch (IndexNotFoundException e) {
+            throw new IndexNotFoundException("index " + ML_MODEL_RELOAD_INDEX + " not found");
```
zane-neo, line 230: add log here.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,271 @@
...
+
+        IndexRequest indexRequest = new IndexRequest(ML_MODEL_RELOAD_INDEX);
+        indexRequest.id(nodeId);
+        indexRequest.source(content);
+        indexRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+        IndexResponse indexResponse = client.execute(IndexAction.INSTANCE, indexRequest).actionGet(5000);
+
+        if (indexResponse.status() == RestStatus.CREATED || indexResponse.status() == RestStatus.OK) {
+            log.debug("node id:{} insert retry times successfully", nodeId);
```
zane-neo, line 266: why using debug level here? Change it to info.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: The reason to reuse an existing thread pool is to reduce the resource consumption, this is the generic thread pool in opensearch: `threadPool.generic()`, threadPool in `MachineLearningPlugin`.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,271 @@
...
+     * the node must be a ml node
+     * @param nodeId node id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeId(String nodeId) {
+        if (!isExistedIndex(ML_TASK_INDEX)) {
+            return;
+        }
+
+        SearchResponse response = queryTask(nodeId);
```
zane-neo, line 136: Can the queryTask get an IndexNotFoundException? Can we leverage this and reduce the isExistedIndex call?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,271 @@
...
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE, it -> enableAutoReLoadModel = it);
+
+        autoReLoadModel();
+    }
+
+    /**
+     * the main method: model auto reloading
+     */
+    public void autoReLoadModel() {
```
zane-neo, line 99: Make this private since we don't expect other classes to use this directly.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: got it

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: it we throw the exception, it is no need to log again

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: except in try-catch, we just use debug to min the effort by log

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: in queryTask, there have been catch IndexNotFoundException

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: ok

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: Suddenly, I think it no need to catch `IndexNotFoundException`. Because I have checked index whether existed at the begin of this method when the method is invoked at the first time. So in the part of query executed, the index must have existed~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I wrote `threadPool.generic().submit(() -> autoReLoadModelByNodeId(localNodeId));`, is it right?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: yes,we used it when we invoked action.

common/src/main/java/org/opensearch/ml/common/MLTask.java
```diff
@@ -262,6 +269,9 @@ public static MLTask parse(XContentParser parser) throws IOException {
                 case IS_ASYNC_TASK_FIELD:
                     async = parser.booleanValue();
                     break;
+                    case IS_AUTO_RELOAD_FIELD:
```
zane-neo, line 272: looks like this is a format issue.

common/src/main/java/org/opensearch/ml/common/MLTask.java
```diff
@@ -282,6 +292,7 @@ public static MLTask parse(XContentParser parser) throws IOException {
                 .error(error)
                 .user(user)
                 .async(async)
+            .autoReload(autoReload)
```
zane-neo, line 295: format

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
zane-neo: Can we move the `CollectionUtils.isEmpty(hits)` to above check together?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     *  auto reload 1 model under the node id
+     * @param localNodeId node id
+     * @param modelId model id
+     */
+    @VisibleForTesting
+    void autoReLoadModelByNodeAndModelId(String localNodeId, String modelId) throws IllegalArgumentException {
+        MLLoadModelRequest mlLoadModelRequest = new MLLoadModelRequest(modelId, new String[] { localNodeId }, false, false, true);
+        AtomicReference<LoadModelResponse> loadModelResponse = new AtomicReference<>();
```
zane-neo, line 177: Why do we need `AtomicReference` instead of `Supplier`, this method doesn't seems to be race conditioned.

model-collapse (changes requested):
some changes required

MAINTAINERS.md
```diff
@@ -1,8 +1,13 @@
...
 
-## Maintainers
-| Maintainer | GitHub ID | Affiliation |
-| --------------- | --------- | ----------- |
-| Hanguang Zhang | [zhanghg08](https://github.com/zhanghg08) | Amazon |
-| Jing Zhang | [jngz-es](https://github.com/jngz-es) | Amazon |
-| Weicong Sun | [weicongs-amazon](https://github.com/weicongs-amazon) | Amazon |
+This document contains a list of maintainers in this repo. See [opensearch-project/.github/RESPONSIBILITIES.md](https://github.com/opensearch-project/.github/blob/main/RESPONSIBILITIES.md#maintainer-responsibilities) that explains what the role of maintainer means, what maintainers do in this and other repos, and how they should be doing it. If you're interested in contributing, and becoming a maintainer, see [CONTRIBUTING](CONTRIBUTING.md).
+
+## Current Maintainers
```
ylwu-amzn, line 5: This file change seems not for this PR. Rebase first?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -30,131 +30,134 @@ public class CommonValue {
 
     public static final String ML_MODEL_INDEX = ".plugins-ml-model";
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
+    public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
```
ylwu-amzn, line 33: Adding new index for a specific tasks is not extensible. Is it possible to reuse current index?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -30,131 +30,134 @@ public class CommonValue {
...
-            + "        \"type\": \"nested\",\n"
-            + "        \"properties\": {\n"
-            + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
-            + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-            + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-            + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
-            + "        }\n"
-            + "      }\n";
+        + CommonValue.USER
+        + "\": {\n"
```
ylwu-amzn, line 40: These changes are just formatting code. Suggest keep the code format stable, otherwise everyone could change the format, that increase code review effort.

ylwu-amzn:
There are many changes not belong to this PR. Please rebase first to make the PR clean. Will be easier to review.

PR #656, opened 2022-12-31
change task worker node to list; add target worker node to cache

ylwu-amzn:

### Description
1. ML task may run on multiple nodes. We save all worker nodes as String in ML task doc which is not easy to use. This PR changed worker nodes in MLTask as List<String>
2. Add target worker node to cache
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLTask.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLTask.java
@@ -20,6 +20,8 @@ import org.opensearch.ml.common.dataset.MLInputDataType;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
 
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.USER;
@@ -54,7 +56,7 @@ public class MLTask implements ToXContentObject, Writeable {
     private Float progress;
     private final String outputIndex;
     @Setter
-    private String workerNode;
+    private List<String> workerNodes;
     private final Instant createTime;
     private Instant lastUpdateTime;
     @Setter
@@ -72,7 +74,7 @@ public class MLTask implements ToXContentObject, Writeable {
         MLInputDataType inputType,
         Float progress,
         String outputIndex,
-        String workerNode,
+        List<String> workerNodes,
         Instant createTime,
         Instant lastUpdateTime,
         String error,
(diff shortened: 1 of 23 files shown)


b4sjoo (approved):

b4sjoo (approved):

b4sjoo (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelCache.java
```diff
@@ -29,16 +30,30 @@ public class MLModelCache {
...
     private final Queue<Double> predictRequestDurationQueue;
 
     public MLModelCache() {
+        targetWorkerNodes = ConcurrentHashMap.newKeySet();
         workerNodes = ConcurrentHashMap.newKeySet();
         modelInferenceDurationQueue = new ConcurrentLinkedQueue<>();
         predictRequestDurationQueue = new ConcurrentLinkedQueue<>();
     }
 
+    public void setTargetWorkerNodes(List<String> targetWorkerNodes) {
```
jngz-es, line 45: Here targetWorkerNodes is a private concurrent set, so I assume MLModelCache needs to be threadsafe. In this case, if setTargetWorkerNodes can be run by multi-threads, we need to protect it by synchronize, otherwise, the targetWorkerNodes could be wrong.

ylwu-amzn: This method will be used by `MLModelCacheHelper.initModelState` method, which already has `synchronize`

common/src/main/java/org/opensearch/ml/common/MLTask.java
```diff
@@ -245,7 +248,15 @@ public static MLTask parse(XContentParser parser) throws IOException {
...
-                    workerNode = parser.text();
+                    if (XContentParser.Token.START_ARRAY == parser.currentToken()) {
+                        workerNodes = new ArrayList<>();
+                        while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
+                            workerNodes.add(parser.text());
+                        }
+                    } else {
+                        String[] nodes = parser.text().split(",");
+                        workerNodes = Arrays.asList(nodes);
+                    }
```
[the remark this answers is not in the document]
ylwu-amzn: For old ML tasks, the worker nodes saved as String, for example "node1,node2,node3". This branch is for BWC. When user get old ML task, will go to this branch.

jngz-es (approved):

PR #662, opened 2023-01-06
add ML task timeout setting and clean up expired tasks from cache

ylwu-amzn:

### Description
For some edge case like worker node crashed, the ML task may stay in the cache forever. This PR fixed that issue by adding timeout setting (default 10 minutes). Sync up job will check ML task in cache expired or not. If expired, will reset task and model status and remove task from cache to avoid memory leak. 

Also fixed one issue for reloading a model which already loaded on some node. For example,  user can load model to node1, then target worker nodes is `[ node1 ]`, after model loaded, user may load model again on node2. Then target worker node becomes `[ node2 ]`. This is not expected. This PR checks if the new target worker nodes include all old loaded nodes (for this example, it's `[ node 1]`). If not, will throw exception. So if user load model again, they can input `[ node1, node2 ]` in load model API, and the target worker nodes will be `[ node1, node2 ]`.  If user input `[ node2 ]` in load model API, will throw exception as `node1` not included.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/syncup/TransportSyncUpOnNodeAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/syncup/TransportSyncUpOnNodeAction.java
@@ -204,6 +204,16 @@ public class TransportSyncUpOnNodeAction extends
                         modelState = MLModelState.PARTIALLY_LOADED;
                     } else {
                         modelState = MLModelState.LOADED;
+                        if (mlTask.getWorkerNodes().size() < workerNodes.length) {
+                            log
+                                .warn(
+                                    "Model loaded on more nodes than target worker nodes. taskId:{}, modelId: {}, workerNodes: {}, targetWorkerNodes: {}",
+                                    taskId,
+                                    modelId,
+                                    Arrays.toString(workerNodes),
+                                    Arrays.toString(mlTask.getWorkerNodes().toArray(new String[0]))
+                                );
+                        }
                     }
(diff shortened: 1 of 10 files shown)


rbhavna (approved):

plugin/src/main/java/org/opensearch/ml/action/syncup/TransportSyncUpOnNodeAction.java
```diff
@@ -148,11 +166,52 @@ private MLSyncUpNodeResponse createSyncUpNodeResponse(MLSyncUpNodesRequest loadM
...
+                    String modelId = mlTask.getModelId();
+                    String[] workerNodes = mlModelManager.getWorkerNodes(modelId);
+                    MLModelState modelState;
+                    if (workerNodes == null || workerNodes.length == 0) {
+                        modelState = MLModelState.LOAD_FAILED;
+                    } else if (mlTask.getWorkerNodes().size() > workerNodes.length) {
+                        modelState = MLModelState.PARTIALLY_LOADED;
+                    } else {
+                        modelState = MLModelState.LOADED;
+                    }
```
jngz-es, line 207: Should we add the branch (mlTask.getWorkerNodes().size() < workerNodes.length)? If entering, log this abnormal status.

rbhavna (approved):

jngz-es (approved):

rbhavna (approved):

jngz-es (approved):

Issue #648, opened 2022-12-28
[FEATURE] Add circuit breaker for native memory usage


ylwu-amzn:
**Is your feature request related to a problem?**
When load model, it will consume native memory. If all native memory used, operating system will kill OpenSearch automatically. 

**What solution would you like?**
We should build circuit breaker for native memory usage just like the JVM heap circuit breaker. If native memory usage exceeds some threshold, we will throw exception

**What alternatives have you considered?**
no

**Do you have any additional context?**
OsService class provides operation system memory usage. We can use that to build a circuit breaker.

b4sjoo:
Feature PR merged, we can close this.

PR #689, opened 2023-01-11
Add native memory circuit breaker.

jngz-es:
Add native memory circuit breaker.
Refactor all breakers from common to plugin.
Add dynamic setting for native memory circuit breaker.

### Description
Add native memory circuit breaker.
Refactor all breakers from common to plugin.
Add dynamic setting for native memory circuit breaker.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/648
 
### Check List
  - [ ] All tests pass


ylwu-amzn:
```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"persistent setting [plugins.ml_commons.native_mem_threshold], not recognized"}],"type":"illegal_argument_exception","reason":"persistent setting [plugins.ml_commons.native_mem_threshold], not recognized"},"status":400}
```
Sicheng added BWC test, which test from 2.4.0,  in 2.4.0, we don't have the native memory threshold setting

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -53,4 +53,7 @@ private MLCommonsSettings() {}
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
+
+    public static final Setting<Integer> ML_COMMONS_NATIVE_MEM_THRESHOLD = Setting
+        .intSetting("plugins.ml_commons.native_mem_threshold", 90, 0, 100, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
ylwu-amzn, line 58: `mem` seems not so clear. how about change to 
```suggestion
        .intSetting("plugins.ml_commons.native_memory_threshold", 90, 0, 100, Setting.Property.NodeScope, Setting.Property.Dynamic);
```

jngz-es: sure, will change it to memory.

plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
```diff
@@ -32,17 +32,17 @@ public DiskCircuitBreaker(long threshold, String diskDir) {
...
-        return  ML_DISK_CB;
+        return ML_DISK_CB;
     }
 
     @Override
     public boolean isOpen() {
         try {
             return AccessController.doPrivileged((PrivilegedExceptionAction<Boolean>) () -> {
-                return (new File(diskDir).getFreeSpace()/1024/1024/1024) < getThreshold();  // in GB
+                return (new File(diskDir).getFreeSpace() / 1024 / 1024 / 1024) < getThreshold();  // in GB
```
ylwu-amzn, line 42: minor: we can create constant for the disk threshold rather than recalculate for every request.
```
public static final long DEFAULT_FREE_DISK_SPACE_THRESHOLD = 5l * 1024 * 1024 * 1024;
```

jngz-es: It's from original code not for this PR. Sure, I can add it.

plugin/src/main/java/org/opensearch/ml/breaker/MLCircuitBreakerService.java
```diff
@@ -65,18 +78,21 @@ public MLCircuitBreakerService init(Path path) {
...
      *
-     * @return the name of any open circuit breaker; otherwise return null
+     * @return any open circuit breaker; otherwise return null
      */
-    public String checkOpenCB() {
+    public ThresholdCircuitBreaker checkOpenCB() {
         for (CircuitBreaker breaker : breakers.values()) {
             if (breaker.isOpen()) {
-                return breaker.getName();
+                return (ThresholdCircuitBreaker) breaker;
```
ylwu-amzn, line 95: `breaker` class is `CircuitBreaker`, this line enforce it to `ThresholdCircuitBreaker`, seems not so extensible. If we have new child class of `CircuitBreaker`, this line will throw error.

Why need to change the return type from `String` to `ThresholdCircuitBreaker`?

jngz-es: Want to get more information like the current threshold when breaker is triggered to help debug.

plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
jngz-es: I prefer to add a contant only for 1024 * 1024 * 1024, as the original logic is using G as default unit.

plugin/src/main/java/org/opensearch/ml/breaker/MLCircuitBreakerService.java
ylwu-amzn: I think we can add a warning log to help

ylwu-amzn (approved):

b4sjoo (approved):

PR #700, opened 2023-01-18
check model state before deleting a model

rbhavna:
Signed-off-by: Bhavana Ramaram <rbhavna@amazon.com>

### Description
This PR checks the condition for model state before deleting a model. If the model is in loaded/loading/uploading/partially loaded state, it throws exception and tells user to unload the model first. Previously, the delete API would delete the model irrespective of any model state which was no desirable.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
@@ -5,8 +5,11 @@
 
 package org.opensearch.ml.action.models;
 
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.common.MLModel.MODEL_ID_FIELD;
+import static org.opensearch.ml.utils.MLNodeUtils.createXContentParserFromRegistry;
+import static org.opensearch.ml.utils.RestActionUtils.getFetchSourceContext;
 
 import lombok.AccessLevel;
 import lombok.experimental.FieldDefaults;
@@ -18,18 +21,26 @@ import org.opensearch.action.ActionListener;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.delete.DeleteRequest;
 import org.opensearch.action.delete.DeleteResponse;
+import org.opensearch.action.get.GetRequest;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.client.Client;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.NamedXContentRegistry;
+import org.opensearch.common.xcontent.XContentParser;
 import org.opensearch.index.query.TermsQueryBuilder;
 import org.opensearch.index.reindex.BulkByScrollResponse;
 import org.opensearch.index.reindex.DeleteByQueryAction;
 import org.opensearch.index.reindex.DeleteByQueryRequest;
+import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.exception.MLResourceNotFoundException;
+import org.opensearch.ml.common.model.MLModelState;
 import org.opensearch.ml.common.transport.model.MLModelDeleteAction;
 import org.opensearch.ml.common.transport.model.MLModelDeleteRequest;
+import org.opensearch.ml.common.transport.model.MLModelGetRequest;
 import org.opensearch.rest.RestStatus;
+import org.opensearch.search.fetch.subphase.FetchSourceContext;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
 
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,70 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+        this.xContentRegistry = xContentRegistry;
     }
 
     @Override
     protected void doExecute(Task task, ActionRequest request, ActionListener<DeleteResponse> actionListener) {
         MLModelDeleteRequest mlModelDeleteRequest = MLModelDeleteRequest.fromActionRequest(request);
         String modelId = mlModelDeleteRequest.getModelId();
-
-        DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_INDEX, modelId);
+        MLModelGetRequest mlModelGetRequest = new MLModelGetRequest(modelId, true);
```
ylwu-amzn, line 76: Do we need to return model content here?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,70 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+        GetRequest getRequest = new GetRequest(ML_MODEL_INDEX).id(modelId).fetchSourceContext(fetchSourceContext);
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-            client.delete(deleteRequest, new ActionListener<DeleteResponse>() {
-                @Override
-                public void onResponse(DeleteResponse deleteResponse) {
-                    deleteModelChunks(modelId, deleteResponse, actionListener);
-                }
+            client.get(getRequest, ActionListener.runBefore(ActionListener.wrap(r -> {
+                log.debug("Completed Get Model Request, id:{}", modelId);
```
ylwu-amzn, line 82: Seems no need to add log here.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,70 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+
+                if (r != null && r.isExists()) {
+                    try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
+                        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                        MLModel mlModel = MLModel.parse(parser);
+                        MLModelState mlModelState = mlModel.getModelState();
+                        System.out.println(mlModelState);
```
ylwu-amzn, line 89: Don't use system print. Remove this line?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,70 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+                        if (mlModelState.equals(MLModelState.LOADED)
+                            || mlModelState.equals(MLModelState.LOADING)
+                            || mlModelState.equals(MLModelState.UPLOADING)
```
ylwu-amzn, line 92: Remove this line? One use case: user upload a model from local file, it split into 10 chunks, 4 chunks uploaded, then user change mind to delete the model. The model is on UPLOADING state, but we should allow user delete the model.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,70 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+                            });
+                        }
+                    } catch (Exception e) {
+                        log.error("Failed to parse ml model" + r.getId(), e);
+                        actionListener.onFailure(e);
                     }
-                    actionListener.onFailure(e);
                 }
-            });
+            }, e -> { actionListener.onFailure(new MLResourceNotFoundException("Fail to find model")); }), () -> context.restore()));
```
ylwu-amzn, line 121: It's possible that user deleted model meta doc first, then failed to delete model chunk when run line 103. That's why we add line 110 which will re-delete model chunks even model meta doc not found. I think we should follow the same logic here.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
rbhavna: Yeah we can skip it.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
rbhavna: This is for the outer wrap when the get model request does not find a model. Line 110 corresponds to delete request. In this case, the delete request is not called yet. So, skipped deleting anything and just threw an exception.

rbhavna:
Closing this PR and creating a new one with comments addressed and new changes: https://github.com/opensearch-project/ml-commons/pull/725

Issue #703, opened 2023-01-23
Refresh load model state


ylwu-amzn:
**Is your feature request related to a problem?**
When load model, there are some edge cases the model state is not accurate:
1. Node crashed when loading model, then the model may stay on LOADING state forever.  Should change it to LOAD_FAILED.
2. After model loaded, some ML node crashed, then the model will still stay on LOADED state. Should change it to PARTIALLY_LOADED state
3. After model loaded, all ML nodes crashed or cluster restart, then the model will still stay on LOADED state. Should change it to LOAD_FAILED state.

**What solution would you like?**
Refresh model state in sync up job.

**What alternatives have you considered?**
No

**Do you have any additional context?**
No

PR #704, opened 2023-01-23
refresh model load state in sync up cron job

ylwu-amzn:

### Description
When load model, there are some edge cases the model state is not accurate:

1. Node crashed when loading model, then the model may stay on LOADING state forever. Should change it to LOAD_FAILED.
2. After model loaded, some ML node crashed, then the model will still stay on LOADED state. Should change it to PARTIALLY_LOADED state
3. After model loaded, all ML nodes crashed or cluster restart, then the model will still stay on LOADED state. Should change it to LOAD_FAILED state.

This PR also adds two new fields in model index: 
1. `planning_worker_node_count`: plan to deploy model to how many nodes 
2. `current_worker_node_count`: model is running on how many nodes

### Issues Resolved
close https://github.com/opensearch-project/ml-commons/issues/703
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -30,7 +30,7 @@ public class CommonValue {
 
     public static final String ML_MODEL_INDEX = ".plugins-ml-model";
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
-    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 2;
+    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 3;
     public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 1;
     public static final String USER_FIELD_MAPPING = "      \""
             + CommonValue.USER
@@ -85,6 +85,12 @@ public class CommonValue {
             + MLModel.MODEL_CONTENT_SIZE_IN_BYTES_FIELD
             + "\" : {\"type\": \"long\"},\n"
             + "      \""
+            + MLModel.PLANNING_WORKER_NODE_COUNT_FIELD
+            + "\" : {\"type\": \"integer\"},\n"
+            + "      \""
+            + MLModel.CURRENT_WORKER_NODE_COUNT_FIELD
+            + "\" : {\"type\": \"integer\"},\n"
+            + "      \""
             + MLModel.MODEL_CONFIG_FIELD
             + "\" : {\"properties\":{\""
             + MODEL_TYPE_FIELD + "\":{\"type\":\"keyword\"},\""
@@ -99,6 +105,9 @@ public class CommonValue {
             + "\" : {\"type\": \"keyword\"},\n"
             + "      \""
             + MLModel.CREATED_TIME_FIELD
+            + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
+            + "      \""
+            + MLModel.LAST_UPDATE_TIME_FIELD
             + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
             + "      \""
             + MLModel.LAST_UPLOADED_TIME_FIELD
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -107,7 +107,7 @@ public class CommonValue {
             + MLModel.CREATED_TIME_FIELD
             + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
             + "      \""
-            + MLModel.LAST_UPDATE_TIME_FIELD
+            + MLModel.LAST_UPDATED_TIME_FIELD
             + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
             + "      \""
             + MLModel.LAST_UPLOADED_TIME_FIELD
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -44,6 +44,7 @@ public class MLModel implements ToXContentObject {
     public static final String MODEL_CONTENT_HASH_VALUE_FIELD = "model_content_hash_value";
     public static final String MODEL_CONFIG_FIELD = "model_config";
     public static final String CREATED_TIME_FIELD = "created_time";
+    public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
     public static final String LAST_UPLOADED_TIME_FIELD = "last_uploaded_time";
     public static final String LAST_LOADED_TIME_FIELD = "last_loaded_time";
     public static final String LAST_UNLOADED_TIME_FIELD = "last_unloaded_time";
@@ -51,6 +52,8 @@ public class MLModel implements ToXContentObject {
     public static final String MODEL_ID_FIELD = "model_id";
     public static final String CHUNK_NUMBER_FIELD = "chunk_number";
     public static final String TOTAL_CHUNKS_FIELD = "total_chunks";
+    public static final String PLANNING_WORKER_NODE_COUNT_FIELD = "planning_worker_node_count";
+    public static final String CURRENT_WORKER_NODE_COUNT_FIELD = "current_worker_node_count";
 
     private String name;
     private FunctionName algorithm;
@@ -66,6 +69,7 @@ public class MLModel implements ToXContentObject {
     private String modelContentHash;
     private MLModelConfig modelConfig;
     private Instant createdTime;
+    private Instant lastUpdateTime;
     private Instant lastUploadedTime;
     private Instant lastLoadedTime;
     private Instant lastUnloadedTime;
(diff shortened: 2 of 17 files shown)


plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -91,6 +129,139 @@ public void run() {
...
+                                res -> { refreshModelState(modelWorkerNodes, loadingModels); },
+                                e -> { log.error("Failed to init model index", e); }
+                            )
+                    );
+            }
         }, e -> { log.error("Failed to sync model routing", e); }));
     }
+
+    @VisibleForTesting
+    void refreshModelState(Map<String, Set<String>> modelWorkerNodes, Map<String, Set<String>> loadingModels) {
```
Zhangxunmt, line 148: This function is a bit too long. Why not break it into smaller ones?

rbhavna (dismissed):

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -101,6 +107,9 @@ public class CommonValue {
             + MLModel.CREATED_TIME_FIELD
             + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
             + "      \""
+            + MLModel.LAST_UPDATE_TIME_FIELD
```
dhrubo-os, line 110: with keeping the similar convention with `CREATED_TIME_FIELD` and `LAST_UPLOADED_TIME_FIELD` may be we can rename to : `LAST_UPDATED_TIME_FIELD`

common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -44,13 +44,16 @@ public class MLModel implements ToXContentObject {
     public static final String MODEL_CONTENT_HASH_VALUE_FIELD = "model_content_hash_value";
     public static final String MODEL_CONFIG_FIELD = "model_config";
     public static final String CREATED_TIME_FIELD = "created_time";
+    public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
```
dhrubo-os, line 47: same here

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -27,7 +27,7 @@ private MLCommonsSettings() {}
     public static final Setting<Integer> ML_COMMONS_SYNC_UP_JOB_INTERVAL_IN_SECONDS = Setting
         .intSetting(
             "plugins.ml_commons.sync_up_job_interval_in_seconds",
-            10,
+            3,
```
Zhangxunmt, line 30: Reducing the sync-up interval to 3 seconds to catch up load failures sooner?

common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -44,13 +44,16 @@ public class MLModel implements ToXContentObject {
...
     public static final String LAST_UPLOADED_TIME_FIELD = "last_uploaded_time";
     public static final String LAST_LOADED_TIME_FIELD = "last_loaded_time";
     public static final String LAST_UNLOADED_TIME_FIELD = "last_unloaded_time";
 
     public static final String MODEL_ID_FIELD = "model_id";
     public static final String CHUNK_NUMBER_FIELD = "chunk_number";
     public static final String TOTAL_CHUNKS_FIELD = "total_chunks";
+    public static final String PLANNING_WORKER_NODE_COUNT_FIELD = "planning_worker_node_count";
```
dhrubo-os, line 55: May be we can add some comment to explain what did we mean by `Planning worker`?

common/src/main/java/org/opensearch/ml/common/transport/sync/MLSyncUpNodeResponse.java
```diff
@@ -13,26 +13,31 @@
...
 import java.io.IOException;
+import java.util.Map;
 
 @Log4j2
 @Getter
 public class MLSyncUpNodeResponse extends BaseNodeResponse  {
 
     private String modelStatus;
     private String[] loadedModelIds;
+    private String[] runningLoadModelIds;
```
dhrubo-os, line 24: Can we add any comment for this field?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -5,31 +5,58 @@
...
+    @VisibleForTesting
+    Semaphore updateModelStateSemaphore;
 
-    public MLSyncUpCron(Client client, DiscoveryNodeHelper nodeHelper) {
+    public MLSyncUpCron(Client client, ClusterService clusterService, DiscoveryNodeHelper nodeHelper, MLIndicesHandler mlIndicesHandler) {
         this.client = client;
+        this.clusterService = clusterService;
         this.nodeHelper = nodeHelper;
+        this.mlIndicesHandler = mlIndicesHandler;
+        this.updateModelStateSemaphore = new Semaphore(1);
```
dhrubo-os, line 59: should we add any fairness setting here?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
dhrubo-os: +1. There are lot of `if-else` clauses which are hard to follow. May be add few comments also?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
ylwu-amzn: This is to keep consistent with MLTask.LAST_UPDATE_TIME_FIELD. I think it's reasonable to keep consistent with other fields of ML model. Will change this part.

common/src/main/java/org/opensearch/ml/common/MLModel.java
ylwu-amzn: Same comment

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: We did some testing reduce to 1 second won't impact cluster performance too much. Frontend needs shorter interval to get latest model state.

common/src/main/java/org/opensearch/ml/common/MLModel.java
ylwu-amzn: Sure, will add some comments for this and also `currentWorkerNodeCount`

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
ylwu-amzn: This is to control only one sync up job can update model state to avoid race condition.

common/src/main/java/org/opensearch/ml/common/transport/sync/MLSyncUpNodeResponse.java
ylwu-amzn: sure, will add

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
ylwu-amzn: Sure, will break to some functions

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
dhrubo-os: I saw Semaphore has another parameter `fair` :

    fair - true if this semaphore will guarantee first-in first-out granting of permits under contention, else false
    
  Let's say there's 3 sync up jobs (job1, job2, job3) and job1 is running. Don't we need to make sure that job2 should run after job1 is finished not job3?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -91,6 +129,171 @@ public void run() {
...
+
+        String modelId,
+        MLModelState state,
+        Long lastUpdateTime,
+        int planningWorkerNodeCount,
+        int currentWorkerNodeCountInIndex
+    ) {
+        Set<String> loadTaskNodes = loadingModels.get(modelId);
+        if (loadTaskNodes != null && loadTaskNodes.size() > 0 && state != MLModelState.LOADING) {
+            // no
```
dhrubo-os, line 233: can we make a clear comment here?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -91,6 +129,171 @@ public void run() {
...
+                // If model deployed to some node/nodes, but not deployed to all nodes planned by user,
+                // then set model state as PARTIALLY_LOADED.
+                return MLModelState.PARTIALLY_LOADED;
+            } else if (planningWorkerNodeCount > 0 && currentWorkerNodeCount >= planningWorkerNodeCount && state != MLModelState.LOADED) {
+                if (currentWorkerNodeCount > planningWorkerNodeCount) {
+                    // This case should not happen that model loaded to more nodes than planned. So log this as warning if
+                    // it happens.
+                    log
+                        .warn(
+                            "Model {} loaded on more nodes [{}] than planing worker node[{}]",
```
dhrubo-os, line 257: [nit] than planing worker -> than planning worker

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
ylwu-amzn: That's not necessary. We just need to make sure only 1 job updating the model state is fine. Job interval is relatively big enough (3s) that race condition is a rare case. If race condition happens, next job run will refresh model state again, no need to guarantee the first-in first-out.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
dhrubo-os: Just curious to know how this warning is useful to customers. What do we expect user to do if they see this warning?

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
ylwu-amzn: That case is not expected to happen. But if user found this edge case happens, this error can give some clue like when this error happened. Then we can analyze other log like when user started loading model, if user re-loaded model etc to find out the root cause.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
ylwu-amzn: Seems forgot to include this change when commit. Will fix this part in next PR.

PR #711, opened 2023-01-25
the dev of [FEATURE]Auto reload model when cluster rebooted/node rejoin

wujunshen:

### Description
the new feature: 
 
### Issues Resolved
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -36,7 +36,6 @@ public class CommonValue {
     public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
     public static final String NODE_ID_FIELD = "node_id";
     public static final String MODEL_LOAD_RETRY_TIMES_FIELD = "retry_times";
-    public static final Integer ML_MODEL_RELOAD_MAX_RETRY_TIMES = 2;
     public static final String USER_FIELD_MAPPING = "      \""
             + CommonValue.USER
             + "\": {\n"
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -96,6 +96,9 @@ public class CommonValue {
             + MLModel.CURRENT_WORKER_NODE_COUNT_FIELD
             + "\" : {\"type\": \"integer\"},\n"
             + "      \""
+            + MLModel.PLANNING_WORKER_NODES_FIELD
+            + "\": {\"type\": \"keyword\"},\n"
+            + "      \""
             + MLModel.MODEL_CONFIG_FIELD
             + "\" : {\"properties\":{\""
             + MODEL_TYPE_FIELD + "\":{\"type\":\"keyword\"},\""
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -32,6 +32,11 @@ public class CommonValue {
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
     public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 3;
     public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 1;
+
+    public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
+    public static final String NODE_ID_FIELD = "node_id";
+    public static final String MODEL_LOAD_RETRY_TIMES_FIELD = "retry_times";
+    public static final Integer ML_MODEL_RELOAD_MAX_RETRY_TIMES = 2;
     public static final String USER_FIELD_MAPPING = "      \""
             + CommonValue.USER
             + "\": {\n"
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
@@ -19,7 +19,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 
-import lombok.Data;
 import lombok.extern.log4j.Log4j2;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
@@ -143,7 +142,6 @@ public class MLModelAutoReLoader {
      */
     @VisibleForTesting
     void autoReLoadModelByNodeId(String localNodeId) throws ExecutionException, InterruptedException {
-        // StepListener<IndicesExistsResponse> indicesExistsResponseStepListener = new StepListener<>();
         StepListener<SearchResponse> queryTaskStep = new StepListener<>();
         StepListener<SearchResponse> getReTryTimesStep = new StepListener<>();
         StepListener<IndexResponse> saveLatestReTryTimesStep = new StepListener<>();
@@ -157,15 +155,14 @@ public class MLModelAutoReLoader {
 
         getReTryTimes(localNodeId, ActionListener.wrap(getReTryTimesStep::onResponse, getReTryTimesStep::onFailure));
 
-        Result result = new Result();
         queryTaskStep.whenComplete(searchResponse -> {
             SearchHit[] hits = searchResponse.getHits().getHits();
             if (CollectionUtils.isEmpty(hits)) {
                 return;
             }
-            result.setHits(hits);
 
             getReTryTimesStep.whenComplete(getReTryTimesResponse -> {
+                int reTryTimes = 0;
                 // if getReTryTimesResponse is null,it means we get reTryTimes at the first time,and the index
                 // .plugins-ml-model-reload doesn't exist,so we should let reTryTimes be zero(init value)
                 // we don't do anything
@@ -173,29 +170,26 @@ public class MLModelAutoReLoader {
                 // .plugins-ml-model-reload,so we get the value of the field MODEL_LOAD_RETRY_TIMES_FIELD
                 if (getReTryTimesResponse != null) {
                     Map<String, Object> sourceAsMap = getReTryTimesResponse.getHits().getHits()[0].getSourceAsMap();
-                    result.setReTryTimes((Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD));
+                    reTryTimes = (Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD);
                 }
 
                 // According to the node id to get retry times, if more than the max retry times, don't need to retry
                 // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
-                if (result.getReTryTimes() > autoReLoadMaxReTryTimes) {
+                if (reTryTimes > autoReLoadMaxReTryTimes) {
                     log.info("have exceeded max retry times, always failure");
                     return;
                 }
 
-                int reTryTimes = 0;
-                try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, result.getHits()[0].getSourceRef())) {
+                try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, hits[0].getSourceRef())) {
                     ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                     MLTask mlTask = MLTask.parse(parser);
 
                     autoReLoadModelByNodeAndModelId(localNodeId, mlTask.getModelId());
 
                     // if reload the model successfully,the number of unsuccessful reload should be reset to zero.
-                    result.setReTryTimes(reTryTimes);
+                    reTryTimes = 0;
                 } catch (MLException e) {
-                    reTryTimes = result.getReTryTimes();
                     reTryTimes++;
-                    result.setReTryTimes(reTryTimes);
                     log.error("Can't auto reload model in node id {} ,has tried {} times\nThe reason is:{}", localNodeId, reTryTimes, e);
                 }
 
@@ -331,11 +325,5 @@ public class MLModelAutoReLoader {
             }
             indexResponseActionListener.onFailure(new MLException("node id:" + localNodeId + " insert retry times unsuccessfully"));
         }, indexResponseActionListener::onFailure));
-    }
-
-    @Data
-    static class Result {
-        private SearchHit[] hits;
-        private int reTryTimes = 0;
     }
 }
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
@@ -253,8 +253,13 @@ public class MLModelAutoReLoader {
         QueryBuilder queryBuilder = QueryBuilders
             .boolQuery()
             .must(QueryBuilders.matchPhraseQuery("task_type", "LOAD_MODEL"))
-            .must(QueryBuilders.matchPhraseQuery("state", "COMPLETED"))
-            .must(QueryBuilders.matchPhraseQuery("worker_node", localNodeId));
+            .must(QueryBuilders.matchPhraseQuery("worker_node", localNodeId))
+            .must(
+                QueryBuilders
+                    .boolQuery()
+                    .should(QueryBuilders.matchPhraseQuery("state", "COMPLETED"))
+                    .should(QueryBuilders.matchPhraseQuery("state", "COMPLETED_WITH_ERROR"))
+            );
         searchSourceBuilder.query(queryBuilder);
 
         SortBuilder<FieldSortBuilder> sortBuilderOrder = new FieldSortBuilder("create_time").order(SortOrder.DESC);
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
@@ -1,0 +1,352 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.ml.model;
+
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.ML_MODEL_RELOAD_INDEX;
+import static org.opensearch.ml.common.CommonValue.ML_TASK_INDEX;
+import static org.opensearch.ml.common.CommonValue.MODEL_LOAD_RETRY_TIMES_FIELD;
+import static org.opensearch.ml.common.CommonValue.NODE_ID_FIELD;
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE;
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_MODEL_RELOAD_MAX_RETRY_TIMES;
+import static org.opensearch.ml.utils.MLNodeUtils.createXContentParserFromRegistry;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import lombok.Data;
+import lombok.extern.log4j.Log4j2;
+
+import org.opensearch.action.ActionListener;
+import org.opensearch.action.StepListener;
+import org.opensearch.action.admin.indices.exists.indices.IndicesExistsAction;
+import org.opensearch.action.admin.indices.exists.indices.IndicesExistsRequestBuilder;
+import org.opensearch.action.admin.indices.exists.indices.IndicesExistsResponse;
+import org.opensearch.action.index.IndexAction;
+import org.opensearch.action.index.IndexRequestBuilder;
+import org.opensearch.action.index.IndexResponse;
+import org.opensearch.action.search.SearchAction;
+import org.opensearch.action.search.SearchRequestBuilder;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.support.WriteRequest;
+import org.opensearch.client.Client;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.CollectionUtils;
+import org.opensearch.common.xcontent.NamedXContentRegistry;
+import org.opensearch.common.xcontent.XContentParser;
+import org.opensearch.index.IndexNotFoundException;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.cluster.DiscoveryNodeHelper;
+import org.opensearch.ml.common.MLTask;
+import org.opensearch.ml.common.transport.load.MLLoadModelAction;
+import org.opensearch.ml.common.transport.load.MLLoadModelRequest;
+import org.opensearch.ml.utils.MLNodeUtils;
+import org.opensearch.rest.RestStatus;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
+import org.opensearch.search.sort.SortBuilder;
+import org.opensearch.search.sort.SortOrder;
+import org.opensearch.threadpool.ThreadPool;
+
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * Manager class for ML models and nodes. It contains ML model auto reload operations etc.
+ */
+@Log4j2
+public class MLModelAutoReLoader {
+
+    private final Client client;
+    private final ClusterService clusterService;
+    private final NamedXContentRegistry xContentRegistry;
+    private final DiscoveryNodeHelper nodeHelper;
+    private final ThreadPool threadPool;
+    private volatile Boolean enableAutoReLoadModel;
+    private volatile Integer autoReLoadMaxReTryTimes;
+
+    /**
+     * constructor method， init all the params necessary for model auto reloading
+     *
+     * @param clusterService   clusterService
+     * @param threadPool       threadPool
+     * @param client           client
+     * @param xContentRegistry xContentRegistry
+     * @param nodeHelper       nodeHelper
+     * @param settings         settings
+     */
+    public MLModelAutoReLoader(
+        ClusterService clusterService,
+        ThreadPool threadPool,
+        Client client,
+        NamedXContentRegistry xContentRegistry,
+        DiscoveryNodeHelper nodeHelper,
+        Settings settings
+    ) {
+        this.clusterService = clusterService;
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+        this.nodeHelper = nodeHelper;
+        this.threadPool = threadPool;
+
+        enableAutoReLoadModel = ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE.get(settings);
+        autoReLoadMaxReTryTimes = ML_MODEL_RELOAD_MAX_RETRY_TIMES.get(settings);
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE, it -> enableAutoReLoadModel = it);
+
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(ML_MODEL_RELOAD_MAX_RETRY_TIMES, it -> autoReLoadMaxReTryTimes = it);
+    }
+
+    /**
+     * the main method: model auto reloading
+     */
+    public void autoReLoadModel() {
+        log.info("enableAutoReLoadModel: {} ", enableAutoReLoadModel);
+
+        // if we don't need to reload automatically, just return without doing anything
+        if (!enableAutoReLoadModel) {
+            return;
+        }
+
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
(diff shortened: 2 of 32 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+            .addSettingsUpdateConsumer(ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE, it -> enableAutoReLoadModel = it);
+
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(ML_MODEL_RELOAD_MAX_RETRY_TIMES, it -> autoReLoadMaxReTryTimes = it);
+    }
+
+    /**
+     * the main method: model auto reloading
+     */
+    public void autoReLoadModel() {
+        log.info("enableAutoReLoadModel: {} ", enableAutoReLoadModel);
```
ylwu-amzn, line 112: Log message is not so readable. How about changing to "Auto reload model enabled: {}"

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
+        if (!MLNodeUtils.isMLNode(clusterService.localNode())) {
+            return;
+        }
+
+        String localNodeId = clusterService.localNode().getId();
+        // auto reload all models of this local ml node
+        threadPool.generic().submit(() -> {
```
ylwu-amzn, line 126: Why not use load model thread pool https://github.com/opensearch-project/ml-commons/blob/2.5/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L448

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+            try {
+                autoReLoadModelByNodeId(localNodeId);
+            } catch (ExecutionException | InterruptedException e) {
+                throw new RuntimeException(e);
```
ylwu-amzn, line 130: Add error log here?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+     * @param indexName index name
+     */
+    void isExistedIndex(String indexName, ActionListener<IndicesExistsResponse> actionListener) {
+        IndicesExistsRequestBuilder indicesExistsRequestBuilder = new IndicesExistsRequestBuilder(
+            client,
+            IndicesExistsAction.INSTANCE,
+            indexName
+        );
+
+        indicesExistsRequestBuilder.execute(ActionListener.wrap(actionListener::onResponse, actionListener::onFailure));
```
ylwu-amzn, line 318: How about just checking if index exists in cluster metadata ? Refer to PR #717

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+            .setSource(content)
+            .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+        indexRequestBuilder.execute(ActionListener.wrap(indexResponse -> {
+            if (indexResponse.status() == RestStatus.CREATED || indexResponse.status() == RestStatus.OK) {
+                log.info("node id:{} insert retry times successfully", localNodeId);
+                indexResponseActionListener.onResponse(indexResponse);
+                return;
+            }
+            indexResponseActionListener.onFailure(new RuntimeException("node id:" + localNodeId + " insert retry times unsuccessfully"));
```
ylwu-amzn, line 343: Change to  `MLException` ?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_MODEL_AUTO_RELOAD_ENABLE, it -> enableAutoReLoadModel = it);
+
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(ML_MODEL_RELOAD_MAX_RETRY_TIMES, it -> autoReLoadMaxReTryTimes = it);
+    }
+
+    /**
+     * the main method: model auto reloading
+     */
+    public void autoReLoadModel() {
```
ylwu-amzn, line 111:
```suggestion
    public void reLoadModel() {
```

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+        log.info("enableAutoReLoadModel: {} ", enableAutoReLoadModel);
+
+        // if we don't need to reload automatically, just return without doing anything
+        if (!enableAutoReLoadModel) {
+            return;
+        }
+
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
+        if (!MLNodeUtils.isMLNode(clusterService.localNode())) {
```
ylwu-amzn, line 120: We can support running model on data node (check this [doc](https://opensearch.org/docs/latest/ml-commons-plugin/cluster-settings/#run-tasks-and-models-on-ml-nodes-only) ). This line means we won't reload model if user running model on data node. Can you confirm with PM if this is ok?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+                    // if reload the model successfully,the number of unsuccessful reload should be reset to zero.
+                    result.setReTryTimes(reTryTimes);
+                } catch (RuntimeException e) {
+                    reTryTimes = result.getReTryTimes();
+                    reTryTimes++;
+                    result.setReTryTimes(reTryTimes);
+                    log.error("Can't auto reload model in node id {} ,has tried {} times\nThe reason is:{}", localNodeId, reTryTimes, e);
+                }
+
+                // Store the latest value of the reTryTimes and node id under the index ".plugins-ml-model-reload"
```
ylwu-amzn, line 199: Any reason we have to persist retryTimes to index ? Is it ok to just cache retry times in memory?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+
+                int reTryTimes = 0;
+                try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, result.getHits()[0].getSourceRef())) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                    MLTask mlTask = MLTask.parse(parser);
+
+                    autoReLoadModelByNodeAndModelId(localNodeId, mlTask.getModelId());
+
+                    // if reload the model successfully,the number of unsuccessful reload should be reset to zero.
+                    result.setReTryTimes(reTryTimes);
```
ylwu-amzn, line 191: Load model is async, I don't think we are sure model loaded successfully even line 188 `autoReLoadModelByNodeAndModelId(localNodeId, mlTask.getModelId());` doesn't throw exception.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+        }
+        MLLoadModelRequest mlLoadModelRequest = new MLLoadModelRequest(modelId, allNodeIdList.toArray(new String[] {}), false, false);
+
+        client
+            .execute(
+                MLLoadModelAction.INSTANCE,
+                mlLoadModelRequest,
+                ActionListener
+                    .wrap(response -> log.info("the model {} is auto reloading under the node {} ", modelId, localNodeId), exception -> {
+                        log.error("fail to reload model " + modelId + " under the node " + localNodeId + "\nthe reason is: " + exception);
```
ylwu-amzn, line 232: I don't see `retryTimes + 1` here, is that correct? I think we should count this failure in retry times

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+
+                // According to the node id to get retry times, if more than the max retry times, don't need to retry
+                // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+                if (result.getReTryTimes() > autoReLoadMaxReTryTimes) {
+                    log.info("have exceeded max retry times, always failure");
+                    return;
+                }
+
+                int reTryTimes = 0;
+                try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, result.getHits()[0].getSourceRef())) {
```
ylwu-amzn, line 184: From `result.getHits()[0]`, seems it will only reload the first model?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+     * @param localNodeId one of query condition
+     */
+    @VisibleForTesting
+    void queryTask(String localNodeId, ActionListener<SearchResponse> searchResponseActionListener) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().from(0).size(1);
+
+        QueryBuilder queryBuilder = QueryBuilders
+            .boolQuery()
+            .must(QueryBuilders.matchPhraseQuery("task_type", "LOAD_MODEL"))
+            .must(QueryBuilders.matchPhraseQuery("state", "COMPLETED"))
```
ylwu-amzn, line 253: I think we should also consider `COMPLETED_WITH_ERROR` which means the model isn't loaded to all workers nodes, but loaded on some worker nodes successfully.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -32,6 +32,11 @@ public class CommonValue {
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
     public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 3;
     public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 1;
+
+    public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
+    public static final String NODE_ID_FIELD = "node_id";
+    public static final String MODEL_LOAD_RETRY_TIMES_FIELD = "retry_times";
+    public static final Integer ML_MODEL_RELOAD_MAX_RETRY_TIMES = 2;
```
ylwu-amzn, line 39: Do we need this constant as we have setting `ML_MODEL_RELOAD_MAX_RETRY_TIMES`?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: I think it is another function,if I use load model thread poo,it maybe will effect the original process of  model load

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: ok, added it~

Zhangxunmt (commented):
Is it possible to test this auto reload in a local cluster? like creating a 2 nodes cluster, and kill one of the OS process in 1 node, etc?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+ */
+@Log4j2
+public class MLModelAutoReLoader {
+
+    private final Client client;
+    private final ClusterService clusterService;
+    private final NamedXContentRegistry xContentRegistry;
+    private final DiscoveryNodeHelper nodeHelper;
+    private final ThreadPool threadPool;
+    private volatile Boolean enableAutoReLoadModel;
```
dhrubo-os, line 72: `reload` is a full word. Maybe we don't need to do camel casing for reload. `enableAutoReloadModel`?

Same comment for `retry`

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+                // .plugins-ml-model-reload,so we get the value of the field MODEL_LOAD_RETRY_TIMES_FIELD
+                if (getReTryTimesResponse != null) {
+                    Map<String, Object> sourceAsMap = getReTryTimesResponse.getHits().getHits()[0].getSourceAsMap();
+                    result.setReTryTimes((Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD));
+                }
+
+                // According to the node id to get retry times, if more than the max retry times, don't need to retry
+                // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+                if (result.getReTryTimes() > autoReLoadMaxReTryTimes) {
+                    log.info("have exceeded max retry times, always failure");
```
dhrubo-os, line 179: How about: log.info("Node: {} has reached to the max retry limit, failed to load models", localNodeId)

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
Zhangxunmt: Do you need to update the retry times in the ML_MODEL_RELOAD_INDEX to 0 after loading successfully? For instance, the reload succeeded at the 3rd time so you need to reset the retry value in the index from 2 to 0?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: ok, it is a very useful suggestion, I can reduce the complexity of coding implement, thanks.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: because when the ml-node happened reboot,the info of retrying time will be lost in cache if we put them in cache

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: because the method `autoReLoadModelByNodeAndModelId` will throw exception to upper method,and the line188 mentioned in the previous comment will be caught, so `retyrTimes+1` is written in the catch statement(the line 195)

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: This `retryTimes` is defined based on the dimension of `node`, not `node+model`.
In other words, there is only one value of the `retryTimes` for each ml node

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: Okay, I found that there may be a time difference, when I wrote the code，the  enumeration class `MLTaskState` does not exist `COMPLETED_WITH_ERROR`, I add this to my code

common/src/main/java/org/opensearch/ml/common/CommonValue.java
wujunshen: yes, u r right, We don't need it in class `CommonValue`. But we need it in `opensearch.yml`,so we can let user define the value by himself.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
```diff
@@ -0,0 +1,352 @@
...
+                log.info("node id:{} insert retry times successfully", localNodeId);
+                indexResponseActionListener.onResponse(indexResponse);
+                return;
+            }
+            indexResponseActionListener.onFailure(new RuntimeException("node id:" + localNodeId + " insert retry times unsuccessfully"));
+        }, indexResponseActionListener::onFailure));
+    }
+
+    @Data
+    static class Result {
```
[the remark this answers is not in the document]
wujunshen: I found after referring to PR https://github.com/opensearch-project/ml-commons/pull/717,  We will not need this `Result` any more.so  I refactored it~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: that's cool

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: yes，the code have had “if success，reset to zero” function

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
ylwu-amzn: `threadPool.generic()` is not dedicated for ML. Using this thread pool may impact other OpenSearch tasks. Suggest change to ML dedicated load model thread pool.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
ylwu-amzn: @wujunshen Is it ok to not support model auto reload on data node?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,329 @@
...
+        String localNodeId = clusterService.localNode().getId();
+        // auto reload all models of this local ml node
+        threadPool.generic().submit(() -> {
+            try {
+                autoReloadModelByNodeId(localNodeId);
+            } catch (ExecutionException | InterruptedException e) {
+                log
+                    .error(
+                        "the model auto-reloading has exception,and the root cause message is: {}",
+                        ExceptionUtils.getRootCauseMessage(e)
```
ylwu-amzn, line 131: How about print the full exception stack trace here? Just print out the root cause seems not easy to debug.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,329 @@
...
+
+    /**
+     * auto reload 1 model under the node id
+     *
+     * @param localNodeId node id
+     * @param modelId     model id
+     */
+    @VisibleForTesting
+    void autoReloadModelByNodeAndModelId(String localNodeId, String modelId) throws MLException {
+        String[] allNodeIds = nodeHelper.getAllNodeIds();
```
ylwu-amzn, line 216: `nodeHelper.getAllNodeIds()` will return all nodes , not just ML nodes. Should we reload model on all nodes?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: can I use `ExceptionUtils.getMessage(e)`?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: ok, I will modify it. Let the collection just have all ids of ml node.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: I think for a while,and changed it to `ExceptionUtils.getStackTrace(e)` at last.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,331 @@
...
+                if (getReTryTimesResponse != null) {
+                    Map<String, Object> sourceAsMap = getReTryTimesResponse.getHits().getHits()[0].getSourceAsMap();
+                    retryTimes = (Integer) sourceAsMap.get(MODEL_LOAD_RETRY_TIMES_FIELD);
+                }
+
+                // According to the node id to get retry times, if more than the max retry times, don't need to retry
+                // that the number of unsuccessful reload has reached the maximum number of times, do not need to reload
+                if (retryTimes > autoReloadMaxRetryTimes) {
+                    log.info("Node: {} has reached to the max retry limit, failed to load models", localNodeId);
+                    return;
```
Zhangxunmt, line 178: Before return, should we check how long the node has been in the max retry status and reset to 0 after a substantial time? It looks to me the node will never reload forever once reached maximum retry times.

wujunshen: in the first comment of yours, I found ml_task index and ml_model index are both definition in [ml_task index](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/indices/MLIndex.java), so if I add 2 new fields in ml_task,the ml_model index will have these 2 fields,too. It sounds that give ml_model index redundant attributes.
in the second comment of yours, when we discussed the design earlier, if the maximum retry times is reached, instead of automatically reloading, the model need to be loaded manually.

Zhangxunmt: Cool. Let's keep this logic then. But we should try to define a new type of ML Task for auto reload, and reuse MLTask  to store the max_retry field, etc.

Zhangxunmt (dismissed):

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
zane-neo: Print the entire exception stack is useful and convenient to locate and debug issues, you can change to `log
.error("the model auto-reloading has exception,and the root cause message is: {}", e`

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,329 @@
...
+        SortBuilder<FieldSortBuilder> sortBuilderOrder = new FieldSortBuilder("create_time").order(SortOrder.DESC);
+        searchSourceBuilder.sort(sortBuilderOrder);
+
+        SearchRequestBuilder searchRequestBuilder = new SearchRequestBuilder(client, SearchAction.INSTANCE)
+            .setIndices(ML_TASK_INDEX)
+            .setSource(searchSourceBuilder);
+
+        searchRequestBuilder.execute(ActionListener.wrap(searchResponseActionListener::onResponse, exception -> {
+            log.error("index {} not found, the reason is {}", ML_TASK_INDEX, exception);
+            throw new IndexNotFoundException("index " + ML_TASK_INDEX + " not found");
```
zane-neo, line 270: We can't confirm this is `IndexNotFoundException`, please throw a MLException instead, and please wrap the original exception into the MLException like this:`throw new MLException(exception)`

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,329 @@
...
+            .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+        indexRequestBuilder.execute(ActionListener.wrap(indexResponse -> {
+            if (indexResponse.status() == RestStatus.CREATED || indexResponse.status() == RestStatus.OK) {
+                log.info("node id:{} insert retry times successfully", localNodeId);
+                indexResponseActionListener.onResponse(indexResponse);
+                return;
+            }
+            indexResponseActionListener.onFailure(new MLException("node id:" + localNodeId + " insert retry times unsuccessfully"));
+        }, indexResponseActionListener::onFailure));
```
zane-neo, line 329: Please add logs here when receiving `indexRequestBuilder.execute` exception.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        // At opensearch startup, get local node id, if not ml node,we ignored, just return without doing anything
+        if (!MLNodeUtils.isMLNode(clusterService.localNode())) {
+            return;
+        }
+
+        String localNodeId = clusterService.localNode().getId();
+        // auto reload all models of this local ml node
+        threadPool.generic().submit(() -> {
```
zane-neo, line 126: Please change this threadpool to `UPLOAD_THREAD_POOL` in MachineLearningPlugin since this is dedicated for uploading models.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: ok, I use `threadPool.executor(LOAD_THREAD_POOL)`

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReLoader.java
wujunshen: have committed the code

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -32,6 +32,11 @@ public class CommonValue {
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
     public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 3;
     public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 1;
+
+    public static final String ML_MODEL_RELOAD_INDEX = ".plugins-ml-model-reload";
```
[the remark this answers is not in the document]
wujunshen: After the communication with you and charlie, we will elaborate

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: cool～I will modify it according to what you said

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: thanks~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: ok, I have changed it~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: done~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
wujunshen: en ,yaliang have said it~ I have committed the latest code

zane-neo (approved):

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,331 @@
...
+
+    /**
+     * query task index, and get the result of "task_type"="LOAD_MODEL" and "state"="COMPLETED" and
+     * "worker_node" match nodeId
+     *
+     * @param localNodeId one of query condition
+     */
+    @VisibleForTesting
+    void queryTask(String localNodeId, ActionListener<SearchResponse> searchResponseActionListener) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().from(0).size(1);
```
ylwu-amzn, line 246: This query only return the latest load model task. If user have 3 models, this query only return 1 latest load model task for 1 model, the other 2 models' tasks won't be returned. So we can't reload all 3 models, just reload 1 model. Is that correct?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReloader.java
```diff
@@ -0,0 +1,331 @@
...
+    private final Client client;
+    private final ClusterService clusterService;
+    private final NamedXContentRegistry xContentRegistry;
+    private final DiscoveryNodeHelper nodeHelper;
+    private final ThreadPool threadPool;
+    private volatile Boolean enableAutoReloadModel;
+    private volatile Integer autoReloadMaxRetryTimes;
+
+    /**
+     * constructor method， init all the params necessary for model auto reloading
```
ylwu-amzn, line 75: This `，` after `constructor method` is not US-ASCII. That will cause infra team's CI workflow failure. https://github.com/opensearch-project/ml-commons/issues/736

PR #724, opened 2023-02-06
tune model config: change pooling mode to optional

ylwu-amzn:

### Description
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/ONNXSentenceTransformerTextEmbeddingTranslator.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/ONNXSentenceTransformerTextEmbeddingTranslator.java
@@ -36,7 +36,7 @@ public class ONNXSentenceTransformerTextEmbeddingTranslator implements ServingTr
     private String modelType;
 
     public ONNXSentenceTransformerTextEmbeddingTranslator(TextEmbeddingModelConfig.PoolingMode poolingMode, boolean normalizeResult, String modelType) {
-        this.poolingMode = poolingMode;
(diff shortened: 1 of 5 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/ONNXSentenceTransformerTextEmbeddingTranslator.java
```diff
@@ -36,7 +36,7 @@ public class ONNXSentenceTransformerTextEmbeddingTranslator implements ServingTr
     private String modelType;
 
     public ONNXSentenceTransformerTextEmbeddingTranslator(TextEmbeddingModelConfig.PoolingMode poolingMode, boolean normalizeResult, String modelType) {
-        this.poolingMode = poolingMode;
+        this.poolingMode = poolingMode == null ? TextEmbeddingModelConfig.PoolingMode.MEAN : poolingMode;
```
dhrubo-os, line 39: For onnx it looks like depending on the poolingMode, output varies a lot. I'm wondering for onnx upload if we should make the`pooling_mode` & `normalize_result` field mandatory from the config file?

dhrubo-os (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_embedding/ONNXSentenceTransformerTextEmbeddingTranslator.java
ylwu-amzn: Actually I'm not sure if there are any models which don't need these two fields or not. Maybe should leave it open to user for now unless we are 100% sure.

Zhangxunmt (approved):

PR #725, opened 2023-02-07
check state before deleting model or task

rbhavna:
Signed-off-by: Bhavana Goud Ramaram <rbhavna@amazon.com>

### Description
This PR check the condition for model state before deleting a model. If the model is in loaded/loading/partially loaded state, it throws exception and tells user to unload the model first. Previously, the delete API would delete the model irrespective of any model state which was no desirable.

Also checks condition for task state before deleting a task. If task is in running state, it throws an exception.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
@@ -5,8 +5,11 @@
 
 package org.opensearch.ml.action.models;
 
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.common.MLModel.MODEL_ID_FIELD;
+import static org.opensearch.ml.utils.MLNodeUtils.createXContentParserFromRegistry;
+import static org.opensearch.ml.utils.RestActionUtils.getFetchSourceContext;
 
 import lombok.AccessLevel;
 import lombok.experimental.FieldDefaults;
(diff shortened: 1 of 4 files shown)


Zhangxunmt (approved):
LGTM. But to be on the safe side, are you still planning to check the model and task status in the cache memory to make sure they are not deleted? You can add it in a separate PR. https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/model/MLModelCache.java, and,  https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/task/MLTaskCache.java.

rbhavna:
> LGTM. But to be on the safe side, are you still planning to check the model and task status in the cache memory to make sure they are not deleted? You can add it in a separate PR. https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/model/MLModelCache.java, and, https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/task/MLTaskCache.java.

Yes, I think we might have to do it.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -44,36 +55,66 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
...
+                            });
+                        }
+                    } catch (Exception e) {
+                        log.error("Failed to parse ml model" + r.getId(), e);
+                        actionListener.onFailure(e);
                     }
-                    actionListener.onFailure(e);
                 }
-            });
+            }, e -> { actionListener.onFailure(new MLResourceNotFoundException("Fail to find model")); }), () -> context.restore()));
```
ylwu-amzn, line 117: From line 81, we are going to restore context before running listener to delete model. Will it fail on security enabled cluster?

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
rbhavna: Tested it on security enabled cluster and it works. Adding test results here.
Screenshot 2023-02-20 at 1 13 16 PM

Issue #787, opened 2023-03-08
Change response code for user errors


ylwu-amzn:
Some of the user errors result in 5xx errors. These errors should be 4xx. So far, the following cases have been identified:

* The customer provides invalid node id while performing model load operation.
* The customer tries to run inferences on a model that needs to be loaded but is not loaded by the customer yet.
* The customer performs any ML operation without provisioning ML nodes.

PR #790, opened 2023-03-09
change exception type; get function name from cache

ylwu-amzn:
### Description
1. change exception type to return 4xx error code rather than 5xx
2. get function name from cache
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/787
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -43,6 +43,7 @@ import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.Semaphore;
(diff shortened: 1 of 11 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -787,6 +788,10 @@ public FunctionName getModelFunctionName(String modelId) {
         return modelCacheHelper.getFunctionName(modelId);
     }
 
+    public Optional<FunctionName> getOptionalModelFunctionName(String modelId) {
```
jngz-es, line 791: Why we need this wrap function, not directly use modelCacheHelper.getOptionalFunctionName()?

ylwu-amzn: Model manager provides more functions which will make future development easier.

jngz-es: Make sense, thanks.

jngz-es (approved):

Zhangxunmt (approved):

Issue #805, opened 2023-03-16
[BUG] Model state becomes UNLOADED when unload model from partial ML nodes


ylwu-amzn:
If unload model from partial ML nodes, the model state becomes UNLOADED which is not correct. It should be PARTIALLY_LOADED.

PR #806, opened 2023-03-16
set model state as partially loaded if unload model from partial nodes

ylwu-amzn:
### Description
If user deploy model to multiple ML nodes, then unload model from one ML node, the model state will become `UNLOADED`. This is not correct. We should set the model state as `PARTIALLY_LOADED`. This PR fixed this bug.
 
### Issues Resolved
[[List any issues this PR will resolve]](https://github.com/opensearch-project/ml-commons/issues/805)
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/unload/TransportUnloadModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/unload/TransportUnloadModelAction.java
@@ -97,8 +97,19 @@ public class TransportUnloadModelAction extends
     ) {
         if (responses != null) {
             Map<String, List<String>> removedNodeMap = new HashMap<>();
+            Map<String, Integer> modelWorkNodeCounts = new HashMap<>();
             responses.stream().forEach(r -> {
                 Set<String> notFoundModels = new HashSet<>();
+                Map<String, Integer> nodeCounts = r.getModelWorkerNodeCounts();
+                if (nodeCounts != null) {
+                    for (Map.Entry<String, Integer> entry : nodeCounts.entrySet()) {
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/org/opensearch/ml/action/unload/TransportUnloadModelAction.java
```diff
@@ -128,7 +139,11 @@ protected UnloadModelNodesResponse newResponse(
                     BulkRequest bulkRequest = new BulkRequest();
                     for (String modelId : removedNodeMap.keySet()) {
                         UpdateRequest updateRequest = new UpdateRequest();
-                        updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(ImmutableMap.of(MODEL_STATE_FIELD, MLModelState.UNLOADED));
+                        int removedNodeCount = removedNodeMap.get(modelId).size();
+                        MLModelState mlModelState = modelWorkNodeCounts.get(modelId) > removedNodeCount
```
Zhangxunmt, line 143: No need to check if removedNodeCount > 0 here?  
modelWorkNodeCounts.get(modelId) > removedNodeCount && removedNodeCount > 0.

Zhangxunmt (approved):

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/action/unload/TransportUnloadModelAction.java
ylwu-amzn: removedNodeCount will always > 0, check line 121

PR #813, opened 2023-03-21
add exclude nodes setting

### LINKED ISSUE #812: Exclude ML nodes
Sometimes, user don't want to run ML tasks on some nodes. For example, they want to replace some nodes with other new more powerful nodes. So they can exclude the nodes first to avoid new ML requests routes to them. 
We can add a new cluster setting, so user can specify which nodes they want to block to run ML task.

ylwu-amzn:
### Description
Add exclude nodes setting.
If user don't want to run ML tasks on some ML nodes, they can use exclude the nodes with this setting. 
 
### Issues Resolved
#812 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/DiscoveryNodeHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/DiscoveryNodeHelper.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.ml.cluster;
 
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_EXCLUDE_NODE_NAMES;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ONLY_RUN_ON_ML_NODE;
 
 import java.util.ArrayList;
@@ -19,6 +20,7 @@ import lombok.extern.log4j.Log4j2;
 import org.opensearch.cluster.ClusterState;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.Strings;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.ml.common.CommonValue;
 import org.opensearch.ml.utils.MLNodeUtils;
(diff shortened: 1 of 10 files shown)


plugin/src/main/java/org/opensearch/ml/cluster/DiscoveryNodeHelper.java
jngz-es, line 1: Do we support a node as mix of ml and data roles?

Zhangxunmt (dismissed):

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
```diff
@@ -74,7 +74,7 @@ public void dispatch(ActionListener<DiscoveryNode> actionListener) {
 
     public void dispatchPredictTask(String[] nodeIds, ActionListener<DiscoveryNode> actionListener) {
         if (nodeIds == null || nodeIds.length == 0) {
-            throw new IllegalArgumentException("Model not loaded yet");
+            throw new IllegalArgumentException("no eligible node to run predict request");
```
jngz-es, line 77: Is it also possible to enter this branch if model not loaded?

plugin/src/main/java/org/opensearch/ml/cluster/DiscoveryNodeHelper.java
ylwu-amzn: Yes, if a node has role of ml, then it can run ML tasks, no matter it has other role or not.

plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java
ylwu-amzn: No, it will fail earlier.

plugin/src/main/java/org/opensearch/ml/cluster/DiscoveryNodeHelper.java
```diff
@@ -68,6 +78,26 @@ public DiscoveryNode[] getEligibleNodes() {
...
+        for (DiscoveryNode node : nodes) {
+            if (excludedNodeNames != null && excludedNodeNames.contains(node.getName())) {
+                continue;
+            }
+            if (MLNodeUtils.isMLNode(node)) {
+                eligibleNodes.add(node.getId());
+            }
+            if (!onlyRunOnMLNode && node.isDataNode() && isEligibleDataNode(node)) {
+                eligibleNodes.add(node.getId());
+            }
```
jngz-es, line 96: If a node has both ml and data roles, eligibleNodes could have duplicated nodes. Should we avoid the duplicate?

ylwu-amzn: Good catch, will fix

Zhangxunmt (approved):

PR #824, opened 2023-03-23
Change the ziputil dependency to fix a potential security concern

b4sjoo:
### Description
We've written our own ZipUtil class, where unzip method used Apache Commons’ ZipFile classes instead of ZipInputStream class. This can fix a potential security vulnerability.

### Issues Resolved
This resolved a sev3 ticket.
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,53 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.Enumeration;
+
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipFile;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * A util class contains file related operations.
```
ylwu-amzn, line 19: This java doc seems not so accurate, should be zip file related ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@Log4j2
+public class ZipUtils {
+    public static void unzip(File zipFile, Path dest) {
```
ylwu-amzn, line 23: Add java doc for public method

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            ZipFile file = new ZipFile(zipFile);
+            Enumeration<ZipArchiveEntry> en = file.getEntries();
+            ZipArchiveEntry ze;
+            while (en.hasMoreElements()) {
+                ze = en.nextElement();
+                String name = ze.getName();
+                if (name.contains("..")) {
```
ylwu-amzn, line 31: What does this mean? 

Do you have any test case covering this branch? Same question for line 35

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,53 @@
...
+                                "Parent path should never be null: " + f);
+                    }
+                    Files.createDirectories(parentFile);
+                    InputStream is = file.getInputStream(ze);
+                    Files.copy(is, f, StandardCopyOption.REPLACE_EXISTING);
+                    is.close();
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e.getMessage(), e);
```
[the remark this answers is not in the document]
b4sjoo: I just follow some samples here, changed it into `IllegalArgumentException` and specified error message.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
b4sjoo: This is because previous djl unzip class wrote like this. I think it is for security concern but I cannot reproduce any file like that, so I cannot write a test case to cover it. Maybe delete it?

ml-algorithms/src/test/java/org/opensearch/ml/engine/utils/ZipUtilsTest.java
```diff
@@ -0,0 +1,36 @@
...
+        Files.createDirectories(output);
+        ZipUtils.unzip(file, output);
+    }
+
+    @Test
+    public void testUnzipFile() throws IOException, URISyntaxException {
+        File testZipFile = new File(Objects.requireNonNull(getClass().getResource("foo.zip")).toURI());
+        Path output = Paths.get("build/output");
+        Files.createDirectories(output);
+        ZipUtils.unzip(testZipFile, output);
```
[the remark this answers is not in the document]
b4sjoo: Added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
b4sjoo: Added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
b4sjoo: Corrected.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.engine.utils;
+
+import java.io.*;
```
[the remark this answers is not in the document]
b4sjoo: Corrected.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,58 @@
...
+                                "Parent path should never be null: " + f);
+                    }
+                    Files.createDirectories(parentFile);
+                    InputStream is = file.getInputStream(ze);
+                    Files.copy(is, f, StandardCopyOption.REPLACE_EXISTING);
+                    is.close();
+                }
+            }
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Input file should never be null", e);
```
ylwu-amzn, line 55: Log the error?

Is "null file" is the only cause for IOException. How about make the error message more general ?

b4sjoo: What about "Wrong input file"

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public static void unzip(File zipFile, Path dest) {
+        try {
+            ZipFile file = new ZipFile(zipFile);
+            Enumeration<ZipArchiveEntry> en = file.getEntries();
+            ZipArchiveEntry ze;
+            while (en.hasMoreElements()) {
+                ze = en.nextElement();
+                String name = ze.getName();
+                Path f = dest.resolve(name).toAbsolutePath();
```
ylwu-amzn, line 39: minor: generally we don't use single char as variable name except for some special case like `i` in for loop.  How about change to more readable name?

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,60 @@
...
+                if (!f.normalize().startsWith(dest))
+                    throw new RuntimeException("Bad zip entry");
```
ylwu-amzn, line 41: Can you write a test case to cover this branch?

ylwu-amzn: Never mind, I guess the edge case should be like this one https://codeql.github.com/codeql-query-help/java/java-zipslip/#recommendation

If zip entry has name ` ..\sneaky-file` , then we may unzip to a parent folder.

ml-algorithms/src/test/java/org/opensearch/ml/engine/utils/ZipUtilsTest.java
```diff
@@ -0,0 +1,39 @@
...
+        Path path = Paths.get("build/empty.zip");
+        File file = new File(path.toUri());
+        Path output = Paths.get("build/output");
+        Files.createDirectories(output);
+        ZipUtils.unzip(file, output);
+    }
+
+    @Test
+    public void testUnzipFile() throws IOException, URISyntaxException {
+        File testZipFile = new File(Objects.requireNonNull(getClass().getResource("foo.zip")).toURI());
```
dhrubo-os, line 32: We already have multiple zip files in the same resource folder. Do we need to add another zip file only for this test?

b4sjoo: I think the answer is yes because we cannot ensure these zip file will be there forever with same version or name.

dhrubo-os: For example, `all-MiniLM-L6-v2_onnx.zip` is being used in 3 difference tests. So I don't think this file will be disappear suddenly. Even if somebody replace this file, that will fail the test and they can update the changed file name accordingly.

In addition, in production this util will unzip a model zip file then why not use the similar file in test...

b4sjoo: Meanwhile I want to build a small size folder that has an empty sub-folder, a non-empty sub-folder and a file to cover most cases, I don't think we have that kind of zip file.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@Log4j2
+public class ZipUtils {
+
+    /**
+     * Uncompressed a zip file.
+     * @param zipFile zip file to be uncompressed
+     * @param dest the destination path of this uncompress
+     */
+    public static void unzip(File zipFile, Path dest) {
```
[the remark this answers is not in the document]
b4sjoo: I think that's Ok because `mlEngine.getMlCachePath()` eventually goes to this `unzip` method as the `@dest` parameter

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
b4sjoo: Corrected

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ZipUtils.java
b4sjoo: Yes but actually this edge case is really hard to create, linux does not allow me to do that...

ylwu-amzn (approved):

dhrubo-os (approved):

PR #825, opened 2023-03-24
remote inference baseline, connector index, and create connector API

Zhangxunmt:
### Description
Remote inference baseline code to unblock API development. Unit/Integ tests will be added later.

### Issues Resolved
[746]


ylwu-amzn:
How about create a feature branch and develop there first?

PR #836, opened 2023-03-28
bugfix and refactor for the fault tolerance feature

wujunshen:

### Description
the new feature: 
Auto redeploy model when cluster rebooted/node rejoin
 
When a ml node under the opensearch cluster halt down with some unknown reasons. The models under this node will be broken and impact the process of the inference or reduced performance. So we add a new feature: When a ml node halt down, we reboot this ml node, the opensearch on this node will auto redeploy all the models under this node,and user will not reload the model manually. Even  in extreme cases, if the redeploy operation is still unsuccessful, opensearch will also tell the user via logs that the redeploy was unsuccessful.

### Issues Resolved
 
### Check List
  - [x] All tests pass


plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+
+    /**
+     * constructor method， init all the params necessary for model auto redeploying
+     *
+     * @param clusterService clusterService
+     * @param threadPool threadPool
+     * @param client client
+     * @param xContentRegistry xContentRegistry
+     * @param settings settings
+     *                @param nodeEnvironment nodeEnvironment
```
ylwu-amzn, line 90:
```suggestion
     * @param nodeEnvironment nodeEnvironment
```

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.threadpool.ThreadPool;
+
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * Manager class for ML models and nodes. It contains ML model auto redeploy operations etc.
+ *
+ * @author wjunshen
```
ylwu-amzn, line 68: Generally we don't add author in our code base

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+                        }
+                    }
+
+                    modelIds.add(mlModel.getModelId());
+                } catch (Exception e) {
+                    throw new MLException("failed to parse ml model", e);
+                }
+            }
+
+            modelIds.forEach(modelId -> log.info("the model id {} need to deploy or undeploy", modelId));
```
ylwu-amzn, line 195: This log looks confusing, let's remove this log line or make it clear, whether the model need to deploy or undeploy. I would suggest remove this log and add logs before line 204, 210.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /**
+     * auto re-deploy model under the local node id<br>
+     * @param localNodeId local node id
+     * @throws MLException exception
+     */
+    @VisibleForTesting
+    void autoReDeployModel(String localNodeId) throws MLException {
+        // wait the index ".plugins-ml-model" finish initialization and get all the models to be deployed
```
ylwu-amzn, line 164: Seems this comment should be moved to before line 147?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+    * @throws MLException exception
+    */
+    @VisibleForTesting
+    void unDeployModel(String localNodeId, List<String> modelIds) throws MLException {
+        StepListener<MLUndeployModelNodesResponse> unDeployModelStep = new StepListener<>();
+
+        ActionListener<MLUndeployModelNodesResponse> listener = ActionListener
+            .wrap(response -> log.info("the models is auto undeploying under the node {} ", localNodeId), exception -> {
+                log.error("fail to undeploy models under the node id: {} \nthe reason is: {}", localNodeId, exception);
+                throw new MLException(exception);
```
ylwu-amzn, line 332: Why we need to throw exception here?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /**
+      * check whether the local node is ml node
+     *
+     * @param roleName role name
+    * @return if the local node is ml node,return true,else return false
+     */
+    @VisibleForTesting
+    boolean isNode(String roleName) {
```
ylwu-amzn, line 462: How about change the function name to `hasRoleName`?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -68,4 +68,10 @@ private MLCommonsSettings() {}
         .simpleString("plugins.ml_commons.exclude_nodes._name", Setting.Property.NodeScope, Setting.Property.Dynamic);
     public static final Setting<Boolean> ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN = Setting
         .boolSetting("plugins.ml_commons.allow_custom_deployment_plan", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
+    public static final Setting<Boolean> ML_COMMONS_MODEL_AUTO_REDEPLOY_ENABLE = Setting
+        .boolSetting("plugins.ml_commons.model.autoredeploy.enable", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
ylwu-amzn, line 73:
```suggestion
        .boolSetting("plugins.ml_commons.model.auto_redeploy.enable", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
```

ylwu-amzn: Should we disable auto redeploy by default?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: have modified

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: remove

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: has moved

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: moved

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: remove

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: changed

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
wujunshen: have reset to true by default

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -68,4 +68,10 @@ private MLCommonsSettings() {}
...
+
+    public static final Setting<Integer> ML_MODEL_REDEPLOY_MAX_RETRY_TIMES = Setting
+        .intSetting("plugins.ml_commons.model.autoredeploy.retrytimes", 3, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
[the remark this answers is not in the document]
wujunshen: changed

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
wujunshen: sorry,yaliang
If it is set to true by default,the ci process on github will report error,you can see these 2 links:
```
https://github.com/opensearch-project/ml-commons/actions/runs/4583717456/jobs/8098013040?pr=836
https://github.com/opensearch-project/ml-commons/actions/runs/4583717456/jobs/8098013069?pr=836
```
So, I changed to false by default again~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -12,11 +12,10 @@
 import static org.opensearch.ml.plugin.MachineLearningPlugin.DATA_ROLE_NAME;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.DEPLOY_THREAD_POOL;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.ML_ROLE_NAME;
-import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_MODEL_AUTO_REDEPLOY_ENABLE;
-import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_MODEL_AUTO_REDEPLOY_RETRY_TIMES;
-import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ONLY_RUN_ON_ML_NODE;
+import static org.opensearch.ml.settings.MLCommonsSettings.*;
```
ylwu-amzn, line 15: avoid using import *

wujunshen: yes, I am changing this

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,511 @@
...
+    /**
+     * get all models from the index <br>
+     *
+     * @param searchResponseActionListener listener for search response
+     */
+    @VisibleForTesting
+    void queryModels(ActionListener<SearchResponse> searchResponseActionListener) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+
+        QueryBuilder queryBuilder = QueryBuilders.boolQuery().mustNot(QueryBuilders.existsQuery(MLModel.CHUNK_NUMBER_FIELD));
```
ylwu-amzn, line 405: Seems no filter for model state. So we will query all models?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,511 @@
...
+            try {
+                if (onlyRunOnMlNode && hasRoleName(DATA_ROLE_NAME)) {
+                    // if onlyRunOnMlNode is true and node is data node
+                    // undeploy all the models on this local node
+                    modelIdAndNodeIds
+                        .forEach(
+                            (modelId, nodeIds) -> log.info("the model id {} need to undeploy under the node ids: {}", modelId, nodeIds)
+                        );
+                    unDeployModel(localNodeId, modelIdAndNodeIds);
+                } else {
```
ylwu-amzn, line 225: For master node, will it go to this `else` branch?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,511 @@
...
+            if (retryTimes > autoDeployMaxRetryTimes) {
+                log.info("Node: {} has reached to the max retry limit, failed to deploy models", localNodeId);
+                return;
+            }
+
+            String[] modelIds = modelIdAndNodeIds.keySet().toArray(new String[] {});
+            final int finalRetryTimes = retryTimes + 1;
+            List<ActionListener<MLDeployModelResponse>> listeners = Arrays
+                .stream(modelIds)
+                .map(e -> ActionListener.<MLDeployModelResponse>wrap(response -> {
```
ylwu-amzn, line 284: minor: how about change `e` to `modelId` or `id`?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: Yes, our Shanghai team is discussing that all models need to be re-deployed

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: done

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: I will changed it,let the model which the state is REGISTERED,UPLOADED,DEPLOY_FAILED,LOAD_FAILED,UNDEPLOYED,UNLOADED will not be queried~

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,511 @@
...
+
+                    throw new MLException(exception);
+                }))
+                .collect(Collectors.toList());
+
+            for (int i = 0; i < modelIds.length; i++) {
+                String[] nodeIds = new ArrayList<>(modelIdAndNodeIds.get(modelIds[i])).toArray(new String[] {});
+
+                ActionListener<MLDeployModelResponse> currentListener = listeners.get(i);
+                autoDeployModelByNodeAndModelId(nodeIds, modelIds[i], currentListener);
```
ylwu-amzn, line 318: This deploy method is async, it will return task id directly, so that means we will deploy multiple models in parallel?

wujunshen: The previous code looked like this
 ```
for (int i = 0; i < modelIds.length; i++) {
                String currentModelId = modelIds[i].
                ActionListener<MLDeployModelResponse> listener = listeners.get(i).
                nodeIds = new ArrayList<>(modelIdAndNodeIds.get(currentModelId)).toArray(new String[] {}).

                if (i == 0) {
                    autoDeployModelByNodeAndModelId(nodeIds, currentModelId, listener).
                }

                if (i + 1 < modelIds.length) {
                    String nextModelId = modelIds[i + 1].
                    nodeIds = new ArrayList<>(modelIdAndNodeIds.get(nextModelId)).toArray(new String[] {}).
                    ActionListener<MLDeployModelResponse> listenerNext = listeners.get(i + 1).
                    autoDeployModelByNodeAndModelId(nodeIds, nextModelId, listenerNext).
                }
            }

```
I think it can be optimized to the current code, so it's now like this.

File: plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java [hunk 14]
```diff
@@ -0,0 +1,539 @@
...
+
+        // auto redeploy all models of this local node
+        threadPool.executor(DEPLOY_THREAD_POOL).execute(() -> {
+            long sleepTimeUnit = 1000L;
+            int numRetries = 4;
+            int numTries = 0;
+
+            // wait the index ".plugins-ml-model" finish initialization and get all the models to
+            // be deployed
+            while (numTries < numRetries) {
```
zane-neo, line 153: If the auto reload succeed after one try, shouldn't we stop the retry? using while loop here will run 3 times even if the reload succeed, right?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,539 @@
...
+    /**
+     * get all models from the index <br>
+     *
+     * @param searchResponseActionListener listener for search response
+     */
+    @VisibleForTesting
+    void queryModels(ActionListener<SearchResponse> searchResponseActionListener) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+
+        QueryBuilder queryBuilder = QueryBuilders
```
zane-neo, line 424: Is it possible to use only a must condition, e.g. LOADED/PARTIAL_LOADED to build the query? This needs less code, right?

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,539 @@
...
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN, it -> allowCustomDeploymentPlan = it);
+    }
+
+    /** the main method: model auto redeploying */
+    public void autoReDeployModel() {
+        log.info("auto redeploy model enabled: {} ", enableAutoReDeployModel);
+
+        // if we don't need to redeploy automatically, just return without doing anything
+        if (!enableAutoReDeployModel) {
```
zane-neo, line 139: Move the master node check here to make sure all the checks are gathered together, this will make it easy for  reading code.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,539 @@
...
+            return;
+        }
+
+        String localNodeId = nodeEnvironment.nodeId();
```
zane-neo, line 143: If we move the auto reload logic to MLCommonsClusterEventListener and reloaded by master node, we shouldn't using this local node id.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
```diff
@@ -0,0 +1,539 @@
...
+
+            for (SearchHit hit : hits) {
+                try (XContentParser parser = MLNodeUtils.createXContentParserFromRegistry(xContentRegistry, hit.getSourceRef())) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                    MLModel mlModel = MLModel.parse(parser);
+                    String[] planningWorkerNodes = mlModel.getPlanningWorkerNodes();
+                    Set<String> nodeIds = Arrays.stream(planningWorkerNodes).collect(Collectors.toSet());
+
+                    if (isMasterNode) {
+                        if (allowCustomDeploymentPlan && !mlModel.isDeployToAllNodes()) {
```
zane-neo, line 207: When allowCustomDeploymentPlan is false, the load API do not accept nodeIds as parameter, we shouldn't pass any nodeIds in this case.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: in line161, it has `break` keyword. If reload successfully the first time, it will break the while loop at once.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: I found the `termQuery` or `termsQuery` don't work in unit test.So I changed it to `marchPhraseQuery`, let one by one filter

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: Because at here, the whole cluster state has not been initialized, and I use the `clusterservice` class to check whether the node is master node, at this time `clusterService.state().nodes()` maybe throw `nullPointerException`, so we can not do master node check at here

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: that's right~ When the `allowCustomDeploymentPlan` is false,whatever the `onlyRunOnMlNode` is true or false,the `nodeids` paramerter will be the empty collection, I will changed it.

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: But there is another scenario to consider. When a new ml or data node joins the cluster, if `allowCustomDeploymentPlan` is true, we still need to get the local node id and perform re-deploy operation

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: changed it,and I have committed it to this PR

plugin/src/main/java/org/opensearch/ml/model/MLModelAutoReDeployer.java
wujunshen: @ylwu-amzn  I used `StepListener` class to let model deployed one by one, u can see the latest code

PR #849, opened 2023-04-07
Add an array filter to our serialize/deserialize methods and narrow down previous filter

b4sjoo:
### Description
Previous we did not add filter to the depth and byte array length to our deserialization method, which can cause a denial of service attack via very long arrays or nested objects. Meanwhile we have an allowlist that is too general so malicious files could potentially bypass. Now we narrow down the allowlist, add a reject list, and also add a depth and length filter. In long term we will try to change this byte stream deserialization method into json or protocol buffer.
 
### Issues Resolved
This is a short-term fix in regards to a security ticket.
 
### Check List
  - [x] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ModelSerDeSer.java
```diff
@@ -10,10 +10,7 @@
 import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.engine.exceptions.ModelSerDeSerException;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.ObjectOutputStream;
+import java.io.*;
```
ylwu-amzn, line 13: Avoid import *

ylwu-amzn (approved):
LGTM. Thanks

dhrubo-os (approved):

PR #870, opened 2023-04-24
access validation for register/update model-group, register/get/delete/deploy/predict model

rbhavna:
### Description
access validation for register/update model-group, register/get/delete/deploy/predict model

This is not final. Search model group and undeploy model APIs in this PR are not yet completed. Will be adding more changes in subsequent PRs.
 
### Check List
  - [ ] All tests pass


client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -200,13 +200,33 @@ default ActionFuture<SearchResponse> searchModel(SearchRequest searchRequest) {
...
     /**
      * For more info on search model, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/#search-model
      * @param searchRequest searchRequest to search the ML Model
      * @param listener action listener
      */
     void searchModel(SearchRequest searchRequest, ActionListener<SearchResponse> listener);
 
+
+    /**
+     * For more info on search model, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/#search-model
```
dhrubo-os, line 213: change the documentation which is not applicable here. 

Also seems like `searchModel` & `searchModelGroup` both expecting the same parameter and performing the same task. May be we can merge these two functions by `searchItem` or something else?

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
```diff
@@ -7,46 +7,106 @@
...
 public class MLModelGroup implements ToXContentObject {
     public static final String MODEL_GROUP_NAME_FIELD = "name";
     // We use int type for version in first release 1.3. In 2.4, we changed to
     // use String type for version. Keep this old version field for old models.
     public static final String DESCRIPTION_FIELD = "description";
+    public static final String TAGS_FIELD = "tags";
     public static final String LATEST_VERSION_FIELD = "latest_version";
+    public static final String BACKEND_ROLES_FIELD = "backend_roles";
+    public static final String OWNER = "owner";
+    public static final String MODEL_GROUP_ID_FIELD = "model_group_id";
```
dhrubo-os, line 37: Lets add comments on all these variables.

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
```diff
@@ -26,29 +27,29 @@ public class MLRegisterModelGroupInput implements ToXContentObject, Writeable{
     public static final String NAME_FIELD = "name"; //mandatory
     public static final String DESCRIPTION_FIELD = "description";
     public static final String TAGS_FIELD = "tags"; //mandatory
-    public static final String MODEL_IDS_FIELD = "model_ids";
+    public static final String BACKEND_ROLES_FIELD = "backend_roles";
```
dhrubo-os, line 30: please add comments.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -68,4 +68,7 @@ private MLCommonsSettings() {}
         .simpleString("plugins.ml_commons.exclude_nodes._name", Setting.Property.NodeScope, Setting.Property.Dynamic);
     public static final Setting<Boolean> ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN = Setting
         .boolSetting("plugins.ml_commons.allow_custom_deployment_plan", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
```
dhrubo-os, line 71: Add comment about this settings.

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
rbhavna: Thanks Dhurbo. Yeah is something we wanted to discuss with the FE team.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -108,15 +107,50 @@ public TransportDeployModelAction(
...
+
     }
 
     @Override
     protected void doExecute(Task task, ActionRequest request, ActionListener<MLDeployModelResponse> listener) {
         MLDeployModelRequest deployModelRequest = MLDeployModelRequest.fromActionRequest(request);
         String modelId = deployModelRequest.getModelId();
+        log.info(modelId);
+
+        MLModelGetRequest mlModelGetRequest = new MLModelGetRequest(modelId, false);
```
ylwu-amzn, line 124: Why create a new `MLModelGetRequest` ? I see the `mlModelGetRequest` only being used by line 125 to get `mlModelGetRequest.isReturnContent()`. As you have hard coded `returnContent` as false in this line (line 124), then the `mlModelGetRequest.isReturnContent()` will always return `false`, right?

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -108,15 +107,50 @@ public TransportDeployModelAction(
...
+        FetchSourceContext fetchSourceContext = getFetchSourceContext(mlModelGetRequest.isReturnContent());
+        GetRequest getRequest = new GetRequest(ML_MODEL_INDEX).id(modelId).fetchSourceContext(fetchSourceContext);
+        User user = RestActionUtils.getUserContext(client);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            client.get(getRequest, ActionListener.wrap(r -> {
```
ylwu-amzn, line 130: You can use `MLModelManager#getModel` method

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupInput.java
```diff
@@ -0,0 +1,114 @@
...
+        this.modelGroupID = modelGroupID;
+        this.name = name;
+        this.description = description;
+        this.backendRoles = backendRoles;
+    }
+
+    public MLUpdateModelGroupInput(StreamInput in) throws IOException {
+        this.modelGroupID = in.readString();
+        this.name = in.readString();
+        this.description = in.readString();
```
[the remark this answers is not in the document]
rbhavna: yes, will make the change.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static boolean validateModelGroupAccess(User user, String modelGroupId, Client client) {
+        List<String> userBackendRoles = user.getBackendRoles();
+        log.info(userBackendRoles);
+
+        GetRequest getModelGroupRequest = new GetRequest(ML_MODEL_GROUP_INDEX).id(modelGroupId);
+        log.info(modelGroupId);
+        List<String> modelGroupBackendRoles = new ArrayList<>();
+        try {
+            Map<String, Object> source = client.get(getModelGroupRequest).get().getSourceAsMap();
```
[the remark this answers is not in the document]
rbhavna: got it. Will make the change

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -154,6 +160,14 @@ public void searchModel(SearchRequest searchRequest, ActionListener<SearchRespon
         }, listener::onFailure));
     }
 
+    @Override
+    public void searchModelGroup(SearchRequest searchRequest, ActionListener<SearchResponse> listener) {
```
ylwu-amzn, line 164: Why do we need to expose this API in ML client?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -199,6 +217,31 @@ private MLUndeployModelNodeResponse createUndeployModelNodeResponse(MLUndeployMo
 
         Map<String, Integer> modelWorkerNodeCounts = new HashMap<>();
         boolean specifiedModelIds = modelIds != null && modelIds.length > 0;
+
+        Set<String> invalidAccessModels = new HashSet<>();
+        User user = RestActionUtils.getUserContext(client);
```
ylwu-amzn, line 222: As we discussed offline, we can't get user information in this class. I debugged by adding a new transport action similar to deploy model, we can get user information. You can refer to my code https://github.com/ylwu-amzn/ml-commons/commit/862ad390ca9624a3e11a3c2d38974054e8482fc9, you can get user information and verify model group backend role in this class `TransportUndeployModelsAction`.

rbhavna: Thanks Yaliang. I think this should work then. Will add these changes.

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
rbhavna: Sorry my bad. Need to remove it from here. I was making changes similar to that of search model initially and forgot to delete it from here

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLUndeployModelNodesResponse> listener) {
+        MLUndeployModelsRequest undeployModelsRequest = MLUndeployModelsRequest.fromActionRequest(request);
+        String[] modelIds = undeployModelsRequest.getModelIds();
+        String[] targetNodeIds = undeployModelsRequest.getNodeIds();
+        boolean specifiedModelIds = modelIds != null && modelIds.length > 0;
+        modelIds = specifiedModelIds ? modelIds : mlModelManager.getAllModelIds();
+        Set<String> invalidAccessModels = new HashSet<>();
```
ylwu-amzn, line 112: HashSet is not thread safe

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -153,7 +145,15 @@
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.watcher.ResourceWatcherService;
 
-import com.google.common.collect.ImmutableList;
+import java.util.Collection;
```
ylwu-amzn, line 148: Are these new imports necessary? If not, you can remove these

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -0,0 +1,170 @@
...
+
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLUndeployModelNodesResponse> listener) {
+        MLUndeployModelsRequest undeployModelsRequest = MLUndeployModelsRequest.fromActionRequest(request);
+        String[] modelIds = undeployModelsRequest.getModelIds();
+        String[] targetNodeIds = undeployModelsRequest.getNodeIds();
+        boolean specifiedModelIds = modelIds != null && modelIds.length > 0;
+        modelIds = specifiedModelIds ? modelIds : mlModelManager.getAllModelIds();
```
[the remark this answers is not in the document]
ylwu-amzn: Can you test on multi-node cluster and confirm if the code can work or not?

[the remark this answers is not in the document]
rbhavna: Yeah okay. Will do some testing

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
rbhavna: These are not added by my changes. They were already there

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
rbhavna: Oh got it. Will look into other alternatives then. Thanks

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
```diff
@@ -17,39 +17,46 @@
...
+    public static final String ADD_ALL_BACKEND_ROLES = "add_all_backend_roles"; //optional
 
     private String name;
     private String description;
-    private List<String> tags;
-    private List<String> models;
+    private Map<String, Object> tags;
+    private List<String> backendRoles;
+    private Boolean isPublic;
+    private Boolean addAllBackendRoles;
```
dhrubo-os, line 39: [nit] Using "is" or "has" as prefixes for boolean variables can help improve code readability by making it clear that the variable represents a condition or state. May be we can rename this variable?

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
```diff
@@ -17,39 +17,46 @@
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 @Data
 public class MLRegisterModelGroupInput implements ToXContentObject, Writeable{
```
dhrubo-os, line 25: If the responsibility of this class to take input for the MLmodel group registration, then may be we can rename it to:
`MLModelGroupRegisterInput`?

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupInput.java
```diff
@@ -0,0 +1,117 @@
...
+@Data
+public class MLUpdateModelGroupInput implements ToXContentObject, Writeable {
+
+    public static final String MODEL_GROUP_ID = "model_group_id";
+    public static final String NAME = "name";
+    public static final String DESCRIPTION = "description";
+    public static final String BACKEND_ROLES = "backend_roles";
+
+
+    private String modelGroupID;
```
dhrubo-os, line 32: So user can't update tags?

plugin/build.gradle
```diff
@@ -292,7 +309,7 @@ jacocoTestCoverageVerification {
             excludes = jacocoExclusions
             limit {
                 counter = 'BRANCH'
-                minimum = 0.7
```
dhrubo-os, line 295: Can we add a `todo` here at least

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public static boolean isAdmin(User user) {
+        if (user == null) {
+            return false;
+        }
+        if (CollectionUtils.isEmpty(user.getRoles())) {
+            return false;
+        }
+        return user.getRoles().contains("all_access");
```
dhrubo-os, line 93: is it possible to have `ALL_ACCESS`

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+import org.opensearch.ml.common.MLModelGroup;
+import org.opensearch.ml.common.exception.MLValidationException;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+import com.google.common.collect.ImmutableList;
+
+@Log4j2
+public class SecurityUtils {
+
+    public static void validateModelGroupAccess(User user, String modelGroupId, Client client, ActionListener<Boolean> listener) {
```
dhrubo-os, line 36: Looks like this method sends three different values on reponse. Make be we can add some comment on the function?

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+            return false;
+        }
+        return user.getRoles().contains("all_access");
+    }
+
+    public static boolean isOwner(Map<String, Object> owner, User user) {
+        if (user == null || owner == null || !owner.containsKey("name")) {
+            return false;
+        }
+        return StringUtils.equals((String) owner.get("name"), user.getName());
```
dhrubo-os, line 100: this name means `username`?

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public static SearchSourceBuilder addUserBackendRolesFilter(User user, SearchSourceBuilder searchSourceBuilder) {
+        if (user == null) {
+            return searchSourceBuilder;
```
dhrubo-os, line 105: shouldn't we throw error here? Is this expected user to be null sometime?

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -166,4 +171,18 @@ public static Optional<String> getStringParam(RestRequest request, String paramN
...
+     * Generates a user string formed by the username, backend roles, roles and requested tenants separated by '|'
+     * (e.g., john||own_index,testrole|__user__, no backend role so you see two verticle line after john.).
+     * This is the user string format used internally in the OPENSEARCH_SECURITY_USER_INFO_THREAD_CONTEXT and may be
+     * parsed using User.parse(string).
+     * @param client Client containing user info. A public API request will fill in the user info in the thread context.
+     * @return parsed user object
+     */
+    public static User getUserContext(Client client) {
+        String userStr = client.threadPool().getThreadContext().getTransient(ConfigConstants.OPENSEARCH_SECURITY_USER_INFO_THREAD_CONTEXT);
+        logger.debug("Filtering result by " + userStr);
```
dhrubo-os, line 184: what are we filtering here?

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
```diff
@@ -7,46 +7,116 @@
...
+    public static final String CREATED_TIME_FIELD = "created_time"; //model group created time stamp
+    public static final String LAST_UPDATED_TIME_FIELD = "last_updated_time"; //updated whenever a new model version is created
     //SHA256 hash value of model content.
 
-    //TODO: add created time, updated time,
+    @Setter
     private String name;
     private String description;
+    private Map<String, Object> tags;
     private int latestVersion = 0;
```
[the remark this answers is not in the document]
rbhavna: Got it, thanks @dhrubo-os.

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
```diff
@@ -7,46 +7,116 @@
...
+        if (tags != null) {
+            out.writeBoolean(true);
+            out.writeMap(tags);
+        } else {
+            out.writeBoolean(false);
+        }
         out.writeInt(latestVersion);
+        out.writeStringCollection(backendRoles);
+        if (owner != null) {
+            owner.writeTo(out);
```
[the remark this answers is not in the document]
rbhavna: good catch. Will add it

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
rbhavna: something like isAddAllBackendRoles? When user specifies this field in the request, maybe addAllBackendRoles can be easier for user to know its purpose

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
rbhavna: Just named it so to maintain consistency with model level APIs

dhrubo-os: ok, ignore it then.

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupInput.java
rbhavna: Few more fields are to be added here including tags and access fields. Will publish them in next PR

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: User will all access is the admin and that is what we would check in this method

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: Will add some documentation in the next PR

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: yes its the owner's name of the model group user is trying to access

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: Hmm I doubt that case. But AD is using a similar check, so thought I would add it too

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
rbhavna: Just a debugging message to see user details. Filtering means the results(model groups/models) will be filtered based on user's back-end roles

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
dhrubo-os: Customer facing variable/parameter and internal variable doesn't need to be the same name. if the addAllBackendRoles represents a flag indicating whether backend roles should be added, you could rename it to `shouldAddBackendRoles`.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -63,9 +63,29 @@ public class CommonValue {
             "    },\n" +
             "    \""+MLModelGroup.LATEST_VERSION_FIELD+"\": {\n" +
             "      \"type\": \"integer\"\n" +
-            "    }\n" +
+            "    },\n" +
```
zane-neo, line 66: Seems tags mapping is missing

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLRegisterModelGroupInput.java
zane-neo: If a model has all backend roles, doesn't that mean it's a public model, is this `addAllBackendRoles` duplicated with `isPublic`?

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupInput.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+
+@Data
+public class MLUpdateModelGroupInput implements ToXContentObject, Writeable {
+
+    public static final String MODEL_GROUP_ID = "model_group_id";
```
zane-neo, line 26: Should we change these constants to ends up with `_FIELD` to keep consistency?

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupRequest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        ActionRequestValidationException exception = null;
+        if (updateModelGroupInput == null) {
+            exception = addValidationError("Update Model group input can't be null", exception);
+        }
+
+        return exception;
```
zane-neo, line 51:
```suggestion
        if (updateModelGroupInput == null)  return addValidationError("Update Model group input can't be null", exception);
```

common/src/main/java/org/opensearch/ml/common/CommonValue.java
rbhavna: Yes, had to remove tags and backend_roles mapping because they were throwing exception of incorrect mappings. Was working on them. Will add them back in next commit.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -56,13 +59,17 @@
...
+import org.opensearch.ml.utils.RestActionUtils;
+import org.opensearch.ml.utils.SecurityUtils;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 
+;
```
zane-neo, line 71: [nit] remove this semicolon

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zane-neo: The Setting name is `VALIDATE_BACKEDN_ROLE` but the configuration is `filter_by_backend_roles` which are not consistent, can we rename one to match another?

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+                    if (source.containsKey(MLModelGroup.OWNER)) {
+                        owner = (Map) source.get(MLModelGroup.OWNER);
+                    }
+                    if (access != null && StringUtils.equals(access, MLModelGroup.PUBLIC)) {
+                        listener.onResponse(Boolean.TRUE);
+                    } else if (access != null && StringUtils.equals(access, MLModelGroup.PRIVATE)) {
+                        if (isOwner(owner, user)) {
+                            listener.onResponse(Boolean.TRUE);
+                            return;
+                        }
```
zane-neo, line 67: Since you're almost parsing all the fields of MLModelGroup, it's better to use the parse XContent approach to create a MLModelGroup like this:
```
XContentParser parser = MLNodeUtils.createXContentParserFromRegistry(NamedXContentRegistry.EMPTY, x.getSourceRef())
                    ) {
                        MLModelGroup mlModelGroup = MLModelGroup.parse(parser);
```

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+import com.google.common.collect.ImmutableList;
+
+@Log4j2
+public class SecurityUtils {
+
+    public static void validateModelGroupAccess(User user, String modelGroupId, Client client, ActionListener<Boolean> listener) {
+        if (modelGroupId == null || isAdmin(user)) {
+            listener.onResponse(Boolean.TRUE);
```
zane-neo, line 38: Is there any specific reason why using encapsulated boolean type here?

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
zane-neo: Please make the name a constant with comments to avoid magic string issue.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+        List<String> backendRoles = user.getBackendRoles() != null ? user.getBackendRoles() : ImmutableList.of();
+        TermsQueryBuilder userRolesFilterQuery = QueryBuilders.termsQuery(ownerBackendRoleFieldName, backendRoles);
+        boolQueryBuilder.must(userRolesFilterQuery);
+        QueryBuilder query = searchSourceBuilder.query();
+        if (query == null) {
+            searchSourceBuilder.query(boolQueryBuilder);
+        } else if (query instanceof BoolQueryBuilder) {
+            ((BoolQueryBuilder) query).filter(boolQueryBuilder);
+        } else {
+            throw new MLValidationException("Search API does not support queries other than BoolQuery");
```
zane-neo, line 118: We do not support match/match_phrase/term query in search API? This doesn't seem correct.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -108,125 +116,143 @@ public TransportDeployModelAction(
...
-        mlStats.getStat(MLNodeLevelStat.ML_NODE_TOTAL_REQUEST_COUNT).increment();
-        DiscoveryNode[] allEligibleNodes = nodeFilter.getEligibleNodes();
-        Map<String, DiscoveryNode> nodeMapping = new HashMap<>();
-        for (DiscoveryNode node : allEligibleNodes) {
-            nodeMapping.put(node.getId(), node);
-        }
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            mlModelManager.getModel(modelId, null, excludes, ActionListener.wrap(mlModel -> {
+                SecurityUtils.validateModelGroupAccess(user, mlModel.getModelGroupId(), client, ActionListener.wrap(access -> {
+                    if ((filterByEnabled) && (Boolean.FALSE.equals(access))) {
```
zane-neo, line 137: If we can make sure there's no null returned by `validateModelGroupAccess`, then we can use primitive boolean to save the boxing/unboxing overhead.

plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java
```diff
@@ -0,0 +1,89 @@
...
+        MLSearchHandler mlSearchHandler,
+        Client client,
+        Settings settings,
+        ClusterService clusterService
+    ) {
+        super(MLModelGroupSearchAction.NAME, transportService, actionFilters, SearchRequest::new);
+        this.mlSearchHandler = mlSearchHandler;
+        this.client = client;
+        this.clusterService = clusterService;
+        filterByEnabled = ML_COMMONS_VALIDATE_BACKEND_ROLES.get(settings);
```
zane-neo, line 54: We can move this `filterByEnabled` consumer to a helper class to avoid duplicate code in every class.

plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java
```diff
@@ -0,0 +1,89 @@
...
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            validateRole(request, user, listener);
+        } catch (Exception e) {
+            log.error("Failed to search", e);
+            listener.onFailure(e);
+        }
+    }
+
+    private void validateRole(SearchRequest request, User user, ActionListener<SearchResponse> listener) {
+        if (user == null || !filterByEnabled || isAdmin(user)) {
```
zane-neo, line 71: It would be better to move these checkings to a util or helper with reasonable names, it will be more friendly to read the code and reduce duplication.

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportRegisterModelGroupAction.java
```diff
@@ -77,9 +84,35 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
     public void createModelGroup(MLRegisterModelGroupInput input, ActionListener<String> listener) {
         try {
             String modelName = input.getName();
+            User user = RestActionUtils.getUserContext(client);
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                MLModelGroupBuilder builder = MLModelGroup.builder();
+                if (Boolean.TRUE.equals(input.getIsPublic())) {
+                    builder = builder.access(MLModelGroup.PUBLIC);
+                } else if (Boolean.TRUE.equals(input.getAddAllBackendRoles())) {
```
zane-neo, line 92: Why using `else if`, does that mean `isPublic` and 'addAllBackendRoles' are mutually exclusive?

zane-neo, line 92: Can user choose not to add any backend roles which means it's a private model that can only be accessed by the owner self?

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLUpdateModelGroupResponse> listener) {
+        MLUpdateModelGroupRequest updateModelGroupRequest = MLUpdateModelGroupRequest.fromActionRequest(request);
+        MLUpdateModelGroupInput updateModelGroupInput = updateModelGroupRequest.getUpdateModelGroupInput();
+
+        String modelGroupId = updateModelGroupInput.getModelGroupID();
+        GetRequest getModelGroupRequest = new GetRequest(ML_MODEL_GROUP_INDEX).id(modelGroupId);
+
+        if (modelGroupId != null) {
```
zane-neo, line 72: Use reverse check to eliminate the nested ifs, e.g. `if (modelGroupId == null) throw new IllgalArgumentException(xxx)`

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+            User user = RestActionUtils.getUserContext(client);
+            try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                client.get(getModelGroupRequest, ActionListener.wrap(modelGroup -> {
+                    if (modelGroup.isExists()) {
+                        Map<String, Object> source = modelGroup.getSourceAsMap();
+                        Map<String, Object> owner = (Map) source.get(MLModelGroup.OWNER);
+                        String access = (String) source.get(MLModelGroup.ACCESS);
+
+                        if (SecurityUtils.isAdmin(user) || access == "public") {
```
zane-neo, line 81: Use public constant instead of magic string.

plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java
rbhavna: I tried it but it was not working as expectecd as this setting was to be injected into these constructors.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: No specific reason. We can use the primitive ones here.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: Thanks Zan, makes sense. Initially there were only backend_roles field I was using to filter them out. Later had to add more fields.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: Actually if the security plugin is disabled, I guess the user will be null

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupRequest.java
rbhavna: This is something in consistent with all other request classes. Also we might want to return null if updateModelGroupInput != null

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
rbhavna: Need to test few edge cases for this. Maybe using boxing/unboxing could be a safer option. If there is no case null is returned, will change them to primitive ones

common/src/main/java/org/opensearch/ml/common/transport/model_group/MLUpdateModelGroupRequest.java
zane-neo: Then it still can be optimized to:
```
if (updateModelGroupInput == null)  return addValidationError("Update Model group input can't be null", exception);
return null;
```

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
zane-neo: This could cause issue since frontend will be using these queries, we need more discussion on this to make sure there's no blocker for other requirements.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
zane-neo: It would be better to think all the cases through before implementing, test may cover some cases but not all because some cases would be difficult to mimic. Better think through if we need to return null or not here.

plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java
zane-neo: This setting can be injected into these constructors but not have to, it can be encapsulated in a helper class and using a public method to return its value, the helper class can be created in MachineLearningPlugin.java and been injected to other transport action constructors. This is not a p0 issue but if got time, please have a try again and if it's not working, please reach out to me, we can dig this issue together.

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+                            updateModelGroup(modelGroupId, source, updateModelGroupInput, listener);
+                        } else if (user.getName().equals(owner.get("name"))) {
```
zane-neo, line 83: Always add null check to user since it could be null if the security plugin is not enabled.

rbhavna: Good catch. Thanks Zan, will add the check wherever needed

plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java
rbhavna: Okay got it. Will try to implement it. If anything, will reach out to you

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+
+                                Boolean isRolePresent = updateModelGroupInput
+                                    .getBackendRoles()
+                                    .stream()
+                                    .allMatch(user.getBackendRoles().stream().collect(Collectors.toSet())::contains);
+
+                                if (!isRolePresent) {
+                                    log.error("Invalid Backend Roles provided in the input");
+                                    throw new IllegalArgumentException("Invalid Backend Roles provided in the input");
+                                }
```
zane-neo, line 100: These code can be optimized to:
1. check `updateModelGroupInput.getBackendRoles() != null`.
2. if  `updateModelGroupInput.getBackendRoles().size() != 0`, check input backend roles is a subset of user backend roles. If false, throw exception.
3. else set user backend roles to updateModelGroupInput.
This can make the code more performant since the backend roles stream will run only when it's necessary.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: @ylwu any inputs here?

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
rbhavna: Actually made a change to this block of code since the default action for empty back-end roles is now private. Will push in my next commit in a while

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+                            }
+
+                            updateModelGroup(modelGroupId, source, updateModelGroupInput, listener);
+                        } else {
+                            throw new IllegalArgumentException("User doesn't have valid privilege to perform this operation");
```
zane-neo, line 105: Only the model is public or user is admin/owner has the permission to update the model? Others with a subset backend roles can not update the model, correct?

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
zane-neo: Another question is: don't we use addAllBackendRoles flag to determine if we add all the user's backend roles to the model?

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -0,0 +1,147 @@
...
+        if (StringUtils.isNotBlank(updateModelGroupInput.getName())) {
+            source.put(MLUpdateModelGroupInput.NAME, updateModelGroupInput.getName());
+        }
+        if (StringUtils.isNotBlank(updateModelGroupInput.getDescription())) {
+            source.put(MLUpdateModelGroupInput.DESCRIPTION, updateModelGroupInput.getDescription());
+        }
+        if (updateModelGroupInput.getBackendRoles() != null) {
+            source.put(MLUpdateModelGroupInput.BACKEND_ROLES, updateModelGroupInput.getBackendRoles());
+        }
+        UpdateRequest updateModelGroupRequest = new UpdateRequest();
```
zane-neo, line 135: Tags should also be able to be updated.

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public static void validateModelGroupAccess(User user, String modelGroupId, Client client, ActionListener<Boolean> listener) {
+        if (modelGroupId == null || isAdmin(user)) {
+            listener.onResponse(Boolean.TRUE);
+            return;
+        }
+
+        List<String> userBackendRoles = user.getBackendRoles();
+
+        GetRequest getModelGroupRequest = new GetRequest(ML_MODEL_GROUP_INDEX).id(modelGroupId);
```
zane-neo, line 44: We should cache the permission information in cache to ensure the performance, every time reading a model group to check the permission is not performant, especially when it's inference cases.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -198,10 +207,10 @@ private MLUndeployModelNodeResponse createUndeployModelNodeResponse(MLUndeployMo
         String[] modelIds = MLUndeployModelNodesRequest.getModelIds();
 
         Map<String, Integer> modelWorkerNodeCounts = new HashMap<>();
-        boolean specifiedModelIds = modelIds != null && modelIds.length > 0;
-        String[] removedModelIds = specifiedModelIds ? modelIds : mlModelManager.getAllModelIds();
-        if (removedModelIds != null) {
-            for (String modelId : removedModelIds) {
+
+
+        if (modelIds != null) {
```
zane-neo, line 212: This change doesn't seem correct, shouldn't we add security check here?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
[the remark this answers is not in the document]
zane-neo: If the modelIds not passed, shouldn't we throw exception instead of removing all models in the cluster?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -0,0 +1,170 @@
...
+                .getClusterSettings()
+                .addSettingsUpdateConsumer(ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN, it -> allowCustomDeploymentPlan = it);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(ML_COMMONS_VALIDATE_BACKEND_ROLES, it -> filterByEnabled = it);
+
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLUndeployModelNodesResponse> listener) {
+        MLUndeployModelsRequest undeployModelsRequest = MLUndeployModelsRequest.fromActionRequest(request);
+        String[] modelIds = undeployModelsRequest.getModelIds();
```
zane-neo, line 108: Check if customer deployment plan enabled configuration to see if current action can be executed.

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
rbhavna: yes changing the code to accommodate all these scenarios

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
rbhavna: actually even if the model group is public, we should not let other users change back-end roles field. For that reason, I am refactoring the code to only allow owner/admin to update a model group

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
rbhavna: Is there any relation between names of model group and its models? Also for tags, I haven't got much time to work on this. Updating a tag would require updates to its models' tags as well. Is the tags field p0?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
rbhavna: The check is being done in TransportUndeployModelsAction class

plugin/src/main/java/org/opensearch/ml/utils/SecurityUtils.java
rbhavna: Makes sense. Will try to make those changes

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
rbhavna: Yeah there is actually some problem with undeploy here. It only works on single node cluster. Still working on this

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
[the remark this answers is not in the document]
rbhavna: User can specify a node ID instead which means to undeploy all the models on that node, in which case modelIDs wont be passed in the request

PR #871, opened 2023-04-27
Add a setting to enable/disable model url in register API.

jngz-es:

### Description
Add a setting to enable/disable model url in register API. The setting default value is false to not allow registering the model by a url.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
@@ -109,7 +109,7 @@ public class RestMLRegisterModelAction extends BaseRestHandler {
             : MLRegisterModelInput.parse(parser, modelName, version, loadModel);
         if (mlInput.getUrl() != null && !isModelUrlAllowed) {
             throw new IllegalArgumentException(
-                "To upload custom model user needs to enable allow_model_url settings. Otherwise please use opensearch pre-trained models."
+                "To upload custom model user needs to enable allow_registering_model_via_url settings. Otherwise please use opensearch pre-trained models."
             );
         }
         return new MLRegisterModelRequest(mlInput);
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
@@ -108,7 +108,7 @@ public class RestMLRegisterModelAction extends BaseRestHandler {
             ? MLRegisterModelInput.parse(parser, loadModel)
             : MLRegisterModelInput.parse(parser, modelName, version, loadModel);
         if (mlInput.getUrl() != null && !allowModelUrl) {
-            throw new IllegalArgumentException("Don't allow model url.");
+            throw new IllegalArgumentException("Don't allow uploading model via url.");
         }
         return new MLRegisterModelRequest(mlInput);
     }
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
@@ -7,6 +7,7 @@ package org.opensearch.ml.rest;
 
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.ML_BASE_URI;
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ALLOW_MODEL_URL;
 import static org.opensearch.ml.utils.RestActionUtils.PARAMETER_DEPLOY_MODEL;
 import static org.opensearch.ml.utils.RestActionUtils.PARAMETER_MODEL_ID;
 import static org.opensearch.ml.utils.RestActionUtils.PARAMETER_VERSION;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -87,4 +87,7 @@ private MLCommonsSettings() {}
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
+
+    public static final Setting<Boolean> ML_COMMONS_ALLOW_MODEL_URL = Setting
+        .boolSetting("plugins.ml_commons.allow_model_url", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
ylwu-amzn, line 92: Need tech writer's suggestion about the setting name. 

Hi Nate, can you help review if this setting name looks good or not?

This one will control whether user can upload their own model by setting url paramter in upload API. For security concern, we only allow user upload our pre-trained model https://opensearch.org/docs/latest/ml-commons-plugin/pretrained-models/, by specifying model name/version/format
POST /_plugins/_ml/models/_upload
{
  "name": "huggingface/sentence-transformers/all-MiniLM-L12-v2",
  "version": "1.0.1",
}
[12:18](https://amzn-aws.slack.com/archives/D035DCFAJQ6/p1682623081091489)
If user upload model by specifying url like this , our code will throw error
POST /_plugins/_ml/models/_upload
{
  "name": "sentence-transformers/all-MiniLM-L6-v2",
  "version": "1.0.0",
  "description": "test model",
  "model_config": {
    "model_type": "bert",
    "embedding_dimension": 384,
    "framework_type": "sentence_transformers"
  },
  "url": "https://github.com/opensearch-project/ml-commons/raw/2.x/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/all-MiniLM-L6-v2_torchscript_sentence-transformer.zip?raw=true"
}

ylwu-amzn:
@jngz-es DCO(developer certificate of Origin) check failed, seems you forgot to sign off.

plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
```diff
@@ -93,6 +107,9 @@ MLRegisterModelRequest getRequest(RestRequest request) throws IOException {
         MLRegisterModelInput mlInput = modelName == null
             ? MLRegisterModelInput.parse(parser, loadModel)
             : MLRegisterModelInput.parse(parser, modelName, version, loadModel);
+        if (mlInput.getUrl() != null && !allowModelUrl) {
+            throw new IllegalArgumentException("Don't allow model url.");
```
ylwu-amzn, line 111:
```suggestion
            throw new IllegalArgumentException("Don't allow uploading model via url.");
```

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
dhrubo-os: @Naarcha-AWS

ylwu-amzn: Pinged @Naarcha-AWS offline, suggest to change to `plugins.ml_commons.allow_registering_model_via_url`

plugin/src/main/java/org/opensearch/ml/rest/RestMLRegisterModelAction.java
```diff
@@ -107,8 +107,10 @@ MLRegisterModelRequest getRequest(RestRequest request) throws IOException {
         MLRegisterModelInput mlInput = modelName == null
             ? MLRegisterModelInput.parse(parser, loadModel)
             : MLRegisterModelInput.parse(parser, modelName, version, loadModel);
-        if (mlInput.getUrl() != null && !allowModelUrl) {
-            throw new IllegalArgumentException("Don't allow uploading model via url.");
+        if (mlInput.getUrl() != null && !isModelUrlAllowed) {
+            throw new IllegalArgumentException(
+                "To upload custom model user needs to enable allow_model_url settings. Otherwise please use opensearch pre-trained models."
```
dhrubo-os, line 112: I think according to nate's suggestion we need to change the settings name to: `allow_registering_model_via_url` ?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: Good question. @rbhavna , can you add a new setting to control enable/disable the local file uploading ? We can set it as disabled by default. Open source user can enable it if they need.

dhrubo-os: That will add another setting. I'm just wondering if adding more setting also confuse customers or not.

The purpose is to allow models from other source than opensearch traced models or not. In that case why not use the same settings? May be renaming the settings to: `allow_registering_custom_models`?

rbhavna: yeah makes sense. Using the same setting in both register model and upload model meta should be fine I guess.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -87,4 +87,8 @@ private MLCommonsSettings() {}
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
+
+    // This setting is to enable/disable model url in model register API.
+    public static final Setting<Boolean> ML_COMMONS_ALLOW_MODEL_URL = Setting
```
dhrubo-os, line 92: Should we change this variable name also as the settings name is also changed to: `allow_registering_model_via_url` ?

jngz-es: I think that's OK, it is different from the name of setting which doesn't impact any cx experience.

dhrubo-os (approved):

ylwu-amzn (approved):

PR #883, opened 2023-05-04
add pre-trained model list

jngz-es:
### Description
Add pre-trained model list in register api to restrict the model parameters input.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
@@ -171,8 +171,6 @@ public class ModelHelper {
 
                 return config;
             });
-        } catch (Exception e) {
-            throw e;
         } finally {
             deleteFileQuietly(mlEngine.getRegisterModelPath(taskId));
         }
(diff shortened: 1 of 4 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
```diff
@@ -137,6 +138,46 @@ public void downloadPrebuiltModelConfig(String taskId, MLRegisterModelInput regi
...
+
+                List<?> config = null;
+                try (JsonReader reader = new JsonReader(new FileReader(cacheFilePath))) {
+                    config = gson.fromJson(reader, List.class);
+                }
+
+                return config;
+            });
+        } catch (Exception e) {
+            throw e;
```
ylwu-amzn, line 175: If we just throw the exception directly, seems no need to add `catch (Exception e)`

jngz-es: Good point, removed it.

ylwu-amzn (approved):

PR #892, opened 2023-05-11
Bumping gradle from 7.4 to 8.1 to unblock build pipeline

### LINKED ISSUE #891: [BUG] ML-Commons Plugin build failure due to OpenSearch Core breaking changes
**What is the bug?**
ML-Commons Plugin 3.0 build fails due to [jackson update](https://github.com/opensearch-project/OpenSearch/commit/3168396208f849848f2cf0bf33a576a9c0e4d723) in OS core

```
./gradlew build

 FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'opensearch-ml'.
> java.util.concurrent.ExecutionException: org.gradle.api.GradleException: Failed to create Jar file /home/runner/.gradle/caches/jars-9/bc33e077f18aab4d8905d3690c9e67e4/jackson-core-2.15.0.jar.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

```

b4sjoo:
### Description
This PR update `gradle` from 7.4 to 8.1. Only `*.gradle` and `gradle-wrapper.properties` need to be adjusted, as well as a `lombok.config` under root directory you may want to create manually to deal with lombok config constructor deprecation. Other files are-auto generated.
 
### Issues Resolved
This is a draft PR to fix #891 
 
### Check List
  - [ ] All tests pass


lombok.config
```diff
@@ -1,3 +1,880 @@
-# This file is generated by the 'io.freefair.lombok' Gradle plugin
+##
```
ylwu-amzn, line 1: Is this file generated automatically? Seems we have much more content now

plugin/src/test/java/org/opensearch/ml/rest/RestMLCustomModelActionIT.java
```diff
@@ -30,6 +31,7 @@ public void setup() {
         registerModelInput = createRegisterModelInput();
     }
 
+    @Ignore
```
dhrubo-os, line 34: Should we add a `todo` if we plan to remove this at some point?

dhrubo-os:
Not for this PR, but in general may be we should enable`Dependabot` in this repo so that we can get PR's for the version bump in the long run? @ylwu-amzn  what do you think?

ylwu-amzn (approved):
LGTM. Thanks for fixing this!

ylwu-amzn:
> Not for this PR, but in general may be we should enable`Dependabot` in this repo so that we can get PR's for the version bump in the long run? @ylwu-amzn what do you think?

You mean bump OpenSearch version like this one [This Pr not bumping the OpenSearch version](https://github.com/opensearch-project/ml-commons/pull/866)? 

This PR is to bump gradle version

dhrubo-os:
> > Not for this PR, but in general may be we should enable`Dependabot` in this repo so that we can get PR's for the version bump in the long run? @ylwu-amzn what do you think?
> 
> You mean bump OpenSearch version like this one [This Pr not bumping the OpenSearch version](https://github.com/opensearch-project/ml-commons/pull/866)?
> 
> This PR is to bump gradle version

No, I mean something like [this](https://github.com/opensearch-project/opensearch-py-ml/pull/162) which automatically bump version of any dependent libraries.

lombok.config
b4sjoo: This one is not. Nowl `ombok` deprecated auto-generate config. This is the config template that they provide.

dhrubo-os (approved):

plugin/src/test/java/org/opensearch/ml/rest/RestMLCustomModelActionIT.java
b4sjoo: I think that's definitely something we need to add. There are several things we still need to fix, including this, `BWC`, as well as the `buildDeb` and `buildRpm` tasks in `gradle.build`. But we might merge first and add them later since this block others build pipeline for too long

ylwu-amzn:
> 



> > > Not for this PR, but in general may be we should enable`Dependabot` in this repo so that we can get PR's for the version bump in the long run? @ylwu-amzn what do you think?
> > 
> > 
> > You mean bump OpenSearch version like this one [This Pr not bumping the OpenSearch version](https://github.com/opensearch-project/ml-commons/pull/866)?
> > This PR is to bump gradle version
> 
> No, I mean something like [this](https://github.com/opensearch-project/opensearch-py-ml/pull/162) which automatically bump version of any dependent libraries.

We can do some research and have a try. I can see some potential issue like dependency conflicts between core and ML-commons, or neural-search and ml-commons, if we always bump to latest version.

ylwu-amzn:
Some other concern for auto bumping dependency version: the latest version may have some breaking changes, that may bring more effort to fix these breaking changes.

dhrubo-os:
I see, that makes sense.

PR #905, opened 2023-05-23
Model access control dev1

zane-neo:
### Description
Add model group & model version access control and code refactor
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/ModelAccessIdentifier.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.opensearch.ml.common;
+
+import lombok.Getter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum ModelAccessIdentifier {
```
ylwu-amzn, line 15: Minor: How about change the name to "ModelAccessMode"?

ylwu-amzn: I see the sample test request has both `model_access_identifier` and `backend_roles`
```
{
    "name": "model_group_restricted",
    "description": "This is an example description",
    "model_access_identifier": "restricted",
    "backend_roles": [
        "role-2"
    ],
    "tags": {
        "key1": "value1",
        "key2": "value2"
    }
}
```
If backend role is not empty, that means the model is "restricted", right? I think we can simplify the input by removing the `model_access_identifier` for this use case.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -100,6 +100,6 @@ private MLCommonsSettings() {}
             Setting.Property.Dynamic
         );
 
-    public static final Setting<Boolean> ML_COMMONS_VALIDATE_BACKEND_ROLES = Setting
-        .boolSetting("plugins.ml_commons.filter_by_backend_roles", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
+    public static final Setting<Boolean> ML_COMMONS_MODEL_ACCESS_CONTROL_ENABLED = Setting
+        .boolSetting("plugins.ml_commons.model_access_control_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
ylwu-amzn, line 104: We should discuss this change, other plugins are using filter_by_backend_roles

PR #909, opened 2023-05-24
Fix no corresponding backend roles user can update model issue

zane-neo:
### Description
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/model_group/TransportRegisterModelGroupAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/model_group/TransportRegisterModelGroupAction.java
@@ -30,7 +30,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.ml.common.MLModelGroup;
 import org.opensearch.ml.common.MLModelGroup.MLModelGroupBuilder;
 import org.opensearch.ml.common.MLTaskState;
-import org.opensearch.ml.common.ModelAccessIdentifier;
+import org.opensearch.ml.common.ModelAccessMode;
 import org.opensearch.ml.common.transport.model_group.MLRegisterModelGroupAction;
 import org.opensearch.ml.common.transport.model_group.MLRegisterModelGroupInput;
 import org.opensearch.ml.common.transport.model_group.MLRegisterModelGroupRequest;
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -154,15 +154,18 @@ private void updateModelGroup(
     }
 
     private void validateRequestForAccessControl(MLUpdateModelGroupInput input, User user, MLModelGroup mlModelGroup) {
-        if (hasAccessControlChange(input) && !modelAccessControlHelper.isOwner(mlModelGroup.getOwner(), user)) {
-            throw new IllegalArgumentException("Only owner has valid privilege to perform update access control data");
+        if (hasAccessControlChange(input)) {
+            if (!modelAccessControlHelper.isOwner(mlModelGroup.getOwner(), user)) {
```
ylwu-amzn, line 158: Should we also consider admin?

ylwu-amzn (dismissed):

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportRegisterModelGroupAction.java
```diff
@@ -153,14 +153,19 @@ public void createModelGroup(MLRegisterModelGroupInput input, ActionListener<Str
...
     private void validateRequestForAccessControl(MLRegisterModelGroupInput input, User user) {
-        ModelAccessIdentifier modelAccessIdentifier = input.getModelAccessIdentifier();
+        ModelAccessMode modelAccessMode = input.getModelAccessMode();
         Boolean isAddAllBackendRoles = input.getIsAddAllBackendRoles();
-        if (modelAccessIdentifier == null) {
-            throw new IllegalArgumentException("Invalid model access identifier");
-        } else if ((ModelAccessIdentifier.PUBLIC == modelAccessIdentifier || ModelAccessIdentifier.PRIVATE == modelAccessIdentifier)
+        if (modelAccessMode == null) {
+            if (!Boolean.TRUE.equals(isAddAllBackendRoles) && CollectionUtils.isEmpty(input.getBackendRoles())) {
+                throw new IllegalArgumentException("User must specify at least one backend role or make the model public/private");
```
rbhavna, line 160: If we do not specify any access fields, the default should be private actually, not throw exception

zane-neo: I think let user aware of this need more time, explicitly specify the model mode to private would be easy to understand and less confuse to user, let's make this explicit, if there's callout from user, we can change the behavior.

rbhavna: yeah should be fine then

rbhavna (approved):

PR #945, opened 2023-05-31
remote inference connector and virtual model APIs

Zhangxunmt:
### Description
Remote Inference Connector New APIs, including CRUD.
Examples:
~~~
GET /_plugins/_ml/connectors/277XeIgBsNBQ1UoSal6t
DELETE /_plugins/_ml/connectors/277XeIgBsNBQ1UoSal6t
GET /_plugins/_ml/connectors/_search
{
  "query": {
    "match_all": {}
  },
  "size": 1000
}

{ connector Template }
~~~
Virtual Model deploy and predict APIs to integrate with the detached Connectors.
 
 
### Check List
  - [ ] All tests pass


common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -172,4 +180,55 @@ public class CommonValue {
...
+            + DetachedConnector.CONNECTOR_NAME_FIELD
+            + "\" : {\"type\":\"text\",\"fields\":{\"keyword\":{\"type\":\"keyword\",\"ignore_above\":256}}},\n"
+            + "      \""
+            + DetachedConnector.CONNECTOR_VERSION_FIELD
+            + "\" : {\"type\": \"keyword\"},\n"
+            + "      \""
+            + DetachedConnector.CONNECTOR_DESCRIPTION_FIELD
+            + "\" : {\"type\": \"text\"},\n"
+            + "      \""
+            + DetachedConnector.CONNECTOR_PROTOCOL_FIELD
```
ylwu-amzn, line 199: I think protocol should be keyword, like `http` right?

common/src/main/java/org/opensearch/ml/common/connector/template/APISchema.java
```diff
@@ -0,0 +1,159 @@
...
+import java.security.PrivilegedExceptionAction;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.utils.StringUtils.gson;
+
+@Data
+public class APISchema implements ToXContentObject, NamedWriteable {
+    public static final String PARSE_FIELD_NAME = "API_Schema";
```
ylwu-amzn, line 30: Let's follow the same naming pattern, change to lower case and connect with "_", 
```suggestion
    public static final String PARSE_FIELD_NAME = "api_schema";
```

common/src/main/java/org/opensearch/ml/common/CommonValue.java
Zhangxunmt: Yes, updated already.

common/src/main/java/org/opensearch/ml/common/connector/template/APISchema.java
Zhangxunmt: Yes will do in the refresh.

common/src/main/java/org/opensearch/ml/common/connector/template/ConnectorState.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.connector.template;
+
+public enum ConnectorState {
```
ylwu-amzn, line 8: Why do we need to track the connector state as we already have model state ?

Zhangxunmt: Maybe we can reserve this connector state for future "metadata" or "healthcheck" APIs? Once the connector is not healthy, we can mark it as unhealthy? Let's keep it for a while. If we figure that it's not useful at all, we can always delete it.

PR #1068, opened 2023-07-11
fix cannot specify model access control parameters error

ylwu-amzn:
### Description
See such error when register pretrained model without model group id,
```
{
	"error": {
		"root_cause": [
			{
				"reason": "You cannot specify model access control parameters because the Security plugin or model access control is disabled on your cluster."
			}
		],
		"reason": "You cannot specify model access control parameters because the Security plugin or model access control is disabled on your cluster."
	},
	"status": 400
}
```

The reason is `input.getBackendRoles()` is not null, but it's empty.
This PR fixed the bug.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
@@ -193,7 +193,9 @@ public class MLModelGroupManager {
     }
 
     private void validateSecurityDisabledOrModelAccessControlDisabled(MLRegisterModelGroupInput input) {
-        if (input.getModelAccessMode() != null || input.getIsAddAllBackendRoles() != null || input.getBackendRoles() != null) {
+        if (input.getModelAccessMode() != null
+            || input.getIsAddAllBackendRoles() != null
+            || !CollectionUtils.isEmpty(input.getBackendRoles())) {
             throw new IllegalArgumentException(
                 "You cannot specify model access control parameters because the Security plugin or model access control is disabled on your cluster."
             );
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
```diff
@@ -193,7 +193,9 @@ public void validateUniqueModelGroupName(String name, ActionListener<Boolean> li
     }
 
     private void validateSecurityDisabledOrModelAccessControlDisabled(MLRegisterModelGroupInput input) {
-        if (input.getModelAccessMode() != null || input.getIsAddAllBackendRoles() != null || input.getBackendRoles() != null) {
+        if (input.getModelAccessMode() != null
+            || input.getIsAddAllBackendRoles() != null
+            || !CollectionUtils.isEmpty(input.getBackendRoles())) {
```
zane-neo, line 198: Seems strange, we better not allow `backend_roles` field show up in user's request instead of passing a empty list, can you change to this behavior?

ylwu-amzn: Thanks, think the same. I submit a new commit which add unit test for model group and fix some bug

zane-neo: I think we can change this back to `input.getBackEndRoles() != null`, right?

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
ylwu-amzn: I think keep as `CollectionUtils.isEmpty` is safer. Do you see any bug if we keep `CollectionUtils.isEmpty`?

zane-neo: I mean we better make the field not showing up when access control is not enabled, user can pass the `backend_roles` field with empty list when using `CollectionUtils.isEmpty`, I prefer to make the field not shown.

zane-neo (approved):

Issue #1179, opened 2023-08-02
[BUG] Orphaned ML Models 


dtaivpp:
**What is the bug?**
When a model upload fails there is an issue where the model still exists but it cannot be removed/unloaded. 

**How can one reproduce the bug?**
1. Upload a model on (what I guess is an unsupported platform?) M1 macbook. 
```
POST /_plugins/_ml/models/_upload 
{
  "name": "TEST_MODEL",
  "version": "1.0.0",
  "description": "MiniLM",
  "model_format": "TORCH_SCRIPT",
  "model_config": {
    "model_type": "bert",
    "embedding_dimension": 384,
    "framework_type": "sentence_transformers"
  },
  "url": "https://github.com/opensearch-project/ml-commons/raw/2.x/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/all-MiniLM-L6-v2_torchscript_sentence-transformer.zip?raw=true"
}
```

This creates a Task ID that can be tracked. Viewing the task id it has clearly failed. 
<img width="546" alt="Screenshot 2023-08-02 at 4 31 27 PM" src="https://github.com/opensearch-project/ml-commons/assets/17506770/6a6f4543-df82-47cc-9d6f-9b108444eaf2">

The above outputs indicates the TaskID we had been given is actually the ModelID. Attempting to `_unload` the model with what I believe is really the TaskID  yields the following: 
<img width="608" alt="Screenshot 2023-08-02 at 4 36 02 PM" src="https://github.com/opensearch-project/ml-commons/assets/17506770/bf461df7-85b9-4076-b42a-fa471af08695">

We cannot load a new model with that name however as it is still in the system: 
<img width="990" alt="Screenshot 2023-08-02 at 4 33 01 PM" src="https://github.com/opensearch-project/ml-commons/assets/17506770/d3e37733-d4f2-4464-9167-237f4b86fc9e">

**What is the expected behavior?**
If a model creation fails I expect to either be able to delete the failed upload or to upload a new model with the same name without issues. 

**What is your host/environment?**
 - OS: MacOS (M1)
 - Version 2.9

dhrubo-os:
There are few confusions. I'm going to clarify all those one by one.

1. `Model content changed` issue: 
[...]

PR #1187, opened 2023-08-07
Fix breaking change caused by opensearch core

zane-neo:
### Description
Fix breaking changes caused by opensearch core change.
 
### Issues Resolved
NA
 
### Check List
  - [ ] All tests pass


common/src/main/java/org/opensearch/ml/common/connector/Connector.java
```diff
@@ -93,7 +93,7 @@ static Connector fromStream(StreamInput in) throws IOException {
     }
 
     static Connector createConnector(XContentBuilder builder, String connectorProtocol) throws IOException {
-        String jsonStr = Strings.toString(builder);
+        String jsonStr = StringUtils.xContentBuilderToString(builder);
```
dhrubo-os, line 96: Can't we do : `builder.toString()`. Core also did the same: https://github.com/opensearch-project/OpenSearch/commit/d090a7f1e8be0df161f3d48f264bd0943766417e

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -82,4 +84,8 @@ public static Map<String, String> getParameterMap(Map<String, ?> parameterObjs)
         }
         return parameters;
     }
+
+    public static String xContentBuilderToString(XContentBuilder builder) {
```
dhrubo-os, line 88: I believe we don't need this method.

common/src/test/java/org/opensearch/ml/common/TestHelper.java
```diff
@@ -26,15 +26,15 @@ public static <T> void testParse(ToXContentObject obj, Function<XContentParser,
...
+        XContentBuilder builder = XContentFactory.jsonBuilder();
         if (wrapWithObject) {
             builder.startObject();
         }
         obj.toXContent(builder, ToXContent.EMPTY_PARAMS);
         if (wrapWithObject) {
             builder.endObject();
         }
-        String jsonStr = Strings.toString(builder);
+        String jsonStr = StringUtils.xContentBuilderToString(builder);
```
dhrubo-os, line 37: same: `builder.toString()`

dhrubo-os (commented):
spotless is failing.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
saratvemulapalli: XContentBuilder already supports `toString`.

common/src/test/java/org/opensearch/ml/common/TestHelper.java
```diff
@@ -26,15 +26,15 @@ public static <T> void testParse(ToXContentObject obj, Function<XContentParser,
     }
 
     public static <T> void testParse(ToXContentObject obj, Function<XContentParser, T> function, boolean wrapWithObject) throws IOException {
-        XContentBuilder builder = XContentFactory.contentBuilder(XContentType.JSON);
+        XContentBuilder builder = XContentFactory.jsonBuilder();
```
saratvemulapalli, line 29:
```suggestion
        XContentBuilder builder = MediaTypeRegistry.contentBuilder(XContentType.JSON);
```

saratvemulapalli:
I didnt realize there are multiple PRs already trying to fix the build, I was chasing a bug #1179 and ended up with compile problems. 
Feel free to use changes from this PR: https://github.com/opensearch-project/ml-commons/pull/1189

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
saratvemulapalli: Yup thats the right way to do it. 
The static method was refactored and moved to `XContentBuilder`

common/src/test/java/org/opensearch/ml/common/TestHelper.java
dhrubo-os: Just curious to know why we should prefer `MediaTypeRegistry.contentBuilder(XContentType.JSON);` over `XContentFactory.jsonBuilder();`? as both are doing the same.

```
    public static XContentBuilder jsonBuilder() throws IOException {
        return MediaTypeRegistry.contentBuilder(XContentType.JSON);
    }
```

saratvemulapalli: In this case it doesnt matter since we are using `XContentType.JSON` anyway. But incase we'd like to use a different `MediaType` the plugin is abstracted out about the format OpenSearch engine uses. So I would prefer `MediaType` where ever possible.

saratvemulapalli: With this change: https://github.com/opensearch-project/OpenSearch/pull/9156 it would make `XContentType.JSON` generic.

zane-neo: I prefer to use the previous one because this is more like a syntactic sugar, and in ml-commons we support only json format for now, we can use `MediaTypeRegistry` in the future if we need to support different media type.

nknize: > I prefer to use the previous one because this is more like a syntactic sugar,

I think this is fine for now. I was leaning toward eliminating `XContentFactory` altogether so as to not tempt a developer to use it in `opensearch-core` (and create the circular library dependency problem) but we already enforce no circular dependencies with the `opensearch-core` library. What we don't enforce are circular dependencies between other libraries. 

The intent is for `opensearch-core` to be the "center of gravity". For example, downstream plugin/extensions takes a required dependency on core and optional dependencies on the other libraries (e.g., x-content, geo, telementry, aggregations, mappers, replication, storage). This keeps the downstream opensearch dependencies lean. Why do we care? Because if someone wants to run an OpenSearch as a function (e.g., lambda, or cloud function) package size limits affect bootstrap times, class loading, and overall customer cost. Developers can afford to be more memory wreckless if the project only runs on self hosted, but load times and package sizes matter in cloud native deployments. This is why I'm so adamant on breaking down the monolith and not building a bloated comprehensive `opensearch-sdk`.

dhrubo-os:
> I didnt realize there are multiple PRs already trying to fix the build, I was chasing a bug https://github.com/opensearch-project/ml-commons/issues/1179 and ended up with compile problems.
Feel free to use changes from this PR: https://github.com/opensearch-project/ml-commons/pull/1189

@saratvemulapalli this is not actually a bug. I will reply today to the issue in details today.

dhrubo-os:
```
ava.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([5287C72090A89EF1:B9B8FAEEEA734B7B]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.opensearch.ml.bwc.MLCommonsBackwardsCompatibilityIT.testBackwardsCompatibility(MLCommonsBackwardsCompatibilityIT.java:223)
  2> NOTE: leaving temporary files on disk at: /home/runner/work/ml-commons/ml-commons/plugin/build/testrun/mlCommonsBwcCluster#fullRestartClusterTask
```

@zane-neo BWC test is failing. Are we planning to fix this?

ylwu-amzn (approved):
LGTM. Thanks for fixing that breaking change. For BWC failure, that seems a different issue, can fix in a separate PR

model-collapse (approved):

PR #1194, opened 2023-08-09
fixing metrics

dhrubo-os:
### Description
[fixing metrics]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -318,9 +318,10 @@ public class MLModelManager {
      * @param mlTask      ML task
      */
     public void registerMLModel(MLRegisterModelInput registerModelInput, MLTask mlTask) {
-        mlStats.getStat(MLNodeLevelStat.ML_NODE_TOTAL_REQUEST_COUNT).increment();
+
         checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);
         try {
+            mlStats.getStat(MLNodeLevelStat.ML_NODE_TOTAL_REQUEST_COUNT).increment();
             mlStats.getStat(MLNodeLevelStat.ML_NODE_EXECUTING_TASK_COUNT).increment();
             mlStats.createCounterStatIfAbsent(mlTask.getFunctionName(), REGISTER, ML_ACTION_REQUEST_COUNT).increment();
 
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -836,6 +837,7 @@ public void deployModel(
 
     private void handleDeployModelException(String modelId, FunctionName functionName, ActionListener<String> listener, Exception e) {
         mlStats.createCounterStatIfAbsent(functionName, ActionName.DEPLOY, MLActionLevelStat.ML_ACTION_FAILURE_COUNT).increment();
+        mlStats.getStat(MLNodeLevelStat.ML_NODE_TOTAL_FAILURE_COUNT).increment();
```
ylwu-amzn, line 840: We should only track the real internal server failure. For example, some exception are expected (which means that's not a code bug) like illegal input parameter. You can refer to other place

dhrubo-os: Agree, addressed.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -689,7 +689,10 @@ private void deleteModel(String modelId) {
     }
 
     private void handleException(FunctionName functionName, String taskId, Exception e) {
-        mlStats.createCounterStatIfAbsent(functionName, REGISTER, MLActionLevelStat.ML_ACTION_FAILURE_COUNT).increment();
+        if (!(e instanceof MLLimitExceededException) && !(e instanceof MLResourceNotFoundException)) {
```
ylwu-amzn, line 692: Thanks , how about also check `IllegalArgumentException` which is generally user's input is wrong, we should not treat it as some internal failure.

dhrubo-os: Sounds good. In that case, may be we can add same exception [here](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java#L44-L54) also, what do you think?

ylwu-amzn: Sure, sounds good

ylwu-amzn (approved):

rbhavna (approved):

PR #1197, opened 2023-08-10
add eligible node role settings

ylwu-amzn:
### Description
Per PM's suggestion, remote model should be able to run on data or ML node by default. But we should also keep local model run on ML node by default. So we still keep `plugins.ml_commons.only_run_on_ml_node` as true, but only control local model. Add more setting to provide granular level control on remote and local model.

1. Add `plugins.ml_commons.task_dispatcher.eligible_node_role.remote_model` which controls only remote model can run on which node. Default value is `["data", "ml"]` which means the remote model can run on data node and ML node by default.
2. Add `plugins.ml_commons.task_dispatcher.eligible_node_role.local_model` which controls only remote model can run on which node. But we have `plugins.ml_commons.only_run_on_ml_node` which controls the model can only run on ML node or not. This is to provide BWC, For local model, 1/ if `plugins.ml_commons.only_run_on_ml_node` is `true`, we will always run it on ML node. 2/ if `plugins.ml_commons.only_run_on_ml_node` is `false`, will run model on nodes defined in `plugins.ml_commons.task_dispatcher.eligible_node_role.local_model`.  
 
 

 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
+++ b/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
@@ -128,6 +128,22 @@ public final class MLCommonsSettings {
             Setting.Property.Dynamic
         );
 
+    /**
+     * Per PM's suggestion, remote model should be able to run on data or ML node by default.
+     * But we should also keep local model run on ML node by default. So we still keep
+     * "plugins.ml_commons.only_run_on_ml_node" as true, but only control local model.
+     * Add more setting to provide granular level control on remote and local model.
+     *
+     * 1. Add "plugins.ml_commons.task_dispatcher.eligible_node_role.remote_model" which controls
+     * only remote model can run on which node. Default value is ["data", "ml"] which means the
+     * remote model can run on data node and ML node by default.
+     * 2. Add "plugins.ml_commons.task_dispatcher.eligible_node_role.local_model" which controls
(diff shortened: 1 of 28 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -127,4 +127,22 @@ private MLCommonsSettings() {}
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
+
+    public static final Setting<List<String>> ML_COMMONS_REMOTE_MODEL_ELIGIBLE_NODE_ROLES = Setting
```
dhrubo-os, line 131: Can we add some comments for these settings (summary of what you have in the PR description) so that it can't be confusing later? 

Will we deprecate `ML_COMMONS_ONLY_RUN_ON_ML_NODE` in the long run?

dhrubo-os (approved):
Just have a minor comment. Overall looks good to me.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: Sure, will add some comment, deprecating that depends on if we have dedicated ML node.

rbhavna (approved):

PR #1208, opened 2023-08-16
add feature flag for remote inference

Zhangxunmt:
### Description
Add a new class for feature flags in ml-commons. It has remote inference feature flag for now which has setting to block APIs for connector creation, model registration, model deploy and model predictions.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
@@ -11,6 +11,7 @@ import static org.opensearch.ml.common.MLTaskState.FAILED;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.DEPLOY_THREAD_POOL;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN;
 import static org.opensearch.ml.task.MLTaskManager.TASK_SEMAPHORE_TIMEOUT;
+import static org.opensearch.ml.utils.MLExceptionUtils.REMOTE_INFERENCE_DISABLED_ERR_MSG;
 
 import java.time.Instant;
 import java.util.ArrayList;
@@ -52,6 +53,7 @@ import org.opensearch.ml.common.transport.deploy.MLDeployModelResponse;
 import org.opensearch.ml.engine.ModelHelper;
 import org.opensearch.ml.helper.ModelAccessControlHelper;
 import org.opensearch.ml.model.MLModelManager;
+import org.opensearch.ml.settings.MLFeatureEnabledSetting;
 import org.opensearch.ml.stats.MLNodeLevelStat;
 import org.opensearch.ml.stats.MLStats;
 import org.opensearch.ml.task.MLTaskDispatcher;
@@ -83,6 +85,7 @@ public class TransportDeployModelAction extends HandledTransportAction<ActionReq
 
     private volatile boolean allowCustomDeploymentPlan;
     private ModelAccessControlHelper modelAccessControlHelper;
+    private MLFeatureEnabledSetting mlFeatureEnabledSetting;
 
     @Inject
     public TransportDeployModelAction(
@@ -99,7 +102,8 @@ public class TransportDeployModelAction extends HandledTransportAction<ActionReq
         MLModelManager mlModelManager,
         MLStats mlStats,
         Settings settings,
-        ModelAccessControlHelper modelAccessControlHelper
+        ModelAccessControlHelper modelAccessControlHelper,
+        MLFeatureEnabledSetting mlFeatureEnabledSetting
     ) {
         super(MLDeployModelAction.NAME, transportService, actionFilters, MLDeployModelRequest::new);
         this.transportService = transportService;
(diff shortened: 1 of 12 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -108,6 +108,9 @@ private MLCommonsSettings() {}
             Setting.Property.Dynamic
         );
 
+    public static final Setting<Boolean> ML_COMMONS_REMOTE_INFERENCE_ENABLED = Setting
```
ylwu-amzn, line 111: I see AD plugin also has setting to disable the whole plugin https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/settings/ADEnabledSetting.java#L35C5-L35C81

Can we also add that for disabling whole plugin?

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -117,6 +122,9 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
      */
     @VisibleForTesting
     MLPredictionTaskRequest getRequest(String modelId, String algorithm, RestRequest request) throws IOException {
+        if (algorithm.equals("REMOTE") && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
```
dhrubo-os, line 125: [nit] let's not use magic string here.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
Zhangxunmt: Let's get in this feature flag first given the tight timeline. The plugin level disabling can be added in a different PR later.

dhrubo-os: Let's add details comments about the settings we are adding and the coverage like (to block APIs for connector creation, model registration, model deploy and model predictions)

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public MLFeatureEnabledSetting(ClusterService clusterService, Settings settings) {
+        isRemoteInferenceEnabled = ML_COMMONS_REMOTE_INFERENCE_ENABLED.get(settings);
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_REMOTE_INFERENCE_ENABLED, it -> isRemoteInferenceEnabled = it);
+    }
+
+    /**
+     * Whether the remote inference feature is enabled.  If disabled, time series plugin rejects RESTful requests for this feature.
```
dhrubo-os, line 27: `If disabled, time series plugin rejects RESTful requests for this feature.`  --> Fix this please

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
```diff
@@ -18,6 +18,8 @@
 public class MLExceptionUtils {
 
     public static final String NOT_SERIALIZABLE_EXCEPTION_WRAPPER = "NotSerializableExceptionWrapper: ";
+    public static final String REMOTE_INFERENCE_DISABLED_ERR_MSG =
```
dhrubo-os, line 21: How about : `Remote Inference is currently disabled. To enable it, update the setting "plugins.ml_commons.remote_inference_enabled" to true.`

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
ylwu-amzn: Sure, can you create a github issue to track this?

How about we follow the same naming pattern like AD, for example `"plugins.anomaly_detection.hcad_cold_start_interpolation.enabled"` , the pattern is `plugins.<plugin_name>.<feature_name>.enabled`

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -114,6 +118,7 @@ public TransportDeployModelAction(
         this.mlModelManager = mlModelManager;
         this.mlStats = mlStats;
         this.modelAccessControlHelper = modelAccessControlHelper;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
```
ylwu-amzn, line 121: Need to add setting update consumer. Check line 125 of this class

ylwu-amzn: Sorry, after reading the whole code changes, I know why you write like this way. I guess you are trying to avoid some duplicate code by creating a new `MLFeatureEnabledSetting` class.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
ylwu-amzn:
```suggestion
        if (FunctionName.REMOTE.name().equals(algorithm) && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
```

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
Zhangxunmt: Yes you are right. More new feature flags can be added into this class in the future releases.

PR #1209, opened 2023-08-16
add feature flags for remote inference

Zhangxunmt:
### Description
Add Remote Inference feature flag. It adds a check up for the feature flag setting before invoking "Create Connector", "Register Model", "Deploy Model" and "Predict Model" APIs. 
 
Tested this PR using a local cluster.

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
+++ b/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
@@ -152,6 +152,7 @@ import org.opensearch.ml.rest.RestMLUndeployModelAction;
 import org.opensearch.ml.rest.RestMLUpdateModelGroupAction;
 import org.opensearch.ml.rest.RestMLUploadModelChunkAction;
 import org.opensearch.ml.settings.MLCommonsSettings;
+import org.opensearch.ml.settings.MLFeatureEnabledSetting;
 import org.opensearch.ml.stats.MLClusterLevelStat;
 import org.opensearch.ml.stats.MLNodeLevelStat;
 import org.opensearch.ml.stats.MLStat;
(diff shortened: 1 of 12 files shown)


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -470,7 +476,11 @@ public List<RestHandler> getRestHandlers(
...
         RestMLSearchTaskAction restMLSearchTaskAction = new RestMLSearchTaskAction();
         RestMLProfileAction restMLProfileAction = new RestMLProfileAction(clusterService);
-        RestMLRegisterModelAction restMLRegisterModelAction = new RestMLRegisterModelAction(clusterService, settings);
+        RestMLRegisterModelAction restMLRegisterModelAction = new RestMLRegisterModelAction(
+            clusterService,
+            settings,
+            mlFeatureEnabledSetting
+        );
         RestMLDeployModelAction restMLDeployModelAction = new RestMLDeployModelAction();
         RestMLUndeployModelAction restMLUndeployModelAction = new RestMLUndeployModelAction(clusterService, settings);
```
dhrubo-os, line 485: Just realized, don't we need to check mlFeatureEnabledSetting in undeploy action also?

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
ylwu-amzn: I think undeploy should be safe. If we disable remote inference feature, and user still have some remote model running, we should still allow them to undeploy.

ylwu-amzn (approved):

PR #1215, opened 2023-08-18
updating request field for the connectors

dhrubo-os:
### Description
[updating request field for the connectors]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelOnNodeAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelOnNodeAction.java
@@ -24,6 +24,7 @@ import org.opensearch.cluster.node.DiscoveryNodes;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ml.breaker.MLCircuitBreakerService;
 import org.opensearch.ml.common.FunctionName;
@@ -161,13 +162,15 @@ public class TransportDeployModelOnNodeAction extends
                     .build();
                 MLForwardRequest deployModelDoneMessage = new MLForwardRequest(mlForwardInput);
 
-                transportService
-                    .sendRequest(
-                        getNodeById(coordinatingNodeId),
-                        MLForwardAction.NAME,
-                        deployModelDoneMessage,
-                        new ActionListenerResponseHandler<>(taskDoneListener, MLForwardResponse::new)
-                    );
+                try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                    transportService
+                        .sendRequest(
+                            getNodeById(coordinatingNodeId),
+                            MLForwardAction.NAME,
+                            deployModelDoneMessage,
+                            new ActionListenerResponseHandler<>(taskDoneListener, MLForwardResponse::new)
+                        );
+                }
             }, e -> {
                 MLForwardInput mlForwardInput = MLForwardInput
                     .builder()
@@ -179,13 +182,15 @@ public class TransportDeployModelOnNodeAction extends
                     .build();
                 MLForwardRequest deployModelDoneMessage = new MLForwardRequest(mlForwardInput);
 
-                transportService
-                    .sendRequest(
-                        getNodeById(coordinatingNodeId),
-                        MLForwardAction.NAME,
-                        deployModelDoneMessage,
-                        new ActionListenerResponseHandler<>(taskDoneListener, MLForwardResponse::new)
-                    );
+                try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                    transportService
+                        .sendRequest(
+                            getNodeById(coordinatingNodeId),
+                            MLForwardAction.NAME,
+                            deployModelDoneMessage,
+                            new ActionListenerResponseHandler<>(taskDoneListener, MLForwardResponse::new)
+                        );
+                }
             })
         );
 


common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -264,6 +265,64 @@ public  <T> T createPredictPayload(Map<String, String> parameters) {
...
+                result = result.replace("${parameters." + paramName + "}", modifiedValue);
+            }
+        }
+
+        result = StringEscapeUtils.unescapeJava(result); // Unescape Unicode escape sequences
+
+        return result;
+    }
+
+    protected boolean isNotPlainString(String value) {
```
ylwu-amzn, line 292: If user input `{"key1": "123"}`, we will detect `"123"` as not plain String?

ylwu-amzn: From line 282 
`isNotPlainString(paramValue) ? paramValue : "\"" + paramValue + "\"";`

It will return `123`  rather than `"123"` which not what user expected.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -264,6 +265,64 @@ public  <T> T createPredictPayload(Map<String, String> parameters) {
         return (T) parameters.get("http_body");
     }
 
+    protected String applyCustomSubstitution(Map<String, String> parameters, String inputPayload) {
+
+        if (parameters == null) {
+            throw new IllegalArgumentException("Some parameter placeholder not filled in payload: input");
```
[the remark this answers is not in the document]
dhrubo-os: Yeah, that's a good point. We don't need to throw exception here. I was mostly trying to reflect this [test case](https://github.com/opensearch-project/ml-commons/blob/2.9/common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java#L168-L173)

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
dhrubo-os: 1. So if user provides a number with `""` (`"123"`), although it's a number we still should send value as `"123"`?
2. if user provide a number without `""` (123), in that case I suppose we want to send the value as `123`?
3. What about boolean values? Same idea?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -254,8 +256,7 @@ public  <T> T createPredictPayload(Map<String, String> parameters) {
         if (predictAction.isPresent() && predictAction.get().getRequestBody() != null) {
             String payload = predictAction.get().getRequestBody();
             payload = fillNullParameters(parameters, payload);
-            StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
-            payload = substitutor.replace(payload);
+            payload = applyCustomSubstitution(parameters, payload);
             if (!isJson(payload)) {
```
austintlee, line 260: In my local testing, I found this can return false on "valid" json.  The issue was that parameter values needed to be escaped.  I can make a separate PR for that, but thought Id mention it here since there's related code in this PR for escaping parameter values.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -264,6 +265,64 @@ public  <T> T createPredictPayload(Map<String, String> parameters) {
...
+        while (matcher.find()) {
+            String paramName = matcher.group(1);
+            String paramValue = parameters.get(paramName);
+            if (paramValue != null) {
+                String modifiedValue = isNotPlainString(paramValue) ? paramValue : "\"" + paramValue + "\"";
+                result = result.replace("${parameters." + paramName + "}", modifiedValue);
+            }
+        }
+
+        result = StringEscapeUtils.unescapeJava(result); // Unescape Unicode escape sequences
```
austintlee, line 287: Here, I am not sure if this is sufficient to avoid valid values causing `isJson()` to fail as I mentioned above.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -264,6 +265,64 @@ public  <T> T createPredictPayload(Map<String, String> parameters) {
...
     }
 
+    protected String applyCustomSubstitution(Map<String, String> parameters, String inputPayload) {
+
+        if (parameters == null) {
+            throw new IllegalArgumentException("Some parameter placeholder not filled in payload: input");
+        }
+        String result = inputPayload;
+
+        Pattern pattern = Pattern.compile("\\$\\{parameters\\.(\\w+)\\}");
```
austintlee, line 275: Maybe pull this out as a class static?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
dhrubo-os: In my testing I saw, at this point payload is supposed to replaced with the actual value:

like: `"{ \"texts\": ${parameters.texts}, \"truncate\": ${parameters.truncate} }"`  --> `"{ \"texts\": [\"hello world\"], \"truncate\": \"END\" }"` which is a valid JSON.

Could you please share your use case where you get false for a "valid" JSON?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
dhrubo-os: Sure, I can do that.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
dhrubo-os: Could you please explain this little bit more for me?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
austintlee: I am basically taking a bunch of text from my OpenSearch cluster (as search results/hits) and passing it as part of the `messages` parameter to OpenAI (search hit -> message).  I don't know which character in my text caused the issue, but to get around it, I applied a patch locally to always do `escapeJson` on the messages and got around the issue.  So, what I am trying to say is that if the text is not properly escaped for Json, you can get IllegalArgumentException.  I guess this is not a comment specific to your PR.  I am not pointing out an issue with the custom substitution logic per se.  Let me know if you want me to open a new issue to track this.

ylwu-amzn: @austintlee I think you are using 2.x branch. We have some bug fix in 2.9 branch, we just backported them to 2.x branch. Please rebase your code and retest.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
ylwu-amzn: For example "{ \"texts\": ${parameters.texts}, \"truncate\": ${parameters.truncate} }" , if user input "truncate": "123" in parameters when predict, will we pass \"truncate\": 123" to model with your PR?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
austintlee: Yaliang answered my question above.

PR #1225, opened 2023-08-21
Neural search default process support

zane-neo:
### Description
ml-commons has two default pre/post process function which are for OpenAI and Cohere and written in painless script. There's no default pre/post process function for neural search plugin, and if user want to use neural search with remote model, user has to write complex painless script which is a heavy burden. This PR do two things: 
1. Add default pre/post process function for neural search plugin and user only need to define a simple string like: `connector.pre_process.neural_search.text_embedding` or `connector.post_process.neural_search.text_embedding` to use neural search plugin easily.
2. Change default painless script to java code to make them more readable to both coder and user.
 
### Issues Resolved
NA
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -19,7 +19,7 @@ dependencies {
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '4.4.0'
 
     compileOnly group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
-    compileOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     compileOnly group: 'org.json', name: 'json', version: '20230227'
 }
 
--- a/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
@@ -18,6 +18,7 @@ import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -44,9 +45,12 @@ public class StringUtils {
         return utf8EncodedString;
     }
 
-    public static Map<String, Object> fromJson(String jsonStr, String defaultKey) {
+    public static Map<String, Object> fromJson(String input, String defaultKey) {
+        if (!isJson(input)) {
+            return Collections.singletonMap(defaultKey, input);
+        }
         Map<String, Object> result;
-        JsonElement jsonElement = JsonParser.parseString(jsonStr);
+        JsonElement jsonElement = JsonParser.parseString(input);
         if (jsonElement.isJsonObject()) {
             result = GsonUtil.fromJson(jsonElement, Map.class);
         } else if (jsonElement.isJsonArray()) {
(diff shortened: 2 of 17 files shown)


common/build.gradle
```diff
@@ -19,7 +19,7 @@ dependencies {
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '4.4.0'
 
     compileOnly group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
-    compileOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
```
ylwu-amzn, line 22: Why have to change this part? Don't suggest change common package if it doesn't break your code.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -48,13 +45,16 @@ public static String toUTF8(String rawString) {
         return utf8EncodedString;
     }
 
-    public static Map<String, Object> fromJson(String jsonStr, String defaultKey) {
+    public static Map<String, Object> fromJson(String input, String defaultKey) {
+        if (!isJson(input)) {
+            return Collections.singletonMap(defaultKey, input);
```
ylwu-amzn, line 50: By adding this if branch, this function extends to support more cases which is not Map or List. That's changing the original design. Will it cause any issue? Why we have to do this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -54,91 +58,112 @@ public static RemoteInferenceInputDataSet processInput(MLInput mlInput, Connecto
...
                 } else {
-                    newParameters.put(entry.getKey(), escapeJson(entry.getValue()));
+                    newParameters.put(key, escapeJson(value));
                 }
             });
             inputData.setParameters(newParameters);
         }
         return inputData;
     }
+    private static RemoteInferenceInputDataSet processTextDocsInput(TextDocsInputDataSet inputDataSet, Connector connector, Map<String, String> parameters, ScriptService scriptService) {
```
ylwu-amzn, line 83: minor: add empty line before this function

zane-neo:
> > Add default pre/post process function for neural search plugin and user only need to define a simple string like: connector.pre_process.neural_search.text_embedding or connector.post_process.neural_search.text_embedding to use neural search plugin easily.
> 
> We have default pre/process function for neural search. These are for neural-search https://github.com/opensearch-project/ml-commons/blob/2.9/common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java#L14-L15

The functions above are specific for OpenAI and Cohere, for user own deployed models(user can define the model output data structure), we're pre-defining the data structure and its process function. E.g. If user follow the pre-defined output data structure: List<List<Float>>, note that this data structure is different with OpenAI and Cohere response, then the default function `connector.post_process.neural_search.text_embedding` can be used to post process. If we don't provide this pre-defined data structure and process function, user needs to create complex painless script to process output. So here we're adding something new to make it easy to use for users.

common/build.gradle
zane-neo: Tested with neural-search plugin, if don't change this, there'll be `java. lang. NoClassDefFoundError: com/google/gson/Gson`, changing this can fix this issue.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
zane-neo: The original design is only accepting json format response? If so, it's ok to remove this.

zane-neo:
Closing this PR since a new PR to main branch has created: https://github.com/opensearch-project/ml-commons/pull/1274

PR #1227, opened 2023-08-21
update connector API

Zhangxunmt:
### Description
Update connector API to partially update a connector in the ml-connector index. Verified it works with updating existing fields, adding new fields and deleting fields.
 
Example:
POST /_plugins/_ml/connectors/_update/4s0vGooBTRq0xHburjWW
{
  "credential": {
        "openAI_key": "new key"
    }
}

GET /_plugins/_ml/connectors/4s0vGooBTRq0xHburjWW
{
  ...
  "credential": {
        "openAI_key": "new key"
    },
  ...
}

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.connector;
+
+import lombok.Builder;
+import lombok.Getter;
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.core.common.io.stream.InputStreamStreamInput;
+import org.opensearch.core.common.io.stream.OutputStreamStreamOutput;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.XContentParser;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Map;
+
+import static org.opensearch.action.ValidateActions.addValidationError;
+
+@Getter
+public class MLUpdateConnectorRequest extends ActionRequest {
+    String connectorId;
+    Map<String, Object> updateContent;
+
+    @Builder
+    public MLUpdateConnectorRequest(String connectorId, Map<String, Object> updateContent) {
+        this.connectorId = connectorId;
+        this.updateContent = updateContent;
+    }
+
+    public MLUpdateConnectorRequest(StreamInput in) throws IOException {
+        super(in);
+        this.connectorId = in.readString();
+        this.updateContent = in.readMap();
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(this.connectorId);
+        out.writeMap(this.getUpdateContent());
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        ActionRequestValidationException exception = null;
+
+        if (this.connectorId == null) {
+            exception = addValidationError("ML connector id can't be null", exception);
+        }
+
+        return exception;
+    }
+
+    public static MLUpdateConnectorRequest parse(XContentParser parser, String connectorId) throws IOException {
+        Map<String, Object> dataAsMap = null;
+        dataAsMap = parser.map();
+
+        return MLUpdateConnectorRequest.builder().connectorId(connectorId).updateContent(dataAsMap).build();
+    }
+
+    public static MLUpdateConnectorRequest fromActionRequest(ActionRequest actionRequest) {
+        if (actionRequest instanceof MLUpdateConnectorRequest) {
+            return (MLUpdateConnectorRequest) actionRequest;
+        }
+
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+             OutputStreamStreamOutput osso = new OutputStreamStreamOutput(baos)) {
+            actionRequest.writeTo(osso);
+            try (StreamInput input = new InputStreamStreamInput(new ByteArrayInputStream(baos.toByteArray()))) {
+                return new MLUpdateConnectorRequest(input);
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException("failed to parse ActionRequest into MLUpdateConnectorRequest", e);
+        }
+    }
+}
--- a/plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.action.connector;
+
+import static org.opensearch.ml.common.CommonValue.ML_CONNECTOR_INDEX;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.DocWriteResponse;
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.action.update.UpdateRequest;
+import org.opensearch.action.update.UpdateResponse;
+import org.opensearch.client.Client;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.ml.common.transport.connector.MLUpdateConnectorAction;
+import org.opensearch.ml.common.transport.connector.MLUpdateConnectorRequest;
+import org.opensearch.ml.helper.ConnectorAccessControlHelper;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportService;
+
+import lombok.AccessLevel;
+import lombok.experimental.FieldDefaults;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+public class UpdateConnectorTransportAction extends HandledTransportAction<ActionRequest, UpdateResponse> {
+    Client client;
+    NamedXContentRegistry xContentRegistry;
+
+    ConnectorAccessControlHelper connectorAccessControlHelper;
+
+    @Inject
+    public UpdateConnectorTransportAction(
+        TransportService transportService,
+        ActionFilters actionFilters,
+        Client client,
+        NamedXContentRegistry xContentRegistry,
+        ConnectorAccessControlHelper connectorAccessControlHelper
+    ) {
+        super(MLUpdateConnectorAction.NAME, transportService, actionFilters, MLUpdateConnectorRequest::new);
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+        this.connectorAccessControlHelper = connectorAccessControlHelper;
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<UpdateResponse> listener) {
+        MLUpdateConnectorRequest mlUpdateConnectorAction = MLUpdateConnectorRequest.fromActionRequest(request);
+        String connectorId = mlUpdateConnectorAction.getConnectorId();
+        UpdateRequest updateRequest = new UpdateRequest(ML_CONNECTOR_INDEX, connectorId);
+        updateRequest.doc(mlUpdateConnectorAction.getUpdateContent());
+        updateRequest.docAsUpsert(true);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            connectorAccessControlHelper.validateConnectorAccess(client, connectorId, ActionListener.wrap(hasPermission -> {
+                if (Boolean.TRUE.equals(hasPermission)) {
+                    client.update(updateRequest, getUpdateResponseListener(connectorId, listener, context));
+                } else {
+                    listener
+                        .onFailure(
+                            new IllegalArgumentException("You don't have permission to update the connector, connector id: " + connectorId)
+                        );
+                }
+            }, exception -> {
+                log.error("You don't have permission to update the connector for connector id: " + connectorId, exception);
+                listener.onFailure(exception);
+            }));
+        } catch (Exception e) {
+            log.error("Failed to update ML connector " + connectorId, e);
+            listener.onFailure(e);
+        }
+    }
+
+    private ActionListener<UpdateResponse> getUpdateResponseListener(
+        String connectorId,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
+                log.info("Connector id:{} failed update", connectorId);
+                actionListener.onResponse(updateResponse);
+                return;
+            }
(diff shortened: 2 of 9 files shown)


ylwu-amzn:
Github CI failed

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
+                log.info("Connector id:{} failed update", connectorId);
+                actionListener.onResponse(updateResponse);
+                return;
+            }
+            log.info("Completed Update Connector Request, connector id:{} updated", connectorId);
```
dhrubo-os, line 92: How about: `log.info("Successfully updated the connector with ID: {}", connectorId)`

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private ActionListener<UpdateResponse> getUpdateResponseListener(
+        String connectorId,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
+                log.info("Connector id:{} failed update", connectorId);
```
dhrubo-os, line 88: log.info("Failed to update the connector with ID: {}", connectorId) ?

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+                if (Boolean.TRUE.equals(hasPermission)) {
+                    client.update(updateRequest, getUpdateResponseListener(connectorId, listener, context));
+                } else {
+                    listener
+                        .onFailure(
+                            new IllegalArgumentException("You don't have permission to update the connector, connector id: " + connectorId)
+                        );
+                }
+            }, exception -> {
+                log.error("You don't have permission to update the connector for connector id: " + connectorId, exception);
```
dhrubo-os, line 72: How about?

log.error("Permission denied: Unable to update the connector with ID {}. Details: {}", connectorId, exception)

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+        updateRequest.docAsUpsert(true);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            connectorAccessControlHelper.validateConnectorAccess(client, connectorId, ActionListener.wrap(hasPermission -> {
+                if (Boolean.TRUE.equals(hasPermission)) {
+                    client.update(updateRequest, getUpdateResponseListener(connectorId, listener, context));
+                } else {
+                    listener
+                        .onFailure(
+                            new IllegalArgumentException("You don't have permission to update the connector, connector id: " + connectorId)
```
dhrubo-os, line 68: log.error("Permission denied: Unable to update the connector with ID {}. Details: {}", connectorId, exception)

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,99 @@
...
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
+                log.info("Connector id:{} failed update", connectorId);
+                actionListener.onResponse(updateResponse);
+                return;
+            }
+            log.info("Completed Update Connector Request, connector id:{} updated", connectorId);
+            actionListener.onResponse(updateResponse);
+        }, exception -> {
+            log.error("Failed to update ML connector: " + connectorId, exception);
```
dhrubo-os, line 95: log.error("Failed to update ML connector with ID {}. Details: {}", connectorId, exception)

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateConnectorAction.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @VisibleForTesting
+    private MLUpdateConnectorRequest getRequest(RestRequest request) throws IOException {
+        if (!mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
+            throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
+        }
+
+        if (!request.hasContent()) {
+            throw new IOException("Update Connector request has empty body");
```
dhrubo-os, line 69: throw new IOException("Failed to update connector: Request body is empty");

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateConnectorAction.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        String connectorId = getParameterId(request, PARAMETER_CONNECTOR_ID);
+        if (Strings.isBlank(connectorId)) {
+            throw new IOException("Update Connector request has no connector Id");
```
dhrubo-os, line 74: throw new IOException("Failed to update connector: Connector ID is missing");

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateConnectorAction.java
Zhangxunmt: Updated all messages based on comments.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
Zhangxunmt: updated

Zhangxunmt:
Updated messages. The CI will pass.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Map;
+
+import static org.opensearch.action.ValidateActions.addValidationError;
+
+@Getter
+public class MLUpdateConnectorRequest extends ActionRequest {
+    String connectorId;
+    Map<String, Object> updateContent;
```
ylwu-amzn, line 29: How about using `Connector newConnector` ?

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
```diff
@@ -0,0 +1,83 @@
...
+        if (this.connectorId == null) {
+            exception = addValidationError("ML connector id can't be null", exception);
+        }
+
+        return exception;
+    }
+
+    public static MLUpdateConnectorRequest parse(XContentParser parser, String connectorId) throws IOException {
+        Map<String, Object> dataAsMap = null;
+        dataAsMap = parser.map();
```
ylwu-amzn, line 63: Do we accept any key/value for the new connector content? Should we do some validation ? For example checking if it's valid fiel in connector.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,95 @@
...
+        MLUpdateConnectorRequest mlUpdateConnectorAction = MLUpdateConnectorRequest.fromActionRequest(request);
+        String connectorId = mlUpdateConnectorAction.getConnectorId();
+        UpdateRequest updateRequest = new UpdateRequest(ML_CONNECTOR_INDEX, connectorId);
+        updateRequest.doc(mlUpdateConnectorAction.getUpdateContent());
+        updateRequest.docAsUpsert(true);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            connectorAccessControlHelper.validateConnectorAccess(client, connectorId, ActionListener.wrap(hasPermission -> {
+                if (Boolean.TRUE.equals(hasPermission)) {
+                    client.update(updateRequest, getUpdateResponseListener(connectorId, listener, context));
```
ylwu-amzn, line 60: What if the connector is being used by some model? After updating, the connector content in index is not consistent with the content in memory.  Should we only allow user update connector when the model undeployed?

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
Zhangxunmt: Yes it supports add/update/delete any field in the OS document. If the value is null, the update request will remove the key. If the key is a new field, it will adds to the connector.

Zhangxunmt: agree, we should put the same connector validation here during the update.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
Zhangxunmt: Yes this is a nice validation before updating.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
Zhangxunmt: Define a new connector would require the user to redefine the whole connector body. We should make it easier just like how the Update API works in the OS Document. User only needs to put the updated/new/deleted field in the request like rotating the credential key, and we will update the connector in the index accordingly.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
ylwu-amzn: > If the key is a new field, it will adds to the connector.

This seems risky, user can hack by adding many new random fields

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
ylwu-amzn: I think we can use same way of current create connector https://github.com/opensearch-project/ml-commons/blob/2.x/plugin/src/main/java/org/opensearch/ml/action/connector/TransportCreateConnectorAction.java#L96

Zhangxunmt: Can I just do the following to update using the current create connector? or I have to reenter the whole connector template?

POST /_plugins/_ml/connectors/_update/xxxx
{
   "credential": {
        "openAI_key": "new_key"
    }
}

b4sjoo (dismissed):

b4sjoo (approved):

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -74,6 +87,44 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
...
+        ActionListener<UpdateResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        SearchRequest searchRequest = new SearchRequest(ML_MODEL_INDEX);
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder();
+        BoolQueryBuilder boolQueryBuilder = QueryBuilders.boolQuery();
+        boolQueryBuilder.must(QueryBuilders.matchQuery(MLModel.CONNECTOR_ID_FIELD, connectorId));
+        boolQueryBuilder.must(QueryBuilders.idsQuery().addIds(mlModelManager.getAllModelIds()));
+        sourceBuilder.query(boolQueryBuilder);
+        searchRequest.source(sourceBuilder);
```
ylwu-amzn, line 102: By default , search request will only return 10 results, can you set size as maximum value 10000?

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+    private void updateUndeployedConnector(
+        String connectorId,
+        UpdateRequest updateRequest,
+        ActionListener<UpdateResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        SearchRequest searchRequest = new SearchRequest(ML_MODEL_INDEX);
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder();
+        BoolQueryBuilder boolQueryBuilder = QueryBuilders.boolQuery();
+        boolQueryBuilder.must(QueryBuilders.matchQuery(MLModel.CONNECTOR_ID_FIELD, connectorId));
```
ylwu-amzn, line 99: Should we also check model status ?

Zhangxunmt: It's already included in this search. The line 100 makes sure that the search results matches the model_ids that are in the cache.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
Zhangxunmt: The security concerns will be reviewed and addressed in the Pen tests.

dhrubo-os: Let's add `TODO` in that case.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
dhrubo-os: Sorry for the confusion, I didn't clearly understand why line 100 is relevant here? 

Why can't we just search in the model index: where CONNECTOR_ID_FIELD = connectorId and model status = deployed or partially deployed?

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+        boolQueryBuilder.must(QueryBuilders.idsQuery().addIds(mlModelManager.getAllModelIds()));
+        sourceBuilder.query(boolQueryBuilder);
+        searchRequest.source(sourceBuilder);
+
+        client.search(searchRequest, ActionListener.wrap(searchResponse -> {
+            SearchHit[] searchHits = searchResponse.getHits().getHits();
+            if (searchHits.length == 0) {
+                client.update(updateRequest, getUpdateResponseListener(connectorId, listener, context));
+            } else {
+                log.error(searchHits.length + " models are still using this connector, please undeploy the models first!");
```
dhrubo-os, line 109: I feel like best user experience will be, if we show the deployed model ids also to the customer. Otherwise with this message user again needs to query to find out the deployed models.

Customer might not have any idea which models are deployed now with this connector idea. If we show the model ids, that could be helpful for customer.

Zhangxunmt: line 100 will use all the deployed and deploying models to search ML-Model index that match "connector_id=<updated connector id>". The model id shouldn't be included in the message otherwise app sec reviewer will raise concerns about the customer info leakage in the logs.

dhrubo-os: not sure how a random model id can concern appsec which is not quite a customer info. 
Anyway, I was mostly referring to the `new MLValidationException(` part where we can send these model ids.

The goal is to make this easy for customers. That shouldn't concern appsec at least.

dhrubo-os: > `all the deployed and deploying models`

does this get a inclusive list of all the deployed and deploying models in `all the nodes`?

PR #1274, opened 2023-09-01
Add neural search default processor for non OpenAI/Cohere scenario

zane-neo:
### Description
ml-commons has two default pre/post process function which are for OpenAI and Cohere and written in painless script. There's no default pre/post process function for neural search plugin text embedding case, if user want to use neural search with remote model to text embedding, user has to write complex painless script which is a heavy burden. This PR do two things:

1. Add default pre/post process function for neural search plugin text embedding and once user defines their model output to List<List<Float>> as suggested, then in the connector pre/post process can be defined to strings like: `connector.pre_process.neural_search.text_embedding` or `connector.post_process.neural_search.text_embedding`, which is pretty easy for user to use neural search text embedding.
2. Change default OpenAI & Cohere painless script to java code to make them more readable to both coder and user.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1304
 
### Check List
  - [ ] All tests pass


jngz-es:
I have a high level comment. From my understanding, the processors of connector is only about the format transformation between ml-commons and llms like OpenAI, Cohere etc. About the format transform part, I don't think neural search is a dependency for ml-commons, the only dependency should be on what llm is using. In other words, what format transforming looks like is only dependent on the llm we are using. We may have default pre/post processors, but not specific to neural search, it should be generic.

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -5,61 +5,64 @@
...
-                "        if (i < embeddings.length - 1) {\n" +
-                "          builder.append(\",\");\n" +
-                "        }\n" +
-                "      }\n" +
-                "      builder.append(\"]\");\n" +
-                "      \n" +
-                "      return builder.toString();\n    ");
+        JSON_PATH_EXPRESSION.put(OPENAI_EMBEDDING, "$.data[*].embedding");
+        JSON_PATH_EXPRESSION.put(COHERE_EMBEDDING, "$.embeddings");
+        JSON_PATH_EXPRESSION.put(NEURAL_SEARCH_EMBEDDING, "$[*]");
```
jngz-es, line 32: Not sure for what model we expect the result is List<List<Float>>? How does the customer use this model?

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -6,44 +6,37 @@
...
-
-        PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT, "\n    StringBuilder builder = new StringBuilder();\n" +
-                        "    builder.append(\"\\\"\");\n" +
-                        "    builder.append(params.text_docs[0]);\n" +
-                        "    builder.append(\"\\\"\");\n" +
-                        "    def parameters = \"{\" +\"\\\"input\\\":\" + builder + \"}\";\n" +
-                        "    return  \"{\" +\"\\\"parameters\\\":\" + parameters + \"}\";");
+        PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT, cohereTextEmbeddingPreProcess());
+        PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT, openAiTextEmbeddingPreProcess());
+        PRE_PROCESS_FUNCTIONS.put(NEURAL_SEARCH_EMBEDDING_INPUT, openAiTextEmbeddingPreProcess());
```
jngz-es, line 32: Do we only need a default pre-processor, not neural search?

zane-neo:
> I have a high level comment. From my understanding, the processors of connector is only about the format transformation between ml-commons and llms like OpenAI, Cohere etc. About the format transform part, I don't think neural search is a dependency for ml-commons, the only dependency should be on what llm is using. In other words, what format transforming looks like is only dependent on the llm we are using. We may have default pre/post processors, but not specific to neural search, it should be generic.

The processor could be for non llms, e.g. our pre-trained model. Neural search is not a dependency for ml-commons, here the neural search might not a good name since it sounds more generic, but the main purpose here is to give user a default approach to transform the inputs from neural search plugin, e.g. TextEmbeddingProcessor to the expected format that user deployed pre-trained model in thirdparty platform, e.g. SageMaker.

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
zane-neo: For text embedding model, including OpenAI/Cohere embedding and our pre-trained text embedding models.

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
zane-neo: This is for neural search plugin text embedding use case, we can fine-tune the name though.

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -6,44 +6,37 @@
...
+import java.util.function.Function;
 
 public class MLPreProcessFunction {
 
-    private static Map<String, String> PRE_PROCESS_FUNCTIONS;
+    private static final Map<String, Function<List<String>, Map<String, Object>>> PRE_PROCESS_FUNCTIONS = new HashMap<>();
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
 
+    public static final String NEURAL_SEARCH_EMBEDDING_INPUT = "connector.pre_process.neural_search.text_embedding";
```
dhrubo-os, line 19: `connector.pre_process.default`.

May be we can add more comment to describe how are we making this default pre-processing function. But I assume people can easily assume if no other pre-processing works, it should fall back to `default` one.

jngz-es (approved):

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -5,61 +5,64 @@
...
 import java.util.Map;
+import java.util.function.Function;
 
 public class MLPostProcessFunction {
 
-    private static Map<String, String> POST_PROCESS_FUNCTIONS;
     public static final String COHERE_EMBEDDING = "connector.post_process.cohere.embedding";
     public static final String OPENAI_EMBEDDING = "connector.post_process.openai.embedding";
 
+    public static final String NEURAL_SEARCH_EMBEDDING = "connector.post_process.default";
```
dhrubo-os, line 22: IS this default post process function will only be for embedding? If yes, then may be we can say: `connector.post_process.default.embedding`?

and the static variable name can be: `DEFAULT_EMBEDDING` (to match other two?)

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -5,61 +5,64 @@
...
-                "                 \"\\\"data_type\\\":\\\"\" + dataType + \"\\\",\" +\n" +
-                "                 \"\\\"shape\\\":\" + shape + \",\" +\n" +
-                "                 \"\\\"data\\\":\" + params.data[0].embedding +\n" +
-                "                 \"}\";\n" +
-                "      return json;\n    ");
+    public static Function<List<List<Float>>, List<ModelTensor>> buildModelTensorList() {
+        return numbersList -> {
+            List<ModelTensor> modelTensors = new ArrayList<>();
+            if (numbersList == null) {
+                throw new IllegalArgumentException("NumbersList is null when applying build-in post process function!");
```
dhrubo-os, line 42: what do we mean by numberlist? Is that the list of embeddings? `numbers` is bit confusing here.

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -6,44 +6,37 @@
...
+import java.util.function.Function;
 
 public class MLPreProcessFunction {
 
-    private static Map<String, String> PRE_PROCESS_FUNCTIONS;
+    private static final Map<String, Function<List<String>, Map<String, Object>>> PRE_PROCESS_FUNCTIONS = new HashMap<>();
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
 
+    public static final String NEURAL_SEARCH_EMBEDDING_INPUT = "connector.pre_process.default";
```
dhrubo-os, line 19: same comment here (like post_process) about renaming for pre_process default.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -48,13 +43,13 @@ public static String toUTF8(String rawString) {
         return utf8EncodedString;
     }
 
-    public static Map<String, Object> fromJson(String jsonStr, String defaultKey) {
+    public static Map<String, Object> fromJson(String input, String defaultKey) {
```
dhrubo-os, line 46: any reason to change the variable name? may be we can rename to: `inputJsonStr`?

common/src/main/java/org/opensearch/ml/common/utils/GsonUtil.java
```diff
@@ -0,0 +1,30 @@
+package org.opensearch.ml.common.utils;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.stream.JsonReader;
+
+public class GsonUtil {
```
dhrubo-os, line 7: Not quite sure what are we exactly achieving with extending this class? May be we can add more comments to describe importance of this class?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -54,91 +58,112 @@ public static RemoteInferenceInputDataSet processInput(MLInput mlInput, Connecto
...
         return inputData;
     }
+    private static RemoteInferenceInputDataSet processTextDocsInput(TextDocsInputDataSet inputDataSet, Connector connector, Map<String, String> parameters, ScriptService scriptService) {
+        List<String> docs = new ArrayList<>(inputDataSet.getDocs());
+        Optional<ConnectorAction> predictAction = connector.findPredictAction();
+        if (predictAction.isEmpty()) {
+            throw new IllegalArgumentException("no predict action found");
+        }
+        String preProcessFunction = predictAction.get().getPreProcessFunction();
+        if (preProcessFunction == null) {
```
dhrubo-os, line 90: I assume end user doesn't need to provide this `preProcessFunction`? By default it will be attached to the predictAction?

dhrubo-os:
Are we planning to merge this PR in main?

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
zane-neo: Changed the name to embeddings.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
zane-neo: I was going to support non json format response, so changed the parameter name for better representation. Then found we don't have such case after discussion, removed the code in function body but missed this parameter name, changed back to jsonStr.

common/src/main/java/org/opensearch/ml/common/utils/GsonUtil.java
zane-neo: I think this is common pattern to create a util class instead of `new` the target class everywhere, the beneficial are:
1. Less memory usage since only have one Gson instance in the GsonUtil.
2. Better encapsulation and better code organization since all Gson related code are in one place.
3. Cleaner code: no need to create Gson instance everywhere.

dhrubo-os: I'm not quite sure about the beneficial pointed out here:

1. We are already doing the same [here](https://github.com/opensearch-project/ml-commons/blob/2.x/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java#L27-L29). I don't think we improved anything specially here. 
2. In all the `fromJson` methods we are just invoking the `fromJson` method from `Gson`. So I don't think we are achieving better code organization here rather than adding extra code. 
3. Same as 1

jngz-es (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: Good finding, this can be attached to the predictAction by default.

common/src/main/java/org/opensearch/ml/common/utils/GsonUtil.java
zane-neo: I missed this one, but I find another case [here](https://github.com/opensearch-project/ml-commons/blob/ff0d73460ee03e7f550c9383529a6627f4ac7ed3/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java#L55). I think if we all follow the principle to use the Gson in StringUtils, then this GsonUtil is no more needed, I'll change this back but let's pay attention to this in the future code review.

zane-neo:
> Are we planning to merge this PR in main branch first?

Yes, this will be merged to main branch and then backport to 2.x branch.

jngz-es (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -54,91 +58,110 @@ public static RemoteInferenceInputDataSet processInput(MLInput mlInput, Connecto
...
+            throw new IllegalArgumentException("no predict action found");
+        }
+        String preProcessFunction = predictAction.get().getPreProcessFunction();
+        preProcessFunction = preProcessFunction == null ? MLPreProcessFunction.DEFAULT_EMBEDDING_INPUT : preProcessFunction;
+        if (MLPreProcessFunction.contains(preProcessFunction)) {
+            Map<String, Object> buildInFunctionResult = MLPreProcessFunction.get(preProcessFunction).apply(docs);
+            return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(buildInFunctionResult)).build();
+        } else {
+            if (preProcessFunction.contains("${parameters")) {
+                StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
```
dhrubo-os, line 96: Looks like we are using this `${parameters.` in multiple places, what do you think about using a Static variable for this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ScriptUtils.java
```diff
@@ -5,52 +5,36 @@
 
 package org.opensearch.ml.engine.utils;
 
-import com.google.gson.Gson;
-import org.opensearch.ml.common.connector.MLPostProcessFunction;
-import org.opensearch.ml.common.connector.MLPreProcessFunction;
-import org.opensearch.ml.common.utils.StringUtils;
```
dhrubo-os, line 11: Not sure why do we have to remove it from here and then use that in the code?

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -6,44 +6,37 @@
...
+import java.util.function.Function;
 
 public class MLPreProcessFunction {
 
-    private static Map<String, String> PRE_PROCESS_FUNCTIONS;
+    private static final Map<String, Function<List<String>, Map<String, Object>>> PRE_PROCESS_FUNCTIONS = new HashMap<>();
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
 
+    public static final String DEFAULT_EMBEDDING_INPUT = "connector.pre_process.default.embedding";
```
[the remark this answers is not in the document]
zane-neo: done

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: this is existing code just moved to different place, beside this there are many other places using this string, we should make it a static variable but maybe in next PR

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ScriptUtils.java
zane-neo: Because the Pre/Post process function has nothing to do with painless script, they are java functions, ScriptUtils' single responsibility is to execute customized painless script.

dhrubo-os: I mean we removed import : `import org.opensearch.ml.common.utils.StringUtils;`

then used in the code like: `org.opensearch.ml.common.utils.StringUtils.fromJson(resultJson, "result");`

any advantage on using like this?

zane-neo: Oh, this one. I think it's because I was using apache common StringUtils in this class as well so make this a fully qualified name, but apache one was removed, so we can use import StringUtils now.

dhrubo-os (approved):

ylwu-amzn (approved):

PR #1283, opened 2023-09-05
feature/conversation backport to 2.x

### LINKED ISSUE #1150: [RFC] Conversations and Generative AI in OpenSearch
# Introduction

The recent advances in Large Language Models (LLMs) have enabled developers to utilize natural language in their applications with better quality and ability. As ChatGPT has shown, these LLMs strongly enable use cases involving summarization and conversation. However, when prompting LLMs to answer fact-based questions (applications we call “conversational search”), we find that there are significant shortcomings for enterprise-grade applications. 

First, the major LLMs are not trained on datasets that are not exposed to the internet, and therefore do not have the context to answer questions on private data. Most enterprise data falls into this category. Second, the way in which LLMs answer questions based on their training data gives rise to “hallucinations” and false answers, which are not acceptable in applications for mission critical use cases. 

End-users love the ability to converse using colloquial language with an application to get answers to questions or find interesting search results, but require up-to-date information and accuracy. A solution to this problem is through Retrieval Augmented Generation (RAG), where an application sends an LLM a superset of correct information in response to a prompt, and the LLM is used to summarize and extract information from this set (instead of probabilistically determining an answer). 

We believe OpenSearch could be a great platform for building conversational search applications, and aligns well with the RAG approach. It already offers semantic search capabilities using its vector database and k-NN plug-in, alongside enterprise-grade security and scalability. This is a great building block for the “source of truth” information retrieval component of RAG. However, it currently lacks the primitives and crisp APIs to easily enable the conversational element. 

Although there are libraries that allow for building this functionality at the application layer (e.g. LangChain), we believe the best developer experience would be to enable this directly in OpenSearch. We consider the “G” in a RAG pipeline as LLM-based post-processing to enable direct question answering, summarization, and a conversational experience on top of OpenSearch semantic search. This enables end-users to interact with their data in OpenSearch in new ways. Furthermore, we believe developers may want to use different LLMs, and that the choice of model should be pluggable.

Through using plugins and search pipelines, we propose an architecture in this RFC to expose easily consumable APIs for conversational search, history, and storage. We segment it into a few components, including: 1/search query rewriting using generative AI and conversational context, 2/question answering and summarization of OpenSearch semantic search queries using generative AI, and 3/a concept of “conversational memory” to easily store the state of conversations and add additional interactions. Conversational Memory will also support conversational applications that have multiple agents operating together, giving a single source of truth for conversation state.

## Goals

1/ Developers can easily build conversational search applications (e.g. knowledge-base search, informational chatbot, etc.) using OpenSearch and their choice of generative AI model using well-defined REST APIs. Some of these applications will be an ongoing conversation, while others will be one-shot (and the history of interactions is not important).

2/ Developers can use OpenSearch to support multi-agent conversational architectures, which require a single “source of truth” for conversational history. Multi-agent architectures will have other agents besides that for semantic search with OpenSearch (e.g. an agent that queries the public internet). These developers need an easy API to manage conversational history, both in adding interactions to conversations and exploring history of those conversations.

3/ Developers can easily obtain OpenSearch (semantic) search results alongside the generative AI question answering, so they can show the source documents and enable the end user to explore the source material.

## Non-Goals

1/ Building a general LLM application toolkit in OpenSearch. Our goal is just to enable conversational search and the related dependency of conversational memory.

2/ LLM hosting. LLMs take significant resources and should be operated outside of an OpenSearch cluster. We also hope to use the ML-Commons remote inference feature rather than implement our own connectors.

3/ A conversational search application platform. Our goal is to expose crisp APIs to make building applications that use conversational search easy, but not create the end application itself.

# Proposed Architecture

- An addition to the ml-commons plugin that provides a CRUD API to store and access conversation history (”memory”).
- An addition of a new Search Pipeline implementation to ml-commons that uses conversational memory and large language models for question answering.
- An addition to ml-commons that enables users to have conversations through a new Conversation API.

<image>

Aryn Conversation Plugins v2

### Conversational Memory API (Chat History)

Conversational memory is the storage for conversations, which are an ordered list of interactions. Conversational memory makes it easy to add new interactions to a conversation or explore previous interactions. For example, you would need conversational memory to write a chatbot, since it takes the previous interactions in a conversation as part of the context for generating a future response. At a high level, this mostly resembles a generic read/write store, and we will use an OpenSearch index for it. However, the interesting nuance is in the data itself, which we will describe next.

A conversation is represented as a list of interactions, ordered chronologically. Each conversation will also include some metadata, like the start time and the number of interactions.

The basic elements of an interaction are an input and a response, representing the human input to an AI agent and that agent’s response. We’ll also include any additional prompting that was used in the interaction, the agent that was used in this interaction, and possible arbitrary metadata that the agent may want to include. For example, a conversational search agent may include the actual search results as metadata for a user search query (which is an interaction).

Each `ConversationMetadata` and `Interaction` will have access controls linked to the specific user that creates them. Only Alice can add to and read from conversations that Alice owns. The main rationale for this is that Alice’s conversation will potentially include information from all documents Alice has access to, so her conversations’ access controls are maximally the intersection of Alice’s access rights. We plan to leverage OpenSearch’s existing access control mechanisms for this.

The plan is to maintain 2 indices - 1 for `ConversationMetadata` and 1 for `Interaction`.

```
structure ConversationMetadata {
    conversationId: ConversationId
    numInteractions: Integer
    createTime: Timestamp
    lastInteractionTime: Timestamp
    name: String
}

structure Interaction {
    conversationId: ConversationId
    interactionId: InteractionId
    input: String
    prompt: String
    response: String
    agent: String
    time: Timestamp
    attributes: InteractionAttributes
}
```
**API**

The operations for conversational memory are similar to the usual CRUD operations for a datastore. `CreateInteraction` will update the appropriate `ConversationMetadata` to have a correct `lastInteractionTime` and `numInteractions`

```
/// Creates a new conversation and returns its id
operation CreateConversation {
    input: CreateConversationInput
    output: CreateConversationOutput
}

@input
structure CreateConversationInput {
    name: String
}

@output
structure CreateConversationOutput {
    conversationId: ConversationId
}

/// Returns the list of all conversations
operation GetConversations {
    input: GetConversationsInput
    output: GetConversationsOutput
}

@input
structure GetConversationsInput {
    nextToken: String
    maxResults: Integer
}

@output 
structure GetConversationsOutput {
    conversations: List[ConversationMetadata]
    nextToken: String
}

/// Adds an interaction to a conversation and returns its id
operation CreateInteraction {
    input: CreateInteractionInput
    output: CreateInteractionOutput
}

@input 
structure CreateInteractionInput  {
    @required
    @httpLabel
    conversationId: ConversationId
    input: String
    prompt: String
    response: String
    agent: String
    attributes: InteractionAttributes
}

@output
structure CreateInteractionOutput {
    interactionId: InteractionId
}

/// Returns the list of interactions associated with a conversation
operation GetInteractions {
    input: GetInteractionsInput
    output: GetInteractionsOutput
}

@input
structure GetInteractionsInput {
    @required 
    @httpLabel
    conversationId: ConversationId
    nextToken: String
    maxResults: Integer
}

@output
structure GetInteractionsOutput {
    metadata: ConversationMetadata
    interactions: List[Interaction]
    nextToken: String
}

operation DeleteConversation {
    input: DeleteConversationInput
    output: DeleteConversationOutput
}

@input
structure DeleteConversationInput {
    @required
    @httpLabel
    conversationId: ConversationId
}

@output
structure DeleteConversationOutput {
    success: Boolean
}
```
We do not propose having an update API for conversation metadata, and we treat this as immutable. We believe that users would prefer to just create a new conversation than update parameters on an existing one.

### Search Pipeline extension

The conversational search path essentially consists of an OpenSearch query, with some pre- and post-processing. Search Pipelines, introduced in 2.8, are a tool for pre- and post-processing in the query path, so we have chosen to use that mechanism to implement conversational search.

We have chosen to implement the question answering component of RAG in the form of query result rewrites.  We are introducing a new response processor that sends the top search results, and optionally some previous conversation history to the LLM to generate a response in the conversation. We are also introducing a new response processor that iterates over search hits and interacts with an LLM to produce an answer for each result with a score.  Finally, we are introducing a request processor to rephrase the user’s query, taking into account the conversation history. We will rely on the remote inference feature proposed in https://github.com/opensearch-project/ml-commons/issues/882 for answer generation.

Based on different patterns we have seen with applications, we designed this API to support “one-off” and “multi-shot” conversations. Users can have “one-off” question answering interactions, where the prior context is not included, via a search pipeline that uses this new question answering processor. Users can also have “multi-shot” conversations where interactions are stored in conversational memory and are used as additional context that is sent to the model along with each search query. Users will need to use the Conversational Search plugin to create a conversation and pass the conversationId to the search pipeline in order to retain all the interactions associated with it.

In addition to the conversation ID, users can also pass a “prompt” parameter for any prompt engineering alongside their search query.

```
GET wiki-simple-paras/_search?search_pipeline=convo_qa_pipeline
{
  "_source": ["title", "text"],
  "query" : {
    "neural": {
      "text_vector": {
         "query_text": "When was Abraham Lincoln born?",
         "k": 10,
         "model_id": "<text-embedding-model-id>"
      }
    }
  },
  "ext": {
     "question_answering_parameters": {
         "question": "When was Abraham Lincoln born?"
     },
     "conversation" : {
         "id": "...",
         "prompt": "..."
		}
  }
}
```
The search pipeline includes pre and post processing steps. The pre-processing step uses generative AI to rewrite the search query submitted by the user, taking into account the conversation history if a conversation was specified. This allows things like antecedent replacement (”When was he born?” → “When was Abraham Lincoln born?”, if the prior question was “Who was Abraham Lincoln?”).

The post-processing step is a processor that takes the search results, optionally performs a lookup against the conversational memory, and then sends this data to the LLM configured by the user. We believe different users will want to use different LLMs, so this will be pluggable.

### Conversation API

The point of this API is to provide conversational search as a relatively simple endpoint, hooking pieces together such that the user can easily build an application with it. It takes a search query (or some other kind of human input), performs a search against OpenSearch, and then feeds those search results into an LLM and returns the answer. All of this work is done in the search pipeline underneath - so the API is just a wrapper - but we feel this kind of an API would be helpful to developers who just want an easy REST API.

We would like to return search results as well as the LLM response. This differs from most existing systems that return only answers, and it allows clients to perform validations or additional downstream processing.

```
/// Ask a question and get a GenAI response grounded in search results
operation Query {
    input: QueryInput
    output: QueryOutput
}

structure QueryInput {
    index: String
    conversationId: ConversationId
    query: String
    prompt: String
    filter: String
    numResults: Integer
}

structure QueryOutput {
    response: String
		rewrittenQuery: String
    searchResults: DocList
    interactionId: InteractionId
}

/// List of docs used to answer the question
list DocList {
    member: Document
}
```
# Discussion

1. **Performance:** LLM inference takes on the order of seconds; if you have sufficiently high traffic, that can increase to minutes or more as an LLM hosting service rate-limits or a hosted model becomes resource constrained. Five people using this at the same time could have the potential to completely stall each other out. We’ll try to be fault-tolerant as regards this, but a lot of the onus may fall on the users and the LLM hosters to work out how to get higher LLM throughput.
2. **Ordering:** Since LLM inference can take a while, a user might get impatient and ask a bunch of search queries before the first search query has returned an answer; and the answers might come back from the LLM out of order. We will write only complete interactions, meaning the order that messages come back from the LLM. The client should disallow multiple queries at once (in a conversation) to prevent this.
3. **Dependencies:** This relies on the relatively new search pipeline and remote inference features. Accordingly, this probably only works for OpenSearch ≥ 2.9, with the appropriate ML-Commons installation. We’re also hoping to get the pipelines themselves into Search-Processors; in which case that plugin also becomes a dependency. Lastly, the high-level Conversational API depends on the Conversational Pipeline, and they both depend on the Conversational Memory plugin, which we think should be its own plugin. We’ll put out some resources on building once we figure it out.

# Summary

In this RFC we gave a proposal for bringing conversational search into OpenSearch. Our proposal consists of three components: 1/ an API for conversational memory stored in OpenSearch, 2/ an OpenSearch search pipeline for Retrieval-Augmented Generation (RAG), and 3/ an API that provides a simple one-shot API for conversational search applications. We would appreciate any feedback, suggestions, and comments towards integrating this cleanly with the rest of the OpenSearch ecosystem and making it the best it can be.

Thanks!

# Requested Feedback

- Does this feature set cover the set of use cases for generative AI applications that you want to build? We have been focused on search applications and we’re interested in how much the community wants to go beyond exposing conversational search and conversational memory building blocks at this time.
- We believe the search pipeline is great mechanism to define a RAG pipelines, but we also felt that a conversational API that invokes this pipeline would be helpful for developers to more easily build conversational search applications. We’d love feedback on if we should add more to this API, or conversely if it’s even needed in providing an easy developer experience.
- This approach for RAG introduces a several cross plugin dependencies. There has been talk in the community about moving away from the plugin architecture for OpenSearch, and we want to make sure this approach is aligned with the higher-level architectural goals of the project. We’d appreciate feedback on this topic.

HenryL27:
### Description
Add Conversation RAG Pipeline and Memory APIs to ml-commons behind a feature flag
 
### Issues Resolved
#1150 
 
### Check List
  - [ ] All tests pass


HenryL27:
I messed up the DCO

PR #1288, opened 2023-09-06
fix commons-lang3 package version to fix jarhell issue

b4sjoo:
### Description
Follows https://github.com/opensearch-project/security-analytics/pull/535 to fix jarhell issue
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -56,7 +56,7 @@ dependencies {
     implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}")
     implementation group: 'com.google.guava', name: 'guava', version: '32.0.1-jre'
     implementation group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
-    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: "${versions.commonslang}"
     implementation group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.19.0"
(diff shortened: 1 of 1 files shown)


Zhangxunmt (approved):

plugin/build.gradle
```diff
@@ -189,7 +189,7 @@ integTest {
 }
 
 testClusters.integTest {
-    testDistribution = "ARCHIVE"
+    testDistribution = "INTEG_TEST"
```
saratvemulapalli, line 192: Why change to `INTEG_TEST`?

b4sjoo: If not we will get this error in integ_test:
```
| Caused by: java.lang.IllegalStateException: jar hell!
| class: com.amazonaws.encryptionsdk.CryptoMaterialsManager
| jar1: /local/home/seasonsg/dev/ml-commons/plugin/build/testclusters/integTest-0/distro/3.0.0-ARCHIVE/lib/aws-encryption-sdk-java-2.4.0.jar
| jar2: /local/home/seasonsg/dev/ml-commons/plugin/build/testclusters/integTest-0/distro/3.0.0-ARCHIVE/plugins/.installing-2105471977921602206/aws-encryption-sdk-java-2.4.0.jar
```

saratvemulapalli: hm.. I suspect we are going to see problems. 
`INTEG_TEST` does not pull in modules which might work for us for tests but when we have to run it with the distribution(release) it will have modules installed and we will end up with JarHell.
Could you manually try installing the plugin with OpenSearch 2.x locally assembled with this change? 
Can you verify we dont hit the problem ?

b4sjoo: I see. This fix is introduced by @HenryL27 so I think we need to involve him in the loop. If so do you have any idea on fixing this jarhell issue?

plugin/build.gradle
```diff
@@ -322,7 +322,7 @@ jacocoTestCoverageVerification {
 check.dependsOn jacocoTestCoverageVerification
 
 configurations.all {
-    resolutionStrategy.force 'org.apache.commons:commons-lang3:3.10'
+    resolutionStrategy.force 'org.apache.commons:commons-lang3:3.13.0'
```
zane-neo, line 325: Do we still need to force the version if we already use `"${versions.commonslang}"` to get the version dynamically?

plugin/build.gradle
zane-neo: Tried in my local, with all the changes in this PR, using distribution won't have jar hell issue.

b4sjoo:
Close this due to the core PR has been reverted.

PR #1315, opened 2023-09-09
Fix unassigned ml system shard replicas

b4sjoo:
### Description
In previous ml system shard replicas always stays on an arbitrary number and cannot expand automatically, which will cause unassigned shards and yellow cluster problem. This PR can resolve this issue.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -117,7 +117,7 @@ public class MLIndicesHandler {
                                                     internalListener.onFailure(exception);
                                                 }));
                                         } else {
-                                            internalListener.onFailure(new MLException("Failed to update index " + indexName));
+                                            internalListener.onFailure(new MLException("Failed to update index: " + indexName));
                                         }
                                     }, exception -> {
                                         log.error("Failed to update index " + indexName, exception);
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -68,7 +68,7 @@ public class MLIndicesHandler {
     public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener) {
         String indexName = index.getIndexName();
         String mapping = index.getMapping();
-        final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-6");
+        final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-5");
 
         try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
             ActionListener<Boolean> internalListener = ActionListener.runBefore(listener, () -> threadContext.restore());
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -109,10 +109,10 @@ public class MLIndicesHandler {
                                                     internalListener.onResponse(true);
                                                 } else {
                                                     internalListener
-                                                        .onFailure(new MLException("Failed to update index setting: " + indexName));
+                                                        .onFailure(new MLException("Failed to update index setting for: " + indexName));
                                                 }
                                             }, exception -> {
-                                                log.error("Failed to update index setting: " + indexName, exception);
+                                                log.error("Failed to update index setting for: " + indexName, exception);
                                                 internalListener.onFailure(exception);
                                             }));
                                     }, exception -> {
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -116,6 +116,8 @@ public class MLIndicesHandler {
                                                     log.error("Failed to update index setting for: " + indexName, exception);
                                                     internalListener.onFailure(exception);
                                                 }));
+                                        } else {
+                                            internalListener.onFailure(new MLException("Failed to update index" + indexName));
                                         }
                                     }, exception -> {
                                         log.error("Failed to update index " + indexName, exception);
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -117,10 +117,10 @@ public class MLIndicesHandler {
                                                     internalListener.onFailure(exception);
                                                 }));
                                         } else {
-                                            internalListener.onFailure(new MLException("Failed to update index" + indexName));
+                                            internalListener.onFailure(new MLException("Failed to update index " + indexName));
                                         }
                                     }, exception -> {
-                                        log.error("Failed to update index: " + indexName, exception);
+                                        log.error("Failed to update index " + indexName, exception);
                                         internalListener.onFailure(exception);
                                     })
                                 );
--- a/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
+++ b/plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
@@ -98,23 +98,25 @@ public class MLIndicesHandler {
                                 .putMapping(
                                     new PutMappingRequest().indices(indexName).source(mapping, XContentType.JSON),
                                     ActionListener.wrap(response -> {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -68,6 +68,7 @@ public void initMLConfigIndex(ActionListener<Boolean> listener) {
     public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener) {
         String indexName = index.getIndexName();
         String mapping = index.getMapping();
+        final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-6");
```
ylwu-amzn, line 71: So the maximum replica is 6?

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -97,12 +98,19 @@ public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener)
...
+                                        updateSettingRequest.indices(indexName).settings(indexSettings);
+                                        client
+                                            .admin()
+                                            .indices()
+                                            .updateSettings(updateSettingRequest, ActionListener.wrap(updateResponse -> {
+                                                if (response.isAcknowledged()) {
+                                                    indexMappingUpdated.get(indexName).set(true);
+                                                    internalListener.onResponse(true);
+                                                } else {
+                                                    internalListener.onFailure(new MLException("Failed to update index: " + indexName));
```
ylwu-amzn, line 111: Tune the error message 
```suggestion
                                                    internalListener.onFailure(new MLException("Failed to update index setting: " + indexName));
```

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -97,12 +98,19 @@ public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener)
...
+                                                }
+                                            }, internalListener::onFailure));
```
ylwu-amzn, line 113: Suggest log the error first, check line 115

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
ylwu-amzn: Seems no need to recreate this map for every request, create a constant field?

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -68,6 +68,7 @@ public void initMLConfigIndex(ActionListener<Boolean> listener) {
     public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener) {
         String indexName = index.getIndexName();
         String mapping = index.getMapping();
+        final Map<String, Object> indexSettings = Map.of("index.auto_expand_replicas", "0-5");
```
dhrubo-os, line 71: Can we add any comment why we chose `0-5`?

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -97,12 +98,19 @@ public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener)
...
+                                        updateSettingRequest.indices(indexName).settings(indexSettings);
+                                        client
+                                            .admin()
+                                            .indices()
+                                            .updateSettings(updateSettingRequest, ActionListener.wrap(updateResponse -> {
+                                                if (response.isAcknowledged()) {
+                                                    indexMappingUpdated.get(indexName).set(true);
+                                                    internalListener.onResponse(true);
+                                                } else {
+                                                    internalListener.onFailure(new MLException("Failed to update index setting: " + indexName));
```
dhrubo-os, line 111: How about:

`internalListener.onFailure(new MLException("Failed to update index settings for: " + indexName));`

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
dhrubo-os: +1

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -41,8 +40,9 @@ public class MLIndicesHandler {
 
     private static final Map<String, AtomicBoolean> indexMappingUpdated = new HashMap<>();
     static {
-        indexMappingUpdated.put(ML_MODEL_INDEX, new AtomicBoolean(false));
-        indexMappingUpdated.put(ML_TASK_INDEX, new AtomicBoolean(false));
+        for (MLIndex mlIndex : MLIndex.values()) {
```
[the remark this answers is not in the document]
b4sjoo: This is suggested by @ylwu-amzn

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
b4sjoo: Looks good, I will use this

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
b4sjoo: To be consistent with AOS default settings

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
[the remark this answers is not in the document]
ylwu-amzn: MLIndex is enum. We need to get all index names and add to `indexMappingUpdated`

ylwu-amzn (dismissed):

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -97,12 +98,23 @@ public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener)
                                 .putMapping(
                                     new PutMappingRequest().indices(indexName).source(mapping, XContentType.JSON),
                                     ActionListener.wrap(response -> {
-                                        if (response.isAcknowledged()) {
-                                            indexMappingUpdated.get(indexName).set(true);
-                                            internalListener.onResponse(true);
-                                        } else {
-                                            internalListener.onFailure(new MLException("Failed to update index: " + indexName));
-                                        }
+                                        UpdateSettingsRequest updateSettingRequest = new UpdateSettingsRequest();
```
ylwu-amzn, line 101: I see you removed `if (response.isAcknowledged()) {`, I think we should keep this checking

b4sjoo: In line 107

b4sjoo: Or you think we need to check twice?

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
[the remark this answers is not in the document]
dhrubo-os: Oh I see. That's what I was curious about.

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
ylwu-amzn: If index mapping not sucessfull, no need to update setting, I think we should twice

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/indices/MLIndicesHandler.java
```diff
@@ -98,10 +99,23 @@ public void initMLIndexIfAbsent(MLIndex index, ActionListener<Boolean> listener)
                                     new PutMappingRequest().indices(indexName).source(mapping, XContentType.JSON),
                                     ActionListener.wrap(response -> {
                                         if (response.isAcknowledged()) {
-                                            indexMappingUpdated.get(indexName).set(true);
-                                            internalListener.onResponse(true);
-                                        } else {
-                                            internalListener.onFailure(new MLException("Failed to update index: " + indexName));
+                                            UpdateSettingsRequest updateSettingRequest = new UpdateSettingsRequest();
```
ylwu-amzn, line 102: For for this line, but you need to also add back these
```
else {	                                            client
                                            internalListener.onFailure(new MLException("Failed to update index: " + indexName));
```

ylwu-amzn (approved):
LGTM. Thanks for the change!

dhrubo-os (approved):

PR #1329, opened 2023-09-12
return model id in registering remote model

Zhangxunmt:
### Description
This change only applies to registering remote models. Registering a remote model doesn't need Asynchronize, it should return model_id right away in the response. 

**Before this change:**
{
    "name": "openAI-GPT-3.5 completions",
    "function_name": "remote",
    "description": "test model",
    "connector_id": "kUFFL4kB4ubqQRzeQ_r-"
}
Response:
{
  "task_id": "l0FIL4kB4ubqQRzeKvpV",
  "status": "CREATED"
}

**After this change:**
{
    "name": "openAI-GPT-3.5 completions",
    "function_name": "remote",
    "description": "test model",
    "connector_id": "kUFFL4kB4ubqQRzeQ_r-"
}
Response:
{
  "task_id": "l0FIL4kB4ubqQRzeKvpV",
  "status": "CREATED",
  "model_id": "xxxxxxxxxxxxxx",
}

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelResponse.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelResponse.java
@@ -18,15 +18,18 @@ import java.io.IOException;
 @Getter
 public class MLRegisterModelResponse extends ActionResponse implements ToXContentObject {
     public static final String TASK_ID_FIELD = "task_id";
+    public static final String MODEL_ID_FIELD = "model_id";
     public static final String STATUS_FIELD = "status";
 
     private String taskId;
     private String status;
+    private String modelId;
 
     public MLRegisterModelResponse(StreamInput in) throws IOException {
         super(in);
         this.taskId = in.readString();
         this.status = in.readString();
+        this.modelId = in.readOptionalString();
     }
 
     public MLRegisterModelResponse(String taskId, String status) {
@@ -34,10 +37,17 @@ public class MLRegisterModelResponse extends ActionResponse implements ToXConten
         this.status= status;
     }
 
+    public MLRegisterModelResponse(String taskId, String status, String modelId) {
+        this.taskId = taskId;
+        this.status= status;
+        this.modelId = modelId;
+    }
+
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         out.writeString(taskId);
         out.writeString(status);
+        out.writeOptionalString(modelId);
     }
 
     @Override
@@ -45,6 +55,9 @@ public class MLRegisterModelResponse extends ActionResponse implements ToXConten
         builder.startObject();
         builder.field(TASK_ID_FIELD, taskId);
         builder.field(STATUS_FIELD, status);
+        if (modelId != null) {
+            builder.field(MODEL_ID_FIELD, modelId);
+        }
         builder.endObject();
         return builder;
     }
--- a/plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
@@ -233,6 +233,7 @@ public class TransportRegisterModelAction extends HandledTransportAction<ActionR
                 throw new IllegalArgumentException("URL can't match trusted url regex");
             }
         }
+        System.out.println("registering the model");
         boolean isAsync = registerModelInput.getFunctionName() != FunctionName.REMOTE;
         MLTask mlTask = MLTask
             .builder()
@@ -249,8 +250,8 @@ public class TransportRegisterModelAction extends HandledTransportAction<ActionR
             mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
                 String taskId = response.getId();
                 mlTask.setTaskId(taskId);
-                mlModelManager.registerMLModel(registerModelInput, mlTask);
-                listener.onResponse(new MLRegisterModelResponse(taskId, MLTaskState.CREATED.name()));
+                System.out.println("mlModelManager calls registerMLRemoteModel");
+                mlModelManager.registerMLRemoteModel(registerModelInput, mlTask, listener);
             }, e -> {
                 logException("Failed to register model", e, log);
                 listener.onFailure(e);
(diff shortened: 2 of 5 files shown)


b4sjoo (approved):

rbhavna (approved):

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelResponse.java
```diff
@@ -18,33 +18,46 @@
...
     private String taskId;
     private String status;
+    private String modelId;
 
     public MLRegisterModelResponse(StreamInput in) throws IOException {
         super(in);
         this.taskId = in.readString();
         this.status = in.readString();
+        this.modelId = in.readOptionalString();
     }
```
saratvemulapalli, line 33: Feels like a good improvement for remote models, do we really have a taskId which is doing real work for remote models?

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -249,8 +250,8 @@ private void registerModel(MLRegisterModelInput registerModelInput, ActionListen
             mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
                 String taskId = response.getId();
                 mlTask.setTaskId(taskId);
-                mlModelManager.registerMLModel(registerModelInput, mlTask);
-                listener.onResponse(new MLRegisterModelResponse(taskId, MLTaskState.CREATED.name()));
+                System.out.println("mlModelManager calls registerMLRemoteModel");
```
saratvemulapalli, line 253: Nit: Probably unintended?

dhrubo-os, line 253: Are we keeping this line?

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -233,6 +233,7 @@ private void registerModel(MLRegisterModelInput registerModelInput, ActionListen
                 throw new IllegalArgumentException("URL can't match trusted url regex");
             }
         }
+        System.out.println("registering the model");
```
dhrubo-os, line 236: Are we keeping this?

ylwu-amzn: Let's avoid using system out print

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -388,6 +450,95 @@ public void registerMLModel(MLRegisterModelInput registerModelInput, MLTask mlTa
...
+            .doc(modelGroupSourceMap);
+
+        return updateModelGroupRequest;
+    }
+
+    private int incrementLatestVersion(Map<String, Object> modelGroupSourceMap) {
+        return (int) modelGroupSourceMap.get(MLModelGroup.LATEST_VERSION_FIELD) + 1;
+    }
+
+    private void indexRemoteModel(
```
dhrubo-os, line 479: Do we need to the other method: `private void indexRemoteModel(MLRegisterModelInput registerModelInput, MLTask mlTask, String modelVersion)` as we are invoking this method only.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -222,30 +224,22 @@ public void registerModelMeta(MLRegisterModelMetaInput mlRegisterModelMetaInput,
...
+                            UpdateRequest updateModelGroupRequest = createUpdateModelGroupRequest(
+                                modelGroupSource,
+                                modelGroupId,
+                                modelGroup.getSeqNo(),
+                                modelGroup.getPrimaryTerm(),
+                                updatedVersion
+                            );
+
+                            client.update(updateModelGroupRequest, ActionListener.wrap(r -> {
+                                uploadMLModelMeta(mlRegisterModelMetaInput, updatedVersion + "", listener);
```
dhrubo-os, line 238: regarding : `updatedVersion + ""` is the goal is to convert int to string, may be we can follow standard way like: `Integer.toString(updatedVersion)`?

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
Zhangxunmt: Good catch! I think I forgot to delete. They will be cleaned up for sure.

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -249,8 +250,8 @@ private void registerModel(MLRegisterModelInput registerModelInput, ActionListen
             mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
                 String taskId = response.getId();
                 mlTask.setTaskId(taskId);
-                mlModelManager.registerMLModel(registerModelInput, mlTask);
-                listener.onResponse(new MLRegisterModelResponse(taskId, MLTaskState.CREATED.name()));
+                System.out.println("mlModelManager calls registerMLRemoteModel");
+                mlModelManager.registerMLRemoteModel(registerModelInput, mlTask, listener);
```
dhrubo-os, line 254: Is this the expectation that, only remote model registration will be asynchronous? If yes, can we please add a comment here also?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
Zhangxunmt: Yes it's needed for registering non-remote models. Actually if you use the other method, the model_id will not be included and the model registering could be asynchronize.

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
Zhangxunmt: Yes I guarante these redundant lines will be removed in my next PR.

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
Zhangxunmt: Just the opposite, remote model registration is synchronous only so the model_id can be returned right away. Comments will be added in the next PR.

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelResponse.java
saratvemulapalli: Curious what kind of validations ?

PR #1336, opened 2023-09-13
include deployment status in deploy API response

Zhangxunmt:
### Description
This is to add deployment status with minimum impact to the current code and API workflows. Verified it works with local environment. 

# Before this change: 
POST /_plugins/_ml/models/mUFIL4kB4ubqQRzeKvr1/_deploy

sample response for remote model
{
  "task_id": "OP5JL4kB28KP1SGMupr3",
  "status": "CREATED"
}

This is ridiculous - deploying a remote model doesn't need to be asyncrhonize. I need to see the deployment status right away without sending another API request to query the Task Id!

# After this change: 
POST /_plugins/_ml/models/mUFIL4kB4ubqQRzeKvr1/_deploy

sample response for remote model
{
  "task_id": "OP5JL4kB28KP1SGMupr3",
  "task_type" : "DEPLOY_MODEL"
  "status": "COMPLETED"
}

For non remote models, the response will be 
{
  "task_id": "OP5JL4kB28KP1SGMupr3",
  "task_type" : "DEPLOY_MODEL"
  "status": "CREATED"
}

If the deployment fails, the response will send back the error message right away.

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
@@ -213,9 +213,14 @@ public class TransportDeployModelAction extends HandledTransportAction<ActionReq
                         mlTaskManager.createMLTask(mlTask, ActionListener.wrap(response -> {
                             String taskId = response.getId();
                             mlTask.setTaskId(taskId);
+                            if (algorithm == FunctionName.REMOTE) {
+                                mlTaskManager.add(mlTask, nodeIds);
+                                deployRemoteModel(mlModel, mlTask, localNodeId, eligibleNodes, deployToAllNodes, listener);
+                                return;
+                            }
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -23,6 +23,8 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import javax.swing.*;
```
b4sjoo, line 26: Can we avoid to import `javax.swing.*` here?

Zhangxunmt: Good catch! Removed this redundant line in the refresh!

b4sjoo (approved):

jngz-es (approved):

PR #1350, opened 2023-09-18
Update Model API

b4sjoo:
### Description
This PR should enable model updating through model update API

### Check List
  - [ ] All tests pass


plugin/build.gradle
```diff
@@ -301,7 +301,7 @@ jacocoTestCoverageVerification {
             excludes = jacocoExclusions
             limit {
                 counter = 'BRANCH'
-                minimum = 0.7  //TODO: change this value to 0.7
+                minimum = 0.0  //TODO: change this value to 0.7
```
dhrubo-os, line 304: Is this PR blocking anybody? If not, then let's add sufficient tests to meet the bar and then we can start reviewing.

b4sjoo: Relaxing this threshold due to tensive AppSec review date. I am actively working on tests and will keep pushing them. I will move back this line(or set to the highest number that coverage test can pass) before merging.

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
```diff
@@ -0,0 +1,274 @@
...
+    protected void doExecute(Task task, ActionRequest request, ActionListener<UpdateResponse> actionListener) {
+        MLUpdateModelRequest updateModelRequest = MLUpdateModelRequest.fromActionRequest(request);
+        MLUpdateModelInput updateModelInput = updateModelRequest.getUpdateModelInput();
+        String modelId = updateModelInput.getModelId();
+        UpdateRequest updateRequest = new UpdateRequest(ML_MODEL_INDEX, modelId);
+
+        try {
+            updateRequest.doc(updateModelInput.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
b4sjoo: This was suggested by intelliJ automatically, changed.

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
```diff
@@ -0,0 +1,274 @@
...
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            client.get(getModelRequest, ActionListener.wrap(r -> {
+                if (r != null && r.isExists()) {
+                    try (XContentParser parser = MLNodeUtils.createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
+                        String algorithmName = "";
+                        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                        if (r.getSource() != null && r.getSource().get(ALGORITHM_FIELD) != null) {
+                            algorithmName = r.getSource().get(ALGORITHM_FIELD).toString();
+                        } else
+                            throw new RuntimeException("FUNCTION_NAME_FIELD not found for this model, model ID " + modelId);
```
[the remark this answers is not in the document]
b4sjoo: Making sense to me, changed the exception to failure response. I think status code should be 5XX due to this failure belongs to server side

common/src/test/java/org/opensearch/ml/common/transport/model/MLUpdateModelInputTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.search.SearchModule;
+import org.opensearch.ml.common.model.MLModelConfig;
+import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
+
+public class MLUpdateModelInputTest {
+
+    private MLUpdateModelInput updateModelInput;
+    private final String expectedInputStr = "{\"model_id\":\"test-model_id\",\"description\":\"description\",\"name\":\"name\",\"model_group_id\":\"modelGroupId\",\"model_config\":" +
```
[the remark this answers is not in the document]
b4sjoo: No we can't. The model ID is not allowed to update, if you sent an update request like this it will directly ignore the model_id in your request body and will only use the model_id you provide in your API endpoint.

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
```diff
@@ -0,0 +1,277 @@
...
+        } else {
+            actionListener
+                .onFailure(
+                    new MLValidationException(
+                        "User doesn't have privilege to perform this operation on this function category: "
+                            + mlModel.getAlgorithm().toString()
+                    )
+                );
+        }
+    }
```
jngz-es, line 227: What about refactoring to the following? if you early fail as my previous comment mentioned, you can remove the first branch also.

if (mlModel.getAlgorithm() != TEXT_EMBEDDING && mlModel.getAlgorithm() != REMOTE) {
            actionListener
                    .onFailure(
                            new MLValidationException(
                                    "User doesn't have privilege to perform this operation on this function category: "
                                            + mlModel.getAlgorithm().toString()
                            )
                    );
        } else if (relinkConnectorId == null) {
            updateModelWithOrWithoutRelinkModelGroup(modelId, relinkModelGroupId, user, updateRequest, actionListener, context);
        } else if (mlModel.getAlgorithm() == TEXT_EMBEDDING) {
            actionListener
                    .onFailure(new IllegalArgumentException("Trying to update the connector or connector_id field on a local model"));
        } else {
            updateModelWithRelinkStandAloneConnector(
                    modelId,
                    relinkModelGroupId,
                    relinkConnectorId,
                    mlModel,
                    user,
                    updateRequest,
                    actionListener,
                    context
            );
        }

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
```diff
@@ -0,0 +1,277 @@
...
+            client.get(getModelRequest, ActionListener.wrap(r -> {
+                if (r != null && r.isExists()) {
+                    try (XContentParser parser = MLNodeUtils.createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
+                        String algorithmName = "";
+                        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                        if (r.getSource() != null && r.getSource().get(ALGORITHM_FIELD) != null) {
+                            algorithmName = r.getSource().get(ALGORITHM_FIELD).toString();
+                        } else {
+                            actionListener
+                                .onFailure(new RuntimeException("FUNCTION_NAME_FIELD not found for this model, model ID " + modelId));
```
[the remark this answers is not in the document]
b4sjoo: Nice catch, I will change it.

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
```diff
@@ -0,0 +1,277 @@
...
+                                if (Boolean.TRUE.equals(hasPermission)) {
+                                    updateRemoteOrTextEmbeddingModel(
+                                        modelId,
+                                        updateModelInput,
+                                        mlModel,
+                                        user,
+                                        updateRequest,
+                                        actionListener,
+                                        context
+                                    );
```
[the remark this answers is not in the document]
b4sjoo: I wrote like this due to code readability, I want each branch to be as short as possible and the condition to be as clear as possible, let me think about another way

plugin/src/main/java/org/opensearch/ml/action/models/TransportUpdateModelAction.java
b4sjoo: I moved it to outer block.

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,280 @@
...
+                    }
+                }, exception -> {
+                    log.error("Permission denied: Unable to update the connector with ID {}. Details: {}", relinkConnectorId, exception);
+                    actionListener.onFailure(exception);
+                }));
+        } else {
+            actionListener
+                .onFailure(
+                    new IllegalArgumentException("This remote does not have a connector_id field, maybe it uses an internal connector.")
+                );
```
jngz-es, line 190: Why don't we allow users to update model from inner connector to a standard connector?

jngz-es (commented):
Do we undeploy the model when updating it? I think we should do it.

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
b4sjoo: We will support update internal connector eventually after update connector API has been finalized, due to this part is a completely copy-paste to that one.

b4sjoo:
> Do we undeploy the model when updating it? I think we should do it.

Nice catch. This should be further discussed.

dhrubo-os:
Tests are failing..

b4sjoo:
> Tests are failing..

Hmmm...let me check

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
b4sjoo: So not supporting inner connector(for now) is by design.

b4sjoo:
> Do we undeploy the model when updating it? I think we should do it.

We will undeploy the model.

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
```diff
@@ -0,0 +1,177 @@
...
+    private String modelId;
+    private String description;
+    private String name;
+    private String modelGroupId;
+    private MLModelConfig modelConfig;
+    private String connectorId;
+
+    @Builder(toBuilder = true)
+    public MLUpdateModelInput(String modelId, String description, String name, String modelGroupId, MLModelConfig modelConfig, String connectorId) {
+        this.modelId = modelId;
```
ylwu-amzn, line 47: Add some validation? For example the model id is mandatory, What if all fields are null except `modelId`?

common/src/test/java/org/opensearch/ml/common/transport/model/MLUpdateModelInputTest.java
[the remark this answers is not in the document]
ylwu-amzn: Seems not necessary to put model id in request body, similar to update detectyor API, https://opensearch.org/docs/latest/observing-your-data/ad/api/#update-detector `PUT _plugins/_anomaly_detection/detectors/<detectorId>` , we can use `PUT _plugins/_ml/models/<model_Id>`

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
b4sjoo: I think that's unnecessary. Update the model with an empty body is ok due to nothing has changed. It will do no harm to anything. (OS built-in update allows update with an empty body)

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,326 @@
...
+        MLModel mlModel,
+        User user,
+        UpdateRequest updateRequest,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        SearchRequest searchRequest = new SearchRequest(ML_MODEL_INDEX);
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder();
+        BoolQueryBuilder boolQueryBuilder = QueryBuilders.boolQuery();
+        boolQueryBuilder.must(QueryBuilders.matchQuery(MLModel.MODEL_ID_FIELD, modelId));
```
ylwu-amzn, line 181: MLModel.modelId is only for model chunk only. I think it should not being used here

common/src/test/java/org/opensearch/ml/common/transport/model/MLUpdateModelInputTest.java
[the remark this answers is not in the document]
b4sjoo: Besides, we used POST method here due to implementing OS built-in update method.  https://opensearch.org/docs/latest/api-reference/document-apis/update-document/

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,326 @@
...
+                    user,
+                    updateRequest,
+                    actionListener,
+                    context
+                );
+            }
+        }
+    }
+
+    private void updateModelWithOrWithoutRelinkModelGroup(
```
ylwu-amzn, line 278: What does this function do? `relinkModelGroupId` means the new model group id? 
Suggest to avoid using new word to describe the same thing to avoid confusion. "Relink" may confuse as we don't have concept to "link" model to model group. We just add/register a model to a model group.

b4sjoo: What about `registerNewModelGroupId`?

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,326 @@
...
+            actionListener.onFailure(e);
+        }
+        updateRequest.docAsUpsert(true);
+        User user = RestActionUtils.getUserContext(client);
+
+        MLModelGetRequest mlModelGetRequest = new MLModelGetRequest(modelId, false);
+        FetchSourceContext fetchSourceContext = getFetchSourceContext(mlModelGetRequest.isReturnContent());
+        GetRequest getModelRequest = new GetRequest(ML_MODEL_INDEX).id(modelId).fetchSourceContext(fetchSourceContext);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
ylwu-amzn, line 107: You stash context here, but seems not restore context

b4sjoo: So do we need a `ActionListener.runBefore`?

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
```diff
@@ -0,0 +1,177 @@
...
+            out.writeBoolean(true);
+            modelConfig.writeTo(out);
+        } else {
+            out.writeBoolean(false);
+        }
+        out.writeOptionalString(connectorId);
+    }
+
+    public static MLUpdateModelInput parse(XContentParser parser, String modelId) throws IOException {
+        String description = null;
```
HenryL27, line 105: maybe have this call the parse method without `modelId` and then set `modelId` in the result?

b4sjoo: What do you mean by "set `modelId` in the result?"

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,326 @@
...
+                        if (r.getSource() != null && r.getSource().get(ALGORITHM_FIELD) != null) {
+                            algorithmName = r.getSource().get(ALGORITHM_FIELD).toString();
+                        } else {
+                            actionListener.onFailure(new RuntimeException("ALGORITHM_FIELD not found for this model, model ID " + modelId));
+                        }
+                        MLModel mlModel = MLModel.parse(parser, algorithmName);
+                        if (mlModel.getAlgorithm() == TEXT_EMBEDDING || mlModel.getAlgorithm() == REMOTE) {
+                            modelAccessControlHelper
+                                .validateModelGroupAccess(user, mlModel.getModelGroupId(), client, ActionListener.wrap(hasPermission -> {
+                                    if (Boolean.TRUE.equals(hasPermission)) {
```
HenryL27, line 122: can we just say `if(hasPermission)`?

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -0,0 +1,326 @@
...
+    ) {
+        SearchRequest searchRequest = new SearchRequest(ML_MODEL_INDEX);
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder();
+        BoolQueryBuilder boolQueryBuilder = QueryBuilders.boolQuery();
+        boolQueryBuilder.must(QueryBuilders.matchQuery(MLModel.MODEL_ID_FIELD, modelId));
+        boolQueryBuilder.must(QueryBuilders.idsQuery().addIds(mlModelManager.getAllModelIds()));
+        sourceBuilder.query(boolQueryBuilder);
+        searchRequest.source(sourceBuilder);
+
+        client.search(searchRequest, ActionListener.wrap(searchResponse -> {
```
HenryL27, line 186: might need another stashContext/restoreContext here depending how you resolve https://github.com/opensearch-project/ml-commons/pull/1350/files#r1344812309

HenryL27: oh I see that gets passed through the entire chain and restored at the very end, nvm

HenryL27: tho I think ActionListener.onFailure doesn't throw its exception so the unhappy cases may still drop the thread?

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateModelAction.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public String getName() {
+        return ML_UPDATE_MODEL_ACTION;
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList
+            .of(new Route(RestRequest.Method.POST, String.format(Locale.ROOT, "%s/models/_update/{%s}", ML_BASE_URI, PARAMETER_MODEL_ID)));
```
rbhavna, line 40: Its better to use "PUT" method and avoid using _update in the path. You can refer to model_group update API. This convention has been followed by other plugins like AD

ylwu-amzn (approved):

rbhavna (approved):

PR #1364, opened 2023-09-21
[#1341]: Add dynamic setting for disk circuit breaker 

### LINKED ISSUE #1341: Add dynamic setting for disk circuit breaker
Disk circuit breaker ([code link](https://github.com/opensearch-project/ml-commons/blob/2.x/plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java#L21)) has a hard coded threshold as 5 GB. We should make it as a dynamic setting.

rawwar:
### Description
Adds dynamic setting for disk circuit breaker 
 
### Issues Resolved
#1341 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
+++ b/plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
@@ -5,11 +5,14 @@
 
 package org.opensearch.ml.breaker;
 
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_DISK_SHORTAGE_THRESHOLD;
 import java.io.File;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.ml.common.exception.MLException;
 
 /**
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
```diff
@@ -20,6 +23,7 @@ public class DiskCircuitBreaker extends ThresholdCircuitBreaker<Long> {
     private static final String ML_DISK_CB = "Disk Circuit Breaker";
     public static final long DEFAULT_DISK_SHORTAGE_THRESHOLD = 5L;
     private static final long GB = 1024 * 1024 * 1024;
+    private volatile float diskShortageThreshold = 5L;
```
jngz-es, line 26: Why is a float variable here given a long constant?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -165,4 +165,7 @@ private MLCommonsSettings() {}
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
+    
+    public static final Setting<Float> ML_COMMONS_DISK_SHORTAGE_THRESHOLD = Setting
+        .floatSetting("plugins.ml_commons.disk_shortage_threshold", 5L, 10L, 1L, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
jngz-es, line 170: Can we have a longSetting here? Why the min value is 10L and the max value is 1L for this setting? Did you have a successful test on this setting change on your local cluster? Or can we have a test case to cover it?

plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java
rawwar: I will fix it.

Issue #1303, opened 2023-09-08
[FEATURE] tags on model group&version


wujunshen:
*Goals*

At present, we can allow users to upload and deploy many models, and carry out subsequent train, predict and other operations.

However, if users want to manage these models, for example, they want to categorize them based on applicable business scenarios, and quickly query and filter out which models are applicable to a specific business scenario, this may be very difficult or increase the user's effort.

Therefore, we introduce the concept of tag, which allows users to put a specific tag on some models, so that they can better manage models and quickly find the models they need.

And now in ml-commons, there are two system level indexes, model group and model version, on both of which we can allow users to tag, and let both of them reuse tags with each other.

For example, if we add a tag to a model group, then all the model versions in the group can use that tag, and for the same tag that is used, different model versions can each give different tag values or content. In the query, users can first query the value or content of the label, quickly find the model version that they need to use, which is more convenient and faster than directly querying the model, and then through the naked eye to determine whether they need.

*Solution*

The tag is divided into 3 parts

* tag key
* tag type (only String and Number two types, Number is float type in java)
* tag value

When performing CRUD operations on the model group, we operate on the key&type. When performing CRUD operations on the model version, we operate on the key&value.

**Changes to Index Mapping**

Since model group and  model version indexes already exist in OS(`.plugins-ml-model-group` and `.plugins-ml-model`), we need to modify their mapping definitions in order to adapt them to the new tags field first.

The `.plugins-ml-model` and `.plugins-ml-model-group`  index data structure of the tags field can be defined as a list, where the elements are nested object of tags(the attributes include key&type in `.plugins-ml-model-group` and key&value in `.plugins-ml-model`), or as a map, where the elements use key-value(key-type in  `.plugins-ml-model-group` and key-value in `.plugins-ml-model`).

If we design tags as a list:

Advantages

* Clearer structure, each tag is an independent object, containing all relevant information.
* Easy to query, you can directly query the attributes of the tag object.
* It is easier to add and delete tags.

Disadvantages

* Occupy more storage space.
* It is not easy to do aggregation analysis, you need to expand the list first.

If we design tags as a map:

Advantages

* Less storage space.
* Easier to do aggregation analysis, can be directly based on the key for statistics.

Disadvantages

* The structure is not clear enough, the key and value are scattered, the information is not centralized.
* Query is relatively complex, you need to get the key first, and then get the value.
* Adding and deleting tags requires updating both key and value.

[...]

PR #1370, opened 2023-09-25
[FEATURE] tags on model group

wujunshen:
see https://github.com/opensearch-project/ml-commons/issues/1303


common/src/main/java/org/opensearch/ml/common/CommonValue.java
zane-neo, line 1: Do not change the format of this file, the format change caused a lot of differences.

PR #1401, opened 2023-09-27
allow input null for text docs input

ylwu-amzn:
### Description
We need to support null value for some use case, for example multi-modal model needs text and image input. User who use TextDocsInputDataSet can use such format `[text, image]`, the first item is text and second is image, but they could be null, for example user may only input image, the input will be `[null, image]`

 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
@@ -114,7 +114,11 @@ public class TextDocsMLInput extends MLInput {
                 case TEXT_DOCS_FIELD:
                     ensureExpectedToken(XContentParser.Token.START_ARRAY, parser.currentToken(), parser);
                     while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
-                        docs.add(parser.text());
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
```diff
@@ -114,7 +114,11 @@ public TextDocsMLInput(XContentParser parser, FunctionName functionName) throws
                 case TEXT_DOCS_FIELD:
                     ensureExpectedToken(XContentParser.Token.START_ARRAY, parser.currentToken(), parser);
                     while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
-                        docs.add(parser.text());
+                        if (parser.currentToken() == null || parser.currentToken() == XContentParser.Token.VALUE_NULL) {
```
dhrubo-os, line 117: What happens if the input is `[null, null]`?

ylwu-amzn: The depends on the model. If model can't accept, it will throw exception.

dhrubo-os (approved):

rbhavna (approved):

Issue #844, opened 2023-04-04
[BUG]Model content hash can't match original hash value


dhrubo-os:
**What is the bug?**
`Model content hash can't match original hash value`

**How can one reproduce the bug?**

I tried with the current code base. I executed the command `./gradlew run` to test.

1. First I upload the model

```
POST /_plugins/_ml/models/_upload
{
  "name": "huggingface/sentence-transformers/all-MiniLM-L12-v2",
  "version": "1.0.1",
  "model_format": "TORCH_SCRIPT"
}
```

2. Then Load the model in memory. This time it works fine. `http://localhost:9200/_plugins/_ml/models/l-bnSYcBQg5VYC5uIxWy/_load`

3. I also generated the embedding with the loaded model and it works fine.

4. Now I unload the model: `http://localhost:9200/_plugins/_ml/models/l-bnSYcBQg5VYC5uIxWy/_unload`
5. And now I try to deploy again. Then I face the error: 

```
[2023-04-03T18:42:42,801][ERROR][o.o.m.m.MLModelManager   ] [integTest-0] Model content hash can't match original hash value
[2023-04-03T18:42:42,804][ERROR][o.o.m.a.f.TransportForwardAction] [integTest-0] deploy model failed on all nodes, model id: l-bnSYcBQg5VYC5uIxWy
[2023-04-03T18:42:42,804][INFO ][o.o.m.a.f.TransportForwardAction] [integTest-0] deploy model done with state: DEPLOY_FAILED, model id: l-bnSYcBQg5VYC5uIxWy
[2023-04-03T18:42:42,804][INFO ][o.o.m.a.d.TransportDeployModelOnNodeAction] [integTest-0] deploy model task done m-buSYcBQg5VYC5uMhUu

```

**What is the expected behavior?**
Model should load again.

**What is your host/environment?**
 - Version [e.g. 2.7]
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

saratvemulapalli:
@dhrubo-os Im trying to understand how ml-commons works and this issue seemed like a good one to pick up. 

I am trying to reproduce the problem, but I am unable to. 
Here are the steps:

1. `./gradlew run`
2. Toggle ml to run on data nodes
```
PUT  {{Host}}/_cluster/settings
{
    "persistent": {
        "plugins.ml_commons.only_run_on_ml_node": false
    }
}
```
3. Upload a model
```
[...]

PR #1414, opened 2023-10-02
Ignoring Redeploy test on MacOS due to known failures

saratvemulapalli:
### Description
Redeploying Model fails on MacOS which we officially don't support for ML-Commons yet. 
 
### Issues Resolved
Part of #844 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestMLDeployModelActionIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestMLDeployModelActionIT.java
@@ -27,6 +27,8 @@ public class RestMLDeployModelActionIT extends MLCommonsRestTestCase {
 
     @Before
     public void setup() throws IOException {
+        // Skip test if running on Mac OS, https://github.com/opensearch-project/ml-commons/issues/844
+        Assume.assumeFalse(System.getProperty("os.name").startsWith("Mac OS X"));
         mlRegisterModelGroupInput = MLRegisterModelGroupInput.builder().name("testGroupID").description("This is test Group").build();
         registerModelGroup(client(), TestHelper.toJsonString(mlRegisterModelGroupInput), registerModelGroupResult -> {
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestMLDeployModelActionIT.java
```diff
@@ -9,7 +9,9 @@
 import java.io.IOException;
 import java.util.Map;
 
+import org.junit.Assume;
 import org.junit.Before;
+import org.junit.Ignore;
```
dhrubo-os, line 14: Do we need this?

saratvemulapalli: Ignore the Ignore :).

plugin/src/test/java/org/opensearch/ml/rest/RestMLDeployModelActionIT.java
```diff
@@ -34,6 +36,9 @@ public void setup() throws IOException {
     }
 
     public void testReDeployModel() throws InterruptedException, IOException {
+        // Skip test if running on Mac OS, https://github.com/opensearch-project/ml-commons/issues/844
+        Assume.assumeFalse(System.getProperty("os.name").startsWith("Mac OS X"));
```
dhrubo-os, line 39: What do you think about adding this in the Setup? For mac, if we are ignoring the test, then may be we don't need to register the model group and subsequent lines in Setup?

saratvemulapalli: There could another test which could still run, but for now I agree it makes sense.

dhrubo-os (approved):

rbhavna (approved):

PR #1415, opened 2023-10-02
model registry fixes

rbhavna:
### Description
This PR fixes the following issues in Model Registry:

1. **Delete model group if the model version registry fails**: Whenever user registers a first model version, a model group is automatically created. Currently, if the version registration fails for some reason, the model group still remains in the index. This PR ensures the model group is created successfully only if model version registration is successful.
2. **Register new version to the model group based on the name provided**: After this PR, we let user register new version to an existing model group with the name alone even if model group ID is not provided. But if user does not have access to the model group in this case, we throw exception asking user to give a different name since user does not have access to the provided name
3. **Do not delete model group automatically with the last version of the model group**: After this PR, even if the last existing version of a model group is deleted, the model group still remains without getting deleted.
4. **Throw exception during update model group when model group does not exist**

 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
@@ -52,7 +52,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     public static final String ACCESS_MODE_FIELD = "access_mode";
     public static final String BACKEND_ROLES_FIELD = "backend_roles";
     public static final String ADD_ALL_BACKEND_ROLES_FIELD = "add_all_backend_roles";
-
+    public static final String IS_THIS_VERSION_CREATING_MODEL_GROUP = "is_this_version_creating_model_group";
     private FunctionName functionName;
     private String modelName;
     private String modelGroupId;
@@ -72,6 +72,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     private List<String> backendRoles;
     private Boolean addAllBackendRoles;
     private AccessMode accessMode;
+    private Boolean isThisVersionCreatingModelGroup;
 
     @Builder(toBuilder = true)
     public MLRegisterModelInput(FunctionName functionName,
@@ -89,7 +90,8 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
                                 String connectorId,
                                 List<String> backendRoles,
                                 Boolean addAllBackendRoles,
-                                AccessMode accessMode
+                                AccessMode accessMode,
+                                Boolean isThisVersionCreatingModelGroup
     ) {
         if (functionName == null) {
             this.functionName = FunctionName.TEXT_EMBEDDING;
@@ -122,6 +124,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
         this.backendRoles = backendRoles;
         this.addAllBackendRoles = addAllBackendRoles;
         this.accessMode = accessMode;
+        this.isThisVersionCreatingModelGroup = isThisVersionCreatingModelGroup;
     }
 
 
@@ -152,6 +155,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
         if (in.readBoolean()) {
             this.accessMode = in.readEnum(AccessMode.class);
         }
+        this.isThisVersionCreatingModelGroup = in.readOptionalBoolean();
     }
 
     @Override
@@ -197,6 +201,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
         } else {
             out.writeBoolean(false);
         }
+        out.writeOptionalBoolean(isThisVersionCreatingModelGroup);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -136,14 +136,51 @@ public TransportRegisterModelAction(
...
+        ActionListener<MLRegisterModelResponse> listener,
+        Boolean isModelNameAlreadyExisting
+    ) {
+        User user = RestActionUtils.getUserContext(client);
         modelAccessControlHelper
             .validateModelGroupAccess(user, registerModelInput.getModelGroupId(), client, ActionListener.wrap(access -> {
                 if (!access) {
-                    log.error("You don't have permissions to perform this operation on this model.");
-                    listener.onFailure(new IllegalArgumentException("You don't have permissions to perform this operation on this model."));
+                    if (isModelNameAlreadyExisting) {
```
dhrubo-os, line 170: I don't know if we should be super verbose here. Because some body might be surprised, I gave a model name and why it's being used as model group. What do you think about this message idea:

```
The model name 'xyz' is unavailable because it is already in use by another model group (Group ID: 'group_id'), to which you don't have access. Since you did not provide a model group ID, the system defaults to using the model name to register a new model group. Please choose a different name to proceed.
```

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -818,7 +815,7 @@ private void deployModelAfterRegistering(MLRegisterModelInput registerModelInput
         client.execute(MLDeployModelAction.INSTANCE, request, listener);
     }
 
-    private void deleteModel(String modelId) {
+    private void deleteModel(String modelId, String modelGroupID, Boolean isThisVersionCreatingModelGroup) {
```
dhrubo-os, line 818: I would suggest to add comment in the code to explain this use case.

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
```diff
@@ -52,7 +52,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     public static final String ACCESS_MODE_FIELD = "access_mode";
     public static final String BACKEND_ROLES_FIELD = "backend_roles";
     public static final String ADD_ALL_BACKEND_ROLES_FIELD = "add_all_backend_roles";
-
+    public static final String IS_THIS_VERSION_CREATING_MODEL_GROUP = "is_this_version_creating_model_group";
```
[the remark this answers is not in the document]
rbhavna: yeah seems better. Will change it

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
rbhavna: yeah sure

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
```diff
@@ -72,6 +72,7 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     private List<String> backendRoles;
     private Boolean addAllBackendRoles;
     private AccessMode accessMode;
+    private Boolean isThisVersionCreatingModelGroup;
```
ylwu-amzn, line 75: What's this for?

rbhavna: We use this later to delete the model group if version registration fails

ylwu-amzn: Can user set this parameter?

rbhavna: It is optional. This is mainly for us to use later in the code. Even if user sets it wrong, we will set it right based on the other input parameters provided

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -90,39 +89,42 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUpda
...
-                client.get(getModelGroupRequest, ActionListener.wrap(modelGroup -> {
-                    if (modelGroup.isExists()) {
-                        try (
-                            XContentParser parser = MLNodeUtils
-                                .createXContentParserFromRegistry(NamedXContentRegistry.EMPTY, modelGroup.getSourceAsBytesRef())
-                        ) {
-                            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
-                            MLModelGroup mlModelGroup = MLModelGroup.parse(parser);
+        GetRequest getModelGroupRequest = new GetRequest(ML_MODEL_GROUP_INDEX).id(modelGroupId);
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
ylwu-amzn, line 93: I see stash context here, but no restore?

rbhavna: Oh yeah. Let me use the wrappedListener here

ylwu-amzn (approved):

PR #1417, opened 2023-10-02
Conversation memory client

### LINKED ISSUE #1282: Move ConversationalMemoryClient to ml-client
As suggested in #1275, this client, which is currently only used by the RAG pipeline, can move to opensearch-ml-client so that it can be used by other applications.

@jngz-es @dhrubo-os do you guys think this should be merged with MachineLearningClient or are you thinking of a separate client.  By the way, MachineLearningClient currently gets built as a shadow jar and it is not meant for internal use.  The main thing I want to discuss here is whether or not we should have an external version of this memory client that makes it easy for applications to interact with memory.

Related: #1150

HenryL27:
### Description
Add a client for accessing conversation memory to the MLClient; similar pattern to the admin clients, e.g.
```
CreateConversationResponse response = machineLearningClient
                                       .memory()
                                       .createConversation(new CreateConversationRequest())
                                       .actionGet();
```

 
### Issues Resolved
#1282 
 
### Check List
  - [ x] All tests pass
- [ x] New functionality has been documented.
  - [ x] New functionality has javadoc added

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -14,6 +14,7 @@ plugins {
 
 dependencies {
     implementation project(':opensearch-ml-common')
+    implementation project(':opensearch-ml-memory')
     compileOnly group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
     testImplementation group: 'junit', name: 'junit', version: '4.13.2'
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '4.4.0'
--- a/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
+++ b/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
@@ -226,4 +226,10 @@ public interface MachineLearningClient {
      * @param listener action listener
      */
     void searchTask(SearchRequest searchRequest, ActionListener<SearchResponse> listener);
+
+    /**
+     * Get the memory client
+     * @return A Memory client for accessing conversation memory
+     */
+    MemoryClient memory();
 }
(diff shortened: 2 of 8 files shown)


HenryL27:
@jngz-es @dhrubo-os would be nice to get this added by 2.11 for @minalsha and the AI-Workflow team.
lmk if I should rename some stuff; maybe `MemoryClient -> ConversationMemoryClient`?

austintlee:
Two questions regarding this change:

1/ Should we move all transport and request/response classes for memory into org.opensearch.ml.common?
2/ Should we move all memory operations into the top level directly under MachineLearningClient as opposed to `.memory().<operation>`?

HenryL27:
we could do either of those things. I went for what I thought was the least invasive, although requiring that anyone who wants to use the memory client depend on the memory subproject is maybe a little invasive. Would be interested to hear @ylwu-amzn 's thoughts

jngz-es:
I prefer to Austin's idea. And I think we don't have to rush this pr into 2.11, right? @ylwu-amzn , we can have more discussion about it, as we will have a general memory layer.

HenryL27:
Austin had two ideas about this. You want both, or one? I agree, if there's gonna be a general memory layer the client should probably sit over that, although I think that it makes sense to separate it out of the full ML client. I'll probably end up dropping this PR and raising against main at some point, so maybe we can discuss what the memory client looks like on #1282 ?

jngz-es:
> Austin had two ideas about this. You want both, or one? I agree, if there's gonna be a general memory layer the client should probably sit over that, although I think that it makes sense to separate it out of the full ML client. I'll probably end up dropping this PR and raising against main at some point, so maybe we can discuss what the memory client looks like on #1282 ?

Both. Yes, we can have more discussions on that issue.

austintlee:
@jngz-es @ylwu-amzn We also want to unblock the AI Workflow team so let's try to converge on this ASAP.  Is there a write-up on the general memory layer?

ylwu-amzn:
@Zhangxunmt  can you help take a look?

Zhangxunmt (commented):
Can you pls rebase and pass the CI?

Zhangxunmt (approved):

HenryL27:
Closing as this was intended to support the flow framework, which has been built without it.

PR #1418, opened 2023-10-03
restore thread context before running action listener

ylwu-amzn:
### Description
We see reindex will throw error "wrong user in reader context". That's because after stashing thread context, the user info will be gone. Security plugin will recheck user info when reindex, then throw this error.  We should restore thread context before running action listener. 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/metrics_correlation/MetricsCorrelation.java
```diff
@@ -225,14 +225,14 @@ void registerModel(ActionListener<MLRegisterModelResponse> listener) throws Inte
             XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
             modelGroup.toXContent(builder, ToXContent.EMPTY_PARAMS);
             createModelGroupRequest.source(builder);
-            client.index(createModelGroupRequest, ActionListener.wrap(r -> {
+            client.index(createModelGroupRequest, ActionListener.runBefore(ActionListener.wrap(r -> {
```
dhrubo-os, line 228: May be we should add a similar comment in the code also what you have in the PR?

dhrubo-os (approved):

rbhavna (approved):

PR #1443, opened 2023-10-06
add status code to model tensor

ylwu-amzn:
### Description
Add http status code to model tensor, so client can know if the remote model returns correctly.

For example, if Cohere text embeding model exceed quota limit, it will return 429 status code. 
Screenshot 2023-10-06 at 9 16 31 AM

 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/output/model/ModelTensors.java
+++ b/common/src/main/java/org/opensearch/ml/common/output/model/ModelTensors.java
@@ -7,6 +7,7 @@ package org.opensearch.ml.common.output.model;
 
 import lombok.Builder;
 import lombok.Getter;
+import lombok.Setter;
 import org.opensearch.common.bytes.BytesReference;
 import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.io.stream.StreamInput;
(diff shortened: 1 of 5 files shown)


vibrantvarun:
Can you post the api request and response here?

b4sjoo (approved):

dhrubo-os (approved):

common/src/main/java/org/opensearch/ml/common/output/model/ModelTensors.java
```diff
@@ -41,6 +45,9 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             }
             builder.endArray();
         }
+        if (statusCode != null) {
```
navneet1v, line 48: status code are great. But, we need some more details on how client should respond to the status code. Like what are the retryable status code, which are permanent failures?

Another thing, this status code is exact status code from remote model, which is inconvenient from client like neural search. Either we should have ML commons also provide a status code around what it thinks of the error. Because different models can give different status code and then neural search cannot understand each of the model as it doesn't know what it really means and what response should be sent to customer.

PR #1459, opened 2023-10-06
if model version fails to register, update model group accordingly

rbhavna:
### Description
Delete/Update model group if the model version registration fails: Whenever user registers a first model version, a model group is automatically created. Currently, if the version registration fails for some reason, the model group still remains in the index. This PR ensures the model group is created successfully only if model version registration is successful. 

Also, if user is trying to register to a pre-existing model group and the registration fails for some reason, the latest_version field in the model group should not get updated. This PR ensures this change.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
+++ b/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
@@ -127,7 +127,8 @@ public final class MLCommonsSettings {
                 .of(
                     "^https://runtime\\.sagemaker\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
                     "^https://api\\.openai\\.com/.*$",
-                    "^https://api\\.cohere\\.ai/.*$"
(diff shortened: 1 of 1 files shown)


ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -831,6 +831,40 @@ private void deleteModel(String modelId) {
...
+            updateLatestVersionInModelGroup(
+                modelGroupID,
+                modelVersion,
+                ActionListener
+                    .wrap(r -> log.debug("model group updated, response {}", r), e -> log.error("Failed to update model group", e))
+            );
+        }
+    }
+
+    private void updateLatestVersionInModelGroup(String modelGroupID, String modelVersion, ActionListener<UpdateResponse> listener) {
```
dhrubo-os, line 855: Is it possible to make this method more generic to update the model version in both ways. (increment and decrement both)?

PR #1472, opened 2023-10-10
Performance enhacement for predict action by caching model info

zane-neo:
### Description
ML-commons predict API needs to check user permissions before the actual inference happen but the current implementation is not efficient since we need to read index for each single predict request. And fetching model info with getModel leverages the `GET` threadpool in OpenSearch Core, under heavy traffic, there might be rejection exception thrown since the threadpool is saturated.
 
### Issues Resolved
NA
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -92,6 +92,7 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
             ActionListener<MLModel> modelActionListener = new ActionListener<>() {
                 @Override
                 public void onResponse(MLModel mlModel) {
+                    context.restore();
                     modelCacheHelper.setModelInfo(modelId, mlModel);
                     FunctionName functionName = mlModel.getAlgorithm();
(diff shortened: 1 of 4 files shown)


ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -87,42 +88,66 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLTask
...
+                            }
+                        }, e -> {
+                            log.error("Failed to Validate Access for ModelId " + modelId, e);
+                            wrappedListener.onFailure(e);
+                        }));
+                }
 
+                @Override
+                public void onFailure(Exception e) {
+                    log.error("Failed to find model " + modelId, e);
```
ylwu-amzn, line 117: Should restore context here too, or use runBefore

ylwu-amzn: Sorry , my bad, it's already there in wrappedListenner
```
ActionListener<MLTaskResponse> wrappedListener = ActionListener.runBefore(listener, () -> context.restore());
```

PR #1482, opened 2023-10-11
fix no worker node exception for remote embedding model

ylwu-amzn:
### Description
For remote embedding model, error happens when ingest documents with neural-search ingest pipeline. Error says no worker nodes found for the model. But actually model already deployed.  neural-search ingest pipeline will send text_embedding request to ml-commons, in ml-commons we will translate it into remote model request. In ml-commons we can't use function name from neural-search(text-embedding), we should use the model's real function name. 

Related PR #1472 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -137,9 +137,16 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
         String requestId = mlPredictionTaskRequest.getRequestID();
         log.debug("receive predict request " + requestId + " for model " + mlPredictionTaskRequest.getModelId());
         long startTime = System.nanoTime();
+        // For remote text embedding model, neural search will set mlPredictionTaskRequest.getMlInput().getAlgorithm() as
+        // TEXT_EMBEDDING. In ml-commons we should always use the real function name of model: REMOTE. So we try to get
+        // from model cache first.
+        FunctionName functionName = modelCacheHelper
+            .getOptionalFunctionName(modelId)
+            .orElse(mlPredictionTaskRequest.getMlInput().getAlgorithm());
         mlPredictTaskRunner
             .run(
-                mlPredictionTaskRequest.getMlInput().getAlgorithm(),
+                // This is by design to NOT use mlPredictionTaskRequest.getMlInput().getAlgorithm() here
+                functionName,
                 mlPredictionTaskRequest,
                 transportService,
                 ActionListener.runAfter(wrappedListener, () -> {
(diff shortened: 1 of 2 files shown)


rbhavna (dismissed):

dhrubo-os (dismissed):

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -431,8 +431,10 @@ public boolean getDeployToAllNodes(String modelId) {
     }
 
     public void setModelInfo(String modelId, MLModel mlModel) {
-        MLModelCache mlModelCache = getExistingModelCache(modelId);
-        mlModelCache.setModelInfo(mlModel);
+        MLModelCache mlModelCache = modelCaches.get(modelId);
+        if (mlModelCache != null) {
```
ylwu-amzn, line 435: @zane-neo , we should check if model cache exist or not

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -137,9 +137,16 @@ private void executePredict(
...
+        // TEXT_EMBEDDING. In ml-commons we should always use the real function name of model: REMOTE. So we try to get
+        // from model cache first.
+        FunctionName functionName = modelCacheHelper
+            .getOptionalFunctionName(modelId)
+            .orElse(mlPredictionTaskRequest.getMlInput().getAlgorithm());
         mlPredictTaskRunner
             .run(
-                mlPredictionTaskRequest.getMlInput().getAlgorithm(),
+                // This is by design to NOT use mlPredictionTaskRequest.getMlInput().getAlgorithm() here
+                functionName,
```
zane-neo, line 149: We should fine-tune this part either in neural search side or in ml-commons side, more and more twisted code could cause more and more issues.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
zane-neo: I think we also need to fine tuning the error message in `getExistingModelCache` method, when model is not in cache, it should be a not deployed model, we should return the should deploy model info, so that in the future other method can use this method safely.

rbhavna (approved):

zane-neo (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
zane-neo: For current code, I think either local or remote needs to be deployed first, are you saying some feature will be implemented in the future?

PR #1485, opened 2023-10-11
fix for delete model group API throwing incorrect error when model index not created

rbhavna:
### Description
fix for delete model group API throwing incorrect error when model index not created
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
@@ -5,10 +5,9 @@
 
 package org.opensearch.ml.action.model_group;
 
-import static org.opensearch.ml.common.CommonValue.ML_MODEL_GROUP_INDEX;
-import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
-import static org.opensearch.ml.utils.RestActionUtils.PARAMETER_MODEL_GROUP_ID;
-
+import lombok.AccessLevel;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
```diff
@@ -84,26 +83,14 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
-                            });
+                            deleteModelGroup(deleteRequest, modelGroupId, wrappedListener);
                         } else {
                             throw new MLValidationException("Cannot delete the model group when it has associated model versions");
                         }
 
                     }, e -> {
                         if (e instanceof IndexNotFoundException) {
-                            wrappedListener.onFailure(new MLResourceNotFoundException("Fail to find model group"));
+                            deleteModelGroup(deleteRequest, modelGroupId, wrappedListener);
```
dhrubo-os, line 93: When index is not found why are we trying to delete the model again?

rbhavna: The index not found is the model version index. So here before deleting a model group, we are searching model index to check if there are any associated model versions in which case we do not delete the model group. But if index itself does not exit, it means there are no model versions indeed on the cluster and we can go ahead with deleting the model group

dhrubo-os (approved):

ylwu-amzn (approved):

PR #1488, opened 2023-10-11
removed empty host check, inherently within httpHost object generation

Arnav-Gr0ver:
### Description
if statement to check if httpHost name is not needed as an exception is already raised when creating the httpHost object for null, empty, and whitespace hostNames
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactoryTests.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactoryTests.java
@@ -102,5 +102,4 @@ public class MLHttpClientFactoryTests {
         HttpHost httpHost = new HttpHost("api.openai.com:65537", -1, "https");
         MLHttpClientFactory.validateSchemaAndPort(httpHost);
     }
-
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
@@ -57,10 +57,6 @@ public class MLHttpClientFactory {
 
     @VisibleForTesting
     protected static void validateSchemaAndPort(HttpHost host) {
-        if (Strings.isBlank(host.getHostName())) {
-            log.error("Remote inference host name is empty!");
-            throw new IllegalArgumentException("Host name is empty!");
-        }
         String scheme = host.getSchemeName();
         if ("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme)) {
             String[] hostNamePort = host.getHostName().split(":");


HenryL27:
why do we need to remove this check? 
where do the exceptions get thrown from other than here?
will this make it easier or harder to debug things?

Arnav-Gr0ver:
> why do we need to remove this check? where do the exceptions get thrown from other than here? will this make it easier or harder to debug things?

Hello Henry!

Apologies for the late reply, in regards to your questions:

1. the check, i.e. the if statement with _isBlank_ is redundant in regards to the checks already within the HttpHost class for the hostname (there is a method _containsNoBlanks_ which already throws the exceptions we wanted to test for)

2. 
```
public HttpHost(final String hostname, final int port, final String scheme) {
        super();
        this.hostname   = Args.containsNoBlanks(hostname, "Host name");
        this.lcHostname = hostname.toLowerCase(Locale.ROOT);
        if (scheme != null) {
            this.schemeName = scheme.toLowerCase(Locale.ROOT);
        } else {
            this.schemeName = DEFAULT_SCHEME_NAME;
        }
        this.port = port;
        this.address = null;
    }
```

and the containsNoBlanks method

```
public static <T extends CharSequence> T containsNoBlanks(final T argument, final String name) {
        if (argument == null) {
            throw new IllegalArgumentException(name + " may not be null");
        }
        if (argument.length() == 0) {
            throw new IllegalArgumentException(name + " may not be empty");
        }
        if (TextUtils.containsBlanks(argument)) {
            throw new IllegalArgumentException(name + " may not contain blanks");
        }
        return argument;
    }
```

3. I don't have enough experience with the codebase to definitively say xyz would be easier, but I was running into some issues when writing unit tests for this code block - the exception was not being thrown and the lines weren't executing hence impacting coverage - and I thought that there was already enough specificity about the type of invalid hostName in _containsNoBlanks_ over the general "host name is empty".

Would appreciate your feedback on this, and I went over this issue with @dhrubo-os & @jngz-es if they'd like to weigh in!

jngz-es:
Thanks for this pr, it looks good to me. Just minor comment, you can put the issue link under the **Issues Resolved** section.

ylwu-amzn:
Thanks @Arnav-Gr0ver , I think we should ask @zane-neo to help review too as he wrote this part.  Zan, can you help review if it's ok to remove this check?

zane-neo (approved):

dhrubo-os:
@Arnav-Gr0ver DCO is missing.

austintlee:
LGTM.

austintlee (dismissed):

ml-algorithms/src/test/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactoryTests.java
```diff
@@ -9,13 +9,16 @@
...
+import java.net.Inet4Address;
 import java.net.InetAddress;
+import org.junit.rules.ExpectedException;
+
 import java.net.UnknownHostException;
 
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.fail;
+import java.lang.reflect.Method;
+import static org.junit.Assert.*;
```
dhrubo-os, line 21: let's not import `.*`. https://www.baeldung.com/intellij-disable-wildcard-import

Arnav-Gr0ver: Sorry about that, I was doing some testing and believe I accidentally left the statement in, I revised the import

dhrubo-os: Seems like you missed the DCO again?

dhrubo-os (dismissed):

zane-neo (approved):

dhrubo-os (approved):

PR #1529, opened 2023-10-18
Read function Name from pretrained model

xinyual:
### Description
Currently if we register pretrained model without url, it will set the default function name to text_embedding since we only have text embedding pretrained model. But now we have sparse encoding, so we need to read the function name from pretrained model config.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
```diff
@@ -87,7 +87,8 @@ public void downloadPrebuiltModelConfig(String taskId, MLRegisterModelInput regi
                         .url(modelZipFileUrl)
                         .deployModel(deployModel)
                         .modelNodeIds(modelNodeIds)
-                        .modelGroupId(modelGroupId);
+                        .modelGroupId(modelGroupId)
+                        .functionName(FunctionName.from((String) config.get("model_task_type")));;
```
dhrubo-os, line 91: Can't we get Function name from `registerModelInput` like the way we get other inputs from 57-62 lines?

xinyual: The registerModelInput is from the request body json. So we can request customer to provide it. But if we want to keep the request convention to only contain "name, version, model_format", we can only read it from pretrained config.

ylwu-amzn: What if `config.get("model_task_type")` is null?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -609,7 +609,7 @@ private void uploadModel(MLRegisterModelInput registerModelInput, MLTask mlTask,
 
     private void registerModelFromUrl(MLRegisterModelInput registerModelInput, MLTask mlTask, String modelVersion) {
         String taskId = mlTask.getTaskId();
-        FunctionName functionName = mlTask.getFunctionName();
+        FunctionName functionName = registerModelInput.getFunctionName();
```
ylwu-amzn, line 612: Any reason to change to `registerModelInput` ? Will it break BWC?

xinyual: If we don't provide url, the function name from ML task would be read from request body, while the modelInput would be generated from config. If we still use the function name from ml task, the name would still be null and the default would be "text_embedding". I have tried locally for body with url and without url, both worked for me.

ylwu-amzn: MLTask must track the correct function name. Can you check if the function name in MLTask correct or not?

xinyual: No. I have tested it again. If the request body is like:
{
  "name": "amazon/neural-sparse/opensearch-neural-sparse-encoding-v1",
  "version": "1.0.0",
}

The function name inside the ml task would be text_embedding.

ylwu-amzn: I mean, after changing to `registerModelInput.getFunctionName();`, is the function name in MLTask correct now?

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
dhrubo-os: I feel like may be we should address [this section](https://github.com/opensearch-project/ml-commons/blob/2.x/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java#L97-L100)

currently we default to text embedding, which doesn't seem right. What happens if we start adding different pre-trained models like right now we added splade model.

xinyual:
Force push to address sign-off.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
xinyual: Already done by set function name inside mltask and rewrite to ml index.

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
xinyual: I think currently the logic is ok. The only scenario we don't input function name is when we use pretrained model, but if we fix the bug, the function name will be read from config and it is correct. So it's ok to keep a default value.

dhrubo-os: It's ok based on a assumption that, every pre-trained model will have `config.get("model_task_type")`.  I don't know how true that is.

zane-neo: > What if `config.get("model_task_type")` is null?

This should be fine, the FunctionName.valueOf is surrounded by try catch, and all exception case will be throwing `Wrong function name`  error.

zane-neo (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
xinyual: Until now, every pretrained model has its own and we can require the future contributor to have this. I agree we should address this section. But I don't have a clear solution towards this problem.

zane-neo: > I feel like may be we should address [this section](https://github.com/opensearch-project/ml-commons/blob/2.x/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java#L97-L100)
> 
> currently we default to text embedding, which doesn't seem right. What happens if we start adding different pre-trained models like right now we added splade model.

Can we make the function name mandatory, when it's null, throw exception instead of setting to a default value?

dhrubo-os: one solution could be: we can add function_name in our [model listing](https://artifacts.opensearch.org/models/ml-models/model_listing/pre_trained_models.json) and then get the function name from there.

xinyual: > > I feel like may be we should address [this section](https://github.com/opensearch-project/ml-commons/blob/2.x/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java#L97-L100)
> > currently we default to text embedding, which doesn't seem right. What happens if we start adding different pre-trained models like right now we added splade model.
> 
> Can we make the function name mandatory, when it's null, throw exception instead of setting to a default value?

No. When we register pretrained model,  we don't provide function name so in some scenario it would be null. We need a default value.

xinyual: > one solution could be: we can add function_name in our [model listing](https://artifacts.opensearch.org/models/ml-models/model_listing/pre_trained_models.json) and then get the function name from there.

I think both work. Model listing and pretrained config are both files we maintain, so it's just a option to read from which file in our s3 bucket.

dhrubo-os: I see what you mean now. We get this `model_task_type` from [config.json](https://artifacts.opensearch.org/models/ml-models/huggingface/sentence-transformers/all-mpnet-base-v2/1.0.1/torch_script/config.json) like this. Yeah this should work.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -756,6 +757,14 @@ private void registerPrebuiltModel(MLRegisterModelInput registerModelInput, MLTa
             throw new IllegalArgumentException("This model is not in the pre-trained model list, please check your parameters.");
         }
         modelHelper.downloadPrebuiltModelConfig(taskId, registerModelInput, ActionListener.wrap(mlRegisterModelInput -> {
+            mlTask.setFunctionName(mlRegisterModelInput.getFunctionName());
+            mlTaskManager
+                .updateMLTask(
```
dhrubo-os, line 762: Can you add testing for this? Code coverage is dropping.

xinyual: I have added one. Please check it.

ylwu-amzn (approved):

dhrubo-os:
There's a merge conflict.

xinyual:
> There's a merge conflict.

I guess we could do it now. I have merged from main branch.

dhrubo-os (approved):

zane-neo (approved):

PR #1533, opened 2023-10-20
Add execute task api for MLClient

jackiehanyang:
### Description
Add execute task api for MLClient
 
### Issues Resolved
 
### Check List
  - [x] All tests pass

### DIFF
--- a/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
+++ b/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
@@ -26,6 +26,7 @@ import org.opensearch.core.action.ActionResponse;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.MLTask;
+import org.opensearch.ml.common.input.Input;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.input.parameter.MLAlgoParams;
 import org.opensearch.ml.common.output.MLOutput;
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -182,6 +186,18 @@ public void registerModelGroup(
...
     }
 
+    /**
+     * Execute metrics correlation algorithm
+     * @param name Function name
+     * @param metricsCorrelationInput metrics correlation input
+     * @param listener a listener to be notified of the result
+     */
+    @Override
+    public void executeAlgorithm(FunctionName name, Input metricsCorrelationInput, ActionListener<MLExecuteTaskResponse> listener) {
```
dhrubo-os, line 196: From here it seems like we are opening up `execute` api only for metrics correlation algorithm? Should we open up `execute api` in general (the way we open up train/predict api for all algorithms)? May be not tied to only metrics correlation algorithm?

@ylwu-amzn what do you think?

jackiehanyang: There's no execute api in general. Already synced up with yaliang on this solution offline.

dhrubo-os: I see, if that's the case can we add more comment and TODO in the code?

ylwu-amzn: @jackiehanyang I think we have some miscommunication here. We only have metrics correlation now, but you should not name the parameter name as `metricsCorrelationInput` here, that's too specific. Suggest change to a more general name like `input` 
```suggestion
    public void executeAlgorithm(FunctionName name, Input input, ActionListener<MLExecuteTaskResponse> listener) {
```

jackiehanyang: updated the parameter name here - https://github.com/opensearch-project/ml-commons/pull/1540/files

PR #1554, opened 2023-10-26
Add process function for bedrock

zane-neo:
### Description
As BedRock is going to production, text embedding API in Bedrock will be available, currently there isn't pre/post process function for BedRock so it's difficult for customer to configure a correct text embedding connector. This PR adds pre/post process function for BedRock, one thing to highlight in this PR is: BedRock doesn't support batch texts to embedding so we need to refactor the existing methods to support this when neural search side passing a list of strings to embed. A valid connector request body looks like:
```
{
  ....
  "actions": [
          {
              "action_type": "predict",
              "method": "POST",
              "url": "xxx",
              "pre_process_function": "connector.pre_process.bedrock.embedding",
              "post_process_function": "connector.post_process.bedrock.embedding",
              "headers": {
              },
              "request_body": "{\"inputText\": \"${parameters.inputText}\"}"
          }
      ]
}
```
### Issues Resolved
 
### Check List
  - [ ] All tests pass


common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -18,38 +18,62 @@ public class MLPostProcessFunction {
...
     public static final String COHERE_EMBEDDING = "connector.post_process.cohere.embedding";
     public static final String OPENAI_EMBEDDING = "connector.post_process.openai.embedding";
-
+    public static final String BEDROCK_EMBEDDING = "connector.post_process.bedrock.embedding";
     public static final String DEFAULT_EMBEDDING = "connector.post_process.default.embedding";
 
     private static final Map<String, String> JSON_PATH_EXPRESSION = new HashMap<>();
 
-    private static final Map<String, Function<List<List<Float>>, List<ModelTensor>>> POST_PROCESS_FUNCTIONS = new HashMap<>();
+    private static final Map<String, Function<List<?>, List<ModelTensor>>> POST_PROCESS_FUNCTIONS = new HashMap<>();
```
dhrubo-os, line 26: Why do we need `<?>`? What other type are we expecting here? In the tensor datatype if we are ensuring `dataType(MLResultDataType.FLOAT32)` why do we need to change here?

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -18,38 +18,62 @@ public class MLPostProcessFunction {
...
+        JSON_PATH_EXPRESSION.put(BEDROCK_EMBEDDING, "$.embedding");
+        POST_PROCESS_FUNCTIONS.put(OPENAI_EMBEDDING, buildMultipleResultModelTensor());
+        POST_PROCESS_FUNCTIONS.put(COHERE_EMBEDDING, buildMultipleResultModelTensor());
+        POST_PROCESS_FUNCTIONS.put(DEFAULT_EMBEDDING, buildMultipleResultModelTensor());
+        POST_PROCESS_FUNCTIONS.put(BEDROCK_EMBEDDING, buildSingleResultModelTensor());
     }
 
-    public static Function<List<List<Float>>, List<ModelTensor>> buildModelTensorList() {
-        return embeddings -> {
+    public static Function<List<?>, List<ModelTensor>> buildSingleResultModelTensor() {
```
dhrubo-os, line 40: I didn't clearly understand why do we need two separate functions here?

Could you please explain what's the problem are we facing here?

Zhangxunmt (commented):
LGTM.
1. Can you add an example of the Predict API using this bedrock pre-processing and post-processing functions, and show the responses? This will make the documentation easier when needed. 
2. It looks like this won't impact other models in OpenAI, etc. But to be on the safe side, did you run any tests to make sure this is BWC due to the "buildModelTensorList" changing to "buildMultipleResultModelTensor"?

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
zane-neo: BedRock has different response with other platforms, BedRock doesn't support `list of string` input so the result is a single dimension list instead of double dimension `list of float`, to adapt to this difference, we need to change the type to a generic type here.

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
zane-neo: I was intended to separate these two methods to handle different data structure responses, but it's not optimal indeed, I have merged them into one by encapsulating a single dimension list into a list.

zane-neo:
> LGTM.
> 
> 1. Can you add an example of the Predict API using this bedrock pre-processing and post-processing functions, and show the responses? This will make the documentation easier when needed.
> 2. It looks like this won't impact other models in OpenAI, etc. But to be on the safe side, did you run any tests to make sure this is BWC due to the "buildModelTensorList" changing to "buildMultipleResultModelTensor"?

Added sample connector, predict request response samples and also tested on other models like OpenAI and Cohere, they all working as expected.

ylwu-amzn (approved):

Zhangxunmt (approved):

Issue #1470, opened 2023-10-09
[FEATURE] Support for timeout in http connector


austintlee:
**Is your feature request related to a problem?**
Occasionally, a request to OpenAI just sits for 10 minutes and fails with a "bad gateway" error.

**What solution would you like?**
I would like to specify the maximum amount of time an http connector should wait before it cancels a request.

**What alternatives have you considered?**

**Do you have any additional context?**

ylwu-amzn:
@zane-neo , Zan, since you built the http client factory, can you give some suggestion/solution for this?

ylwu-amzn:
@zane-neo  as you find other http client perfromance issue, can you create issue for tracking that?

zane-neo:
> @zane-neo , Zan, since you built the http client factory, can you give some suggestion/solution for this?

I think we can let user configure these, including: connection timeout, read timeout, connection pool size etc, I will create a PR to address this.

zane-neo:
> @zane-neo as you find other http client perfromance issue, can you create issue for tracking that?

Found it here: https://github.com/opensearch-project/ml-commons/issues/1537

austintlee:
#1835

ylwu-amzn:
Done in this PR https://github.com/opensearch-project/ml-commons/pull/1835

Issue #1537, opened 2023-10-23
[FEATURE] Performance issue in CloseableHttpClient


zane-neo:
**Is your feature request related to a problem?**
When performing benchmark test with heavy load in remote inference case, there's a chance to encounter the `Timeout to get connection` issue like below:
```
Caused by: org.apache.http.conn.ConnectionPoolTimeoutException: Timeout waiting for connection from pool
        at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.leaseConnection(PoolingHttpClientConnectionManager.java:316) ~[httpclient-4.5.14.jar:4.5.14]
        at org.apache.http.impl.conn.PoolingHttpClientConnectionManager$1.get(PoolingHttpClientConnectionManager.java:282) ~[httpclient-4.5.14.jar:4.5.14]
        at software.amazon.awssdk.http.apache.internal.conn.ClientConnectionRequestFactory$DelegatingConnectionRequest.get(ClientConnectionRequestFactory.java:92) ~[apache-client-2.20.19.jar:?]
        at software.amazon.awssdk.http.apache.internal.conn.ClientConnectionRequestFactory$InstrumentedConnectionRequest.get(ClientConnectionRequestFactory.java:69) ~[apache-client-2.20.19.jar:?]
        at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:190) ~[httpclient-4.5.14.jar:4.5.14]
        at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186) ~[httpclient-4.5.14.jar:4.5.14]
        at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185) ~[httpclient-4.5.14.jar:4.5.14]
        at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83) ~[httpclient-4.5.14.jar:4.5.14]
        at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:56) ~[httpclient-4.5.14.jar:4.5.14]
        at software.amazon.awssdk.http.apache.internal.impl.ApacheSdkHttpClient.execute(ApacheSdkHttpClient.java:72) ~[apache-client-2.20.19.jar:?]
        at software.amazon.awssdk.http.apache.ApacheHttpClient.execute(ApacheHttpClient.java:254) ~[apache-client-2.20.19.jar:?]
[...]

PR #1558, opened 2023-10-27
enable customize http client parameters

zane-neo:
### Description
Currently remote inference using CloseableHttpClient with default configurations both for AWS protocol and http protocol,  the drawback is customer can't configure the client base on their own use case. Sometimes issue can be found thrown from httpclient, this feature enables customer to configure their own parameters for httpclient.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1470
https://github.com/opensearch-project/ml-commons/issues/1537
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
@@ -3,15 +3,11 @@ package org.opensearch.ml.engine.algorithms.remote;
 import lombok.Getter;
 import lombok.Setter;
 
+@Setter
+@Getter
 public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor{
-    @Setter
-    @Getter
     private Integer maxConnections;
-    @Setter
-    @Getter
     private Integer connectionTimeoutInMillis;
-    @Setter
-    @Getter
     private Integer readTimeoutInMillis;
 
     public void validate() {
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
@@ -1,0 +1,28 @@
+package org.opensearch.ml.engine.algorithms.remote;
+
+import lombok.Getter;
+import lombok.Setter;
+
+public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor{
+    @Setter
+    @Getter
+    private Integer maxConnections;
+    @Setter
+    @Getter
+    private Integer connectionTimeoutInMillis;
+    @Setter
+    @Getter
+    private Integer readTimeoutInMillis;
+
+    public void validate() {
+        if (connectionTimeoutInMillis == null) {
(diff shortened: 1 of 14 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -177,4 +177,25 @@ private MLCommonsSettings() {}
     // Feature flag for enabling search processors for Retrieval Augmented Generation using OpenSearch and Remote Inference.
     public static final Setting<Boolean> ML_COMMONS_RAG_PIPELINE_FEATURE_ENABLED =
         GenerativeQAProcessorConstants.RAG_PIPELINE_FEATURE_ENABLED;
+
+    public static final Setting<Integer> ML_COMMONS_HTTP_CLIENT_CONNECTION_TIMEOUT_IN_MILLI_SECOND = Setting
```
dhrubo-os, line 181: What if customer wants timeout = 60 ms for remote model A, and timeout = 80 ms for remote model B?

May be customer can provide these parameters during setting up the connector? We'll have some default values, but customer can provide their expected values also if they want?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -0,0 +1,28 @@
+package org.opensearch.ml.engine.algorithms.remote;
+
+import lombok.Getter;
+import lombok.Setter;
+
+public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor{
```
Zhangxunmt, line 6: Why not just put @Setter and @Getter on top of this class so it applies to all field.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -0,0 +1,28 @@
+package org.opensearch.ml.engine.algorithms.remote;
```
Zhangxunmt, line 1: Is this missing the Copyright statement on top of this new class?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
Zhangxunmt: Why not include these three params in the request body of the "Predict" call to each LLM? If they are absent in the request, we can use the default values like these settings.

zane-neo: This can be implemented but we'd better make these changes in a different PR, for this PR the target is to enable customer to set these parameters to avoid exceptions.

ylwu-amzn: @zane-neo So we have three options now
1. Add a cluster setting like what this PR does
2. Add these parameters to connector configuration, from @dhrubo-os 
3. User can set these parameters in request body.

For my understanding, we already have a lot of cluster settings, maybe let's not add more cluster settings? Seems adding to connector level is more flexible, and if we add to connector parameters, we can also support option3 by default.

dhrubo-os: +1 on this.

PR #1582, opened 2023-11-02
CatIndexTool implementation

### LINKED ISSUE #1547: [FEATURE] Tool development for Cat Index Tool

Agent framework (#1161 ) needs tools to extend functionality. This issue propose to add cat index tools to it.
The cat index tool should return main information of index

- health status
- shard info: primary/replica shard count, size
- doc count

Build tool in ml-commons to retrieve all index information.
Agent framework (#1161 ) will provide a general tool framework. We can build search monitor/alerts tools with that framework.
 

1. Build against the Tool interface on the `agent_framework_dev` branch: https://github.com/opensearch-project/ml-commons/blob/feature/agent_framework_dev/spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java

2. Review example tools in the Agent Framework POC commit: https://github.com/opensearch-project/ml-commons/commit/1b85cffb07977a3aedb61a41637cb1b24f6cd016 where there are 6 tools under `ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/`. A good example is the `CatIndices` tool: https://github.com/opensearch-project/ml-commons/blob/feature/agent_framework_dev/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java which shows how to use the NodeClient (IndicesAdminClient) to fetch Cluster Health statistics.

N/A

#1161

dbwiddis:
### Description

Implements Cat Index Tool with edge case handling and tests
 
### Issues Resolved

Fixes #1547 
 
### Check List
  - [x] All tests pass


dbwiddis:
Test failures unassociated with this PR:
```
46 tests completed, 1 failed, 9 skipped
Tests with failures:
 - org.opensearch.ml.rest.RestMLDeployModelActionIT.testReDeployModel
```
Tests pass locally:
```
➜  ml-commons git:(cat-index) ./gradlew :opensearch-ml-algorithms:test --tests 'CatIndexToolTests'   
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 8.1.1
  OS Info               : Mac OS X 13.6 (x86_64)
  JDK Version           : 17 (Amazon Corretto JDK)
  JAVA_HOME             : /Library/Java/JavaVirtualMachines/amazon-corretto-17.jdk/Contents/Home
  Random Testing Seed   : BBF6D34AA5D9DD5B
  In FIPS 140 mode      : false
=======================================

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/8.1.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 5s
11 actionable tasks: 1 executed, 10 up-to-date
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
```diff
@@ -73,80 +70,98 @@ public Object parse(Object o) {
...
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        List<String> indexList = gson.fromJson(parameters.get("indices"), List.class);
-        String[] indices = parameters.containsKey("indices")? indexList.toArray(new String[0]) : new String[]{};
-
+        String[] indices = null;
+        if (parameters.containsKey("indices")) {
+            @SuppressWarnings("unchecked")
+            List<String> indexList = gson.fromJson(parameters.get("indices"), List.class);
+            indices = indexList.toArray(new String[0]);
```
dhrubo-os, line 77: what happens if a request comes to cat an system index?

dbwiddis: Huh. Looks like I need to use the getSettings approach [here](https://github.com/opensearch-project/OpenSearch/blob/8673fa937db405b8d614f8d4a02c0aa52587c037/server/src/main/java/org/opensearch/rest/action/cat/RestIndicesAction.java#L146)

dbwiddis: Fixed to match REST API

jngz-es:
A high level question. It turns out the implementation is same as RestIndicesAction in OpenSearch that should be OK for now. But how can we track the future changes of RestIndicesAction in OS? Can we have a way to reuse the code of RestIndicesAction to avoid the divergence between those two implementations?

dbwiddis:
> A high level question. It turns out the implementation is same as RestIndicesAction in OpenSearch that should be OK for now. But how can we track the future changes of RestIndicesAction in OS? Can we have a way to reuse the code of RestIndicesAction to avoid the divergence between those two implementations?

Great question:
1. In an ideal world, the REST action would have very little logic in it, and would just be a thin interface to the client API that would handle all the details.   However that's essentially what's being done with the sub-action calls and we're just stitching together "hey we want to call these 4 transport client APIs and get bits and pieces from each" and that's hard to standardize.
2. We could make the methods in the REST action public (I'm not fond of this option)
3. We could move the "cat index logic" to a common util class.  (I'm not excited about this in general, but for "legacy apis" seems a middle ground.)
4. We could implement the tool closer to the actual cat indices logic in OpenSearch using the ML Commons SPI (i'm not excited about this as it adds a dependency)
5. We could align on some combination of 3 and 4.
6. We consider this a rare exception and just live with it.  Maybe add a comment to the OpenSearch code that says "Hey, if you change this code, also change it in X other file." (That kind of stuff exists already in OpenSearch with multiple Request/Response classes for different client implementations.)

jngz-es:
Thanks Dan for those detailed options I really like. I don't think this is a blocking issue, so we are good to go for now. Could we add a TODO comment here for the future reference?

jngz-es (approved):
Thanks Dan, LGTM.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
```diff
@@ -73,126 +86,210 @@ public Object parse(Object o) {
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        List<String> indexList = gson.fromJson(parameters.get("indices"), List.class);
-        String[] indices = parameters.containsKey("indices")? indexList.toArray(new String[0]) : new String[]{};
+        // TODO: This logic exactly matches the OpenSearch _cat/indices REST action. If code at
```
dhrubo-os, line 89: >This logic exactly matches the OpenSearch _cat/indices REST action.

If that's true, what's the issue if we invoke this rest api, get the result and perform necessary post processing and send that to the llm? 

Sorry I might not have enough context on this.

dhrubo-os: If we want to add more functionalities such as _cat/shards, _cat/nodes, _cat/master, _cat/allocation, _cat/segments, _cat/transform, etc., will we continue to maintain separate code bases?

jngz-es: Hi Dhrubo, currently we don't have a way to directly call rest api in plugin, as the plugin doesn't know how to provide credentials needed in rest api. So Dan provided some solutions here we can address them later. I don't think it is a blocking issue for these kind of tools development for now.

dhrubo-os: Thanks @jngz-es for the clarification. I assume we can't invoke those APIs through transport layer either?

dbwiddis: @dhrubo-os we are invoking thru transport, the problem is that there is a lot of logic in between REST call and the transport.  We either have to duplicate the logic or find a common place to put it.

dbwiddis: @dhrubo-os to be more clear there is no transport API exactly corresponding to `_cat/indices`.  That REST API calls 4 transport APIs: settings to use as an index filter, and health (for that one stat), metadata (for uuid), and stats (for everything else), and then combine all the return values.  

In an ideal world we use the existing code, either by exposing currently private methods or by moving them to a public util class or by creating a new "parent" transport action above the logic that calls the other actions, etc.

dhrubo-os: Thanks for the clarification @dbwiddis . Approving the PR.

dhrubo-os (approved):

PR #1586, opened 2023-11-03
support step size for embedding model which outputs less embeddings

ylwu-amzn:
### Description
For some model , they may took multiple input docs and output less embeddings. For example,  multi-modal model will take 2 input docs, but output just 1 embedding. 
Add a step size parameter to control the process step for integration with neural-search.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import static org.opensearch.ml.engine.algorithms.remote.ConnectorUtils.processInput;
 
@@ -41,7 +42,14 @@ public interface RemoteConnectorExecutor {
                 if (tempTensorOutputs.size() > 0 && tempTensorOutputs.get(0).getMlModelTensors() != null) {
                     tensorCount = tempTensorOutputs.get(0).getMlModelTensors().size();
                 }
-                processedDocs += Math.max(tensorCount, 1);
+                // This is to support some model which takes N text docs and embedding size is less than N-1.
+                // We need to tell executor what's the step size for each model run.
+                Map<String, String> parameters = getConnector().getParameters();
+                int stepSize = 1;
+                if (parameters != null) {
+                    stepSize = Integer.parseInt(Optional.ofNullable(parameters.get("text_docs_processed_step_size")).orElse("1"));
+                }
+                processedDocs += Math.max(tensorCount, stepSize);
                 tensorOutputs.addAll(tempTensorOutputs);
             }
 
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
@@ -47,7 +47,7 @@ public interface RemoteConnectorExecutor {
                 Map<String, String> parameters = getConnector().getParameters();
                 int stepSize = 1;
                 if (parameters != null) {
-                    stepSize = Integer.parseInt(Optional.ofNullable(parameters.get("text_docs_processed_step_size")).orElse("1"));
+                    stepSize = Integer.parseInt(Optional.ofNullable(parameters.get("input_docs_processed_step_size")).orElse("1"));
                 }
                 processedDocs += Math.max(tensorCount, stepSize);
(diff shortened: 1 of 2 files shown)


rbhavna (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -41,7 +42,14 @@ default ModelTensorOutput executePredict(MLInput mlInput) {
                 if (tempTensorOutputs.size() > 0 && tempTensorOutputs.get(0).getMlModelTensors() != null) {
                     tensorCount = tempTensorOutputs.get(0).getMlModelTensors().size();
                 }
-                processedDocs += Math.max(tensorCount, 1);
+                // This is to support some model which takes N text docs and embedding size is less than N-1.
+                // We need to tell executor what's the step size for each model run.
+                Map<String, String> parameters = getConnector().getParameters();
+                int stepSize = 1;
+                if (parameters != null) {
+                    stepSize = Integer.parseInt(Optional.ofNullable(parameters.get("input_docs_processed_step_size")).orElse("1"));
```
dhrubo-os, line 50: `input_docs_processed_step_size` --> Just curious how do we get this variable?

dhrubo-os (dismissed):

dhrubo-os (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
ylwu-amzn: This is some reserved key for connector parameters, just like `region` for sagemaker model

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutorTest.java
```diff
@@ -190,4 +192,46 @@ public void executePredict_TextDocsInput() throws IOException {
...
+                + "        }    ],\n"
+                + "    \"model\": \"text-embedding-ada-002-v2\",\n" + "    \"usage\": {\n" + "        \"prompt_tokens\": 5,\n"
+                + "        \"total_tokens\": 5\n" + "    }\n" + "}";
+        StatusLine statusLine = new BasicStatusLine(new ProtocolVersion("HTTP", 1, 1), 200, "OK");
+        when(response.getStatusLine()).thenReturn(statusLine);
+        HttpEntity entity = new StringEntity(modelResponse);
+        when(response.getEntity()).thenReturn(entity);
+        when(executor.getHttpClient()).thenReturn(httpClient);
+        when(executor.getConnector()).thenReturn(connector);
+        MLInputDataset inputDataSet = TextDocsInputDataSet.builder().docs(Arrays.asList("test doc1", "test doc2")).build();
```
dhrubo-os, line 230: May be we should provide 4 documents to see if we are getting 2 outputs? But we can do it later too.

rbhavna (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -41,7 +42,14 @@ default ModelTensorOutput executePredict(MLInput mlInput) {
                 if (tempTensorOutputs.size() > 0 && tempTensorOutputs.get(0).getMlModelTensors() != null) {
                     tensorCount = tempTensorOutputs.get(0).getMlModelTensors().size();
                 }
-                processedDocs += Math.max(tensorCount, 1);
+                // This is to support some model which takes N text docs and embedding size is less than N-1.
+                // We need to tell executor what's the step size for each model run.
+                Map<String, String> parameters = getConnector().getParameters();
+                if (parameters != null && parameters.containsKey("input_docs_processed_step_size")) {
+                    processedDocs += Integer.parseInt(parameters.get("input_docs_processed_step_size"));
```
krishna-ggk, line 49: 1. Is this standard param? Can different connector have different parameter name to represent step size?
2. What is the value returned is invalid int or out of range? (-ve value)

ylwu-amzn: 1. This will be a reserved param for AI connector, just like the `region` for AWS region and `access_key` for AWS access key. All AI connector have to use this parameter to represent step size, just like they have to use `access_key` to represent AWS access key.
2. Good point, will add check logic, the step value should be greater than 0, not necessary to be less than input doc size.

ylwu-amzn: Addressed in this PR https://github.com/opensearch-project/ml-commons/pull/1587

PR #1587, opened 2023-11-03
validate step size

ylwu-amzn:
### Description
Validate step size
This is to address comments from this PR https://github.com/opensearch-project/ml-commons/pull/1586
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
@@ -42,11 +42,16 @@ public interface RemoteConnectorExecutor {
                 if (tempTensorOutputs.size() > 0 && tempTensorOutputs.get(0).getMlModelTensors() != null) {
                     tensorCount = tempTensorOutputs.get(0).getMlModelTensors().size();
                 }
-                // This is to support some model which takes N text docs and embedding size is less than N-1.
+                // This is to support some model which takes N text docs and embedding size is less than N.
                 // We need to tell executor what's the step size for each model run.
                 Map<String, String> parameters = getConnector().getParameters();
                 if (parameters != null && parameters.containsKey("input_docs_processed_step_size")) {
-                    processedDocs += Integer.parseInt(parameters.get("input_docs_processed_step_size"));
+                    int stepSize = Integer.parseInt(parameters.get("input_docs_processed_step_size"));
+                    // We need to check the parameter on runtime as parameter can be passed into predict request
+                    if (stepSize <= 0) {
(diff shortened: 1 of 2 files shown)


rbhavna (approved):

zane-neo (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -42,11 +42,16 @@ default ModelTensorOutput executePredict(MLInput mlInput) {
...
                 // We need to tell executor what's the step size for each model run.
                 Map<String, String> parameters = getConnector().getParameters();
                 if (parameters != null && parameters.containsKey("input_docs_processed_step_size")) {
-                    processedDocs += Integer.parseInt(parameters.get("input_docs_processed_step_size"));
+                    int stepSize = Integer.parseInt(parameters.get("input_docs_processed_step_size"));
+                    // We need to check the parameter on runtime as parameter can be passed into predict request
+                    if (stepSize <= 0) {
+                        throw new IllegalArgumentException("Invalid parameter: input_docs_processed_step_size. It must be positive integer.");
+                    }
+                    processedDocs += stepSize;
```
krishna-ggk, line 54: 1. parseInt can throw exception as well - can we handle that too?
2. I understand it is param for one type of connector, but should it be more generalized and property of connector?

ylwu-amzn: 1. It should ok to not handle, it will throw exception and will not crash the service. User should always follow our blueprint, sample blueprint https://github.com/opensearch-project/ml-commons/blob/2.x/docs/remote_inference_blueprints/bedrock_connector_anthropic_claude_blueprint.md. We will publish a new blueprint for multi-modal one soon.
2. Parameters map is one property of connector. This step size is just one parameter in it.

krishna-ggk: > Parameters map is one property of connector. This step size is just one parameter in it.

I don’t see input_docs_processed_step_size in blueprint - probably I still didn’t understand this clearly. Shouldn't this be generalized param we extract?





[17:15](https://amzn-aws.slack.com/archives/D03JLF8RENQ/p1699011954114609)
Wouldn’t it be different param we need to check for different connector? The code currently is in [RemoteConnectorExecutor.java](https://github.com/opensearch-project/ml-commons/pull/1587/files#diff-a5fee3d3840ca1ebf7229073c3a53af51fc3d1a99c960820c93d25613f67532e) (generic)

Issue #1161, opened 2023-07-28
[RFC] Agent framework


ylwu-amzn:
In 2.9, ml-commons support remote inference with connector. This issue designs how to build a general Agent framework by leveraging remote LLM.

# Why we need to build Agent?

To solve a complex problem, the process generally is hard to be predefined. We need to find some way to solve the problem step by step, identify potential solutions to reach a resolution. 

# Architecture

![ml-commons-cot-arc (1)](https://github.com/opensearch-project/ml-commons/assets/49084640/d2df5240-50f8-4d43-8d91-eb88499c62b2)

# Components

## Model

ml-commons released remote inference feature in 2.9. We can create remote model with LLM connector. For example create a remote model with OpenAI Chat model.

## Prompt Template

Define prompt template for LLM.

1. Prompt repo: generic prompt template
2. Prompt index: user’s custom prompt template

User can refer to the prompt template with prompt id in “prompt repo” or “prompt index”.

## Agent

Agent is a coordinator which uses LLM to reason what action to take to solve problem, then coordinate action execution. The action execution sequence could be not predefined/hard-coded. We plan to make the framework flexible to support multiple Agent types (like flow, CoT etc). But for the first phase, we can start from supporting conversational ReAct Agent.

## Tools

Tool is a function which can be executed by Agent. We will define a `Tool` framework in ml-commons. We will build some general built-in tools, for example

* `OpenSearchIndexTool` which will gather OpenSearch index information like index status
* `VectorDBTool` which will support running vector search
* `SeachIndexTool` which support search OpenSearch index

The `Tool` framework will be extensible. Other plugins can build their own Tools by implementing `Tool` interface. For example AnomalyDetection plugin can build `AnomalyResultTool` which can query anomaly detection result and analyze, and build "AnomalyDetectorTool" to create anomaly detector. Some use cases to use these tool
1. User ask "How many anomalies detected in last 24 hours". Then Agent will use `AnomalyResultTool` to query anomaly result index about how many anomalies detected in last 24 hours.
2. User ask "I have one index: my_log_data, it will receive realtime log data, can you help create an anomaly detector and monitor the anomaly and send out alarm to my email xxx@yyy.com". Then Agent will use `AnomalyDetectorTool` to create anomaly detector for this index, then use `AlertingMonitorTool` create a anomaly monitor with detector and user's email.

ml-commons will be the central place to manage all of these tools.  It will provide tool management functions:

* List all tools
* Get metadata of a specific tool
[...]

PR #1609, opened 2023-11-08
Implement Index Mapping Tool

dbwiddis:
### Description

Implements a tool to fetch index settings and mappings.
 
### Issues Resolved

In support of #1161
 
### Check List
  - [x] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.engine.tools;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.logging.log4j.util.Strings;
+import org.opensearch.action.admin.indices.get.GetIndexRequest;
+import org.opensearch.action.admin.indices.get.GetIndexResponse;
+import org.opensearch.action.support.IndicesOptions;
+import org.opensearch.client.Client;
+import org.opensearch.cluster.metadata.MappingMetadata;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import static org.opensearch.action.support.clustermanager.ClusterManagerNodeRequest.DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT;
+import static org.opensearch.ml.common.utils.StringUtils.gson;
+
+@ToolAnnotation(IndexMappingTool.NAME)
+public class IndexMappingTool implements Tool {
+    public static final String NAME = "IndexMappingTool";
+
+    private static final String DEFAULT_DESCRIPTION = "Use this tool to get index mapping information.";
+    @Setter
+    @Getter
+    private String name = IndexMappingTool.NAME;
+    @Getter
+    @Setter
+    private String description = DEFAULT_DESCRIPTION;
+    @Getter
+    private String type;
+    @Getter
+    private String version;
+    private Client client;
+    @SuppressWarnings("unused")
+    private String modelId;
+    @Setter
+    private Parser<?, ?> inputParser;
+    @Setter
+    private Parser<?, ?> outputParser;
+    @SuppressWarnings("unused")
+    private ClusterService clusterService;
+
+    public IndexMappingTool(Client client, ClusterService clusterService, String modelId) {
+        this.client = client;
+        this.clusterService = clusterService;
+        this.modelId = modelId;
+
+        outputParser = new Parser<>() {
+            @Override
+            public Object parse(Object o) {
+                @SuppressWarnings("unchecked")
+                List<ModelTensors> mlModelOutputs = (List<ModelTensors>) o;
+                return mlModelOutputs.get(0).getMlModelTensors().get(0).getDataAsMap().get("response");
+            }
+        };
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        @SuppressWarnings("unchecked")
+        List<String> indexList = parameters.containsKey("index")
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -0,0 +1,190 @@
...
+        };
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        @SuppressWarnings("unchecked")
+        List<String> indexList = parameters.containsKey("index")
+            ? gson.fromJson(parameters.get("index"), List.class)
+            : Collections.emptyList();
+        final String[] indices = indexList.toArray(Strings.EMPTY_ARRAY);
```
jngz-es, line 78: If indices is an empty list, should we early return the result, not go through the following logic then return empty result? Early returning can improve the tool execution performance in edge cases.

dbwiddis: Good point. I originally copied the code from cat indices where "empty" meant "match all the indices".  But I guess that isn't the case here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -0,0 +1,198 @@
...
+    @Setter
+    @Getter
+    private String name = IndexMappingTool.NAME;
+    @Getter
+    @Setter
+    private String description = DEFAULT_DESCRIPTION;
+    @Getter
+    private String type;
+    @Getter
+    private String version;
```
arjunkumargiri, line 46: Rather than defining these fields for each tool can we create an abstract tool with type and version defined?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -0,0 +1,198 @@
...
+    private Client client;
+    @SuppressWarnings("unused")
+    private String modelId;
```
arjunkumargiri, line 49: Curious why modelId is required if it is not used?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -0,0 +1,198 @@
...
+            public void onResponse(GetIndexResponse getIndexResponse) {
+                try {
+                    // Handle empty response
+                    if (getIndexResponse.indices().length == 0) {
+                        @SuppressWarnings("unchecked")
+                        T empty = (T) ("There were no results searching the index parameter [" + parameters.get("index") + "].");
+                        listener.onResponse(empty);
+                        return;
+                    }
+                    StringBuilder sb = new StringBuilder();
```
arjunkumargiri, line 103: Should we keep the output of tool follow json format and let the model understand from json format? this will help simplify each tool by directly returning json response

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
dbwiddis: Good question @arjunkumargiri .  These were not part of the `Tool` interface when I started this PR, and @jngz-es made a direct commit to this dev branch [here](https://github.com/opensearch-project/ml-commons/commit/90d9b310811c682f55ee5415d53ad0be8f671156#diff-f3f27d4e953486c2798c6edef8641779b3211dabbe5374c5bfe1737b3aab2f25) without any documentation of what they mean, so I added them here and in #1582 just to have something that implements the interface.

I was under the understanding that this dev branch was for rapid iteration (thus the non-PR maintainer commit linked above).  I'd really like to not hold up this PR until there is full documentation for the new committed interface that I'm required to implement.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
dbwiddis: Partly because I'm just copying over from the CatIndices where it was also unused but was based on @ylwu-amzn's POC code where it was unused as well.

There have been no detailed requirements written for what I'm writing this PR against, and I understand the `Tool` interface isn't finalized, so I'm doing my best to guess at what will eventually be needed.   

This can probably be deleted but I'd rather have it here unused and delete later than delete now and re-add later if we find out it's needed.

See previous comment on doing my best to rapidly iterate and keep up with changing requirements rather than waiting for everything to be finalized.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
dbwiddis: Good question, the requirements for the output have not been specified and the format I gave was in response to my question for how it should be.  If we want a different format it needs to be clearly specified for me to write code against.  @ylwu-amzn @jngz-es

ylwu-amzn: > Should we keep the output of tool follow json format and let the model understand from json format

Unless we do see the benefit of this, I would suggest not enforcing this. Keep it flexible and we could fine tune later.

> the requirements for the output have not been specified and the format I gave was in response to my question for how it should be

It's hard to tell what's the exact format which will works the best without testing and fine tuning. Let's not worry too much about format for now. We can iterate fast and fine tune later.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
ylwu-amzn: modelId is not required. The current code is not cleaned up. May have some testing code. Agree that let's don't wait, we can move forward and fine tune later.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
arjunkumargiri: Agree with moving forward on dev branch. However, I would prefer to define a unified format for all tool output so that it is easier for other tool development. If we are inclined on moving tuning the output later, I would prefer tools output in json format and let agent implement a common mechanism to covert from json to other formats.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
jngz-es: Good suggestion, @arjunkumargiri . As @dbwiddis  mentioned, this dev branch is for rapid development to unblock different teams, I can do refactoring afterwards.

ylwu-amzn (approved):

ylwu-amzn:
This PR opens for a long time. I don't see more comments on it. Just merged it to unblock Dan. You can add more comments, @dbwiddis , you can fix in new PR if necessary.

Issue #1589, opened 2023-11-03
[FEATURE] Support local cross-encoder model


HenryL27:
**Is your feature request related to a problem?**
We're trying to put a bunch of local model types in ml-commons (#1164). One such type is a [cross-encoder](https://www.sbert.net/examples/applications/cross-encoder/README.html). This will allow us to support [reranking](https://github.com/opensearch-project/neural-search/issues/485) in the neural-search plugin, which a lot of people have asked for.

**What solution would you like?**
Will be able to upload a custom cross-encoder model, deploy it, and use it with the upcoming neural search reranking processor.

**What alternatives have you considered?**
External hosting: Still would have to deal with the inputs and outputs, and then we also get the pleasure of figuring out some solution for externally hosting cross-encoders.

LTR: That can do reranking, but not cross-encoder reranking, so this isn't where that thought belongs.

**Do you have any additional context?**

HenryL27:
Gonna implement this as a new ml-algorithm / function name: `TEXT_SIMILARITY`. Cross-encoders are one technique to do this, but not the only one. Essentially it's just defined as (text1, text2) -> similarity_socre
(As opposed to embedding models which are (text) -> vector, and then similarity is an inner product of two vectors)

HenryL27:
A bunch of requests have come in for remote reranking models too, so I'll make sure that connectors can deal with TextSimilarityDatasets appropriately as well

ylwu-amzn:
Make sure local model and remote model can be switched smoothly .

ylwu-amzn:
> Make sure local model and remote model can be switched smoothly .

This is done in this PR https://github.com/opensearch-project/ml-commons/pull/1954

dhrubo-os:
>Make sure local model and remote model can be switched smoothly .

@HenryL27 will check remote model and then we can close this issue.

Issue #1164, opened 2023-07-29
[RFC] Support more local model types


ylwu-amzn:
Currently, ml-commons only supports uploading text-embedding models. However, we believe there are other models that could be valuable additions to our platform:

- Summarization model: This model can summarize a document or lengthy content and return a concise version.
- Question answering model: If you have a question related to a context or set of documents, this model can provide you with accurate answers.
- Text classification: This model is designed to classify text, such as sentiment analysis, where it assigns labels like 'positive,' 'negative,' or 'neutral.'
- Named entity recognition (NER) model: For identifying various entities in a text, such as organizations, persons, locations, and more.
- Image embedding model: This model translates an image into a vector representation for easier analysis.

Please comment on this issue if you require support for other local models or vote for the model you need the most.

hijakk:
+1. Most of these would be valuable in my use cases, as well as language identification. Possibly langid is a different beast than this approach should support? 

Image vectorization is a potentially tough use case as images can be large and including them in base64 natively in documents can dramatically inflate document size on disk rather than providing a reference pointer to external storage (such as s3).

ylwu-amzn:
From https://github.com/opensearch-project/ml-commons/issues/1150#issuecomment-1660940973

@asfoorial suggests
> I suggest to keep the door open for LLM hosting as there is a trend to get LLMs smaller with quantization yet achieve reasonable performance. I would say they will be hostable in ml nodes or other dedicated nodes.

nateynateynate:
They all get a thumbs up from me, but I actually would love to see image embedding. I'm fascinated by it.

HungryHowies:
+1

austintlee:
Are cross encoders covered under "rerank model"?

austintlee:
@dhrubo-os (tagging you since you went over some basics on this with the OCI students)

Would it make it easier to produce ML input/output classes for all these different models if we used Smithy to define them and have it generate the classes.  Just wondering what we can do to expedite progress on this using some common framework.

TrungBui59:
@ylwu-amzn @dhrubo-os I am interested in working on supporting the Question-answering model. Can you guys give me some hints on what I should do? Currently, I am thinking of following the approach that we used to support the text-embedding model

PR #1615, opened 2023-11-10
Add cross encoder support

HenryL27:
### Description
Adds support for (huggingface) cross encoders to ml-commons. Uses a new function name (`TEXT_SIMILARITY`) which takes as input a list of text pairs and spits out 1-dimensional tensors representing the similarity of the items in each pair. E.g.
```POST /_plugins/_ml/models/UWEvuosBD5KonuYrtwGX/_predict 
{
  "query_text": "today is sunny"
  "text_docs": [ 
    "today is sunny",
    "today is july fifth",
    "it is winter"
  ] 
}
```
yields
```
{
  "inference_results": [
    {
      "output": [
        {
          "name": "logits",
          "data_type": "FLOAT32",
          "shape": [
            1
          ],
          "data": [
            10.939743
          ],
          "byte_buffer": {
            "array": "MAkvQQ==",
            "order": "LITTLE_ENDIAN"
          }
        }
      ]
    },
    {
      "output": [
        {
          "name": "logits",
          "data_type": "FLOAT32",
          "shape": [
            1
          ],
          "data": [
            -6.067284
          ],
          "byte_buffer": {
            "array": "MSfCwA==",
            "order": "LITTLE_ENDIAN"
          }
        }
      ]
    },
    {
      "output": [
        {
          "name": "logits",
          "data_type": "FLOAT32",
          "shape": [
            1
          ],
          "data": [
            -11.261627
          ],
          "byte_buffer": {
            "array": "oC80wQ==",
            "order": "LITTLE_ENDIAN"
          }
        }
      ]
    }
  ]
}
```
This was using the model `cross-encoder/ms-marco-TinyBERT-L-2-v2` - the config I used to upload it looked like
```
{
  "name": model_name,
  "version": "1.0.0",
  "description": "Cross Encoder text similarity model",
  "function_name": "TEXT_SIMILARITY",
  "model_content_hash_value": hash_value,
  "model_config": {
    "model_type": "bert",
    "embedding_dimension": 1,
    "framework_type": "huggingface_transformers",
    "all_config": cfg.to_json_string(),
  }
}
```
 
### Issues Resolved
- Cross encoder issue: #1589 
- Model Types meta issue: #1164 
- Neural-search cross encoder reranking RFC: [opensearch-project/neural-search/485](https://github.com/opensearch-project/neural-search/issues/485#issue-1976944678)
 
### Check List
  - [ x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/FunctionName.java
+++ b/common/src/main/java/org/opensearch/ml/common/FunctionName.java
@@ -5,6 +5,9 @@
 
 package org.opensearch.ml.common;
 
+import java.util.HashSet;
+import java.util.Set;
+
 public enum FunctionName {
     LINEAR_REGRESSION,
     KMEANS,
@@ -36,9 +39,12 @@ public enum FunctionName {
      * @return true for deep learning model.
      */
     public static boolean isDLModel(FunctionName functionName) {
-        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE || functionName == TEXT_SIMILARITY) {
-            return true;
-        }
-        return false;
+        Set<FunctionName> dlmodels = new HashSet<>(Set.of(
+            TEXT_EMBEDDING,
+            TEXT_SIMILARITY,
+            SPARSE_ENCODING,
+            SPARSE_TOKENIZE
+        ));
+        return dlmodels.contains(functionName);
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/FunctionName.java
+++ b/common/src/main/java/org/opensearch/ml/common/FunctionName.java
@@ -17,6 +17,7 @@ public enum FunctionName {
     RCF_SUMMARIZE,
     LOGISTIC_REGRESSION,
     TEXT_EMBEDDING,
+    TEXT_SIMILARITY,
     SPARSE_ENCODING,
     SPARSE_TOKENIZE,
     METRICS_CORRELATION,
@@ -35,7 +36,7 @@ public enum FunctionName {
      * @return true for deep learning model.
      */
     public static boolean isDLModel(FunctionName functionName) {
-        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE) {
+        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE || functionName == TEXT_SIMILARITY) {
             return true;
         }
         return false;
--- a/common/src/main/java/org/opensearch/ml/common/FunctionName.java
+++ b/common/src/main/java/org/opensearch/ml/common/FunctionName.java
@@ -34,17 +34,18 @@ public enum FunctionName {
         }
     }
 
+    private static final HashSet<FunctionName> DL_MODELS = new HashSet<>(Set.of(
+        TEXT_EMBEDDING,
+        TEXT_SIMILARITY,
+        SPARSE_ENCODING,
+        SPARSE_TOKENIZE
+    ));
+
     /**
      * Check if model is deep learning model.
      * @return true for deep learning model.
      */
     public static boolean isDLModel(FunctionName functionName) {
-        Set<FunctionName> dlmodels = new HashSet<>(Set.of(
-            TEXT_EMBEDDING,
-            TEXT_SIMILARITY,
-            SPARSE_ENCODING,
-            SPARSE_TOKENIZE
-        ));
-        return dlmodels.contains(functionName);
+        return DL_MODELS.contains(functionName);
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -52,7 +52,7 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
     public TextSimilarityInputDataSet(StreamInput in) throws IOException {
         super(MLInputDataType.TEXT_SIMILARITY);
         int size = in.readInt();
-        this.pairs = new ArrayList<Pair<String, String>>(size);
+        this.pairs = new ArrayList<Pair<String, String>>();
         for(int i = 0; i < size; i++) {
             String query = in.readString();
             String context = in.readString();
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -46,7 +46,7 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
         Objects.requireNonNull(textDocs);
         Objects.requireNonNull(queryText);
         if(textDocs.isEmpty()) {
-            throw new IllegalArgumentException("pairs must be nonempty");
+            throw new IllegalArgumentException("No text documents provided");
         }
         this.textDocs = textDocs;
         this.queryText = queryText;
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -36,9 +36,9 @@ import lombok.experimental.FieldDefaults;
 @InputDataSet(MLInputDataType.TEXT_SIMILARITY)
 public class TextSimilarityInputDataSet extends MLInputDataset {
     
-    private List<String> textDocs;
+   List<String> textDocs;
 
-    private String queryText;
+   String queryText;
 
     @Builder(toBuilder = true)
     public TextSimilarityInputDataSet(String queryText, List<String> textDocs) {
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -38,14 +38,14 @@ import lombok.experimental.FieldDefaults;
 @InputDataSet(MLInputDataType.TEXT_SIMILARITY)
 public class TextSimilarityInputDataSet extends MLInputDataset {
     
-    private ModelResultFilter resultsFilter;
+    private ModelResultFilter resultFilter;
 
     private List<Pair<String, String>> pairs;
 
     @Builder(toBuilder = true)
-    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs, ModelResultFilter resultsFilter) {
+    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs, ModelResultFilter resultFilter) {
         super(MLInputDataType.TEXT_SIMILARITY);
-        this.resultsFilter = resultsFilter;
+        this.resultFilter = resultFilter;
         Objects.requireNonNull(pairs);
         if(pairs.isEmpty()) {
             throw new IllegalArgumentException("pairs must be nonempty");
@@ -63,9 +63,9 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
             this.pairs.add(Pair.of(query, context));
         }
         if(in.readBoolean()) {
-            this.resultsFilter = new ModelResultFilter(in);
+            this.resultFilter = new ModelResultFilter(in);
         } else {
-            this.resultsFilter = null;
+            this.resultFilter = null;
         }
     }
 
@@ -77,9 +77,9 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
             out.writeString(p.getLeft());
             out.writeString(p.getRight());
         }
-        if(this.resultsFilter != null) {
+        if(this.resultFilter != null) {
             out.writeBoolean(true);
-            resultsFilter.writeTo(out);
+            resultFilter.writeTo(out);
         } else {
             out.writeBoolean(false);
         }
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -46,7 +46,7 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
         Objects.requireNonNull(textDocs);
         Objects.requireNonNull(queryText);
         if(textDocs.isEmpty()) {
-            throw new IllegalArgumentException("No text documents provided");
+            throw new IllegalArgumentException("No text documents were provided");
         }
         this.textDocs = textDocs;
         this.queryText = queryText;
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -22,7 +22,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.ml.common.annotation.InputDataSet;
@@ -37,36 +36,40 @@ import lombok.experimental.FieldDefaults;
 @InputDataSet(MLInputDataType.TEXT_SIMILARITY)
 public class TextSimilarityInputDataSet extends MLInputDataset {
     
-    private List<Pair<String, String>> pairs;
+    private List<String> textDocs;
+
+    private String queryText;
 
     @Builder(toBuilder = true)
-    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs) {
+    public TextSimilarityInputDataSet(String queryText, List<String> textDocs) {
         super(MLInputDataType.TEXT_SIMILARITY);
-        Objects.requireNonNull(pairs);
-        if(pairs.isEmpty()) {
+        Objects.requireNonNull(textDocs);
+        Objects.requireNonNull(queryText);
+        if(textDocs.isEmpty()) {
             throw new IllegalArgumentException("pairs must be nonempty");
         }
-        this.pairs = pairs;
+        this.textDocs = textDocs;
+        this.queryText = queryText;
     }
 
     public TextSimilarityInputDataSet(StreamInput in) throws IOException {
         super(MLInputDataType.TEXT_SIMILARITY);
+        this.queryText = in.readString();
         int size = in.readInt();
-        this.pairs = new ArrayList<Pair<String, String>>();
+        this.textDocs = new ArrayList<String>();
         for(int i = 0; i < size; i++) {
-            String query = in.readString();
             String context = in.readString();
-            this.pairs.add(Pair.of(query, context));
+            this.textDocs.add(context);
         }
     }
 
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
-        out.writeInt(this.pairs.size());
-        for (Pair<String, String> p : this.pairs) {
-            out.writeString(p.getLeft());
-            out.writeString(p.getRight());
+        out.writeString(queryText);
+        out.writeInt(this.textDocs.size());
+        for (String doc : this.textDocs) {
+            out.writeString(doc);
         }
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -26,7 +26,6 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.ml.common.annotation.InputDataSet;
-import org.opensearch.ml.common.output.model.ModelResultFilter;
 
 import lombok.AccessLevel;
 import lombok.Builder;
@@ -38,14 +37,11 @@ import lombok.experimental.FieldDefaults;
 @InputDataSet(MLInputDataType.TEXT_SIMILARITY)
 public class TextSimilarityInputDataSet extends MLInputDataset {
     
-    private ModelResultFilter resultFilter;
-
     private List<Pair<String, String>> pairs;
 
     @Builder(toBuilder = true)
-    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs, ModelResultFilter resultFilter) {
+    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs) {
         super(MLInputDataType.TEXT_SIMILARITY);
-        this.resultFilter = resultFilter;
         Objects.requireNonNull(pairs);
         if(pairs.isEmpty()) {
             throw new IllegalArgumentException("pairs must be nonempty");
@@ -62,11 +58,6 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
             String context = in.readString();
             this.pairs.add(Pair.of(query, context));
         }
-        if(in.readBoolean()) {
-            this.resultFilter = new ModelResultFilter(in);
-        } else {
-            this.resultFilter = null;
-        }
     }
 
     @Override
@@ -76,12 +67,6 @@ public class TextSimilarityInputDataSet extends MLInputDataset {
         for (Pair<String, String> p : this.pairs) {
             out.writeString(p.getLeft());
             out.writeString(p.getRight());
-        }
-        if(this.resultFilter != null) {
-            out.writeBoolean(true);
-            resultFilter.writeTo(out);
-        } else {
-            out.writeBoolean(false);
         }
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
+++ b/common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2023 Aryn
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.opensearch.ml.common.dataset;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.ml.common.annotation.InputDataSet;
+import org.opensearch.ml.common.output.model.ModelResultFilter;
+
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.experimental.FieldDefaults;
+
+@Getter
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@InputDataSet(MLInputDataType.TEXT_SIMILARITY)
+public class TextSimilarityInputDataSet extends MLInputDataset {
+    
+    private ModelResultFilter resultsFilter;
+
+    private List<Pair<String, String>> pairs;
+
+    @Builder(toBuilder = true)
+    public TextSimilarityInputDataSet(List<Pair<String, String>> pairs, ModelResultFilter resultsFilter) {
+        super(MLInputDataType.TEXT_SIMILARITY);
+        this.resultsFilter = resultsFilter;
+        Objects.requireNonNull(pairs);
+        if(pairs.isEmpty()) {
+            throw new IllegalArgumentException("pairs must be nonempty");
+        }
+        this.pairs = pairs;
+    }
+
+    public TextSimilarityInputDataSet(StreamInput in) throws IOException {
+        super(MLInputDataType.TEXT_SIMILARITY);
+        int size = in.readInt();
+        this.pairs = new ArrayList<Pair<String, String>>(size);
+        for(int i = 0; i < size; i++) {
+            String query = in.readString();
+            String context = in.readString();
+            this.pairs.add(Pair.of(query, context));
+        }
+        if(in.readBoolean()) {
+            this.resultsFilter = new ModelResultFilter(in);
+        } else {
+            this.resultsFilter = null;
+        }
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeInt(this.pairs.size());
+        for (Pair<String, String> p : this.pairs) {
+            out.writeString(p.getLeft());
+            out.writeString(p.getRight());
+        }
+        if(this.resultsFilter != null) {
+            out.writeBoolean(true);
+            resultsFilter.writeTo(out);
+        } else {
+            out.writeBoolean(false);
+        }
+    }
+}
--- a/common/src/main/java/org/opensearch/ml/common/input/MLInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/input/MLInput.java
@@ -8,6 +8,8 @@ package org.opensearch.ml.common.input;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+
+import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.xcontent.XContentBuilder;
@@ -21,6 +23,7 @@ import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.SearchQueryInputDataset;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
+import org.opensearch.ml.common.dataset.TextSimilarityInputDataSet;
 import org.opensearch.ml.common.input.parameter.MLAlgoParams;
 import org.opensearch.search.builder.SearchSourceBuilder;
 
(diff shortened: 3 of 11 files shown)


common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public TextSimilarityInputDataSet(StreamInput in) throws IOException {
+        super(MLInputDataType.TEXT_SIMILARITY);
+        int size = in.readInt();
+        this.pairs = new ArrayList<Pair<String, String>>();
+        for(int i = 0; i < size; i++) {
+            String query = in.readString();
+            String context = in.readString();
+            this.pairs.add(Pair.of(query, context));
```
austintlee, line 59: Is there a use case where `query` is _not_ the same in each of these pairs in a given input dataset?  Or is this string always going to be the same (repeated `size` times)?

austintlee (commented):
One minor question, but overall looks great!

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
HenryL27: I can't think of one off the top of my head, but I didn't want to preclude the possibility. Maybe you want to try out a bunch of queries over a couple docs and get like a cartesian product of scores?

austintlee (dismissed):

ylwu-amzn:
```
{
  "text_pairs": [ 
    ["today is sunny", "today is sunny"],
    ["today is sunny", "today is july fifth"],
    ["today is sunny", "it is winter"]
  ] 
}
```
Have you considered some other option like using current text docs input , but you process the docs two by two.

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
```diff
@@ -55,6 +58,8 @@ public class MLInput implements Input {
     public static final String TARGET_RESPONSE_POSITIONS_FIELD = "target_response_positions";
     // Input text sentences for text embedding model
     public static final String TEXT_DOCS_FIELD = "text_docs";
+    // Input text pairs for text similarity model
+    public static final String TEXT_PAIRS_FIELD = "text_pairs";
```
ylwu-amzn, line 62: Can we reuse text_docs to keep the interface simpler?

HenryL27:
> Have you considered some other option like using current text docs input, but you process the docs two by two.

@ylwu-amzn I thought about this. Ultimately I decided against it because 
1. I don't want people mixing up their model types. Cross-Encoders and Bi-Encoders have different input types and different output types, so let's formalize that in the code, rather than shoehorning a dataset type into a model it wasn't intended for. 
2. Still the same amount of data to serialize/deserialize, and the code isn't all that much of a diff.

By the same token, using `text_pairs` in the API forces the user to understand that this is a different kind of model that does different things. (esp since it's shaped differently I think it makes a lot of sense to keep it as a separate name. If I was making it just a double-long `TextDocsDS` then calling it `text_docs` would make sense)

HenryL27:
@ylwu-amzn @navneet1v @austintlee per this morning's meeting updated predict API:
```
POST /_plugins/_ml/models/<id>/_predict
{
  "query_text": "today is sunny",
  "text_docs": [
    "today is sunny",
    "today is july fifth",
    "it is winter"
  ]
}
```
Should mesh better with reranking, etc; Cohere connector should now not require any 'painless' scripting

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
HenryL27: @ylwu-amzn see updated interface

common/src/main/java/org/opensearch/ml/common/FunctionName.java
```diff
@@ -35,7 +36,7 @@ public static FunctionName from(String value) {
      * @return true for deep learning model.
      */
     public static boolean isDLModel(FunctionName functionName) {
-        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE) {
+        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE || functionName == TEXT_SIMILARITY) {
```
navneet1v, line 39: why we choose a name called as TEXT_SIMILARITY?

navneet1v: isn't this a re-ranking?

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,75 @@
...
+import lombok.experimental.FieldDefaults;
+
+@Getter
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@InputDataSet(MLInputDataType.TEXT_SIMILARITY)
+public class TextSimilarityInputDataSet extends MLInputDataset {
+    
+    private List<String> textDocs;
+
+    private String queryText;
```
navneet1v, line 41: when we add AccessLevel.PRIVATE you can remove `private` from the class variables.

```suggestion
@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
@InputDataSet(MLInputDataType.TEXT_SIMILARITY)
public class TextSimilarityInputDataSet extends MLInputDataset {
    
   List<String> textDocs;

   String queryText;
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public Output processOutput(TranslatorContext ctx, NDList list) {
+        Output output = new Output(200, "OK");
+
+        List<ModelTensor> outputs = new ArrayList<>();
+        Iterator<NDArray> iterator = list.iterator();
+        while (iterator.hasNext()) {
+            NDArray ndArray = iterator.next();
+            String name = ndArray.getName();
```
dhrubo-os, line 75: In your output (issue description), it's showing `logits`.  This doesn't seem quite meaningful. May be we can add function name here?

dhrubo-os: In addition, does this only support torchscript or both (torchscript & onnx)?

ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_similarity/TinyBERT-CE.zip
dhrubo-os, line 1: I understand this is important for testing. I'm just worried that ml-commons plugin size is getting bigger and bigger. If we add more model supports in the long run, then adding all the test resource files will increase the plugin size significantly bigger. 

@ylwu-amzn do you have any thoughts on this?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
dhrubo-os: +1 on this.

HenryL27: I chose `TEXT_SIMILARITY` because I think that best captures what this algorithm is about. Sure, we use it for Reranking, but you can use it for other stuff, too. For instance, one thing we've been using cross-encoders for is computing the semantic similarity between answers generated by our RAG systems and human-annotated answers that we know are correct, which is not a reranking use-case. Of course, we don't do that on the cluster, so maybe it's moot, but I wanted to pick a name that captures all use-cases of cross-encoders (or any models that spit out numbers representing how similar two pieces of text are). wdyt?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
```diff
@@ -17,6 +17,7 @@ public enum FunctionName {
     RCF_SUMMARIZE,
     LOGISTIC_REGRESSION,
     TEXT_EMBEDDING,
+    TEXT_SIMILARITY,
```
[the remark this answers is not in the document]
HenryL27: see above

common/src/main/java/org/opensearch/ml/common/FunctionName.java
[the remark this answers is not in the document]
HenryL27: or switch? is that the same thing?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
HenryL27: haven't tried with ONNX - never compiled an ONNX model before, so I don't know. But I _think_ I didn't do anything that would preclude it?

ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_similarity/TinyBERT-CE.zip
HenryL27: fwiw, this is one of the smallest models out there (TinyBERT-L-2-v2) - it's just two layers. But yeah, I was surprised to find all these zips when I was figuring out the testing

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
HenryL27: ack to logits. changing to similarity in next push, but I wanna resolve the function name thing first

common/src/main/java/org/opensearch/ml/common/FunctionName.java
dhrubo-os: I understand your point. But for semantic search, we also perform the "text similarity" with comparing large numbers of texts in our vector store. So IMO, text similarity can be achieved through bi-encoder & cross encoder both. In fact, with the efficiency concern bi-encoder is widely used for text similarity. 

But for Re-ranking cross encoder is widely used. So I would vote to rename this to `RE-RANK` algorithm and customer can choose which algorithm to choose to obtain their text similarity goal.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
dhrubo-os: If you look at the text embedding algorithm, we needed to add pooling layer information for onnx translator. So can we try to check if this will also work with onnx format or not?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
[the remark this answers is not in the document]
dhrubo-os: I prefer MAP, `DL_MODELS`

HenryL27: I  guess my concern with calling it `RERANK` is that there are also other reranking interfaces and models - not all are just reranking text. e.g. more traditional things like xgboost. idk. probably it's fine. maybe `TEXT_COMPARISON`?

HenryL27: > But for Re-ranking cross encoder is widely used. So I would vote to rename this to RE-RANK algorithm and customer can choose which algorithm to choose to obtain their text similarity goal.

When I proposed cross-encoder reranking to the neural search folks their main feedback was that reranking should be generalizable to more kinds of models than cross-encoders (see [this rfc](https://github.com/opensearch-project/neural-search/issues/485)), so I'm wary of synonymizing reranking with cross-encoders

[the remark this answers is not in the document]
HenryL27: ack

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2023 Aryn
```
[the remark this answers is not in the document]
HenryL27: as I understand, all this does is protect us on the off-chance that AWS does what Elastic did a few years ago and switches out the license or something. This is coming from Mehul, who oversaw that whole transition (elastic fork -> opensearch).

common/src/main/java/org/opensearch/ml/common/FunctionName.java
dhrubo-os: I'm not seeing any issue honestly. I would like to see this way: Currently we are going to support cross-encoder for re-ranking. We can extend `re-ranking` with other mechanism later when we have more clear requirement.

@navneet1v what do you think? Do you see any issue?

HenryL27: @navneet1v unless I'm misunderstanding, this enum is meant to be 1:1 with the model interface. So for example you can swap out MiniLM for openai-ada (local vs remote) within the bounds of `TEXT_EMBEDDING`, because both models take a list of strings and give back a list of vectors. 

This is an interface that takes a string and a list of strings and gives back a list of numbers (representing the similarity of the first string with list, though I guess the interface doesn't demand that). You could just as easily replace the cross encoder with something like a BLEU score (although you'd need to implement that algorithm) and not break the contract.

@dhrubo-os is that a correct understanding of FunctionName?

dhrubo-os: In my opinion, this is 1:N relationship with different model types. For example, for text embedding we are supporting multiple sentence embedding model types, like sentence transformers model, clip models. 

Let's take an example of `REMOTE`. This interface is supporting multiple model types including multiple LLMs as a remote model. 


@ylwu-amzn please correct me if I'm wrong.

ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_similarity/TinyBERT-CE.zip
ylwu-amzn: Agree that we should pay attention to the model size. Adding small model looks ok, the zip is only for testing, will not be packaged to release artifact.

common/src/main/java/org/opensearch/ml/common/FunctionName.java
ylwu-amzn: The function name is to group models with same interface and task types. Maybe some models for different tasks have same interface, but it maybe confused if we group them to one enum. I would suggest name for general task purpose. @HenryL27 , this is a long discussion, can you summarize what options we have now and explain the reason why people prefer the name?

HenryL27: sure. summary:
We're debating the name of the function that cross-encoders should use. Options:

1. `TEXT_SIMILARITY` (original choice): This is a model that measures the similarity of text. ergo, text similarity.
2. `RE_RANK`: This is a model used for reranking. ergo, re-rank
3. `TEXT_COMPARISON`: This is a model that compares pieces of text (similarity can be measured by vector distance and is therefore too broad. Although I guess an implementation of this algorithm that uses a bi-encoder and vector similarity is just as valid as the cross-encoder implementation), but the output can mean whatever you want it to mean, it's just a number.

The drawback of `RE_RANK` is that it seems to preclude other uses of cross-encoders, and also other kinds of reranking models.

The drawback of `TEXT_*` is that is obfuscates the main use-case (re-ranking)

@dhrubo-os did I miss anything?

ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_similarity/TinyBERT-CE.zip
HenryL27: btw, how do we add "pretrained models"?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
HenryL27: k. looks like onnx doesn't work. (can upload the model but the output is messed up). will enforce TORCH_SCRIPT format here and support ONNX in the future, sg?

HenryL27: the end of the response:
```
      {
          "name": "similarity",
          "data_type": "FLOAT32",
          "shape": [
            128
          ],
          "data": [
            0.9998161,
            -0.9851868,
            ...
            0.99590653,
            -0.18241836
          ],
          "byte_buffer": {
            "array": "8/N/PzQ1fL+s5n8/pvl0v1c7cD+h7HM/MpJ+P0KlST+1HHu/iD5vvy3Tfj/JEnS/CiV3vwAAgL9ulXM/Bf16P7vuRj+1AXu/SF9RPyUwfD/dVn+/5XZ0P73ybL8xt3+/vOx/vygae7/pgH4/Aad/P++yTD+IfnC/dQ9pv9yTfT9zwXm/TP9/vzL/fz85ln+/rGhMv6D9cL8U3n2//Tp4P5bjYj9GYXK/HSN/P8mkf78AAIA/KVJ+P+jtfj/C4X+/kxh/v0oTeT8SOnC/RE13v3NFfj9uJHQ/DQZ5PwAAgL8AAIA/A1x7v4yXfz8lqn8/fxt9v5byfr9LdX0/1ntzPwiqeT/3/38//bV/P68HYz/X73a/hal+P6O3f78cYXw/LFZ/Pzm1fL92+32/mIF8v+mwRb+OXHo/kPt7v31Ofb/PeXw/p1t3PxeBfL+Z/nU/tex/v3r8fj/hn3+/8kl8P0e9dT/ms3E/AACAP1t2db9ckni/DUZ4P//afb+AlXw/Jt1/P8qUf79v1n8/drd7v7FnV7+FXl+/8j9/P6uYfj/X3Hm/tWB6P5SvbT9Uwn8/Hudlv39Ffr9hmHw/TNp1P7nyfz8rtH8/AACAP8Anfj/RQXk/4S99P1bme79591e/kwN+v/r3fz83r36/Wmx+P9LIfz8i5G8/u/N+P+HLOr4=",
            "order": "LITTLE_ENDIAN"
          }
        }
      ]
    }
  ]
}
```
(my terminal cut off the top).
But yes, we'll probably need to add some custom onnx output translator to select the right entry in this tensor list (there were like nine of them? But varying sizes so maybe it gives you the hidden states is that possible?) and possibly convert to single number somehow. I would have thought the classification head would get pulled along with the model at compile time so hopefully it won't be too bad. I guess it's also possible my compile script was bad...
```
def _prepare_cross_encoder_onnx(model_name):
    tk = AutoTokenizer.from_pretrained(model_name)
    model = AutoModelForSequenceClassification.from_pretrained(model_name)
    mname = Path(model_name).name
    
    tk.save_pretrained(f"/tmp/{mname}-tokenizer")
    transformers.convert_graph_to_onnx.convert(
        framework="pt",
        model=model_name,
        output=Path(f"/tmp/onnx/{mname}.onnx"),
        opset=15,
    )
    r = requests.get("https://github.com/opensearch-project/opensearch-py-ml/raw/main/LICENSE")

    with ZipFile(f"/tmp/{mname}.zip", "w") as f:
        f.write(f"/tmp/onnx/{mname}.onnx", arcname=f"{mname}.onnx")
        f.write(f"/tmp/{mname}-tokenizer/tokenizer.json", arcname="tokenizer.json")
        f.writestr("LICENSE", r.content)
    shutil.rmtree(f"/tmp/{mname}-tokenizer")
    shutil.rmtree("/tmp/onnx/")
    return f"/tmp/{mname}.zip"
```
btw are we allowed to just slap an Apache license on all these models? That seems fishy to me

HenryL27: aha I guess it does drop the classification head, since it's just running off of the name. lemme see if there's a way to onnx-ify with that piece. (might be able to simplify the embedding models too?)

common/src/main/java/org/opensearch/ml/common/FunctionName.java
HenryL27: @ylwu-amzn I think we just need you to make a decision here. fwiw I got chatGPT to vote too [cross encoder naming](https://platform.openai.com/playground/p/qpkL9gjOr8zKF7tDW8rzziUu?model=gpt-3.5-turbo&mode=chat)

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityTranslator.java
HenryL27: ok, if we use `torch.onnx.export` then onnx works just fine. I think the issue is that `transformers.convert_graph_to_onnx.convert` drops the classification head at the end of the model. I think this would also fix embedding models so we don't have to do the pooling by hand.

dhrubo-os: Cool, thanks for deep diving here. Could you please cut an issue [here](https://github.com/opensearch-project/opensearch-py-ml/issues) so that we can change this method to use `torch.onnx.export`. We should change that method as I remember `convert_graph_to_onnx` is going to be deprecated too.

HenryL27: https://github.com/opensearch-project/opensearch-py-ml/issues/347

common/src/main/java/org/opensearch/ml/common/FunctionName.java
austintlee: @dhrubo-os `re-ranking` is _not_ what the cross encoder algorithm does.  The output of the algorithm is, as you say, widely used for re-ranking purposes.

However, I agree that `text similarity` is not a uniquely identifying feature of the cross encoder.  But given the above options, I would choose 1 (1 and 3 are pretty much the same to me).  We can maybe throw in `lossless` to distinguish cross encoders from bi encoders, but I am not sure if that's 100% accurate.  I'm thinking of terms like "lossless", "HD", "hi-res", "slow", "rich".  Would adding one of those help?  Again, if we are choosing one, my vote is for 1.

dhrubo-os: Text similarity does seem quite generic. In our semantic search implementation, we are indeed performing text similarity analysis, right? This enum refers to either a function or an algorithm name. However, my interest lies more in segregating the functionality. For instance, currently, we're achieving this through the use of a cross-encoder, but it's pertinent to note that in the future, we might explore and integrate additional methods to enhance this capability. The crucial aspect to keep in mind is that this functionality is exclusively dedicated to the re-ranking process, irrespective of the underlying method we employ to achieve it.

Think about `Question-answering` use case, In QA, text similarity algorithms might be used to match a user's question with similar questions or to find passages in a text that are similar to the question. Is that case, will we fill QA also under this giant `Text similarity` umbrella?

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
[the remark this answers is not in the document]
ylwu-amzn: > protect us on the off-chance that AWS does what Elastic did a few years ago and switches out the license or something

No need to worry that actually, if you check Elastic, community can still use the open source version even they changed license, otherwise we can't do "elastic fork -> opensearch".  If you worry this part, I think we should suggest not adding Aryn to license header. What if some other company make a change to this file, they may also prefer to add their license. Then in future how can Aryn tell which company holds the license for this file? It looks not reasonable that not allowing other guys modifying your code, right?  

I guess the other reason for adding Aryn is to show the credit of this feature. We are planning to build some way to show attribution/appreciation for features from community contributors. For example, we can maintain a contribution list file like this

| Feature      | Description |  Contributor |
| ----------- | ----------- | ----------- |
| Conversation search      | This feature will...       | Austin (from Aryn.ai) Henry (from Aryn.ai) |

Any suggestion?

ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_similarity/TinyBERT-CE.zip
ylwu-amzn: @dhrubo-os  can help publish pretrained model

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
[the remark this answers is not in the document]
HenryL27: we already decided this with @sean-zheng-amazon and @mashah

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
```diff
@@ -243,6 +265,9 @@ public static MLInput parse(XContentParser parser, String inputAlgoName) throws
             ModelResultFilter filter = new ModelResultFilter(returnBytes, returnNumber, targetResponse, targetResponsePositions);
             inputDataSet = new TextDocsInputDataSet(textDocs, filter);
         }
+        if (algorithm == FunctionName.TEXT_SIMILARITY) {
```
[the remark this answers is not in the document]
HenryL27: @dhrubo-os jacoco is showing red for like, this entire section (despite my attempts to the contrary). I think what's happening is that the classLoader stuff at the top of this method preempts all the MLInput parsing logic and relegates it to the subclass. We should probly just remove this stuff, idk?

rawwar:
Adding it here again as I commented on a "Resolved comment". 
@HenryL27  , Can you please point to the discussions where this is decided and what exactly is the decision?

HenryL27:
> Adding it here again as I commented on a "Resolved comment". @HenryL27 , Can you please point to the discussions where [this](https://github.com/opensearch-project/ml-commons/pull/1615/files#r1411447361)(To add Aryn copyright in the header) is decided and what exactly is the decision?

sure, [this thread](https://github.com/opensearch-project/ml-commons/pull/1196#discussion_r1289386124)

rawwar:
> > Adding it here again as I commented on a "Resolved comment". @HenryL27 , Can you please point to the discussions where [this](https://github.com/opensearch-project/ml-commons/pull/1615/files#r1411447361)(To add Aryn copyright in the header) is decided and what exactly is the decision?
> 
> sure, [this thread](https://github.com/opensearch-project/ml-commons/pull/1196#discussion_r1289386124)

If I am contributing some code that is reusing or extending the functionality provided in the classes that Aryn contributed, do i have to keep attributing Aryn there as well?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
ylwu-amzn: @HenryL27 @dhrubo-os what's the final decision?

common/src/main/java/org/opensearch/ml/common/FunctionName.java
```diff
@@ -35,9 +39,12 @@ public static FunctionName from(String value) {
      * @return true for deep learning model.
      */
     public static boolean isDLModel(FunctionName functionName) {
-        if (functionName == TEXT_EMBEDDING || functionName == SPARSE_ENCODING || functionName == SPARSE_TOKENIZE) {
-            return true;
-        }
-        return false;
+        Set<FunctionName> dlmodels = new HashSet<>(Set.of(
```
ylwu-amzn, line 42: Create a static constant for DL models? No need to create a new HashSet for each `isDLModel` call.

common/src/main/java/org/opensearch/ml/common/FunctionName.java
HenryL27: `text similarity`

common/src/main/java/org/opensearch/ml/common/FunctionName.java
HenryL27: ack

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
```diff
@@ -157,6 +161,20 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
                             builder.field(TARGET_RESPONSE_POSITIONS_FIELD, targetPositions.toArray(new Integer[0]));
                         }
                     }
+                    break;
+                case TEXT_SIMILARITY:
```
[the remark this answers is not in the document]
HenryL27: ack

common/src/main/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInput.java
```diff
@@ -0,0 +1,116 @@
...
+                    break;
+                case QUERY_TEXT_FIELD: 
+                    queryText = parser.text();
+                default:
+                    parser.skipChildren();
+                    break;
+            }
+        }        
+        if(docs.isEmpty()) {
+            throw new IllegalArgumentException("no text docs");
```
dhrubo-os, line 108: [nit] what do you think about: `No text documents are provided`

common/src/main/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInput.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        if(queryText == null) {
+            throw new IllegalArgumentException("no query text");
```
dhrubo-os, line 111: [nit] what do you think about: `No text query is provided`

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,75 @@
...
+
+   String queryText;
+
+    @Builder(toBuilder = true)
+    public TextSimilarityInputDataSet(String queryText, List<String> textDocs) {
+        super(MLInputDataType.TEXT_SIMILARITY);
+        Objects.requireNonNull(textDocs);
+        Objects.requireNonNull(queryText);
+        if(textDocs.isEmpty()) {
+            throw new IllegalArgumentException("pairs must be nonempty");
```
dhrubo-os, line 49: [nit] can we add some more user-friendly exception message?

common/src/test/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInputTest.java
```diff
@@ -0,0 +1,134 @@
...
+    public void testParseJson_NoPairs_ThenFail() throws IOException {
+        String json = "{\"algorithm\":\"TEXT_SIMILARITY\",\"query_text\":\"today is sunny\",\"text_docs\":[]}";
+        XContentParser parser = XContentType.JSON.xContent()
+                .createParser(new NamedXContentRegistry(new SearchModule(Settings.EMPTY,
+                        Collections.emptyList()).getNamedXContents()), null, json);
+        parser.nextToken();
+
+        IllegalArgumentException e = assertThrows(IllegalArgumentException.class,
+            () -> MLInput.parse(parser, input.getFunctionName().name()));
+        assert (e.getMessage().equals("no text docs"));
```
dhrubo-os, line 119: Can we have another test for missing query text too?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityCrossEncoderModelTest.java
```diff
@@ -0,0 +1,291 @@
...
+    private Encryptor encryptor;
+
+    @Before
+    public void setUp() throws URISyntaxException {
+        mlCachePath = Path.of("/tmp/ml_cache" + UUID.randomUUID());
+        encryptor = new EncryptorImpl("m+dWmfmnNRiNlOdej/QelEkvMTyH//frS2TBeS2BP4w=");
+        mlEngine = new MLEngine(mlCachePath, encryptor);
+        model = MLModel
+            .builder()
+            .modelFormat(MLModelFormat.TORCH_SCRIPT)
```
dhrubo-os, line 92: Can we have another test for ONNX too?

common/src/main/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInput.java
HenryL27: ack

common/src/main/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInput.java
HenryL27: ack

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
HenryL27: ack

common/src/test/java/org/opensearch/ml/common/input/nlp/TextSimilarityMLInputTest.java
HenryL27: sure

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_similarity/TextSimilarityCrossEncoderModelTest.java
HenryL27: sure

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,75 @@
...
+
+   String queryText;
+
+    @Builder(toBuilder = true)
+    public TextSimilarityInputDataSet(String queryText, List<String> textDocs) {
+        super(MLInputDataType.TEXT_SIMILARITY);
+        Objects.requireNonNull(textDocs);
+        Objects.requireNonNull(queryText);
+        if(textDocs.isEmpty()) {
+            throw new IllegalArgumentException("No text documents provided");
```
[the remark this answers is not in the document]
HenryL27: whoops. TinyBERT-CE.zip is a pt model so I'm naming it to TinyBERT-CE-pt.zip (to test both frameworks we need a model in each)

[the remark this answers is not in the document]
dhrubo-os: pt is the torchscript, right?

dhrubo-os:
Thanks for working on this. Approved.

dhrubo-os (approved):

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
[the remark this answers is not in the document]
austintlee: This was decided prior to the 2.10 release as part of a PR that we contributed.  I believe the initial discussion was done in that PR, but was later concluded over a call.

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
```diff
@@ -0,0 +1,75 @@
...
+        for(int i = 0; i < size; i++) {
+            String context = in.readString();
+            this.textDocs.add(context);
+        }
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(queryText);
```
austintlee, line 69: What happens if `queryText` is an empty string?  Does reading it back (in.readString) work as expected?

austintlee (approved):
LGTM (with one minor question.  You can answer and resolve.)

common/src/main/java/org/opensearch/ml/common/dataset/TextSimilarityInputDataSet.java
HenryL27: yep, you get an empty string back out

PR #1625, opened 2023-11-14
hidden_model

dhrubo-os:
### Description
[Creating hidden model functionalities

Tested all the functionalities. 

Test is not covered yet. Will address test after unblocking others.
]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
@@ -15,6 +15,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.AccessMode;
 import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.model.MLModelConfig;
 import org.opensearch.ml.common.model.MLModelFormat;
@@ -27,6 +28,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
+import java.util.Objects;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.connector.Connector.createConnector;
@@ -83,6 +85,8 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     private AccessMode accessMode;
     private Boolean doesVersionCreateModelGroup;
 
+    private Boolean isHidden;
+
     @Builder(toBuilder = true)
     public MLRegisterModelInput(FunctionName functionName,
                                 String modelName,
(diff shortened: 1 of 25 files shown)


common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
```diff
@@ -263,23 +267,23 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field(URL_FIELD, url);
         }
         if (hashValue != null) {
-            builder.field(HASH_VALUE_FIELD, hashValue);
+            builder.field(MLModel.MODEL_CONTENT_HASH_VALUE_FIELD, hashValue);
         }
         if (modelFormat != null) {
-            builder.field(MODEL_FORMAT_FIELD, modelFormat);
+            builder.field(MLModel.MODEL_FORMAT_FIELD, modelFormat);
```
ylwu-amzn, line 273: Why use MLModel fields rather than the filed in this class?

dhrubo-os: Reuse the same variable?

plugin/src/main/java/org/opensearch/ml/action/handler/MLSearchHandler.java
```diff
@@ -96,6 +97,17 @@ public void search(SearchRequest request, ActionListener<SearchResponse> actionL
                 Optional.ofNullable(request.source()).map(SearchSourceBuilder::fetchSource).map(FetchSourceContext::includes).orElse(null),
                 excludes.toArray(new String[0])
             );
+
+            // Check if the original query is not null before adding it to the must clause
```
ylwu-amzn, line 101: Is this a todo?

plugin/src/main/java/org/opensearch/ml/action/handler/MLSearchHandler.java
```diff
@@ -96,6 +97,17 @@ public void search(SearchRequest request, ActionListener<SearchResponse> actionL
...
+            BoolQueryBuilder queryBuilder = QueryBuilders.boolQuery();
```
ylwu-amzn, line 102: Could the `queryBuilder` be null?

plugin/src/main/java/org/opensearch/ml/action/handler/MLSearchHandler.java
dhrubo-os: no, it was the explanation why did I do that.

ylwu-amzn: Got it, this is a comment for line 103

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
ylwu-amzn: Check if the variable value is the same with this class

ylwu-amzn: If reuse, we can remove the duplicate variable. Let's keep one PR for one thing?

dhrubo-os: I just replaced where variable value is the same with this class.

dhrubo-os: >If reuse, we can remove the duplicate variable. Let's keep one PR for one thing? Refactoring is not highest priority for now.

Sure, reverted the changes.

ylwu-amzn (approved):

PR #1627, opened 2023-11-14
Cluster restart model auto redeploy

zane-neo:
### Description
Support cluster level restart model auto redeploy.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1641
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterManagerEventListener.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterManagerEventListener.java
@@ -14,6 +14,7 @@ import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.lifecycle.LifecycleListener;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.ml.autoredeploy.MLModelAutoReDeployer;
 import org.opensearch.ml.engine.encryptor.Encryptor;
 import org.opensearch.ml.indices.MLIndicesHandler;
@@ -22,7 +23,6 @@ import org.opensearch.threadpool.ThreadPool;
 
 import lombok.extern.log4j.Log4j2;
 
-import java.util.Arrays;
 import java.util.List;
 
 @Log4j2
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -685,6 +686,7 @@ public List<Setting<?>> getSettings() {
                 MLCommonsSettings.ML_COMMONS_ENABLE_INHOUSE_PYTHON_MODEL,
                 MLCommonsSettings.ML_COMMONS_MODEL_AUTO_REDEPLOY_ENABLE,
                 MLCommonsSettings.ML_COMMONS_MODEL_AUTO_REDEPLOY_LIFETIME_RETRY_TIMES,
+                MLCommonsSettings.ML_COMMONS_MODEL_AUTO_REDEPLOY_SUCCESS_RATIO,
```
Zhangxunmt, line 689: This setting is not used anywhere in this PR?

Zhangxunmt (commented):
Is the flag plugins.ml_commons.model_auto_redeploy.enable for cluster level or node level after this change?

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
zane-neo: Didn't expose this setting when implementing the auto redeploy feature, so exposing it now to enable customer to change this ratio value.

plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterManagerEventListener.java
```diff
@@ -62,13 +70,28 @@ public MLCommonsClusterManagerEventListener(
...
+            if (syncModelRoutingCron == null) {
+                startSyncModelRoutingCron();
+            }
+        });
+        mlModelAutoReDeployer.setStartCronJobListener(listener);
+        String localNodeId = clusterService.localNode().getId();
+        threadPool
+            .schedule(
+                () -> mlModelAutoReDeployer.buildAutoReloadArrangement(List.of(localNodeId), localNodeId),
+                TimeValue.timeValueSeconds(jobInterval),
```
ylwu-amzn, line 87: Can we start auto reload immediately ?

zane-neo: No, cluster manager ready doesn't mean cluster is ready, if we start auto reload immediately we can get a `cluster not initialized` etc exception when querying needs reload models.

zane-neo:
> Is the flag plugins.ml_commons.model_auto_redeploy.enable for cluster level or node level after this change?

Both level, this change is more like an enhancement to model auto redeploy to support cluster level.

ylwu-amzn (approved):

dhrubo-os:
Just curious to know if this will also re-deploy the hidden model (what we are building now? Currently only domain manager with having the admin SSL certificate can deploy the hidden model.

Zhangxunmt (approved):

ylwu-amzn:
> Just curious to know if this will also re-deploy the hidden model (what we are building now? Currently only domain manager with having the admin SSL certificate can deploy the hidden model.

I think we have no way to support hidden model auto redeployment, domain manager should build some logic outside of cluster.

PR #1636, opened 2023-11-15
Refactor tools factory implementation to a common factory class

arjunkumargiri:
### Description
Refactor tools factory implementation to a common factory class. Refactored to move away from each tool having to implement a factory to constructor based object creation. When a tool gets added it needs to be registered with `ToolsFactory` class. Any external tools will be automatically be added to `ToolsFactory`

Updated `Tool` interface for  `run()` and `validate()` method to include ToolSpec parameters. This helps to create a single instance of tool and reuse it for all tool executions. Given that the factory will maintain a single tool instance, each tool needs to be thread safe. To achieve the same, VectorDBTool is refactored to be thread safe by moving class variables to `run` method input. To create a new tool `run` and `validate` are the only mandatory methods that needs to be implemented and boilerplate functionalities are implemented through `AbstractTool`. 

### Check List
  - [x] All tests pass

### Testing
Added unit tests to validate that factory method will return tool instance.

Note: Build is failing due to existing integration test failures in memory package.


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class ToolsFactory {
+    private static final Map<String, Tool> AGENT_TOOLS = new HashMap<>();
+
+    public ToolsFactory(Client client, ScriptService scriptService, ClusterService clusterService,
+                        NamedXContentRegistry xContentRegistry, Map<String, Tool> externalTools) {
+        AGENT_TOOLS.put(MLModelTool.TYPE, new MLModelTool(client));
```
ylwu-amzn, line 20: Do we only have 1 instance for each tool?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
```diff
@@ -0,0 +1,37 @@
...
+        AGENT_TOOLS.put(CatIndexTool.TYPE, new CatIndexTool(client, clusterService));
+        AGENT_TOOLS.put(VectorDBTool.TYPE, new VectorDBTool(client, xContentRegistry));
+        AGENT_TOOLS.put(PainlessScriptTool.TYPE, new PainlessScriptTool(client, scriptService));
+        AGENT_TOOLS.putAll(externalTools);
+
+    }
+
+    public Tool getTool(final String toolName) {
+        return  Objects.requireNonNull(AGENT_TOOLS.get(toolName), "Unsupported tool: " + toolName);
+    }
```
Hailong-am, line 31: The keys of AGENT_TOOLS are type, should we change the variable to type?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
arjunkumargiri: This depends upon tool implementation. For example, VectorDBTool was not thread safe but have updated the tool as part of this CR to be thread safe.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
arjunkumargiri: Will rename variable `toolName` -> `agentType`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLReActAgentRunner.java
```diff
@@ -159,19 +160,20 @@ private void runAgent(MLAgent mlAgent, Map<String, String> params, ActionListene
...
             }
-            if (!toolFactories.containsKey(toolSpec.getType())) {
-                listener.onFailure(new IllegalArgumentException("No tool factory found for " + toolSpec.getType()));
-                return;
-            }
-            Tool tool = toolFactories.get(toolSpec.getType()).create(toolParams);
-            tool.setName(toolSpec.getName());
+            try {
+                Tool tool = toolsFactory.getTool(toolSpec.getType());
+                tool.setName(toolSpec.getName());
```
Hailong-am, line 165: ToolSepc name could be null, need if condition check here to avoid override to null.

arjunkumargiri: Added null check in `getTool` function.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractTool.java
```diff
@@ -0,0 +1,77 @@
+package org.opensearch.ml.engine.tools;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.Tool;
+
+import java.util.Map;
+
+public abstract class AbstractTool implements Tool {
```
Hailong-am, line 10: It's better move this class to spi module, so that tools in plugin own repository still can re-use it

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
Hailong-am: It's better mentioned in comments, each plugin need to be thread safe.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractTool.java
arjunkumargiri: Good suggestion, done!

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
arjunkumargiri: Added documentation to Tool and Abstract tool interfaces.

Hailong-am (commented):
Have abstractTool class looks good to me. To refactor tools factory and force tools need to thread safe, that would better have feedback from @ylwu-amzn

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
```diff
@@ -9,7 +9,7 @@
 import java.util.Map;
 
 /**
- * General tool interface.
+ * General tool interface. Tool implementation needs to be thread safe.
```
ylwu-amzn, line 12: > Tool implementation needs to be thread safe.

This is a high requirement for tool developer. Before this refactoring , we don't require this as we create new Tool for each request. I would suggest keep it simpler for tool developer.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolsFactory.java
ylwu-amzn: Before this refactoring, we allow user to define multiple Tools with same type. For example, user can define tools like this
```
[
{
      "type": "VectorDBTool",
      "name": "opensearch_vectordb",
},
{
      "type": "VectorDBTool",
      "name": "aws_vectordb",
}
]
```

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
arjunkumargiri: Agreed that tool developers needs to be careful with maintaining global states  but Singleton design pattern is a standard design pattern and has the benefit of using less memory overhead. Tool implementation needs to be stateless and should only perform the operation of calling other plugins/opensearch to fetch required information. If tools could be stateless do you see any concerns of maintaining a singleton tool?

ylwu-amzn: Seems you missed this comment https://github.com/opensearch-project/ml-commons/pull/1636#discussion_r1406940721

PR #1644, opened 2023-11-15
Fix only superadmin can use update model api bug

b4sjoo:
### Description
Fix the update model api issue due to restoring context too early.
 
### Check List
  - [ ] All tests pass


plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -179,7 +179,7 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
             },
                 e -> actionListener
                     .onFailure(new MLResourceNotFoundException("Failed to find model to update with the provided model id: " + modelId))
-            ), () -> context.restore()));
+            ));
```
ylwu-amzn, line 182: Why remove this?

b4sjoo: We've restored the context in line 488, if we restore here it will cause big

ylwu-amzn: Suggest keep restore here. Then add stash context before running update request.

ylwu-amzn (approved):

Issue #1553, opened 2023-10-26
[BUG] Class Cast Exception when invoking Register API through MLClient


owaiskazi19:
**What is the bug?**
Getting a class cast exception as mentioned below using the [register](https://github.com/opensearch-project/ml-commons/blob/main/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java#L231) API of MLClient in our plugin. Two class loaders are getting loaded in the cluster

```
java.net.FactoryURLClassLoader @475f5672
java.net.FactoryURLClassLoader @320be73
```
whereas our plugin is importing `MLRegisterModelResponse` just once from the location 
```
jar:file:/home/ubuntu/OpenSearch/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-flow-framework/opensearch-ml-client-3.0.0.0-SNAPSHOT.jar!/org/opensearch/ml/common/transport/register/MLRegisterModelResponse.class
```

Code:
```java
MLRegisterModelInput mlInput = MLRegisterModelInput.builder()
                .functionName(functionName)
                .modelName(modelName)
                .description(description)
                .connectorId(connectorId)
                .build();

mlClient.register(mlInput, actionListener);

```

COMPLETE ERROR:

```
[2023-10-25T20:13:30,038][ERROR][o.o.f.t.ProvisionWorkflowTransportAction] [ip-172-31-56-214] Provisioning failed for workflow rlt4aIsB-BC8dYg-Eww9 : java.util.concurrent.ExecutionException: org.opensearch.flowframework.exception.FlowFrameworkException: class org.opensearch.ml.common.transport.register.MLRegisterModelResponse cannot be cast to class org.opensearch.ml.common.transport.register.MLRegisterModelResponse (org.opensearch.ml.common.transport.register.MLRegisterModelResponse is in unnamed module of loader java.net.FactoryURLClassLoader @475f5672; org.opensearch.ml.common.transport.register.MLRegisterModelResponse is in unnamed module of loader java.net.FactoryURLClassLoader @320be73)
```

Found a PR similar to this issue already addressed Class Cast Exception: https://github.com/opensearch-project/ml-commons/pull/127
**How can one reproduce the bug?**
1. Create a MachineLearningNodeClient Object and invoke register API with the required input.

**What is the expected behavior?**
Register API should return MLRegisterModelResponse without any exception.

**What is your host/environment?**
 - OS: ubuntu
[...]

PR #1647, opened 2023-11-16
Register agent API support for MLClient

arjunkumargiri:
### Description
Register agent API support for MLClient. This will allow flow framework to integrate with register agent API.
 

### Check List
  - [x] All tests pass


client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -253,6 +256,16 @@ public void getTool(String toolName, ActionListener<ToolMetadata> listener) {
         client.execute(MLGetToolAction.INSTANCE, mlToolGetRequest, getMlGetToolResponseActionListener(listener));
     }
 
+    @Override
+    public void registerAgent(MLRegisterAgentRequest mlRegisterAgentRequest, ActionListener<MLRegisterAgentResponse> listener) {
+        client
+            .execute(
+                MLRegisterAgentAction.INSTANCE,
+                mlRegisterAgentRequest,
+                ActionListener.wrap(listener::onResponse, listener::onFailure)
```
ylwu-amzn, line 265: I think this way will fail if calling from AI flow side. Can you verify?

arjunkumargiri: Can you please provide more details on why this would fail?

@dbwiddis can you please check if this mechanism would fail?

dbwiddis: @arjunkumargiri haven't checked yet, but "probably".

See https://github.com/opensearch-project/ml-commons/pull/1616/files for the pattern to use to avoid using different class loaders for the Action listener.

CC: @owaiskazi19 @joshpalis

joshpalis: I would suggest following this [commit](https://github.com/opensearch-project/ml-commons/commit/882246c04daacc98eb9c72fbdcabcce497856423) instead, since the `GetMLTask` API responds with an `MLTask` rather than a response object

dbwiddis: > Can you please provide more details on why this would fail?

See https://github.com/opensearch-project/ml-commons/issues/1553 for reference and other examples.

I'm not sure of the exact inner workings of the JVM causing this, and it may have to do with which order the plugins are loaded on bootstrap, but the pattern you use can cause the Response class to be loaded from a different class loader than the one declared in flow framework which is calling this method, and somehow creating a separate method manages to fix it.

TLDR: in `client.execute()` use a getter method to get the action listener instead of declaring it inline.

arjunkumargiri: I would prefer to return the response object as it is possible that other attributes are added to response object in future. Do we know what flow framework is loading duplicate jars which is leading to class cast exception?

owaiskazi19: @arjunkumargiri replied to the issue. Also, yes this would fail on flow framework side and to avoid we should follow the similar pattern as this PR: https://github.com/opensearch-project/ml-commons/pull/1560/files

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -337,4 +339,22 @@ default ActionFuture<ToolMetadata> getTool(String toolName) {
      * @param listener action listener
      */
     void getTool(String toolName, ActionListener<ToolMetadata> listener);
+
+    /**
+     * Registers new agent and returns ActionFuture.
+     * @param mlRegisterAgentRequest Register agent input, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/#register-agent
+     * @return the result future
+     */
+    default ActionFuture<MLRegisterAgentResponse> registerAgent(MLRegisterAgentRequest mlRegisterAgentRequest) {
```
owaiskazi19, line 348: Based on the previous patterns of the APIs, we generally create the request in `MachineLearningNodeClient`
```suggestion
    default ActionFuture<MLRegisterAgentResponse> registerAgent(MLAgent mlAgent) {
```

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -253,6 +256,16 @@ public void getTool(String toolName, ActionListener<ToolMetadata> listener) {
         client.execute(MLGetToolAction.INSTANCE, mlToolGetRequest, getMlGetToolResponseActionListener(listener));
     }
 
+    @Override
+    public void registerAgent(MLRegisterAgentRequest mlRegisterAgentRequest, ActionListener<MLRegisterAgentResponse> listener) {
+        client
```
owaiskazi19, line 261: We can create a `MLRegisterAgentRequest` request here by using the `MLAgent` passed from the `MachineLearningClient`

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
arjunkumargiri: Updated PR to unblock flow integration. However, it will be good to identify why duplicate jars are pulled into flow framework?

owaiskazi19 (approved):
LGTM

ylwu-amzn (approved):
LGTM. Thanks a lot for the change

ylwu-amzn:
BTW, the Github CI failed, maybe need to rebase ?

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
dbwiddis: > Updated PR to unblock flow integration. However, it will be good to identify why duplicate jars are pulled into flow framework?

I think I answered this on another issue but I will repeat it here.
- there is no duplicate JAR, there are two different class loaders involved
- OpenSearch bootstraps plugins with their own class loader which is a child of the OpenSearch parent class loader.  This keeps plugin classloaders isolated from each other.
- so two different Java `.class` objects are created in memory representing the identical compiled Java class from the JAR.  When they are passed between plugins the JVM thinks they are being cast to each other.

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
arjunkumargiri: Done.

PR #1672, opened 2023-11-22
separate get Trace actions into a new rest API

Zhangxunmt:
### Description
Only return final results in the "List Conversation" API. Add a new API of "Get Trace" to return only traces for an interaction. 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesAction.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesAction.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.memory.action.conversation;
+
+import org.opensearch.action.ActionType;
+
+/**
+ * Action to return the traces associated with an interaction
+ */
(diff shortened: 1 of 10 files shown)


memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesAction.java
```diff
@@ -0,0 +1,23 @@
...
+import org.opensearch.action.ActionType;
+
+/**
+ * Action to return the traces associated with an interaction
+ */
+public class GetTracesAction extends ActionType<GetTracesResponse> {
+    /** Instance of this */
+    public static final GetTracesAction INSTANCE = new GetTracesAction();
+    /** Name of this action */
+    public static final String NAME = "cluster:admin/opensearch/ml/memory/trace/get";
```
ylwu-amzn, line 17: Trace is only for conversation agent, to make it general, maybe build a general search API for interaction data ?

Zhangxunmt: The general search API already exists. This "get trace" is only meant for the conversation agent to let FE use. The "list interaction" will only return interactions that have "null" trace_number (final response). In other agents, the "list interactions" API still works to list all interactions because the "trace_number" is null since it's only used in the conversational context.

Issue #1614, opened 2023-11-09
[FEATURE] Memory Enhancement to Support more LLM applications


Zhangxunmt:
In the Agent Framework, we want to enhance Ml-Commons to be able to interact with all kinds of LLMs like Claude, BedRock, etc. The use case of the agent framework will not be limited in the Search scenario, but also in Chatbot, Forecasting etc. To this end, the memory component needs to be extended and refactored to support more applications as a new data layer in Ml-Commons between the public APIs and system indices. This document focus on the design of this memory layer that supports Agent Framework in Ml-Commons.

### Architecture

A memory system needs to support two basic actions: reading and writing. Recall that every agent defines some core execution logic that expects certain inputs. Some of these inputs come directly from the user, but some of these inputs can come from memory. An agent will interact with its memory system twice in a given run.

1. AFTER receiving the initial user inputs but BEFORE executing the core logic, an agent will READ from its memory system and augment the user inputs.
2. AFTER executing the core logic but BEFORE returning the answer, an agent will WRITE the inputs and outputs of the current run to memory, so that they can be referred to in future runs.
<img width="934" alt="Screenshot 2023-11-01 at 6 37 11 PM" src="https://github.com/opensearch-project/ml-commons/assets/38896112/db228b0c-1887-4168-8906-e758b59f1b99">

As outlined in the high level Design, CRUD-like APIs will be added to the Memory system for two resources: Conversation and Interaction. Conversations are made up of interactions. An interaction represents a pair of messages: a human input and an artificial intelligence (AI) response. These restful APIs already exists in the conversational search. To support Agent Framework, the mapping and schema of the Conversation and Interaction will be updated with backward compatibility. 

The sequential flow of these APIs can be summarized as below. We plan to support three types of Query Algorithms.

* (P0) Most Recent Top N - store all interactions and return the most recent of them. The number of interactions returned is configurable.
* (P2) Summary - generate a summary of the most recent interactions. The user can configure the number of interactions, -1 means all history messages.
* (P2) VectorSearch - return the most relevant interactions by neural search.

![memory](https://github.com/opensearch-project/ml-commons/assets/38896112/13fa1ad9-e457-45ac-afbc-45e1453d7963)

### System Index and Mappings

[...]

PR #1678, opened 2023-11-22
rename the conversation/interaction indices to memory meta/message

Zhangxunmt:
### Description
Rename the conversation and interaction indices to memory meta/message based on the discussion in https://github.com/opensearch-project/ml-commons/issues/1614
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


austintlee:
Do we have a name for things we're storing in this new index other than interactions?

@navneet1v I think you suggested the name change.  Can you sign off on the new names of the indexes?

dhrubo-os:
fix spotless.

navneet1v:
> Do we have a name for things we're storing in this new index other than interactions?
> 
> @navneet1v I think you suggested the name change. Can you sign off on the new names of the indexes?

looks good from my side

austintlee:
@Zhangxunmt what else are we storing in the 'message' index?  Will there be a new resource type called 'message'?

dhrubo-os:
> > Do we have a name for things we're storing in this new index other than interactions?
> > @navneet1v I think you suggested the name change. Can you sign off on the new names of the indexes?
> 
> looks good from my side

We have 

> > Do we have a name for things we're storing in this new index other than interactions?
> > @navneet1v I think you suggested the name change. Can you sign off on the new names of the indexes?
> 
> looks good from my side

now index name: `plugins-ml-memory-message`

How much does it make sense to have `interaction_id` as a id field?
Same for `plugins-ml-memory-meta`, id field is `conversation_id` :)

 I hope in the long run, it doesn't get more confusing.

Zhangxunmt:
The spotless error is coming from the indexMapping Tool. @dhrubo-os 
The conversation_id and interaction_id still make sense because they align with the current name convention in the conversational search https://opensearch.org/docs/latest/ml-commons-plugin/conversational-search/, unless we are going to rename this feature. @dhrubo-os 

The change of the index name only reflects that this index is not only for conversation/interaction. It can be reused for other memory related data like a message (not necessary an interaction),  or anything that can be stored as history data. We don't have a concrete idea of what the "message" application_type looks like though. Let's assume this "message" scope is larger than "interaction"? @austintlee

austintlee (approved):
I am fine, but if this can wait a few days, I would also get a sign-off from @HenryL27 .

Zhangxunmt:
@HenryL27 , do you still have concerns for this name change?

PR #1701, opened 2023-11-28
update sorting from desc to asc for list interactions and traces

Zhangxunmt:
### Description
Update the default sort from DESC to ASC in listing interactions and traces, based on the requests from FE.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
@@ -176,7 +176,7 @@ public class MLMemoryManager {
         searchRequest.source(searchSourceBuilder);
 
         searchRequest.source().size(lastNInteraction);
-        searchRequest.source().sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.DESC);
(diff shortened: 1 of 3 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -176,7 +176,7 @@ void innerGetFinalInteractions(String conversationId, int lastNInteraction, Acti
         searchRequest.source(searchSourceBuilder);
 
         searchRequest.source().size(lastNInteraction);
-        searchRequest.source().sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.DESC);
+        searchRequest.source().sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.ASC);
```
Hailong-am, line 180: @Zhangxunmt this change may break `getFinalInteractions`, based on the parameter of this method, it fetch latest N interactions. After this change, it will fetch first N interactions.

Hailong-am: I would like to keep using DESC and make it sort as ASC in line 186 with `result.add(0, Interaction.fromSearchHit(hit));`

PR #1702, opened 2023-11-28
Add Delete Connector to MLClient

dbwiddis:
### Description

Adds MLClient method for delete connector
 
### Check List
  - [x] All tests pass

### DIFF
--- a/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
+++ b/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
@@ -36,6 +36,8 @@ import org.opensearch.ml.common.transport.MLTaskResponse;
 import org.opensearch.ml.common.transport.agent.MLRegisterAgentAction;
 import org.opensearch.ml.common.transport.agent.MLRegisterAgentRequest;
 import org.opensearch.ml.common.transport.agent.MLRegisterAgentResponse;
+import org.opensearch.ml.common.transport.connector.MLConnectorDeleteAction;
+import org.opensearch.ml.common.transport.connector.MLConnectorDeleteRequest;
 import org.opensearch.ml.common.transport.connector.MLCreateConnectorAction;
 import org.opensearch.ml.common.transport.connector.MLCreateConnectorInput;
 import org.opensearch.ml.common.transport.connector.MLCreateConnectorRequest;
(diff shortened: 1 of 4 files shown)


dbwiddis:
Failing test unrelated to this PR:
```
Tests with failures:
 - org.opensearch.ml.action.prediction.PredictionITTests.testPredictionWithSearchInput_LogisticRegression
```

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
```diff
@@ -243,6 +245,14 @@ public void createConnector(MLCreateConnectorInput mlCreateConnectorInput, Actio
         client.execute(MLCreateConnectorAction.INSTANCE, createConnectorRequest, getMlCreateConnectorResponseActionListener(listener));
     }
 
+    @Override
+    public void deleteConnector(String connectorId, ActionListener<DeleteResponse> listener) {
```
dhrubo-os, line 249: I think this will create the class loader issue (two objects of the same class). Can we follow the same way like `createConnector`. Please check `getMlCreateConnectorResponseActionListener` as an example?

dbwiddis: This is different because it uses an OpenSearch class `DeleteResponse`.   See `SearchResponse` for example.

dbwiddis: The problem with trying to use the example you gave is that as an OpenSearch class it does not have the `fromActionResponse()` method implemented here in the local response class, so I'd have to create that response on OpenSearch.

However, I believe as a child class loader of the OpenSearch class loader is may not have the same problem as having the class in two different plugins.

I can possibly take the same serialize/deserialize approach internally, but would like to merge as is in order to be able to actually test this.

dhrubo-os: I see, approving then.

dhrubo-os (approved):

client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java
dbwiddis: As a folllow-up, I've implemented the delete connector step in Flow Framework and did not encounter a Class Cast exception, so my theory appears correct.  

```
[2023-11-28T17:43:36,976][INFO ][o.o.f.w.ProcessNode      ] [88665a43e3fa] Starting workflow_step_2.
[2023-11-28T17:43:37,021][INFO ][o.o.m.a.c.DeleteConnectorTransportAction] [88665a43e3fa] Completed Delete Connector Request, connector id:-v6-GIwBk0tyGR8G9Nm2 deleted
```

So for now this PR looks good.  I reserve the right to find out later that I'm wrong, but I suspect the serialize/deserialize workaround is only required when a class declared in MLClient is accessed from multiple plugins.

PR #1709, opened 2023-11-29
Support regenerate for execute API

Hailong-am:
### Description

Support regenerate on existing interaction, `regenerate_interaction_id` and `memory_id` must be provide for regenerate case.

```json
POST http://localhost:9200/_plugins/_ml/agents/MBWOSIwBAnQBpyKtmAkn/_execute
{
  "parameters": {
    "memory_id": "A5CgV4wBiufF8-BkxC0r",
    "regenerate_interaction_id": "BJCgV4wBiufF8-BkxC2b",
    "verbose": false
  }
}
```

Response 
```json
{
    "inference_results": [
        {
            "output": [
                {
                    "name": "response",
                    "dataAsMap": {
                        "response": "Your OpenSearch cluster contains many security audit log indices from late 2023, with most replicas in a yellow status. The cluster and index storage sizes and document counts appear normal.",
                        "additional_info": {}
                    }
                },
                {
                    "name": "QuestionSuggestor",
                    "result": "[\"What are some ways I can analyze these security audit logs further?\",\"Can you help me understand why the replicas are in yellow status?\"]"
                }
            ]
        }
    ]
}
```

 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -106,16 +108,31 @@ public void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Obje
         List<MLToolSpec> toolSpecs = mlAgent.getTools();
         String memoryType = mlAgent.getMemory().getType();
         String memoryId = params.get(MEMORY_ID);
+        // for regenerate, original interaction id must provide
+        String regenerateInteractionId = params.get(REGENERATE_INTERACTION_ID);
```
arjunkumargiri, line 112: Can interaction regeneration logic be handled in MLAgenExecutor instead of Chat agent? In future there could be new agent types and we should not re-implement regeneration logic in each agent type

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -106,16 +108,31 @@ public void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Obje
...
                 Iterator<Interaction> iterator = r.iterator();
+                boolean interactionIdFound = false;
                 while (iterator.hasNext()) {
                     Interaction next = iterator.next();
+                    if (next.getId().equals(regenerateInteractionId)) {
+                        log.info("Regenerate for existing interaction {}", regenerateInteractionId);
+                        interactionIdFound = true;
+                        // if no new question provided, use original question
+                        params.computeIfAbsent(QUESTION, key -> next.getInput());
+                        // there may have other new interactions happened after this interaction
```
arjunkumargiri, line 133: Will question be specified in case of re-generation request? if so why would it be different from previous interaction question?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -286,4 +289,47 @@ public void updateInteraction(String interactionId, Map<String, Object> updateCo
...
+     * Delete interaction with its trace data
+     * @param interactionId interaction id
+     * @param listener callback for delete result
+     */
+    public void deleteInteraction(String interactionId, ActionListener<Boolean> listener) {
+        BulkRequest bulkRequest = new BulkRequest(indexName);
+        bulkRequest.add(new DeleteRequest(indexName, interactionId));
+
+        innerGetTraces(interactionId, ActionListener.wrap(traces -> {
+            traces.forEach(trace -> bulkRequest.add(new DeleteRequest(indexName, trace.getId())));
```
arjunkumargiri, line 303: Instead of bulk request, can we use delete_by_query to reduce number of operations?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
Hailong-am: For current UI design, question will not been passed as parameter when regenerate, it will use original one. But think what chatGPT and other chatbot does, user could edit the question and regenerate, to support that as well, adding this logic here.

if user don't have specified the question, use original one, otherwise use new one.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
Hailong-am: @arjunkumargiri Instead of put logic in `MLAgenExecutor`, we can have a base class for all kinds of agent runner, put common logic in that class would make more sense.

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Hailong-am: make sense, updated to `delete_by_query`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
arjunkumargiri: Agreed, makes sense.

Hailong-am: @arjunkumargiri Since the root interaction been moved to `MLAgentRunner`,regenerate is coupled with root interaction, move the logic to that class also.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,41 +122,51 @@ public void execute(Input input, ActionListener<Output> listener) {
...
+                                        log.info("Regenerate for existing interaction {}", regenerateInteractionId);
+                                        getQuestionFromInteraction(memoryId, interactionId, inputDataSet);
+                                    });
+
                                     // Create root interaction ID
                                     ConversationIndexMessage msg = ConversationIndexMessage
                                         .conversationIndexMessageBuilder()
                                         .type(appType)
-                                        .question(question)
+                                        .question(inputDataSet.getParameters().get(QUESTION))
```
xluo-aws, line 152: Why not use question?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,41 +122,51 @@ public void execute(Input input, ActionListener<Output> listener) {
...
                                 ConversationIndexMemory.Factory conversationIndexMemoryFactory =
                                     (ConversationIndexMemory.Factory) memoryFactoryMap.get(memorySpec.getType());
-                                conversationIndexMemoryFactory.create(question, memoryId, appType, ActionListener.wrap(memory -> {
+                                conversationIndexMemoryFactory.create(question, memoryId, appType, wrap(memory -> {
                                     inputDataSet.getParameters().put(MEMORY_ID, memory.getConversationId());
-                                    ConversationIndexMemory conversationIndexMemory = (ConversationIndexMemory) memory;
+
+                                    // get regenerate interaction question
+                                    Optional.ofNullable(regenerateInteractionId).ifPresent(interactionId -> {
+                                        log.info("Regenerate for existing interaction {}", regenerateInteractionId);
```
xluo-aws, line 144: Use interactionId is more readable?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -156,6 +175,8 @@ public void execute(Input input, ActionListener<Output> listener) {
                                 ActionListener<Object> agentActionListener = createAgentActionListener(listener, outputs, modelTensors);
                                 executeAgent(inputDataSet, mlAgent, agentActionListener);
                             }
+                        } catch (Exception ex) {
+                            listener.onFailure(ex);
```
xluo-aws, line 179: Log the ex?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: question will be override at line 204 for regenerate case

xluo-aws: Could you add comment for this specific regenerate case? It takes a while to figure out why we can't use the question in request directly

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -119,6 +123,10 @@ public void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Obje
                 Iterator<Interaction> iterator = r.iterator();
                 while (iterator.hasNext()) {
                     Interaction next = iterator.next();
+                    if (next.getId().equals(regenerateInteractionId)) {
+                        // there may have other new interactions happened after this interaction
```
xluo-aws, line 127: Could you update the comment to be more clear?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: thanks, comments added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: would like be consistent with parameter name

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: thanks, error log added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
Hailong-am: sure. comments added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,41 +122,52 @@ public void execute(Input input, ActionListener<Output> listener) {
                             MLMemorySpec memorySpec = mlAgent.getMemory();
                             String memoryId = inputDataSet.getParameters().get(MEMORY_ID);
                             String parentInteractionId = inputDataSet.getParameters().get(PARENT_INTERACTION_ID);
+                            String regenerateInteractionId = inputDataSet.getParameters().get(REGENERATE_INTERACTION_ID);
                             String appType = mlAgent.getAppType();
                             String question = inputDataSet.getParameters().get(QUESTION);
 
+                            if (memoryId == null && regenerateInteractionId != null) {
+                                listener.onFailure(new MLValidationException("Memory id must provide for regenerate"));
+                            }
```
jngz-es, line 131: Missed a return express here?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,41 +122,52 @@ public void execute(Input input, ActionListener<Output> listener) {
...
+
                             if (memorySpec != null
                                 && memorySpec.getType() != null
                                 && memoryFactoryMap.containsKey(memorySpec.getType())
                                 && (memoryId == null || parentInteractionId == null)) {
                                 ConversationIndexMemory.Factory conversationIndexMemoryFactory =
                                     (ConversationIndexMemory.Factory) memoryFactoryMap.get(memorySpec.getType());
-                                conversationIndexMemoryFactory.create(question, memoryId, appType, ActionListener.wrap(memory -> {
+                                conversationIndexMemoryFactory.create(question, memoryId, appType, wrap(memory -> {
```
jngz-es, line 139: Keep the existing style of using ActionListener.wrap()? You only changed this wrap, but there are still some ActionListener.wrap() and ActionListener.* somewhere else in this file.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -169,6 +192,24 @@ public void execute(Input input, ActionListener<Output> listener) {
...
+     * @param memoryId conversation id
+     * @param interactionId interaction id
+     * @param inputDataSet input parameters
+     */
+    private void getQuestionFromInteraction(String memoryId, String interactionId, RemoteInferenceInputDataSet inputDataSet) {
+        try {
+            PlainActionFuture<GetInteractionResponse> future = PlainActionFuture.newFuture();
+            client.execute(GetInteractionAction.INSTANCE, new GetInteractionRequest(memoryId, interactionId), future);
+
+            GetInteractionResponse interactionResponse = future.get();
```
jngz-es, line 206: I don't think we should have any blocking invocation in an opensearch thread. Can you change to non-blocking way like listener?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
jngz-es: Yeah, I was confused too. I think you can also set question to the original question meanwhile updating inputDataSet.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: Good catch, will add return here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: refactor to listener way and it's more clear now.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: this is unintentional change, change it back.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
Hailong-am: sure, changed it to non-blocking way.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,41 +122,52 @@ public void execute(Input input, ActionListener<Output> listener) {
                             MLMemorySpec memorySpec = mlAgent.getMemory();
                             String memoryId = inputDataSet.getParameters().get(MEMORY_ID);
                             String parentInteractionId = inputDataSet.getParameters().get(PARENT_INTERACTION_ID);
+                            String regenerateInteractionId = inputDataSet.getParameters().get(REGENERATE_INTERACTION_ID);
                             String appType = mlAgent.getAppType();
                             String question = inputDataSet.getParameters().get(QUESTION);
 
+                            if (memoryId == null && regenerateInteractionId != null) {
+                                listener.onFailure(new MLValidationException("Memory id must provide for regenerate"));
```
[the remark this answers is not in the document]
Hailong-am: `regenerate` means user want LLM to reprocess the chat question and generate answer, a picture may more clear

image

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -113,9 +116,14 @@ public void execute(Input input, ActionListener<Output> listener) {
                             MLMemorySpec memorySpec = mlAgent.getMemory();
                             String memoryId = inputDataSet.getParameters().get(MEMORY_ID);
                             String parentInteractionId = inputDataSet.getParameters().get(PARENT_INTERACTION_ID);
+                            String regenerateInteractionId = inputDataSet.getParameters().get(REGENERATE_INTERACTION_ID);
                             String appType = mlAgent.getAppType();
                             String question = inputDataSet.getParameters().get(QUESTION);
 
+                            if (memoryId == null && regenerateInteractionId != null) {
```
dhrubo-os, line 123: What if both are null? Do we still want to continue?

Hailong-am: if `regenerateInteractionId` is null, it's a normal flow, memory will automatically create if needed

dhrubo-os (dismissed):

Hailong-am:
done in #1816

PR #1733, opened 2023-12-05
Fix sort on get trace API and flaky test of conversation

Hailong-am:
### Description

1. Fix sort on get trace Api, in the case two traces with create time are `"create_time": "2023-12-05T05:14:42.823172Z"` and `"create_time": "2023-12-05T05:14:42.823833Z"` they are equally same date due to mapping field of [`create_time`](https://github.com/opensearch-project/ml-commons/blob/feature/agent_framework_dev/common/src/main/java/org/opensearch/ml/common/conversation/ConversationalIndexConstants.java#L97) is `date` not `date_nanos`, either we change type to `date_nanos` or add one more sort field.

2. Add refresh policy to update interaction api so that get conversation api will able to fetch latest data.

3. fix flaky test for conversation

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
@@ -337,7 +337,10 @@ public class InteractionsIndex {
 
         request.source(searchSourceBuilder);
         request.source().from(from).size(maxResults);
-        request.source().sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.ASC);
+        request
+            .source()
+            .sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.ASC)
+            .sort(ConversationalIndexConstants.INTERACTIONS_TRACE_NUMBER_FIELD, SortOrder.ASC);
         try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
             ActionListener<List<Interaction>> internalListener = ActionListener.runBefore(listener, () -> threadContext.restore());
             ActionListener<SearchResponse> al = ActionListener.wrap(response -> {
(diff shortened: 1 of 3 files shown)


Hailong-am:
@HenryL27 @Zhangxunmt please help to review

memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
```diff
@@ -337,7 +337,10 @@ public void getTraces(String interactionId, int from, int maxResults, ActionList
 
         request.source(searchSourceBuilder);
         request.source().from(from).size(maxResults);
-        request.source().sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.ASC);
+        request
+            .source()
+            .sort(ConversationalIndexConstants.INTERACTIONS_CREATE_TIME_FIELD, SortOrder.ASC)
```
ylwu-amzn, line 342: Do we still need sort by create time if we add sort by trace number ?

Hailong-am: make sense, trace number should enough, remove create_time sort.

Zhangxunmt (approved):
lgtm. This sorting may need to apply to "get Interactions" as well.

Hailong-am:
> lgtm. This sorting may need to apply to "get Interactions" as well.

@Zhangxunmt there is no other sort key than create_time for interactions, do you mean we update the mappings of `create_time` filed to `date_nanos`?

Zhangxunmt:
@Hailong-am , I mean the problem you described ideally applies to all docs that are sorted based on "date" field type, which includes conversation and interactions. However this race condition may be extremely rare for conversation/interaction because they are human-generated, so what you have changed in using the trace_number should be just enough.

ylwu-amzn (approved):

PR #1742, opened 2023-12-07
use truncate flag in huggingface tokenizer

HenryL27:
### Description
Turns on the truncation flag for the tokenizer used for nlp tasks. If the generated encoding is too long, there can be problems (index out of bounds when looking up roberta positional embeddings).
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/SentenceTransformerTranslator.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/SentenceTransformerTranslator.java
@@ -25,7 +25,12 @@ public abstract class SentenceTransformerTranslator implements ServingTranslator
     @Override
     public void prepare(TranslatorContext ctx) throws IOException {
         Path path = ctx.getModel().getModelPath();
-        tokenizer = HuggingFaceTokenizer.builder().optPadding(true).optTokenizerPath(path.resolve("tokenizer.json")).build();
+        tokenizer = HuggingFaceTokenizer
+            .builder()
+            .optPadding(true)
+            .optTruncation(true)
+            .optTokenizerPath(path.resolve("tokenizer.json"))
+            .build();
     }
 
     @Override


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/SentenceTransformerTranslator.java
```diff
@@ -25,7 +25,12 @@ public Batchifier getBatchifier() {
     @Override
     public void prepare(TranslatorContext ctx) throws IOException {
         Path path = ctx.getModel().getModelPath();
-        tokenizer = HuggingFaceTokenizer.builder().optPadding(true).optTokenizerPath(path.resolve("tokenizer.json")).build();
+        tokenizer = HuggingFaceTokenizer
+            .builder()
+            .optPadding(true)
+            .optTruncation(true)
```
ylwu-amzn, line 31: Why add this?

dhrubo-os: @HenryL27 Isn't it same when we add truncation information in the tokenizer.json file like this:

```
"truncation": {
    "direction": "Right",
    "max_length": 128,
    "strategy": "LongestFirst",
    "stride": 0
  },
  ```
  
 Can you please share your tokenizer file you are using?

Comment (HenryL27) [reply ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/SentenceTransformerTranslator.java:31]: If I don't and the context is too long, I get sometimes index-out-of-bounds when using a roberta-based model to infer (there are 514 learned positional embeddings so if you go over the djl engine breaks)

Comment (HenryL27) [reply ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/SentenceTransformerTranslator.java:31]: you're right, I've got 
```
"truncation": null
```

dhrubo-os: Yup, which is why we added [this method ](https://github.com/opensearch-project/opensearch-py-ml/blob/main/opensearch_py_ml/ml_models/sentencetransformermodel.py#L735-L763) to update the tokenizer.json file for our pre-trained model. 

I feel like in future, we should also remove this padding from here, as that info should also come from tokenizer.json file.

HenryL27: why shouldn't we do this though?

dhrubo-os: what if customer doesn't want the truncation? As a platform, how can we support this?

HenryL27:
closing this as the tokenizer file handles it (as long as it's compiled correctly :)

PR #1751, opened 2023-12-12
Update tool selction prompt

Hailong-am:
### Description
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -334,8 +334,8 @@ private void runReAct(
                     Map<String, ?> dataAsMap = tmpModelTensorOutput.getMlModelOutputs().get(0).getMlModelTensors().get(0).getDataAsMap();
 
                     String thought = dataAsMap.get("thought") + "";
-                    String action = dataAsMap.get("action") + "";
-                    String actionInput = dataAsMap.get("action_input") + "";
+                    String action = dataAsMap.get("tool") + "";
```
ylwu-amzn, line 337: Why have to change this part ? Is this causing any failure?

ylwu-amzn: Thought/action is standard word, don't suggest change this part

Hailong-am: Action may lead misunderstanding to LLM, it could means the detailed action to take, not the tool to use.

Hailong-am: During the integration testing, we found sometimes LLM could not select correct tool, for example, 
```json
{
    "response": "no access to this tool ",
    "origin": "split the opensearch_dashboards_sample_data_ecommerce index into multiple smaller indexes based on a field",
}
```
origin here means the action LLM response to agent framework, it's more like a detailed action we need to take. Change prompt from `action to take` to `tool to use` and name from `action` to `tool` may help on tool selection.

ylwu-amzn: hi, Hailong, have you verified this change will help on the tool selection? I don't think we should change the word, Action is some standard word used by langchain too. Can we use other way like fine tune prompt.

xluo-aws: We have tuned prompt but seems not very effective, this is the most effective solution we found so far. We can put it on hold if anyone else has a better solution. Choose an invalid string as action is the major issue we found so far during testing.

Hailong-am: close it, it will handled by https://github.com/opensearch-project/ml-commons/pull/1813

PR #1593, opened 2023-11-04
return parsing exception 400 for parsing errors

Zhangxunmt:
### Description
Fix the 500 internal service error reported in the pen test. All parsing errors should return 400 error code showing it's coming from users.

 
Test: 
~~~
PUT /_plugins/_ml/connectors/zmMCmIsBaumhjugAg9cu
{
  "description": null
}
Response:
{
  "error": {
    "root_cause": [
      {
        "type": "parse_exception",
        "reason": "Can't get text on a VALUE_NULL at 2:18"
      }
    ],
    "type": "parse_exception",
    "reason": "Can't get text on a VALUE_NULL at 2:18"
  },
  "status": 400
}
~~~

~~~
PUT /_plugins/_ml/connectors/zmMCmIsBaumhjugAg9cu
{
  "description": "Update description"
}
Response:
{
  "_index": ".plugins-ml-connector",
  "_id": "zmMCmIsBaumhjugAg9cu",
  "_version": 3,
  "result": "updated",
  "_shards": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "_seq_no": 2,
  "_primary_term": 2
}
~~~

### Issues Resolved
[[List any issues this PR will resolve]](https://talos.security.aws.a2z.com/#/talos/engagement/arn:aws:talos-engagement:engagement/1224cf1a-48cb-4cfd-bd70-d170bbaa2b93/finding/8f543629-903f-4c57-a03d-df61b8b20601)
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1761, opened 2023-12-14
Memory Manager and Update Memory Actions/APIs

Zhangxunmt:
### Description
Add more actions in the memory layer for update Memories with 100% coverage. Memory Manager for Agents is also included. 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
@@ -29,6 +29,8 @@ public class ActionConstants {
     public final static String RESPONSE_CONVERSATION_LIST_FIELD = "conversations";
     /** name of list on interactions in all responses */
     public final static String RESPONSE_INTERACTION_LIST_FIELD = "interactions";
+    /** name of list on interactions in all responses */
+    public final static String RESPONSE_TRACES_LIST_FIELD = "traces";
     /** name of interaction Id field in all responses */
     public final static String RESPONSE_INTERACTION_ID_FIELD = "interaction_id";
 
@@ -56,20 +58,28 @@ public class ActionConstants {
     public final static String SUCCESS_FIELD = "success";
 
     private final static String BASE_REST_PATH = "/_plugins/_ml/memory/conversation";
+    private final static String BASE_REST_INTERACTION_PATH = "/_plugins/_ml/memory/interaction";
     /** path for create conversation */
     public final static String CREATE_CONVERSATION_REST_PATH = BASE_REST_PATH + "/_create";
     /** path for get conversations */
     public final static String GET_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/_list";
+    /** path for update conversations */
+    public final static String UPDATE_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/{conversation_id}/_update";
+    /** path for put interaction */
     /** path for create interaction */
     public final static String CREATE_INTERACTION_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_create";
     /** path for get interactions */
     public final static String GET_INTERACTIONS_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_list";
+    /** path for get interactions */
+    public final static String GET_TRACES_REST_PATH = "/_plugins/_ml/memory/trace" + "/{interaction_id}/_list";
     /** path for delete conversation */
     public final static String DELETE_CONVERSATION_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_delete";
     /** path for search conversations */
     public final static String SEARCH_CONVERSATIONS_REST_PATH = BASE_REST_PATH + "/_search";
     /** path for search interactions */
     public final static String SEARCH_INTERACTIONS_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_search";
+    /** path for update interactions */
+    public final static String UPDATE_INTERACTIONS_REST_PATH = BASE_REST_INTERACTION_PATH + "/{interaction_id}/_update";
     /** path for get conversation */
     public final static String GET_CONVERSATION_REST_PATH = BASE_REST_PATH + "/{conversation_id}";
     /** path for get interaction */
--- a/common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
@@ -29,7 +29,7 @@ public class ActionConstants {
     public final static String RESPONSE_CONVERSATION_LIST_FIELD = "conversations";
     /** name of list on interactions in all responses */
     public final static String RESPONSE_INTERACTION_LIST_FIELD = "interactions";
-    /** name of list on interactions in all responses */
+    /** name of list on traces in all responses */
     public final static String RESPONSE_TRACES_LIST_FIELD = "traces";
     /** name of interaction Id field in all responses */
     public final static String RESPONSE_INTERACTION_ID_FIELD = "interaction_id";
@@ -65,12 +65,11 @@ public class ActionConstants {
     public final static String GET_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/_list";
     /** path for update conversations */
     public final static String UPDATE_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/{conversation_id}/_update";
-    /** path for put interaction */
     /** path for create interaction */
     public final static String CREATE_INTERACTION_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_create";
     /** path for get interactions */
     public final static String GET_INTERACTIONS_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_list";
-    /** path for get interactions */
+    /** path for get traces */
     public final static String GET_TRACES_REST_PATH = "/_plugins/_ml/memory/trace" + "/{interaction_id}/_list";
     /** path for delete conversation */
     public final static String DELETE_CONVERSATION_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_delete";
--- a/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
@@ -20,6 +20,8 @@ import java.security.PrivilegedExceptionAction;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class StringUtils {
 
--- a/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
@@ -20,8 +20,6 @@ import java.security.PrivilegedExceptionAction;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 public class StringUtils {
 
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
@@ -19,6 +19,7 @@ import org.opensearch.ml.common.conversation.Interaction;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.NonNull;
 
 /**
  * Action Response for get traces for an interaction
@@ -26,6 +27,7 @@ import lombok.Getter;
 @AllArgsConstructor
 public class GetTracesResponse extends ActionResponse implements ToXContentObject {
     @Getter
+    @NonNull
     private List<Interaction> traces;
     @Getter
     private int nextToken;
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.memory.action.conversation;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.opensearch.core.action.ActionResponse;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.ToXContent;
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.ml.common.conversation.ActionConstants;
+import org.opensearch.ml.common.conversation.Interaction;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Action Response for get traces for an interaction
+ */
+@AllArgsConstructor
+public class GetTracesResponse extends ActionResponse implements ToXContentObject {
+    @Getter
+    private List<Interaction> traces;
+    @Getter
+    private int nextToken;
+    private boolean hasMoreTokens;
+
+    /**
+     * Constructor
+     * @param in stream input; assumes GetTracesResponse.writeTo was called
+     * @throws IOException if there's not a G.I.R. in the stream
+     */
+    public GetTracesResponse(StreamInput in) throws IOException {
+        super(in);
+        traces = in.readList(Interaction::fromStream);
+        nextToken = in.readInt();
+        hasMoreTokens = in.readBoolean();
+    }
+
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeList(traces);
+        out.writeInt(nextToken);
+        out.writeBoolean(hasMoreTokens);
+    }
+
+    /**
+     * Are there more pages in this search results
+     * @return whether there are more traces in this search
+     */
+    public boolean hasMorePages() {
+        return hasMoreTokens;
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params params) throws IOException {
+        builder.startObject();
+        builder.startArray(ActionConstants.RESPONSE_TRACES_LIST_FIELD);
+        for (Interaction trace : traces) {
+            trace.toXContent(builder, params);
+        }
+        builder.endArray();
+        if (hasMoreTokens) {
+            builder.field(ActionConstants.NEXT_TOKEN_FIELD, nextToken);
+        }
+        builder.endObject();
+        return builder;
+    }
+
+}
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.memory.action.conversation;
+
+import java.util.List;
+
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.client.Client;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.ml.common.conversation.Interaction;
+import org.opensearch.ml.memory.ConversationalMemoryHandler;
+import org.opensearch.ml.memory.index.OpenSearchConversationalMemoryHandler;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportService;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class GetTracesTransportAction extends HandledTransportAction<GetTracesRequest, GetTracesResponse> {
+    private Client client;
+    private ConversationalMemoryHandler cmHandler;
+
+    /**
+     * Constructor
+     * @param transportService for inter-node communications
+     * @param actionFilters for filtering actions
+     * @param cmHandler Handler for conversational memory operations
+     * @param client OS Client for dealing with OS
+     * @param clusterService for some cluster ops
+     */
+    @Inject
+    public GetTracesTransportAction(
+        TransportService transportService,
+        ActionFilters actionFilters,
+        OpenSearchConversationalMemoryHandler cmHandler,
+        Client client,
+        ClusterService clusterService
+    ) {
+        super(GetTracesAction.NAME, transportService, actionFilters, GetTracesRequest::new);
+        this.client = client;
+        this.cmHandler = cmHandler;
+    }
+
+    @Override
+    public void doExecute(Task task, GetTracesRequest request, ActionListener<GetTracesResponse> actionListener) {
+        int maxResults = request.getMaxResults();
+        int from = request.getFrom();
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().newStoredContext(true)) {
+            ActionListener<GetTracesResponse> internalListener = ActionListener.runBefore(actionListener, () -> context.restore());
+            ActionListener<List<Interaction>> al = ActionListener.wrap(tracesList -> {
+                internalListener.onResponse(new GetTracesResponse(tracesList, from + maxResults, tracesList.size() == maxResults));
+            }, e -> { internalListener.onFailure(e); });
+            cmHandler.getTraces(request.getInteractionId(), from, maxResults, al);
+        } catch (Exception e) {
+            log.error("Failed to get traces for conversation " + request.getInteractionId(), e);
+            actionListener.onFailure(e);
+        }
+    }
+}
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
@@ -10,7 +10,6 @@ import java.util.List;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.client.Client;
-import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.action.ActionListener;
@@ -33,15 +32,13 @@ public class GetTracesTransportAction extends HandledTransportAction<GetTracesRe
      * @param actionFilters for filtering actions
      * @param cmHandler Handler for conversational memory operations
      * @param client OS Client for dealing with OS
-     * @param clusterService for some cluster ops
      */
     @Inject
     public GetTracesTransportAction(
         TransportService transportService,
         ActionFilters actionFilters,
         OpenSearchConversationalMemoryHandler cmHandler,
-        Client client,
-        ClusterService clusterService
+        Client client
     ) {
         super(GetTracesAction.NAME, transportService, actionFilters, GetTracesRequest::new);
         this.client = client;
@@ -53,6 +50,7 @@ public class GetTracesTransportAction extends HandledTransportAction<GetTracesRe
         int maxResults = request.getMaxResults();
         int from = request.getFrom();
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().newStoredContext(true)) {
+            // TODO: check this newStoredContext() method and remove it if it's redundant
             ActionListener<GetTracesResponse> internalListener = ActionListener.runBefore(actionListener, () -> context.restore());
             ActionListener<List<Interaction>> al = ActionListener.wrap(tracesList -> {
                 internalListener.onResponse(new GetTracesResponse(tracesList, from + maxResults, tracesList.size() == maxResults));
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
@@ -6,7 +6,6 @@
 package org.opensearch.ml.memory.action.conversation;
 
 import static org.opensearch.action.ValidateActions.addValidationError;
-import static org.opensearch.ml.common.conversation.ConversationalIndexConstants.APPLICATION_TYPE_FIELD;
 import static org.opensearch.ml.common.conversation.ConversationalIndexConstants.META_NAME_FIELD;
 
 import java.io.ByteArrayInputStream;
@@ -36,7 +35,7 @@ public class UpdateConversationRequest extends ActionRequest {
     private String conversationId;
     private Map<String, Object> updateContent;
 
-    private static final Set<String> allowedList = new HashSet<>(Arrays.asList(META_NAME_FIELD, APPLICATION_TYPE_FIELD));
+    private static final Set<String> allowedList = new HashSet<>(Arrays.asList(META_NAME_FIELD));
 
     @Builder
     public UpdateConversationRequest(String conversationId, Map<String, Object> updateContent) {
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
@@ -14,6 +14,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -40,13 +41,13 @@ public class UpdateConversationRequest extends ActionRequest {
     @Builder
     public UpdateConversationRequest(String conversationId, Map<String, Object> updateContent) {
         this.conversationId = conversationId;
-        this.updateContent = filterMapContent(updateContent);
+        this.updateContent = filterUpdateContent(updateContent);
     }
 
     public UpdateConversationRequest(StreamInput in) throws IOException {
         super(in);
         this.conversationId = in.readString();
-        this.updateContent = filterMapContent(in.readMap());
+        this.updateContent = filterUpdateContent(in.readMap());
     }
 
     @Override
@@ -62,6 +63,9 @@ public class UpdateConversationRequest extends ActionRequest {
 
         if (this.conversationId == null) {
             exception = addValidationError("conversation id can't be null", exception);
+        }
+        if (this.updateContent == null) {
+            exception = addValidationError("Update conversation content can't be null", exception);
         }
 
         return exception;
@@ -89,8 +93,11 @@ public class UpdateConversationRequest extends ActionRequest {
         }
     }
 
-    private Map<String, Object> filterMapContent(Map<String, Object> mapContent) {
-        return mapContent
+    private Map<String, Object> filterUpdateContent(Map<String, Object> updateContent) {
+        if (updateContent == null) {
+            return new HashMap<>();
+        }
+        return updateContent
             .entrySet()
             .stream()
             .filter(map -> allowedList.contains(map.getKey()))
--- a/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.memory.action.conversation;
+
+import static org.opensearch.action.ValidateActions.addValidationError;
+import static org.opensearch.ml.common.conversation.ConversationalIndexConstants.APPLICATION_TYPE_FIELD;
+import static org.opensearch.ml.common.conversation.ConversationalIndexConstants.META_NAME_FIELD;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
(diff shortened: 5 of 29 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,218 @@
...
+ * Memory manager for Memories. It contains ML memory related operations like create, read interactions etc.
+ */
+@Log4j2
+@AllArgsConstructor
+public class MLMemoryManager {
+
+    private Client client;
+    private ClusterService clusterService;
+    private ConversationMetaIndex conversationMetaIndex;
+    private final String indexName = ConversationalIndexConstants.INTERACTIONS_INDEX_NAME;
```
ylwu-amzn, line 59: Why wrap another static constant with this final String ?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
```diff
@@ -0,0 +1,98 @@
...
+public class UpdateInteractionRequest extends ActionRequest {
+    private String interactionId;
+    private Map<String, Object> updateContent;
+
+    private static final Set<String> allowedList = new HashSet<>(Arrays.asList(INTERACTIONS_ADDITIONAL_INFO_FIELD));
+
+    @Builder
+    public UpdateInteractionRequest(String interactionId, Map<String, Object> updateContent) {
+        this.interactionId = interactionId;
+        this.updateContent = filterMapContent(updateContent);
```
ylwu-amzn, line 42: What if `updateContent` is null ? Do we allow null updateContent?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
```diff
@@ -0,0 +1,98 @@
...
+        super(in);
+        this.interactionId = in.readString();
+        this.updateContent = filterMapContent(in.readMap());
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(this.interactionId);
+        out.writeMap(this.getUpdateContent());
```
ylwu-amzn, line 55: what if updateContent is null?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionTransportAction.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private ActionListener<UpdateResponse> getUpdateResponseListener(
+        String interactionId,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
```
ylwu-amzn, line 58: We can only tell the interaction updated succesfully with `updateResponse.getResult() == DocWriteResponse.Result.UPDATED`, for other cases like `updateResponse == null` or `updateResponse.getResult() == null` or `updateResponse.getResult() != DocWriteResponse.Result.UPDATED` , we can't tell if the interaction updated successfully or not.
Suggest change to 
```suggestion
            if (updateResponse != null && updateResponse.getResult() == DocWriteResponse.Result.UPDATED) {
                log.info("Successfully updated the interaction with ID: {}", interactionId);
                actionListener.onResponse(updateResponse);
            } else {
                log.info("Failed to update the interaction with ID: {}", interactionId);
                actionListener.onResponse(updateResponse);
            }
```

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionTransportAction.java
```diff
@@ -0,0 +1,70 @@
...
+                log.info("Failed to update the interaction with ID: {}", interactionId);
+                actionListener.onResponse(updateResponse);
+                return;
+            }
+            log.info("Successfully updated the interaction with ID: {}", interactionId);
+            actionListener.onResponse(updateResponse);
+        }, exception -> {
+            log.error("Failed to update ML interaction with ID {}. Details: {}", interactionId, exception);
```
ylwu-amzn, line 66: If need to log exception stack trace, should use `log.error("message", exception)`

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
```diff
@@ -0,0 +1,98 @@
...
+            actionRequest.writeTo(osso);
+            try (StreamInput input = new InputStreamStreamInput(new ByteArrayInputStream(baos.toByteArray()))) {
+                return new UpdateInteractionRequest(input);
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException("failed to parse ActionRequest into UpdateInteractionRequest", e);
+        }
+    }
+
+    private Map<String, Object> filterMapContent(Map<String, Object> mapContent) {
```
ylwu-amzn, line 91: Change method and parameter name ?

```suggestion
    private Map<String, Object> filterUpdateContent(Map<String, Object> updateContent) {
```

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
```diff
@@ -0,0 +1,68 @@
...
+        UpdateConversationRequest updateConversationRequest = UpdateConversationRequest.fromActionRequest(request);
+        String conversationId = updateConversationRequest.getConversationId();
+        UpdateRequest updateRequest = new UpdateRequest(ConversationalIndexConstants.META_INDEX_NAME, conversationId);
+        updateRequest.doc(updateConversationRequest.getUpdateContent());
+        updateRequest.docAsUpsert(true);
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            client.update(updateRequest, getUpdateResponseListener(conversationId, listener, context));
+        } catch (Exception e) {
+            log.error("Failed to update Conversation for conversation id {}. Details {}:", conversationId, e);
```
ylwu-amzn, line 45: Use `log.error("<message>", e)` to log the exception stack trace.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    private ActionListener<UpdateResponse> getUpdateResponseListener(
+        String conversationId,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() != DocWriteResponse.Result.UPDATED) {
```
ylwu-amzn, line 56: Refer to https://github.com/opensearch-project/ml-commons/pull/1761/files#r1426367685

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
```diff
@@ -0,0 +1,68 @@
...
+                log.info("Failed to update the Conversation with ID: {}", conversationId);
+                actionListener.onResponse(updateResponse);
+                return;
+            }
+            log.info("Successfully updated the Conversation with ID: {}", conversationId);
```
ylwu-amzn, line 61: Should this be debug level?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -20,6 +20,8 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
```
ylwu-amzn, line 23: Why add line 23, 24 ?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import lombok.Builder;
+import lombok.Getter;
+
+@Getter
+public class UpdateConversationRequest extends ActionRequest {
+    private String conversationId;
+    private Map<String, Object> updateContent;
+
+    private static final Set<String> allowedList = new HashSet<>(Arrays.asList(META_NAME_FIELD, APPLICATION_TYPE_FIELD));
```
ylwu-amzn, line 38: Do we allow update APPLICATION_TYPE_FIELD ?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
```diff
@@ -0,0 +1,99 @@
...
+            actionRequest.writeTo(osso);
+            try (StreamInput input = new InputStreamStreamInput(new ByteArrayInputStream(baos.toByteArray()))) {
+                return new UpdateConversationRequest(input);
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException("failed to parse ActionRequest into UpdateConversationRequest", e);
+        }
+    }
+
+    private Map<String, Object> filterMapContent(Map<String, Object> mapContent) {
```
ylwu-amzn, line 92:
```suggestion
    private Map<String, Object> filterUpdateContent(Map<String, Object> updateContent) {
```

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
```diff
@@ -0,0 +1,99 @@
...
+public class UpdateConversationRequest extends ActionRequest {
+    private String conversationId;
+    private Map<String, Object> updateContent;
+
+    private static final Set<String> allowedList = new HashSet<>(Arrays.asList(META_NAME_FIELD, APPLICATION_TYPE_FIELD));
+
+    @Builder
+    public UpdateConversationRequest(String conversationId, Map<String, Object> updateContent) {
+        this.conversationId = conversationId;
+        this.updateContent = filterMapContent(updateContent);
```
ylwu-amzn, line 43: What if `updateContent` is null ? Do we allow null updateContent?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
```diff
@@ -0,0 +1,99 @@
...
+        super(in);
+        this.conversationId = in.readString();
+        this.updateContent = filterMapContent(in.readMap());
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(this.conversationId);
+        out.writeMap(this.getUpdateContent());
```
ylwu-amzn, line 56: What if updateContent is null?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Zhangxunmt: This is removed in the refactor.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
```diff
@@ -0,0 +1,75 @@
...
+    private boolean hasMoreTokens;
+
+    /**
+     * Constructor
+     * @param in stream input; assumes GetTracesResponse.writeTo was called
+     * @throws IOException if there's not a G.I.R. in the stream
+     */
+    public GetTracesResponse(StreamInput in) throws IOException {
+        super(in);
+        traces = in.readList(Interaction::fromStream);
```
ylwu-amzn, line 41: Could traces be null ?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
```diff
@@ -0,0 +1,75 @@
...
+     * @return whether there are more traces in this search
+     */
+    public boolean hasMorePages() {
+        return hasMoreTokens;
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params params) throws IOException {
+        builder.startObject();
+        builder.startArray(ActionConstants.RESPONSE_TRACES_LIST_FIELD);
```
ylwu-amzn, line 63: Should we check if traces is null or empty ?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
```diff
@@ -0,0 +1,66 @@
...
+        super(GetTracesAction.NAME, transportService, actionFilters, GetTracesRequest::new);
+        this.client = client;
+        this.cmHandler = cmHandler;
+    }
+
+    @Override
+    public void doExecute(Task task, GetTracesRequest request, ActionListener<GetTracesResponse> actionListener) {
+        int maxResults = request.getMaxResults();
+        int from = request.getFrom();
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().newStoredContext(true)) {
```
ylwu-amzn, line 55: Is it necessary to do this here?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+        Map<String, String> additionalInfo,
+        String parentIntId,
+        Integer traceNum,
+        ActionListener<CreateInteractionResponse> actionListener
+    ) {
+        Preconditions.checkNotNull(conversationId);
+        Preconditions.checkNotNull(input);
+        Preconditions.checkNotNull(response);
+        // additionalInfo cannot be null as flat object
+        additionalInfo = (additionalInfo == null) ? new HashMap<>() : additionalInfo;
```
ylwu-amzn, line 86: Why need to create a new HashMap if `additionalInfo` is null?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Get the interactions associate with this conversation, sorted by recency
+     * @param conversationId the conversation whose interactions to get
+     * @param lastNInteraction Return how many interactions
+     * @param actionListener get all the final interactions that are not traces
+     */
+    public void getFinalInteractions(String conversationId, int lastNInteraction, ActionListener<List<Interaction>> actionListener) {
+        Preconditions.checkNotNull(conversationId);
+        Preconditions.checkArgument(lastNInteraction > 0, "lastN must be at least 1.");
+        log.info("Getting Interactions, conversationId {}, lastN {}", conversationId, lastNInteraction);
```
ylwu-amzn, line 117: change to `debug` level?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    /**
+     * Get the interactions associate with this conversation, sorted by recency
+     * @param conversationId the conversation whose interactions to get
+     * @param lastNInteraction Return how many interactions
+     * @param actionListener get all the final interactions that are not traces
+     */
+    public void getFinalInteractions(String conversationId, int lastNInteraction, ActionListener<List<Interaction>> actionListener) {
```
ylwu-amzn, line 114: From the method name,  this method is to get the final interaction, but the comments says "the interactions associate with this conversation".  Which does this method do ?

ylwu-amzn: By reading code of `innerGetInteractions` of InteractionsIndex.java,  the "final interaction" means the interaction rather than the trace data. Suggest explain more details in the comments

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the interactions associate with this conversation, sorted by recency
+     * @param parentInteractionId the parent interaction id whose traces to get
+     * @param actionListener get all the trace interactions that are only traces
+     */
+    public void getTraces(String parentInteractionId, ActionListener<List<Interaction>> actionListener) {
+        Preconditions.checkNotNull(parentInteractionId);
+        log.info("Getting traces for conversationId {}", parentInteractionId);
```
ylwu-amzn, line 137: Change to debug level

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+
+        ActionListener<GetTracesResponse> al = ActionListener.wrap(getTracesResponse -> {
+            actionListener.onResponse(getTracesResponse.getTraces());
+        }, e -> { actionListener.onFailure(e); });
+
+        try {
+            client.execute(GetTracesAction.INSTANCE, new GetTracesRequest(parentInteractionId), al);
+        } catch (RuntimeException exception) {
```
ylwu-amzn, line 145: Why only catch runtime exception?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
```diff
@@ -0,0 +1,164 @@
...
+        Preconditions.checkArgument(lastNInteraction > 0, "lastN must be at least 1.");
+        log.info("Getting Interactions, conversationId {}, lastN {}", conversationId, lastNInteraction);
+
+        ActionListener<GetInteractionsResponse> al = ActionListener.wrap(getInteractionsResponse -> {
+            actionListener.onResponse(getInteractionsResponse.getInteractions());
+        }, e -> { actionListener.onFailure(e); });
+
+        try {
+            client.execute(GetInteractionsAction.INSTANCE, new GetInteractionsRequest(conversationId, lastNInteraction), al);
+        } catch (RuntimeException exception) {
```
ylwu-amzn, line 125: Why only catch runtime exception?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
ylwu-amzn: See such logic in multiple places, please check if that's ok for only catching runtime exception.

plugin/src/main/java/org/opensearch/ml/rest/RestMemoryUpdateConversationAction.java
```diff
@@ -0,0 +1,59 @@
...
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        UpdateConversationRequest updateConversationRequest = getRequest(request);
+        return restChannel -> client
+            .execute(UpdateConversationAction.INSTANCE, updateConversationRequest, new RestToXContentListener<>(restChannel));
+    }
+
+    @VisibleForTesting
+    private UpdateConversationRequest getRequest(RestRequest request) throws IOException {
+        if (!request.hasContent()) {
+            throw new OpenSearchParseException("Failed to update conversation: Request body is empty");
```
ylwu-amzn, line 49: What's the error status code for `OpenSearchParseException` ?  We should throw bad request error (400), maybe use `IllegalArgumentException` ?

plugin/src/main/java/org/opensearch/ml/rest/RestMemoryUpdateInteractionAction.java
```diff
@@ -0,0 +1,60 @@
...
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        UpdateInteractionRequest updateInteractionRequest = getRequest(request);
+        return restChannel -> client
+            .execute(UpdateInteractionAction.INSTANCE, updateInteractionRequest, new RestToXContentListener<>(restChannel));
+    }
+
+    @VisibleForTesting
+    private UpdateInteractionRequest getRequest(RestRequest request) throws IOException {
+        if (!request.hasContent()) {
+            throw new OpenSearchParseException("Failed to update interaction: Request body is empty");
```
ylwu-amzn, line 49: Same comment of https://github.com/opensearch-project/ml-commons/pull/1761/files#r1427543701

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
Zhangxunmt: 1. The rest request has to have content. Otherwise an error will throw at the Rest layer. So the request.contentParser().map() will always return the content. 
2. If this updateContent is empty after the filter, the update API will succeed with the "updated" status but essentially nothing changed. I have verified this behavior in my local environment. 
To be on the safe side, I added a check to make sure this updateContent is never Null. Otherwise it will throw an exception.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
Zhangxunmt: I added a check in the constructor in the filterUpdateContent(), which makes the updateContent an empty map if it's null. Also a null value check is added in the validate() function to throw an exception for the UpdateInteractionRequest.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionTransportAction.java
Zhangxunmt: good call out. Updated.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
Zhangxunmt: I think both info and debug should Ok. The updateResponse will return the status and we only pass the results back.   Even the updateResponse is not "Updated", we still finished the process and nothing to debug from our side. The log.error will help to debug anyways.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
Zhangxunmt: No it won't be null. this traces is generated from search results organizing into a initialized LinkedMap<>(). It only can be empty in the worst case. Also I added a @NonNull annotation in the constructor in the new commit.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesResponse.java
Zhangxunmt: addressed in the above comment.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
Zhangxunmt: it seems unnecessary but that's because we are not checking the users in the context for traces. The reason is that when user wants to get traces that means he is already seeing the interaction that was validated by "get or list Interaction access control". However, let's still keep this restore context just in case we will need to check users or some context info for edge cases arising in the future.

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Zhangxunmt: I think that was due to a bug fix. The flatobject cannot be null otherwise the index will fail. There was error reported before this line was added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Zhangxunmt: make sense. updated.

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Zhangxunmt: added this explain in the comment title. Also the @param actionListener has that explain.

plugin/src/main/java/org/opensearch/ml/rest/RestMemoryUpdateConversationAction.java
Zhangxunmt: https://github.com/opensearch-project/ml-commons/pull/1593. This was discussed in the earlier app sec review items. The OpenSearchParseException returns error code of 400. The IOException we used earlier throws 500 which is not accurate.

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/MLMemoryManager.java
Zhangxunmt: It should be Ok. The only reason it will catch error here is this client.execute method fails at runtime. All the methods in Transport Layer and Memory Layer have catch exceptions inside. 
To be on the safe side, I updated to Catch more general Exception e.
I remember quite some time ago, another SDE in my old team said that the catch exception should avoid using Exception because that couldn't give you enough info.

plugin/src/main/java/org/opensearch/ml/rest/RestMemoryUpdateInteractionAction.java
Zhangxunmt: the same addressed above.

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
```diff
@@ -29,6 +29,8 @@ public class ActionConstants {
     public final static String RESPONSE_CONVERSATION_LIST_FIELD = "conversations";
     /** name of list on interactions in all responses */
     public final static String RESPONSE_INTERACTION_LIST_FIELD = "interactions";
+    /** name of list on interactions in all responses */
```
dhrubo-os, line 32: change the comment?

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
```diff
@@ -56,20 +58,28 @@ public class ActionConstants {
...
 
     private final static String BASE_REST_PATH = "/_plugins/_ml/memory/conversation";
+    private final static String BASE_REST_INTERACTION_PATH = "/_plugins/_ml/memory/interaction";
     /** path for create conversation */
     public final static String CREATE_CONVERSATION_REST_PATH = BASE_REST_PATH + "/_create";
     /** path for get conversations */
     public final static String GET_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/_list";
+    /** path for update conversations */
+    public final static String UPDATE_CONVERSATIONS_REST_PATH  = BASE_REST_PATH + "/{conversation_id}/_update";
+    /** path for put interaction */
```
dhrubo-os, line 68: what is this comment about?

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
```diff
@@ -56,20 +58,28 @@ public class ActionConstants {
...
     /** path for create interaction */
     public final static String CREATE_INTERACTION_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_create";
     /** path for get interactions */
     public final static String GET_INTERACTIONS_REST_PATH = BASE_REST_PATH + "/{conversation_id}/_list";
+    /** path for get interactions */
```
dhrubo-os, line 73: is this comment correct?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param client OS Client for dealing with OS
+     * @param clusterService for some cluster ops
+     */
+    @Inject
+    public GetTracesTransportAction(
+        TransportService transportService,
+        ActionFilters actionFilters,
+        OpenSearchConversationalMemoryHandler cmHandler,
+        Client client,
+        ClusterService clusterService
```
dhrubo-os, line 44: Are we using this anywhere?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
dhrubo-os: let's add a comment in the code?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
```diff
@@ -0,0 +1,68 @@
...
+        String conversationId,
+        ActionListener<UpdateResponse> actionListener,
+        ThreadContext.StoredContext context
+    ) {
+        return ActionListener.runBefore(ActionListener.wrap(updateResponse -> {
+            if (updateResponse != null && updateResponse.getResult() == DocWriteResponse.Result.UPDATED) {
+                log.info("Successfully updated the Conversation with ID: {}", conversationId);
+                actionListener.onResponse(updateResponse);
+            } else {
+                log.info("Failed to update the Conversation with ID: {}", conversationId);
```
dhrubo-os, line 60: In which cases this can happen? During that time don't we expect exception?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        return exception;
+    }
+
+    public static UpdateInteractionRequest parse(XContentParser parser, String interactionId) throws IOException {
+        Map<String, Object> dataAsMap = null;
+        dataAsMap = parser.map();
+
+        if (dataAsMap == null) {
+            throw new OpenSearchParseException("Failed to parse UpdateInteractionRequest due to Null update content");
```
dhrubo-os, line 79: Can we add a test?

jngz-es (approved):

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
Zhangxunmt: This is reserved to allow conversations in the conversational search to add APPLICATION_TYPE_FIELD. We can remove this field in the allowedList after GA.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
Zhangxunmt: I added a check in the filterUpdateContent(Map<String, Object> updateContent) method. If updateContent is null, it returns a new Map with empty content to avoid null pointer. Also the ActionRequestValidationException validate() method will throw an exception if updateContent is null.

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
Zhangxunmt: I think this line is redundant. Just removed it.

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
Zhangxunmt: Good catch!

common/src/main/java/org/opensearch/ml/common/conversation/ActionConstants.java
Zhangxunmt: updated.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
Zhangxunmt: This is a leftover after the refactor. It was used earlier on. I just removed it.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
Zhangxunmt: For example, the client.threadPool().getThreadContext() failed, this catch will catch and log the messages. Anything exception that is not catched in the cmHandler.getTraces() will go here.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
Zhangxunmt: @Test
    public void testConstructor_NullUpdateContent() throws IOException {

This test is added. Probably not exactly what you wanted :). The constructor does not allow this dataAsMap to null. We can add another tests to construct a null updateContent in the next PR.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java
ylwu-amzn: > allow conversations in the conversational search to add APPLICATION_TYPE_FIELD.

Why conversation search need to update APPLICATION_TYPE_FIELD ? I think it's just saving interactions to one conversation. Will it update the conversation application type?

jngz-es (approved):

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
dhrubo-os: That part I understand too. I'm more curious to know about `anything exception that` cases. Because any kind of exception we are catching. But in this clause,  exception won't happen but still somehow we won't get the update response. So trying to understand in which cases this can happen?

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
dhrubo-os: No, I'm not asking for that test. I am asking to extend this test `testParse_Success` where we provide a parser which could be null, after parsing (`parser.map()`)

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
ylwu-amzn: @dhrubo-os  It will be hard to tell and construct edge case for this exception. I think some edge case like out of disk space, then the update will fail, or some circuit breaker like memory triggered may also cause update failure.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
ylwu-amzn: > provide a parser which could be null

@dhrubo-os , you mean test `parser` is null case ?
@Zhangxunmt Is it possible that `parser` will be null?

dhrubo-os: Yeah.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
dhrubo-os: I see, don't we get exception during those cases too? I'm just wondering, update failed, but we still sending success response with just a log. Is this reasonable or not.

Zhangxunmt: This log.error() means the update itself returned error but Ml-Commons finished processing the request successfully because we received the updateResponse. If we get any exceptions during the request, the actionListener will report error back to the customer through actionListener.onFailure(exception).

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
Zhangxunmt: This parser only comes from the RestAction. In the Rest layer, the RestRequest is guaranteed not be Null, otherwise exception is throw before reaching this point. We can manually create a request with Null parser, but this would only exists in the test.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
dhrubo-os: [nit] let's a comment in the code

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationTransportAction.java
dhrubo-os: [nit] Let's add `log.error` then instead of `log.info` at least.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateInteractionRequest.java
dhrubo-os: [nit] In summary,  after `dataAsMap = parser.map();`, `dataAsMap` shouldn't be null. May be we can simplify the code by removing that unnecessary null check.

dhrubo-os:
apply `spotlessApply`

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
Zhangxunmt: This was added by an earlier change that was removed later. They are redundant now so I removed them in the newest commit.

memory/src/main/java/org/opensearch/ml/memory/action/conversation/GetTracesTransportAction.java
Zhangxunmt: Added comment.

dhrubo-os:
let's not amend the commit. It's hard to track which is being updated in this commit.

Zhangxunmt:
> let's not amend the commit. It's hard to track which is being updated in this commit.

The recent ones are all related to comments updates. Nothing really changed.

dhrubo-os (approved):

ylwu-amzn (approved):

Issue #1774, opened 2023-12-18
Machine Learning Dashboard shows partially response after restarting non-ml nodes


zane-neo:
**What is the bug?**
Machine Learning Dashboard shows partially response after restarting non-ml nodes, more details can be found here: https://forum.opensearch.org/t/model-is-partially-responding-in-case-of-non-ml-node-restart/17030

**How can one reproduce the bug?**
1. Setup a cluster with 2.11.x version with data nodes and ml nodes.
2. Register a local model and deploy the model.
3. restart a data node and invoke profile API: /_plugins/_ml/profile/models/<your_model_id>?view=model
4. Check the `target_worker_nodes` and `worker_nodes`, you can see the difference.

**What is the expected behavior?**
`target_worker_nodes` and `worker_nodes` should be same.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

dhrubo-os:
resolving this issue in respect of https://github.com/opensearch-project/ml-commons/pull/1775

PR #852, opened 2023-04-13
Add model auto redeploy feature

zane-neo:
### Description
This is a new feature in ml-commons which is to enable the model auto redeploy when nodes join cluster based on different configurations, e.g. plugins.ml_commons.model_auto_redeploy.enable, plugins.ml_commons.model_auto_redeploy.lifetime_retry_times etc. This feature can make sure the crash/new node been deployed with running models automatically to save user efforts.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/832
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1775, opened 2023-12-18
Fix partially response issue in profile API result

zane-neo:
### Description
Fix partially response issue in profile API result. The root cause is when cluster nodes receive cluster change event(nodes add in this case), the added nodes will be put into the cache's target_worker_nodes blindly(without checking if the new nodes are eligible or not), and then trigger the auto redeploy job. The auto redeploy job can make sure the models are only deployed to eligible nodes and the deploy job can make sure the target_worker_nodes is correct.
If the added nodes put into cache, it won't get updated unless user performs model deploy, and for profile API and ML Dashboard(depends on profile API results), the target_worker_nodes and worker_nodes will show difference(target_worker_nodes has new nodes added even the node is not an eligible nodes for model running), and the model status will show as partial_response on ML Dashboard.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1774
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
@@ -67,7 +67,6 @@ public class MLCommonsClusterEventListener implements ClusterStateListener {
             mlModelManager.removeWorkerNodes(removedNodeIds, false);
         } else if (delta.added()) {
             List<String> addedNodesIds = delta.addedNodes().stream().map(DiscoveryNode::getId).collect(Collectors.toList());
-            mlModelManager.addModelWorkerNodes(addedNodesIds);
             mlModelAutoReDeployer.buildAutoReloadArrangement(addedNodesIds, state.getNodes().getClusterManagerNodeId());
         }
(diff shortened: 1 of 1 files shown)


ylwu-amzn:
Can you add more details to describe the issue? Seems this is from https://github.com/opensearch-project/ml-commons/pull/852/files#diff-e21ed0b0090f097417526680f7427b721ce8b9b0c51cdd8d75546ae289b5a22b

ylwu-amzn (approved):

dhrubo-os:
Sorry I still didn't understand clearly the issue. Could you please describe a bit more?

zane-neo:
> Sorry I still didn't understand clearly the issue. Could you please describe a bit more?

Added more details in description, if you still have question, please let me know the specific question since this issue indeed has several aspects.

dhrubo-os:
> > Sorry I still didn't understand clearly the issue. Could you please describe a bit more?
> 
> Added more details in description, if you still have question, please let me know the specific question since this issue indeed has several aspects.

Thanks. This makes sense now. Approving.

dhrubo-os (approved):

PR #1800, opened 2023-12-21
Model & user level throttling

b4sjoo:
### Description
This PR should enable model and user level rate limiting on TextEmbedding model and all Remote model.

### Check List
  - [ ] All tests pass


dhrubo-os:
Is this PR ready to review? Spotless missing. Is that 111 files? Or other commits also added here?

b4sjoo:
> Is this PR ready to review? Spotless missing. Is that 111 files? Or other commits also added here?

Main codes are ready to review, just some UTs are missing. This PR including some UT names refactor, but overall they are only about the throttling feature itself.

b4sjoo:
Yeah it's a super large PR...

dhrubo-os:
> Yeah it's a super large PR...

Is this possible to breakdown into smaller PR's? Then we can have more focused review.

b4sjoo:
> > Yeah it's a super large PR...
> 
> Is this possible to breakdown into smaller PR's? Then we can have more focused review.

I found it really hard to separate them into different PRs, because git does not allow separate PR in the same branch. However since they are strongly coupled together, separate branch seems impossible.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -54,12 +55,14 @@ public class CommonValue {
...
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
     public static final Integer ML_MODEL_GROUP_INDEX_SCHEMA_VERSION = 2;
-    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 8;
+    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 10;
     public static final String ML_CONNECTOR_INDEX = ".plugins-ml-connector";
     public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 2;
     public static final Integer ML_CONNECTOR_SCHEMA_VERSION = 2;
     public static final String ML_CONFIG_INDEX = ".plugins-ml-config";
     public static final Integer ML_CONFIG_INDEX_SCHEMA_VERSION = 2;
+    public static final String ML_MODEL_CONTROLLER_INDEX = ".plugins-ml-model-controller";
```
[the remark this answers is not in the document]
b4sjoo: Our goal here is to use a name that is general enough so that in the future we won't build another index. This has been discussed in our design review meeting.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -54,12 +55,14 @@ public class CommonValue {
     public static final String ML_MODEL_INDEX = ".plugins-ml-model";
     public static final String ML_TASK_INDEX = ".plugins-ml-task";
     public static final Integer ML_MODEL_GROUP_INDEX_SCHEMA_VERSION = 2;
-    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 8;
+    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 10;
```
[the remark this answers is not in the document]
b4sjoo: Will do. Bumped twice due to testing...

common/src/main/java/org/opensearch/ml/common/CommonValue.java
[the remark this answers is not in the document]
ylwu-amzn: For general purpose how about `.plugins-ml-controller` ? As we have agent and connector, they may also need controller

common/src/main/java/org/opensearch/ml/common/transport/controller/MLCreateModelControllerAction.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.ml.common.transport.controller;
+
+import org.opensearch.action.ActionType;
+
+public class MLCreateModelControllerAction extends ActionType<MLCreateModelControllerResponse>{
+    public static final MLCreateModelControllerAction INSTANCE = new MLCreateModelControllerAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/create_model_controller";
```
ylwu-amzn, line 11:
```suggestion
    public static final String NAME = "cluster:admin/opensearch/ml/controllers/_create";
```

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -222,6 +225,15 @@ public class CommonValue {
...
             + "      \""
+            + MLModel.IS_ENABLED_FIELD
+            + "\" : {\"type\": \"boolean\"},\n"
+            + "      \""
+            + MLModel.IS_MODEL_CONTROLLER_ENABLED_FIELD
+            + "\" : {\"type\": \"boolean\"},\n"
+            + "      \""
+            + MLModel.MODEL_RATE_LIMITER_CONFIG_FIELD
+            + "\" : {\"type\": \"flat_object\"},\n"
+            + "      \""
```
Zhangxunmt, line 236: Is it possible to combine these fields into 1 field saving some spaces? for example, if this rate limiter config field is empty, we can treat this model as controller disabled, etc.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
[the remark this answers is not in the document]
dhrubo-os: That's a good point. I agree with this. In addition, I feel like controller is a wide term? What kind of control this controller should have? Like we have model access control. Is this usage control? Or any kind of control in general?

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
```diff
@@ -0,0 +1,132 @@
...
+
+@Data
+public class MLModelController implements ToXContentObject, Writeable {
+
+    public static final String MODEL_ID_FIELD = "model_id"; // mandatory
+    public static final String USER_RATE_LIMITER_CONFIG = "user_rate_limiter_config";
+
+    @Getter
+    private String modelId;
+    private Map<String, MLRateLimiter> userRateLimiterConfig;
```
Zhangxunmt, line 38: Add a comment here showing the map key is for users, and value is for RateLimiter?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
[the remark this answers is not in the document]
b4sjoo: Sounds like a good name to me, I agree with `.plugins-ml-controller`

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
b4sjoo: Will do

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -530,7 +538,7 @@ private void indexRemoteModel(
                     String modelId = modelMetaRes.getId();
                     mlTask.setModelId(modelId);
                     log.info("create new model meta doc {} for upload task {}", modelId, taskId);
-                    mlTaskManager.updateMLTask(taskId, ImmutableMap.of(MODEL_ID_FIELD, modelId, STATE_FIELD, COMPLETED), 5000, true);
+                    mlTaskManager.updateMLTask(taskId, Map.of(MODEL_ID_FIELD, modelId, STATE_FIELD, COMPLETED), 5000, true);
```
Zhangxunmt, line 541: Why change to Map.of()? This also returns an immutable Map so nothing changed?

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
```diff
@@ -0,0 +1,141 @@
...
+    public static MLRateLimiter update(MLRateLimiter rateLimiter, MLRateLimiter updateContent) {
+        if (rateLimiter == null) {
+            return updateContent;
+        } else {
+            rateLimiter.update(updateContent);
+            return rateLimiter;
+        }
+    }
+
+    public static boolean isUpdatable(MLRateLimiter rateLimiter, MLRateLimiter updateContent) {
```
dhrubo-os, line 113: let's add comment for these methods specifying in which cases we need this

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
```diff
@@ -0,0 +1,148 @@
...
+        builder.startObject();
+        builder.field(MODEL_ID_FIELD, modelId);
+        if (userRateLimiterConfig != null) {
+            builder.field(USER_RATE_LIMITER_CONFIG, userRateLimiterConfig);
+        }
+        builder.endObject();
+        return builder;
+    }
+
+    public static boolean isDeployRequiredAfterUpdate(MLModelController modelController, MLModelController updateContent) {
```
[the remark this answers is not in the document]
b4sjoo: Correct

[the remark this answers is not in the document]
b4sjoo: Any update which will lead to a complete rate limiter construction will initiate a re-deploy

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -222,6 +225,15 @@ public class CommonValue {
             + MODEL_MAX_LENGTH_FIELD + "\":{\"type\":\"integer\"},\""
             + ALL_CONFIG_FIELD + "\":{\"type\":\"text\"}}},\n"
             + "      \""
+            + MLModel.IS_ENABLED_FIELD
+            + "\" : {\"type\": \"boolean\"},\n"
+            + "      \""
+            + MLModel.IS_MODEL_CONTROLLER_ENABLED_FIELD
+            + "\" : {\"type\": \"boolean\"},\n"
+            + "      \""
```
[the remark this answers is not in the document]
b4sjoo: MODEL.IS_ENABLED is the quota flag field. It's the field that admin actively set it to tell if the model can receive the request or not. MODEL.IS_MODEL_CONTROLLER_ENABLED_FIELD is the field passively set to indicate if there's a controller related to this model that need to be deployed when we deploy the model.

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
b4sjoo: Sure

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -75,6 +79,165 @@ public synchronized void setModelState(String modelId, MLModelState state) {
...
+    public synchronized void updateUserRateLimiterMap(String modelId, Map<String, TokenBucket> updateUserRateLimiterMap) {
+        log.debug("Updating the user level rate limiter for Model {}", modelId);
+        Map<String, TokenBucket> userRateLimiterMap = getExistingModelCache(modelId).getUserRateLimiterMap();
+        if (userRateLimiterMap != null) {
+            userRateLimiterMap.putAll(updateUserRateLimiterMap);
+        } else {
+            throw new OpenSearchStatusException(
+                "Model controller doesn't exist for the model. Consider calling create model controller api instead. Model ID: " + modelId,
+                RestStatus.CONFLICT
+            );
```
Zhangxunmt, line 158: Should we just add the Model Controller if it does not exist here? Update includes adding controller?

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -75,6 +79,165 @@ public synchronized void setModelState(String modelId, MLModelState state) {
...
+
+    /**
+     * Update the user rate limiter map for a single user.
+     * If the user rate limiter map doesn't exist for the model, consider calling setUserRateLimiterMap instead.
+     *
+     * @param modelId model id
+     * @param user user
+     * @param rateLimiter rate limiter
+     */
+    public synchronized void updateUserRateLimiterMap(String modelId, String user, TokenBucket rateLimiter) {
```
Zhangxunmt, line 170: A general question for all the update/get in the MLModelCache, do we need to check the data consistency with the indices before returning the results? Or they are all taken care in the transport layer actions?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -1079,6 +1042,65 @@ public void deployModel(
...
+        Map<String, Object> params = setUpParameterMap(modelId);
+        Predictable predictable = mlEngine.deploy(mlModel, params);
+        modelCacheHelper.setPredictor(modelId, predictable);
+    }
+
+    private Map<String, Object> setUpParameterMap(String modelId) {
+        TokenBucket modelRateLimiter = getModelRateLimiter(modelId);
+        Map<String, TokenBucket> userRateLimiterMap = getUserRateLimiterMap(modelId);
+
+        Map<String, Object> params = new HashMap<>() {
```
ylwu-amzn, line 1076: Let's use the standard way to initialize HashMap 
```
Map<String, Object> params = new HashMap<>();
params.put(ML_ENGINE, mlEngine);
params.put(SCRIPT_SERVICE, scriptService);
params.put(CLIENT, client);
```

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -1079,6 +1042,65 @@ public void deployModel(
         }
     }
 
+    private void deployRemoteOrBuiltInModel(MLModel mlModel, Integer eligibleNodeCount, ActionListener<String> wrappedListener) {
+        String modelId = mlModel.getModelId();
+        setupModelRateLimiter(modelId, eligibleNodeCount, mlModel.getModelRateLimiterConfig());
+        if (mlModel.getConnector() != null || FunctionName.REMOTE != mlModel.getAlgorithm()) {
```
ylwu-amzn, line 1048: What does this mean?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -984,42 +949,42 @@ public void deployModel(
...
-                            XCONTENT_REGISTRY,
-                            xContentRegistry,
-                            CLUSTER_SERVICE,
-                            clusterService
-                        );
                     // deploy remote model with internal connector or model trained by built-in algorithm like kmeans
-                    if (mlModel.getConnector() != null || FunctionName.REMOTE != mlModel.getAlgorithm()) {
-                        setupPredictable(modelId, mlModel, params);
-                        wrappedListener.onResponse("successful");
+                    if (mlModel.getIsModelControllerEnabled() != null && mlModel.getIsModelControllerEnabled()) {
```
ylwu-amzn, line 963:
```suggestion
                    if (BooleanUtils.isTrue(mlModel.getIsModelControllerEnabled())) {
```

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
b4sjoo: As method name said, it only covered remote and built-in model like Kmeans. Local model is covered in other part

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
b4sjoo: Ok

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
```diff
@@ -0,0 +1,148 @@
...
+                            throw new RuntimeException(e);
+                        }
+                    });
+                    break;
+                default:
+                    parser.skipChildren();
+                    break;
+            }
+        }
+        // Model ID can only be set through RestRequest.
```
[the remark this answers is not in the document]
b4sjoo: Model ID is set through sending the request url.

common/src/main/java/org/opensearch/ml/common/transport/controller/MLDeployModelControllerAction.java
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.controller;
+
+import org.opensearch.action.ActionType;
+
+public class MLDeployModelControllerAction extends ActionType<MLDeployModelControllerNodesResponse> {
+    public static final MLDeployModelControllerAction INSTANCE = new MLDeployModelControllerAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/deploy_model_controllers";
```
ylwu-amzn, line 12:
```suggestion
    public static final String NAME = "cluster:admin/opensearch/ml/controllers/deploy";
```

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
```diff
@@ -0,0 +1,141 @@
...
+        if (!isUpdatable(rateLimiter, updateContent)) {
+            return false;
+        } else {
+            return rateLimiter.isRateLimiterConstructable() || updateContent.isRateLimiterConstructable()
+                    || (rateLimiter.getRateLimitUnit() != null && updateContent.getRateLimitNumber() != null)
+                    || (rateLimiter.getRateLimitNumber() != null && updateContent.getRateLimitUnit() != null);
+        }
+    }
+
+    public boolean isRateLimiterConstructable() {
```
ylwu-amzn, line 134: This name looks confusing. I think you mean `isValid()`?

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
```diff
@@ -0,0 +1,141 @@
...
+        return (this.rateLimitUnit != null && this.rateLimitNumber != null);
+    }
+
+    public boolean isRateLimiterEmpty() {
```
ylwu-amzn, line 138: We don't need so verbose name
```suggestion
    public boolean isEmpty() {
```

common/src/main/java/org/opensearch/ml/common/transport/controller/MLModelControllerDeleteAction.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+package org.opensearch.ml.common.transport.controller;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.action.delete.DeleteResponse;
+
+public class MLModelControllerDeleteAction extends ActionType<DeleteResponse> {
+    public static final MLModelControllerDeleteAction INSTANCE = new MLModelControllerDeleteAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/model_controllers/delete";
```
ylwu-amzn, line 13:
```suggestion
    public static final String NAME = "cluster:admin/opensearch/ml/controllers/delete";
```

ylwu-amzn: Let's use consistent pattern `cluster:admin/opensearch/ml/<resouce_type>/<action>`

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -350,6 +362,17 @@ public class CommonValue {
             + "    }\n"
             + "}";
 
+    public static final String ML_MODEL_CONTROLLER_INDEX_MAPPING = "{\n"
+            + "    \"_meta\": {\"schema_version\": "
+            + ML_MODEL_CONTROLLER_INDEX_SCHEMA_VERSION
```
ylwu-amzn, line 367: Do we need model id field?

common/src/main/java/org/opensearch/ml/common/transport/controller/MLUndeployModelControllerAction.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.controller;
+
+import org.opensearch.action.ActionType;
+
+public class MLUndeployModelControllerAction extends ActionType<MLUndeployModelControllerNodesResponse> {
```
ylwu-amzn, line 10: So user can undeploy model controller separately ?  Which mean the model deployed , but undeploy the model controller, then no rate limiting any more?

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
```diff
@@ -32,11 +33,12 @@ public class MLUpdateModelInput implements ToXContentObject, Writeable {
     public static final String MODEL_VERSION_FIELD = "model_version"; // passively set when register model to a new model group
     public static final String MODEL_NAME_FIELD = "name"; // optional
     public static final String MODEL_GROUP_ID_FIELD = "model_group_id"; // optional
+    public static final String IS_ENABLED_FIELD = "is_enabled"; // optional
+    public static final String MODEL_RATE_LIMITER_CONFIG_FIELD = "model_rate_limiter_config"; // optional
     public static final String MODEL_CONFIG_FIELD = "model_config"; // optional
-    public static final String CONNECTOR_FIELD = "connector"; // optional
+    public static final String UPDATED_CONNECTOR_FIELD = "updated_connector"; // passively set when updating the internal connector
```
ylwu-amzn, line 39: When we update model description, we don't need a new filed `updated_description`. Why do we need this new filed when update connector?

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
```diff
@@ -45,25 +47,29 @@ public class MLUpdateModelInput implements ToXContentObject, Writeable {
...
     private String name;
     private String modelGroupId;
+    private Boolean isEnabled;
+    private MLRateLimiter modelRateLimiterConfig;
     private MLModelConfig modelConfig;
-    private Connector connector;
+    private Connector updatedConnector;
     private String connectorId;
-    private MLCreateConnectorInput connectorUpdateContent;
+    private MLCreateConnectorInput connector;
```
ylwu-amzn, line 55: This part was for new connector content. From the name this line means the original connector content , as I see line 53 means the new connector content `private Connector updatedConnector;` .   Is the understanding correct ? Why switch the meaning of these two fields?

common/src/main/java/org/opensearch/ml/common/transport/update_cache/MLUpdateModelCacheAction.java
```diff
@@ -9,7 +9,7 @@
 
 public class MLUpdateModelCacheAction extends ActionType<MLUpdateModelCacheNodesResponse> {
     public static final MLUpdateModelCacheAction INSTANCE = new MLUpdateModelCacheAction();
-    public static final String NAME = "cluster:admin/opensearch/ml/models/update_model_cache";
+    public static final String NAME = "cluster:admin/opensearch/ml/update_model_cache";
```
ylwu-amzn, line 12: Let's use consistent name 
`cluster:admin/opensearch/ml/<resouce_type>/<action>`

`cluster:admin/opensearch/ml/models/update_cache`

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
```diff
@@ -0,0 +1,249 @@
...
+                                .error(
+                                    "Failed to undeploy model controller from cache and delete the model controller for model {}",
+                                    modelId,
+                                    e
+                                );
+                            actionListener.onFailure(e);
+                        }), context::restore)
+                    );
+            } else {
+                deleteModelController(modelId, actionListener);
```
ylwu-amzn, line 195: So this method also support delete model for not deployed model ?  If yes, the method name is not accurate, just change to `deleteModelControllerWithDeployedModel`. If you want to highlight that you have some special logic for deployed model, please add more comments. no need to add everything to the method name.

common/src/main/java/org/opensearch/ml/common/transport/controller/MLUndeployModelControllerAction.java
b4sjoo: The deploy and undeploy of the model controller are passive behavior. User cannot actively trigger them.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
b4sjoo: It's the doc id. But when `GET` model cotroller it will print up.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
b4sjoo: This is not possible because this field is used as a flag to notify model if we need to auto-deploy/undeploy its corresponding controller.

common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -50,9 +51,13 @@ public class MLModel implements ToXContentObject {
...
     public static final String MODEL_STATE_FIELD = "model_state";
     public static final String MODEL_CONTENT_SIZE_IN_BYTES_FIELD = "model_content_size_in_bytes";
-    //SHA256 hash value of model content.
+    // SHA256 hash value of model content.
     public static final String MODEL_CONTENT_HASH_VALUE_FIELD = "model_content_hash_value";
 
+    // Model level quota and throttling control
+    public static final String IS_ENABLED_FIELD = "is_enabled";
+    public static final String MODEL_RATE_LIMITER_CONFIG_FIELD = "model_rate_limiter_config";
+    public static final String IS_MODEL_CONTROLLER_ENABLED_FIELD = "is_model_controller_enabled";
```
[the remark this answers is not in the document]
b4sjoo: Will do

common/src/main/java/org/opensearch/ml/common/transport/controller/MLUndeployModelControllerAction.java
ylwu-amzn: Add some comments to explain when will "deploy and undeploy of the model controller" be triggered ?

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
ylwu-amzn: Discussed offline with Sicheng. This filed is not user's input. It's constructed by code with the original connector and user's input new connector content (the `CONNECTOR_FIELD`).

ylwu-amzn: Generally we use input class only for user's input data. This implementation is different. Suggest add more comments so others know user can't input which fields.

ylwu-amzn: If you have more time, suggest add another class for internal usage. Keep this class only for user's input

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
```diff
@@ -182,17 +207,23 @@ public static MLUpdateModelInput parse(XContentParser parser) throws IOException
...
+                    break;
+                case MODEL_RATE_LIMITER_CONFIG_FIELD:
+                    modelRateLimiterConfig = MLRateLimiter.parse(parser);
+                    break;
                 case MODEL_CONFIG_FIELD:
                     modelConfig = TextEmbeddingModelConfig.parse(parser);
                     break;
-                case CONNECTOR_FIELD:
-                    connector = Connector.createConnector(parser);
+                case UPDATED_CONNECTOR_FIELD:
```
ylwu-amzn, line 219: User can never input updated connector, this is constructed by our code internally. Please remove this to avoid confusion.

b4sjoo: Ack

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
```diff
@@ -0,0 +1,249 @@
...
+        }
+    }
+
+    private void deleteModelController(String modelId, ActionListener<DeleteResponse> actionListener) {
+        DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_CONTROLLER_INDEX, modelId);
+        client.delete(deleteRequest, new ActionListener<>() {
+            @Override
+            public void onResponse(DeleteResponse deleteResponse) {
+                log.info("Model controller for model {} successfully deleted from index, result: {}", modelId, deleteResponse.getResult());
+                mlModelManager.updateModel(modelId, Map.of(MLModel.IS_MODEL_CONTROLLER_ENABLED_FIELD, false));
```
ylwu-amzn, line 209: What if this step failed ? Should we add action listener for this method ?

plugin/src/main/java/org/opensearch/ml/action/controller/GetModelControllerTransportAction.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLModelControllerGetResponse> actionListener) {
+        MLModelControllerGetRequest modelControllerGetRequest = MLModelControllerGetRequest.fromActionRequest(request);
+        String modelId = modelControllerGetRequest.getModelId();
+        FetchSourceContext fetchSourceContext = getFetchSourceContext(modelControllerGetRequest.isReturnContent());
+        GetRequest getRequest = new GetRequest(ML_MODEL_CONTROLLER_INDEX).id(modelId).fetchSourceContext(fetchSourceContext);
+        User user = RestActionUtils.getUserContext(client);
+        String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
```
ylwu-amzn, line 78: minor: keep this creating variable line close to the line where it's being used : line 87

plugin/src/main/java/org/opensearch/ml/action/controller/UndeployModelControllerTransportAction.java
```diff
@@ -0,0 +1,121 @@
...
+        return createUndeployModelControllerNodeResponse(request.getUndeployModelControllerNodesRequest());
+    }
+
+    private MLUndeployModelControllerNodeResponse createUndeployModelControllerNodeResponse(
+        MLUndeployModelControllerNodesRequest undeployModelControllerNodesRequest
+    ) {
+        String modelId = undeployModelControllerNodesRequest.getModelId();
+
+        Map<String, String> modelControllerUndeployStatus = new HashMap<>();
+        modelControllerUndeployStatus.put(modelId, "received");
```
ylwu-amzn, line 112: I don't see the status changed in this class. Is that by design?

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
b4sjoo: You can check the message inside, both the error and response will go into the log. I think it's fine because when doing model deploying I handled this edge case. Once it failed to be updated to false it can never set back to false actively, unless user manually create and delete the controller again.

plugin/src/main/java/org/opensearch/ml/action/controller/UndeployModelControllerTransportAction.java
```diff
@@ -0,0 +1,121 @@
...
+
+        String localNodeId = clusterService.localNode().getId();
+
+        mlModelManager.undeployModelController(modelId, ActionListener.wrap(r -> {
+            log.info("Successfully undeployed model controller for model {} on node {}", modelId, localNodeId);
+        }, e -> { log.error("Failed to undeploy model controller for model {} on node {}", modelId, localNodeId, e); }));
```
ylwu-amzn, line 118: From this line, if exception happens on node, just log it, won't throw exception. For such case, will the node added bo `undeployModelControllerNodesResponse.failures` ?

Refer to plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java line 234
`for (FailedNodeException failedNodeException : undeployModelControllerNodesResponse.failures()) {`

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private String[] getUndeployModelControllerFailedNodesList(
+        MLUndeployModelControllerNodesResponse undeployModelControllerNodesResponse
+    ) {
+        if (undeployModelControllerNodesResponse == null) {
+            return getAllNodes();
+        } else {
+            List<String> nodeIds = new ArrayList<>();
+            for (FailedNodeException failedNodeException : undeployModelControllerNodesResponse.failures()) {
```
ylwu-amzn, line 234: Have you tested this ?

plugin/src/main/java/org/opensearch/ml/action/controller/UndeployModelControllerTransportAction.java
b4sjoo: Yes, I followed other node response class. The node exception will be handled by a failure class. I used that for logging and responsing.

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
b4sjoo: Yep

b4sjoo: It is covered by UT

plugin/src/main/java/org/opensearch/ml/action/controller/GetModelControllerTransportAction.java
b4sjoo: Got it

plugin/src/main/java/org/opensearch/ml/action/controller/UndeployModelControllerTransportAction.java
b4sjoo: I checked other transport on node action class, it seems they don't throw exception, only find a way to handle them, so I think it will be fine

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteModelControllerTransportAction.java
b4sjoo: Ok I will add comments to it, because `deleteModelController` more fits to the direct delete method I think

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
b4sjoo: Yeah agree. We can moved out from parser first, but I will investigate another way to handle it.

common/src/main/java/org/opensearch/ml/common/transport/controller/MLUndeployModelControllerAction.java
b4sjoo: Ack

common/src/main/java/org/opensearch/ml/common/transport/controller/MLDeployModelControllerAction.java
b4sjoo: Changed

common/src/main/java/org/opensearch/ml/common/transport/controller/MLCreateModelControllerAction.java
b4sjoo: Changed

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
b4sjoo: It's by PE's recommendation.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
b4sjoo: By default is pass all.

common/src/main/java/org/opensearch/ml/common/transport/controller/MLModelControllerDeleteAction.java
b4sjoo: Changed

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
b4sjoo: Changed

common/src/main/java/org/opensearch/ml/common/transport/update_cache/MLUpdateModelCacheAction.java
b4sjoo: Changed

plugin/src/main/java/org/opensearch/ml/action/controller/GetModelControllerTransportAction.java
b4sjoo: It seems all the code keeps this line out of the try catch block, will keep aligned to this.

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelInput.java
b4sjoo: Discussed this topic offline

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
b4sjoo: Changed

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
b4sjoo: We will only update the cache once we confirm the index has been updated. That's how we control the consistency.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
b4sjoo: Changed

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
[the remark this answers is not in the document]
b4sjoo: Added

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
```diff
@@ -0,0 +1,155 @@
...
+@Data
+public class MLModelController implements ToXContentObject, Writeable {
+
+    public static final String MODEL_ID_FIELD = "model_id"; // mandatory
+    public static final String USER_RATE_LIMITER_CONFIG = "user_rate_limiter_config";
+
+    @Getter
+    private String modelId;
+    // The String is the username field where the MLRateLimiter is its corresponding rate limiter config.
+    private Map<String, MLRateLimiter> userRateLimiterConfig;
```
Zhangxunmt, line 40: In cases that multiple users have the same userRateLimiterConfig, or a single user has the same RateLimiter for multipleModels, do we need to send multiple redundant data to create or update the controller?

plugin/src/main/java/org/opensearch/ml/action/update_cache/UpdateModelCacheTransportAction.java
```diff
@@ -106,20 +106,15 @@ private MLUpdateModelCacheNodeResponse createUpdateModelCacheNodeResponse(
         MLUpdateModelCacheNodesRequest mlUpdateModelCacheNodesRequest
     ) {
         String modelId = mlUpdateModelCacheNodesRequest.getModelId();
```
Zhangxunmt, line 108: Why not use the mlModelManger to updateModelCache directly? What is the reason behind that we have to wrap this into a transport action call? I see that it's not taking any inputs from Users or Rest Layer.

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
Zhangxunmt: Just want to check if there was any discussion to put the MLRateLimiter into the .plugins-ml-config system index. In this way you can store all the users as a set or list in one RateLimiter that are associated, and also you only need to use a RateLimiter id as reference in the MLModel. Isn't it redundant to have a Controller defined inside MLModel, and this controller also contains the same modelId?

Zhangxunmt:
added a few comments from user experience and data model in the OS index. Do you have a quip document to track all the APIs usages and examples that will be added in the PR?

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
b4sjoo: > In cases that multiple users have the same userRateLimiterConfig, or a single user has the same RateLimiter for multipleModels, do we need to send multiple redundant data to create or update the controller?
I am afraid we have to at this time. Maybe we can figure out a way as a feature enhancement in the future.

b4sjoo: `.plugins-ml-config` is not appropriate because it stored MASTER_KEY in it. So the access control will be difficult there. I think we did the similar discussion in our design review.

plugin/src/main/java/org/opensearch/ml/action/update_cache/UpdateModelCacheTransportAction.java
b4sjoo: If this is a single node cluster then yes. However this is not the case in multi-node cluster. This is because we have to use the `NodeRequest` to ensure that we send the request to all necessary nodes.

b4sjoo:
> added a few comments from user experience and data model in the OS index. Do you have a quip document to track all the APIs usages and examples that will be added in the PR?

Addressed. For the API usage will added to the same quip we used to demo later today.

common/src/main/java/org/opensearch/ml/common/controller/MLModelController.java
Zhangxunmt: Actually it shouldn't be that difficult. The ml-config system index has no public API built for it so no one can see it. But we can also build APIs that ensure this master_key is never visible.

b4sjoo: Oh I get you. Yeah I think this is doable, but we may take some additional effort to change the api and I think we should do some further discussion on any potential issues on this decision.

Zhangxunmt (approved):
This can be merged with 2 major concerns from my view:
1. There will be data redundancy in the Ml-Model because both user/model throttling are bind within the Ml-Model. It's not obvious that we can decouple them easily in the future.
2. The APIs added in the PR should be all optional for users to run ml-commons. There shouldn't be any required setup for throttling unless someone needs to use throttling.

b4sjoo:
> This can be merged with 2 major concerns from my view:
> 
> There will be data redundancy in the Ml-Model because both user/model throttling are bind within the Ml-Model. It's not obvious that we can decouple them easily in the future.
> The APIs added in the PR should be all optional for users to run ml-commons. There shouldn't be any required setup for throttling unless someone needs to use throttling.

We can further discuss on the first one. For the second one, at this time all throttling related apis are optional for our user. The default behavior is pass for all requests (same as the previous).

rbhavna (approved):

PR #1812, opened 2023-12-27
Add ML agent tools

arjunkumargiri:
### Description
Add ML agent tools to main branch. RAG tool is still pending to be merged.

Added unit test to ensure code coverage.
 
Note: Build failure doesn't seem to be related to this commit and has failed on previous commit with same error: https://github.com/opensearch-project/ml-commons/pull/1800. Need to look into root cause of build failure.
 
### Check List
  - [x] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/VectorDBTool.java
```diff
@@ -0,0 +1,114 @@
...
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This tool supports neural search with embedding models and knn index.
+ */
+@Log4j2
+@Getter
+@Setter
+@ToolAnnotation(VectorDBTool.TYPE)
+public class VectorDBTool extends AbstractRetrieverTool {
```
ylwu-amzn, line 29: This tool moved to skills https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/VectorDBTool.java

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchAlertsTool.java
```diff
@@ -0,0 +1,155 @@
...
+import org.opensearch.ml.common.spi.tools.AbstractTool;
+import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@ToolAnnotation(SearchAlertsTool.TYPE)
+public class SearchAlertsTool extends AbstractTool {
```
ylwu-amzn, line 31: This tool will be moved to skills repo.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/VectorDBTool.java
arjunkumargiri: Updated. Is the plan to move all tools to skills repo?

ylwu-amzn: no, we plan to move tools which depend on other plugins to skills repo

ml-algorithms/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     implementation group: 'io.protostuff', name: 'protostuff-collectionschema', version: '1.8.0'
     testImplementation group: 'junit', name: 'junit', version: '4.13.2'
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '5.3.1'
+    testImplementation group: 'org.mockito', name: 'mockito-inline', version: '4.4.0'
```
Zhangxunmt, line 42: Not against using this version, but is there a reason for using '4.4.0' rather than the latest '5.x.x' version?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/VisualizationsTool.java
```diff
@@ -0,0 +1,144 @@
...
+import org.opensearch.search.SearchHits;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@ToolAnnotation(VisualizationsTool.TYPE)
+public class VisualizationsTool extends AbstractTool {
```
Hailong-am, line 35: this tool been moved to skills repo

ml-algorithms/build.gradle
dhrubo-os: I believe `mockito-core` should already have this `mockito-inline` ability in `5.3.1` version. Could you please check?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/NeuralSparseTool.java
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+
+        @Override
+        public NeuralSparseTool create(Map<String, Object> params) {
+            String index = (String) params.get(INDEX_FIELD);
+            String embeddingField = (String) params.get(EMBEDDING_FIELD);
+            String[] sourceFields = gson.fromJson((String) params.get(SOURCE_FIELD), String[].class);
+            String modelId = (String) params.get(MODEL_ID_FIELD);
+            Integer docSize = params.containsKey(DOC_SIZE_FIELD) ? Integer.parseInt((String) params.get(DOC_SIZE_FIELD)) : 2;
```
Zhangxunmt, line 91: Is null allowed for all these variables? should some checks added here to ensure modelId is not null, etc?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/PainlessScriptTool.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        public void init(Client client, ScriptService scriptService) {
+            this.client = client;
+            this.scriptService = scriptService;
+        }
+
+        @Override
+        public PainlessScriptTool create(Map<String, Object> map) {
```
Zhangxunmt, line 84: seems the map is not used at all in the creation. A member variable is missing or this map can be removed from this method?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,159 @@
...
+                            Map<String, Object> docContent = new HashMap<>();
+                            docContent.put("_index", hit.getIndex());
+                            docContent.put("_id", hit.getId());
+                            docContent.put("_score", hit.getScore());
+                            docContent.put("_source", hit.getSourceAsMap());
+                            return gson.toJson(docContent);
+                        });
+                        contextBuilder.append(doc).append("\n");
+                    }
+                    listener.onResponse((T) gson.toJson(contextBuilder.toString()));
```
Zhangxunmt, line 113: should adding log.info here to trace that the tool finished execution?  The same applies to other tools too.

ml-algorithms/build.gradle
arjunkumargiri: This was present in `agent_framework_dev` branch, will check if this dependency needs to be explicitly added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/PainlessScriptTool.java
arjunkumargiri: `create(Map<String, Object> map)` is the interface definition of `Tool.Factory` interface. We can consider creating another `create()`function in `Tool.Factory` interface which does not take in any argument. We can take this code cleanup separately.

ml-algorithms/build.gradle
```diff
@@ -64,8 +69,11 @@ dependencies {
     implementation 'software.amazon.awssdk:auth'
     implementation 'software.amazon.awssdk:apache-client'
     implementation 'com.amazonaws:aws-encryption-sdk-java:2.4.1'
-    implementation 'com.jayway.jsonpath:json-path:2.8.0'
+    implementation('com.jayway.jsonpath:json-path:2.8.0') {
+        exclude group: 'org.ow2.asm', module: 'asm'
+    }
     implementation group: 'org.json', name: 'json', version: '20231013'
+    implementation group: 'org.jsoup', name: 'jsoup', version: '1.16.1'
```
ylwu-amzn, line 76: I added this to feature branch ([code commit](https://github.com/opensearch-project/ml-commons/commit/1b85cffb07977a3aedb61a41637cb1b24f6cd016#diff-68650256ee78318934f0579cf938c57e39a60b4b5452b0ec6d76830cf3100940)) for some testing purpose. We can remove it when merge to main branch.

build.gradle
```diff
@@ -23,6 +23,7 @@ buildscript {
         }
 
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
+        kotlin_version = System.getProperty("kotlin.version", "1.8.21")
```
ylwu-amzn, line 26: We can remove this as alerting tool moved to skills repo

ml-algorithms/build.gradle
```diff
@@ -48,6 +49,10 @@ dependencies {
     implementation("ai.djl.onnxruntime:onnxruntime-engine:0.21.0") {
         exclude group: "com.microsoft.onnxruntime", module: "onnxruntime"
     }
+    implementation("org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}") {
```
ylwu-amzn, line 52: Same here, we can remove this as alerting tool moved to skills repo

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/NeuralSparseTool.java
arjunkumargiri: Good suggestion. Added.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
arjunkumargiri: It might be better for agent framework to log details on tool execution. This would centrally manage all tool related logging.

ml-algorithms/build.gradle
arjunkumargiri: Done.

build.gradle
arjunkumargiri: Done.

ml-algorithms/build.gradle
arjunkumargiri: Removed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/VisualizationsTool.java
arjunkumargiri: Removed

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,160 @@
...
+    protected abstract String getQueryBody(String queryText);
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        try {
+            String question = parameters.get(INPUT_FIELD);
+            try {
+                question = gson.fromJson(question, String.class);
+            } catch (Exception e) {
+                // throw new IllegalArgumentException("wrong input");
```
dhrubo-os, line 82: Are we going to do anything here?

ml-algorithms/build.gradle
```diff
@@ -64,7 +65,7 @@ dependencies {
     implementation 'software.amazon.awssdk:auth'
     implementation 'software.amazon.awssdk:apache-client'
     implementation 'com.amazonaws:aws-encryption-sdk-java:2.4.1'
-    implementation 'com.jayway.jsonpath:json-path:2.8.0'
+    implementation('com.jayway.jsonpath:json-path:2.8.0')
```
dhrubo-os, line 68: why do we need `(`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,160 @@
...
+@Getter
+@Setter
+public abstract class AbstractRetrieverTool extends AbstractTool {
+    public static final String DEFAULT_DESCRIPTION = "Use this tool to search data in OpenSearch index.";
+    public static final String INPUT_FIELD = "input";
+    public static final String INDEX_FIELD = "index";
+    public static final String SOURCE_FIELD = "source_field";
+    public static final String DOC_SIZE_FIELD = "doc_size";
+    public static final int DEFAULT_DOC_SIZE = 2;
+    public static final int DEFAULT_K = 10;
```
dhrubo-os, line 48: Can we add comment for these two variables?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/NeuralSparseTool.java
```diff
@@ -0,0 +1,134 @@
...
+ * This tool supports neural_sparse search with sparse encoding models and rank_features field.
+ */
+@Log4j2
+@Getter
+@Setter
+@ToolAnnotation(NeuralSparseTool.TYPE)
+public class NeuralSparseTool extends AbstractRetrieverTool {
+    public static final String TYPE = "NeuralSparseTool";
+    public static final String MODEL_ID_FIELD = "model_id";
+    public static final String EMBEDDING_FIELD = "embedding_field";
```
dhrubo-os, line 33: Should we add a default description for this tool?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/PainlessScriptTool.java
```diff
@@ -0,0 +1,100 @@
...
+import lombok.NonNull;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@ToolAnnotation(PainlessScriptTool.TYPE)
+public class PainlessScriptTool extends AbstractTool {
+    public static final String TYPE = "PainlessScriptTool";
+    public static final String SCRIPT = "script";
+    public static final String SCRIPT_PARAMS = "script_params";
+    private static String DEFAULT_DESCRIPTION = "Use this tool to get index information.";
```
dhrubo-os, line 31: This description doesn't seem quite relevant.

spi/build.gradle
```diff
@@ -46,7 +47,8 @@ dependencies {
     compileOnly "org.opensearch:opensearch:${opensearch_version}"
 
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
-    testImplementation "org.apache.logging.log4j:log4j-core:${versions.log4j}"
+    testImplementation group: 'junit', name: 'junit', version: '4.13.2'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '4.4.0'
```
dhrubo-os, line 51: may be we can add the same mockito version we have in the build.gradle file of ml-algorithms?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,163 @@
...
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * Abstract tool supports search paradigms in neural-search plugin.
+ */
+@Log4j2
+@Getter
+@Setter
+public abstract class AbstractRetrieverTool extends AbstractTool {
```
ylwu-amzn, line 41: This tool moved to skills repo https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/NeuralSparseTool.java
```diff
@@ -0,0 +1,135 @@
...
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This tool supports neural_sparse search with sparse encoding models and rank_features field.
+ */
+@Log4j2
+@Getter
+@Setter
+@ToolAnnotation(NeuralSparseTool.TYPE)
+public class NeuralSparseTool extends AbstractRetrieverTool {
```
ylwu-amzn, line 30: This tool moved to skills repo https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/NeuralSparseSearchTool.java

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AbstractRetrieverTool.java
arjunkumargiri: It is confusing that tools are split between ml-commons and skills repo, it might better to move all tools to skills repo even if there are no dependencies to other plugins.

ylwu-amzn: I think keeping the common tools in ml-commons could make integ test of agent framework easier. Otherwise we have another circular dependency between ml-commons and skills repo. For common tools like agent tool and ml model tool, they are very lightweight. 

The logic is very like the way how we organize OpenSearch plugins: we have some common plugins in OpenSearch core ([code link](https://github.com/opensearch-project/OpenSearch/tree/main/plugins)) , but we also have separate repo for individual/heavy plugins.

ylwu-amzn: The plan is moving all tools depending on other plugins to the new skills repo. The new repo is mainly to solve the circular dependency between ml-commons and other plugins which depend on ml-commons. For example , SQL plugin depends on ml-commons client now, if we add PPL tool to ml-commons, then ml-commons will also depend on SQL plugin. That will be circular dependency for integ test. We discussed with infra team about other options like using custom build workflow to build ml-commons client, then build SQL, then build ml-commons plugin. But that's not easy from infra team and hard to maintain. That's why we build the new skills repo , so we can use current way to build ml-commons first, then the ml-commons client will be ready. Then we can build SQL plugin. When both ml-commons and SQL ready, we can build the new skills repo.

arjunkumargiri: I agree on skills repo being the owner of tools and it will help simplify tool management. However, maintaining some subset of tools in ml-commons just for integ tests doesn't seem a valid usecase. If required integ tests can create dummy tools to bypass circular dependency between ml-commons and skills package.

PR #1813, opened 2023-12-28
Bug fix: Incorrect tool selection by LLM

arjunkumargiri:
### Description
Bug fix: Incorrect tool selection by LLM
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/PromptTemplate.java
```diff
@@ -5,10 +5,10 @@ public class PromptTemplate {
...
     public static final String PROMPT_TEMPLATE = "\n\nHuman:${parameters.prompt.prefix}\n\n${parameters.prompt.suffix}\n\nAssistant:";
     public static final String PROMPT_TEMPLATE_TOOL_RESPONSE =
-        "TOOL RESPONSE: \n---------------------\n${parameters.observation}\n\nUSER'S INPUT\n--------------------\n\nOkay, so what is the response to my last comment? If using information obtained from the tools you must mention it explicitly without mentioning the tool names - I have forgotten all TOOL RESPONSES! Remember to respond with a markdown code snippet of a json blob with a single action, and NOTHING else.";
+        "TOOL RESPONSE: \n---------------------\n${parameters.observation}\n\nUSER'S INPUT\n--------------------\n\nOkay, so what is the response to my last comment? If using information obtained from the tools you must mention it explicitly without mentioning the tool names. Remember to respond with a markdown code snippet of a json blob with a single action, and NOTHING else.";
```
xluo-aws, line 13: Not sure if you want to remove this prompt, seems it's added on purpose.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/PromptTemplate.java
```diff
@@ -5,10 +5,10 @@ public class PromptTemplate {
...
...FORMAT INSTRUCTIONS\n----------------------------\nOutput a JSON markdown code snippet containing a valid JSON object in one of two formats:\n\n**Option 1:**\nUse this if you want the human to use a tool.\nMarkdown code snippet formatted in the following schema:\n\n```json\n{\n    \"thought\": string, // think about what to do next: if you know the final answer just return \"Now I know the final answer\", otherwise suggest which tool to use.\n    \"action\": string, // The action to take. Must be one of these tool names: [${parameters.tool_names}], do NOT use any other name for action except the tool names. If the required tool is not part of available tools then use Option 2 to generate final response with available information.\n    \"action_input\": string // The input to the action. May be a stringified object.\n}\n```\n\n**Option #2:**\nUse this if you want to respond directly and conversationally to the human. Markdown code snippet formatted in the following schema:\n\n```json\n{\n    \"thought\": \"Now I know the final answer\",\n    \"final_answer\": string, // summarize and return the final answer in a sentence with details, don't just return a number or a word. \n}\n```";
```
xluo-aws, line 8: Confirmed this new prompt is able to address two problems we have before.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/PromptTemplate.java
arjunkumargiri: @jngz-es do you know why this prompt was added? I ran into few issues with this prompt where LLM said it no longer has access to tool output and it was not able to generate a valid response.

xluo-aws: Arjun, do you have examples to trigger the no access to tool problem? I was able to reproduce it consistently before, but it has been fixed by your fix to PROMPT_FORMAT_INSTRUCTION. I reverted your change to PROMPT_TEMPLATE_TOOL_RESPONSE and it continues to work fine.

arjunkumargiri: I was not able to consistently reproduce the problem, even before applying the fix I was only transiently running into the bug.

dhrubo-os:
@arjunkumargiri can we close this pull request? do we need this anymore?

PR #1816, opened 2023-12-28
support regenerate for chatbot

Hailong-am:
### Description
Regenerate api for chatbot, `memory_id` and `regenerate_interaction_id` is needed

```json
POST /_plugins/_ml/agents/<agent_id>/_execute

{
 "parameters": {
  "memory_id": "<memory_id>",
  "regenerate_interaction_id": "<regenerate_interaction_id>"
 }
}
```

```json
{
 "inference_results": [
  {
   "output": [
    {
     "name": "memory_id",
     "result": "CjJIqowBPJb9bj-RF685"
    },
    {
     "name": "response",
     "result": "Thought: Let me check the index sizes"
    },
    {
     "name": "response",
     "result": "\nObservation: health\tstatus\tindex\tuuid\tpri\trep\tdocs.count\tdocs.deleted\tstore.size\tpri.store.size\nyellow\topen\t.plugins-ml-memory-message\tPpjdHagoTeu-6e0cV_XPMQ\t1\t1\t18\t0\t37.9kb\t37.9kb\ngreen\topen\t.plugins-ml-model-group\tG3WmQfOOQKu5wjFMNxZ5Ww\t1\t0\t1\t0\t5.6kb\t5.6kb\nyellow\topen\t.plugins-ml-memory-meta\tHM_eIsqbShaJrf3hhgPesw\t1\t1\t1\t1\t12.9kb\t12.9kb\ngreen\topen\t.plugins-ml-config\tPmv5mywfToGcy9J_EEq80w\t1\t0\t1\t0\t3.9kb\t3.9kb\ngreen\topen\t.plugins-ml-model\tuOxxq_g7SLuPoIV2gQCFRw\t1\t0\t1\t1\t32kb\t32kb\ngreen\topen\t.plugins-ml-agent\tNwRqlaDUQO-u5jY1grJsJQ\t1\t0\t2\t0\t26.5kb\t26.5kb\ngreen\topen\t.plugins-ml-task\tdY3Ie4YUTfuOY4639AE0Bg\t1\t0\t3\t2\t28.6kb\t28.6kb\ngreen\topen\t.plugins-ml-connector\tb9cc8YjOTjWg4lDijk9DtQ\t1\t0\t1\t0\t21.7kb\t21.7kb\n"
    },
    {
     "name": "response",
     "result": "Now I know the final answer"
    },
    {
     "name": "response",
[...]

Issue #1802, opened 2023-12-22
Fine tune GetTool and ListTools API


jackiehanyang:
Back in August, when many things were still experimental, we hardcoded four internal tools along with one external tool in the GetTool API and ListTools API for Tools Management. 

This setup is still present in the feature branch - feature/agent_framework. After merging into main, we need to identify the gap between the current Tools implementation and the hardcoded tools we previously used, and change the implementation of these two API accordingly. 

Some gaps are mentioned as unresolved comments in this [PR](https://github.com/opensearch-project/ml-commons/pull/1764). Please feel free to add more comments here on what we need to do to fine-tune the Tools Management piece.

Hailong-am:
Not sure who will be the consumer of this API, tool name is not enough for register an agent, if it should used by ai flow.

dhrubo-os:
I think we should first identify the gap between the current Tools implementation and the hardcoded tools we previously used, solve this gap and then reflect the current PR accordingly and then merge. IMO, it's not better practice to merge this static tools in main. In the dev branch we just did POC. Let's not just merge the POC code in main. 

cc @ylwu-amzn

PR #1764, opened 2023-12-15
Add GetTool and ListTools API

jackiehanyang:
### Description
This pr merges GetTool and ListTools API
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1818, opened 2023-12-29
 Add GetTool API and ListTools API

jackiehanyang:
### Description
This pr moves GetTool and ListTools API from feature_dev branch. All code should be identical except this pr removes the hardcoded internal tools part. Also, renaming all `externalTools` attribute as `toolMetadataList` to reduce confusion as we don't differentiate internal or external tools when building GetTool and ListTools result. Instead, we pass internal tools and external tools together as a whole tool factories to these two rest APIs.

Had to close the [original pr](https://github.com/opensearch-project/ml-commons/pull/1764) due to too much merge conflict, and wanted to make sure all code is from feature_dev branch and it don't get messed up during resolving merge conflicts. You can find references in the original pr conversations.

**Note**: Build failure doesn't seem to be related to this commit and has failed on previous commit with same error: https://github.com/opensearch-project/ml-commons/pull/1800. Root cause in https://github.com/opensearch-project/ml-commons/pull/1816/files#r1437937707

### Testing
```
GET http://localhost:9200/_plugins/_ml/tools/MLModelTool
{
	"name": "MLModelTool",
	"description": "Use this tool to run any model.",
	"type": "MLModelTool",
	"version": "undefined"
}
```
```
GET http://localhost:9200/_plugins/_ml/tools
{
	"name": "MLModelTool",
	"description": "Use this tool to run any model.",
	"type": "MLModelTool",
	"version": "undefined"
}{
	"name": "AgentTool",
	"description": "Use this tool to run any agent.",
	"type": "AgentTool",
	"version": "undefined"
}{
	"name": "CatIndexTool",
	"description": "Use this tool to get index information.",
	"type": "CatIndexTool",
	"version": "undefined"
}
```
 
### Issues Resolved
related issues - https://github.com/opensearch-project/ml-commons/issues/1802
 
### Check List
  - [x] All tests pass


ylwu-amzn:
https://github.com/opensearch-project/ml-commons/actions/runs/7359436543/job/20034209306?pr=1818
```
* What went wrong:
Execution failed for task ':opensearch-ml-plugin:jacocoTestCoverageVerification'.
> Rule violated for class org.opensearch.ml.action.tools.ListToolsTransportAction: lines covered ratio is 0.2, but expected minimum is 0.8
  Rule violated for class org.opensearch.ml.action.tools.GetToolTransportAction: lines covered ratio is 0.1, but expected minimum is 0.8
  Rule violated for class org.opensearch.ml.rest.RestMLGetToolAction: lines covered ratio is 0.5, but expected minimum is 0.8
  Rule violated for class org.opensearch.ml.rest.RestMLListToolsAction: lines covered ratio is 0.5, but expected minimum is 0.8
```

common/src/main/java/org/opensearch/ml/common/ToolMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+
+
+public class ToolMetadata implements ToXContentObject, Writeable {
+
+    public static final String TOOL_NAME_FIELD = "name";
+    public static final String TOOL_DESCRIPTION_FIELD = "description";
+
+    @Getter
+    private String name;
```
ylwu-amzn, line 28: Jing did a refactor and changed name to type([code link](https://github.com/opensearch-project/ml-commons/commit/90d9b310811c682f55ee5415d53ad0be8f671156#diff-f3f27d4e953486c2798c6edef8641779b3211dabbe5374c5bfe1737b3aab2f25)). Change to type?

jackiehanyang: Jing's change still kept `name`, just added a new field `type`. Do we want to include `type` in ToolMetadata and show it in GetTool and ListTools response?

dhrubo-os: I think we should show `type` & `version` both.

jngz-es: Yeah, we need both type and name.

jackiehanyang: Should we add a default version each tool? Currently it's set to null - https://github.com/opensearch-project/ml-commons/blob/8352e69ae4edd2b615933ff2947d61d1d9546740/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java#L78

jackiehanyang: Added type and version information in the latest revision

plugin/src/main/java/org/opensearch/ml/tools/ListToolsTransportAction.java
```diff
@@ -0,0 +1,47 @@
...
+
+        List<ToolMetadata> toolsList = mlToolsGetRequest.getToolMetadataList();
+
+        try {
+            listener.onResponse(MLToolsListResponse.builder().toolMetadata(toolsList).build());
+        } catch (Exception e) {
+            log.error("Failed to get tools list", e);
+            listener.onFailure(e);
+        }
+    }
```
Zhangxunmt, line 46: Do we really need this transport layer action if it's only returning the toolsList from the ListToolsRequest? It seems like the response is already included in the Request, so this response can be returned right away when the Request is constructed?

plugin/src/main/java/org/opensearch/ml/tools/GetToolTransportAction.java
```diff
@@ -0,0 +1,56 @@
...
+            ToolMetadata theTool = toolsList
+                .stream()
+                .filter(tool -> tool.getName().equals(toolName))
+                .findFirst()
+                .orElseThrow(NoSuchElementException::new);
+            listener.onResponse(MLToolGetResponse.builder().toolMetadata(theTool).build());
+        } catch (Exception e) {
+            log.error("Failed to get tool", e);
+            listener.onFailure(e);
+        }
```
Zhangxunmt, line 53: Why not just doing this in the rest layer and return the response right away? Or whoever calls this action can do it directly from the client side?

plugin/src/main/java/org/opensearch/ml/tools/ListToolsTransportAction.java
jackiehanyang: Simply followed the current ml-commons API structure: from rest layer to transport layer. If we can guarantee that there won't be use case of calling this transport action directly in the future, then we can talk about a plan to only leave the rest layer action and remove this one

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/tools/GetToolTransportAction.java
jackiehanyang: same as above comment, we can return the response right away from the client side. But will it be a short-term solution and make it hard for future iterations?

common/src/test/java/org/opensearch/ml/common/transport/tools/MLToolGetRequestTests.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+        };
+        MLToolGetRequest.fromActionRequest(actionRequest);
+    }
+
+    @Test
+    public void validate_Exception_NullToolName() {
+        MLToolGetRequest mlToolGetRequest = MLToolGetRequest.builder().build();
+        ActionRequestValidationException exception = mlToolGetRequest.validate();
+        assertEquals("Validation Failed: 1: Tool name can't be null;", exception.getMessage());
```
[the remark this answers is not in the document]
jackiehanyang: It's validating the tool name in the get request cannot be null, not the name of the actual tool cannot be null. So it's different than the code example you pasted and it doesn't apply in this case.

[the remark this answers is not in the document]
mingshl: Fair enough

PR #1835, opened 2024-01-04
Adding connector http timeout in the connector level

dhrubo-os:
### Description
[Adding connector http timeout in the connector level]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


zane-neo:
LGTM

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+import org.opensearch.ml.common.connector.Connector;
+import org.opensearch.ml.common.connector.ConnectorHttpClientConfig;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor {
+    private Integer maxConnections;
```
ylwu-amzn, line 17: Can we use `ConnectorHttpClientConfig` here ? So we don't need to add more and more fields to this class

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -61,8 +61,18 @@ public class HttpJsonConnectorExecutor implements RemoteConnectorExecutor {
     @Getter
     private Client client;
 
+    private CloseableHttpClient httpClient;
+
     public HttpJsonConnectorExecutor(Connector connector) {
+        super.validate(connector);
         this.connector = (HttpConnector) connector;
+        this.httpClient = MLHttpClientFactory
+            .getCloseableHttpClient(super.getConnectionTimeoutInMillis(), super.getReadTimeoutInMillis(), super.getMaxConnections());
```
ylwu-amzn, line 70: Can we use `getHttpClientConfig` method of Connector class ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -98,6 +98,8 @@ default void setModelRateLimiter(TokenBucket modelRateLimiter) {}
 
     default void setUserRateLimiterMap(Map<String, TokenBucket> userRateLimiterMap) {}
 
+    default void initialize() {}
```
ylwu-amzn, line 101: Seems this method is not being used, delete it?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
dhrubo-os: Good idea, addressed that.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
dhrubo-os: Addressed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
dhrubo-os: removed.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
```diff
@@ -0,0 +1,122 @@
...
+@EqualsAndHashCode
+public class ConnectorHttpClientConfig implements ToXContentObject, Writeable {
+
+    public static final String MAX_CONNECTION_FIELD = "max_connection";
+    public static final String CONNECTION_TIMEOUT_FIELD = "connection_timeout";
+    public static final String READ_TIMEOUT_FIELD = "read_timeout";
+
+    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
+    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(20000);
+    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(20000);
```
zane-neo, line 33: Waiting for 20s for user seems not a good idea, is this the suggested default value from OpenSearch?

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -42,6 +45,11 @@ public class MLCreateConnectorInput implements ToXContentObject, Writeable {
     public static final String OWNER_FIELD = "owner";
     public static final String ACCESS_MODE_FIELD = "access_mode";
     public static final String DRY_RUN_FIELD = "dry_run";
+    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
+    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(10000);
+    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(10000);
```
zane-neo, line 50: Why this default value is different with above ones?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -0,0 +1,26 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor {
+    private ConnectorHttpClientConfig httpClientConfig;
+
+    public void validate(Connector connector) {
```
Zhangxunmt, line 19: should we rename this to "initialization"? No Exception is handled or thrown so this function looks more like an initialization rather than validation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -68,7 +71,27 @@ public AwsConnectorExecutor(Connector connector, SdkHttpClient httpClient) {
     }
 
     public AwsConnectorExecutor(Connector connector) {
-        this(connector, new DefaultSdkHttpClientBuilder().build());
+        super.validate(connector);
+        this.connector = (AwsConnector) connector;
+        Duration connectionTimeout = Duration.ofMillis(super.getHttpClientConfig().getConnectionTimeout());
+        Duration readTimeout = Duration.ofMillis(super.getHttpClientConfig().getReadTimeout());
+        try (
```
Zhangxunmt, line 78: aren't adding a catch for this try? Also there is the existing constructor "public AwsConnectorExecutor(Connector connector, SdkHttpClient httpClient)". Should we apply default client configs for this httpClient too for the sake of BWC?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
dhrubo-os: https://opensearch.org/docs/1.2/opensearch/rest-api/cluster-settings/#:~:text=Default%20is%2030%20seconds.



From here, it looks like the default value is 30 seconds. I'm planning to make this 30 seconds unless we have strong reasoning not to do this.

ylwu-amzn: The other way is to keep the default http timeout, we don't need to use 
```
this.readTimeout = Objects.requireNonNullElse(readTimeout, READ_TIMEOUT_DEFAULT_VALUE);
```
We just let `this.readTimeout = readTimeout` , if it's null, we don't need to set it, then client will use default timeout.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public static final String MAX_CONNECTION_FIELD = "max_connection";
+    public static final String CONNECTION_TIMEOUT_FIELD = "connection_timeout";
+    public static final String READ_TIMEOUT_FIELD = "read_timeout";
+
+    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
+    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+
+    private Integer maxConnections;
```
samuel-oci, line 35: why not using primitive `int` instead of `Integer`, it is usually recommended unless trying to support `null` value use case?
Also it can prevent unnecessary boxing/unboxing later on.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
```diff
@@ -0,0 +1,122 @@
...
+        Integer readTimeout
+    ) {
+        this.maxConnections = maxConnections;
+        this.connectionTimeout = connectionTimeout;
+        this.readTimeout = readTimeout;
+
+    }
+
+    public ConnectorHttpClientConfig(StreamInput input) throws IOException {
+        this.maxConnections = input.readOptionalInt();
```
samuel-oci, line 52: See comment above regarding use of primitive `int` instead of `Integer` to prevent unnecessary boxing and unboxing of primitive values (and prevent null situation).

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
dhrubo-os: addressed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
dhrubo-os: added exception.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
dhrubo-os: default value of int is 0. The target of this `maxConnections` or the other values is: is it is null, then it will automatically use the default value of httpclient. If we use int, then for null it will default to 0 which is not aligning with the goal.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
```diff
@@ -0,0 +1,122 @@
...
+            builder.field(CONNECTION_TIMEOUT_FIELD, connectionTimeout);
+        }
+        if (readTimeout != null) {
+            builder.field(READ_TIMEOUT_FIELD, readTimeout);
+        }
+        return builder.endObject();
+    }
+
+    public static ConnectorHttpClientConfig fromStream(StreamInput in) throws IOException {
+        ConnectorHttpClientConfig action = new ConnectorHttpClientConfig(in);
```
ylwu-amzn, line 87: Why use `action` as variable name?

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -43,6 +43,8 @@ public abstract class AbstractConnector implements Connector {
     public static final String BACKEND_ROLES_FIELD = "backend_roles";
     public static final String OWNER_FIELD = "owner";
     public static final String ACCESS_FIELD = "access";
+    public static final String HTTP_CLIENT_CONFIG_FIELD = "http_client_config";
```
ylwu-amzn, line 46: Should we just name it as `config` ?  We may support other protocol in future which is not http

dhrubo-os: renamed it to `client_config`

common/src/main/java/org/opensearch/ml/common/connector/ConnectorHttpClientConfig.java
dhrubo-os: renamed

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -0,0 +1,121 @@
...
+    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
+    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+
+    private Integer maxConnections;
+    private Integer connectionTimeout;
+    private Integer readTimeout;
+
+    @Builder(toBuilder = true)
+    public ConnectorClientConfig(
```
Zhangxunmt, line 39: Should we add a checkup here for the values? It seems all values are allowed to build this clientConfig. In some cases we should block CX to use a super large timeout value to avoid malicious requests?

Zhangxunmt (commented):
Once the ClientConfig is set in the connector, have you checked if this config can be updated using the UpdateConnector API? If not, we should add this field in the Update API to allow updates.

dhrubo-os:
> Once the ClientConfig is set in the connector, have you checked if this config can be updated using the UpdateConnector API? If not, we should add this field in the Update API to allow updates.

Yup, I tested withing providing the following body in the update request. It successfully updates the values.

```
{
  "client_config": {
    "max_connection": 1,
    "connection_timeout": 10000,
    "read_timeout": 10000
  }
  
}
```

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
dhrubo-os: Good point. I'll start a thread on this with our security engineer. If there's anything we need to address, we can do that in the following PR.

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -248,6 +266,13 @@ public void writeTo(StreamOutput output) throws IOException {
         }
         output.writeBoolean(dryRun);
         output.writeBoolean(updateConnector);
+        if (connectorClientConfig != null) {
```
b4sjoo, line 269: Maybe add a version filter here to avoid bwc issue?

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -277,5 +302,8 @@ public MLCreateConnectorInput(StreamInput input) throws IOException {
         }
         dryRun = input.readBoolean();
         updateConnector = input.readBoolean();
+        if (input.readBoolean()) {
```
b4sjoo, line 305: Same here, you might need to consider add a version filter

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
dhrubo-os: We have default values [here](https://github.com/opensearch-project/ml-commons/blob/1dd9f344484be9464b9c937b7ad32ab3ea6b44a6/common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java). Could you please explain a bit more why BWC issue could happen here?

b4sjoo: It's not because if there is value or not, it's because you are introduce new field here. If a master node with this version sends this input to a older data node, it will throw an exception due to cannot process

b4sjoo:
Approved with major concern: we need to add a version filter to check the input/output stream version in order to avoid bwc issue before this code deployed to prod. According to the code owner, in this case he will publish a separate PR for this version filter in 2.x branch, because in main branch 2.13 version is not ready due to some unknown issue.

b4sjoo (approved):

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -156,6 +156,9 @@ public class CommonValue {
                         + AbstractConnector.CREDENTIAL_FIELD
                         + "\" : {\"type\": \"flat_object\"},\n"
                         + "      \""
+                        + AbstractConnector.CLIENT_CONFIG_FIELD
```
ylwu-amzn, line 159: Should bump `ML_CONNECTOR_SCHEMA_VERSION`

ylwu-amzn: Fixed in PR https://github.com/opensearch-project/ml-commons/pull/2228

PR #1845, opened 2024-01-09
Add kotlin stblib dependency for SearchAlertTool

Hailong-am:
### Description
Add kotlin stblib dependency for SearchAlertTool, issue comes from skills repo and need to fix in ml-commons. The reason of this fix described in [issue](https://github.com/opensearch-project/skills/issues/116) 
 
### Issues Resolved
https://github.com/opensearch-project/skills/issues/116
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ buildscript {
         }
 
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
+        kotlin_version = System.getProperty("kotlin.version", "1.8.21")
     }
 
     repositories {
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -24,6 +24,9 @@ dependencies {
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
     implementation "org.opensearch:common-utils:${common_utils_version}"
+    implementation ("org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}") {
+       exclude group: "org.jetbrains", module: "annotations"
+    }
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
     implementation group: 'org.reflections', name: 'reflections', version: '0.9.12'
     implementation group: 'org.tribuo', name: 'tribuo-clustering-kmeans', version: '4.2.1'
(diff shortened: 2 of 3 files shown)


Hailong-am:
https://github.com/opensearch-project/ml-commons/pull/1846 will fix the CI

ohltyler:
@Hailong-am is this the overall right approach? Why should the common utils parent classloader be ML commons? Ideally it should just come from common utils plugin itself to prevent this weird dependency model. Thoughts?

Hailong-am:
> @Hailong-am is this the overall right approach? Why should the common utils parent classloader be ML commons? Ideally it should just come from common utils plugin itself to prevent this weird dependency model, so any explicit classes needed in skills plugin are defined in skills plugin's build.gradle instead of relying on changes in ml commons. Thoughts?

@ohltyler  First of all, each plugin has its own classloader to load all lib/class under its directory and have opensearch core classloader as its parent classloader.

Second, common-utils is a jar lib not a plugin, so there is no common-utils classloader. 
Skills repos is an extension of ml-commons, ML classloader is the parent of skill classloader.

**As `common-uitls` been a dependency of ml-common before skills repos created, skill repo can't not add `common-utils` as runtime dependency otherwise it will have jar shell error.**

based on that, common-utils lib loaded by ml classloader, any missing dependency of common-utils need add to ml-commons, this may not very straightforward.

ohltyler:
> > @Hailong-am is this the overall right approach? Why should the common utils parent classloader be ML commons? Ideally it should just come from common utils plugin itself to prevent this weird dependency model, so any explicit classes needed in skills plugin are defined in skills plugin's build.gradle instead of relying on changes in ml commons. Thoughts?
> 
> @ohltyler First of all, each plugin has its own classloader to load all lib/class under its directory and have opensearch core classloader as its parent classloader.
> 
> Second, common-utils is a jar lib not a plugin, so there is no common-utils classloader. Skills repos is an extension of ml-commons, ML classloader is the parent of skill classloader.
> 
> **As `common-uitls` been a dependency of ml-common before skills repos created, skill repo can't not add `common-utils` as runtime dependency otherwise it will have jar shell error.**
> 
> based on that, common-utils lib loaded by ml classloader, any missing dependency of common-utils need add to ml-commons, this may not very straightforward.

Got it, it makes sense the ML classloader is the parent of skills classloader. and because common-utils is already part of ML commons, this is where that needs to be changed. Still seems slightly odd to need to add a skills-plugin related change here, but this is an edge case, so I'm ok. Thanks for all of the clarifications!

ohltyler (approved):

ylwu-amzn (approved):

Hailong-am:
> Do you mean adding this kotlin dependency in skills repo depends on the classes in common-utils, so it has to be included in ml-commons? This type of fix is not scalable because ml-commons would bloats infinitely due to more and more tools (theoretically infinite) in Skills repo.  

No, just for common utils related or dependency that ml common already has and still missing other dependencies.

If there's a new dependency We can add it skills directly, it will load by skills repo class loader.

Issue #1854, opened 2024-01-11
[BUG] Remove unnecessary dependency on Guava


samuel-oci:
**What is the bug?**
There is an unnecessary dependency on Guava throughout the project that doesn't seem to contribute any capability or make anything more concise. Since all functionalities of Guava used in this project have already been moved to core Java and a few to apache commons which is also a dependency.

This dependency also requires shading the guava artifact and adds multiple exclusion statement and additional unnecessary dependency on shading gradle plugin.
It's also a good idea to remove Guava dependencies to avoid any collisions with runtimes of other plugins.
Moreover the shading is making it difficult for the IDE to detect the dependency on the relocated jar which happens during the repacking but breaks during source code analysis.

**How can one reproduce the bug?**
N/A

**What is the expected behavior?**
N/A

**What is your host/environment?**
N/A

**Do you have any screenshots?**
N/A

samuel-oci:
Please assign to me, I have a PR ready to send.

PR #1855, opened 2024-01-11
Remove guava dependency

sam-herman:
### Description
Removing unnecessary guava dependencies
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1854
 
### Check List
- [ X] New functionality includes testing.
  - [X ] All tests pass
- [X ] New functionality has been documented.
  - [X ] New functionality has javadoc added
- [X ] Commits are signed per the DCO using --signoff 

### DIFF
--- a/memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
@@ -61,8 +61,6 @@ import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.search.sort.SortOrder;
 
-import com.google.common.annotations.VisibleForTesting;
-
 import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 
@@ -275,7 +273,7 @@ public class InteractionsIndex {
         conversationMetaIndex.checkAccess(conversationId, accessListener);
     }
 
-    @VisibleForTesting
+    // VisibleForTesting
     void innerGetInteractions(String conversationId, int from, int maxResults, ActionListener<List<Interaction>> listener) {
         SearchRequest request = Requests.searchRequest(INTERACTIONS_INDEX_NAME);
 
@@ -369,7 +367,7 @@ public class InteractionsIndex {
      * @param maxResults how many interactions to get per search query
      * @param listener receives the list of all interactions in the conversation
      */
-    @VisibleForTesting
+    // VisibleForTesting
     void getAllInteractions(String conversationId, int maxResults, ActionListener<List<Interaction>> listener) {
         ActionListener<List<Interaction>> al = nextGetListener(conversationId, 0, maxResults, listener, new LinkedList<>());
         innerGetInteractions(conversationId, 0, maxResults, al);
@@ -385,7 +383,7 @@ public class InteractionsIndex {
      * @param result partially built list of interactions
      * @return an ActionListener to handle the next search query
      */
-    @VisibleForTesting
+    // VisibleForTesting
     ActionListener<List<Interaction>> nextGetListener(
         String conversationId,
         int from,
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
@@ -18,6 +18,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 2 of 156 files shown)


ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
```diff
@@ -129,9 +127,9 @@ public void executePredict_RemoteInferenceInput_NullResponse() throws IOExceptio
             .url("http://test.com/mock")
             .requestBody("{\"input\": \"${parameters.input}\"}")
             .build();
-        Map<String, String> credential = ImmutableMap
+        Map<String, String> credential = Map
             .of(ACCESS_KEY_FIELD, encryptor.encrypt("test_key"), SECRET_KEY_FIELD, encryptor.encrypt("test_secret_key"));
-        Map<String, String> parameters = ImmutableMap.of(REGION_FIELD, "us-west-2", SERVICE_NAME_FIELD, "sagemaker");
+        Map<String, String> parameters = Map.of(REGION_FIELD, "us-west-2", SERVICE_NAME_FIELD, "sagemaker");
```
Zhangxunmt, line 132: Can we keep this ImmutableMap at least for these ones? It doesn't look like we want to update them in the code. We can use the equivalent in aws sdk through "import software.amazon.awssdk.utils.ImmutableMap"

plugin/src/main/java/org/opensearch/ml/action/forward/TransportForwardAction.java
```diff
@@ -12,10 +12,7 @@
 import static org.opensearch.ml.utils.MLExceptionUtils.toJsonString;
 
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
Zhangxunmt: +1.

Zhangxunmt (commented):
Did you compare the opensearch-ml zip size before and after this removal? We can remove it for now but we probably will add some of them back to use Guava cache for better cache management if the zip size doesn't change much.

memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
```diff
@@ -275,7 +273,7 @@ public void getInteractions(String conversationId, int from, int maxResults, Act
         conversationMetaIndex.checkAccess(conversationId, accessListener);
     }
 
-    @VisibleForTesting
+    // VisibleForTesting
```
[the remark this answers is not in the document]
samuel-oci: @dhrubo-os curious what are the advantages you see of `ImmutableMap` over `Map`? If it's the immutability you seek than `Map.of()` also creates an immutable collection.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
Zhangxunmt: Good question. The Map.of() only gives you the very basic immutableMap which is not convenient to handle Null value, factory pattern, builder pattern etc. I think you will find more advantages by searching the difference online. because AWS already provided the same ImmutableMap, why not use it without adding any new dependencies?

samuel-oci: I think it's better to use core language unless there is a clear reason to add new external dependency with a clear use case to show why those are needed. So far I haven't seen one and honestly the shaded repackage thing breaks my IDE (besides mess of exclusions in the gradle.build), which is kinda annoying as well.

Specifically regarding the AWS SDK vs Guava, I would rather keep Guava rather than rely on AWS SDK implementation.

With that being said, if the community doesn't want to change this dependency I'm fine abandoning this as well.

Hailong-am: > shaded repackage thing breaks my IDE

try to using latest IntelliJ IDEA version of 2023.3.2, the shadow jar was fix in this version.

samuel-oci:
> Did you compare the opensearch-ml zip size before and after this removal? We can remove it for now but we probably will add some of them back to use Guava cache for better cache management if the zip size doesn't change much.

Hi @Zhangxunmt this PR is not really have the zip size optimization as it's main focus, but rather it was more meant about making the build file clean from exclusion and shading acrobatics and avoiding dependencies that can collide with opensearch runtime.

Now if we intend on keeping Guava going forward for various other reasons (e.g. Cache) then should we use runtime dependency instead of compile time?
You can pass the guava to testImplementation instead and avoid the exclusions and shading?

ylwu-amzn:
Some new PR merged to main branch. Rebase first to make the CI pass?

Zhangxunmt:
@samuel-oci Can you rebase this PR and merge? It's been a long time. Not sure if it has new conflicts with recent commits. Currently still no plan to use guava for cache.

memory/src/main/java/org/opensearch/ml/memory/index/InteractionsIndex.java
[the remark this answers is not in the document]
dhrubo-os: Immutability, thread-safe, better performance

dhrubo-os:
@sam-herman are you still working on this issue? We are planning to close this PR if there's no update/progress in this month of January. Thanks.

Issue #1787, opened 2023-12-19
[BUG] When Hitting ML Connector Search Endpoint, returns index not found


LucasWang750:
**What is the bug?**
A clear and concise description of the bug.
When Hitting ML Connector Search Endpoint, returns index not found

**How can one reproduce the bug?**
1.Run POST GET /_plugins/_ml/connectors/_search before connector is created 
2.Returns 404 error :
An error occurred while initializing cluster bootstrap: NotFoundError(404, 'index_not_found_exception', 'no such index [.plugins-ml-connector]', .plugins-ml-connector, index_or_alias)

**What is the expected behavior?**
Return no connectors

**What is your host/environment?**
 - OS: Mac OS X
 - Version: OpenSearch 2.11
 - Plugins

Zhangxunmt:
This behavior is expected because you haven't created a connector yet. The connector index is created when the first connector is created. This is not a service error so the error code 404 is returned. This is like you visiting a web service that hasn't been created yet.

ylwu-amzn:
The error `NotFoundError(404, 'index_not_found_exception', 'no such index [.plugins-ml-connector]', .plugins-ml-connector, index_or_alias)` expose the implementation details and looks confusing to community user. I think we can fine tune the error message to make it easier to understand like "no connector created"

ylwu-amzn:
@samuel-oci will help on this

samuel-oci:
Sure thing @ylwu-amzn feel free to assign it to me.

PR #1870, opened 2024-01-13
Issue 1787 fixing connector endpoint returns index not found

samuel-oci:
### Description
When connector is not created yet, return an empty search response instead of a broken 404 with cryptic index not found message.
More description in: https://github.com/opensearch-project/ml-commons/issues/1787

Minor additional issues fixed:
- Ability to run integTest with cluster debugging. 
For example:
`./gradlew :opensearch-ml-plugin:integTest --tests org.opensearch.ml.rest.RestMLSearchConnectorActionIT.testSearchConnector_beforeConnectorCreated -Dcluster.debug=1`
- Show log output during integTest

### Issues Resolved
[[List any issues this PR will resolve]](https://github.com/opensearch-project/ml-commons/issues/1787)
 
### Check List
  - [x ] All tests pass
- [x ] New functionality has been documented.
  - [x ] New functionality has javadoc added

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -189,6 +189,10 @@ integTest {
     if (System.getProperty("test.debug") != null) {
         jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'
     }
+
+    // Set this to true this if you want to see the logs in the terminal test output.
+    // note: if left false the log output will still show in your IDE
+    testLogging.showStandardStreams = true
 }
 
 testClusters.integTest {
@@ -198,7 +202,7 @@ testClusters.integTest {
     // When running integration tests it doesn't forward the --debug-jvm to the cluster anymore
     // i.e. we have to use a custom property to flag when we want to debug elasticsearch JVM
     // since we also support multi node integration tests we increase debugPort per node
-    if (System.getProperty("opensearch.debug") != null) {
+    if (System.getProperty("cluster.debug") != null) {
         def debugPort = 5005
         nodes.forEach { node ->
             node.jvmArgs("-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=*:${debugPort}")
--- a/plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
@@ -84,19 +84,22 @@ public class SearchConnectorTransportAction extends HandledTransportAction<Searc
 
             ActionListener<SearchResponse> doubleWrappedListener = ActionListener.wrap(wrappedListener::onResponse, e -> {
                 if (e instanceof IndexNotFoundException) {
-                    log.debug("Connectors index not created yet, therefore we will swallow the exception and return an empty search result");
+                    log
+                        .debug(
+                            "Connectors index not created yet, therefore we will swallow the exception and return an empty search result"
+                        );
                     final InternalSearchResponse internalSearchResponse = InternalSearchResponse.empty();
                     final SearchResponse emptySearchResponse = new SearchResponse(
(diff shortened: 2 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
```diff
@@ -78,12 +81,38 @@ private void search(SearchRequest request, ActionListener<SearchResponse> action
                 excludes.toArray(new String[0])
             );
             request.source().fetchSource(rebuiltFetchSourceContext);
+
+            ActionListener<SearchResponse> doubleWrappedListener = ActionListener.wrap(wrappedListener::onResponse, e -> {
+                if (e instanceof IndexNotFoundException) {
```
Hailong-am, line 86: better to use `ExceptionHelp.unwrap` to unwrap the exception for multiple node case.

plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
```diff
@@ -78,12 +81,38 @@ private void search(SearchRequest request, ActionListener<SearchResponse> action
...
+                        0,
+                        0,
+                        0,
+                        0,
+                        null,
+                        new ShardSearchFailure[] {},
+                        SearchResponse.Clusters.EMPTY,
+                        null
+                    );
+                    wrappedListener.onResponse(emptySearchResponse);
```
Hailong-am, line 104: return empty result with status 200 may confusing user, better to keep rest status as 404 with customized error message.

samuel-oci: > return empty result with status 200 may confusing user, better to keep rest status as 404 with customized error message.

Why should user get or expect to get 404 if all they did was to ask which connectors are available?
The return of empty result is inline with the expected behavior described in the issue https://github.com/opensearch-project/ml-commons/issues/1787

plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
Hailong-am: just update to `unwrapCause`

samuel-oci: Sounds good, will do

samuel-oci: @Hailong-am done

ylwu-amzn:
The code looks good to me. How about enhancing other search APIs too:  search model / model group/ tasks?

samuel-oci:
> The code looks good to me. How about enhancing other search APIs too: search model / model group/ tasks?

@ylwu-amzn sounds good, would you like me to create issues for those as well?

plugin/build.gradle
```diff
@@ -198,7 +202,7 @@ testClusters.integTest {
     // When running integration tests it doesn't forward the --debug-jvm to the cluster anymore
     // i.e. we have to use a custom property to flag when we want to debug elasticsearch JVM
     // since we also support multi node integration tests we increase debugPort per node
-    if (System.getProperty("opensearch.debug") != null) {
+    if (System.getProperty("cluster.debug") != null) {
```
Hailong-am, line 205: have you tried with option `-debug-jvm` that should work also.

Hailong-am (approved):

plugin/build.gradle
samuel-oci: Yeah, I did try that. But unfortunately it doesn't seem to be passed to the actual cluster being spun by the integ test.
So the test itself is debuggable but not the cluster spun by it.

HenryL27 (approved):

ylwu-amzn (approved):

PR #1882, opened 2024-01-17
Fix long pending issue when deleting model

zane-neo:
### Description
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ylwu-amzn:
Github CI failed
```
[ant:jacocoReport] Rule violated for class org.opensearch.ml.action.models.DeleteModelTransportAction.1: branches covered ratio is 0.5, but expected minimum is 0.7
```

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -183,13 +185,7 @@ void deleteModelChunks(String modelId, DeleteResponse deleteResponse, ActionList
...
                 && (r.getSearchFailures() == null || r.getSearchFailures().size() == 0)) {
                 log.debug("All model chunks are deleted for model {}", modelId);
-                if (deleteResponse != null) {
-                    // If model metaData not found and deleteResponse is null, do not return here.
-                    // ResourceNotFound is returned to notify that this model was deleted.
-                    // This is a walk around to avoid cleaning up model leftovers. Will revisit if
-                    // necessary.
-                    actionListener.onResponse(deleteResponse);
-                }
+                actionListener.onResponse(deleteResponse);
```
ylwu-amzn, line 188: So the long pending is caused by the `deleteResponse` is null case ? Curious when the `deleteResponse` will be null?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -212,22 +208,22 @@ private void returnFailure(BulkByScrollResponse response, String modelId, Action
...
+                } else {
+                    actionListener.onResponse(deleteResponse);
+                }
                 deleteController(modelId);
             }
 
             @Override
             public void onFailure(Exception e) {
                 log.error("Failed to delete model meta data for model: " + modelId, e);
-                if (e instanceof ResourceNotFoundException) {
```
ylwu-amzn, line 227: I think we should not delete this part. This is for hand edge cases that model meta doc deleted but failed to delete model chunks.

zane-neo: When catching exception: https://github.com/opensearch-project/ml-commons/blob/6bff20d5ea2dd16d40904d7eef1c22ffeda5b9ea/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L228

zane-neo: Can we return error response and give user the guidance how to clean up instead of we handle this for user, you can't ensure the clean up can  be succeed.

ylwu-amzn: > This is for hand edge cases that model meta doc deleted but failed to delete model chunks.

Not quite get you, what guidance we should give to user if model meta deleted ?

zane-neo: I think the correct direction is: When deleting a model, we should delete model meta and chunks and controller, only when all three are deleted we return success, otherwise we return failure, and tell user `model chunks or controller not cleaned up, please try again.`, and user can try delete model chunks and controller until succeed.

ylwu-amzn: Make sense. I think the old logic is the same with your idea, we delete model meta first, then delete chunks, if anyone failed, we throw exception. Took a look at code, the model controller part will not throw exception when delete model. You can fix this.

ylwu-amzn: DeleteModelTransportAction.java
` deleteController(modelId);` doesn't have `actionListener` , that looks not reasonable 
```
private void deleteModel(String modelId, ActionListener<DeleteResponse> actionListener) {
        DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_INDEX, modelId);
        client.delete(deleteRequest, new ActionListener<DeleteResponse>() {
            @Override
            public void onResponse(DeleteResponse deleteResponse) {
                deleteModelChunks(modelId, deleteResponse, actionListener);
                deleteController(modelId);
            }

            @Override
            public void onFailure(Exception e) {
                log.error("Failed to delete model meta data for model: " + modelId, e);
                if (e instanceof ResourceNotFoundException) {
                    deleteModelChunks(modelId, null, actionListener);
                    deleteController(modelId);
                }
                actionListener.onFailure(e);
            }
        });
    }
```

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -212,22 +210,68 @@ private void returnFailure(BulkByScrollResponse response, String modelId, Action
         actionListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.INTERNAL_SERVER_ERROR));
     }
 
-    private void deleteModel(String modelId, ActionListener<DeleteResponse> actionListener) {
-        DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_INDEX, modelId);
-        client.delete(deleteRequest, new ActionListener<DeleteResponse>() {
+    private void deleteModel(String modelId, FunctionName functionName, ActionListener<DeleteResponse> actionListener) {
+        // Always delete model chunks and model controller first, because deleting metadata first user is not able clean up model chunks and
```
ylwu-amzn, line 214: This could have such issue: model content deleted, but model meta still there. User can still see this model with search API. Then user may want to deploy this model again. It will throw error which may confuse user. 

That's why we want to delete model meta first. If model chunk deleted failed, we throw error and user need to re-delete the model, that's why we catch the not found exception when delete model meta and re-delete the model chunk in original code.

zane-neo: If we delete model metadata first and model chunk or controller are not deleted, then there's no chance user can delete these, since the getRequest won't query results back: https://github.com/opensearch-project/ml-commons/blob/c53d586e50d7d38ccd4d5448c09bc015cce49992/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L104C24-L104C34. 
For deploy model, the model query also exclude the model content: https://github.com/opensearch-project/ml-commons/blob/c53d586e50d7d38ccd4d5448c09bc015cce49992/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java#L1444, I think it should be fine, but even if this has problem, I think we'd better to add message in deploy API.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -214,47 +214,101 @@ private void returnFailure(BulkByScrollResponse response, String modelId, Action
...
+            CountDownLatch countDownLatch = new CountDownLatch(2);
+            AtomicBoolean bothDeleted = new AtomicBoolean(true);
+            ActionListener<Boolean> countDownActionListener = ActionListener.wrap(b -> {
+                countDownLatch.countDown();
+                bothDeleted.compareAndSet(true, b);
+                if (countDownLatch.getCount() == 0) {
+                    if (bothDeleted.get()) {
+                        log.debug("model chunks and model controller for model {} deleted successfully", modelId);
+                        if (deleteResponse != null) {
+                            actionListener.onResponse(deleteResponse);
```
ylwu-amzn, line 253: Seems `deleteResponse` could be null, will it cause any error? What information user can see for null?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -214,47 +214,101 @@ private void returnFailure(BulkByScrollResponse response, String modelId, Action
...
+                        } else {
+                            actionListener.onFailure(new OpenSearchStatusException("Failed to find model", RestStatus.NOT_FOUND));
+                        }
+                    } else {
+                        actionListener
+                            .onFailure(
+                                new IllegalStateException("Failed to delete model chunks or model controller, please try again: " + modelId)
```
ylwu-amzn, line 260: This looks complicated for user to understand which one failed. 
How about just delete model chunk first, then delete model controller ? Throw exception if any delete fails

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -214,47 +214,101 @@ private void returnFailure(BulkByScrollResponse response, String modelId, Action
...
         });
     }
 
+    private void deleteModelChunksAndController(
+        FunctionName functionName,
+        ActionListener<DeleteResponse> actionListener,
+        String modelId,
+        DeleteResponse deleteResponse
+    ) {
+        if (FunctionName.REMOTE != functionName) {
```
ylwu-amzn, line 243: How about simplify this just always delete chunks and controllers?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: Only when model not found, deleteResponse will be null, in this case we try to delete model chunks and model controller, and then return 400 "model not found" to user.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: We know remote model doesn't have chunks, the query to it's its chunk can be saved to improve a little bit performance. But I am ok to remove this check if you prefer to not check this.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: How about refine this error message to more generic, e.g. "model is not all cleaned up, please try again"?

ylwu-amzn: Sounds good

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
ylwu-amzn: I think adding another delete won't bring a lot impact for performance, it can make the code easier to maintain without such if else

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -174,7 +175,7 @@ protected MLUndeployModelNodesResponse newResponse(
                             deployToAllNodes.put(modelId, false);
                         }
                         updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(updateDocument);
-                        bulkRequest.add(updateRequest);
+                        bulkRequest.add(updateRequest).setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
```
ylwu-amzn, line 178: Do you see anywhere else missed this?

ylwu-amzn (approved):

model-collapse (approved):

PR #1893, opened 2024-01-21
system error handling

jngz-es:
### Description
Catch all system error from agent when executing and wrap them into ErrorMessage to reply.
Tested on local testing cluster.

Some examples:

Old:
{
  "error": {
    "root_cause": [
      {
        "reason": "Tool not found: VectorDBTool"
      }
    ],
    "reason": "Tool not found: VectorDBTool"
  },
  "status": 400
}

New:
"{\n  \"error\": {\n    \"reason\": \"Invalid Request\",\n    \"details\": \"Tool not found: VectorDBTool\",\n    \"type\": \"IllegalArgumentException\"\n  },\n  \"status\": 400\n}"

Old:
{
  "inference_results": [
    {
      "output": [
        {
          "result": """ ..."""
        }
      ]
    }
  ]
}

New:
{
  "inference_results": [
    {
      "output": [
        {
          "name": "response",
          "result": "..."
        }
      ]
    }
  ]
}
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


plugin/src/main/java/org/opensearch/ml/rest/RestMLExecuteAction.java
```diff
@@ -87,4 +116,16 @@ MLExecuteTaskRequest getRequest(RestRequest request) throws IOException {
...
+    private void sendResponse(RestChannel channel, MLExecuteTaskResponse response) throws Exception {
+        channel.sendResponse(new RestToXContentListener<MLExecuteTaskResponse>(channel).buildResponse(response));
+    }
+
+    private void reportError(final RestChannel channel, final Exception e, final RestStatus status) {
+        channel.sendResponse(new BytesRestResponse(status, ErrorMessageFactory.createErrorMessage(e, status.getStatus()).toString()));
+    }
+
+    private static boolean isClientError(Exception e) {
+        return e instanceof NullPointerException || e instanceof IllegalArgumentException || e instanceof IllegalAccessException;
```
mingshl, line 129: trying to understand why nullPointerException isClientError?

jngz-es: Yeah, it is hard to differentiate but more likely caused by bad request, something like missing configuration etc.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.utils.error;
+
+import org.json.JSONObject;
```
Hailong-am, line 8: I am not sure whether we should leverage `ToXContentObject` interface which provided by OpenSearch to do json/yml convert and try to not include a new lib.

plugin/src/main/java/org/opensearch/ml/rest/RestMLExecuteAction.java
ylwu-amzn: I think it's more safe to remove null pointer exception. If that NPE error is from our code, we should not say this is caused by client side

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
jngz-es: I just tried to use ToXContentObject but the related methods would throw exception again, so I want to keep this raw json handling.

plugin/src/main/java/org/opensearch/ml/rest/RestMLExecuteAction.java
jngz-es: OK, delete NPE from client error.

plugin/build.gradle
```diff
@@ -63,6 +63,7 @@ dependencies {
     implementation group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.19.0"
+    implementation group: 'org.json', name: 'json', version: '20231013'
```
saratvemulapalli, line 66: Lets use jackson databind which is already packaged currently[1] with ml-commons ? 
We try to avoid using multiple JSON creators and parsers in the core. 

[1] https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java#L50

jngz-es: Makes sense. Just removed this dependency and used ObjectMapper instead of JsonObject.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
jngz-es: As Sarat suggested, I removed json library dependency and used ObjectMapper.

ylwu-amzn (approved):

b4sjoo (approved):

PR #1921, opened 2024-01-25
Fix string.format wrong parameter position

Hailong-am:
### Description
Fix string.format wrong parameter position, the Locale should be first parameter
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/MLEngine.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/MLEngine.java
@@ -58,7 +58,7 @@ public class MLEngine {
 
     public String getPrebuiltModelConfigPath(String modelName, String version, MLModelFormat modelFormat) {
         String format = modelFormat.name().toLowerCase(Locale.ROOT);
-        return String.format("%s/%s/%s/%s/config.json", MODEL_REPO, modelName, version, format, Locale.ROOT);
+        return String.format(Locale.ROOT, "%s/%s/%s/%s/config.json", MODEL_REPO, modelName, version, format);
     }
 
     public String getPrebuiltModelPath(String modelName, String version, MLModelFormat modelFormat) {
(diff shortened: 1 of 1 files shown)


Hailong-am:
@ylwu-amzn Search the whole codebase also, there have few places use string.format without specifying Locale, do you think we should add for those?

ml-algorithms/src/main/java/org/opensearch/ml/engine/MLEngine.java
```diff
@@ -58,7 +58,7 @@ public String getPrebuiltModelMetaListPath() {
 
     public String getPrebuiltModelConfigPath(String modelName, String version, MLModelFormat modelFormat) {
         String format = modelFormat.name().toLowerCase(Locale.ROOT);
-        return String.format("%s/%s/%s/%s/config.json", MODEL_REPO, modelName, version, format, Locale.ROOT);
+        return String.format(Locale.ROOT, "%s/%s/%s/%s/config.json", MODEL_REPO, modelName, version, format);
```
samuel-oci, line 61: Can you add a simple UT that shows the behavior that is fixed? It's usually a common industry practice when fixing a bug to make sure that there is going forward a UT that clearly document the behavior change.

samuel-oci (changes requested):

ml-algorithms/src/main/java/org/opensearch/ml/engine/MLEngine.java
Hailong-am: Actually, this fix didn't change the behavior of the method. Put locale at last that will ignored by String.format method as parameters length is greater than conversions length, the superfluous one will ignored.

For UT since there already have [one](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java#L70) which will cover this change.

samuel-oci: Sounds good, another comment, might not be directly related to this change, but as a future FYI
this method seem like it should be static because it has no dependency on object level instance.
If we want to keep it non-static we should leverage existing parameters in the object instance instead.

HenryL27 (approved):

ylwu-amzn:
> @ylwu-amzn Search the whole codebase also, there have few places use string.format without specifying Locale, do you think we should add for those?

Thanks, we should add this

ylwu-amzn (approved):

Issue #1839, opened 2024-01-05
[FEATURE] Replace blocking httpclient with async httpclient in remote inference


zane-neo:
**Is your feature request related to a problem?**
Community user brings up a performance issue [here](https://forum.opensearch.org/t/how-to-register-sparse-encoding-model-in-aws-opensearch/17112/9?u=zane_neo), which reveals a performance issue in HttpClient of remote inference. The flow of the prediction can be illustrated as below:
![blocking-httpclient drawio (1)](https://github.com/opensearch-project/ml-commons/assets/114061051/e0919ea8-4b57-44bc-94b6-a0b3f11c394e)

There are two major issues here: 
1. The connection pool size of HttpClient is 20 by default which can cause `timeout waiting for connection` described here: https://github.com/opensearch-project/ml-commons/issues/1537.
2. The blocking HttpClient is a bottleneck since the predict pool thread size by default is `2 * num of vCPUs`, this isn't a big value since local model prediction is a CPU bound operation. But for remote inference, it's an IO bound operation and the thread pool size is relatively small.
For issue1, we can enable user to update the configuration of max_connections to handle more parallel predict requests.
For issue2, we can increase the predict thread pool size to bigger number to increase the parallelism, but this is not optimal because more threads would cause more context switch and degrade the overall system performance.

**What solution would you like?**
Replace the blocking HttpClient with async HttpClient. With async HttpClient, both two issues above can be handled perfectly, there's no connection pool in async HttpClient and we don't need to change the default predict thread pool size since async HttpClient has better performance with only a few threads.
AWS async HttpClient: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration-crt.html

**What alternatives have you considered?**
Increasing the predict thread pool size and make this a system setting and configurable to user.

**Do you have any additional context?**
NA

zane-neo:
@model-collapse @ylwu-amzn @dhrubo-os @austintlee Please chime in.

dhrubo-os:
> there's no connection pool in async HttpClient

Could you please explain why? In your provided link, I can see `maxConcurrency(100)` was set for both (async and sync)

zane-neo:
A mistake here, async httpclient also has connection pools, the 100 is an example value.

zane-neo:
Benchmark results of replacing sync http-client to async

Test settings

commons settings

* Benchmark doc count: 100k
[...]

PR #1958, opened 2024-01-30
Change httpclient to async

zane-neo:
### Description
Change sync httpclient to async to improve remote inference performance
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1839
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/test/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactoryTests.java
```diff
@@ -6,99 +6,145 @@
...
         try {
-            MLHttpClientFactory.validateIp("0254.020.00.01");
-        } catch (IllegalArgumentException e) {
+            SdkHttpFullRequest request = SdkHttpFullRequest
+                .builder()
+                .method(POST)
+                .uri(URI.create("http://0177.1/v1/completions"))
+                .contentStreamProvider(requestBody.contentStreamProvider())
+                .build();
+        } catch (NullPointerException e) {
```
ylwu-amzn, line 54: Will this throw 500 error ? Seems not correct, as it caused by wrong input

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -62,74 +54,38 @@ public class AwsConnectorExecutor implements RemoteConnectorExecutor {
...
-            Map<String, String> headers = connector.getDecryptedHeaders();
-            if (headers != null) {
-                for (String key : headers.keySet()) {
-                    builder.putHeader(key, headers.get(key));
-                }
-            }
-            SdkHttpFullRequest request = builder.build();
-            HttpExecuteRequest executeRequest = HttpExecuteRequest
+            SdkHttpFullRequest request = ConnectorUtils.buildSdkRequest(connector, parameters, payload, POST);
+            MLHttpClientFactory.validateIp(request.getUri().getHost());
```
ylwu-amzn, line 73: Add more comments?

ml-algorithms/src/main/java/org/opensearch/ml/engine/Predictable.java
```diff
@@ -31,7 +34,13 @@ public interface Predictable {
      * @param mlInput input data
      * @return predicted results
      */
-    MLOutput predict(MLInput mlInput);
+    default MLOutput predict(MLInput mlInput) {
+        throw new IllegalStateException("Method is not implemented");
+    }
+
+    default void predict(MLInput mlInput, MLTask mlTask, ActionListener<MLTaskResponse> actionListener) {
+
```
ylwu-amzn, line 42: Add default action? 
```
actionListener.onFailure(new IllegalStateException("Method is not implemented")); 
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Override
+    public void onError(Throwable error) {
+        log.error(error.getMessage(), error);
+        actionListener
+            .onFailure(
+                new OpenSearchStatusException(
+                    "Error on communication with remote model: " + error.getMessage(),
+                    RestStatus.INTERNAL_SERVER_ERROR
```
ylwu-amzn, line 93: Why always return this internal server error?

ylwu-amzn: Can we translate the status code to this RestStatus

ml-algorithms/src/main/java/org/opensearch/ml/engine/Predictable.java
zane-neo: Added and changed method parameter to remove unused mlTask and renamed the method to avoid conflict with `public MLOutput predict(MLInput mlInput, MLModel model)`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: Done

ml-algorithms/src/test/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactoryTests.java
zane-neo: Changed the validate code a little bit to make it more compatible with existing UTs

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -245,4 +260,52 @@ public static SdkHttpFullRequest signRequest(
...
+        Connector connector,
+        Map<String, String> parameters,
+        String payload,
+        SdkHttpMethod method
+    ) throws UnknownHostException {
+        String endpoint = connector.getPredictEndpoint(parameters);
+        Pattern pattern = Pattern.compile("(?:(\\w+)://)?((\\w+\\.)*\\w+)(?::(\\w+))?");
+        Matcher matcher = pattern.matcher(endpoint);
+        if (matcher.find()) {
+            String protocol = matcher.group(1);
```
dhrubo-os, line 274: This seems bit brittle. What do you think about something like this:

```
String endpoint = "https://example.com:8080/path/to/resource?query=123";
            URL url = new URL(endpoint);

            // Getting protocol, host, and port
            String protocol = url.getProtocol();
            String host = url.getHost();
            int port = url.getPort(); // This will return -1 if the port is not explicitly specified in the URL
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -61,83 +56,50 @@ public class HttpJsonConnectorExecutor implements RemoteConnectorExecutor {
...
-                        String predictEndpoint = connector.getPredictEndpoint(parameters);
-                        request = new HttpPost(predictEndpoint);
-                        String charset = parameters.containsKey("charset") ? parameters.get("charset") : "UTF-8";
-                        HttpEntity entity = new StringEntity(payload, charset);
-                        ((HttpPost) request).setEntity(entity);
-                    } catch (Exception e) {
-                        throw new MLException("Failed to create http request for remote model", e);
-                    }
+                    log.debug("original payload to remote model: " + payload);
+                    request = ConnectorUtils.buildSdkRequest(connector, parameters, payload, POST);
```
dhrubo-os, line 80: add test?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -29,52 +32,63 @@
...
+        ActionListener<List<ModelTensors>> tensorActionListener = ActionListener.wrap(r -> {
+            actionListener.onResponse(new MLTaskResponse(new ModelTensorOutput(r)));
+        }, actionListener::onFailure);
+        Map<Integer, ModelTensors> modelTensors = new ConcurrentHashMap<>();
         if (mlInput.getInputDataset() instanceof TextDocsInputDataSet) {
             TextDocsInputDataSet textDocsInputDataSet = (TextDocsInputDataSet) mlInput.getInputDataset();
-            int processedDocs = 0;
-            while (processedDocs < textDocsInputDataSet.getDocs().size()) {
+            Tuple<Integer, Integer> calculatedChunkSize = calculateChunkSize(textDocsInputDataSet);
+            CountDownLatch countDownLatch = new CountDownLatch(calculatedChunkSize.v1());
```
dhrubo-os, line 48: Can we add detailed comment regarding `calculatedChunkSize.v1()` vs `calculatedChunkSize.v2()`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,63 @@
...
-    public static CloseableHttpClient getCloseableHttpClient() {
-        return createHttpClient();
+    public static SdkAsyncHttpClient getAsyncHttpClient() {
+        try {
+            return AccessController
+                .doPrivileged(
+                    (PrivilegedExceptionAction<SdkAsyncHttpClient>) () -> NettyNioAsyncHttpClient.builder().maxConcurrency(100).build()
+                );
+        } catch (PrivilegedActionException e) {
+            return null;
```
dhrubo-os, line 34: what happens when we return null? Add test?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,63 @@
...
-                int port = Integer.parseInt(hostNamePort[1]);
-                if (port < 0 || port > 65536) {
-                    log.error("Remote inference port out of range: " + port);
-                    throw new IllegalArgumentException("Port out of range: " + port);
-                }
-            }
-        } else {
-            log.error("Remote inference scheme not supported: " + scheme);
-            throw new IllegalArgumentException("Unsupported scheme: " + scheme);
+        if (!NumberUtils.isDigits(portStr)) {
```
dhrubo-os, line 50: why can't we return `80` or `443` in the first place? Why do we need to return as string?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -61,83 +56,50 @@ public class HttpJsonConnectorExecutor implements RemoteConnectorExecutor {
...
+        Map<Integer, ModelTensors> tensorOutputs,
+        WrappedCountDownLatch countDownLatch,
+        ActionListener<List<ModelTensors>> actionListener
+    ) {
         try {
-            AtomicReference<String> responseRef = new AtomicReference<>("");
-            AtomicReference<Integer> statusCodeRef = new AtomicReference<>();
-
-            HttpUriRequest request;
+            SdkHttpFullRequest request;
```
khoaisohd, line 76: I though this is protocol is for general HTTP, why we are using AWS SDK request?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,63 @@
...
+            return AccessController
+                .doPrivileged(
+                    (PrivilegedExceptionAction<SdkAsyncHttpClient>) () -> NettyNioAsyncHttpClient.builder().maxConcurrency(100).build()
+                );
+        } catch (PrivilegedActionException e) {
+            return null;
+        }
     }
 
-    private static CloseableHttpClient createHttpClient() {
```
khoaisohd, line 35: Are we going to stop supporting HTTP protocol?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+import com.google.gson.GsonBuilder;
+
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+import software.amazon.awssdk.http.SdkHttpFullResponse;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.http.async.SdkAsyncHttpResponseHandler;
+
+@Log4j2
+public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandler {
```
khoaisohd, line 45: This is used in RemoteConnectorExecutor for all protocols. What happened we are introducing a new connector protocol that has nothing to do with AWS SDK protocol

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -245,4 +260,52 @@ public static SdkHttpFullRequest signRequest(
...
     }
+
+    public static SdkHttpFullRequest buildSdkRequest(
+        Connector connector,
+        Map<String, String> parameters,
+        String payload,
+        SdkHttpMethod method
+    ) throws UnknownHostException {
+        String endpoint = connector.getPredictEndpoint(parameters);
+        Pattern pattern = Pattern.compile("(?:(\\w+)://)?((\\w+\\.)*\\w+)(?::(\\w+))?");
```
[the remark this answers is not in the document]
zane-neo: I don't think it's necessary since this is only used in this method and appears once, only when it's been using for multiple places I think we need to move it to class static variable.

ml-algorithms/build.gradle
```diff
@@ -63,12 +63,13 @@ dependencies {
         }
     }
 
-    implementation platform('software.amazon.awssdk:bom:2.21.15')
+    implementation platform('software.amazon.awssdk:bom:2.22.12')
```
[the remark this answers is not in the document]
zane-neo: It's a upgrade, from 2.21 to 2.22.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: Seems good, will test this approach.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
zane-neo: Previously we have separate code for aws and general http/https, but in fact they can be merged together, aws sdk request only have different headers with general one's, so we can use aws sdk request for general ones as well.

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
zane-neo: They're merged into one implementation, http/https still supported.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: Same above, they're merged together.

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
zane-neo: Changed this as the protocol, host and port retrieval approach changed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/Predictable.java
```diff
@@ -31,7 +33,13 @@ public interface Predictable {
      * @param mlInput input data
      * @return predicted results
      */
-    MLOutput predict(MLInput mlInput);
+    default MLOutput predict(MLInput mlInput) {
+        throw new IllegalStateException("Method is not implemented");
+    }
+
+    default void asyncPredict(MLInput mlInput, ActionListener<MLTaskResponse> actionListener) {
+        throw new IllegalStateException("Method is not implemented");
```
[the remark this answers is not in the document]
zane-neo: added

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
zane-neo: The inference request will return MLException here: https://github.com/opensearch-project/ml-commons/pull/1958/files#diff-6c364bbd919d4840502401e3ada03134cd3c3f71571cbd458c4e2a3aea0da745R87 and https://github.com/opensearch-project/ml-commons/pull/1958/files#diff-dbff9b113bf7a6cdba66b054415ce4f40be8499f4164b7b2c72de815294c81e9R102. It's not easy to add UT for this since we don't have a convenient way to mock static methods.

dhrubo-os: [isSuperAdminUser](https://github.com/opensearch-project/ml-commons/blob/0903d5da4bc9fb8051621de05759dbdd36613972/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java#L228) is an static method. Corresponding test: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/test/java/org/opensearch/ml/utils/RestActionUtilsTests.java#L294-L310


While using this static method to other class, I created a stub ([example](https://github.com/opensearch-project/ml-commons/blob/0903d5da4bc9fb8051621de05759dbdd36613972/plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java#L162-L166)), Now we can use that stubbing method in test. ([example](https://github.com/opensearch-project/ml-commons/blob/0903d5da4bc9fb8051621de05759dbdd36613972/plugin/src/test/java/org/opensearch/ml/action/models/GetModelTransportActionTests.java#L151))

zane-neo: Added tests for this case.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -245,4 +258,47 @@ public static SdkHttpFullRequest signRequest(
...
+        int port = url.getPort();
+        MLHttpClientFactory.validate(protocol, host, port);
+        String charset = parameters.getOrDefault("charset", "UTF-8");
+        RequestBody requestBody;
+        if (payload != null) {
+            requestBody = RequestBody.fromString(payload, Charset.forName(charset));
+        } else {
+            requestBody = RequestBody.empty();
+        }
+        if (SdkHttpMethod.POST == method && "0".equals(requestBody.optionalContentLength().get().toString())) {
```
dhrubo-os, line 281: `"0".equals(requestBody.optionalContentLength().get().toString())` --> Why do we need to convert to string to check?

dhrubo-os: if `RequestBody.empty()`, what do we get from `requestBody.optionalContentLength().get()`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -61,83 +56,50 @@ public class HttpJsonConnectorExecutor implements RemoteConnectorExecutor {
...
+                .build();
+            AccessController.doPrivileged((PrivilegedExceptionAction<CompletableFuture<Void>>) () -> httpClient.execute(executeRequest));
         } catch (RuntimeException e) {
             log.error("Fail to execute http connector", e);
-            throw e;
+            actionListener.onFailure(e);
         } catch (Throwable e) {
             log.error("Fail to execute http connector", e);
-            throw new MLException("Fail to execute http connector", e);
+            actionListener.onFailure(new MLException("Fail to execute http connector", e));
```
dhrubo-os, line 102: IS that 500 level or 400 level exception?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+                                RestStatus.fromCode(status)
+                            )
+                        );
+                } catch (PrivilegedActionException e) {
+                    actionListener.onFailure(new OpenSearchStatusException(e.getMessage(), RestStatus.fromCode(statusCode)));
+                }
+            }
+        } else {
+            // non batch API.
+            for (Map.Entry<Integer, ModelTensors> entry : sortedMap.entrySet()) {
```
dhrubo-os, line 171: Just curious, why don't we check `HttpStatus` when tensorOutputs.size() > 1 ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,55 @@
...
-        String scheme = host.getSchemeName();
-        if ("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme)) {
-            String[] hostNamePort = host.getHostName().split(":");
-            if (hostNamePort.length > 1 && NumberUtils.isDigits(hostNamePort[1])) {
-                int port = Integer.parseInt(hostNamePort[1]);
-                if (port < 0 || port > 65536) {
-                    log.error("Remote inference port out of range: " + port);
-                    throw new IllegalArgumentException("Port out of range: " + port);
-                }
+    public static void validate(String protocol, String host, int port) throws UnknownHostException {
```
dhrubo-os, line 35: can we give more meaningful name? Validating what?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,55 @@
...
+        if (protocol != null && !"http".equalsIgnoreCase(protocol) && !"https".equals(protocol)) {
```
dhrubo-os, line 36: why for http: `equalsIgnoreCase` and for https it's only `equals`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
```diff
@@ -8,79 +8,55 @@
...
+            log.error("Remote inference protocol is not http or https: " + protocol);
+            throw new IllegalArgumentException("Protocol is not http or https: " + protocol);
+        }
+        if (port == -1) {
```
dhrubo-os, line 40: Can we add a comment regarding this? when port will be -1?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
dhrubo-os: Can we add comment about this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: > `"0".equals(requestBody.optionalContentLength().get().toString())` --> Why do we need to convert to string to check?

The toString() can be removed.

zane-neo: > if `RequestBody.empty()`, what do we get from `requestBody.optionalContentLength().get()`?

it's 0.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: Current implementation sends multiple requests to remote model at the same time, some of them can succeed and some of them can fail. I used a simple solution to use 200 as http status always and user can extract the actual result from the response list, e.g. a sample response could be:
```
{
    "inference_results": [
        {
            "output": [
                {
                    "dataAsMap": {
                        "error_message": {"Runtime exception"}
                    }
                }
            ]
        },
        {
            "output": [
                {
                    "name": "sentence_embedding",
                    "data_type": "FLOAT32",
                    "shape": [
                        768
                    ],
                    "data": [
                        0.08624682,
                        0.28323436,
                        -0.061060455,
                        ...
                    ]
                }
            ]
        }
    ]
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zane-neo: Added comment here: https://github.com/opensearch-project/ml-commons/pull/1958/files/59f16f0ee6313bc6c12a5630c393fac7ea26cd1e#diff-a5fee3d3840ca1ebf7229073c3a53af51fc3d1a99c960820c93d25613f67532eR76

ml-algorithms/src/main/java/org/opensearch/ml/engine/httpclient/MLHttpClientFactory.java
zane-neo: This validates a few things I don't think we need to explicitly specify everything in the method name(which can make the method name very long), when one's reading the code and saw the method and its parameters, one should be able to understand the things we're validating. I don't see any confusion or inconvenience this method name will bring.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
ylwu-amzn: > Current implementation sends multiple requests to remote model at the same time, some of them can succeed and some of them can fail.

Why send multiple requests at the same time ?

ml-algorithms/build.gradle
```diff
@@ -63,12 +63,13 @@ dependencies {
...
     implementation 'software.amazon.awssdk:auth'
     implementation 'software.amazon.awssdk:apache-client'
     implementation 'com.amazonaws:aws-encryption-sdk-java:2.4.1'
     implementation 'com.jayway.jsonpath:json-path:2.9.0'
     implementation group: 'org.json', name: 'json', version: '20231013'
+    implementation group: 'software.amazon.awssdk', name: 'netty-nio-client', version: '2.22.12'
```
samuel-oci, line 72: why dependency on awssdk here instead of the Netty project directly?
https://netty.io/

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: > > Current implementation sends multiple requests to remote model at the same time, some of them can succeed and some of them can fail.
> 
> Why send multiple requests at the same time ?

Previously the requests are send one by one which is natural but not efficient with the sync httpclient. Now we changed this to async httpclient and send multiple requests is a natural approach because we don't need to wait for prior responses.

zane-neo: > > I used a simple solution to use 200 as http status always
> 
> Which part defines this logic ? The old code has such logic
> 
> ```
>             if (statusCode < 200 || statusCode >= 300) {
>                 throw new OpenSearchStatusException(REMOTE_SERVICE_ERROR + modelResponse, RestStatus.fromCode(statusCode));
>             }
> ```
> 
> Do you keep the same logic in your new code?

This logic changed. E.g. if we have a requests with 10 sub requests to remote model and previously the 10 requests are sent sequentially and anyone fails will cause the whole request fail, for example first 9 are succeed and the last one fails, then the whole request fail.
The current logic changed since the httpclient is async and we don't need to blocking waiting for any response(for efficiency), and we send 10 requests to remote model at the same time and then collect all the responses and process them together. So if 9 requests succeed and one fails, we return success response(9 success, 1 fail) to user instead of all fail.

dhrubo-os: What happens if 9 fails and 1 succeeds?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
dhrubo-os: The link seems broken?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -252,4 +267,47 @@ public static SdkHttpFullRequest signRequest(
...
+        int port = url.getPort();
+        MLHttpClientFactory.validate(protocol, host, port);
+        String charset = parameters.getOrDefault("charset", "UTF-8");
+        RequestBody requestBody;
+        if (payload != null) {
+            requestBody = RequestBody.fromString(payload, Charset.forName(charset));
+        } else {
+            requestBody = RequestBody.empty();
+        }
+        if (SdkHttpMethod.POST == method && 0 == requestBody.optionalContentLength().get()) {
```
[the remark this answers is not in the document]
zane-neo: No, even when payload is null, we create requestBody with `RequestBody.empty()` which is not null, and I use this method on purpose to make sure the requestBody never be null.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -216,7 +223,15 @@ public static ModelTensors processOutput(
...
             connector.parseResponse(filteredResponse, modelTensors, scriptReturnModelTensor);
         }
-        return ModelTensors.builder().mlModelTensors(modelTensors).build();
+        return ModelTensors.builder().statusCode(RestStatus.OK.getStatus()).mlModelTensors(modelTensors).build();
+    }
+
+    public static ModelTensors processErrorResponse(String errorResponse) {
+        return ModelTensors
+            .builder()
+            .statusCode(RestStatus.INTERNAL_SERVER_ERROR.getStatus())
```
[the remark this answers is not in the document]
zane-neo: We should use original status code, will add status code as parameter to this method.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: The overall status code is still 200, but the inner status for each request is different, 9 fails have non 200 status_code and 1 succeed has 200.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zane-neo: Yeah, should be caused by new commits, the correct link: https://github.com/opensearch-project/ml-commons/pull/1958/files#diff-a5fee3d3840ca1ebf7229073c3a53af51fc3d1a99c960820c93d25613f67532eR73-R77

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    protected class MLResponseSubscriber implements Subscriber<ByteBuffer> {
+        private Subscription subscription;
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            this.subscription = s;
+            s.request(Long.MAX_VALUE);
```
dhrubo-os, line 184: What does this mean? what happens if we set min value?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        @Override
+        public void onNext(ByteBuffer byteBuffer) {
+            responseBody.append(StandardCharsets.UTF_8.decode(byteBuffer));
```
dhrubo-os, line 189: Should we make this default charsets or this can be user dependent?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: This is a definition which tells the publisher how many at most the subscriber can fetch in one time, you can find more here: https://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/org/reactivestreams/Subscription.html. The value must be a positive number and if we set it to a small number it still works but internally the publisher will publish more times until all data are sent to subscriber. Set it to a bigger value can make this process more efficient.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: We didn't make this user dependent in current code: https://github.com/opensearch-project/ml-commons/blob/a0358e26f2bc5c9be800fe05ce9938b41142edb5/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java#L110, and in network transport case, UTF8 is the standard encoding choice I think we can make it default.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private Map<String, String> parameters;
+
+    private Map<Integer, ModelTensors> tensorOutputs;
+
+    private Connector connector;
+
+    private ScriptService scriptService;
+
+    private final static Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
```
dhrubo-os, line 63: Can't we re-use this [class](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java#L43)?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+    private void processResponse(
+        Integer statusCode,
+        String body,
+        Map<String, String> parameters,
+        Map<Integer, ModelTensors> tensorOutputs
+    ) {
+        ModelTensors tensors;
+        if (Strings.isBlank(body)) {
+            log.error("Remote model response body is empty!");
+            tensors = processErrorResponse(statusCode, "null");
```
dhrubo-os, line 124: "null" --> This will not make any sense to customers. What do you think about `Remote model response is empty!` as default message?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
dhrubo-os: Will neural search face any issue for this change?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,221 @@
...
+        @Override
+        public void onComplete() {
+            processResponse(statusCode, responseBody.toString(), parameters, tensorOutputs);
+            countDownLatch.getCountDownLatch().countDown();
+            response(tensorOutputs);
+        }
+    }
+
+    private void response(Map<Integer, ModelTensors> tensors) {
+        if (countDownLatch.getCountDownLatch().getCount() == 0) {
```
dhrubo-os, line 215: Does this mean we are done with getting all the embeddings? If yes, can we add a comment?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
dhrubo-os: I was mostly curious to know about v1() vs v2()?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: No, we have to make the gson `disableHtmlEscaping` to ensure not breaking our ITs.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: Sounds good.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: sure

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zane-neo: v1 and v2 are in a tuple object, and I have stated the return value of v1 and v2 in the java doc of this method, they're chunk size and step size.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: This indeed changed the behavior in neural search side, I'll change this part to make it bwc with the existing code logic.

dhrubo-os: Have you already made these changes, or do you need to address them in a separate PR? It would be beneficial to provide more details in the description, given the extensive nature of this PR and the potential for breaking changes. To improve understanding and clarity, please consider adding more information to the issue description, including:

The current user experience and any deviations from the previous implementation, if applicable.
The implications for, or interactions with, our upstream processes (such as Airflow or Neural Search).
This additional information will help in assessing the PR more effectively.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -29,52 +32,68 @@
...
+     * @return Tuple of chunk size and step size.
+     */
+    private Tuple<Integer, Integer> calculateChunkSize(TextDocsInputDataSet textDocsInputDataSet) {
+        int textDocsLength = textDocsInputDataSet.getDocs().size();
+        Map<String, String> parameters = getConnector().getParameters();
+        if (parameters != null && parameters.containsKey("input_docs_processed_step_size")) {
+            int stepSize = Integer.parseInt(parameters.get("input_docs_processed_step_size"));
+            // We need to check the parameter on runtime as parameter can be passed into predict request
+            if (stepSize <= 0) {
+                throw new IllegalArgumentException("Invalid parameter: input_docs_processed_step_size. It must be positive integer.");
```
dhrubo-os, line 85: Can we cover this in test?

common/src/main/java/org/opensearch/ml/common/output/model/ModelTensors.java
```diff
@@ -35,6 +35,12 @@ public ModelTensors(List<ModelTensor> mlModelTensors) {
         this.mlModelTensors = mlModelTensors;
     }
 
+    @Builder
```
ylwu-amzn, line 38: Can we have two `@Builder` annotations on different constructors in the same class with Lombok?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: The changes already made to this PR, so this PR has exactly same behavior with previous code implementation. So this won't change the upstream behavior.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -65,99 +54,43 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
...
-                .contentStreamProvider(requestBody.contentStreamProvider());
-            Map<String, String> headers = connector.getDecryptedHeaders();
-            if (headers != null) {
-                for (String key : headers.keySet()) {
-                    builder.putHeader(key, headers.get(key));
-                }
-            }
-            SdkHttpFullRequest request = builder.build();
-            HttpExecuteRequest executeRequest = HttpExecuteRequest
+            SdkHttpFullRequest request = ConnectorUtils.buildSdkRequest(connector, parameters, payload, POST);
```
Zhangxunmt, line 78: This assumes all invoke remote model is a "Post" request by hardcoding. Very soon we will add more action types with different methods. So do you want to extend this to support more HTTP Method through different action types?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -219,6 +225,13 @@ public static ModelTensors processOutput(
         return ModelTensors.builder().mlModelTensors(modelTensors).build();
     }
 
+    public static ModelTensors processErrorResponse(String errorResponse) {
+        return ModelTensors
+            .builder()
+            .mlModelTensors(List.of(ModelTensor.builder().dataAsMap(Map.of("remote_response", errorResponse)).build()))
```
ylwu-amzn, line 231: Is this part backward compatible?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
ylwu-amzn: I don't remember we have `disableHtmlEscaping` before , why it starts failing now?

ml-algorithms/src/main/java/org/opensearch/ml/engine/Predictable.java
```diff
@@ -31,7 +33,13 @@ public interface Predictable {
      * @param mlInput input data
      * @return predicted results
      */
-    MLOutput predict(MLInput mlInput);
+    default MLOutput predict(MLInput mlInput) {
+        throw new IllegalStateException("Method is not implemented");
```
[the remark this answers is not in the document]
zane-neo: Added tests for this. No, this is a runtime error, compilation won't be impacted by this. Below are reasons: 
First of all, this is an interface method but not all subcalsses need to implement this method, e.g. RemoteModel. So we need to make it a default method, and this method can be accessed in all its implemented classes. Also this method needs return value, there are three options:
1. return null: this is not good since this might take long time for someone to identify the root cause.
2. return a manually created object: this is not good since we don't want to manually create an object.
3. throw exception: this is good since the exception and error message is clear and searchable for developer/user thus they can identify the issue quickly.

In fact, there's another option which is keep this method abstract and all subclasses needs to implement this and based on their case, they can throw UnsupportedOperationException which can also avoid accidentally abuse of the method, but this option we need to change quite a lot of files, and since this also works I choose this simpler option.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
zane-neo: This change is based on the existing code: https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java#L116, I think when we needs to support more different methods, we can change this at that time, current implementation can adapt to that change easily.

ml-algorithms/build.gradle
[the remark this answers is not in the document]
ylwu-amzn: The latest version is [2.25.38](https://mvnrepository.com/artifact/software.amazon.awssdk/bom/2.25.38), should we use latest version?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,239 @@
...
+                    t
+                );
+            processResponse(statusCode, responseBody.toString(), parameters, tensorOutputs);
+            countDownLatch.getCountDownLatch().countDown();
+            response(tensorOutputs);
+        }
+
+        @Override
+        public void onComplete() {
+            processResponse(statusCode, responseBody.toString(), parameters, tensorOutputs);
```
ylwu-amzn, line 225: `processResponse` will return error for some failure case, for example if response body is empty. If these error happens, do we still need to run these code ?
```
            countDownLatch.getCountDownLatch().countDown();
            response(tensorOutputs);
```

line 225-227 duplicate with line 218-220 ? Tune these code and merge to process response method?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,239 @@
...
+                    log.error("Failed to process response body: {}", body, e);
+                    actionListener.onFailure(new MLException("Fail to execute predict in aws connector", e));
+                }
+            }
+        }
+    }
+
+    private void reOrderTensorResponses(Map<Integer, ModelTensors> tensorOutputs) {
+        List<ModelTensors> modelTensors = new ArrayList<>();
+        TreeMap<Integer, ModelTensors> sortedMap = new TreeMap<>(tensorOutputs);
```
ylwu-amzn, line 149: Seems not necessary to create a new TreeMap for every predict request. Can we use such way to iterate `sortedMap` with sorted order 
```
for (int i = 1; i<=sortedMap.size(); i++) {
    ModelTensors tensors = sortedMap.get(i);
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,239 @@
...
+        this.statusCode = sdkResponse.statusCode();
+    }
+
+    @Override
+    public void onStream(Publisher<ByteBuffer> stream) {
+        stream.subscribe(new MLResponseSubscriber());
+    }
+
+    @Override
+    public void onError(Throwable error) {
```
ylwu-amzn, line 100: Simplify this method ?
```
@Override 
public void onError(Throwable error) {

  log.error(error.getMessage(), error);
  
  RestStatus status = (statusCode == null) ? 
    RestStatus.INTERNAL_SERVER_ERROR :
    RestStatus.fromCode(statusCode);

  String msg = "Error communicating with remote model: " + error.getMessage();
  
  actionListener.onFailure(
    new OpenSearchStatusException(msg, status)  
  );

}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -252,4 +263,47 @@ public static SdkHttpFullRequest signRequest(
...
+        if (headers != null) {
+            for (String key : headers.keySet()) {
+                builder.putHeader(key, headers.get(key));
+            }
+        }
+        if (builder.matchingHeaders("Content-Type").isEmpty()) {
+            builder.putHeader("Content-Type", "application/json");
+        }
+        if (builder.matchingHeaders("Content-Length").isEmpty()) {
+            builder.putHeader("Content-Length", requestBody.optionalContentLength().get().toString());
```
ylwu-amzn, line 305: Is this mandatory ? We don't have this part before

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
xinyual: Also find this problem in local env. When one response contains error, the listener will onFailure, and in `response` will onResponse Again, which cause unexpected error log. But I think we need to wait for all response and gather them together instead of stopping once we receive a error response.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: Currently no LLM accepts empty request body, adding this is an enhancement to our code which means we should have this in the first place.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: sure

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,236 @@
...
+                actionListener.onResponse(modelTensors);
+            } else {
+                actionListener.onFailure(buildOpenSearchStatusException(singleTensor));
+            }
+        } else {
+            // non batch API. This is to follow the previously code logic. Previously when making multiple requests to remote model,
+            // either one fails, we will return a failure response.
+            OpenSearchStatusException openSearchStatusException = null;
+            for (Map.Entry<Integer, ModelTensors> entry : sortedMap.entrySet()) {
+                if (entry.getValue().getStatusCode() < HttpStatus.SC_OK
```
xinyual, line 161: Do we need to check it again? If one response is error, we will set `exceptionHolder` and won't get into this function. Seems we can remove line 161-165

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -320,17 +320,26 @@ private void runPredict(
...
+                        internalListener.onResponse(new MLTaskResponse(output));
                     }
-
-                    // Once prediction complete, reduce ML_EXECUTING_TASK_COUNT and update task state
-                    handleAsyncMLTaskComplete(mlTask);
-                    MLTaskResponse response = MLTaskResponse.builder().output(output).build();
-                    internalListener.onResponse(response);
                     return;
                 } catch (Exception e) {
+                    log.error("Failed to predict model " + modelId, e);
```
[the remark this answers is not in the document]
ylwu-amzn: please fix this issue before 2.14 code freeze

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
ylwu-amzn: Not quite understand, if need to check request body, you can check body length directly, why have to add a header ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private void processResponse(
+        Integer statusCode,
+        String body,
+        Map<String, String> parameters,
+        Map<Integer, ModelTensors> tensorOutputs
+    ) {
+        if (Strings.isBlank(body)) {
+            log.error("Remote model response body is empty!");
+            if (executionContext.getExceptionHolder().get() == null)
```
ylwu-amzn, line 111: How about add `{}` to make code more readable ? https://stackoverflow.com/questions/20116639/java-one-line-if-statement

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    private void response(Map<Integer, ModelTensors> tensors) {
+        processResponse(statusCode, responseBody.toString(), parameters, tensorOutputs);
+        executionContext.getCountDownLatch().countDown();
+        // when countdown's count equals to 0 means all responses are received.
+        if (executionContext.getCountDownLatch().getCount() == 0) {
+            if (executionContext.getExceptionHolder().get() != null) {
+                actionListener.onFailure(executionContext.getExceptionHolder().get());
```
ylwu-amzn, line 186: Cool, thanks for fixing this bug

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,194 @@
...
+                return;
+            }
+            reOrderTensorResponses(tensors);
+        } else {
+            log.debug("Not all responses received, left response count is: " + executionContext.getCountDownLatch().getCount());
```
ylwu-amzn, line 191: I think we should return a failure for this case 
```
actionListener.onFailure(...)
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -0,0 +1,194 @@
...
+            responseBody.append(StandardCharsets.UTF_8.decode(byteBuffer));
+            subscription.request(Long.MAX_VALUE);
+        }
+
+        @Override
+        public void onError(Throwable t) {
+            log
+                .error(
+                    "Error on receiving response body from remote: {}",
+                    t instanceof NullPointerException ? "NullPointerException" : t.getMessage(),
```
ylwu-amzn, line 168: the code assumes `t.getMessage()` will never be null. This will hold true in most cases, however, there is a chance that in certain exceptions/error conditions `t.getMessage(`) could be null 

```
t instanceof NullPointerException ? "NullPointerException" : (t.getMessage() != null ? t.getMessage() : "Unknown Error")
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
ylwu-amzn: For this edge case, if we don't have all response back, we should still check if any failure received and return that failure; if no failure, return a general error message

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ExecutionContext.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class ExecutionContext {
```
dhrubo-os, line 18: Let's add some comment about what is this class for?

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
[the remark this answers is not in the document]
dhrubo-os: @zane-neo I raised a PR for this: https://github.com/opensearch-project/ml-commons/pull/2310

Yeah it might not cover all the logs, but most of the customer facing exception. Let's follow this pattern and apply wherever necessary.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: httpclient has timeout parameters by default, I don't see the case don't have all response back. Can you give an example?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: Exception can be NullPointerException, and in that case if jdk version is under 9, it can cause NEP here and user won't able to get response.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ExecutionContext.java
zane-neo: sure

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
zane-neo: Aasync httpclient doesn't set the `Content-Length` automatically, we need to do this manually, otherwise we're not able to get correct response from remote model side.

xinyual (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
ylwu-amzn: I don't have example on mind. Generally we use a expected way to process this. I think it's expected to call onFailure for such error, rather than wait for the timeout exception happens.

ylwu-amzn (approved):

xinyual (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Under the condition that "executionContext.getCountDownLatch().getCount() != 0 and executionContext.getExceptionHolder().get() != null", the handler would still wait for the next responses. This means that when we receive an Exception in between the first and last response, we still wait for the last response to return the error message from the earlier results. This is a wast of time and we should just return a failure in the short circuit way and stop processing the remaining responses.

PR #1981, opened 2024-02-01
do not allow non super admin users to undeploy hidden models

rbhavna:
### Description
do not allow non super admin users to undeploy hidden models
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
@@ -117,7 +117,8 @@ public class TransportUndeployModelsAction extends HandledTransportAction<Action
                 }
             }, listener::onFailure));
         } else {
-            // Only allow user undeploy one model if model access control enabled.
+            // Only allow user to undeploy one model if model access control enabled.
+            // With multiple models, it is difficult to check to which models user has access to.
             if (modelAccessControlHelper.isModelAccessControlEnabled()) {
                 throw new IllegalArgumentException("only support undeploy one model");
             } else {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -93,27 +107,48 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUnde
...
-                        listener.onResponse(new MLUndeployModelsResponse(r));
-                    }, e -> { listener.onFailure(e); }));
+                    undeployModels(targetNodeIds, modelIds, listener);
                 } else {
                     listener.onFailure(new IllegalArgumentException("No permission to undeploy model " + modelId));
                 }
             }, listener::onFailure));
-            return;
+        } else {
+            // Only allow user undeploy one model if model access control enabled.
```
ylwu-amzn, line 120: Add more comments explain why?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -93,27 +107,48 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUnde
         String[] modelIds = undeployModelsRequest.getModelIds();
         String[] targetNodeIds = undeployModelsRequest.getNodeIds();
 
-        if (modelAccessControlHelper.isModelAccessControlEnabled()) {
-            // Only allow user undeploy one model if model access control enabled.
-            if (modelIds == null || modelIds.length != 1) {
-                throw new IllegalArgumentException("only support undeploy one model");
-            }
-
+        if (modelIds.length == 1) {
```
mingshl, line 110: when there is only one model, why there is no need to check whether it's hidden model or not?

rbhavna: We do check it. All validations are done [here](https://github.com/opensearch-project/ml-commons/pull/1981/files#diff-8edb43a502aa8b2462127ccfe0a9a89cbd298123526557155a3547d0dea2dd43L124) when there is only one model

ylwu-amzn (approved):

mingshl (approved):

PR #1987, opened 2024-02-02
deploy remote model if predicting an un_deployed remote model

Zhangxunmt:
### Description
Without changing the code structure, this PR resolves the problem that models are not auto-deployed to new hosts when cluster scale up or restart.
The auto-reload runs periodically but the the models are still in un_deploy status in the reload job gaps and the prediction would fail.

Verified it works in my local cluster. UTs can be added later.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -19,8 +19,12 @@ import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.MLTaskState;
 import org.opensearch.ml.common.exception.MLValidationException;
 import org.opensearch.ml.common.transport.MLTaskResponse;
+import org.opensearch.ml.common.transport.deploy.MLDeployModelAction;
+import org.opensearch.ml.common.transport.deploy.MLDeployModelRequest;
+import org.opensearch.ml.common.transport.deploy.MLDeployModelResponse;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskRequest;
 import org.opensearch.ml.helper.ModelAccessControlHelper;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -159,6 +165,44 @@ public void onFailure(Exception e) {
...
+                    .onFailure(
+                        new IllegalArgumentException(
+                            "Model not ready yet. Please run this first: POST /_plugins/_ml/models/" + modelId + "/_deploy"
+                        )
+                    );
+                return;
+            }
+            // The DeployModel is async, set this maximum wait time for deployment to finish
+            long startTime = System.currentTimeMillis();
+            long maxDuration = 100; // 1 second in milliseconds
```
ylwu-amzn, line 193: wrong comment?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -159,6 +165,44 @@ public void onFailure(Exception e) {
...
+            while (workerNodes == null || workerNodes.length == 0) {
```
ylwu-amzn, line 194: It looks like you're missing a sleep statement inside the while loop. Without it, the loop may consume a lot of CPU resources since it will continuously check the condition without any delay. Adding a Thread.sleep() statement inside the loop would be beneficial to reduce CPU usage.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -159,6 +165,44 @@ public void onFailure(Exception e) {
...
 
+    private void deployRemoteModel(
+        String modelId,
+        FunctionName functionName,
+        ActionListener<MLTaskResponse> wrappedListener,
+        MLPredictionTaskRequest mlPredictionTaskRequest
+    ) {
+        String[] workerNodes = mlModelManager.getWorkerNodes(modelId, functionName, true);
+        if (workerNodes != null && workerNodes.length != 0) {
+            return;
```
ylwu-amzn, line 176:
```suggestion
            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
            return;
```

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -159,6 +165,44 @@ public void onFailure(Exception e) {
...
+        }
+
+        MLDeployModelRequest deployModelRequest = MLDeployModelRequest.builder().modelId(modelId).async(false).dispatchTask(true).build();
```
ylwu-amzn, line 179:
```suggestion
        MLDeployModelRequest deployModelRequest = MLDeployModelRequest.builder().modelId(modelId).async(false).dispatchTask(false).build();
```

PR #2073, opened 2024-02-11
changing error message and error code

dhrubo-os:
### Description
[changing error message and error code. Currently error message is default to the 500 error which is not ideal. So we need to set the correct error which we are receiving from the downstream. ]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -19,6 +19,7 @@ import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.exception.MLResourceNotFoundException;
 import org.opensearch.ml.common.transport.MLTaskResponse;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskRequest;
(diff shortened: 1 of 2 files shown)


dbwiddis (approved):

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -139,7 +141,10 @@ public void onResponse(MLModel mlModel) {
                             }
                         }, e -> {
                             log.error("Failed to Validate Access for ModelId " + modelId, e);
-                            wrappedListener.onFailure(e);
+                            wrappedListener
+                                .onFailure(
+                                    new OpenSearchStatusException("Failed to Validate Access for ModelId " + modelId, RestStatus.FORBIDDEN)
```
ylwu-amzn, line 146: The error message means this is some internal error. Before validating access successfully, how can we know this is forbidden ?

dhrubo-os: Good point, but I'm not sure if `INTERNAL_SERVER_ERROR` will be applicable here too. 

Here it's invoking [validateModelGroupAccess](https://github.com/opensearch-project/ml-commons/blob/8ae85a37d608739e57310bd7b9851c41d26bb06a/plugin/src/main/java/org/opensearch/ml/helper/ModelAccessControlHelper.java#L77-L140) method from where I think various kind of exception (Exception, MLResourceNotFoundException, MLValidationException) can be sent.

Do you have any suggestion?

dbwiddis: > various kind of exception (Exception, MLResourceNotFoundException, MLValidationException) can be sent.
> 
> Do you have any suggestion?

You could do a text search of the message for the exception name and/or if there is a "Status:" field in the message.

samuel-oci:
@ylwu-amzn can you please make sure to fill in the fields in the PR description? e.g. __Description__, __Issues resolved__ etc..

samuel-oci (dismissed):

rbhavna (approved):

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
dhrubo-os: @dbwiddis Could you please take a look at the current commit?

dhrubo-os:
> @dhrubo-os can you please make sure to fill in the fields in the PR description? e.g. **Description**, **Issues resolved** etc..

Added description.

ylwu-amzn (approved):

PR #2102, opened 2024-02-13
enable auto redeploy for hidden model

rbhavna:
### Description
enable auto redeploy for hidden model
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
@@ -38,17 +38,21 @@ public class MLDeployModelRequest extends MLTaskRequest {
     private String modelId;
     private String[] modelNodeIds;
     boolean async;
+    // This is to identify if the get request is initiated by user or not. During auto redeploy, we also perform deploy operation. This field is to distinguish between
+    // these two situations.
+    private final boolean isUserInitiatedDeployRequest;
 
     @Builder
-    public MLDeployModelRequest(String modelId, String[] modelNodeIds, boolean async, boolean dispatchTask) {
+    public MLDeployModelRequest(String modelId, String[] modelNodeIds, boolean async, boolean dispatchTask, boolean isUserInitiatedDeployRequest) {
         super(dispatchTask);
         this.modelId = modelId;
         this.modelNodeIds = modelNodeIds;
         this.async = async;
+        this.isUserInitiatedDeployRequest = isUserInitiatedDeployRequest;
     }
 
     public MLDeployModelRequest(String modelId, boolean async) {
-        this(modelId, null, async, true);
+        this(modelId, null, async, true, true);
     }
 
     public MLDeployModelRequest(StreamInput in) throws IOException {
@@ -56,6 +60,7 @@ public class MLDeployModelRequest extends MLTaskRequest {
         this.modelId = in.readString();
         this.modelNodeIds = in.readOptionalStringArray();
         this.async = in.readBoolean();
+        this.isUserInitiatedDeployRequest = in.readOptionalBoolean();
     }
 
     @Override
@@ -74,6 +79,7 @@ public class MLDeployModelRequest extends MLTaskRequest {
         out.writeString(modelId);
         out.writeOptionalStringArray(modelNodeIds);
         out.writeBoolean(async);
+        out.writeOptionalBoolean(isUserInitiatedDeployRequest);
     }
 
     public static MLDeployModelRequest parse(XContentParser parser, String modelId) throws IOException {
@@ -96,7 +102,7 @@ public class MLDeployModelRequest extends MLTaskRequest {
             }
         }
         String[] nodeIds = nodeIdList == null ? null : nodeIdList.toArray(new String[0]);
-        return new MLDeployModelRequest(modelId, nodeIds, false, true);
+        return new MLDeployModelRequest(modelId, nodeIds, false, true, true);
     }
 
     public static MLDeployModelRequest fromActionRequest(ActionRequest actionRequest) {
--- a/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
@@ -38,7 +38,7 @@ public class MLDeployModelRequest extends MLTaskRequest {
     private String modelId;
     private String[] modelNodeIds;
     boolean async;
-    // This is to identify if the get request is initiated by user or not. During auto redeploy also, we perform deploy operation.
+    // This is to identify if the deploy request is initiated by user or not. During auto redeploy also, we perform deploy operation.
     // This field is mainly to distinguish between these two situations.
     private final boolean isUserInitiatedDeployRequest;
 
--- a/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
@@ -38,8 +38,8 @@ public class MLDeployModelRequest extends MLTaskRequest {
     private String modelId;
     private String[] modelNodeIds;
     boolean async;
-    // This is to identify if the get request is initiated by user or not. During auto redeploy, we also perform deploy operation. This field is to distinguish between
-    // these two situations.
+    // This is to identify if the get request is initiated by user or not. During auto redeploy also, we perform deploy operation.
+    // This field is mainly to distinguish between these two situations.
     private final boolean isUserInitiatedDeployRequest;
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
```diff
@@ -38,24 +38,29 @@ public class MLDeployModelRequest extends MLTaskRequest {
...
         super(dispatchTask);
         this.modelId = modelId;
         this.modelNodeIds = modelNodeIds;
         this.async = async;
+        this.isUserInitiatedDeployRequest = isUserInitiatedDeployRequest;
     }
 
     public MLDeployModelRequest(String modelId, boolean async) {
-        this(modelId, null, async, true);
+        this(modelId, null, async, true, true);
```
dhrubo-os, line 55: [nit] may be add a comment why here `isUserInitiatedDeployRequest` will always be true?

common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
```diff
@@ -38,24 +38,29 @@ public class MLDeployModelRequest extends MLTaskRequest {
...
     }
 
     public MLDeployModelRequest(StreamInput in) throws IOException {
         super(in);
         this.modelId = in.readString();
         this.modelNodeIds = in.readOptionalStringArray();
         this.async = in.readBoolean();
+        this.isUserInitiatedDeployRequest = in.readOptionalBoolean();
```
dhrubo-os, line 63: why optionalboolean? what'll happen if the value isn't present?

plugin/src/test/java/org/opensearch/ml/action/deploy/TransportDeployModelActionTests.java
```diff
@@ -286,6 +286,7 @@ public void testDoExecute_no_permission_hidden_model() {
         MLModel mlModel = mock(MLModel.class);
         when(mlModel.getAlgorithm()).thenReturn(FunctionName.ANOMALY_LOCALIZATION);
         when(mlModel.getIsHidden()).thenReturn(true);
+        when(mlDeployModelRequest.isUserInitiatedDeployRequest()).thenReturn(true);
```
arjunkumargiri, line 289: Add unit test to cover `isUserInitiatedDeployRequest` set to false.

arjunkumargiri (commented):
Can we follow similar logic for non-hidden models to bypass access check in case the request is system initiated? : https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R162

rbhavna:
> Can we follow similar logic for non-hidden models to bypass access check in case the request is system initiated? : https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R162

This is already handled in the validateModelGroupAccess method. But it checks for user being null and skips access check which handles the auto redeploy scenario

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -857,7 +857,7 @@ private void updateModelRegisterStateAsDone(
     void deployModelAfterRegistering(MLRegisterModelInput registerModelInput, String modelId) {
         String[] modelNodeIds = registerModelInput.getModelNodeIds();
         log.debug("start deploying model after registering, modelId: {} on nodes: {}", modelId, Arrays.toString(modelNodeIds));
-        MLDeployModelRequest request = new MLDeployModelRequest(modelId, modelNodeIds, false, true);
+        MLDeployModelRequest request = new MLDeployModelRequest(modelId, modelNodeIds, false, true, true);
```
ylwu-amzn, line 860: What if user define a flow template with deploy hidden model action in it ? Can user deploy any hidden model?

dhrubo-os: Isn't hidden model management (register, deploy, undeploy, delete) is out of scope for AI-Flow?

arjunkumargiri:
> > Can we follow similar logic for non-hidden models to bypass access check in case the request is system initiated? : https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R162
> 
> This is already handled in the validateModelGroupAccess method. But it checks for user being null and skips access check which handles the auto redeploy scenario

Null user checks is a workaround, can we explicitly check if deploy request is not user initiated so that same behavior is maintained for system initiated deployments?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
rbhavna: With a flow template, we will still consider it as user initiated right. So we will still check for access. Only if it is system initiated, we will bypass the access checks

ylwu-amzn: > Isn't hidden model management (register, deploy, undeploy, delete) is out of scope for AI-Flow?

Yes for normal process, but that's not a hard control. What if user just copy the hidden model id and define a flow template to deploy it?

arjunkumargiri (approved):

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
zane-neo: Can user see the hidden model id? I thought only superAdmin can see it.

dhrubo-os: >Can user see the hidden model id? I thought only superAdmin can see it.

model id isn't hidden to users. But model content is hidden to users. During prediction, model id is being used. As user can make the prediction request, they can know the model id eventually.

dhrubo-os: >Yes for normal process, but that's not a hard control. User can copy the hidden model id and define a flow template to deploy it.

We opened this [deploy request ](https://github.com/opensearch-project/ml-commons/blob/main/client/src/main/java/org/opensearch/ml/client/MachineLearningNodeClient.java#L260) to clients and then user initiated request flag is being set to true from code. So every request sent from ai-flow or any other clients should be treated as user initiated request too.

common/src/main/java/org/opensearch/ml/common/transport/deploy/MLDeployModelRequest.java
```diff
@@ -38,24 +38,32 @@ public class MLDeployModelRequest extends MLTaskRequest {
     private String modelId;
     private String[] modelNodeIds;
     boolean async;
+    // This is to identify if the get request is initiated by user or not. During auto redeploy also, we perform deploy operation.
```
dhrubo-os, line 41: [nit] this is a deploy request not a get request.

dhrubo-os:
> > > Can we follow similar logic for non-hidden models to bypass access check in case the request is system initiated? : https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R162
> > 
> > 
> > This is already handled in the validateModelGroupAccess method. But it checks for user being null and skips access check which handles the auto redeploy scenario
> 
> Null user checks is a workaround, can we explicitly check if deploy request is not user initiated so that same behavior is maintained for system initiated deployments?

@rbhavna are we planning to address this?

rbhavna:
> > > > Can we follow similar logic for non-hidden models to bypass access check in case the request is system initiated? : https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R162
> > > 
> > > 
> > > This is already handled in the validateModelGroupAccess method. But it checks for user being null and skips access check which handles the auto redeploy scenario
> > 
> > 
> > Null user checks is a workaround, can we explicitly check if deploy request is not user initiated so that same behavior is maintained for system initiated deployments?
> 
> @rbhavna are we planning to address this?

Yes this has been addressed already https://github.com/opensearch-project/ml-commons/pull/2102/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R147

dhrubo-os (approved):

ylwu-amzn (approved):

Zhangxunmt:
Why this PR is not backported to main?

rbhavna:
> Why this PR is not backported to main?

This was merged in this PR @Zhangxunmt https://github.com/opensearch-project/ml-commons/pull/2136

PR #2113, opened 2024-02-15
Add verification to rate limiter number field

b4sjoo:
### Description
This PR will add verification to rate limiter number field to avoid illegal input

### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
+++ b/common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.common.controller;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
+import org.opensearch.OpenSearchParseException;
 import org.opensearch.core.common.io.stream.Writeable;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
```diff
@@ -49,6 +50,14 @@ public static MLRateLimiter parse(XContentParser parser) throws IOException {
             switch (fieldName) {
                 case LIMIT_FIELD:
                     limit = parser.text();
+                    try {
+                        int limitNumber = Integer.parseInt(limit);
```
jngz-es, line 54: What if the number is zero?

b4sjoo: Zero will simply build a rate limiter that refuse all requests, so essentially this is a legal case.

jngz-es: Makes sense. Do we support fraction, like 1.5?

b4sjoo: Nice catch, I just test, fraction works so I change int into double

jngz-es (approved):

ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/controller/MLRateLimiter.java
```diff
@@ -49,6 +50,14 @@ public static MLRateLimiter parse(XContentParser parser) throws IOException {
             switch (fieldName) {
                 case LIMIT_FIELD:
                     limit = parser.text();
+                    try {
+                        double limitNumber = Double.parseDouble(limit);
```
samuel-oci, line 54: shouldn't we add validation for `UNIT_FIELD` as well?

PR #2123, opened 2024-02-16
1799 asymmetric embeddings

### LINKED ISSUE #1799: [FEATURE] Support for asymmetric embedding models
OpenSearch currently only supports symmetric text embedding models, such as `sentence-transformers/all-MiniLM-L12-v2`. These models treat queries and passages equally at inference time. While they can be trained on datasets of queries and passages in such a way that they learn similar representations for the queries and passages (e.g. `sentence-transformers/msmarco-MiniLM-L-12-v3`), the best performing embedding models on the MTEB board (https://huggingface.co/spaces/mteb/leaderboard) are models that offer different inference "APIs" for queries and passages.

To be able to support these models in OpenSearch we need to be able to define different inference mechanisms for the passage embedding and the query embedding.

Prominent asymmetric models use string prefixes to prime the model to embed queries and passages differently. Cf. https://github.com/FlagOpen/FlagEmbedding/tree/master/FlagEmbedding/llm_embedder#using-transformers or https://huggingface.co/intfloat/e5-large-v2#usage. To be able to support this kind of asymmetric model, I propose to introduce model-specific "embedding templates". These should be part of the model metadata and can be used to "format" the input for the model before running inference.

E.g. for the [e5-family](https://github.com/microsoft/unilm/tree/master/e5) of models, the query template could look like this:

| type | template |
|---|---|
| query |  `query: %s` |
| passage | `passage: %s` |

I propose to add optional fields to the model configuration in the `_register` endpoint in ml-commons. E.g.:

```
{
  ...,
  "model_config": {
    ...,
    "query_template" : "query: %s",
    "passage_template" : "passage: %s",
    ...
  },
  ...
}
```

ml-commons already distinguishes datasets by type (cf. [SearchQueryInputDataset](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/dataset/SearchQueryInputDataset.java) and [TextDocsInputDataSet](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/dataset/TextDocsInputDataSet.java)). On inference time, it should be possible to check whether a particular model has templates or not and depending on the dataset type, apply the correct one using regular Java format strings.

OpenSearch neural-search would then need to be extended, to make sure it uses the correct dataset type for queries and passages. Currently it uses the same type, regardless of the use-case: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L248.

The proposed change has a small surface and only extends the API. I couldn't think of any competitive alternative solution.

No.

br3no:
### Description
This PR adds support for asymmetric dense embedding models such as e5 (please refer to #1799 for more information about asymmetric vs. symmetric embedding models).

After this PR is merged users will be able to register e.g. https://huggingface.co/intfloat/multilingual-e5-small as embedding model. This model requires passages and queries to be differently primed with a prefix string. The model needs to be appropriately packaged for OpenSearch, as described in the documentation.

#### Registering

```
{
    "name": "e5-small-onnx-test",
    "version": "1.0.0",
    "description": "test model",
    "model_format": "ONNX",
    "model_group_id": "Vmdrso0BOzEQv2TKCXqI",
    "model_content_hash_value": "f0ee1fce47e73bc8ae1343b74856ae63b06f2b661ebd09d9e588fff5429764ed",
    "model_config": {
        "model_type": "bert",
        "embedding_dimension": 384,
        "framework_type": "sentence_transformers",
        "query_prefix" : "query: ", 
        "passage_prefix" : "passage: ",
        "all_config" : "{  \"_name_or_path\": \"intfloat/multilingual-e5-small\",  \"architectures\": [    \"BertModel\"  ],  \"attention_probs_dropout_prob\": 0.1,  \"classifier_dropout\": null,  \"hidden_act\": \"gelu\",  \"hidden_dropout_prob\": 0.1,  \"hidden_size\": 384,  \"initializer_range\": 0.02,  \"intermediate_size\": 1536,  \"layer_norm_eps\": 1e-12,  \"max_position_embeddings\": 512,  \"model_type\": \"bert\",  \"num_attention_heads\": 12,  \"num_hidden_layers\": 12,  \"pad_token_id\": 0,  \"position_embedding_type\": \"absolute\",  \"tokenizer_class\": \"XLMRobertaTokenizer\",  \"transformers_version\": \"4.30.2\",  \"type_vocab_size\": 2,  \"use_cache\": true,  \"vocab_size\": 250037}"
    },
    "url": "http://localhost:8000/intfloat-multilingual-e5-small-onnx.zip"
}
```

#### Using the model with the predict API
Embedding a passage:
```
POST /_plugins/_ml/_predict/text_embedding/<MODEL_ID>                                                                             
{                    
  "parameters" : {
    "content_type" : "passage"
  },
  "text_docs":[ "Here comes the sun, doo-doo-doo-doo\nHere comes the sun\nAnd I say, it's alright\n\nLittle darlin', it's been a long, cold, lonely winter\nLittle darlin', it feels like years since it's been here"], 
}
```
Embedding a query:
```
POST /_plugins/_ml/_predict/text_embedding/<MODEL_ID>                                                                             
{                    
  "parameters" : {
    "content_type" : "query"
  },
  "text_docs":["Where is the sun?"], 
}
```

#### Further questions

The following answers are valid for the state of the PR at Feb 19th.

> What occurs if both prefixes are not provided during the model registration process? Including the output from such a scenario could be insightful.

The model registration process has been extended to allow for asymmetric models that require string prefixes for the embedding inference (cf. #1799 for two examples of models that work like this). The knowledge whether a model is asymmetric and requires prefixes cannot be expected to be encoded into OpenSearch – it's reasonable to expect the users to know what they are doing in this regard. So registering a model without prefixes is still accepted, since this is how **symmetric** models work. If users want to upload and register an **asymmetric** model, they need to look into the model's documentation to check what prefixes are expected and pass them on to OpenSearch at the moment of model registration.

> What is the outcome if a prefix is omitted during registration, and we attempt to use that missing prefix in a prediction?

There are two possible outcomes: 

1. if both prefixes are missing in the model's registered configuration, the model is considered to be symmetric. If  an attempt is made to compute an embedding with the model with the usage of parameters that indicate the user wants to use an **asymmetric** model, OpenSearch will respond with an `IllegalArgumentException` stating that `When passing AsymmetricTextEmbeddingParameters, the model requires to be registered with at least one of 'query_prefix' or 'passage_prefix'.` (cf. https://github.com/opensearch-project/ml-commons/pull/2123/files#diff-fb3858368c67c4fecc3c252c6fc7a0d81fb468a7839c30300f224834d9d8a1e8R55)
2. if only one prefix is registered, the model is considered to be asymmetric and only require a prefix for either `query` or `passage`. In this case, if there is an attempt to compute an embedding for the "missing" prefix, nothing bad happens – no prefix is added to the text and the embedding is computed as expected.

> If prefixes are included during model registration but are not specified in the prediction request, what is the impact? Performance degradation is expected, but have you observed any specific effects?

As stated above, if a model was registered with prefixes, it is considered to be asymmetric. If a user requests the computation of an embedding without passing the information whether the text is a `QUERY` or a `PASSAGE`, OpenSearch will throw an `IllegalArgumentException` with the message `The embedding model chosen is asymmetric. To use it, you must declare whether the input is of type 'QUERY' or of type 'PASSAGE'.` (cf. https://github.com/opensearch-project/ml-commons/pull/2123/files#diff-fb3858368c67c4fecc3c252c6fc7a0d81fb468a7839c30300f224834d9d8a1e8R72)

I think it is a good design decision to not silently accept an inference request without the input type declaration if the model being used is asymmetric, as this is most likely a mistake. Yes, there would be a performance degradation if this would be accepted.

> What transpires if prefixes are specified in the model registry and used during prediction, but the registered model does not support asymmetric embedding?

There is not much we can do to prevent users from mistakenly registering a **symmetric** model with prefixes and also use the input type specification at inference time. This is because OpenSearch cannot know if a model is **symmetric** or **asymmetric**. This information is not in any way structurally encoded in the model files – it's only part of a paper and model documentation.

Depending on the lengths of prefixes and texts being embedded, there is a performance degradation to be expected.

But I think it's fair to consider this situation very unlikely to happen.

### Issues Resolved
#1799 
 
### Check List
  - [x] All tests pass


plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java
```diff
@@ -136,6 +137,8 @@ public void testRegisterModelRequest() throws Exception {
         assertEquals("test_model", registerModelInput.getModelName());
         assertEquals("1", registerModelInput.getVersion());
         assertEquals("TORCH_SCRIPT", registerModelInput.getModelFormat().toString());
+        assertEquals("query: ", ((TextEmbeddingModelConfig) registerModelInput.getModelConfig()).getQueryPrefix());
+        assertEquals("passage: ", ((TextEmbeddingModelConfig) registerModelInput.getModelConfig()).getPassagePrefix());
```
HenryL27, line 141: can I get a test of null-ness if these are not specified as well?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingDenseModelTest.java
```diff
@@ -239,6 +241,44 @@ private void initModel_predict_HuggingfaceModel(
...
+        textEmbeddingDenseModel.predict(mlInputPassages);
+        TextDocsInputDataSet queries = (TextDocsInputDataSet) mlInputQueries.getInputDataset();
+        TextDocsInputDataSet passages = (TextDocsInputDataSet) mlInputPassages.getInputDataset();
+
+        assertTrue("all docs should start with query prefix", queries.getDocs().stream().allMatch(doc -> doc.startsWith("query >> ")));
+        assertEquals("passage 0 should remain unchanged", passages.getDocs().get(0), "The meaning of life is 42");
+        assertEquals("passage 1 should remain unchanged", passages.getDocs().get(1), "I won this year's us open");
+
+        textEmbeddingDenseModel.close();
+    }
```
HenryL27, line 280: Can we also add tests that show
a. The right thing happens when there's a document prefix
b. The right thing happens when there's prefixes specified but no params provided

HenryL27 (commented):
lookin pretty good! One thing I think you can do for integration tests is to just treat something like `traced_small_model` as an asymmetric embedding model. That should still work right?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -8,21 +8,35 @@
...
         MLInputDataset inputDataSet = mlInput.getInputDataset();
+        MLAlgoParams mlParams = mlInput.getParameters();
+        if (mlParams != null && mlParams instanceof AsymmetricTextEmbeddingParameters) {
+            addPrefixesToData((AsymmetricTextEmbeddingParameters) mlParams, (TextDocsInputDataSet) inputDataSet);
+        } else if (modelConfig != null
+            && (((TextEmbeddingModelConfig) modelConfig).getPassagePrefix() != null
+                || ((TextEmbeddingModelConfig) modelConfig).getQueryPrefix() != null)) {
+            throw new IllegalArgumentException(
+                "The embedding model chosen is asymmetric. To use it, you must declare whether the input is a query or a passage."
+            );
```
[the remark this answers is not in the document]
br3no: I refactored the check into another function, to make it more readable. Let me know what you think.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +50,19 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+    private void addPrefixesToData(AsymmetricTextEmbeddingParameters mlParams, TextDocsInputDataSet inputDataSet) {
+        // Asymmetric embedding models typically work with "mini-prompts" that prime the model to embed a text
+        // as a query or as a passage. Here we apply the prompt as defined in the model configuration. We default
+        // to passage embedding.
+        TextEmbeddingModelConfig modelConfig = (TextEmbeddingModelConfig) this.modelConfig;
+        String prefix = mlParams.getEmbeddingContentType() == EmbeddingContentType.QUERY
+            ? modelConfig.getQueryPrefix()
+            : modelConfig.getPassagePrefix();
+        if (prefix != null) {
+            inputDataSet.getDocs().replaceAll(doc -> prefix + doc);
```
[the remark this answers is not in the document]
br3no: Okay

plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java
br3no: Okay.

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -121,6 +140,8 @@ public TextEmbeddingModelConfig(StreamInput in) throws IOException{
         }
         normalizeResult = in.readBoolean();
         modelMaxLength = in.readOptionalInt();
+        queryPrefix = in.readOptionalString();
```
ylwu-amzn, line 143: We can focus more on the feature itself in this PR. But later when we release this in new version. We should add version check. @b4sjoo Can you help when release new version.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,52 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+                || ((TextEmbeddingModelConfig) modelConfig).getPassagePrefix() == null
+                    && ((TextEmbeddingModelConfig) modelConfig).getQueryPrefix() == null) {
+                throw new IllegalArgumentException(
+                    "When passing AsymmetricTextEmbeddingParameters, the model requires to be "
+                        + "registered with at least one of `query_prefix` or `passage_prefix`."
+                );
+            }
+            // Passed all checks
+            return true;
+        } else if (mlParams != null) {
```
ylwu-amzn, line 62: Will it cause any problem if remove this checking ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,52 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+            // Should never happen.
+            throw new IllegalArgumentException("TEXT_EMBEDDING algorithm only supports AsymmetricTextEmbeddingParameters.");
+        }
+
+        // mlParams is null, but the model is asymmetric.
+        if (modelConfig != null
+            && (((TextEmbeddingModelConfig) modelConfig).getPassagePrefix() != null
+                || ((TextEmbeddingModelConfig) modelConfig).getQueryPrefix() != null)) {
+            throw new IllegalArgumentException(
+                "The embedding model chosen is asymmetric. To use it, you must declare whether the input is a query or a passage."
```
ylwu-amzn, line 73: This error message seems telling user to input a boolean to define it's query or passage, I think you mean user should input either `query_prefix` or  `passage_prefix`?

ylwu-amzn:
@br3no Can you add more details to description part ? Like sample upload model and predict request/response? You can also add some explanation of the parameters. That could help people understand your PR more easily.

br3no:
The build failed on different tests, and is not failing on my machine 😉... Can someone have a look? Was this a problem with the test infrastructure?

br3no:
Seems like my javadocs change fixed the build. ;)

common/src/main/java/org/opensearch/ml/common/dataset/AsymmetricTextEmbeddingParameters.java
```diff
@@ -0,0 +1,114 @@
...
+ * Asymmetric embedding models treat the input text differently, depending on whether it is a
+ * passage or a query. One example asymmetric model, that requires different prefixes is e5
+ * (cf. https://arxiv.org/pdf/2212.03533.pdf).
+ * <p>
+ * Use this parameter only if the model is asymmetric and has been registered with the corresponding
+ * `query_prefix` and `passage_prefix` configuration parameters.
+ */
+@Data
+@MLAlgoParameter(algorithms = { FunctionName.TEXT_EMBEDDING })
+public class AsymmetricTextEmbeddingParameters implements MLAlgoParams {
```
dhrubo-os, line 37: Let's add a corresponding test class for this class.

common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
```diff
@@ -124,6 +129,14 @@ public TextDocsMLInput(XContentParser parser, FunctionName functionName) throws
                 case RESULT_FILTER_FIELD:
                     resultFilter = ModelResultFilter.parse(parser);
                     break;
+                case ML_PARAMETERS_FIELD:
```
dhrubo-os, line 132: Can we add a test for this case?

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -97,12 +110,18 @@ public static TextEmbeddingModelConfig parse(XContentParser parser) throws IOExc
                 case MODEL_MAX_LENGTH_FIELD:
                     modelMaxLength = parser.intValue();
                     break;
+                case QUERY_PREFIX:
+                    queryPrefix = parser.text();
+                    break;
```
dhrubo-os, line 115: add test.

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -97,12 +110,18 @@ public static TextEmbeddingModelConfig parse(XContentParser parser) throws IOExc
...
+                case PASSAGE_PREFIX:
+                    passagePrefix = parser.text();
+                    break;
```
dhrubo-os, line 118: add test.

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
```diff
@@ -135,6 +135,12 @@ public void downloadPrebuiltModelConfig(
                                     case TextEmbeddingModelConfig.MODEL_MAX_LENGTH_FIELD:
                                         configBuilder.modelMaxLength(((Double) configEntry.getValue()).intValue());
                                         break;
+                                    case TextEmbeddingModelConfig.QUERY_PREFIX:
+                                        configBuilder.queryPrefix(configEntry.getValue().toString());
+                                        break;
```
dhrubo-os, line 140: add test.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,52 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+                    "When passing AsymmetricTextEmbeddingParameters, the model requires to be "
+                        + "registered with at least one of `query_prefix` or `passage_prefix`."
+                );
+            }
+            // Passed all checks
+            return true;
+        } else if (mlParams != null) {
+            // mlParams is not null and not an instance of AsymmetricTextEmbeddingParameters.
+            // Should never happen.
+            throw new IllegalArgumentException("TEXT_EMBEDDING algorithm only supports AsymmetricTextEmbeddingParameters.");
```
dhrubo-os, line 65: This doesn't seem right.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,52 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+            );
+        }
+
+        return false;
+    }
+
+    private TextDocsInputDataSet addPrefixesToData(AsymmetricTextEmbeddingParameters mlParams, TextDocsInputDataSet inputDataSet) {
+        // Asymmetric embedding models typically work with "mini-prompts" that prime the model to embed a text
+        // as a query or as a passage. Here we apply the prompt as defined in the model configuration. We default
+        // to passage embedding.
```
dhrubo-os, line 83: may be we should default to query embedding.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,52 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
         return new ModelTensorOutput(tensorOutputs);
     }
 
+    private boolean isAsymmetricModel(MLAlgoParams mlParams) {
+        if (mlParams instanceof AsymmetricTextEmbeddingParameters) {
+            // Check for the necessary prefixes in modelConfig
+            if (modelConfig == null
```
dhrubo-os, line 52: I think we need to refine this condition.

1. What if model config is not null, but in that model config passage prefix and query prefix isn't provided?
2. Shouldn't we expect both prefixes in the model config, so that later there's no confusion?

dhrubo-os:
Thank you for submitting this PR and for providing a detailed description. Could you please enhance your PR description by adding more details in the following areas?

What occurs if both prefixes are not provided during the model registration process? Including the output from such a scenario could be insightful.

What is the outcome if a prefix is omitted during registration, and we attempt to use that missing prefix in a prediction?

If prefixes are included during model registration but are not specified in the prediction request, what is the impact? Performance degradation is expected, but have you observed any specific effects?

What transpires if prefixes are specified in the model registry and used during prediction, but the registered model does not support asymmetric embedding?

These additions will greatly enhance the comprehensiveness of your PR description, providing clearer insights into the behavior and implications of the feature under various conditions.

common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
br3no: This test implements a serialization round-trip. Don't you think this is covered?
https://github.com/opensearch-project/ml-commons/pull/2123/files#diff-a4fe83870d62cd423be730edbf6fa08b8d36ca848dcc32b8c9de6b56f2eac286R72

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
br3no: This code is covered in common/src/test/java/org/opensearch/ml/common/model/TextEmbeddingModelConfigTests.java. I'm not sure what you mean.

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
br3no: This code is covered in common/src/test/java/org/opensearch/ml/common/model/TextEmbeddingModelConfigTests.java.

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
br3no: This one is a chicken-and-egg problem. This is only testable when an asymmetric model is added to the OpenSearch model server, from which one can download prepackaged models. I can remove this code, but then it will not be possible to use an uploaded model, if the community decides to prepackage one. 

Shall I remove this code?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: Cf. https://github.com/opensearch-project/ml-commons/pull/2123#discussion_r1493348525

What's your suggestion? Shall I remove this path?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: I don't mind to change this. But why do you think this is better?

common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
dhrubo-os: I don't think this file is even being used anywhere. So you can skip comment for this file. I'll plan to remove this file later.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: Your point 1. is precisely what is being checked for in this first `if` statement. 

If the model config is not null, but both passage and query prefixes are not provided, we throw an `IllegalArgumentException` saying the model needs to have at least one of them, if you want to use `AsymmetricTextEmbeddingParameters`.

Regarding 2: I don't know any asymmetric model that only requires **one** of both prefixes, but this doesn't mean there isn't one. It would actually be a pretty reasonable design decision to have `passage` be the default. Because of this I thought that it would make sense to be less strict, since we don't really know how these models will develop in the future. 

If you think this is very important, I can change the condition and require both prefixes to be set. If a model only requires one of them, the user would be required to set the empty string as prefix for the other.

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
dhrubo-os: I think you need to update this test: https://github.com/opensearch-project/ml-commons/blob/main/common/src/test/java/org/opensearch/ml/common/model/TextEmbeddingModelConfigTests.java#L84-L88

ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
dhrubo-os: Leave this for now then. Thanks for the explanation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
dhrubo-os: I think we can remove this check.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
dhrubo-os: I think query will happen more often than passage?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
dhrubo-os: In this condition, we are checking both prefixes (&&), but in the message we are telling either one of these two needs to be set. 


We usually use the same model for query & passage, so from this context both should be present. But now in the second thought, there's no cons to keep this flexible, right?

br3no:
@dhrubo-os I have answered the questions you listed in the PR description. Please let me know if there are still any open questions.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: We're checking if both (`&&`) are `null`, and if so, throwing the exception telling the at least one of them needs to be **not** `null`.

br3no: Yes, I think it's better to keep it flexible.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: Okay...

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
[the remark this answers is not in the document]
br3no: Since the other folks already reviewed the changed code, I resolved this conversation. Feel free to reopen.

common/src/main/java/org/opensearch/ml/common/input/nlp/TextDocsMLInput.java
br3no: Oh, okay. I'll revert my changes then.

dhrubo-os:
@br3no thanks a lot for answering all my questions. Overall looks good to me. Approved.

dhrubo-os (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
HenryL27: for an example of a model that uses a query prefix and no passage prefix: https://huggingface.co/BAAI/bge-large-en#frequently-asked-questions

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
```diff
@@ -36,6 +46,48 @@ public ModelTensorOutput predict(String modelId, MLInput mlInput) throws Transla
...
+        }
+
+        // no AsymmetricTextEmbeddingParameters passed, but the model is asymmetric.
+        if (modelConfig != null
+            && (((TextEmbeddingModelConfig) modelConfig).getPassagePrefix() != null
+                || ((TextEmbeddingModelConfig) modelConfig).getQueryPrefix() != null)) {
+            throw new IllegalArgumentException(
+                "The embedding model chosen is asymmetric. To use it, you must declare whether the input is of type `QUERY` or of type `PASSAGE`."
+            );
+        }
```
Zhangxunmt, line 71: Is this really necessary? Why not just return false in this condition. If the param is not asymmetric, just return false. Also this message is confusing. What if I passed a symmetric param with a "QUERY" by mistake, and this message doesn't help me to understand the problem?

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -37,16 +37,25 @@ public class TextEmbeddingModelConfig extends MLModelConfig {
...
+
+    public TextEmbeddingModelConfig(String modelType, Integer embeddingDimension, FrameworkType frameworkType, String allConfig,
+        PoolingMode poolingMode, boolean normalizeResult, Integer modelMaxLength) {
+        this(modelType, embeddingDimension, frameworkType, allConfig, poolingMode, normalizeResult, modelMaxLength, null, null);
+    }
 
     @Builder(toBuilder = true)
     public TextEmbeddingModelConfig(String modelType, Integer embeddingDimension, FrameworkType frameworkType, String allConfig,
-                                    PoolingMode poolingMode, boolean normalizeResult, Integer modelMaxLength) {
+                                    PoolingMode poolingMode, boolean normalizeResult, Integer modelMaxLength, String queryPrefix, String passagePrefix) {
```
Zhangxunmt, line 58: Why not add a check function here to ensure at least one of the new parameters is not Null? It looks like this check can avoid problems later in loading the asymmetric model.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingDenseModelTest.java
```diff
@@ -239,6 +243,257 @@ private void initModel_predict_HuggingfaceModel(
...
+            assertEquals(
+                "When passing AsymmetricTextEmbeddingParameters, the model requires to be registered with at least one of `query_prefix` or `passage_prefix`.",
+                e.getCause().getMessage()
+            );
+            return;
+        } finally {
+            textEmbeddingDenseModel.close();
+        }
+
+        fail("Expected exception not thrown");
```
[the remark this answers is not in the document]
br3no: I used a `try`-`catch` scheme to keep the tests similar to other sad path tests in the test class, e.g. https://github.com/opensearch-project/ml-commons/blob/247a1c77928b2f25ab50db99ff7587b87f0e4433/ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingDenseModelTest.java#L534

I don't mind changing this. But then, shouldn't we refactor all those tests? Should we do this in the context of this PR?

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
br3no: `TextEmbeddingModelConfig` is used for both **asymmetric** and **symmetric** models. Symmetric models don't have prefixes. The alternative would be to create a new `TextEmbeddingModelConfig` class for asymmetric models. I don't think this would be a cleaner solution, tbh.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java
br3no: > Is this really necessary? Why not just return false in this condition. If the param is not asymmetric, just return false.

This exception will be thrown if users want to compute embeddings using an **asymmetric** model **without** stating what kind of data they are trying to embed. 

Running inference on an asymmetric model without the corresponding prefixes can lead to bad things happening. This is what this exception is avoiding. Returning `false` here would lead to OpenSearch silently accepting this and users wondering why the hell the supposedly superb models are performing bad. I don't think this is a good design choice.

> What if I passed a symmetric param with a "QUERY" by mistake, and this message doesn't help me to understand the problem?

I assume you mean what would happen if someone passed an `AsymmetricTextEmbeddingParameters` object, stating they want to embed a text of type `QUERY` while using a **symmetric** model (symmetric models don't take parameters)? This would lead to the exception in line 55. 

One could argue that this is exaggerated caution, that OpenSearch should accept such requests and simply embed the text without adding any prefixes. I thought about that and came to the conclusion, that it is better to fail here. This would avoid the error scenario that someone uploads and registers an **asymmetric** model and **forgets** to configure the prefixes. Trying to compute embeddings in this case would be then silently accepted by OpenSearch leading to degraded model performance.

> Also this message is confusing

Do you think this is still an issue? If so, could you suggest a change?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingDenseModelTest.java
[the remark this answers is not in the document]
HenryL27: yeah I guess if we're gonna do that refactor we should do it everywhere. ok

HenryL27 (approved):

br3no:
Now that the PR is approved, is there anything I must do to get this merged, or will this happen automatically? What is the process?
I can also contribute some documentation, of course.

HenryL27:
@br3no can you help with the backport PR?
Docs would be nice too, but we'll take that in a separate PR

br3no:
Yes, I’ll have a look.

br3no:
I have been able to replicate the issue. I have opened issue #2317 and PR #2318. @dhrubo-os, could you review it?

PR #2130, opened 2024-02-19
remove more sensitive logs in pipeline/responseProcessor logs

Zhangxunmt:
### Description
Pen testers still pointed out "Finding: Search Pipeline / Response Processor Logs Contains Sensitive Data" not resolved. This PR is to delete the leftovers in the sensitive logs required by the pen tests. 

 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
+++ b/search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
@@ -120,7 +120,6 @@ public class GenerativeQAResponseProcessor extends AbstractProcessor implements 
         if (timeout == null || timeout == GenerativeQAParameters.SIZE_NULL_VALUE) {
             timeout = DEFAULT_PROCESSOR_TIME_IN_SECONDS;
         }
-        log.info("Timeout for this request: {} seconds.", timeout);
 
         String llmQuestion = params.getLlmQuestion();
         String llmModel = params.getLlmModel() == null ? this.llmModel : params.getLlmModel();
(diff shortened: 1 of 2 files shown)


search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
```diff
@@ -120,7 +120,6 @@ public SearchResponse processResponse(SearchRequest request, SearchResponse resp
         if (timeout == null || timeout == GenerativeQAParameters.SIZE_NULL_VALUE) {
             timeout = DEFAULT_PROCESSOR_TIME_IN_SECONDS;
         }
-        log.info("Timeout for this request: {} seconds.", timeout);
```
dhrubo-os, line 123: Why this should be treated as a sensitive information?

HenryL27 (approved):
The system and user prompts are returned directly by just the GET search/pipeline endpoint - they're not meant to be secure. Those variables don't include any index or user data - basically just templates.
i.e. I don't think any of the GenerativeQAResponseProcessor log-deletions are necessary?
Approving but would like to keep these logs if possible

search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java
Zhangxunmt: This one is not directly pointed out in the report but the timeout parameter is a configuration provided by customer in their pipeline. So to be on the safe side we should avoid logging it. Just in case this could be a follow-up in the next round.

dhrubo-os: I don't quite agree with the reason. But approving anyway as this is final hour....

dhrubo-os (approved):

PR #2145, opened 2024-02-21
fix npe issue

jngz-es:
### Description
Fix npe issue.

```
POST /_plugins/_ml/agents/ekQGzY0BLfzPq4u31Dfz/_execute 
{
    "parameter": { }
}

Response,
"{\"error\":{\"reason\":\"System Error\",\"details\":\"Cannot invoke \\\"org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet.getParameters()\\\" because \\\"inputDataSet\\\" is null\",\"type\":\"NullPointerException\"},\"status\":500}"
```
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
@@ -98,7 +98,7 @@ public class MLAgentExecutor implements Executable {
         AgentMLInput agentMLInput = (AgentMLInput) input;
         String agentId = agentMLInput.getAgentId();
         RemoteInferenceInputDataSet inputDataSet = (RemoteInferenceInputDataSet) agentMLInput.getInputDataset();
-        if (inputDataSet.getParameters() == null) {
+        if (inputDataSet == null || inputDataSet.getParameters() == null) {
             throw new IllegalArgumentException("wrong input");
         }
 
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -98,7 +98,7 @@ public void execute(Input input, ActionListener<Output> listener) {
         AgentMLInput agentMLInput = (AgentMLInput) input;
         String agentId = agentMLInput.getAgentId();
         RemoteInferenceInputDataSet inputDataSet = (RemoteInferenceInputDataSet) agentMLInput.getInputDataset();
-        if (inputDataSet.getParameters() == null) {
+        if (inputDataSet == null || inputDataSet.getParameters() == null) {
```
ylwu-amzn, line 101: Add some UT?

dhrubo-os: +1

jngz-es: Added.

jngz-es:
> There is a try block that is missing catch in this execute function. Why not include this line inside the try block and catch this error? Plus, this error message is not indicating that parameter is empty. Why not updating the message to show "parameter cannot be null"?

Hi Xun, we don't want to put it in the try/catch block as we want to quickly/directly fail the agent execution and return 4xx to the frontend for something like invalid user request, not system issues. 
The error message is not accurate, I changed it, thanks.

dhrubo-os (approved):

PR #2154, opened 2024-02-22
prevent exposing internal ip

jngz-es:
### Description
Prevent exposing internal ip addresses.
```
{
    "error": {
        ......,
        "reason": "[opensearch-cluster-master-1][192.168.31.112:9300] Node not connected"
    },
    "status": 500
}
```
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
@@ -51,7 +51,7 @@ public class ErrorMessage {
     }
 
     protected String fetchDetails() {
-        String msg;
+        final String msg;
         // Prevent the method from exposing internal information such as internal ip address etc. that is a security concern.
         if (hasInternalInformation(exception)) {
             msg = decorateMessage(exception);
--- a/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.opensearch.core.rest.RestStatus;
+import org.opensearch.transport.ActionTransportException;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 
@@ -50,12 +51,34 @@ public class ErrorMessage {
     }
 
     protected String fetchDetails() {
-        // Some exception prints internal information (full class name) which is security concern
-        return emptyStringIfNull(exception.getLocalizedMessage());
+        String msg;
+        // Prevent the method from exposing internal information such as internal ip address etc. that is a security concern.
+        if (hasInternalInformation(exception)) {
+            msg = decorateMessage(exception);
+        } else {
+            msg = exception.getLocalizedMessage();
+        }
+
+        return emptyStringIfNull(msg);
     }
 
     private String emptyStringIfNull(String str) {
         return str != null ? str : "";
+    }
+
+    private Boolean hasInternalInformation(Throwable t) {
+        if (t instanceof ActionTransportException) {
+            return true;
+        }
+        return false;
+    }
+
+    private String decorateMessage(Throwable t) {
+        if (t instanceof ActionTransportException) {
+            String regexIPPort = "\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}(:\\d{1,5})?";
+            return emptyStringIfNull(t.getLocalizedMessage()).replaceAll(regexIPPort, "x.x.x.x:x");
+        }
+        return "";
     }
 
     @SneakyThrows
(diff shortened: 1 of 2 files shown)


plugin/src/test/java/org/opensearch/ml/utils/error/ErrorMessageTests.java
```diff
@@ -28,6 +33,22 @@ public void fetchDetails() {
         assertEquals(errorMessage.fetchDetails(), "illegal state");
     }
 
+    @Test
+    public void fetchDetailsWithPrivateIP() throws UnknownHostException {
```
dhrubo-os, line 37: may be add another test, showing without ip address it's not redacting anything else?

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
```diff
@@ -50,14 +51,36 @@ protected String fetchReason() {
...
+
+        return emptyStringIfNull(msg);
     }
 
     private String emptyStringIfNull(String str) {
         return str != null ? str : "";
     }
 
+    private Boolean hasInternalInformation(Throwable t) {
+        if (t instanceof ActionTransportException) {
```
dhrubo-os, line 70: Is this exception coming from Core?

jngz-es: Yes, any derived class from it has ip address.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
```diff
@@ -50,14 +51,36 @@ protected String fetchReason() {
...
+            return true;
+        }
+        return false;
+    }
+
+    private String decorateMessage(Throwable t) {
+        if (t instanceof ActionTransportException) {
+            String regexIPPort = "\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}(:\\d{1,5})?";
```
Hailong-am, line 78: should we consider case for IPv6?

samuel-oci (dismissed):
I would start with creating an issue and explaining why 500 error with the details of which node broke is a problem.
If it's an issue related to a hosted service provider not wanting to have their internal IPs exposed I think we should address this a little bit differently.
For example most managed service providers will prefer to prevent 500 errors exposed all together, but we can't really force everyone using OpenSource code maybe even on premise to not be able to see which node IP broke.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
```diff
@@ -50,14 +51,36 @@ protected String fetchReason() {
     }
 
     protected String fetchDetails() {
-        // Some exception prints internal information (full class name) which is security concern
-        return emptyStringIfNull(exception.getLocalizedMessage());
+        String msg;
```
samuel-oci, line 54: final String msg;

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
```diff
@@ -50,14 +51,36 @@ protected String fetchReason() {
...
+        }
+
+        return emptyStringIfNull(msg);
     }
 
     private String emptyStringIfNull(String str) {
         return str != null ? str : "";
     }
 
+    private Boolean hasInternalInformation(Throwable t) {
```
samuel-oci, line 69: see general comment I made earlier. This probably need to go under a flag of some sort in case a hosted service provider want to turn it on. An organization or individual running in a private environment might actually want this check.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
jngz-es: Good point, I would have another pr to address ipv6.

plugin/src/test/java/org/opensearch/ml/utils/error/ErrorMessageTests.java
jngz-es: I think it's already covered by the existing test case fetchDetails() for this case.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
jngz-es: Thanks Samuel, I got your point. Usually we consider internal IPs as sensitive information for a service provider. Regarding an individual running, they can easily get the information in other ways, as it is a cluster issue like some nodes dropping.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
jngz-es: Added final.

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessage.java
ylwu-amzn: hi, @jngz-es  , this PR open for a long time , how about just add ipv6 support in this PR?

jngz-es: Added ipv6 support.

Zhangxunmt (approved):

b4sjoo (approved):

PR #2156, opened 2024-02-22
Change the index update settings to make it only contain dynamic settings

### LINKED ISSUE #2155: [BUG] BWC tests are failing on neural search when load model api is called
**What is the bug?**
The bwc tests on neural search are failing due to below error
```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.number_of_shards]] for open indices [[.plugins-ml-task/B7ApSs9vTtKUurU54IuV6A]]"}],"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.number_of_shards]] for open indices [[.plugins-ml-task/B7ApSs9vTtKUurU54IuV6A]]"},"status":400}
        at __randomizedtesting.SeedInfo.seed([A532B5BBB1EF0C09:B5AEE3F5FC9D5C50]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:376)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:346)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:321)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.makeRequest(BaseNeuralSearchIT.java:735)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.makeRequest(BaseNeuralSearchIT.java:708)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.loadModel(BaseNeuralSearchIT.java:176)
        at app//org.opensearch.neuralsearch.bwc.SemanticSearchIT.testTextEmbeddingProcessor_E2EFlow(SemanticSearchIT.java:42)
```

1. checkout neural search repo and branch 2.x
2. ./gradlew bwcTestSuite  and then consecutively run for other versions by passing parameter for -Dbwc.version (for 2.9, 2.10, 2.11)
3. Test will fail with above error

Bwc tests should pass successfully

 - OS: linux, windows

This fix might be missing in the bwc versions [https://github.com/opensearch-project/ml-commons/pull/2137](https://github.com/opensearch-project/ml-commons/pull/2137)

b4sjoo:
### Description
Previously our index settings contain both static setting and dynamic setting, which will cause bug when doing index update. So we change the index update settings to make it only contain dynamic settings.
 
### Issues Resolved
This PR will resolve #2155
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
@@ -41,6 +41,9 @@ public class MLIndicesHandler {
     ClusterService clusterService;
     Client client;
     private static final Map<String, AtomicBoolean> indexMappingUpdated = new HashMap<>();
+    // This setting is for index update only, so only dynamic settings should be contained!
+    // For index creation setting, use INDEX_SETTINGS directly.
+    public static final Map<String, Object> UPDATED_INDEX_SETTINGS = Map.of("index.auto_expand_replicas", "0-1");
 
     static {
         for (MLIndex mlIndex : MLIndex.values()) {
@@ -123,7 +126,7 @@ public class MLIndicesHandler {
                                     ActionListener.wrap(response -> {
                                         if (response.isAcknowledged()) {
                                             UpdateSettingsRequest updateSettingRequest = new UpdateSettingsRequest();
-                                            updateSettingRequest.indices(indexName).settings(INDEX_SETTINGS);
+                                            updateSettingRequest.indices(indexName).settings(UPDATED_INDEX_SETTINGS);
                                             client
                                                 .admin()
                                                 .indices()


vibrantvarun:
Do we have to not add this bug fix in changelog.md? @b4sjoo

ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
```diff
@@ -41,6 +41,9 @@ public class MLIndicesHandler {
     ClusterService clusterService;
     Client client;
     private static final Map<String, AtomicBoolean> indexMappingUpdated = new HashMap<>();
+    // This setting is for index update only, so only dynamic settings should be contained!
+    // For index creation setting, use INDEX_SETTINGS directly.
+    public static final Map<String, Object> UPDATED_INDEX_SETTINGS = Map.of("index.auto_expand_replicas", "0-1");
```
dhrubo-os, line 46: Should we keep `UPDATED_INDEX_SETTINGS` and `INDEX_SETTINGS` both at the same place? This class name seems suitable for this.

b4sjoo: In fact I totally agree with this idea, but the issue here is for the INDEX_SETTING variable we have to keep it in the memory interface, or it will cause a circular dependency. So there are two ways, the first one is to store both of them in the memory interface, and the second one is to keep as it is. Keeping in the memory seems more tidier, but because MLIndicesHandler is the only place to use this update setting, to keep it as near the line we use it as possible seems also not a bad idea. Which one do you prefer? @dhrubo-os

dhrubo-os: There's a utils folder in common module. What do you think about creating a `indexUtils` class and add these two variables there? All other modules use common module. What do you think?

b4sjoo: Agree, refactored.

common/src/main/java/org/opensearch/ml/common/utils/IndexUtils.java
```diff
@@ -0,0 +1,19 @@
...
+package org.opensearch.ml.common.utils;
+
+import lombok.extern.log4j.Log4j2;
+import java.util.Map;
+
+@Log4j2
+public class IndexUtils {
+
+    // This setting is for index creation.
+    public static final Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", "1", "index.auto_expand_replicas", "0-1");
```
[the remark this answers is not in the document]
b4sjoo: Thanks, added.

dhrubo-os (approved):

rbhavna (approved):

PR #2193, opened 2024-03-12
onnxruntime depends on architecture, not os

HenryL27:
### Description
I couldn't get the correct onnxruntime lib running opensearch 2.12 docker on my m1 mac. My guess is that since it's built for arm linux (being a docker image), checking whether the operating system is macos was insufficient. There's also some potential of the culprit being that the plugin would be _built_ in a linux container - hence never hitting this if statement.

This has worked for me, but I'd want to check that this change still works for intel mac as well.
 
### Issues Resolved
I haven't raised a bug about it.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -51,9 +51,8 @@ dependencies {
     implementation("ai.djl.onnxruntime:onnxruntime-engine:0.21.0") {
         exclude group: "com.microsoft.onnxruntime", module: "onnxruntime"
     }
-    def os = DefaultNativePlatform.currentOperatingSystem
-    //mac doesn't support GPU
-    if (os.macOsX) {
+    //arm doesn't support GPU
+    if (System.getProperty("os.arch") == "aarch64") {
         dependencies {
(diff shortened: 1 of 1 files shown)


ml-algorithms/build.gradle
```diff
@@ -51,9 +51,8 @@ dependencies {
     implementation("ai.djl.onnxruntime:onnxruntime-engine:0.21.0") {
         exclude group: "com.microsoft.onnxruntime", module: "onnxruntime"
     }
-    def os = DefaultNativePlatform.currentOperatingSystem
-    //mac doesn't support GPU
-    if (os.macOsX) {
+    //arm doesn't support GPU
+    if (System.getProperty("os.arch") == "aarch64") {
```
dhrubo-os, line 55: To make sure why can't we keep both then like:

`if os.macOsX || System.getProperty("os.arch") == "aarch64"`

HenryL27: that makes a lot of sense.

HenryL27:
closing because I can't add to this branch

PR #2202, opened 2024-03-13
enhance connector helper notebook to support 2.9

ylwu-amzn:
### Description
enhance connector helper notebook to support 2.9
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


rbhavna (approved):

Zhangxunmt (approved):
This works. But given that it applies to newer versions after 2.9, my concern is that it makes people think this "model_group" is totally a "chicken rib" because it shares the name with model name. And model name is mostly unique so it ends up with unique model name with a dedicated model group.

ylwu-amzn:
>   And model name is mostly unique so it ends up with unique model name with a dedicated model group.

Model group name is unique, but model name not. One model group can have multiple models with same model name.

minalsha:
@ylwu-amzn why do we need to have this when we have Flow framework which exactly does this right now?

ylwu-amzn:
> @ylwu-amzn why do we need to have this when we have Flow framework which exactly does this right now?

But we don't have flow framework on AWS now, right ? This is to help current AWS cx

PR #2204, opened 2024-03-14
implementing hidden agent

dhrubo-os:
### Description
[When an conversational agent is registered with a hidden model, then the agent will be treated as hidden too. We don't want to expose any information for the hidden model.]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -66,7 +66,7 @@ public class CommonValue {
         public static final Integer ML_CONTROLLER_INDEX_SCHEMA_VERSION = 1;
         public static final String ML_MAP_RESPONSE_KEY = "response";
         public static final String ML_AGENT_INDEX = ".plugins-ml-agent";
-        public static final Integer ML_AGENT_INDEX_SCHEMA_VERSION = 1;
+        public static final Integer ML_AGENT_INDEX_SCHEMA_VERSION = 2;
         public static final String ML_MEMORY_META_INDEX = ".plugins-ml-memory-meta";
         public static final Integer ML_MEMORY_META_INDEX_SCHEMA_VERSION = 1;
         public static final String ML_MEMORY_MESSAGE_INDEX = ".plugins-ml-memory-message";
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -416,6 +416,9 @@ public class CommonValue {
                         + MLAgent.MEMORY_FIELD
                         + "\" : {\"type\": \"flat_object\"},\n"
                         + "      \""
+                        + MLAgent.IS_HIDDEN_FIELD
+                        + "\": {\"type\": \"boolean\"},\n"
+                        + "      \""
                         + MLAgent.CREATED_TIME_FIELD
                         + "\": {\"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
                         + "      \""
--- a/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.ml.common;
 
+import java.util.Locale;
+
 public enum MLAgentType {
     FLOW,
     CONVERSATIONAL,
@@ -15,7 +17,7 @@ public enum MLAgentType {
             throw new IllegalArgumentException("Agent type can't be null");
         }
         try {
-            return MLAgentType.valueOf(value.toUpperCase());
+            return MLAgentType.valueOf(value.toUpperCase(Locale.ROOT));
         } catch (Exception e) {
             throw new IllegalArgumentException("Wrong Agent type");
         }
--- a/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
@@ -11,8 +11,11 @@ public enum MLAgentType {
     CONVERSATIONAL_FLOW;
 
     public static MLAgentType from(String value) {
+        if (value == null) {
+            throw new IllegalArgumentException("Agent type can't be null");
+        }
         try {
-            return MLAgentType.valueOf(value);
+            return MLAgentType.valueOf(value.toUpperCase());
         } catch (Exception e) {
             throw new IllegalArgumentException("Wrong Agent type");
         }
--- a/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLAgentType.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common;
+
+public enum MLAgentType {
+    FLOW,
+    CONVERSATIONAL,
+    CONVERSATIONAL_FLOW;
+
+    public static MLAgentType from(String value) {
+        try {
+            return MLAgentType.valueOf(value);
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Wrong Agent type");
+        }
+    }
+}
--- a/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
@@ -83,6 +83,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         this.createdTime = createdTime;
         this.lastUpdateTime = lastUpdateTime;
         this.appType = appType;
+        // is_hidden field isn't going to be set by user. It will be set by the code.
         this.isHidden = isHidden;
         validate();
     }
@@ -91,15 +92,10 @@ public class MLAgent implements ToXContentObject, Writeable {
         if (name == null) {
             throw new IllegalArgumentException("Agent name can't be null");
         }
-        if (type == null) {
-            throw new IllegalArgumentException("Agent type can't be null");
-        } else {
-            validateMLAgentType(type);
-        }
+        validateMLAgentType(type);
         if (type.equalsIgnoreCase(MLAgentType.CONVERSATIONAL.toString()) && llm == null) {
             throw new IllegalArgumentException("We need model information for the conversational agent type");
         }
-
         Set<String> toolNames = new HashSet<>();
         if (tools != null) {
             for (MLToolSpec toolSpec : tools) {
@@ -114,11 +110,15 @@ public class MLAgent implements ToXContentObject, Writeable {
     }
 
     private void validateMLAgentType(String agentType) {
-        try {
-            MLAgentType.valueOf(agentType.toUpperCase()); // Use toUpperCase() to allow case-insensitive matching
-        } catch (IllegalArgumentException e) {
-            // The typeStr does not match any MLAgentType, so throw a new exception with a clearer message.
-            throw new IllegalArgumentException(agentType + " is not a valid Agent Type");
+        if (type == null) {
+            throw new IllegalArgumentException("Agent type can't be null");
+        } else {
+            try {
+                MLAgentType.valueOf(agentType.toUpperCase()); // Use toUpperCase() to allow case-insensitive matching
+            } catch (IllegalArgumentException e) {
+                // The typeStr does not match any MLAgentType, so throw a new exception with a clearer message.
+                throw new IllegalArgumentException(agentType + " is not a valid Agent Type");
+            }
         }
     }
 
@@ -146,6 +146,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         createdTime = input.readOptionalInstant();
         lastUpdateTime = input.readOptionalInstant();
         appType = input.readOptionalString();
+        // is_hidden field isn't going to be set by user. It will be set by the code.
         if (streamInputVersion.onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_HIDDEN_AGENT)) {
             isHidden = input.readOptionalBoolean();
         }
@@ -187,6 +188,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         out.writeOptionalInstant(createdTime);
         out.writeOptionalInstant(lastUpdateTime);
         out.writeOptionalString(appType);
+        // is_hidden field isn't going to be set by user. It will be set by the code.
         if (streamOutputVersion.onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_HIDDEN_AGENT)) {
             out.writeOptionalBoolean(isHidden);
         }
@@ -225,6 +227,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         if (appType != null) {
             builder.field(APP_TYPE_FIELD, appType);
         }
+        // is_hidden field isn't going to be set by user. It will be set by the code.
         if (isHidden != null) {
             builder.field(MLModel.IS_HIDDEN_FIELD, isHidden);
         }
@@ -285,6 +288,7 @@ public class MLAgent implements ToXContentObject, Writeable {
                 case APP_TYPE_FIELD:
                     appType = parser.text();
                     break;
+                // is_hidden field isn't going to be set by user. It will be set by the code.
                 case IS_HIDDEN_FIELD:
                     isHidden = parser.booleanValue();
                     break;
@@ -311,9 +315,5 @@ public class MLAgent implements ToXContentObject, Writeable {
     public static MLAgent fromStream(StreamInput in) throws IOException {
         MLAgent agent = new MLAgent(in);
         return agent;
-    }
-
-    public void removeModelId() {
-        this.llm = null;
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
@@ -245,12 +245,15 @@ public class MLAgent implements ToXContentObject, Writeable {
     }
 
     private static MLAgent parseCommonFields(XContentParser parser, boolean parseHidden) throws IOException {
-        String name = null, type = null, description = null;
+        String name = null;
+        String type = null;
+        String description = null;
         LLMSpec llm = null;
         List<MLToolSpec> tools = null;
         Map<String, String> parameters = null;
         MLMemorySpec memory = null;
-        Instant createdTime = null, lastUpdateTime = null;
+        Instant createdTime = null;
+        Instant lastUpdateTime = null;
         String appType = null;
         boolean isHidden = false;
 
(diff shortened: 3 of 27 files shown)


ylwu-amzn:
Are you building a hidden agent or removing the hidden model from Get/Search agent response?

dhrubo-os:
> Are you building a hidden agent or removing the hidden model from Get/Search agent response?

hidden agent.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -60,10 +66,42 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
...
+        String llmId;
+        if (agent.getType().equalsIgnoreCase(MLAgentType.CONVERSATIONAL.name()) && agent.getLlm() != null) {
+            llmId = agent.getLlm().getModelId();
+        } else {
+            llmId = null;
+        }
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (llmId != null) {
+                String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
+                mlModelManager.getModel(llmId, null, excludes, ActionListener.wrap(mlModel -> {
```
arjunkumargiri, line 78: Few tools also depends upon model ID. Validity of those model ID is not being checked. What is the plan to add tool validations?

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
```diff
@@ -66,6 +75,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLAgen
                     try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
                         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                         MLAgent mlAgent = MLAgent.parse(parser);
+                        if (mlAgent.getIsHidden() && !isSuperAdmin) {
+                            mlAgent.removeModelId();
```
arjunkumargiri, line 79: If tool contains model then it will still be included as part of response. 

It is better to follow existing behavior of Get hidden model and not return complete agent information.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -75,13 +83,23 @@ public MLAgent(String name,
...
     private void validate() {
         if (name == null) {
-            throw new IllegalArgumentException("agent name is null");
+            throw new IllegalArgumentException("Agent name can't be null");
+        }
+        if (type == null) {
+            throw new IllegalArgumentException("Agent type can't be null");
+        } else {
+            validateMLAgentType(type);
+        }
```
arjunkumargiri, line 98: Nit: can move null check as part of validateMLAgentType

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -246,6 +285,9 @@ public static MLAgent parse(XContentParser parser) throws IOException {
                 case APP_TYPE_FIELD:
                     appType = parser.text();
                     break;
+                case IS_HIDDEN_FIELD:
+                    isHidden = parser.booleanValue();
+                    break;
```
arjunkumargiri, line 290: Would this allow users to pass isHidden field through API?

dhrubo-os: Even if they pass the field, that will be replaced with our code logic:

```
if (llmId != null) {
                String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
                mlModelManager.getModel(llmId, null, excludes, ActionListener.wrap(mlModel -> {
                    processModelAndRegisterAgent(agent, mlModel.getIsHidden(), now, listener, context);
                }, e -> {
                    log.error("Failed to retrieve the ML model with ID: " + llmId, e);
                    listener.onFailure(e);
                }));
            } else {
                processModelAndRegisterAgent(agent, false, now, listener, context);
            }
```

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: Are you saying there could be a use case, where we aren't using any hidden model id in agent, but using hidden model id in tools?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -416,6 +416,9 @@ public class CommonValue {
                         + MLAgent.MEMORY_FIELD
                         + "\" : {\"type\": \"flat_object\"},\n"
                         + "      \""
+                        + MLAgent.IS_HIDDEN_FIELD
```
ylwu-amzn, line 419: Bump `ML_AGENT_INDEX_SCHEMA_VERSION` ?

ylwu-amzn:
CI failed 
[ant:jacocoReport] Rule violated for class org.opensearch.ml.action.agents.TransportSearchAgentAction: branches covered ratio is 0.5, but expected minimum is 0.7

dhrubo-os:
> CI failed
> 
> > Task :opensearch-ml-plugin:jacocoTestCoverageVerification FAILED
> > [ant:jacocoReport] Rule violated for class org.opensearch.ml.action.agents.TransportSearchAgentAction: branches covered ratio is 0.5, but expected minimum is 0.7

Yeah, I'm working on it.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
ylwu-amzn: Suggest not parse this field to avoid some confusion. Someone in future may think we can allow user set this field and mess up the code

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
ylwu-amzn: For hidden model , we don't allow common user get it . For hidden agent, we allow common user get it but just filter out LLM model id ? That seems two different implementation of `hidden`. Should we use different term for this different logic for Agent, like using `is_llm_hidden_model` on Agent level, or move `is_hidden`  to Agent.llm level

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -60,10 +66,42 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
...
+    }
+
+    private void processModelAndRegisterAgent(
+        MLAgent agent,
+        Boolean isHiddenModel,
+        Instant now,
+        ActionListener<MLRegisterAgentResponse> listener,
+        ThreadContext.StoredContext context
+    ) {
+        MLAgent mlAgent = agent.toBuilder().createdTime(now).lastUpdateTime(now).isHidden(isHiddenModel).build();
```
ylwu-amzn, line 101: `isHiddenModel` is for LLM , why we set it on Agent level?

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -60,10 +66,42 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
...
         mlIndicesHandler.initMLAgentIndex(ActionListener.wrap(result -> {
             if (result) {
-                try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                try (context) { // Using try-with-resources to ensure 'context' is restored
```
ylwu-amzn, line 104: Can we not pass the context along into nested code, that makes hard to understand what's the impact scope of context

plugin/src/main/java/org/opensearch/ml/action/agents/TransportSearchAgentAction.java
```diff
@@ -42,6 +45,28 @@ private void search(SearchRequest request, ActionListener<SearchResponse> action
...
+            BoolQueryBuilder queryBuilder = QueryBuilders.boolQuery();
+            if (request.source().query() != null) {
+                queryBuilder.must(request.source().query());
+            }
+
+            // Create a BoolQueryBuilder for the should clauses
+            BoolQueryBuilder shouldQuery = QueryBuilders.boolQuery();
+
+            // Add a should clause to include documents where IS_HIDDEN_FIELD is false
+            shouldQuery.should(QueryBuilders.termQuery(MLAgent.IS_HIDDEN_FIELD, false));
```
ylwu-amzn, line 58: So for search agent API, we won't return hidden Agent. For get Agent API, user can get hidden agent with removing hidden model id ?

dhrubo-os: I think according to arjun's suggestion, I'm going to block the hidden agent get request.

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
dhrubo-os: To make it simple may be we can follow the hidden model nature. As arjun said, blocking the get request for hidden agent and keep the `is_hidden` in the agent level?

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: in my understanding if a agent is created with a hidden model, that agent will also be marked as hidden agent.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
arjunkumargiri: yes, so it is better to hide the agent in general and not just hide the conversational model.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
arjunkumargiri: It is better to not allow isHidden parameter through API and set it based on super admin check.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: Yeah, I'm going to hide hidden agent in general.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: I'm passing the context as a method parameter. That was I can re-use the code.

common/src/main/java/org/opensearch/ml/common/MLAgentType.java
```diff
@@ -0,0 +1,23 @@
...
+    FLOW,
+    CONVERSATIONAL,
+    CONVERSATIONAL_FLOW;
+
+    public static MLAgentType from(String value) {
+        if (value == null) {
+            throw new IllegalArgumentException("Agent type can't be null");
+        }
+        try {
+            return MLAgentType.valueOf(value.toUpperCase());
```
ylwu-amzn, line 18:
```suggestion
            return MLAgentType.valueOf(value.toUpperCase(Locale.ROOT));
```

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
```diff
@@ -66,7 +75,17 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLAgen
                     try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, r.getSourceAsBytesRef())) {
                         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                         MLAgent mlAgent = MLAgent.parse(parser);
-                        actionListener.onResponse(MLAgentGetResponse.builder().mlAgent(mlAgent).build());
+                        if (mlAgent.getIsHidden() && !isSuperAdmin) {
+                            actionListener
+                                .onFailure(
+                                    new OpenSearchStatusException(
+                                        "User doesn't have privilege to perform this operation on this model",
```
ylwu-amzn, line 82:
```suggestion
                                        "User doesn't have privilege to perform this operation on this agent",
```

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -60,10 +66,42 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
 
     private void registerAgent(MLAgent agent, ActionListener<MLRegisterAgentResponse> listener) {
         Instant now = Instant.now();
-        MLAgent mlAgent = agent.toBuilder().createdTime(now).lastUpdateTime(now).build();
+        String llmId;
+        if (agent.getType().equalsIgnoreCase(MLAgentType.CONVERSATIONAL.name()) && agent.getLlm() != null) {
```
ylwu-amzn, line 70: If we add a new agent type which also needs LLM, we need to remember to update this part. Not so easy to maintain.

How about change to 
```suggestion
        if (agent.getLlm() != null) {
```

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
ylwu-amzn: Not quite understand, which code you are re-using? 

We should minimize the scope of `stashContext`.  `stashContext` in Line 75 is for get model in line 78, the action listener in line 78 should use runAfter to restore context to minimize the scope of line 75. 

In this action listener of `initMLAgentIndex` , we should stash it again so it's clear the line 75 is for get model, here we stash again for line 109.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: addressed

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
dhrubo-os: done

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
dhrubo-os: done

common/src/main/java/org/opensearch/ml/common/CommonValue.java
dhrubo-os: done

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: done

common/src/main/java/org/opensearch/ml/common/MLAgentType.java
dhrubo-os: done

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
dhrubo-os: done

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
dhrubo-os: done

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
arjunkumargiri: It is better to rely on super admin to identify if an agent is hidden. It is possible that one of the tools make use of hidden model and based on current logic the agent will not be hidden.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -190,67 +228,61 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+
+    private static MLAgent parseCommonFields(XContentParser parser, boolean parseHidden) throws IOException {
+        String name = null, type = null, description = null;
         LLMSpec llm = null;
         List<MLToolSpec> tools = null;
         Map<String, String> parameters = null;
         MLMemorySpec memory = null;
-        Instant createdTime = null;
-        Instant lastUpdateTime = null;
+        Instant createdTime = null, lastUpdateTime = null;
```
ylwu-amzn, line 253: Why change the code style ? Using consistent code style can make it easier to read.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -190,67 +228,61 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
-                case AGENT_TYPE_FIELD:
-                    type = parser.text();
-                    break;
-                case DESCRIPTION_FIELD:
-                    description = parser.text();
-                    break;
-                case LLM_FIELD:
-                    llm = LLMSpec.parse(parser);
-                    break;
+                case AGENT_NAME_FIELD: name = parser.text(); break;
```
ylwu-amzn, line 263: Put multiple lines into one line looks reduce line numbers but also makes it hard to read, generally we use separate line.

ylwu-amzn (dismissed):
LGTM, have some minor comment for code style

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -30,41 +41,79 @@ public class DeleteAgentTransportAction extends HandledTransportAction<ActionReq
...
+        GetRequest getRequest = new GetRequest(ML_AGENT_INDEX).id(agentId);
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
 
-                @Override
-                public void onFailure(Exception e) {
-                    log.error("Failed to delete ML Agent " + agentId, e);
-                    wrappedListener.onFailure(e);
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            client.get(getRequest, ActionListener.wrap(getResponse -> {
+                if (getResponse.isExists()) {
```
arjunkumargiri, line 69: Nit: Nested if conditions makes it harder to read. I prefer following coding pattern to prevent nesting:

```
if (!getResponse.isExists()) {
  actionListener.onFailure(..);
  return;
}
try(...) {
// remaining code
}
```

arjunkumargiri (approved):

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -30,41 +41,79 @@ public class DeleteAgentTransportAction extends HandledTransportAction<ActionReq
...
+                        if (mlAgent.getIsHidden() && !isSuperAdmin) {
+                            actionListener
+                                .onFailure(
+                                    new OpenSearchStatusException(
+                                        "User doesn't have privilege to perform this operation on this hidden agent",
+                                        RestStatus.FORBIDDEN
+                                    )
+                                );
+                        } else {
+                            // If the agent is not hidden or if the user is a super admin, proceed with deletion
```
Zhangxunmt, line 82: looks like the agent is running without any access control. Anyone can delete an agent when he has the agent ID?

Zhangxunmt (dismissed):

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
dhrubo-os: Yeah, we don't have access control in agent yet.

b4sjoo (approved):

Zhangxunmt (approved):

PR #2206, opened 2024-03-14
auto deployment for remote models

Zhangxunmt:
### Description
This PR is the minimum requirement for automatic deployment. When running a prediction for a remote model,
1. If remote model is not deployed, deploy to the cluster. It takes time to deploy to all nodes based on the deployment plan.
2. If deployment has not reached the local node, deploy to the local node without waiting.
3. worker nodes are concurrent hash map in the ModelCache, so it's thread safe. 
4. Once the deployment is done, either from local deploy or cluster deploy, kick off the prediction.
5. TTL for undeployment and UTs will be in a separate PR.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -116,7 +116,7 @@ public class MLModel implements ToXContentObject {
     private Integer totalChunks; // model chunk doc only
     private Integer planningWorkerNodeCount; // plan to deploy model to how many nodes
     private Integer currentWorkerNodeCount; // model is deployed to how many nodes
-
+    @Getter
```
ylwu-amzn, line 119: Do we really need this ? As the class already has `@Getter`, check line 36

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -135,6 +139,8 @@ public void onResponse(MLModel mlModel) {
                                         } else {
                                             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                         }
+                                    } else if (functionName == FunctionName.REMOTE) {
```
ylwu-amzn, line 142: Add a feature flag to control enable/disable this auto deployment ?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +186,67 @@ public void onFailure(Exception e) {
         }
     }
 
+    private void deployAndPredictRemoteModel(
+        String modelId,
+        ActionListener<MLTaskResponse> wrappedListener,
+        MLPredictionTaskRequest mlPredictionTaskRequest
+    ) {
+        String[] workerNodes = mlModelManager.getWorkerNodes(modelId, FunctionName.REMOTE, true);
```
ylwu-amzn, line 194: Should we use `true` or false for `getWorkerNodes` here?

Zhangxunmt: should be true here. Eligible nodes for remote models are ml nodes and data nodes. Also that's how it's used elsewhere in the "predict".

common/src/main/java/org/opensearch/ml/common/MLModel.java
Zhangxunmt: Yeah it's redundant, there must a glitch in IntelliJ when I tried to get this variable so I added this annotation.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
Zhangxunmt: good point, will add one feature flag in the refresh.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +186,67 @@ public void onFailure(Exception e) {
...
+        if (workerNodes != null && workerNodes.length != 0) {
+            // transport layer deployment finished
+            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+            return;
+        }
+        mlModelManager.getModel(modelId, ActionListener.wrap(mlModel -> {
+            // Obtain the deployment plan
+            String[] workerNodeIds = mlModel.getPlanningWorkerNodes();
```
ylwu-amzn, line 202: Is it possible `workerNodeIds` will be null or empty ? If yes, can this PR handle this case?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -135,6 +139,8 @@ public void onResponse(MLModel mlModel) {
                                         } else {
                                             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                         }
+                                    } else if (functionName == FunctionName.REMOTE) {
+                                        deployAndPredictRemoteModel(modelId, wrappedListener, mlPredictionTaskRequest);
```
dhrubo-os, line 143: Don't we need to check if the model is already deployed or not?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +186,67 @@ public void onFailure(Exception e) {
...
+                        new IllegalArgumentException(
+                            "Deployment Failed for " + modelId + ", please check your remote model before predicting."
+                        )
+                    );
+                return;
+            }
+            String[] workerNodes = mlModelManager.getWorkerNodes(modelId, FunctionName.REMOTE, true);
+            if (workerNodes == null || workerNodes.length == 0) {
+                // The deployment is kicked off, but not finished yet in this node
+                mlModelManager.deployRemoteModelToLocal(modelId, ActionListener.wrap(r -> {
```
ylwu-amzn, line 237: So we will deploy again to local here , will it cause any issue/conflict with the first deployment in line 247?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +186,67 @@ public void onFailure(Exception e) {
...
+                    executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+                }, exception -> {
+                    log.error("Failed to deploy model to local node" + modelId, exception);
+                    wrappedListener.onFailure(exception);
+                }));
+            } else {
+                executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
```
ylwu-amzn, line 244: So if local `workerNodes` is not null and empty, we will run predict request. If I remember correctly, the predict request could be routed to other nodes, should we make sure the `mlPredictionTaskRequest` only run on local ?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
Zhangxunmt: Yes it will deploy to all nodes by default. In the cluster level deploy, the transport layer deployment has the logic to deploy to all when worker nodes are null or empty. https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java#L200 In the local level deploy, this is not needed at all.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
Zhangxunmt: It's checked in the function itself as the first thing.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
Zhangxunmt: This is actually the normal predict case when the model is deployed successfully, so it needs to route the request to the work nodes through RoundRobin. In the case of local deployment, I have changed the mlPredictionTaskRequest to not dispatch tasks so it invokes the prediction only locally.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
Zhangxunmt: Good question. I have been thinking this too. My current observation is that usually the local deploy wins and invoke the model, and later the transport layer deploy arrives and overwrite. I haven't implemented bypassing modelCache updates when a value already exists from an earlier deploy (likely not necessary).

For remote models, here are the two variables that to be updated in the deployment. 
1. Predictable predictor
2. Set<String> workerNodes

The workerNodes is initialized as ConcurrentHashMap.newKeySet() so it's thread safe. The predictor is a Predictor value and it's updated through "public synchronized void setPredictor()" and it's also thread safe. 

Basically, during the short window (xx ms) between the cluster deploy and local deploy, we will always "predict" a model based on the latest local deployment value. Once the cluster level deployment is finished, the "predict" will not deploy anymore.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -1049,6 +1050,24 @@ public void deployModel(
         }
     }
 
+    public void deployRemoteModelToLocal(String modelId, MLModel mlModel, ActionListener<String> listener) {
+        String localNodeId = clusterService.localNode().getId();
```
ylwu-amzn, line 1054: `localNodeId` not being used. remove this line ?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +196,69 @@ public void onFailure(Exception e) {
         }
     }
 
+    private void deployAndPredictRemoteModel(
+        String modelId,
+        ActionListener<MLTaskResponse> wrappedListener,
+        MLPredictionTaskRequest mlPredictionTaskRequest
+    ) {
+        String[] workerNodes = mlModelManager.getWorkerNodes(modelId, FunctionName.REMOTE, true);
+        if (!enableAutomaticDeployment || (mlModelManager.isModelDeployed(modelId) && workerNodes != null && workerNodes.length != 0)) {
```
ylwu-amzn, line 205:
```suggestion
        if (!enableAutomaticDeployment || (mlModelManager.isModelDeployed(modelId) && ArrayUtils.isNotEmpty(workerNodes))) {
```

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +196,69 @@ public void onFailure(Exception e) {
...
+                    executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+                }, exception -> {
+                    log.error("Failed to deploy model to local node" + modelId, exception);
+                    wrappedListener.onFailure(exception);
+                }));
+            } else {
+                executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+            }
+        }, wrappedListener::onFailure);
+        client.execute(MLDeployModelAction.INSTANCE, deployModelRequest, deployModelActionListener);
```
ylwu-amzn, line 259: If we have 10k requests comes same time, they may trigger 10k deploy request ? I think we discussed this issue before, but seems this PR doesn't solve it

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -180,6 +196,69 @@ public void onFailure(Exception e) {
...
+                        new IllegalArgumentException(
+                            "Deployment Failed for " + modelId + ", please check your remote model before predicting."
+                        )
+                    );
+                return;
+            }
+            String[] workerNodes = mlModelManager.getWorkerNodes(modelId, FunctionName.REMOTE, true);
+            if (workerNodes == null || workerNodes.length == 0) {
+                // The deployment is kicked off, but not finished yet in this node
+                mlModelManager.deployRemoteModelToLocal(modelId, mlModel, ActionListener.wrap(r -> {
```
ylwu-amzn, line 248: Check if local node is eligible for deploy model first. For example we should not deploy model on master node

dhrubo-os:
```
 - org.opensearch.ml.action.prediction.TransportPredictionTaskActionTests.testPrediction_MLResourceNotFoundException
 - org.opensearch.ml.action.prediction.TransportPredictionTaskActionTests.testPrediction_OpenSearchStatusException
 - org.opensearch.ml.action.prediction.TransportPredictionTaskActionTests.testPrediction_default_exception
 ```

File: plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java (hunk 10 above)
Comment (Zhangxunmt) [reply plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java:259]: This is a good point. First of all we need to agree that it's the best cx experience for 10k requests to finish with correct results rather than rejecting any of them. With that said, we can limit the number of transport deploy call to be maximum 1 through a synchronizedHashSet tracking each modelId. For the local deploys, we cannot limit the number if they needs to happen before the Transport Layer Deployment finishes. This is the price to pay for invoking a cold model without deployment, and it's not that expensive because the local deploy only executes a few doze of code to setup the "Predictable". 

In the example of 10K concurrent requests, the first request will invoke the "Deploy API" in the transport layer. There won't be 9999 local deploy since we don't have that many threads available. Local deploy will exit in the middle if it finds the deployment from transport deployment is finished. So a small portion of 9999 will be local deploy, and then the transport layer deployment finishes, and all rests invokes becomes normal happen case.

File: plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java (hunk 11 above)
Comment (Zhangxunmt) [reply plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java:248]: Good point, we can reuse the current eligible nodes from the ModelManager to check. Will add this in the next refresh.

File: plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java [hunk 12]
```diff
@@ -1049,6 +1050,24 @@ public void deployModel(
...
+        if (modelCacheHelper.isModelDeployed(modelId)) {
+            listener.onResponse("Success");
+            return;
+        }
+        modelCacheHelper
+            .initModelState(modelId, MLModelState.DEPLOYING, FunctionName.REMOTE, new ArrayList<>(), mlModel.isDeployToAllNodes());
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            ActionListener<String> wrappedListener = ActionListener.runBefore(listener, context::restore);
+            modelCacheHelper.setIsModelEnabled(modelId, mlModel.getIsEnabled());
+            deployRemoteOrBuiltInModel(mlModel, 1, wrappedListener);
```
dhrubo-os, line 1064: Can we add a comment why are we setting a default value here?

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -32,10 +32,13 @@
 @Log4j2
 public class MLModelCacheHelper {
     private final Map<String, MLModelCache> modelCaches;
+
+    private final Set<String> localDeployedModels;
```
ylwu-amzn, line 36: How about use HashMap to avoid `synchronized`, https://github.com/ylwu-amzn/ml-commons/blob/2x_autodeploy/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java#L35

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -242,23 +248,45 @@ private void deployAndPredict(
...
+        MLModel mlModel,
+        MLPredictionTaskRequest mlPredictionTaskRequest,
+        ActionListener<MLTaskResponse> wrappedListener
+    ) {
+        String[] workerNodes = mlModelManager.getWorkerNodes(modelId, FunctionName.REMOTE, true);
+        if (workerNodes == null || workerNodes.length == 0) {
+            // The deployment is kicked off, but not finished yet in this node
+            String localNodeId = clusterService.localNode().getId();
+            if (!mlModelManager.isNodeEligible(localNodeId, FunctionName.REMOTE)) {
+                log.error("Failed to deploy model to the ineligible node : " + localNodeId);
```
ylwu-amzn, line 267: How about we run deploy model to local when task dispatched to worker node ? Refer to my poc branch https://github.com/ylwu-amzn/ml-commons/blame/2x_autodeploy/plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java#L147

Zhangxunmt: It seems better. The only cons is that there will be more inter-nodes traffic latency to find the right node to deploy locally through round robin. But overall that latency should be negligible since some nodes that invoke "prediction" shouldn't be running prediction in the first place.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
Zhangxunmt: Will refresh in the next commit.

jngz-es (approved):

b4sjoo (approved):

PR #2211, opened 2024-03-18
Added Conversation API in MLClient

owaiskazi19:
### Description
Added Conversation API in MLClient
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


owaiskazi19:
@dhrubo-os @jngz-es @rbhavna can you take a look? Thanks

memory/src/main/java/org/opensearch/ml/memory/action/conversation/CreateConversationResponse.java
```diff
@@ -67,4 +73,20 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
         return builder;
     }
 
+    public static CreateConversationResponse fromActionResponse(ActionResponse actionResponse) {
```
dhrubo-os, line 76: Let's add a test.

client/build.gradle
```diff
@@ -16,6 +16,7 @@ plugins {
 dependencies {
     implementation project(path: ":${rootProject.name}-spi", configuration: 'shadow')
     implementation project(path: ":${rootProject.name}-common", configuration: 'shadow')
+    implementation project(path: ":${rootProject.name}-memory")
```
dhrubo-os, line 19: Just curious to know if we should mark this as `shadow`?

owaiskazi19: Memory package doesn't have a publishing task that's why @jngz-es suggested to keep it like this.

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -428,4 +429,22 @@ default ActionFuture<ToolMetadata> getTool(String toolName) {
      */
     void getTool(String toolName, ActionListener<ToolMetadata> listener);
 
+    /**
+     * Create conversational memory for conversation
+     * @param name name of the conversation, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/memory-apis/create-memory/
+     * @return the result future
+     */
+    default ActionFuture<CreateConversationResponse> createConversation(String name) {
```
ylwu-amzn, line 437: I don't think we should add this API. We should create a new conversation when conversation id not set.

amitgalitz: Oh okay Yaliang so if we want users to be able to do one click creation for conversation use case with some defaults for demo/poc purpose, we can just skip this step and not give rag processor the conversation ID?

austintlee: I think you do want the ability to give conversations human-readable names.

cc: @HenryL27

amitgalitz: The main reason I am currently adding this is so we can have default use cases ready to go for conversational search https://github.com/opensearch-project/flow-framework/pull/588

We want to be able to do something like setting up the whole e2e with just:
```
http://localhost:9200/_plugins/_flow_framework/workflow?use_case=conversationl_search&provision=true
{
    "create_connector.credential.key" : "123"
}
```
all fields can be overriden but at least for easy poc/dev I thought it would be useful,

 If we have some nice human-readable name in the processor for memory will it even be seen. I was thinking this part creates the memory id to give here, not conversation ID:
 ```
 GET /my_rag_test_data/_search
{
  "query": {
    "match": {
      "text": "What's the population of NYC metro area in 2023"
    }
  },
  "ext": {
    "generative_qa_parameters": {
      "llm_model": "gpt-3.5-turbo",
      "llm_question": "What's the population of NYC metro area in 2023",
      "memory_id": "znCqcI0BfUsSoeNTntd7",
      "context_size": 5,
      "message_size": 5,
      "timeout": 15
    }
  }
}
```

HenryL27: > I was thinking this part creates the memory id to give here, not conversation ID

memory=conversation. message=interaction. Sorry about the confusion here... we probably need to rename the internal classes to match the API names.

HenryL27 (commented):
Do you wanna add any more APIs than just "CreateConversation" / "CreateMemory"? Also @Zhangxunmt which name scheme we going with?

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
Zhangxunmt: Yes, we need to follow the new names. Use Memory to replace Conversation, and Message to replace Interaction.

Zhangxunmt: Yes, we need to follow the new names. Use Memory to replace Conversation, and Message to replace Interaction.

Zhangxunmt:
Yes, we need to follow the new names. Use Memory to replace Conversation, and Message to replace Interaction.

Zhangxunmt (approved):

austintlee:
@owaiskazi19 Do you want to try to get this into 2.18?

dhrubo-os:
@owaiskazi19 any update on this PR?

owaiskazi19:
Sorry for the delay here. Raised https://github.com/opensearch-project/ml-commons/pull/3475 and closing this

PR #2212, opened 2024-03-18
update the response code to 404 when deleting a memory that is not al…

Zhangxunmt:
…lowed to access

### Description
Return 404 not found when deleting a memory that is not allowed to access. Previously it returned only "success": "false"
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/memory/src/main/java/org/opensearch/ml/memory/index/OpenSearchConversationalMemoryHandler.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/index/OpenSearchConversationalMemoryHandler.java
@@ -21,6 +21,7 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.StepListener;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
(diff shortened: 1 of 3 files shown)


memory/src/main/java/org/opensearch/ml/memory/index/OpenSearchConversationalMemoryHandler.java
```diff
@@ -325,7 +327,14 @@ public void deleteConversation(String conversationId, ActionListener<Boolean> li
                 }, listener::onFailure);
 
             } else {
-                listener.onResponse(false);
+                log.error("No access to delete the memory for " + conversationId);
```
dhrubo-os, line 330: From reading the code it seems like when we don't have have access to that conversation we mark this as "Resource not found exception". 

Is that the correct understanding? If yes, that doesn't seem quite right.

Zhangxunmt: The Olly chatbot pen tester thinks that UNAUTHORIZED(401) would suggests the user that the memory id at least exists, which is considered as leaking information. So returning "Not Found"(404) is a better option because it only indicates that you cannot find it.

dhrubo-os: I see, makes sense. Can we add a comment in the code then? So that later we don't find it confusing again?

jngz-es (approved):

b4sjoo (approved):

Issue #2164, opened 2024-02-27
[FEATURE] Add setting to control enable/disable local model


ylwu-amzn:
**Is your feature request related to a problem?**
We have a feature setting to enable/disable remote inference `plugins.ml_commons.remote_inference.enabled` but we don't have dedicated setting for local model

**What solution would you like?**
We can add a new setting for enable/disable local model. 
Once it disabled, user can't upload/deploy/predict model

**What alternatives have you considered?**
N/A

**Do you have any additional context?**

PR #2232, opened 2024-03-19
add local inference enabling/disabling setting

rbhavna:
### Description
add a setting to enable/disable local model inference
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2164
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
@@ -144,7 +144,7 @@ public class TransportDeployModelAction extends HandledTransportAction<ActionReq
                 Boolean isHidden = mlModel.getIsHidden();
                 if (functionName == FunctionName.REMOTE && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
                     throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
-                } else if (FunctionName.isDLModel(functionName) && !mlFeatureEnabledSetting.isLocalModelInferenceEnabled()) {
+                } else if (FunctionName.isDLModel(functionName) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
                     throw new IllegalStateException(LOCAL_MODEL_DISABLED_ERR_MSG);
                 }
                 if (!isUserInitiatedDeployRequest) {
--- a/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
@@ -11,6 +11,7 @@ import static org.opensearch.ml.common.MLTaskState.FAILED;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.DEPLOY_THREAD_POOL;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ALLOW_CUSTOM_DEPLOYMENT_PLAN;
 import static org.opensearch.ml.task.MLTaskManager.TASK_SEMAPHORE_TIMEOUT;
+import static org.opensearch.ml.utils.MLExceptionUtils.LOCAL_MODEL_DISABLED_ERR_MSG;
 import static org.opensearch.ml.utils.MLExceptionUtils.REMOTE_INFERENCE_DISABLED_ERR_MSG;
 
 import java.time.Instant;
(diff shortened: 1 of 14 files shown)


plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -121,6 +122,8 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
     MLPredictionTaskRequest getRequest(String modelId, String algorithm, RestRequest request) throws IOException {
         if (FunctionName.REMOTE.name().equals(algorithm) && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
             throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
+        } else if (FunctionName.isDLModel(FunctionName.from(algorithm)) && !mlFeatureEnabledSetting.isLocalModelInferenceEnabled()) {
```
ylwu-amzn, line 125: Can we also add check in transport action?

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
```diff
@@ -46,4 +53,12 @@ public boolean isAgentFrameworkEnabled() {
         return isAgentFrameworkEnabled;
     }
 
+    /**
+     * Whether the local model feature is enabled. If disabled, APIs in ml-commons will block local model inference.
+     * @return whether the agent framework is enabled.
```
dhrubo-os, line 58: `agent framework`?

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
rbhavna: yes makes sense. Will do

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
rbhavna: Sorry, will change it

dhrubo-os:
The field name`local_model_inference_enabled` doesn't seem quite right to me because it suggests that only prediction is being managed. In reality, we are also controlling model registration and deployment.

Therefore, names like local_model_management_enabled, local_model_lifecycle_enabled, or local_model_services_enabled seem more appropriate for the operations we want to handle.

However, I realize these suggestions might not align perfectly with the existing remote_model_inference_enabled field.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -143,6 +144,8 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLDepl
                 Boolean isHidden = mlModel.getIsHidden();
                 if (functionName == FunctionName.REMOTE && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
                     throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
+                } else if (FunctionName.isDLModel(functionName) && !mlFeatureEnabledSetting.isLocalModelInferenceEnabled()) {
```
dhrubo-os, line 147: Don't we need to add this check for transport registration and transport prediction?

rbhavna: yes working on it now

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -144,7 +144,7 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLDepl
                 Boolean isHidden = mlModel.getIsHidden();
                 if (functionName == FunctionName.REMOTE && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
                     throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
-                } else if (FunctionName.isDLModel(functionName) && !mlFeatureEnabledSetting.isLocalModelInferenceEnabled()) {
+                } else if (FunctionName.isDLModel(functionName) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
                     throw new IllegalStateException(LOCAL_MODEL_DISABLED_ERR_MSG);
```
dhrubo-os, line 148: just wondering if we should throw `IllegalStateException` or Forbidden exception?

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
```diff
@@ -20,6 +20,8 @@ public class MLExceptionUtils {
     public static final String NOT_SERIALIZABLE_EXCEPTION_WRAPPER = "NotSerializableExceptionWrapper: ";
     public static final String REMOTE_INFERENCE_DISABLED_ERR_MSG =
         "Remote Inference is currently disabled. To enable it, update the setting \"plugins.ml_commons.remote_inference_enabled\" to true.";
+    public static final String LOCAL_MODEL_DISABLED_ERR_MSG =
+        "Local Model Inference is currently disabled. To enable it, update the setting \"plugins.ml_commons.local_model.enabled\" to true.";
```
ylwu-amzn, line 24: Should we also change this part? As we are not just disable inference

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
rbhavna: We can maybe use the same one as remote inference and agent framework to stay consistent?

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
rbhavna: fixed it

ylwu-amzn (approved):

dhrubo-os (approved):

PR #2264, opened 2024-03-22
fix remote register model / circuit breaker 500

HenryL27:
### Description
On very rare occasion the memory circuit breaker triggers when registering a remote model. Before this PR, this would throw an exception, and because registerRemoteModel is a actionListener-based method, the error would be dropped and you get a 500 (internal server error) back. This PR moves the circuit breaker check inside the try block so that the exception is caught and 'actionListen'ed.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -346,8 +346,8 @@ public class MLModelManager {
         MLTask mlTask,
         ActionListener<MLRegisterModelResponse> listener
     ) {
-        checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);
             mlStats.getStat(MLNodeLevelStat.ML_REQUEST_COUNT).increment();
             mlStats.createCounterStatIfAbsent(mlTask.getFunctionName(), REGISTER, ML_ACTION_REQUEST_COUNT).increment();
             mlStats.getStat(MLNodeLevelStat.ML_EXECUTING_TASK_COUNT).increment();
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -346,8 +346,8 @@ public void registerMLRemoteModel(
         MLTask mlTask,
         ActionListener<MLRegisterModelResponse> listener
     ) {
-        checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);
```
dhrubo-os, line 350: Can we have a corresponding test to show that exception is 'actionListen'ed now?

HenryL27: sure

plugin/src/test/java/org/opensearch/ml/model/MLModelManagerTests.java
```diff
@@ -449,6 +451,23 @@ public void testRegisterMLRemoteModel() throws PrivilegedActionException {
         verify(mlTaskManager).updateMLTask(anyString(), anyMap(), anyLong(), anyBoolean());
     }
 
+    public void testRegisterMLRemoteModel_WhenMemoryCBOpen_ThenFail() throws PrivilegedActionException {
```
dhrubo-os, line 454: [nit] should it throw `PrivilegedActionException`?

dhrubo-os: does this test fail, if we put `checkAndAddRunningTask(mlTask, maxRegisterTasksPerNode);` outside of the thread context?

HenryL27: guess not :)

dhrubo-os: > guess not :)

Is that a answer for both of these two questions?

dhrubo-os (approved):

plugin/src/test/java/org/opensearch/ml/model/MLModelManagerTests.java
HenryL27: just an answer to the first. my browser seems to be a rather selective comment-loader. The test fails when you put the `checkAnd...` outside of the try block

austintlee (approved):

PR #2266, opened 2024-03-22
add agent framework security it tests

Zhangxunmt:
### Description
Add Security IT tests for Agent Framework. The tests all verified by a local 2.13 cluster.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
@@ -122,7 +122,7 @@ public class RestMLPredictionAction extends BaseRestHandler {
     MLPredictionTaskRequest getRequest(String modelId, String algorithm, RestRequest request) throws IOException {
         if (FunctionName.REMOTE.name().equals(algorithm) && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
             throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
-        } else if (FunctionName.isDLModel(FunctionName.from(algorithm)) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
+        } else if (FunctionName.isDLModel(FunctionName.from(algorithm.toUpperCase())) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
             throw new IllegalStateException(LOCAL_MODEL_DISABLED_ERR_MSG);
         }
         XContentParser parser = request.contentParser();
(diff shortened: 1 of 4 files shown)


plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
```diff
@@ -746,6 +748,47 @@ public String registerModel(String input) throws IOException {
         return parseTaskIdFromResponse(response);
     }
 
+    public void registerMLAgent(RestClient client, String input, Consumer<Map<String, Object>> function) throws IOException {
```
dhrubo-os, line 751: Just curious why don't we add `@Test` with the test methods here?

dhrubo-os:
Thanks for adding integration tests. Is it possible to include testing for hidden agent use case too?

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -122,7 +122,7 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
     MLPredictionTaskRequest getRequest(String modelId, String algorithm, RestRequest request) throws IOException {
         if (FunctionName.REMOTE.name().equals(algorithm) && !mlFeatureEnabledSetting.isRemoteInferenceEnabled()) {
             throw new IllegalStateException(REMOTE_INFERENCE_DISABLED_ERR_MSG);
-        } else if (FunctionName.isDLModel(FunctionName.from(algorithm)) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
+        } else if (FunctionName.isDLModel(FunctionName.from(algorithm.toUpperCase())) && !mlFeatureEnabledSetting.isLocalModelEnabled()) {
```
ylwu-amzn, line 125: Should we fix in `FunctionName.from(..)` method ? So we don't need to add toUpperCase in multiple places where from method being used.

thedhrub0, line 125: I think we should change [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/FunctionName.java#L36C41-L36C46)

`return valueOf(value.toUpperCase(Locale.ROOT));`

ylwu-amzn: Suggest add Locale `toUpperCase(Locale.ROOT)`

Zhangxunmt: Updated

Zhangxunmt: done

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
Zhangxunmt: not needed once you extended OpenSearchRestTestCase.

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
dhrubo-os: Do we need to keep this: `algorithm.toUpperCase()`?

ylwu-amzn (approved):

PR #2269, opened 2024-03-25
guardrails bug fixes and IT for creating guardrails

jngz-es:
### Description
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/model/Guardrails.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/Guardrails.java
@@ -22,26 +22,22 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
 @Getter
 public class Guardrails implements ToXContentObject {
     public static final String TYPE_FIELD = "type";
-    public static final String ENGLISH_DETECTION_ENABLED_FIELD = "english_detection_enabled";
     public static final String INPUT_GUARDRAIL_FIELD = "input_guardrail";
     public static final String OUTPUT_GUARDRAIL_FIELD = "output_guardrail";
 
     private String type;
-    private Boolean engDetectionEnabled;
     private Guardrail inputGuardrail;
     private Guardrail outputGuardrail;
 
     @Builder(toBuilder = true)
-    public Guardrails(String type, Boolean engDetectionEnabled, Guardrail inputGuardrail, Guardrail outputGuardrail) {
+    public Guardrails(String type, Guardrail inputGuardrail, Guardrail outputGuardrail) {
         this.type = type;
-        this.engDetectionEnabled = engDetectionEnabled;
         this.inputGuardrail = inputGuardrail;
         this.outputGuardrail = outputGuardrail;
     }
 
     public Guardrails(StreamInput input) throws IOException {
         type = input.readString();
-        engDetectionEnabled = input.readBoolean();
         if (input.readBoolean()) {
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/opensearch/ml/common/model/MLGuard.java
```diff
@@ -159,4 +159,9 @@ public Boolean validateStopWordsSingleIndex(String input, String indexName, List
         }
         return hitStopWords.get();
     }
+
+    public enum Type {
```
dhrubo-os, line 163: May be we can add a from method like [this](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/FunctionName.java#L34)

Let's rename it to `MLGuardType`?

common/src/main/java/org/opensearch/ml/common/model/Guardrails.java
```diff
@@ -22,26 +22,22 @@
 @Getter
 public class Guardrails implements ToXContentObject {
     public static final String TYPE_FIELD = "type";
-    public static final String ENGLISH_DETECTION_ENABLED_FIELD = "english_detection_enabled";
```
dhrubo-os, line 25: Just curious to know why are we removing this?

mingshl (commented):
why decided to no longer use english detections?

common/src/main/java/org/opensearch/ml/common/model/MLGuard.java
jngz-es: It's an inner class which is used as MLGuard.Type. So I don't think we need to add MLGuard prefix in the name.

common/src/main/java/org/opensearch/ml/common/model/Guardrails.java
jngz-es: Yeah, I got comments before to remove it as we don't implement this feature right now. So to not confuse users, remove it now.

jngz-es:
> why decided to no longer use english detections?

Just not introduce this feature in the first version. We will add it back in next release.

common/src/main/java/org/opensearch/ml/common/model/MLGuard.java
dhrubo-os: can this MLGuard.Type be used in the Guardrails type also? OR these two types are different?

Zhangxunmt (approved):

mingshl (approved):

plugin/src/test/java/org/opensearch/ml/rest/RestMLGuardrailsIT.java
```diff
@@ -0,0 +1,266 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+public class RestMLGuardrailsIT extends MLCommonsRestTestCase {
+
+    final String OPENAI_KEY = System.getenv("OPENAI_KEY");
+    final String COHERE_KEY = System.getenv("COHERE_KEY");
+
+    final String completionModelConnectorEntity = "{\n"
```
Zhangxunmt, line 33: Move this to Utility class of TestHelper under utils folder?

jngz-es: Yeah, will remove in next pr.

PR #2276, opened 2024-03-26
Add a flag to control auto-deploy behavior

b4sjoo:
### Description
This PR introduced a boolean flag to control auto-deploy behavior. This boolean is wrapped with a new flat object field in order to avoid adding to much field to model index directly in the future. To make model index cleaner and maintainable.
 
### Check List
  - [ ] All tests pass


common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
```diff
@@ -0,0 +1,71 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+
+import java.io.IOException;
+
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+
+@Setter
+@Getter
+public class MLDeploySetting implements ToXContentObject, Writeable {
```
Zhangxunmt, line 25: Have you thought about using a Map in the FlatObject of the "deploy_setting" field? Just like here https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java#L93. In this way we can add more settings/flags in the field going forward without any code change.

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -56,7 +56,7 @@ public class CommonValue {
         public static final String ML_MODEL_INDEX = ".plugins-ml-model";
         public static final String ML_TASK_INDEX = ".plugins-ml-task";
         public static final Integer ML_MODEL_GROUP_INDEX_SCHEMA_VERSION = 2;
-        public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 10;
+        public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 11;
```
ylwu-amzn, line 59: Check if we already bumped this version from 2.12 to 2.13 ?

common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
b4sjoo: Yes, but concern here is too much flexibility might cause security issue.

Zhangxunmt (approved):

jngz-es (approved):

PR #2307, opened 2024-04-10
filtering stats api for hidden model

dhrubo-os:
### Description
[filtering stats api for hidden model]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


jngz-es:
Why does this pr merge to 2.x, not main?

dhrubo-os:
> Why does this pr merge to 2.x, not main?

Do you see any concern? I will backport to main eventually? In 2.x it's easy to test with injecting the code in cluster. For super admin, that's how I need to test.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +146,72 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
         // return model level stats
         if (mlStatsInput.includeModelStats()) {
             for (String modelId : mlStats.getAllModels()) {
-                if (mlStatsInput.retrieveStatsForModel(modelId)) {
-                    Map<ActionName, MLActionStats> actionStatsMap = new HashMap<>();
-                    for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
-                        if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
-                            actionStatsMap.put(entry.getKey(), entry.getValue());
+                // Add action listener to retrieve model details
+                validateAccess(modelId, ActionListener.wrap(hasPermissionToShowStat -> {
```
mingshl, line 155: so the change in here to show hidden model id if hasPermissionToShowStat, when I am the super user?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +146,72 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+                    if (hasPermissionToShowStat) {
+                        if (mlStatsInput.retrieveStatsForModel(modelId)) {
+                            Map<ActionName, MLActionStats> actionStatsMap = new HashMap<>();
+                            for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
+                                if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
+                                    actionStatsMap.put(entry.getKey(), entry.getValue());
```
mingshl, line 161: can the users differentiate which model is the hidden model in the map of model_id and their values?

mingshl (dismissed):
Please add `is_hidden` flag in the next PR. Approve.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +146,72 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+                    // Count down the latch even in case of failure to ensure proper synchronization
+                    latch.countDown();
+                }));
+            }
+            // Wait for all asynchronous calls to complete
+            try {
+                latch.await();
+            } catch (InterruptedException e) {
+                // Handle interruption if necessary
+                Thread.currentThread().interrupt();
```
Zhangxunmt, line 182: Why not use await(long timeout, TimeUnit unit) to avoid infinite waiting loop?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +146,72 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
             }
         }
-
         return new MLStatsNodeResponse(clusterService.localNode(), statValues, algorithmStats, modelStats);
     }
+
+    private void validateAccess(String modelId, ActionListener<Boolean> listener) {
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+        String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
```
Zhangxunmt, line 190: Looks like only the isHidden flag is needed. So why not only return this flag in the MLModel by defining an String[] include?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: added default timeout for 2 minutes.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: the underlying method is checking algorithm field. So I don't think adding includes will be a good option here. Do you see any risk of adding `excludes`? I don't think it will add any extra latency in the query.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +147,78 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+                Thread.currentThread().interrupt();
             }
         }
-
         return new MLStatsNodeResponse(clusterService.localNode(), statValues, algorithmStats, modelStats);
     }
+
+    private void validateAccess(String modelId, ActionListener<Boolean> listener) {
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+        String[] includes = new String[] { MLModel.IS_HIDDEN_FIELD };
```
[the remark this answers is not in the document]
dhrubo-os: reverted.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
Zhangxunmt: In the case of timeout, the results will be the partial stats for all models? e.g. some of model stats are not included in the results?

dhrubo-os: Yeah that's correct. But I don't think it will take 2 minutes to get all the models. What do you think?

Zhangxunmt: Is it trivial to add a flag in the response to show if the stats are complete in summarizing the MLStatsNodesResponse from all nodes? Return partial is fine, but it would be better to indicate that in the response.

dhrubo-os: It don't think it's trivial and also I'm not sure if it's necessary though. If we think 2 minutes isn't enough threshold time, we can increase it to 5 minutes. But in my experience, It didn't take too long to get 1000 models in stat api response.

Zhangxunmt: Your assumption may be valid as of now, but not necessarily true if more stats are added later on in the code. Also if this stats API is running on a schedule that is less than 2 minutes, is it going to trigger problems and delay the response?

dhrubo-os: >Also if this stats API is running on a schedule that is less than 2 minutes

Can you explain a bit more here?

>but not necessarily true if more stats are added later on in the code

how more stats are we expecting?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: Change here is: If it's superadmin user, only that time we will show the stats for hidden model. Other users will not be able to see the hidden model stats.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: user won't differentiate.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -26,15 +33,25 @@
...
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
 public class MLStatsNodesTransportAction extends
     TransportNodesAction<MLStatsNodesRequest, MLStatsNodesResponse, MLStatsNodeRequest, MLStatsNodeResponse> {
     private MLStats mlStats;
     private final JvmService jvmService;
 
+    Client client;
```
ylwu-amzn, line 51: minor: add `private final`  for this line and line 53?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +147,78 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+            mlModelManager.getModel(modelId, null, excludes, ActionListener.runBefore(ActionListener.wrap(mlModel -> {
+                Boolean isHidden = mlModel.getIsHidden();
+                if (isHidden != null && isHidden) {
+                    if (isSuperAdmin) {
+                        listener.onResponse(true);
+                    } else {
+                        listener.onResponse(false);
+                    }
+                } else {
+                    listener.onResponse(true);
```
ylwu-amzn, line 208: From this, we don't check if user have access to this model ?

ylwu-amzn: Seems all users have access unless the model is hidden and they are not a super admin. The method name `validateAccess` seems misleading, how about change to `validateModelStatsViewAccess`

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +147,78 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
         return new MLStatsNodeResponse(clusterService.localNode(), statValues, algorithmStats, modelStats);
     }
+
+    private void validateAccess(String modelId, ActionListener<Boolean> listener) {
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+        String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            mlModelManager.getModel(modelId, null, excludes, ActionListener.runBefore(ActionListener.wrap(mlModel -> {
+                Boolean isHidden = mlModel.getIsHidden();
+                if (isHidden != null && isHidden) {
```
ylwu-amzn, line 201: Seems line 201 to 208 can be simplified as 
```
listener.onResponse(isSuperAdmin || (isHidden != null && !isHidden));
```

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +147,78 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+                            for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
+                                if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
+                                    actionStatsMap.put(entry.getKey(), entry.getValue());
+                                }
+                            }
+                            modelStats.put(modelId, new MLModelStats(actionStatsMap));
                         }
                     }
-                    modelStats.put(modelId, new MLModelStats(actionStatsMap));
+                    // Count down the latch after each asynchronous call completes
```
ylwu-amzn, line 168: Cound down latch for which async call ?

dhrubo-os: for all the get model calls

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +147,78 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
     }
+
+    private void validateAccess(String modelId, ActionListener<Boolean> listener) {
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+        String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            mlModelManager.getModel(modelId, null, excludes, ActionListener.runBefore(ActionListener.wrap(mlModel -> {
+                Boolean isHidden = mlModel.getIsHidden();
+                if (isHidden != null && isHidden) {
+                    if (isSuperAdmin) {
```
ylwu-amzn, line 202: If user is super admin, we don't need to get model from index and check, right ? Just always return true.

If not super admin, we can just filter out the hidden model

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
Zhangxunmt: For example, at some point we will need to query the stats periodically and send the data as metrics to CW for dashboard monitoring, etc. Or at some time, the stats API is called at a large TPS. The waiting in the thread will be a potential problem? We cannot just limit this API to be a very simple use case right?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -126,22 +155,80 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+
+            // Execute the search request asynchronously
+            client.search(searchRequest, al);
+        } catch (Exception e) {
+            // Notify the listener of any unexpected errors
+            listener.onFailure(e);
+        }
+    }
+
+    private SearchRequest buildSearchRequest() {
```
lezzago, line 211: Can we improve the function name so its clear this is building a search request for finding hidden model, so it can be like `buildHiddenModelSearchRequest`

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -127,21 +156,89 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
         if (mlStatsInput.includeModelStats()) {
-            for (String modelId : mlStats.getAllModels()) {
-                if (mlStatsInput.retrieveStatsForModel(modelId)) {
-                    Map<ActionName, MLActionStats> actionStatsMap = new HashMap<>();
-                    for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
-                        if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
-                            actionStatsMap.put(entry.getKey(), entry.getValue());
+            CountDownLatch latch = new CountDownLatch(1);
+            boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+            searchHiddenModels(ActionListener.wrap(hiddenModels -> {
```
ylwu-amzn, line 162: If isSuperAdmin is true, de we need to search hidden models?

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -127,21 +156,89 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
-                    modelStats.put(modelId, new MLModelStats(actionStatsMap));
                 }
+                latch.countDown();
+            }, e -> {
+                log.error("Search Hidden model wasn't successful");
+                latch.countDown();
+            }));
+            // Wait for the asynchronous call to complete
+            try {
+                latch.await(120, TimeUnit.SECONDS);
```
ylwu-amzn, line 183: Do we really need to wait for so long ? Suggest change to a reasonable timeout so the thread not being blocked for a long time and reduce other part.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -127,21 +156,89 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+                                if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
+                                    actionStatsMap.put(entry.getKey(), entry.getValue());
+                                }
+                            }
+                            modelStats.put(modelId, new MLModelStats(actionStatsMap));
                         }
                     }
-                    modelStats.put(modelId, new MLModelStats(actionStatsMap));
                 }
+                latch.countDown();
```
ylwu-amzn, line 176: Move this line to runAfter in line 198 `ActionListener<Set<String>> internalListener = ActionListener.runAfter(listener, () -> threadContext.restore());` ? So we can remove the duplicate line 179

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: I was thinking to add a flag of the hidden model based on that search so that if superadmin pulls the metrics they know which one is the hidden model metrics, but I was thinking to add that flag in another PR, not in this PR.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -127,21 +156,89 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
+        SearchRequest searchRequest = buildHiddenModelSearchRequest();
+        // Use a try-with-resources block to ensure resources are properly released
+        try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
+            // Wrap the listener to restore thread context before calling it
+            ActionListener<Set<String>> internalListener = ActionListener.runAfter(listener, () -> threadContext.restore());
+            // Wrap the search response handler to handle success and failure cases
+            // Notify the listener of any search failures
+            ActionListener<SearchResponse> al = ActionListener.wrap(response -> {
+                // Initialize the result set
+                Set<String> result = new HashSet<>(response.getHits().getHits().length); // Set initial capacity to the number of hits
```
ylwu-amzn, line 203: Should we handle edge cases like model index not created yet(for a brand new cluster), model index created but user deleted all models, etc?

dhrubo-os: In my understanding, that should fail in line no 217 and eventually go to the following section:

```
e -> {
                log.error("Search Hidden model wasn't successful");
                latch.countDown();
            }
```

So we aren't showing any model stats finally.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
ylwu-amzn: Got it, I think you mean add is_hidden flag in stats API response, so user can know the hidden model usage

ylwu-amzn (approved):
LGTM. Thanks for the change

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
mingshl: questions here, so if the users delete all the models, will the model index still exist? 

I am hoping if there is no models in the cluster, it will delete the model index.

dhrubo-os: The index itself will not be deleted automatically.

mingshl (approved):

PR #2310, opened 2024-04-11
filter model id for hidden model cases

dhrubo-os:
### Description
[For hidden model we don't want to show model id in the exception message to the customer. This PR is removing model id for hidden model wherever possible.]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
@@ -191,4 +191,19 @@ public class StringUtils {
     public static boolean containsEscapeMethod(String input) {
         return patternExist(input, "String\\s+escape\\s*\\(\\s*(def|String)\\s+.*?\\)\\s*\\{?");
     }
+
+    /**
+     * This method will define if we should print out model id with the error message or not.
+     * @param errorMessage
+     * @param modelId
+     * @param isHidden
+     * @return
+     */
+    public static String getErrorMessage(String errorMessage, String modelId, boolean isHidden) {
+        if (isHidden) {
+            return errorMessage;
+        } else {
+            return errorMessage + " Model ID: " + modelId;
+        }
+    }
 }
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
@@ -193,7 +193,7 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
                     }
 
                     if (indexResponse.getResult() == DocWriteResponse.Result.CREATED) {
-                        mlModelManager.updateModel(modelId, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
+                        mlModelManager.updateModel(modelId, isHidden, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
                     }
                     if (!ArrayUtils.isEmpty(mlModelCacheHelper.getWorkerNodes(modelId))) {
                         if (isHidden) {
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.action.controller;
 import static org.opensearch.ml.common.CommonValue.ML_CONTROLLER_INDEX;
 import static org.opensearch.ml.common.FunctionName.REMOTE;
 import static org.opensearch.ml.common.FunctionName.TEXT_EMBEDDING;
+import static org.opensearch.ml.common.utils.StringUtils.getErrorMessage;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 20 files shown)


plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -176,17 +176,21 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLDepl
                                 deployModel(deployModelRequest, mlModel, modelId, wrappedListener, listener);
                             }
                         }, e -> {
-                            log.error("Failed to Validate Access for ModelId " + modelId, e);
+                            if (isHidden) {
+                                log.error("Failed to Validate Access for the given model", e);
+                            } else {
+                                log.error("Failed to Validate Access for ModelId " + modelId, e);
+                            }
                             wrappedListener.onFailure(e);
```
jngz-es, line 184: For deploy model api, you removed the model id by "the given model". What about update model api, should we also remove the model id?
`if (isSuperAdmin) {
                                updateRemoteOrTextEmbeddingModel(modelId, updateModelInput, mlModel, user, wrappedListener);
                            } else {
                                wrappedListener
                                    .onFailure(
                                        new OpenSearchStatusException(
                                            "User doesn't have privilege to perform this operation on this model, model ID " + modelId,
                                            RestStatus.FORBIDDEN
                                        )
                                    );
                            }`

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -244,21 +244,21 @@ private void deleteModelChunksAndController(
             bothDeleted.compareAndSet(true, b);
             if (countDownLatch.getCount() == 0) {
                 if (bothDeleted.get()) {
-                    log.debug("model chunks and model controller for model {} deleted successfully", modelId);
```
jngz-es, line 247: Why don't we have model id there for non super admin users? The model id information can help debug and log analysis.

dhrubo-os: whenever I get the model information, i re-used that info to detect if that model is hidden or not. But for delete model we don't perform any get request so I didn't introduce any extra api call.

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
```diff
@@ -100,37 +100,41 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLCrea
...
+                Boolean isHidden = mlModel.getIsHidden();
                 if (functionName == TEXT_EMBEDDING || functionName == REMOTE) {
                     modelAccessControlHelper
                         .validateModelGroupAccess(user, mlModel.getModelGroupId(), client, ActionListener.wrap(hasPermission -> {
                             if (hasPermission) {
                                 if (mlModel.getModelState() != MLModelState.DEPLOYING) {
                                     indexAndCreateController(mlModel, controller, wrappedListener);
+                                } else {
+                                    final String errorMessage =
+                                        "Creating a model controller during its corresponding model in DEPLOYING state is not allowed, please either create the model controller after it is deployed or before deploying it.";
```
ylwu-amzn, line 112: @b4sjoo  Can you help review ? Seems this error message is enough. Do we really need to return the model id in error message ?

ylwu-amzn: This PR adds a lot of if/else checking for just handling error message of hidden model. Looks hard to maintain in future. How about we just use a common error message for both hidden and normal models

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
dhrubo-os: If you are referring to this errorMsg, then cx should know already for which model he/she is creating the controller.

But in general for more debugging purposes, I kept model ID as much as possible for the normal model.

mingshl (approved):

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -191,4 +191,19 @@ public static boolean isEscapeUsed(String input) {
...
+
+    /**
+     * This method will define if we should print out model id with the error message or not.
+     * @param errorMessage
+     * @param modelId
+     * @param isHidden
+     * @return
+     */
+    public static String getErrorMessage(String errorMessage, String modelId, boolean isHidden) {
+        if (isHidden) {
```
ylwu-amzn, line 203: null checking ?

ylwu-amzn:
```suggestion
        if (BooleanUtils.isTrue(isHidden)) {
```

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -191,4 +191,19 @@ public static boolean isEscapeUsed(String input) {
...
     }
+
+    /**
+     * This method will define if we should print out model id with the error message or not.
+     * @param errorMessage
+     * @param modelId
+     * @param isHidden
+     * @return
+     */
+    public static String getErrorMessage(String errorMessage, String modelId, boolean isHidden) {
```
zane-neo, line 202: We should have a more general class initiated with `client` and expose method for logging, the method should automatically check if current user is super admin, because most of the cases we're not able to get the isHidden.

PR #2320, opened 2024-04-15
Fix connector can use subfield of secrectArn to bypass extraction of credential from AWS

b4sjoo:
### Description
We found connector can use subfield of `secretArn` (e.g. `secretArn.key`) to bypass extraction of credential from AWS, so we need to filter out any key string which tries to imitate the subfield of the `secretArn` of the credential map to avoid security issue. UT has added.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
@@ -148,7 +148,14 @@ public class MLCreateConnectorInput implements ToXContentObject, Writeable {
                     parameters = getParameterMap(parser.map());
                     break;
                 case CONNECTOR_CREDENTIAL_FIELD:
-                    credential = parser.mapStrings();
+                    // We need to filter out any key string that is trying to imitate the subfield of the secretArn of the credential map
+                    credential = new HashMap<>();
+                    Map<String, String> credentialKeyToAdd = parser.mapStrings();
+                    for (String key : credentialKeyToAdd.keySet()) {
+                        if (!key.startsWith("secretArn.")) {
+                            credential.put(key, credentialKeyToAdd.get(key));
+                        }
+                    }
                     break;
                 case CONNECTOR_ACTIONS_FIELD:
                     actions = new ArrayList<>();
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
@@ -25,6 +25,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


jngz-es (dismissed):

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -148,7 +148,14 @@ public static MLCreateConnectorInput parse(XContentParser parser, boolean update
                     parameters = getParameterMap(parser.map());
                     break;
                 case CONNECTOR_CREDENTIAL_FIELD:
-                    credential = parser.mapStrings();
+                    // We need to filter out any key string that is trying to imitate the subfield of the secretArn of the credential map
+                    credential = new HashMap<>();
+                    Map<String, String> credentialKeyToAdd = parser.mapStrings();
+                    for (String key : credentialKeyToAdd.keySet()) {
+                        if (!key.startsWith("secretArn.")) {
```
dhrubo-os, line 155: for Opensource we don't care about `secretArn.`, right? Please evaluate again if this code should be here or not? Or how can we make it generic?

R1V3RJ1s: The reason we submit this PR in github instead of gitfarm directly is to avoid future autosync conflict. As our AOS specific code accumulates we are receiving more and more sev2 TT on this issue, so before we are working out a clear SOP I would like to keep the github/gitfarm repo as sync as possible.

mingshl, line 155: if it's the connector to AWS, for example sageMager, it makes sense to not allowing using scretArn.

mingshl (approved):

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -148,7 +150,17 @@ public static MLCreateConnectorInput parse(XContentParser parser, boolean update
                     parameters = getParameterMap(parser.map());
                     break;
                 case CONNECTOR_CREDENTIAL_FIELD:
-                    credential = parser.mapStrings();
+                    // We need to filter out any key string that is trying to imitate the subfield of the secretArn of the credential map
```
dhrubo-os, line 153: [nit] change the comment then?

b4sjoo: Added

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
b4sjoo: All ARN subfield forbid

dhrubo-os (approved):

Zhangxunmt (approved):

PR #2332, opened 2024-04-16
Filtering hidden model info from model profiling for users other than superadmin

dhrubo-os:
### Description
[Filtering hidden model info from model profiling for users other than superadmin]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileRequest.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileRequest.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.action.profile;
 
 import java.io.IOException;
+import java.util.Set;
 
 import org.opensearch.action.support.nodes.BaseNodesRequest;
 import org.opensearch.core.common.io.stream.StreamInput;
@@ -13,15 +14,20 @@ import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.ml.profile.MLProfileInput;
 
 import lombok.Getter;
+import lombok.Setter;
 
 public class MLProfileRequest extends BaseNodesRequest<MLProfileRequest> {
 
     @Getter
     private MLProfileInput mlProfileInput;
+    @Getter
+    @Setter
+    private Set<String> hiddenModelIds;
 
     public MLProfileRequest(StreamInput input) throws IOException {
         super(input);
         mlProfileInput = new MLProfileInput(input);
+        hiddenModelIds = input.readSet(StreamInput::readString);
     }
 
     /**
@@ -37,5 +43,6 @@ public class MLProfileRequest extends BaseNodesRequest<MLProfileRequest> {
     public void writeTo(StreamOutput output) throws IOException {
         super.writeTo(output);
         mlProfileInput.writeTo(output);
+        output.writeCollection(hiddenModelIds, StreamOutput::writeString);
     }
 }
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileRequest.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileRequest.java
@@ -6,11 +6,14 @@
 package org.opensearch.ml.action.profile;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.Set;
 
+import org.opensearch.Version;
 import org.opensearch.action.support.nodes.BaseNodesRequest;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
 import org.opensearch.ml.profile.MLProfileInput;
 
 import lombok.Getter;
@@ -22,12 +25,15 @@ public class MLProfileRequest extends BaseNodesRequest<MLProfileRequest> {
     private MLProfileInput mlProfileInput;
     @Getter
     @Setter
-    private Set<String> hiddenModelIds;
+    private Set<String> hiddenModelIds = Collections.emptySet();
 
     public MLProfileRequest(StreamInput input) throws IOException {
         super(input);
+        Version streamInputVersion = input.getVersion();
         mlProfileInput = new MLProfileInput(input);
-        hiddenModelIds = input.readSet(StreamInput::readString);
+        if (streamInputVersion.onOrAfter(MLRegisterModelInput.MINIMAL_SUPPORTED_VERSION_FOR_AGENT_FRAMEWORK)) {
+            hiddenModelIds = input.readSet(StreamInput::readString);
+        }
     }
 
     /**
@@ -41,8 +47,11 @@ public class MLProfileRequest extends BaseNodesRequest<MLProfileRequest> {
 
     @Override
     public void writeTo(StreamOutput output) throws IOException {
+        Version streamOutputVersion = output.getVersion();
         super.writeTo(output);
         mlProfileInput.writeTo(output);
-        output.writeCollection(hiddenModelIds, StreamOutput::writeString);
+        if (streamOutputVersion.onOrAfter(MLRegisterModelInput.MINIMAL_SUPPORTED_VERSION_FOR_AGENT_FRAMEWORK)) {
+            output.writeCollection(hiddenModelIds, StreamOutput::writeString);
+        }
     }
 }
--- a/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
@@ -8,25 +8,41 @@ package org.opensearch.ml.action.profile;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import org.opensearch.action.FailedNodeException;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.nodes.TransportNodesAction;
+import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.env.Environment;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.CommonValue;
+import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.MLTask;
 import org.opensearch.ml.model.MLModelManager;
 import org.opensearch.ml.profile.MLModelProfile;
 import org.opensearch.ml.profile.MLProfileInput;
 import org.opensearch.ml.task.MLTaskManager;
+import org.opensearch.ml.utils.RestActionUtils;
 import org.opensearch.monitor.jvm.JvmService;
+import org.opensearch.search.SearchHit;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
+
+import com.google.common.annotations.VisibleForTesting;
 
 import lombok.extern.log4j.Log4j2;
 
@@ -36,6 +52,8 @@ public class MLProfileTransportAction extends
     private MLTaskManager mlTaskManager;
     private final JvmService jvmService;
     private final MLModelManager mlModelManager;
+
+    private final Client client;
 
     /**
      * Constructor
@@ -55,7 +73,8 @@ public class MLProfileTransportAction extends
         ActionFilters actionFilters,
         MLTaskManager mlTaskManager,
         Environment environment,
-        MLModelManager mlModelManager
+        MLModelManager mlModelManager,
+        Client client
     ) {
         super(
             MLProfileAction.NAME,
@@ -71,6 +90,7 @@ public class MLProfileTransportAction extends
         this.mlTaskManager = mlTaskManager;
         this.jvmService = new JvmService(environment.settings());
         this.mlModelManager = mlModelManager;
+        this.client = client;
     }
 
     @Override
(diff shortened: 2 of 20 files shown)


plugin/src/main/java/org/opensearch/ml/profile/MLModelProfile.java
```diff
@@ -79,6 +82,9 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         if (memSizeEstimationGPU != null) {
             builder.field("memory_size_estimation_gpu", memSizeEstimationGPU);
         }
+        if (isHidden != null) {
```
mingshl, line 85: why isHidden would be null?

dhrubo-os: This is how we build the builders. If the field is present then we put that in the builder. Check the previous code too. 

In this particular scenario isHidden could be null. When the model profiling is being initialized or being used that time we don't put any information about isHidden. We only set when we call the profile api.

mingshl: for boolean, should we give a default when it's read? so when registering the model, we don't know if a model isHidden or not.

dhrubo-os: This is not model registration object, this is model profile object. We set the isHidden value during model registration.

mingshl (approved):

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
```diff
@@ -103,27 +123,104 @@ private MLProfileNodeResponse createMLProfileNodeResponse(MLProfileRequest mlPro
...
                 }
-            }
-        });
+            });
+        }, e -> { log.error("Search Hidden model wasn't successful"); }), latch);
+
+        try {
+            latch.await(10, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            // Handle interruption if necessary
```
Zhangxunmt, line 164: Still return MLProfileNodeResponse for this node with empty results or a flag in the response showing this node exists in the cluster?

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
```diff
@@ -103,27 +123,104 @@ private MLProfileNodeResponse createMLProfileNodeResponse(MLProfileRequest mlPro
...
+                    if (mlProfileInput.isReturnAllTasks()
+                        || (!mlProfileInput.emptyModels() && targetModelIds.contains(mlTask.getModelId()))) {
+                        log.debug("Runtime task profile is found for model {}", mlTask.getModelId());
+                        mlLocalTasks.put(taskId, mlTask);
+                    }
+                }
+            });
+            Arrays.stream(mlModelManager.getAllModelIds()).forEach(modelId -> {
+                if (isSuperAdmin || !hiddenModels.contains(modelId)) {
+                    if (mlProfileInput.isReturnAllModels() || (!mlProfileInput.emptyModels() && targetModelIds.contains(modelId))) {
```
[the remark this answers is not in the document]
dhrubo-os: This isn't my change. My change is `if (isSuperAdmin || !hiddenModels.contains(modelId)) {`

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
dhrubo-os: it will return with a empty result.

Zhangxunmt: I mean returning "node-id": { models: ..., tasks: ... } rather than returning {}, so the Profile API results show the nodeId so we know the node still exists in the cluster.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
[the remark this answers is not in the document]
Zhangxunmt: But this is still making a nested IF and hurdles the readability?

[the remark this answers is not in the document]
dhrubo-os: I don't think so. However, do you have any better suggestion you want me to follow?

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
dhrubo-os: If any model isn't deployed or there isn't any task, currently it shows `{}`, so I don't think I'm breaking any CX experience.

Zhangxunmt: The difference is that what would be the implication of {} after this change? Is it expired or nothing in the memory at all?

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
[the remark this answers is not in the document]
Zhangxunmt: I think we can just use a function to return boolean like "shouldIncludeModel/Task". All the check ups can go into the function.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileRequest.java
```diff
@@ -6,22 +6,28 @@
...
     @Getter
     private MLProfileInput mlProfileInput;
+    @Getter
+    @Setter
+    private Set<String> hiddenModelIds;
 
     public MLProfileRequest(StreamInput input) throws IOException {
         super(input);
         mlProfileInput = new MLProfileInput(input);
+        hiddenModelIds = input.readSet(StreamInput::readString);
```
ylwu-amzn, line 30: Add BWC version checking? Profile request will broadcast to all nodes, if the cluster has two versions, it may throw exception.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
```diff
@@ -103,27 +112,45 @@ private MLProfileNodeResponse createMLProfileNodeResponse(MLProfileRequest mlPro
...
-            if (mlProfileInput.isReturnAllTasks() || (!mlProfileInput.emptyTasks() && mlProfileInput.getTaskIds().contains(taskId))) {
-                log.debug("Runtime task profile is found for model {}", mlTask.getModelId());
-                mlLocalTasks.put(taskId, mlTask);
-            }
-            if (mlProfileInput.isReturnAllTasks() || (!mlProfileInput.emptyModels() && targetModelIds.contains(mlTask.getModelId()))) {
-                log.debug("Runtime task profile is found for model {}", mlTask.getModelId());
-                mlLocalTasks.put(taskId, mlTask);
+            if (isSuperAdmin || !hiddenModels.contains(mlTask.getModelId())) {
+                if (mlProfileInput.isReturnAllTasks() || (!mlProfileInput.emptyTasks() && mlProfileInput.getTaskIds().contains(taskId))) {
+                    log.debug("Runtime task profile is found for model {}", mlTask.getModelId());
```
ylwu-amzn, line 125: Is this error message correct? Same with line 129

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesRequest.java
```diff
@@ -17,15 +17,20 @@
...
     @Getter
     private MLStatsInput mlStatsInput;
+    @Getter
+    @Setter
+    private Set<String> hiddenModelIds;
 
     public MLStatsNodesRequest(StreamInput in) throws IOException {
         super(in);
         mlStatsInput = new MLStatsInput(in);
+        hiddenModelIds = in.readSet(StreamInput::readString);
```
ylwu-amzn, line 33: Add BWC version check?

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
dhrubo-os: I didn't change anything [here](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java#L108-L115)

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesRequest.java
dhrubo-os: ok adding.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
```diff
@@ -157,29 +158,28 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
-                        if (mlStatsInput.retrieveStatsForModel(modelId)) {
-                            Map<ActionName, MLActionStats> actionStatsMap = new HashMap<>();
-                            for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
-                                if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
-                                    actionStatsMap.put(entry.getKey(), entry.getValue());
-                                }
+            Set<String> hiddenModels = mlStatsNodesRequest.getHiddenModelIds() != null
+                ? mlStatsNodesRequest.getHiddenModelIds()
+                : Collections.emptySet();
+            log.info("List of hidden models from Stats transport action: {}", Arrays.toString(hiddenModels.toArray()));
```
ylwu-amzn, line 165: Change to debug level? Stats API may run regularly, looks not necessary to print this

dhrubo-os: I'm removing this in the next commit. This I used to debug, but forgot to remove.

plugin/src/main/java/org/opensearch/ml/profile/MLModelProfile.java
```diff
@@ -79,6 +82,9 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+            builder.field("is_hidden", isHidden);
```
ylwu-amzn, line 86: Do we need to set `is_hidden: false` for normal model?

plugin/src/main/java/org/opensearch/ml/profile/MLModelProfile.java
```diff
@@ -104,6 +110,7 @@ public MLModelProfile(StreamInput in) throws IOException {
         }
         this.memSizeEstimationCPU = in.readOptionalLong();
         this.memSizeEstimationGPU = in.readOptionalLong();
+        this.isHidden = in.readOptionalBoolean();
```
ylwu-amzn, line 113: Add BWC version check

plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
```diff
@@ -10,19 +10,17 @@
...
-import static org.opensearch.ml.utils.RestActionUtils.onFailure;
 import static org.opensearch.ml.utils.RestActionUtils.splitCommaSeparatedParam;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
ylwu-amzn, line 16: Avoid import *

plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
```diff
@@ -124,33 +129,60 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+                client.search(searchRequest, ActionListener.runBefore(new ActionListener<SearchResponse>() {
+                    @Override
+                    public void onResponse(SearchResponse searchResponse) {
+                        Set<String> hiddenModelIds = new HashSet<>(searchResponse.getHits().getHits().length); // Set initial capacity to
+                        // the number of hits
+                        for (SearchHit hit : searchResponse.getHits()) {
+                            hiddenModelIds.add(hit.getId());
+                        }
+                        mlStatsNodesRequest.setHiddenModelIds(hiddenModelIds);
+                        log.info("List of hidden models from Rest ML Stats Action: {}", Arrays.toString(hiddenModelIds.toArray()));
```
ylwu-amzn, line 144: should we print hidden model id ?

dhrubo-os: No I'm removing this, forgot to remove.

plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
```diff
@@ -124,33 +129,60 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
-                        }, e -> {
-                            String errorMessage = "Failed to get ML model count";
-                            log.error(errorMessage, e);
-                            onFailure(channel, RestStatus.INTERNAL_SERVER_ERROR, errorMessage, e);
-                        })
-                    );
-            } else {
-                getNodeStats(finalMlStatsInput, clusterStatsMap, client, mlStatsNodesRequest, channel);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
```
ylwu-amzn, line 173: Do we need to add this try/catch ? I don't see it in the old code

plugin/src/main/java/org/opensearch/ml/stats/MLModelStats.java
```diff
@@ -25,15 +25,18 @@ public class MLModelStats implements ToXContentFragment, Writeable {
      * Example: {predict: { request_count: 1}}
      */
     private Map<ActionName, MLActionStats> modelStats;
+    private Boolean isHidden;
 
     public MLModelStats(StreamInput in) throws IOException {
         if (in.readBoolean()) {
             this.modelStats = in.readMap(stream -> stream.readEnum(ActionName.class), MLActionStats::new);
         }
+        this.isHidden = in.readOptionalBoolean();
```
ylwu-amzn, line 34: Add BWC version check?

plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
dhrubo-os: editor was showing me: `Unhandled exception: java.io.IOException`

plugin/src/main/java/org/opensearch/ml/profile/MLModelProfile.java
dhrubo-os: changed it to:

```
        if (isHidden != null && isHidden) {
            builder.field("is_hidden", true);
        }
        ```

File: plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java [hunk 14]
```diff
@@ -157,29 +157,24 @@ MLStatsNodeResponse createMLStatsNodeResponse(MLStatsNodesRequest mlStatsNodesRe
...
-                                }
+            Set<String> hiddenModels = mlStatsNodesRequest.getHiddenModelIds() != null
+                ? mlStatsNodesRequest.getHiddenModelIds()
+                : Collections.emptySet();
+            for (String modelId : mlStats.getAllModels()) {
+                if (isSuperAdmin || !hiddenModels.contains(modelId)) {
+                    if (mlStatsInput.retrieveStatsForModel(modelId)) {
+                        Map<ActionName, MLActionStats> actionStatsMap = new HashMap<>();
+                        for (Map.Entry<ActionName, MLActionStats> entry : mlStats.getModelStats(modelId).entrySet()) {
+                            if (mlStatsInput.retrieveStatsForAction(entry.getKey())) {
```
Zhangxunmt, line 169: We should refactor this 5 layer of nested for and ifs. This is far from elegant. At least we can use stream().filter(...) to avoid some.

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
```diff
@@ -103,27 +112,45 @@ private MLProfileNodeResponse createMLProfileNodeResponse(MLProfileRequest mlPro
...
-                log.debug("Runtime model profile is found for model {}", modelId);
-                MLModelProfile modelProfile = mlModelManager.getModelProfile(modelId);
-                if (modelProfile != null) {
-                    mlLocalModels.put(modelId, modelProfile);
+            if (isSuperAdmin || !hiddenModels.contains(modelId)) {
+                if (mlProfileInput.isReturnAllModels() || (!mlProfileInput.emptyModels() && targetModelIds.contains(modelId))) {
+                    log.debug("Runtime model profile is found for model {}", modelId);
+                    MLModelProfile modelProfile = mlModelManager.getModelProfile(modelId);
+                    if (modelProfile != null) {
+                        if (isSuperAdmin && hiddenModels.contains(modelId)) {
```
Zhangxunmt, line 140: It's becoming a 4 layer nested If...else... At least the last two ifs can be merged into one.

plugin/src/main/java/org/opensearch/ml/utils/IndexUtils.java
```diff
@@ -134,4 +138,23 @@ public void getNumberOfDocumentsInIndex(
         }
     }
 
+    public static SearchRequest buildHiddenModelSearchRequest() {
```
Zhangxunmt, line 141: Is this the right place to build the searchRequest? This IndexUtils is only for Index related method. Why not put this in the RestActionUtils?

plugin/src/main/java/org/opensearch/ml/action/profile/MLProfileTransportAction.java
dhrubo-os: Yeah agree, let me simplify this.

plugin/src/main/java/org/opensearch/ml/action/stats/MLStatsNodesTransportAction.java
dhrubo-os: +1, simplifying this.

plugin/src/main/java/org/opensearch/ml/utils/IndexUtils.java
dhrubo-os: well this is also index related search request. I thought of doing in `RestActionUtils` then felt like it can go here.

ylwu-amzn (approved):

PR #2351, opened 2024-04-23
not sending failure message when model index isn't present

dhrubo-os:
### Description
[
For a newly created cluster stats api response:

```
{
    "ml_model_index_status": "non-existent",
    "ml_connector_index_status": "non-existent",
    "ml_model_count": 0,
    "ml_controller_index_status": "non-existent",
    "ml_task_index_status": "non-existent",
    "ml_config_index_status": "green",
    "ml_connector_count": 0,
    "nodes": {
        "TLaJyhhfS0KeqLbdTRMJSA": {
            "ml_request_count": 0,
            "ml_executing_task_count": 0,
            "ml_deployed_model_count": 0,
            "ml_circuit_breaker_trigger_count": 0,
            "ml_jvm_heap_usage": 27,
            "ml_failure_count": 0,
            "algorithms": {},
            "models": {}
        }
    }
}
```

Profile api response:

```
{}
```
]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
@@ -154,7 +154,7 @@ public class RestMLProfileAction extends BaseRestHandler {
 
                     @Override
                     public void onFailure(Exception e) {
-                        onFailed(channel, "Searching model wasn't successful", e);
+                        channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
                     }
 
                 }, threadContext::restore));
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
@@ -154,7 +154,15 @@ public class RestMLProfileAction extends BaseRestHandler {
 
                     @Override
                     public void onFailure(Exception e) {
-                        channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+                        try {
+                            builder.startObject();
+                            builder.endObject();
+                            channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+                        } catch (IOException ex) {
+                            String errorMessage = "Failed to get ML node level profile";
+                            log.error(errorMessage, e);
+                            onFailed(channel, errorMessage, e);
+                        }
                     }
 
                 }, threadContext::restore));
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
@@ -176,7 +176,11 @@ public class RestMLStatsAction extends BaseRestHandler {
 
                     @Override
                     public void onFailure(Exception e) {
-                        onFailed(channel, RestStatus.INTERNAL_SERVER_ERROR, "Searching model wasn't successful", e);
+                        try {
+                            getNodeStats(finalMlStatsInput, clusterStatsMap, client, mlStatsNodesRequest, channel);
+                        } catch (IOException ex) {
+                            onFailed(channel, RestStatus.INTERNAL_SERVER_ERROR, "Failed to retrieve Cluster level metrics", e);
+                        }
                     }
                 }, threadContext::restore));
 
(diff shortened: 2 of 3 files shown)


ylwu-amzn:
Can you add more details and test results to description part ?

dhrubo-os:
> Can you add more details and test results to description part ?

added in the description

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
```diff
@@ -154,7 +154,15 @@ public void onResponse(SearchResponse searchResponse) {
 
                     @Override
                     public void onFailure(Exception e) {
-                        onFailed(channel, "Searching model wasn't successful", e);
+                        try {
+                            builder.startObject();
+                            builder.endObject();
+                            channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
```
ylwu-amzn, line 160: Is this same with the old response?

dhrubo-os: Yes. 

```
{}
```

jngz-es: I didn't get why we return empty content here if failing to search hidden models. In an example case, super-admin call profile api, we could continue to exeute MLProfileAction to return profiles even if failed to search hidden models, as user is super-admin.

plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
```diff
@@ -176,7 +176,11 @@ public void onResponse(SearchResponse searchResponse) {
 
                     @Override
                     public void onFailure(Exception e) {
-                        onFailed(channel, RestStatus.INTERNAL_SERVER_ERROR, "Searching model wasn't successful", e);
+                        try {
+                            getNodeStats(finalMlStatsInput, clusterStatsMap, client, mlStatsNodesRequest, channel);
```
rbhavna, line 180: Shouldn't we check if the failure is due to index not found exception and continue the stats execution? For every other failures, we can throw exception.

plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
dhrubo-os: This is to keep the experience same. Currently if there's any model in the index/cache, we return an empty object when we call profile api. I'm just keeping the experience same.

ylwu-amzn: If failed to search, is it possible the profile API will have other valid response which should be returned to user ? If yes, we should not just return empty `{}`

dhrubo-os: I tested in the newly created cluster (2.13.0) where any model isn't created. If no models in the cluster, we return {}

ylwu-amzn: I mean what if for other reason the search failed, not just for newly created cluster which has no model index

dhrubo-os: Even for other reason if the search fails, we need to make sure we aren't exposing any information of hidden model. If the search fails then we don't have any visibility of which one is the hidden model.

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/rest/RestMLStatsAction.java
rbhavna: okay makes sense. user should know nothing of hidden model. got it

rbhavna (approved):

plugin/src/main/java/org/opensearch/ml/rest/RestMLProfileAction.java
jngz-es: But super-admin should have permission to get everything even if hidden models, right? We shouldn't block super-admin to get information.

dhrubo-os: Yeah that's true. But super admin won't know which one is the hidden model. Because depending on that search we mark hidden model. If the search isn't successful, then that marking won't happen.

Issue #1343, opened 2023-09-15
[Enhancement] Simplify the current user experience and APIs in Ml-Commons


Zhangxunmt:
From customer feedbacks, building ML apps on top of Ml-Commons requires too many upfront setup or pre-requisite APIs calls that looks unnecessary or unwanted to them. Some improvement are done in https://github.com/opensearch-project/ml-commons/issues/1147 and https://github.com/opensearch-project/ml-commons/issues/1146 recently. However, there are still more to enhance to lower the bar for both internal and external customers to onboard Ml-Commons. 

We need to propose a roadmap to simplify as much as possible to help customers, starting with the below pain points: 
1. Allow remote inference without registering models. Users just want to create connector and assume it in their applications to run remote "Predict".
2. Remove the obstacle that are brought in by model_groups. 
3. Get ride of manually deploy/undeploy models that are required in the workflow. 

Feel free to add more thoughts and suggestions!

ylwu-amzn:
One more point: add blueprint id, so user can just input "cohere_embedding" , rather than copy the whole blue print

owaiskazi19:
This is the exact case we are working on. To simplify the setup for registering the model. Please take a look of https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/21 and https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/70. We are trying to automate creating a connector, model group then registering and deploying the remote model.

jonfritz:
@owaiskazi19 as part of the AI Flow project, are you also planning to simplify the APIs as well? Or, are you just creating a UI that interacts with the existing APIs? Thanks!

owaiskazi19:
> @owaiskazi19 as part of the AI Flow project, are you also planning to simplify the APIs as well? Or, are you just creating a UI that interacts with the existing APIs? Thanks!

@jonfritz we are creating wrapper APIs which would simplify the provision of complex setups of OpenSearch. Details for the API can be found [here](https://github.com/opensearch-project/opensearch-ai-flow-framework/issues/49). 
AI Flow project will have a separate backend and not just frontend. User can use the APIs to create a template based on the use case and execute it.

minalsha:
@Zhangxunmt please see https://github.com/opensearch-project/ml-commons/issues/1343#issuecomment-1753428715 by @owaiskazi19 to address the problems you highlighted in this issue.

Zhangxunmt:
Close as completed.

Issue #1148, opened 2023-07-20
[Enhancement] Auto-deploy ML Model when predict


Zhangxunmt:
Currently the ML models are manually "deployed" or "loaded" into the memory which requires customers to manually invoke a "deploy" API before using any ML models. Also after usage, ml-common requires a manual "undeploy" or "unload" from end users. This is adding more overhead to the system and end users to use ml-common in the workflow. 

We should build a auto-deploy mechanism to get rid of these "deploy" and "undeploy" operations in the workflow. Instead, we should auto deploy the model when customers use a model in the first time and setup a TTL to auto-undeploy from the system. In this way, the deploy and un-deploy APIs can be removed from the workflow and user experience are much simplified.

hijakk:
+1, automating management of availability of specific models would simplify operations significantly

owaiskazi19:
@Zhangxunmt thanks for the proposal. This looks like a much asked feature. Couple of questions around the same:
1. How does the API experience will look like for the auto deploy model? Would we have a param when registering a model something like `_register?auto_deploy=true`?

2. Will we still support `deploy` API for users to deploy the model manually or will we deprecate the API?
As part of automating the setup of ml-commons we support [DeployModelStep](https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/DeployModelStep.java) and [UndeployModelSetp](https://github.com/opensearch-project/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/workflow/UndeployModelStep.java) in flow framework. We might need to deprecate them here as well.

Zhangxunmt:
@owaiskazi19 , the BWC is still valid. Nothing is changed from your side. You can still setup deploy and undeploy in the flow frameworks. The API experience remain the same too. The model registration of "_register?auto_deploy=true" is still valid. 

This change only handles the case when a cluster scale up and down, restart, or node replacement, etc. We need to auto-deploy the models in the "Prediction" stage so customers don't need to keep manually deploying again and again after each event.

jngz-es:
https://github.com/opensearch-project/ml-commons/pull/2206

PR #2365, opened 2024-04-26
Add TTL to un-deploy model automatically

Zhangxunmt:
### Description
Reuse the current cron job and transport APIs to implement auto-undeploy ML_Models based on TTL after any model is un-used for a certain time. This is to mate the automatic remote model deployment and save memory for long idling local models. Verified in a single host cluster. Will verify in bigger cluster. 
 
### Issues Resolved
[https://github.com/opensearch-project/ml-commons/issues/1343] , https://github.com/opensearch-project/ml-commons/issues/1148

### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
@@ -25,37 +25,50 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
 @Getter
 public class MLDeploySetting implements ToXContentObject, Writeable {
     public static final String IS_AUTO_DEPLOY_ENABLED_FIELD = "is_auto_deploy_enabled";
-    public static final String MODEL_TTL_FIELD = "model_ttl";
+    public static final String MODEL_TTL_HOURS_FIELD = "model_ttl_hours";
+    public static final String MODEL_TTL_MINUTES_FIELD = "model_ttl_minutes";
     private static final long DEFAULT_TTL_HOUR = -1;
+    private static final long DEFAULT_TTL_MINUTES = -1;
 
     private Boolean isAutoDeployEnabled;
-    private Long modelTTL;  // Time to live in hours
+    private Long modelTTLInHours;  // Time to live in hours
+    private Long modelTTLInMinutes; // in minutes
 
     @Builder(toBuilder = true)
-    public MLDeploySetting(Boolean isAutoDeployEnabled) {
+    public MLDeploySetting(Boolean isAutoDeployEnabled, Long modelTTLInHours, Long modelTTLInMinutes) {
         this.isAutoDeployEnabled = isAutoDeployEnabled;
-        this.modelTTL = DEFAULT_TTL_HOUR;
-    }
-    @Builder(toBuilder = true)
-    public MLDeploySetting(Boolean isAutoDeployEnabled, Long modelTTL) {
-        this.isAutoDeployEnabled = isAutoDeployEnabled;
-        this.modelTTL = modelTTL;
+        this.modelTTLInHours = modelTTLInHours;
+        this.modelTTLInMinutes = modelTTLInMinutes;
+        if (modelTTLInHours == null && modelTTLInMinutes == null) {
+            this.modelTTLInHours = DEFAULT_TTL_HOUR;
+            this.modelTTLInMinutes = DEFAULT_TTL_MINUTES;
+            return;
+        }
+        if (modelTTLInHours == null) {
+            this.modelTTLInHours = 0L;
+        }
+        if (modelTTLInMinutes == null) {
+            this.modelTTLInMinutes = 0L;
+        }
     }
 
     public MLDeploySetting(StreamInput in) throws IOException {
         this.isAutoDeployEnabled = in.readOptionalBoolean();
-        this.modelTTL = in.readOptionalLong();
+        this.modelTTLInHours = in.readOptionalLong();
+        this.modelTTLInMinutes = in.readOptionalLong();
     }
 
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         out.writeOptionalBoolean(isAutoDeployEnabled);
-        out.writeOptionalLong(modelTTL);
+        out.writeOptionalLong(modelTTLInHours);
+        out.writeOptionalLong(modelTTLInMinutes);
     }
 
     public static MLDeploySetting parse(XContentParser parser) throws IOException {
         Boolean isAutoDeployEnabled = null;
-        Long modelTTL = null;
+        Long modelTTLHours = null;
+        Long modelTTLMinutes = null;
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
             String fieldName = parser.currentName();
@@ -64,14 +77,16 @@ public class MLDeploySetting implements ToXContentObject, Writeable {
                 case IS_AUTO_DEPLOY_ENABLED_FIELD:
                     isAutoDeployEnabled = parser.booleanValue();
                     break;
-                case MODEL_TTL_FIELD:
-                    modelTTL = parser.longValue();
+                case MODEL_TTL_HOURS_FIELD:
+                    modelTTLHours = parser.longValue();
+                case MODEL_TTL_MINUTES_FIELD:
+                    modelTTLMinutes = parser.longValue();
                 default:
                     parser.skipChildren();
                     break;
             }
         }
-        return new MLDeploySetting(isAutoDeployEnabled, modelTTL);
+        return new MLDeploySetting(isAutoDeployEnabled, modelTTLHours, modelTTLMinutes);
     }
 
     @Override
@@ -80,8 +95,11 @@ public class MLDeploySetting implements ToXContentObject, Writeable {
         if (isAutoDeployEnabled != null) {
             builder.field(IS_AUTO_DEPLOY_ENABLED_FIELD, isAutoDeployEnabled);
         }
-        if (modelTTL != null) {
-            builder.field(MODEL_TTL_FIELD, modelTTL);
+        if (modelTTLInHours != null) {
+            builder.field(MODEL_TTL_HOURS_FIELD, modelTTLInHours);
+        }
+        if (modelTTLInMinutes != null) {
+            builder.field(MODEL_TTL_MINUTES_FIELD, modelTTLInMinutes);
         }
         builder.endObject();
         return builder;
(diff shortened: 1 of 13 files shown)


plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -154,6 +168,8 @@ public void run() {
                         log.error("Failed to sync model routing", ex);
                     })
                 );
+            // Undeploy expired models
+            undeployExpiredModels(expiredModels.keySet(), modelWorkerNodes);
```
ylwu-amzn, line 172: The current logic is very aggressive: if model expired on any node, we will undeploy it. It's possible that model expires on one node , but the other 99 nodes not yet. 

Suggestion: Let's gather model last access time and TTL from each node. Then create a method to check if model expired on all nodes, we will undeploy model. This will also be flexible for future change. For example if we want to undeploy model when expired on more than half nodes, we can just change the method.

common/src/main/java/org/opensearch/ml/common/transport/sync/MLSyncUpNodeResponse.java
```diff
@@ -38,6 +40,7 @@ public MLSyncUpNodeResponse(StreamInput in) throws IOException {
         this.deployedModelIds = in.readOptionalStringArray();
         this.runningDeployModelIds = in.readOptionalStringArray();
         this.runningDeployModelTaskIds = in.readOptionalStringArray();
+        this.expiredModelIds = in.readOptionalStringArray();
```
ylwu-amzn, line 43: Add BWC version, ask @b4sjoo who knows the best

Zhangxunmt: will do

common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
```diff
@@ -25,37 +25,50 @@
 @Getter
 public class MLDeploySetting implements ToXContentObject, Writeable {
     public static final String IS_AUTO_DEPLOY_ENABLED_FIELD = "is_auto_deploy_enabled";
-    public static final String MODEL_TTL_FIELD = "model_ttl";
+    public static final String MODEL_TTL_HOURS_FIELD = "model_ttl_hours";
+    public static final String MODEL_TTL_MINUTES_FIELD = "model_ttl_minutes";
```
ylwu-amzn, line 29: Why separate hours and minutes? Seems use minutes should be enough.

Zhangxunmt: Initially only used Hours. But it may take too much time to test expiration in hours. Having both hours and minutes cover most cases. Most likely only hours will be used, e.g. 72 hours (6 days), etc.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Zhangxunmt: It costs too much to return lastAccessTime + TTL for each model and process all these info in the syncup Job, because we would need to construct a Map<modelId, Array of (lastAccessTime, TTL) for all nodes> and compute/compare for each entry. On the other hand, without adding new info, we can use the size of work nodes for expired model to do the same thing. If the model is expired in all deployed nodes, we will un-deploy.

common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java
ylwu-amzn: Why not just use minutes ?  24 hours will be 24 * 60,

Zhangxunmt: It should be easier to set a longer time in hours? For 1 week of 7 days, it would be 10080 minutes, but this is not an obvious number to understand the time length.

ylwu-amzn: I would suggest just use the minimal time unit for this, that could be easier. If you add hours, why not add days and months?

Zhangxunmt: Done.

common/src/main/java/org/opensearch/ml/common/transport/sync/MLSyncUpInput.java
```diff
@@ -35,6 +36,8 @@ public class MLSyncUpInput implements Writeable {
     // profile API has consistent data with model index.
     private Map<String, Boolean> deployToAllNodes;
 
+    public static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_TTL = Version.V_2_14_0;
```
ylwu-amzn, line 39: `MINIMAL_SUPPORTED_VERSION_FOR_MODEL_TTL` is more related with `common/src/main/java/org/opensearch/ml/common/model/MLDeploySetting.java`, can we move to `MLDeploySetting` ?

common/src/test/java/org/opensearch/ml/common/model/MLDeployingSettingTests.java
```diff
@@ -109,12 +109,12 @@ public void parseWithIllegalArgumentInteger() throws Exception {
...
     @Test
     public void parseWithIllegalField() throws Exception {
-        final String expectedInputStrWithIllegalField = "{\"is_auto_deploy_enabled\":true," +
+        final String expectedInputStrWithIllegalField = "{\"is_auto_deploy_enabled\":true," + "\"model_ttl_hours\":0," +
                 "\"illegal_field\":\"This field need to be skipped.\"}";
 
         testParseFromJsonString(expectedInputStrWithIllegalField, parsedInput -> {
             try {
-                assertEquals(expectedInputStr, serializationWithToXContent(parsedInput));
+                assertEquals("{\"is_auto_deploy_enabled\":true,\"model_ttl_minutes\":-1}", serializationWithToXContent(parsedInput));
```
ylwu-amzn, line 117: What if `model_ttl_minutes` is 0 ?

common/src/main/java/org/opensearch/ml/common/transport/sync/MLSyncUpInput.java
Zhangxunmt: Updated in the newest commit

common/src/test/java/org/opensearch/ml/common/model/MLDeployingSettingTests.java
Zhangxunmt: Just the opposite, with model_ttl_minutes == 0, the model will be removed in the next cron job running cycle right away. By default, without any given value for this ttl in minutes, model_ttl_minutes == -1 and it will make the model never expire.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -341,6 +345,29 @@ public String[] getLocalDeployedModels() {
...
+            }
+            Duration liveDuration = Duration.between(entry.getValue().getLastAccessTime(), Instant.now());
+            Long ttlInMinutes = mlModel.getDeploySetting().getModelTTLInMinutes();
+            if (ttlInMinutes < 0) {
+                return false;
+            }
+            Duration ttl = Duration.ofMinutes(ttlInMinutes);
+            boolean isModelExpired = liveDuration.getSeconds() >= ttl.getSeconds();
+            return isModelExpired
+                && (mlModel.getModelState() == MLModelState.DEPLOYED || mlModel.getModelState() == MLModelState.PARTIALLY_DEPLOYED);
```
ylwu-amzn, line 367: Why do we care the model status for model TTL ? For example, if we have some bug, and model stays on DEPLOY_FAILED, I think TTL could be a good way to remove model to avoid memory leak.

Zhangxunmt: The Deploy_Failed may indicate some insights to the cluster or ml service. Auto undeploying the model won't fix the problem but only hide it. We shouldn't use this as a tool to clear the failure status? Otherwise, the model will likely be in Failed status when it's deployed again.

ylwu-amzn: We have auto deploy, if model deploy stuck there, why shouldn't we use TTL to undeploy model to remove the unnecessary memory usage?

Zhangxunmt: That's doable. But is it hiding problems that may bite us more in the future? When the deploy stuck, why shouldn't we look into the reasons first and then un-deploy if it's the real solution? Also the auto deploy is only for remote model which takes little memory. So there might be other different reasons for the deployment failure like the one I showed in the meeting?

ylwu-amzn: Not sure what problem we are hidding. This is what you are building to automate the deploy lifecycle with auto deployment and TTL

ylwu-amzn (approved):

b4sjoo (approved):

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -126,6 +137,17 @@ public void run() {
                     }
                 }
             }
+
+            Set<String> modelsToUndeploy = new HashSet<>();
+            for (String modelId : expiredModelToNodes.keySet()) {
+                if (modelWorkerNodes.containsKey(modelId)
+                    && expiredModelToNodes.get(modelId).size() == modelWorkerNodes.get(modelId).size()) {
+                    // this model has expired in all the nodes
+                    modelWorkerNodes.remove(modelId);
```
ylwu-amzn, line 146: Cron job use modelWorkerNodes to check if model has been deployed to any nodes. This will cause sync up cron job set model as deploy_failed.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -341,6 +345,29 @@ public String[] getLocalDeployedModels() {
...
     }
 
+    /**
+     * Get expired models on node.
+     *
+     * @return array of expired model id
+     */
+    public String[] getExpiredModels() {
+        return modelCaches.entrySet().stream().filter(entry -> {
+            MLModel mlModel = entry.getValue().getCachedModelInfo();
```
ylwu-amzn, line 355: Cached model info only set when predict request comes. Check https://github.com/opensearch-project/ml-commons/pull/1472/files

For a deployed model, if no predict request comes, the cached model info is null.  Then the `mlModel` will be null and line 356 will throw null pointer exception, then sync up cron job will get no response.

ylwu-amzn: Add this line `modelCacheHelper.setModelInfo(modelId, mlModel);` when deploy model should fix the issue

PR #2389, opened 2024-05-01
set the mlModel info in the deploy stage

Zhangxunmt:
### Description
Set the mlModel info in the deployment. Also use the cached value outside of mlModel to obtain the ModelStatus.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -386,9 +386,11 @@ public class MLModelCacheHelper {
      */
     public String[] getExpiredModels() {
         return modelCaches.entrySet().stream().filter(entry -> {
-            MLModel mlModel = entry.getValue().getCachedModelInfo();
-            if (mlModel.getDeploySetting() == null) {
-                return false;  // no TTL, never expire
+            MLModelCache modelCache = entry.getValue();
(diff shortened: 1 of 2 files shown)


zane-neo:
Is this a critical issue that we need to release in 2.14?

zane-neo (dismissed):

zane-neo:
format code

dhrubo-os:
apply spotless

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -386,9 +386,11 @@ public String[] getLocalDeployedModels() {
      */
     public String[] getExpiredModels() {
```
dhrubo-os, line 387: Do we have any test for this method?

Zhangxunmt: The syncUpCron will cover this part. Only an unit tests will not find this edge case. It has to be in the scenario of "local model deployed and wait more than 10 seconds to predict".

dhrubo-os:
let's push commit rather than amending the commit.

Zhangxunmt:
> let's push commit rather than amending the commit.

No for spotless. Check the commit history. How many meaningless commits we have for a single PR?

ylwu-amzn (approved):

dhrubo-os (approved):

PR #2402, opened 2024-05-03
Fix local build failure for RestMLInferenceIngestProcessorIT

mingshl:
### Description
fix local build failure for RestMLInferenceIngestProcessorIT

when openai key is null, skip running predict
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
@@ -115,6 +115,10 @@ public class RestMLInferenceIngestProcessorIT extends MLCommonsRestTestCase {
         String index_name = "daily_index";
         createPipelineProcessor(createPipelineRequestBody, "diary_embedding_pipeline");
         createIndex(index_name, createIndexRequestBody);
+        // Skip test if key is null
+        if (OPENAI_KEY == null) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
```diff
@@ -115,6 +115,10 @@ public void testMLInferenceProcessorWithObjectFieldType() throws Exception {
         String index_name = "daily_index";
         createPipelineProcessor(createPipelineRequestBody, "diary_embedding_pipeline");
         createIndex(index_name, createIndexRequestBody);
+        // Skip test if key is null
+        if (OPENAI_KEY == null) {
```
ylwu-amzn, line 119: Can you share the failure message ? Curious why OPENAI_KEY will be null. If this is null, then our IT will always fail.

dhrubo-os: This happens only in local environment, when we don't set the `OPENAI_KEY` and it will be null and the IT will eventually fail. Same scenario for the OS 2.14 build I think as they don't have the `OPENAI_KEY` in their environment, these test will fail : https://build.ci.opensearch.org/blue/rest/organizations/jenkins/pipelines/integ-test/runs/8248/nodes/118/steps/383/log/?start=0

dhrubo-os (approved):

b4sjoo (approved):

PR #2405, opened 2024-05-03
avoid race condition in syncup model state refresh and handle NP of I…

Zhangxunmt:
### Description
Avoid race condition in sync up job to un-deploy expired models and refresh model states. Without this change, the sync up job sometimes sets wrong Model State after model is un-deployed. Default AutoDeploy to enabled when it's a null value. 

Verified and tested in local 2.14 cluster.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
@@ -9,6 +9,7 @@ import static org.opensearch.ml.common.CommonValue.CREATE_TIME_FIELD;
 import static org.opensearch.ml.common.CommonValue.MASTER_KEY;
 import static org.opensearch.ml.common.CommonValue.ML_CONFIG_INDEX;
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
+import static org.opensearch.ml.utils.RestActionUtils.getAllNodes;
 
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -97,6 +99,9 @@ public void run() {
         // gather running model/tasks on nodes
         client.execute(MLSyncUpAction.INSTANCE, gatherInfoRequest, ActionListener.wrap(r -> {
             List<MLSyncUpNodeResponse> responses = r.getNodes();
+            if (r.failures() != null && r.failures().size() != 0) {
+                log.debug("Received {} failures in the sync up response on nodes", r.failures().size());
```
ylwu-amzn, line 103: Seems not very helpful if we don't log the error message

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -168,37 +172,44 @@ public void run() {
...
-                    })
-                );
-            // Undeploy expired models
-            undeployExpiredModels(modelsToUndeploy, modelWorkerNodes);
+            client.execute(MLSyncUpAction.INSTANCE, syncUpRequest, ActionListener.wrap(re -> {
+                log.debug("sync model routing job finished");
+                if (!modelsToUndeploy.isEmpty()) {
+                    // Undeploy expired models
+                    undeployExpiredModels(modelsToUndeploy, modelWorkerNodes, deployingModels);
+                    return;
```
ylwu-amzn, line 180: Why return here ? If we have 10 models, `modelsToUndeploy` has 4 models. We should also call `refreshModelState` for the other 6, right?

Zhangxunmt: The refreshModelState is called in the listener of unDeployModels after the expired models are un-deployed. Check the undeployExpiredModels(), So the modelstate will be refreshed anyways.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Zhangxunmt: updated in the refresh.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -97,6 +99,9 @@ public void run() {
         // gather running model/tasks on nodes
         client.execute(MLSyncUpAction.INSTANCE, gatherInfoRequest, ActionListener.wrap(r -> {
             List<MLSyncUpNodeResponse> responses = r.getNodes();
+            if (r.failures() != null && r.failures().size() != 0) {
+                log.error("Received {} failures in the sync up response on nodes", r.failures().size());
```
ylwu-amzn, line 103: I think you misunderstood me. I mean we should log the error message in r.failures

Zhangxunmt: Added the messages in the logging.

ylwu-amzn (approved):

jngz-es (approved):

PR #2410, opened 2024-05-06
add a flag to distinguish duplicate remote model auto deploy and tran…

Zhangxunmt:
### Description
Adding a flag to avoid race condition of automatic deploying and transport deploying remote models, causing temporary Model deployment state errors. When this flag is true, the transport deploy will continue the deploy and do not throw exceptions. 
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.BooleanUtils;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.TokenBucket;
@@ -299,7 +300,7 @@ public class MLModelCacheHelper {
      */
     public boolean isAutoDeploying(String modelId) {
         MLModelCache modelCache = modelCaches.get(modelId);
-        return modelCache != null && modelCache.getIsAutoDeploying() == true;
+        return modelCache != null && BooleanUtils.isTrue(modelCache.getIsAutoDeploying());
     }
 
     /**
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -61,7 +61,7 @@ public class MLModelCacheHelper {
         List<String> targetWorkerNodes,
         boolean deployToAllNodes
     ) {
-        if (isModelRunningOnNode(modelId)) {
+        if (isModelRunningOnNode(modelId) && !isAutoDeploying(modelId)) {
             throw new MLLimitExceededException("Duplicate deploy model task");
         }
         log.debug("init model state for model {}, state: {}", modelId, state);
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -92,6 +92,7 @@ public synchronized void initModelStateLocal(
         modelCache.setDeployToAllNodes(false);
         modelCache.setLastAccessTime(Instant.now());
         modelCaches.put(modelId, modelCache);
+        setIsAutoDeploying(modelId, true);
```
ylwu-amzn, line 95: Looks `setIsAutoDeploying` method can be removed to simplify. Just need to add `modelCache.setIsAutoDeploying(true)` after line 93 ?

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -279,6 +280,28 @@ public Boolean getIsModelEnabled(String modelId) {
         return modelCache.getIsModelEnabled();
     }
 
+    /**
+     * Set a flag to show if model is in auto deploying status
+     *
+     * @param modelId        model id
+     * @param isModelAutoDeploying auto deploy flag
+     */
+    public synchronized void setIsAutoDeploying(String modelId, Boolean isModelAutoDeploying) {
```
ylwu-amzn, line 289: Do we really need to add this method ?

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -279,6 +280,28 @@ public Boolean getIsModelEnabled(String modelId) {
...
+
+    /**
+     * Check if model is in auto deploying.
+     *
+     * @param modelId model id
+     * @return true if model is auto deploying.
+     */
+    public boolean isAutoDeploying(String modelId) {
+        MLModelCache modelCache = modelCaches.get(modelId);
+        return modelCache != null && modelCache.getIsAutoDeploying() == true;
```
ylwu-amzn, line 302:
```suggestion
        return modelCache != null && BooleanUtils.isTrue(modelCache.getIsAutoDeploying());
```

Zhangxunmt: Updated to "modelCache.getIsAutoDeploying() != null && BooleanUtils.isTrue(modelCache.getIsAutoDeploying())", otherwise it will throw an Null Pointer Exception.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -299,7 +300,7 @@ public synchronized void setIsAutoDeploying(String modelId, Boolean isModelAutoD
      */
     public boolean isAutoDeploying(String modelId) {
         MLModelCache modelCache = modelCaches.get(modelId);
-        return modelCache != null && modelCache.getIsAutoDeploying() == true;
+        return modelCache != null && modelCache.getIsAutoDeploying() != null && BooleanUtils.isTrue(modelCache.getIsAutoDeploying());
```
ylwu-amzn, line 303: `BooleanUtils.isTrue` already covers null case, do we really need to add `modelCache.getIsAutoDeploying() != null` ?

Zhangxunmt: Checked and verified again. Shouldn't need so removed it already.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
Zhangxunmt: Updated.

jngz-es (approved):

b4sjoo (approved):

PR #2415, opened 2024-05-08
add deploySetting in registering local models

Zhangxunmt:
### Description
DeploySetting was not included in registering local models (pre-trained or from custom url), which was fine because it only included autoDeploy flag for remote models. Now we added TTL in the DeploySetting which applies to all models so We need to add it back.

Verified in local dev cluster.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/ModelHelper.java
@@ -27,6 +27,7 @@ import java.util.zip.ZipFile;
 
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.model.MLDeploySetting;
 import org.opensearch.ml.common.model.MLModelConfig;
 import org.opensearch.ml.common.model.MLModelFormat;
 import org.opensearch.ml.common.model.QuestionAnsweringModelConfig;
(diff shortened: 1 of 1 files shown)


Zhangxunmt:
> From the description
> 
> > DeploySetting was not included in registering pre-trained models, which was fine because it only included autoDeploy flag for remote models.
> 
> The code fix seems not just for pre-trained model. Seems we are fixing for all local models, right?

Yes it should be for all models.

b4sjoo (approved):

ylwu-amzn:
> > From the description
> > > DeploySetting was not included in registering pre-trained models, which was fine because it only included autoDeploy flag for remote models.
> > 
> > 
> > The code fix seems not just for pre-trained model. Seems we are fixing for all local models, right?
> 
> Yes it should be for all models.

Got it, you can update the description to make it accurate

dhrubo-os:
> > From the description
> > > DeploySetting was not included in registering pre-trained models, which was fine because it only included autoDeploy flag for remote models.
> > 
> > 
> > The code fix seems not just for pre-trained model. Seems we are fixing for all local models, right?
> 
> Yes it should be for all models. I just verified with registering model from url and it applies to that case too.

What about pre-trained models? What happens for those models? How can we add deploy settings to those models? Do we need to update the documentation as well for pre-trained models?

ylwu-amzn (approved):

ylwu-amzn (approved):

PR #2416, opened 2024-05-08
Fix flaky test of visualization tool

Hailong-am:
### Description
This PR is add waiting for model to be undeployed to fix flaky test introduced by #2363, 

https://github.com/opensearch-project/ml-commons/actions/runs/8933312245/job/24539185907?pr=2402
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/tools/ToolIntegrationWithLLMTest.java
+++ b/plugin/src/test/java/org/opensearch/ml/tools/ToolIntegrationWithLLMTest.java
@@ -8,15 +8,19 @@ package org.opensearch.ml.tools;
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 
+import lombok.SneakyThrows;
 import org.apache.hc.core5.http.ParseException;
 import org.junit.After;
 import org.junit.Before;
 import org.opensearch.client.Response;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.model.MLModelState;
 import org.opensearch.ml.rest.RestBaseAgentToolsIT;
 import org.opensearch.ml.utils.TestHelper;
 
@@ -26,6 +30,10 @@ import lombok.extern.log4j.Log4j2;
 
 @Log4j2
 public abstract class ToolIntegrationWithLLMTest extends RestBaseAgentToolsIT {
+
+    private static final int MAX_TASK_RESULT_QUERY_TIME_IN_SECOND = 60 * 5;
+    private static final int DEFAULT_TASK_RESULT_QUERY_INTERVAL_IN_MILLISECOND = 1000;
+
     protected HttpServer server;
     protected String modelId;
     protected String agentId;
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/opensearch/ml/tools/ToolIntegrationWithLLMTest.java
```diff
@@ -63,9 +71,41 @@ public void stopMockLLM() {
...
 
+    @SneakyThrows
+    private void waitModelUndeployed(String modelId) {
+        Predicate<Response> condition = response -> {
+            try {
+                Map<String, Object> responseInMap = parseResponseToMap(response);
+                String state = responseInMap.get(MLModel.MODEL_STATE_FIELD).toString();
+                return !state.equals(MLModelState.DEPLOYED.toString())
+                    && !state.equals(MLModelState.DEPLOYING.toString())
+                    && !state.equals(MLModelState.PARTIALLY_DEPLOYED.toString());
```
yuye-aws, line 86: Is undeployed state either MLModelState.UNDEPLOYED or MLModelState.DEPLOY_FAILED?

yuye-aws: https://github.com/opensearch-project/neural-search/blob/7c54c86e256cfb43d0e7623bda1f7590c4a18b0e/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java#L932C1-L934C67

Hailong-am: I refer the logic from here https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L326

>Is undeployed state either MLModelState.UNDEPLOYED or MLModelState.DEPLOY_FAILED?

Yes, i think most of case should be these two. There have two more `REGISTERING`,`REGISTERED` allow to delete.

yuye-aws: Since we are undeploying the model, why we will get REGISTERING,REGISTERED state?

yuye-aws: Also from `isModelNotDeployed` function, the model state cannot be LOADED, LOADING and PARTIALLY_LOADED

Hailong-am: As we do undeploy first, the state could only be `UNDEPLOYED` and `DEPLOY_FAILED`, i have changed to check these two state.

plugin/src/test/java/org/opensearch/ml/tools/ToolIntegrationWithLLMTest.java
```diff
@@ -63,9 +72,39 @@ public void stopMockLLM() {
...
+            } catch (Exception e) {
+                return false;
+            }
+        };
+        waitResponseMeetingCondition("GET", "/_plugins/_ml/models/" + modelId, null, condition);
+    }
+
+    @SneakyThrows
+    protected Response waitResponseMeetingCondition(String method, String endpoint, String jsonEntity, Predicate<Response> condition) {
+        for (int i = 0; i < MAX_TASK_RESULT_QUERY_TIME_IN_SECOND; i++) {
```
ylwu-amzn, line 95: The worst case this line will run `MAX_TASK_RESULT_QUERY_TIME_IN_SECOND` times, it will be 300 ? For every retry, it will sleep one second, so it will wait for 300 seconds, that's a long time. Do we really need so long time ? We should fail quickly with a reasonable time

Hailong-am: yeah, reduce to total 30s that should be enough for model to be undeployed.

ylwu-amzn (approved):

rbhavna:
Thanks @Hailong-am for fixing the flaky tests. @ylwu-amzn This issue is not just for flaky tests. The undeploy response is giving `UNDEPLOYED` immediately but we still have to rely on get model to see if the model is indeed undeployed. Could be confusing to users. We should either give a task_id or say something like `UNDEPLOY INITIATED` in the undeploy response if the model is not completely undeployed

dhrubo-os:
> Thanks @Hailong-am for fixing the flaky tests. @ylwu-amzn This issue is not just for flaky tests. The undeploy response is giving `UNDEPLOYED` immediately but we still have to rely on get model to see if the model is indeed undeployed. Could be confusing to users. We should either give a task_id or say something like `UNDEPLOY INITIATED` in the undeploy response if the model is not completely undeployed

Seems like @ylwu-amzn created an [issue](https://github.com/opensearch-project/ml-commons/issues/2419) for this.

dhrubo-os (approved):

dhrubo-os:
Do we need to backport this to 2.14 branch? I don't see any backport label to this PR.

ylwu-amzn:
> Do we need to backport this to 2.14 branch? I don't see any backport label to this PR.

I think this test case seems not flaky in the Release CI workflow. We don't see IT failure report yet. To not block the whole release. I think it's fine to not backport

Issue #2417, opened 2024-05-08
[BUG] BWC issue with async http client change


ylwu-amzn:
**What is the bug?**
We have changed to async http client in 2.14 (https://github.com/opensearch-project/ml-commons/issues/1839), tested with [2.14 RC4](https://github.com/opensearch-project/opensearch-build/issues/4562#issuecomment-2099600151), find the predict result for text embedding model not same with 2.13.

**How can one reproduce the bug?**
1. Create two test clusters: 2.13 and [2.14 RC4](https://github.com/opensearch-project/opensearch-build/issues/4562#issuecomment-2099600151)
2. Follow steps of this issue https://github.com/opensearch-project/ml-commons/issues/2249 to create model
3. Then run this test on 2.13 and 2.14 test cluster
```
POST /_plugins/_ml/_predict/text_embedding/<your_model_id>
{
  "text_docs":[ "today is sunny", "hello world"],
  "return_number": true,
  "target_response": ["sentence_embedding"]
}
```
2.13 result
```
{
  "inference_results": [
    {
      "output": [
        {
          "name": "sentence_embedding",
          "data_type": "FLOAT32",
          "shape": [
            768
          ],
          "data": [
            0.009424215,
            -0.008311393,
            0.06740056,
            ...
           ]
        }
      ],
      "status_code": 200
    },
    {
      "output": [
        {
          "name": "sentence_embedding",
          "data_type": "FLOAT32",
          "shape": [
            768
          ],
          "data": [
            0.010724226,
            0.055782758,
            0.027084162,
            ...
          ]
        }
      ],
      "status_code": 200
    }
  ]
}
```
2.14 result
```
{
  "inference_results": [
    {
      "output": [
        {
          "name": "sentence_embedding",
          "data_type": "FLOAT32",
          "shape": [
            768
          ],
          "data": [
            0.009424215,
            -0.008311393,
            0.06740056,
            ...
          ]
        }
      ],
      "status_code": 200
    }
  ]
}
```

**What is the expected behavior?**
We should not break BWC. Should return same result of 2.13.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

zane-neo:
The issue is caused by the refactor of the httpclient from sync to async, when using sync httpclient with user defined preprocess function, a list of string input will be processed in sequential order, user script always pick up the first element in the list, and the list shrinks when a prediction is done. 
[...]

PR #2418, opened 2024-05-08
Fix user defined preprocess function missing prediction issue

zane-neo:
### Description
When users use self defined preprocess scripts to process inputs, there'll be missing predictions for all inputs in the list except the first one.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2417
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -99,6 +102,15 @@ private Tuple<Integer, Integer> calculateChunkSize(TextDocsInputDataSet textDocs
                 return Tuple.tuple(textDocsLength / stepSize + 1, stepSize);
             }
         } else {
+            Optional<ConnectorAction> predictAction = getConnector().findPredictAction();
```
dhrubo-os, line 105: Can we have unit test for this section?

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -22,6 +23,7 @@ public class MLPreProcessFunction {
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
     public static final String TEXT_DOCS_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.bedrock.embedding";
+    public static final String TEXT_IMAGE_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.multimodal.embedding";
```
dhrubo-os, line 26: Can we raise a separate PR for this functionality only.

ylwu-amzn:
Suggest not merge multiple things in one PR. That makes it hard to rollback/backport

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -99,6 +102,15 @@ private Tuple<Integer, Integer> calculateChunkSize(TextDocsInputDataSet textDocs
...
+            if (predictAction.isEmpty()) {
+                throw new IllegalArgumentException("no predict action found");
+            }
+            String preProcessFunction = predictAction.get().getPreProcessFunction();
+            if (preProcessFunction != null && !MLPreProcessFunction.contains(preProcessFunction)) {
+                // user defined preprocess script, this case, the chunk size is always equals to text docs length.
+                return Tuple.tuple(textDocsLength, 1);
```
ylwu-amzn, line 112: Is this understanding correct ?
1. `textDocsLength` means how many chunks
2. `1` means step or chunk size ?

If correct, why hard code the chunk size as 1 ?  This issue https://github.com/opensearch-project/ml-commons/issues/2417 is an example, it's not the only case. For example some user may process two documents in one chunk.

ylwu-amzn: https://github.com/opensearch-project/ml-commons/blob/2.x/docs/remote_inference_blueprints/bedrock_connector_titan_multimodal_embedding_blueprint.md

zane-neo: textDocsLength is chunks, 1 means step. When user process two documents in one chunk, user has to specify the `input_docs_processed_step_size`, in this case, this won't enter this branch. Multi-modal is a case that user needs to handle two documents in one chunk, I've tested this case and it works well.

dhrubo-os (approved):

xinyual (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
ylwu-amzn: Thanks for testing multi-modal case. 
It's possible that step size > 1 and no `input_docs_processed_step_size`. For example, before async http client, user can confiture pre_process function to read three documents and construct `"inputs": [doc1, doc2, doc3]`, when response comes back, will move to next 3 docs

zane-neo: This is a rare case, and we have workaround which is to add `input_docs_processed_step_size` configuration.

ylwu-amzn: Thanks, based on our discussion, this sounds good

Issue #2358, opened 2024-04-24
[RFC] Enhancement of Multi-Tenancy Capabilities in ML-Commons


dhrubo-os:
# RFC: Enhancement of Multi-Tenancy in ML-Commons Plugin

## 1. Introduction

This RFC details the proposed low-level design for enhancing multi-tenancy in the ML-Commons plugin, aimed at providing robust, scalable, and secure management of machine learning models for each tenant within a highly distributed environment.

## 2. High-Level Context

The proposed enhancements to the ML-Commons plugin are designed to support the evolving requirements of cloud architectures transitioning towards serverless models. This RFC aligns with the trend of enhancing cloud-native plugins to facilitate sophisticated multi-tenancy capabilities.

### Strategic Benefits of Multi-Tenancy:

- **Cost Efficiency**: Multi-tenancy allows for resource consolidation across multiple tenants, reducing costs through economies of scale.
- **Reduced Management Overhead**: Administrative tasks are centralized, significantly reducing management overhead and enhancing operational efficiency.
- **Scalability and Flexibility**: ML-Commons support can be distributed as an application-based microservice, enhancing scalability and flexibility.

On a high level, ML-Common support can be distributed as an application based micro service. 

![Screenshot 2024-04-24 at 4 06 11 PM](https://github.com/opensearch-project/ml-commons/assets/109556906/d5cb7ee9-7a18-494e-93a7-2c207bfb0757)

## 3. Purpose

The necessity for this enhancement arises from the growing demand for platform capabilities that can support multiple tenants simultaneously without compromising on performance, security, or scalability. In this context, a "tenant" refers to an individual customer or distinct operational unit, each with its own secure data and configurations. Allowing multiple tenants to coexist on the same application installation effectively means we can manage separate customers or business units independently yet within a single shared environment. Effective multi-tenancy will enable our infrastructure to:

* Securely Segregate Tenant Data: Ensure that each tenant's data is completely isolated from others, thereby protecting customer information and adhering to data privacy regulations.
* Efficiently Allocate and Manage Computational Resources: Dynamically distribute system resources among tenants based on demand, ensuring optimal performance and cost efficiency.
* Scale Operations Dynamically: Adjust resource allocation and service capacity in real-time as tenant demands change, thus supporting fluctuating workloads without affecting the quality of service.

This approach not only maximizes resource utilization but also enhances operational flexibility, making it possible to cater to a broad spectrum of customer needs efficiently.

## 3. Design Considerations

[...]

PR #2430, opened 2024-05-09
[Feature/multi_tenancy] [WIP] SdkClient interface and implementation

dbwiddis:
### Description

**Work In Progress**

Creates an abstraction for index metadata, see https://github.com/opensearch-project/OpenSearch/issues/13336 and initial implementation for Connector.

The near-term goal is proving out the flexibility of the interface.

Still TODO:
 - Update and Search APIs
 - Thorough exception handling and user error messages
 - More documentation

### Issues Resolved

Part of #2358
 
### Check List
  - [x] All tests pass


plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
```diff
@@ -0,0 +1,104 @@
...
+    private final Client client;
+    private final NamedXContentRegistry xContentRegistry;
+
+    public XContentClient(Client client, NamedXContentRegistry xContentRegistry) {
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+    }
+
+    @Override
+    public CompletionStage<PutCustomResponse> putCustom(PutCustomRequest request) {
```
dhrubo-os, line 50: What about `Update`?

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public CompletionStage<DeleteCustomResponse> deleteCustom(DeleteCustomRequest request) {
+        CompletableFuture<DeleteCustomResponse> future = new CompletableFuture<>();
+        client
+            .delete(
+                new DeleteRequest(request.index(), request.id()),
+                ActionListener
+                    .wrap(
+                        r -> future.complete(new DeleteCustomResponse.Builder().id(r.getId()).deleted(r.getResult() == DELETED).build()),
```
dhrubo-os, line 98: What happens if the deletion fails? How are we sending the error response back to the customer?

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
dbwiddis: Coming soon to a PR near you, but prioritizing initial APIs to support multitenancy work.

common/src/main/java/org/opensearch/sdk/DeleteCustomResponse.java
```diff
@@ -0,0 +1,76 @@
...
+     * Returns the document id
+     * @return the id
+     */
+    public String id() {
+        return id;
+    }
+
+    /**
+     * Returns whether deletion was successful
+     * @return true if deletion was successful, false if the object was not found
```
dhrubo-os, line 30: what happens if the object was found but some reason we couldn't delete.

dbwiddis: The onFailure / exception path will be followed instead.  This is the same behavior as the current DeleteResponse where you can have either DELETED (true) or NOT_FOUND (false) successful response.

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
dbwiddis: The actionListener for the whole execution is being completed exceptionally.  I have not yet gone through to more carefully identify exception possibilities and give more detailed user information.  Again, this is an early-stage WIP.

dhrubo-os:
I'm seeing:

```
/home/runner/work/ml-commons/ml-commons/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java:43: error: HttpConnector is not abstract and does not override abstract method parse(XContentParser) in Custom
public class HttpConnector extends AbstractConnector {
```

Are you planning to look into this?

dbwiddis:
> I'm seeing:
> 
> ```
> /home/runner/work/ml-commons/ml-commons/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java:43: error: HttpConnector is not abstract and does not override abstract method parse(XContentParser) in Custom
> public class HttpConnector extends AbstractConnector {
> ```
> 
> Are you planning to look into this?

Yeah, I missed actually implementing the `parse()` method in my own interface that needs to wrap the (static) xContent parsing methods in `Connector`.  I'll push a fix tomorrow.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -632,6 +634,9 @@ public Collection<Object> createComponents(
             .getClusterSettings()
             .addSettingsUpdateConsumer(MLCommonsSettings.ML_COMMONS_RAG_PIPELINE_FEATURE_ENABLED, it -> ragSearchPipelineEnabled = it);
 
+        SdkClient sdkClient = new SdkClient();
+        sdkClient.setImplementation(new XContentClient(client, xContentRegistry));
+
```
arjunkumargiri, line 639: Add a todo to initialize client based on settings. There should be support to set different implementations.

common/src/main/java/org/opensearch/sdk/SdkClient.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import java.util.concurrent.CompletionStage;
+
+public class SdkClient {
```
arjunkumargiri, line 13: This should be an interface and XContentClient should implement the the interface.

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
```diff
@@ -0,0 +1,104 @@
...
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.sdk.DeleteCustomRequest;
+import org.opensearch.sdk.DeleteCustomResponse;
+import org.opensearch.sdk.GetCustomRequest;
+import org.opensearch.sdk.GetCustomResponse;
+import org.opensearch.sdk.PutCustomRequest;
+import org.opensearch.sdk.PutCustomResponse;
+import org.opensearch.sdk.SdkClient;
+
+public class XContentClient extends SdkClient {
```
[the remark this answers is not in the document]
dbwiddis: No, because `NodeClient` is an implementation. This is a wrapper around `o.o.client.Client` interface.  The implementation is not defined.  (Yeah, there's an `AbstractClient` subclass and `NodeClient` is the only core subclass, but this code doesn't care about that.).  `OpenSearchClient` has specific meaning related to the java client so is not a good choice (in fact, that's probably what I will name the remote cluster http implementation).

This client manipulates the `_source` field containing structured object data in one of four mediatypes, defined by a content-type field, often JSON but possibly YAML or others.  Perhaps `StructuredObjectClient` or `MediaTypeClient` might be better?

I'm happy to entertain naming conversations but don't want to bikeshed too much here.

common/src/main/java/org/opensearch/sdk/SdkClient.java
dbwiddis: Yes, I know and if you look at the commit history you'll see it was initially implemented that way.

The problem I encountered is trying to inject it into the transport actions via `@Inject` annotation, which means I needed to return an `SdkClient` instance via the `createComponents()` extention point.   OpenSearch just uses the runtime class as the key to bind to, which is the concrete implementation, not the class.   It's possible to bind to the interface if you have access to the injector, but plugins don't.  

Eventually this object/implementation will probably be provided to the plugin at implementation, and this whole problem will go away.

I could create a new wrapper class to inject and still build against the interface, which will look a lot like this class.  In the interest of moving forward with the intent to get multi-tenancy solved within the next week, I don't want to spend too much time on this temporary workaround.

common/src/main/java/org/opensearch/sdk/SdkClient.java
```diff
@@ -0,0 +1,31 @@
...
+        return impl.putCustom(request);
+    }
+
+    public CompletionStage<GetCustomResponse> getCustom(GetCustomRequest request) {
+        return impl.getCustom(request);
+    }
+
+    public CompletionStage<DeleteCustomResponse> deleteCustom(DeleteCustomRequest request) {
+        return impl.deleteCustom(request);
+    }
```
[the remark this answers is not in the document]
dbwiddis: We could but I'm not sure of the value of that.  Nothing requires the implementation to act asynchronously, I could always immediately complete the future (and in fact for some error handling/illegal argument cases, would immediately complete exceptionally).    Sync methods would require more exception throwing/handling.

Not opposed to it if we need it but I don't see the need for it (yet).

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
dbwiddis: There should be, but probably not in the plugin (see previous comment on why a concrete class and not an interface).  I think the implementation will be determined elsewhere and the plugin will just be handled an initialized object.

For development/testing it might be nice to have a setting-based switch between two different implementations on-the-fly.

common/src/main/java/org/opensearch/sdk/SdkClient.java
arjunkumargiri: Plugins can make use `createGuiceModules` to inject custom guice modules. In the custom guice module we should be able to inject required implementation. Sample code to override plugin module: https://github.com/arjunkumargiri/ml-commons/commit/49e8f1ad95922f61e6be8b806a1808179f0c6c79#diff-5fb7a6291568399efed0b4be65f70e91206634d3a1e5d1265243f1c204c0d81f

common/src/main/java/org/opensearch/sdk/SdkClient.java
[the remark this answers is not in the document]
arjunkumargiri: I think it is primarily with coding style of the plugin. Given that we are building a generic interface it might be better to support both and let client chose their implementation style.

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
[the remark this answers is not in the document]
arjunkumargiri: Client implementation should provide information about the data store it supports. XContent client provides information that it operates on the _source field but lacks information about the underlying data store.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
arjunkumargiri: It might be better to handle plugin dependencies with in the plugin rather than relying on another component. We might be able to make use of `createGuiceModules` to initialize dependencies within plugin.

common/src/main/java/org/opensearch/sdk/GetCustomRequest.java
```diff
@@ -0,0 +1,124 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.search.fetch.subphase.FetchSourceContext;
+
+public class GetCustomRequest {
+
+    private final String index;
```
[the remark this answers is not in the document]
dbwiddis: In general we're talking about a "table" and a (unique) "primary key" here, with the retrieved data an "object" or "row".   I'm ok with something more generic here but the concepts are the same, and since this is OpenSearch, I don't see any reason to _not_ use OpenSearch terminology of index, document id, and document.   We can obviously amplify in the Javadocs what the more generic term is, as I did in my POC [here](https://github.com/dbwiddis/flow-framework/blob/b7384f25f48a55081bb687eff7417bc967de49bf/src/main/java/org/opensearch/flowframework/indices/DynamoDbUtil.java).

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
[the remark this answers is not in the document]
dbwiddis: > the underlying data store.

How about `ClusterIndexClient`?

common/src/main/java/org/opensearch/sdk/SdkClient.java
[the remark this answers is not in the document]
dbwiddis: Makes sense, but the comment about exception handling remains.  With the `CompletionStage` we clearly can pass exceptions back to the caller, even if executed synchronously.   Should we consistently throw a particular type of exception from these methods, and if so, what type?  I'd lean toward `OpenSearchStatusException` lacking any other preference.

common/src/main/java/org/opensearch/sdk/SdkClient.java
dbwiddis: Nice! I had forgotten that extension point existed!  I'll go ahead and use it.

common/src/main/java/org/opensearch/sdk/SdkClient.java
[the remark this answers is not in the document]
dbwiddis: So, we can't have two methods with the same name and arguments, but different return type.
```java
    public CompletionStage<PutCustomResponse> putCustom(PutCustomRequest request);

    public PutCustomResponse putCustom(PutCustomRequest request);
```
Possible to add an "async" name to the CompletionStage version.

Also we could just have default "synchronous" implementations:
```java
    public CompletionStage<PutCustomResponse> putCustom(PutCustomRequest request);

    default PutCustomResponse putCustomSync(PutCustomRequest request) {
        CompletableFuture<PutCustomResponse> future = new CompletableFuture<>();
        putCustom(request).whenComplete((r, ex) -> {
            if (r == null) {
               future.completeExceptionally(ex);
            } else {
                future.complete(r);
            }            
        });
        try {
            return future.get();
        } catch (InterruptedException ie) {
            FutureUtils.cancel(future);
            Thread.currentThread().interrupt();
            throw new OpenSearchException(ie);
        } catch (java.util.concurrent.ExecutionException e) {
            throw new OpenSearchException(e.getCause());
        }
    }
```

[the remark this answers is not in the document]
arjunkumargiri: yes, this approach looks good.

plugin/src/main/java/org/opensearch/ml/sdkclient/XContentClient.java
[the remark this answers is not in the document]
arjunkumargiri: Sounds good, thanks.

common/src/main/java/org/opensearch/sdk/GetCustomRequest.java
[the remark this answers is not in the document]
arjunkumargiri: Got it, will be good to also add a comment around how the index could map to a table.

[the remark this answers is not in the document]
dbwiddis: I'll fine tune all the docs once we get the functionality demonstrated

dbwiddis:
Ready for incremental review/approval to permit additional multitentant work
 - Integ tests pass on local cluster implementation
   - I originally had more parsing of a data object but really we just need ToXContentObject because to keep things generic I had to return a parser... so that may be removed
 - Needs unit testing/coverage of interfaces/implementations
 - Failing unit tests for connectors are commented out with `// FIXME`, I will get to fixing those
 - Remote cluster get request is entirely untested at the moment...

common/src/main/java/org/opensearch/sdk/DataObject.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentParser;
+
+import java.io.IOException;
+
+/**
+ * A structured object used for Data Exchange.
+ */
+public interface DataObject extends ToXContentObject {
```
arjunkumargiri, line 19: Can ToXContentObject be directly used? What is the benefit of using a new interface `DataObject`?

arjunkumargiri (commented):
Lets add unit tests to the new classes introduced.

common/src/main/java/org/opensearch/sdk/DataObject.java
dbwiddis: > Lets add unit tests to the new classes introduced.

Yes I know this, just trying to not slow up development.

> Can ToXContentObject be directly used? What is the benefit of using a new interface DataObject?

Originally I thought to use the `parse()` method to produce an instance of the object but that proved too difficult with class casting, requiring tons of mucking around with reflection, anonymous classes, etc.   I finally decided to remove that and just return the parser.  So right now there's no added benefit ... but maybe we'll want to include some multitenant stuff in it?

Again, trying to rapidly iterate here on a feature branch to avoid holding other work up.

plugin/src/test/java/org/opensearch/ml/sdkclient/TestDataObject.java
```diff
@@ -0,0 +1,48 @@
+package org.opensearch.ml.sdkclient;
```
dhrubo-os, line 1: Add license

dbwiddis: Would be nice if spotless added the licenses. :|

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
```diff
@@ -0,0 +1,118 @@
...
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.core.common.Strings;
+import org.opensearch.core.index.shard.ShardId;
+
+public class DeleteDataObjectResponse {
+    private final String id;
+    private final ShardId shardId;
```
arjunkumargiri, line 16: Is `shardId` required? This can tie implementation specific to opensearch?

common/src/main/java/org/opensearch/sdk/PutDataObjectRequest.java
```diff
@@ -0,0 +1,86 @@
...
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.core.xcontent.ToXContentObject;
+
+public class PutDataObjectRequest {
+
+    private final String index;
+    private final ToXContentObject dataObject;
```
arjunkumargiri, line 16: Can java bounded generic types be used instead of XContentObject? 

https://docs.oracle.com/javase/tutorial/java/generics/bounded.html 

Generic type can enable strong type checking while performing CRUD operations.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -0,0 +1,71 @@
...
+    public static final String REGION = "REGION";
+
+    private final String remoteMetadataEndpoint;
+    private final String region;
+
+    /**
+     * Instantiate this module using environment variables
+     */
+    public SdkClientModule() {
+        this(System.getenv(REMOTE_METADATA_ENDPOINT), System.getenv(REGION));
```
arjunkumargiri, line 38: Optionally can cluster settings be used to fetch metadata endpoint?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param region The region
+     */
+    SdkClientModule(String remoteMetadataEndpoint, String region) {
+        this.remoteMetadataEndpoint = remoteMetadataEndpoint;
+        this.region = region;
+    }
+
+    @Override
+    protected void configure() {
+        boolean local = Strings.isNullOrEmpty(remoteMetadataEndpoint);
```
arjunkumargiri, line 53: Looks good for initial implementation. Lets add a follow up task to support other types of implementations. Ideally adding a new implementation should not require changes to plugin code base.

plugin/src/test/java/org/opensearch/ml/action/connector/TransportCreateConnectorActionTests.java
```diff
@@ -213,7 +217,7 @@ public void test_execute_connectorAccessControlEnabled_success() {
             return null;
         }).when(client).index(any(IndexRequest.class), isA(ActionListener.class));
         action.doExecute(task, request, actionListener);
-        verify(actionListener).onResponse(any(MLCreateConnectorResponse.class));
+        // FIXME verify(actionListener).onResponse(any(MLCreateConnectorResponse.class));
```
dhrubo-os, line 220: What's the reason for commenting these?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
dbwiddis: Unfortunately, yes. I didn't add anything I didn't have to.  

This was required for `DeleteConnectorTransportAction` which processes an `ActionListener<DeleteResponse>` (which is eventually parsed and returned to the user) as a response to the delete API:
```json
{
    "_index": "the_index",
    "_id": "vZAyLY8BaPsQZP-UDULE",
    "_version": 2,
    "result": "deleted",
    "_shards": {
        "total": 1,
        "successful": 1,
        "failed": 0
    },
    "_seq_no": 1,
    "_primary_term": 1
}
```
It's not allowed to be null:
```java
this.shardId = Objects.requireNonNull(shardId);
```
I can manufacture one (and the builder provides the facility to do that) but it's gotta be there. 😦

common/src/main/java/org/opensearch/sdk/PutDataObjectRequest.java
dbwiddis: What value would that bring?

I did initially use generics, primarily to enable a typed return value from the GET implementation.  However, that failed because the `parse()` (or sometimes `fromXContent()`) implementations are static methods.  I could use reflection to call those methods, but they do not have a consistent signature.  

For example, `Connector` is an interface and has a static `createConnector()` method to do the parsing.   

I can't enforce signatures on static methods like I can instance methods with an interface... which I did try briefly with a `parse()` method required to be implemented (that just wrapped the static method).

But for me to call that I need to create an instance; but `Connector` is an interface so I'd have to create an anonymous implementation of all the abstract methods... which are very specific to `Connector` and impossible to put in a generic implementation.

So I finally just decided to return the `parser` from the Get method and let the caller use it to create the object.

That gave me a leftover `DataObject` that no longer used its parse method that you suggested removing [here](https://github.com/opensearch-project/ml-commons/pull/2430#discussion_r1602169911).

So what's the benefit here in an `Object<? extends ToXContentObject> foo` vs. `ToXContentObject foo`?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
dbwiddis: Assuming you have easy access to the cluster settings from the plugin, yes.  In the current plugin environment we have settings available, but in earlier Extensions SDK work we had to create some overhead to subscribe to them.

Lots of ways to do this, I just based this implementation on [Your POC here](https://github.com/arjunkumargiri/ml-commons/commit/49e8f1ad95922f61e6be8b806a1808179f0c6c79#diff-c548f309dba4a5c4be8a5d19a7c7bd23649f7183c7a6819d585e779c49829d31).

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
dbwiddis: Adding a new implementation requires adding the implementation class always, so I don't think there's any way to get around updating this class to choose from among the options based on whatever criteria we decide on (settings, environment variable, command line argument, etc.).  Not sure how we can future-proof this as it's an extension point that runs at startup.

dbwiddis:
Current status:
 - manually tested create/get/delete on local cluster
   - "happy path" works well except the shard stats (failed, successful, total) are returning null to the user on deletion
   - GET on nonexistent connector fails on null content (String.length() at GetConnectorTransportAction.java:83)
 - validated remote cluster client is implemented based on system settings
   - API calls failing because I don't yet have credentials provided

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
arjunkumargiri: Adding new implementations to plugin repo could increase the size of the plugin. Instead we should be able to dynamically load implementation during runtime. Ideally these implementations should be loaded during runtime.

common/src/main/java/org/opensearch/sdk/PutDataObjectRequest.java
arjunkumargiri: Makes sense. Generic types would have helped with type checking but I agree that it not a primary requirement. Only additional challenge might be with client having to parse the request but opensearch plugins already operate this way, so it shouldn't be a concern.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
dbwiddis: > Instead we should be able to dynamically load implementation during runtime. 

Agreed, particularly since there's a lot of specific configuration around these (user credentials, access keys, SSL keystore and keystore password, etc.). 

For now I've just switched this to a basic http RestClient implementation to get the rest of the code working, and will circle back to properly doing this later.

PR #2431, opened 2024-05-09
add IT for remote model automatic deploy with TTL

Zhangxunmt:
### Description
Add IT to auto deploy a remote model in Predict, and auto un-deploy it through TTL.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
@@ -15,6 +15,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.hc.core5.http.message.BasicHeader;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.rules.ExpectedException;
 import org.opensearch.client.Response;
@@ -171,19 +172,6 @@ public class RestMLRemoteInferenceIT extends MLCommonsRestTestCase {
         assertEquals((Double) 1.0, (Double) ((Map) ((Map) responseMap.get("hits")).get("total")).get("value"));
     }
 
-    public void testRegisterRemoteModel() throws IOException, InterruptedException {
-        Response response = createConnector(completionModelConnectorEntity);
-        Map responseMap = parseResponseToMap(response);
-        String connectorId = (String) responseMap.get("connector_id");
-        response = registerRemoteModel("openAI-GPT-3.5 completions", connectorId);
-        responseMap = parseResponseToMap(response);
-        String taskId = (String) responseMap.get("task_id");
-        waitForTask(taskId, MLTaskState.COMPLETED);
-        response = getTask(taskId);
-        responseMap = parseResponseToMap(response);
-        assertNotNull(responseMap.get("model_id"));
-    }
-
     public void testDeployRemoteModel() throws IOException, InterruptedException {
         Response response = createConnector(completionModelConnectorEntity);
         Map responseMap = parseResponseToMap(response);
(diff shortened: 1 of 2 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
```diff
@@ -841,6 +875,46 @@ public static Response registerRemoteModel(String name, String connectorId) thro
...
+            + "  \"model_group_id\": \""
+            + modelGroupId
+            + "\",\n"
+            + "  \"version\": \"1.0.0\",\n"
+            + "  \"description\": \"test model\",\n"
+            + "  \"connector_id\": \""
+            + connectorId
+            + "\",\n"
+            + "  \"deploy_setting\": "
+            + " { \"model_ttl_minutes\": "
```
dhrubo-os, line 910: `model_ttl_minutes` can this be float? Like if I want `0.5` minutes => 30 seconds?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
```diff
@@ -237,6 +238,39 @@ public void testPredictRemoteModel() throws IOException, InterruptedException {
...
+        responseList = (List) responseMap.get("choices");
+        if (responseList == null) {
+            assertTrue(checkThrottlingOpenAI(responseMap));
+            return;
+        }
+        responseMap = (Map) responseList.get(0);
+        assertFalse(((String) responseMap.get("text")).isEmpty());
+
+        getModelProfile(modelId, verifyRemoteModelDeployed());
+        TimeUnit.SECONDS.sleep(71);
```
dhrubo-os, line 270: Why do we have to wait 11 more seconds?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
dhrubo-os: In addition, I'm curious to know more about this `model_ttl_minutes`. Let's say, I set `model_ttl_minutes` to 1 minutes and the model got deployed. Now at around 50th Second I made a prediction request with that model. In that case will the model still get undeployed after 10 seconds or it will start waiting for another 60 seconds?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
```diff
@@ -237,6 +238,39 @@ public void testPredictRemoteModel() throws IOException, InterruptedException {
         assertFalse(((String) responseMap.get("text")).isEmpty());
     }
 
+    public void testPredictWithAutoDeployAndTTL_RemoteModel() throws IOException, InterruptedException {
```
dhrubo-os, line 241: Can't we use any existing test? So that we don't need any extra api call again? Also waiting for 71 seconds!!!!

May be use one test and keep the model Id public so that at the end of this whole IT test, we can see if that model is already un deployed or not for not being used?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
Zhangxunmt: TTL resets when the model is used in Prediction. It's only based on idling time. The number of ttl is a long number because Java Duration doesn't support float number. If you need to use seconds, the time unit needs to change to seconds. But TTL Time granularity at minutes is fine enough.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
Zhangxunmt: auto undeploy is in the cron job which runs every 10 seconds in the IT. So the max time is ttl + 10.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
Zhangxunmt: It's in my plan to remove some other duplicate tests to save time.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRemoteInferenceIT.java
```diff
@@ -457,6 +428,7 @@ public void testOpenAIEditsModel() throws IOException, InterruptedException {
         assertFalse(((String) responseMap.get("content")).isEmpty());
     }
 
+    @Ignore
```
dhrubo-os, line 431: Any reason for ignoring these tests?

Zhangxunmt: Running the text-embedding model should be enough for OpenAI. No need to verify every other models to reduce the IT time. This moderations model is not quite used anyways for now.

jngz-es (approved):

mingshl (approved):

Issue #2436, opened 2024-05-10
[BUG] Flaky test PredictionITTests


Hailong-am:
**What is the bug?**

PredictionITTests is flaky, CI runs https://github.com/opensearch-project/ml-commons/actions/runs/9025991264/job/24802601995?pr=2435

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
```
REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:test' --tests "org.opensearch.ml.action.prediction.PredictionITTests.testPredictionWithDataFrame_BatchRCF" -Dtests.seed=CC4F8C51F54E62C2 -Dtests.security.manager=false -Dtests.locale=fr-FR -Dtests.timezone=Pacific/Niue -Druntime.java=17
    java.lang.AssertionError: Shard [.plugins-ml-config][0] is still locked after 5 sec waiting
```

```
[2024-05-09T15:11:50,845][INFO ][o.o.m.a.p.PredictionITTests] [testPredictionWithDataFrame_BatchRCF] after test
  1> [2024-05-09T15:11:50,845][INFO ][o.o.t.OpenSearchTestClusterRule] [testPredictionWithDataFrame_BatchRCF] [PredictionITTests#testPredictionWithDataFrame_BatchRCF]: cleaning up after test
  1> [2024-05-09T15:11:50,885][WARN ][o.o.c.r.a.AllocationService] [node_s0] Falling back to single shard assignment since batch mode disable or multiple custom allocators set
  1> [2024-05-09T15:11:50,894][INFO ][o.o.c.m.MetadataDeleteIndexService] [node_s0] [.plugins-ml-model/JrJHF6J_TEKPCBru8sqBRA] deleting index
  1> [2024-05-09T15:11:50,894][INFO ][o.o.c.m.MetadataDeleteIndexService] [node_s0] [iris_data_for_prediction_it/baRt5UxBTXCsYtA4d-bt_w] deleting index
  1> [2024-05-09T15:11:50,894][WARN ][o.o.c.r.a.AllocationService] [node_s0] Falling back to single shard assignment since batch mode disable or multiple custom allocators set
  1> [2024-05-09T15:11:50,905][INFO ][o.o.t.s.M.Listener       ] [node_s1] [.plugins-ml-model][0] start check index
[...]

PR #2437, opened 2024-05-10
fix flaky test of PredictionITTests and RestConnectorToolIT

Hailong-am:
### Description

1. Add per-defined cluster setting to disable MLSyncUpJob for JVM private OpenSearch Cluster, so that it will not create `.plugin-ml-config` index automatically every 10 seconds.
2. Skip RestConnectorToolIT if `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are not set
3. Add retry for creating connector if ML encryption master key has not been initialized yet

### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2436
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/action/MLCommonsIntegTestCase.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/MLCommonsIntegTestCase.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.action;
 import static org.opensearch.ml.common.input.parameter.regression.LogisticRegressionParams.ObjectiveType.LOGMULTICLASS;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_NATIVE_MEM_THRESHOLD;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_ONLY_RUN_ON_ML_NODE;
+import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_SYNC_UP_JOB_INTERVAL_IN_SECONDS;
 import static org.opensearch.ml.utils.RestActionUtils.getAllNodes;
 import static org.opensearch.ml.utils.TestData.TARGET_FIELD;
 import static org.opensearch.ml.utils.TestData.TIME_FIELD;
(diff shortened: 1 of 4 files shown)


ylwu-amzn:
@Hailong-am Can you rebase with main branch and check if IT can pass?

plugin/src/test/java/org/opensearch/ml/action/MLCommonsIntegTestCase.java
```diff
@@ -93,15 +94,27 @@
...
-public class MLCommonsIntegTestCase extends OpenSearchIntegTestCase {
+public class MLCommonsIntegTestCase extends ParameterizedStaticSettingsOpenSearchIntegTestCase {
     private Gson gson = new Gson();
 
+    /**
+     * set ML_COMMONS_SYNC_UP_JOB_INTERVAL_IN_SECONDS to 0 to disable ML_COMMONS_SYNC_UP_JOB
+     * the cluster will be pre-created with the settings at startup
+     */
+    public MLCommonsIntegTestCase() {
+        super(Settings.builder().put(ML_COMMONS_SYNC_UP_JOB_INTERVAL_IN_SECONDS.getKey(), 0).build());
```
ylwu-amzn, line 111: Will this set sync up job interval as 0 for all integ test cases ? If yes, we should avoid this

Hailong-am: Here are the implementations of this base class, I am not quite sure they will depends on `MLSyncUpCron`
image

ylwu-amzn: Hard code this in the base class looks risky. Can we add this to the child class which is flaky?

Hailong-am: sure, move the settings to `PredictionITTests`, will add it one by one if we have other similar in future.

ylwu-amzn:
> Add per-defined cluster setting to disable MLSyncUpJob for JVM private OpenSearch Cluster, so that it will not create .plugin-ml-config index automatically every 10 seconds.

This seems a code bug. Can you confirm this is just impacting IT ?

ylwu-amzn (approved):
Approve to unblock falied IT for 2.15 release

rbhavna (approved):

Issue #2429, opened 2024-05-09
[Enhancement] Error handling for Sagemaker throttle exception


ylwu-amzn:
**Is your feature request related to a problem?**
Some user reports 4xx error from Sagemaker when bulk ingest data https://github.com/opensearch-project/ml-commons/issues/2249

The reason is Sagemaker has concurrency limit which is dynamic and based on the number of instances and instance types used by the endpoint. When exceeds the limit, Sagemaker will return 400 error code and null error message `{"message": null}` ([doc link](https://docs.aws.amazon.com/sagemaker/latest/APIReference/CommonErrors.html))

**What solution would you like?**
Sagemaker error message for throttle exception is null, that's not easy to understand. The error code `400` is misleading, generally we expect 429 error code. 

Before Sagemaker team making enhancement, we can add some specific logic in ml-commons to handle such error and return readable error message to user.

**What alternatives have you considered?**
Ask Sagemaker team to return readable error message and correct error code 429.

**Do you have any additional context?**
No

Issue #2249, opened 2024-03-20
[BUG] Neural search: 4xx error ingesting data with Sagemaker external model


tiagoshin:
**What is the bug?**
I'm trying to use a model hosted in a Sagemaker endpoint in the same AWS Account as the Opensearch cluster to perform a Neural search.
The issue that I observe is that, while ingesting data into the index, I observe the following error for many documents:
```
        {
            "index": {
                "_index": "my-index",
                "_id": "id",
                "status": 400,
                "error": {
                    "type": "status_exception",
                    "reason": "Error from remote service: {\"message\":null}"
                }
            }
        }
``` 
I don't see any logs in OpenSearch error logs, and I don't see any 4xx or 5xx requests in Sagemaker. 
This error only happens with a reasonable amount of data in bulk ingestion, which in this case is 250 records. When I ingest only 20 records, it works.
I already tested getting some documents that failed and tried to ingest them separately, and it worked. So, the issue is not with the document or with the Sagemaker model.

**How can one reproduce the bug?**
1. First, deploy the `bge-base-en-v1.5` embedding model in Sagemaker using this python script:
```
from sagemaker.jumpstart.model import JumpStartModel

model_id = "huggingface-sentencesimilarity-bge-base-en-v1-5"
env = {
	'MMS_JOB_QUEUE_SIZE': '100000',
}
text_embedding_model = JumpStartModel(
    model_id=model_id,
    env=env,
    role="<YOUR-SAGEMAKER-ROLE>",
)

predictor = text_embedding_model.deploy(
	initial_instance_count=1, 
	instance_type='ml.g5.xlarge'
)

```
2. Once it's deployed, get the SageMaker endpoint.
3. Create a Sagemaker connector in OpenSearch:
```
POST {{host}}/_plugins/_ml/connectors/_create
{
  "name": "Amazon Sagemaker connector",
  "description": "The connector to Sagemaker",
  "version": 1,
  "protocol": "aws_sigv4",
  "credential": {
    "roleArn": "<YOUR-ROLE>"
  },
  "parameters": {
    "region": "<YOUR-REGION>",
    "service_name": "sagemaker"
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "headers": {
        "content-type": "application/json"
      },
      "url": "<YOUR-SAGEMAKER-ENDPOINT>",
[...]

Issue #2438, opened 2024-05-11
[RFC] Add retry with backoff for SageMaker throttling exception to mitigate the data lost problem


zhichao-aws:
## Problem Statement
related issue: #2429 
Some OpenSearch users deploy their ML models at SageMaker to encode documents. But some times users get throttling exceptions, especially during ingestion, even with bulk size 100. It is hard to diagnose the problem, and the failed documents won't be ingested (may lead to data lost). This problem harms the user experience with several main perspectives:

### 1. Error log is hard to diagnose
With a bulk request, the throttled documents fail, but other documents success. And the entire bulk request don’t throw exception. The response looks like this:
```json
{
  "took": 71,
  "ingest_took": 4444,
  "errors": true,
  "items": [
    {
      "index": {
        "_index": "scifact",
        "_id": "33",
        "status": 400,
        "error": {
          "type": "status_exception",
          "reason": "Error from remote service: {\"message\":null}"
        }
      }
    },
    ...
  ]
}
```
If the exception is not handled by client, data is lost. The exception information is in the response header and is hard to diagnose.

### 2. Users get throttling exception even without large workload
As mentioned before, we can still get the throttling exception even with a single bulk request of size 100. The SageMaker endpoint can definitely handle it (in fact, the maximum throughput is ~300 docs/s in our experiments). And this harms customer experience a lot, they have to split one bulk request to smaller ones. 

## Solution
SageMaker throttles the requests by max concurrent request numbers. I set up an endpoint with pseudo handler, which just sleep 50ms and return a fix vector. This pseudo handler consumes little CPU/GPU resources, but I can still get throttling exception.

We can devide the throttling scenerio by 2 cases: 1. users do have large workload 2. users don't have large workload, but they get throttled because the bulk request invoke the SageMaker for each document, leading to large concurrent requests. For case 1, the user has to upgrade to a stronger SageMaker instance or use more instances. For case 2, we can mitigate this problem by adding strategy of retry with exponential backoff.

## experiments
I did some experiments at my side to verify the feasibility of retry with backoff.
I use the [neural sparse model](https://huggingface.co/opensearch-project/opensearch-neural-sparse-encoding-v1) and ml.g4dn.xlarge. (deployment script reference [link](https://github.com/zhichao-aws/neural-search/tree/neural_sparse_sagemaker/neural_sparse_sagemaker_example), here I turn off the async logging for stable performance). All documents consist of fixed 128 tokens.
[...]

PR #2442, opened 2024-05-13
hanlde the throttling error in the response header

Zhangxunmt:
### Description
Handle the throttling error in the header when the error is only returned in the headers and response body is empty. This is for a specific case discussed in the issues below. 
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2429, https://github.com/opensearch-project/ml-commons/issues/2249
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections.MapUtils;
 import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.util.Strings;
 import org.opensearch.OpenSearchStatusException;
@@ -80,6 +81,23 @@ public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandle
         SdkHttpFullResponse sdkResponse = (SdkHttpFullResponse) response;
         log.debug("received response headers: " + sdkResponse.headers());
         this.statusCode = sdkResponse.statusCode();
+        if (MapUtils.isEmpty(sdkResponse.headers())) {
+            return;
+        }
+        List<String> errorsInHeader = sdkResponse.headers().get("x-amzn-ErrorType");
+        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
+            return;
+        }
+        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
+        if (containsThrottlingException) {
+            actionListener
+                .onFailure(
+                    new OpenSearchStatusException(
+                        REMOTE_SERVICE_ERROR + "The request was denied due to request throttling.",
+                        RestStatus.fromCode(statusCode)
+                    )
+                );
+        }
     }
 
     @Override
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
@@ -109,6 +109,8 @@ public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandle
         if (errorsInHeader == null || errorsInHeader.isEmpty()) {
             return;
         }
+        // Check the throttling exception from AMZN servers, e.g. sageMaker.
+        // See [https://github.com/opensearch-project/ml-commons/issues/2429] for more details.
         boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
         if (containsThrottlingException && executionContext.getExceptionHolder().get() == null) {
             log.error("Remote server returned error code: {}", statusCode);
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
@@ -39,6 +39,7 @@ import software.amazon.awssdk.http.async.SdkAsyncHttpResponseHandler;
 
 @Log4j2
 public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandler {
+    public static final String AMZ_ERROR_HEADER = "x-amzn-ErrorType";
     @Getter
     private Integer statusCode;
     @Getter
@@ -81,22 +82,9 @@ public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandle
         SdkHttpFullResponse sdkResponse = (SdkHttpFullResponse) response;
         log.debug("received response headers: " + sdkResponse.headers());
         this.statusCode = sdkResponse.statusCode();
-        if (MapUtils.isEmpty(sdkResponse.headers())) {
-            return;
-        }
-        List<String> errorsInHeader = sdkResponse.headers().get("x-amzn-ErrorType");
-        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
-            return;
-        }
-        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
-        if (containsThrottlingException) {
-            actionListener
-                .onFailure(
-                    new OpenSearchStatusException(
-                        REMOTE_SERVICE_ERROR + "The request was denied due to request throttling.",
-                        RestStatus.fromCode(statusCode)
-                    )
-                );
+        if (statusCode < HttpStatus.SC_OK || statusCode > HttpStatus.SC_MULTIPLE_CHOICES) {
+            handleThrottlingInHeader(sdkResponse);
+            // add more handling here for other exceptions in headers
         }
     }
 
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -80,6 +81,23 @@ public void onHeaders(SdkHttpResponse response) {
         SdkHttpFullResponse sdkResponse = (SdkHttpFullResponse) response;
         log.debug("received response headers: " + sdkResponse.headers());
         this.statusCode = sdkResponse.statusCode();
+        if (MapUtils.isEmpty(sdkResponse.headers())) {
```
zhichao-aws, line 84: Should we only check the header after checking the statusCode?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -80,6 +81,23 @@ public void onHeaders(SdkHttpResponse response) {
...
+        }
+        List<String> errorsInHeader = sdkResponse.headers().get("x-amzn-ErrorType");
+        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
+            return;
+        }
+        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
+        if (containsThrottlingException) {
+            actionListener
+                .onFailure(
+                    new OpenSearchStatusException(
```
zhichao-aws, line 95: How about create a exception class like RemoteThrottlingException, so that we can identify the exception type quickly without checking its string content. For example, we need to check its type to implement retry logic https://github.com/opensearch-project/ml-commons/issues/2438

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -80,6 +81,23 @@ public void onHeaders(SdkHttpResponse response) {
...
+            return;
+        }
+        List<String> errorsInHeader = sdkResponse.headers().get("x-amzn-ErrorType");
+        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
+            return;
+        }
+        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
+        if (containsThrottlingException) {
+            actionListener
+                .onFailure(
```
zhichao-aws, line 94: I have a big concern here. Although we call actionListener.onFailure in onHeaders, this doesn't prevent the onStream being invoked(I confirmed it using debugger tool). And we'll still subscribe the response and trigger the actionListener.onFailure again at the MLResponseSubscriber. And this can lead to bugs to call actionListener.onFailure twice.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Haven't fully understand your question. But without this line, it would throw NPE when obtaining the errors in the header.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Fair enough. But if I miss declaring this new class, please add it in your next PR for the retry implementation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Thanks for the callout. I had the same concern but didn't verify it yet yesterday. Already have this fixed in the new commit.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: I mean we can only check the header after checking the status code < 200 or > 300. Since most responses don't have throttling exception.
Not a blocking issue, just a suggestion

Zhangxunmt: good suggestion. Updated in the new commit.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -95,6 +101,29 @@ public void onError(Throwable error) {
...
+        List<String> errorsInHeader = sdkResponse.headers().get(AMZ_ERROR_HEADER);
+        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
+            return;
+        }
+        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
+        if (containsThrottlingException && executionContext.getExceptionHolder().get() == null) {
+            log.error("Remote server returned error code: {}", statusCode);
+            executionContext
+                .getExceptionHolder()
+                .compareAndSet(
```
ylwu-amzn, line 117: The line 146 will set the exception as http body `{"message":null}`. Have you confirmed this line will happen after line 146 ? If not, the error will still be `{"message":null}`, right ? Have you done some manual testing if IT is hard/flaky to add?

Zhangxunmt: Actually it will not anymore because the executionContext.getExceptionHolder() isn't null anymore. Line 146 only sets the exception when there is nothing in the ExceptionHolder. But for the throttling errors, this PR has set an exception in the ExceptionHolder so it will be not be set anymore (line 146 will not invoke). I have verified it using a multi-node cluster.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -95,6 +101,29 @@ public void onError(Throwable error) {
...
 
+    private void handleThrottlingInHeader(SdkHttpFullResponse sdkResponse) {
+        if (MapUtils.isEmpty(sdkResponse.headers())) {
+            return;
+        }
+        List<String> errorsInHeader = sdkResponse.headers().get(AMZ_ERROR_HEADER);
+        if (errorsInHeader == null || errorsInHeader.isEmpty()) {
+            return;
+        }
+        boolean containsThrottlingException = errorsInHeader.stream().anyMatch(str -> str.startsWith("ThrottlingException"));
```
dhrubo-os, line 112: Will this be able to catch these kind of exceptions:

1. google.api.core.retry.RetryingExecutor$ThrottlingException:
2. Microsoft.Azure.Cosmos.ThrottlingException
3. org.opensearch.action.ThrottlingException

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -38,6 +39,7 @@
 
 @Log4j2
 public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandler {
+    public static final String AMZ_ERROR_HEADER = "x-amzn-ErrorType";
```
dhrubo-os, line 42: what is this error about? Can we add more context?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
ylwu-amzn: sounds good if onHeaders execute first

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Hailong-am: As `errorsInHeader` is coming from header key `x-amzn-ErrorType`, this should be amazon specific key, need @Zhangxunmt to double confirm.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Verified with multiple test with reasonable TPS concurrency, the onHeader() always executes before onStream(). So when the exception is caught at the headers, it will not be override in the onStream() when receiving the request body.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: Please check the PR description, it has a lot of context in the aforementioned github issues.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
Zhangxunmt: +1 with Hailong. Please check the PR description for more context. This is to handle the throttling exception from SageMaker. For other platforms, they likely use different status code (likely 429) or different key in the header or messages in the request body for throttling responses.

zhichao-aws (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
dhrubo-os: The method name `handleThrottlingInHeader` is too generic and doesn't convey the specific use case of handling throttling exceptions from Amazon SageMaker service. I suggest renaming the method to `handleSageMakerThrottlingException` or `handleAmazonServiceThrottlingException` to make it more clear and self-explanatory.

Additionally, it would be helpful to add a brief comment explaining the purpose of this method and why it was introduced. Something like:

```
/**
 * Handles throttling exceptions from Amazon SageMaker service by checking the 'x-amzn-ErrorType' header.
 * This method was introduced to handle remote server throttling errors, which were not being properly
 * propagated to the client. See [link to the relevant issue or PR] for more details.
 */
private void handleSageMakerThrottlingException(SdkHttpFullResponse sdkResponse) {
    // ...
}
```

Adding such comments will make it easier for future developers to understand the context and rationale behind this code, without having to dig through the issue tracker or commit history.

Zhangxunmt: This throttling issue could be applicable for other model servers too, so using an Amazon specific name will give hurdle to handle other cases. With this general name, it's easier to add more checkups for other non-amzn throttling keys in the header. The same for adding this comment on top of the method. 
I think we can provide the issue link with brief intro inside the function when obtaining the amazon header.

jngz-es (approved):

b4sjoo (approved):

PR #2450, opened 2024-05-16
adding independent node settings

dhrubo-os:
### Description
[adding `plugins.ml_commons.independent_node` to identify if that's a single tenant or multi-tenant use case.]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
@@ -18,6 +18,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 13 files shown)


plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -308,4 +315,22 @@ public static void wrapListenerToHandleSearchIndexNotFound(Exception e, ActionLi
...
+
+    public static String getTenantID(ClusterService clusterService, Settings settings, RestRequest restRequest) {
+        if (IsIndependentNode(clusterService, settings)) {
+            String tenantId = restRequest.getHeaders().get(Constants.TENANT_ID).get(0);
+            if (tenantId == null) {
+                throw new OpenSearchStatusException("Tenant ID can't be null", RestStatus.INTERNAL_SERVER_ERROR);
+            }
+            return tenantId;
+        } else {
+            return null;
```
rbhavna, line 332: What should be the flow when it is not an independent node and null is returned here?

dhrubo-os: for single tenant case, tenant id will be null and that will be stored in the index as null.

Issue #2428, opened 2024-05-09
[FEATURE] Split remote inference text list if its number exceeds user configured limitation


chishui:
**Is your feature request related to a problem?**
Most of ML services may set a limitation on maximum length of text in a batch request. For example, Cohere explicitly sets 96 for it's embedding API ([ref](https://docs.cohere.com/reference/embed)), OpenAI has a limitation of 2048 ([ref](https://community.openai.com/t/embeddings-api-max-batch-size/655329)). Exceeding the limitation could cause a failed request.

With [chunking processor feature](https://github.com/opensearch-project/neural-search/issues/548) and [batch ingestion feature](https://github.com/opensearch-project/neural-search/issues/743), we are more likely to run into the issue that the number of texts in a single batch request could exceed server's limitation. 

So we'd like to have a solution to cut texts into small batches if total number of them in a batch request exceeds the limitation.

**What solution would you like?**
#### 1. Support a new parameter in connector to set the maximum batch size limitation from remote server.
```json
POST /_plugins/_ml/connectors/_create
{
  ...
  "parameters": {
     "max_batch_size": 96
  },
  ...
}
```
If `max_batch_size` is not set, we don't cut texts into sub batches.
#### 2. Cut texts into small batches
if `max_batch_size` is set and total number of texts exceed the number it sets, we use `max_batch_size` to chunk texts. There will be `total number of texts / max_batch_size` batches and each batch has text number no greater than `max_batch_size`.

#### 3. Sort texts based on text length before cutting
As @model-collapse explained [here](https://github.com/opensearch-project/OpenSearch/issues/12457#issuecomment-1980472336), LLM has better performance if the batched texts have similar text length. We also did a benchmark to test the performance of sorting texts before cutting them into small batches and sending batch requests to a SageMaker model. We found sorting could improve ingesting latency by 5.5%.

- SageMaker host type: g5.xlarge
- Processor: Sparse Encoding
- Benchmark Setup
  - Bulk size: 160 
  - client: 1
  - batch size: 16

|                    Metrics | no sort | sort before making batches |
| -------------------------: | -------: | --------------------: |
|    Min Throughput (docs/s) |    391.85 |                 205.16 |
|   Mean Throughput (docs/s) |    445.6 |                466.79 |
| Median Throughput (docs/s) |   445.02 |                470.9 |
|    Max Throughput (docs/s) |    494.83 |                527.66 |
|           Latency P50 (ms) |  373.039 |               357.348 |
[...]

PR #2455, opened 2024-05-17
[Feature] Split remote inference text list if its number exceeds user configured limitation

chishui:
### Description
We'd like to have a solution to cut texts into small batches if total number of them in a batch request exceeds the limitation
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2428
 
### Check List
  - [x] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
@@ -20,6 +20,7 @@ import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.util.Strings;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.exception.MLException;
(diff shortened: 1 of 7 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -137,10 +138,25 @@ private void processResponse(
...
+
+        // step 2: restore to original order as textDocs might be sorted
+        List<Integer> originalOrderIndexes = executionContext.getOriginalOrder();
+        List<ModelTensors> modelTensorsList = Arrays.asList(modelTensors);
+        if (CollectionUtils.isEmpty(originalOrderIndexes)) {
+            actionListener.onResponse(modelTensorsList);
+        } else {
+            // if the originalOrder is not empty, reorder based on it
+            List<ModelTensors> sortedModelTensors = Arrays.asList(new ModelTensors[modelTensorsList.size()]);
+            assert (originalOrderIndexes.size() == modelTensors.length);
```
xinyual, line 154: The `originalOrderIndexes`  size is the doc number, while `modelTensors` size is request size. So it would cause some error here. A simple way to reproduce is to try predict a four size input and set connector max batch size to 2. We may need to fetch all tensor and re-create ModelTensors Again.

One more small detail here: What is the list of `modelTensors` size here we want to keep? If we keep the size equal to request number, it may confuse user because actually sort and modelTensors is not the actual remote response (we re-sort it). So maybe keep all tensor into one modelTensors can be a choice.

chishui: Thanks for catching this, it is a bug, will fix in next revision

chishui:
Will implement the logic from neural search side. Please refer to https://github.com/opensearch-project/ml-commons/issues/2428#issuecomment-2132554851

Closing now

PR #2458, opened 2024-05-17
fix error message with unwrapping the root cause

### LINKED ISSUE #2443: [BUG] agent execution error message
**What is the bug?**
Agent execution api will return the following error message if the root cause exception doesn't occur in coordinate node and is wrapped into remote transport exception.
```
"{\"error\":{\"reason\":\"Invalid Request\",\"details\":\"[d0aa9b912dfc3dacaf11334ad978ea34][x.x.x.x:x][cluster:admin/opensearch/ml/execute]\",\"type\":\"RemoteTransportException\"},\"status\":400}"
```

jngz-es:
### Description
Unwrap the root cause to return when agent execution gets an exception.
 
### Issues Resolved
resolve #2443
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
@@ -59,4 +59,8 @@ public class MLExceptionUtils {
             log.error(errorMessage, e);
         }
     }
+
+    public static Throwable getRootCause(Throwable t) {
+        return ExceptionUtils.getRootCause(t);
+    }
 }
--- a/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessageFactory.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessageFactory.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.utils.error;
 
 import org.opensearch.OpenSearchException;
+import org.opensearch.ml.utils.MLExceptionUtils;
 
 import lombok.experimental.UtilityClass;
 
@@ -23,22 +24,9 @@ public class ErrorMessageFactory {
         int st = status;
         if (t instanceof OpenSearchException) {
             st = ((OpenSearchException) t).status().getStatus();
-        } else {
-            t = unwrapCause(e);
         }
+        t = MLExceptionUtils.getRootCause(t);
 
         return new ErrorMessage(t, st);
-    }
-
-    protected static Throwable unwrapCause(Throwable t) {
-        Throwable result = t;
-        if (result instanceof OpenSearchException) {
-            return result;
-        }
-        if (result.getCause() == null) {
-            return result;
-        }
-        result = unwrapCause(result.getCause());
-        return result;
     }
 }
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestMLExecuteActionTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestMLExecuteActionTests.java
@@ -30,6 +30,7 @@ import org.opensearch.client.node.NodeClient;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.Strings;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.input.Input;
 import org.opensearch.ml.common.transport.execute.MLExecuteTaskAction;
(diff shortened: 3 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
```diff
@@ -59,4 +59,8 @@ public static void logException(String errorMessage, Exception e, Logger log) {
             log.error(errorMessage, e);
         }
     }
+
+    public static Throwable getRootCause(Throwable t) {
```
ylwu-amzn, line 63: seems not necessary to create a new method for one line of code

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessageFactory.java
```diff
@@ -23,22 +24,9 @@ public static ErrorMessage createErrorMessage(Throwable e, int status) {
         int st = status;
         if (t instanceof OpenSearchException) {
             st = ((OpenSearchException) t).status().getStatus();
-        } else {
-            t = unwrapCause(e);
         }
+        t = MLExceptionUtils.getRootCause(t);
```
Hailong-am, line 28: would recommend to use `ExceptionsHelper.unwrapCause()` which provided by core library

https://github.com/opensearch-project/OpenSearch/blob/3fe66747b1523c74e06bc3465ef2ab682048f76b/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java#L127-L145

zhichao-aws: +1

jngz-es: MLExceptionUtils.getRootCause(t) wraps the above function. What I was thinking is the existing class MLExceptionUtils is responsible for hosting all exception utilities in ml-commons, so I just simply wrap the above core function into this class to use for the entire ml-commons plugin.

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
jngz-es: See my below comment please.

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessageFactory.java
jngz-es: Also it can provide flexibility for ml-commons plugin. If we needed to change the logic, we only needed to change the function MLExceptionUtils.getRootCause(t), not everywhere the core function used.

Hailong-am: > MLExceptionUtils.getRootCause(t) wraps the above function 

I think the logic between `ExceptionUtils.getRootCause(t)` and `ExceptionsHelper.unwrapCause()` is different. `getRootCause` is to get the root cause but `unwrapCause` used for unwrap `OpenSearchWrapperException` and not go deeper. They may produce different result.

jngz-es: Hi Hailong, I went through those two implementations and didn't see big differences. The only difference is about `ExceptionsHelper.unwrapCause()` having 10 times limitation on the extracting loop.

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/utils/error/ErrorMessageFactory.java
Hailong-am: > The only difference is about ExceptionsHelper.unwrapCause() having 10 times limitation on the extracting loop.

I guess the difference is that ExceptionsHelper.unwrapCause() will stop if the cause is not instance of `OpenSearchWrapperException`. I don't know if there has a case the cause is not `OpenSearchWrapperException` and still have inner cause.

jngz-es: `OpenSearchWrapperException` is reusing the standard `throwable` `getCause` interface to unwrap. I prefer to use the method from the library `org.apache.commons.lang3.exception` which covers the method `ExceptionsHelper.unwrapCause()`.

Zhangxunmt (approved):

yuye-aws:
@jngz-es Shall we also backport this PR to 2.14 and 2.15 branch? The multi node IT is currently excluded for search index tool.

PR #2459, opened 2024-05-17
[Feature/multi_tenancy] SdkClient interface with local and remote implementations for Connector

dbwiddis:
### Description
 - Creates a new `SdkClient` Interface with Put, Get, and Delete operations (more to come later)
 - Adds implementations for both Local (NodeClient) and Remote (OpenSearch Java Client) clusters
 - Uses the destination-agnostic interface for Connector Put (Index), Get, and Delete actions
 
### Issues Resolved

Part of implementation in support of RFC https://github.com/opensearch-project/OpenSearch/issues/13336

### Previous Comments

PR #2430 was auto-closed due to a branch deletion, some comments are there.
 
### Guided Review
1. Start with the `SdkClient` interface and the request/response classes in the `org.opensearch.sdk` package.  These are eventually intended to be moved outside of this plugin to a common repository (probably in the OpenSearch repo as a new module) but remain here during initial development to ease development iteration.
2. The changes to the Plugin initialization (inject the local client and a client module) are probably going to stay in somewhat a similar state, but the SdkClientModule is currently temporary (creating an http client without TLS/SSL) pending a more properly thought out long term pluggable/injectable solution.
3. Glance at the `org.opensearch.ml.action.connector` changes to get a quick idea of how the new interface calls the clients.  I have not yet done detailed exception handling here. That's a TODO below.
4. Look at the LocalClusterIndicesClient which essentially wraps current NodeClient implementation.
5. Look at the RemoteClusterIndicesClient which tries to make using a remote cluster via the Java Client transparent to the user.
6. Look at everything else.

### How to manually test locally

Checkout this PR and create the zip file for it
```
gh pr checkout 2459
./gradlew assemble
```

Publish the latest OpenSearch 2.x branch snapshot to local (change `darwin-tar` to your own OS if needed)
```
cd ~/git/OpenSearch 
git fetch upstream
git checkout 2.x
git pull upstream 2.x
./gradlew clean localDistro
cd distribution/archives/darwin-tar/build/install/opensearch-2.15.0-SNAPSHOT
```

Install the plugin (change the path as necessary for your own local environment)
```
./bin/opensearch-plugin install file:///Users/widdisd/git/ml-commons/plugin/build/distributions/opensearch-ml-2.15.0.0-SNAPSHOT.zip
```
Start OpenSearch and test the local implementation using your favorite REST client (curl, Postman, etc.)
```
bin/opensearch
```

Ctrl-C to close.   Switch to using the remote client by making this change to your environment:
```
export REMOTE_METADATA_ENDPOINT=http://127.0.0.1:9200
```

Restart OpenSearch and test the remote implementation using your favorite REST client (curl, Postman, etc.).  The Connector Put, Get, and Delete will use the remote client even though they're still operating on the same cluster; other APIs will still work locally.
```
bin/opensearch
```

### TODO Items 
Will continue to work on these in a new PR if this one needs to be merged to unblock other work
 - [x] Fix a bug where shard stats return null on deletion
 - [x] Update currently commented-out Connector Unit tests with proper mocks/tests of new functionality
 - [x] Re-enable code coverage checks
 - [x] Implement ability to specify a thread pool executor as part of the interface
 - [ ] Properly handle all exceptions
 - [ ] Extend functionality to Update and Search

### Check List
  - [x] All tests pass


dhrubo-os:
```
Build file 'D:\a\ml-commons\ml-commons\plugin\build.gradle' line: 43

* What went wrong:
Could not compile build file 'D:\a\ml-commons\ml-commons\plugin\build.gradle'.
> startup failed:
  build file 'D:\a\ml-commons\ml-commons\plugin\build.gradle': 43: Unexpected input: '{' @ line 43, column 14.
     dependencies {
                  ^

**Comment by dbwiddis** [comment]
> Unexpected input: '{'

The problems with doing manual merge and missing little things.  I also ran spotless but failed to actually commit the changed files. This may be correlated with the fact that it's Friday after 5pm. 😁 🍺

File: plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java [hunk 1]
```diff
@@ -53,11 +56,13 @@ public DeleteConnectorTransportAction(
         TransportService transportService,
         ActionFilters actionFilters,
         Client client,
```
ylwu-amzn, line 58: Should we remove this ?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
```diff
@@ -0,0 +1,141 @@
...
+package org.opensearch.sdk;
+
+import org.opensearch.action.support.replication.ReplicationResponse.ShardInfo;
+import org.opensearch.core.common.Strings;
+import org.opensearch.core.index.shard.ShardId;
+
+public class DeleteDataObjectResponse {
+    private final String id;
+    private final ShardId shardId;
+    private final ShardInfo shardInfo;
```
arjunkumargiri, line 18: Why are these fields required? Shard details are specific to opensearch which will make it difficult to extend the interface to non-opensearch data stores.

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
arjunkumargiri, line 58: Nit: client is no longer used and can be removed. Same comment for other transport action classes.

plugin/src/main/java/org/opensearch/ml/action/connector/TransportCreateConnectorAction.java
arjunkumargiri, line 1: Line 127 performs mlIndicesHandler.initMLConnectorIndex(), which would create connector index if missing. mlIndicesHandler class needs to be updated to make use of SDKClient.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -673,7 +682,8 @@ public Collection<Object> createComponents(
                 clusterManagerEventListener,
                 mlCircuitBreakerService,
                 mlModelAutoRedeployer,
-                cmHandler
+                cmHandler,
+                localClusterIndicesClient
```
arjunkumargiri, line 686: Instead of hard coding to local indices client, need to make use of SDKClient interface to dynamically load client based SDK client initialized in SDKClientModule.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -0,0 +1,84 @@
...
+    SdkClientModule(String remoteMetadataEndpoint, String region) {
+        this.remoteMetadataEndpoint = remoteMetadataEndpoint;
+        this.region = region;
+    }
+
+    @Override
+    protected void configure() {
+        boolean local = Strings.isNullOrEmpty(remoteMetadataEndpoint);
+        if (local) {
+            bind(SdkClient.class).to(LocalClusterIndicesClient.class);
```
arjunkumargiri, line 57: Should bind to instance of class instead of class?

plugin/src/main/java/org/opensearch/ml/sdkclient/LocalClusterIndicesClient.java
```diff
@@ -0,0 +1,127 @@
...
+                log.info("Getting {} from {}", request.id(), request.index());
+                GetResponse getResponse = client.get(new GetRequest(request.index(), request.id())).actionGet();
+                if (getResponse == null || !getResponse.isExists()) {
+                    return new GetDataObjectResponse.Builder().id(request.id()).build();
+                }
+                XContentParser parser = jsonXContent
+                    .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, getResponse.getSourceAsString());
+                log.info("Retrieved data object");
+                return new GetDataObjectResponse.Builder().id(getResponse.getId()).parser(Optional.of(parser)).build();
+            } catch (OpenSearchStatusException | IndexNotFoundException notFound) {
```
dhrubo-os, line 101: So for `RestStatus.UNAUTHORIZED`, `RestStatus.FORBIDDEN`, `RestStatus.CONFLICT`, `RestStatus.INTERNAL_SERVER_ERROR`, `RestStatus.BAD_REQUEST`, we will throw `notFound`? This doesn't seem right to me.

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
```diff
@@ -0,0 +1,124 @@
...
+                log.info("Deleting {} from {}", request.id(), request.index());
+                DeleteResponse deleteResponse = openSearchClient.delete(deleteRequest);
+                log.info("Deletion status for id {}: {}", deleteResponse.id(), deleteResponse.result());
+                ShardInfo shardInfo = new ShardInfo(
+                    deleteResponse.shards().total().intValue(),
+                    deleteResponse.shards().successful().intValue()
+                );
+                return new DeleteDataObjectResponse.Builder()
+                    .id(deleteResponse.id())
+                    .shardId(deleteResponse.index())
```
dhrubo-os, line 115: [nit] if we are just trying to put some random value for other remote stores, let's a comment.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
dhrubo-os: +1

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -0,0 +1,84 @@
...
+                .setStrictDeprecationMode(true)
+                .setHttpClientConfigCallback(httpClientBuilder -> {
+                    try {
+                        return httpClientBuilder.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE);
+                    } catch (Exception e) {
+                        throw new OpenSearchException(e);
+                    }
+                })
+                .build();
+            ObjectMapper objectMapper = new ObjectMapper().setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE);
```
dhrubo-os, line 78: can this introduce any insecure de-serialization?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
dbwiddis: I previously answered that question when you asked it [here](https://github.com/opensearch-project/ml-commons/pull/2430#discussion_r1603931384).

The return value to the end user is created from a `DeleteResponse` object, which _requires_ a non-null ShardId:
```java
this.shardId = Objects.requireNonNull(shardId);
```
So I have to have a Shard ID. I can create one, or I can pass along the actual one that is returned from cluster-based OpenSearch.   Which do you prefer?   I see no good reason to delete information that exists.  It can be ignored for other implementations.

Similarly, for `ShardInfo`, the return value from the Delete Connector API looks like this:
```json
{
    "_id": "qU00aowBH9rhI2IwnrH3",
    "_version": 2,
    "result": "deleted",
    "_shards": {
        "total": 1,
        "successful": 1,
        "failed": 0
    },
    "_seq_no": 2,
    "_primary_term": 1
}
```

I can either pass along those actual statistics or make them up.  Which is easier to explain to a customer, that a statistic might be irrelevant for their storage option, or that we're taking away information they used to have?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
```diff
@@ -0,0 +1,141 @@
...
+            return this;
+        }
+
+        /**
+         * Adds a generated shard id to this builder
+         * @param indexName the index name to generate a shard id
+         * @return the updated builder
+         */
+        public Builder shardId(String indexName) {
+            this.shardId = new ShardId(indexName, Strings.UNKNOWN_UUID_VALUE, 0);
```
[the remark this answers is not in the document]
dbwiddis: Answered above and also [here](https://github.com/opensearch-project/ml-commons/pull/2430#discussion_r1603993547), but I"ll repeat:

The `DeleteResponse` object requires a non-null ShardId.
```java
this.shardId = Objects.requireNonNull(shardId);
```
When there isn't one provided (such as DynamoDb) this code creates a simulated one.

common/src/main/java/org/opensearch/sdk/GetDataObjectResponse.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.core.xcontent.XContentParser;
+
+import java.util.Optional;
+
+public class GetDataObjectResponse {
+    private final String id;
+    private final Optional<XContentParser> parser;
```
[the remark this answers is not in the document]
dbwiddis: You still get a `GetResponse` even if you didn't find an object.  The (Cluster) `GetResponse` has a field `isExists()` that checks whether there is actually a result present.    The (Java Client) `GetResponse` has a similar field `found()` which serves the same purpose.

I can either have a `null` parser and another boolean field that you have to check, or I can have an `Optional`.  I dislike passing around nulls.   So you check if the optional `isPresent()` and then you get it to parse.

The Optional not being present means the document/item/data was not found.

common/src/main/java/org/opensearch/sdk/SdkClient.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    default GetDataObjectResponse getDataObject(GetDataObjectRequest request) {
+        try {
+            return getDataObjectAsync(request).toCompletableFuture().join();
+        } catch (CompletionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            throw new OpenSearchException(cause);
```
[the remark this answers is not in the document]
dbwiddis: I haven't yet aligned exception handling. That's on my to do list. 

Please see the PR description where I mention this, and a few TODOs in the test classes.

I plan on not changing the user experience.  I put in placeholders during my initial testing.

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
dbwiddis: No, because we still need it to access other key things like Thread Pools, Thread Context (and User), etc. as well as potentially any `client.execute()` calls to other Transport Actions.   The `SdkClient` is only abstracting out the data access (CRUD-S) features of the `NodeClient`.

dhrubo-os:
Merging this PR in the branch so that I can continue my development on top of this PR. @dbwiddis please address the changes (if needed) in separate PR. Thanks.

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
dbwiddis: It very much is still used.
```java
```
```java
client.threadPool().executor(GENERAL_THREAD_POOL)
```

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
```diff
@@ -120,26 +125,24 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
-                        return;
+            sdkClient
+                .deleteDataObjectAsync(
+                    new DeleteDataObjectRequest.Builder().index(deleteRequest.index()).id(deleteRequest.id()).build(),
+                    client.threadPool().executor(GENERAL_THREAD_POOL)
+                )
+                .whenComplete((r, throwable) -> {
+                    context.restore();
+                    if (throwable != null) {
+                        actionListener.onFailure(new RuntimeException(throwable));
```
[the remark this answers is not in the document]
dbwiddis: See above that I haven't fully hashed out the exception handling.  I used placeholders until I got things working.  I left a rather detailed initial comment on this PR.

plugin/src/main/java/org/opensearch/ml/action/connector/GetConnectorTransportAction.java
```diff
@@ -66,50 +66,60 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLConn
...
-                        if (connectorAccessControlHelper.hasPermission(user, mlConnector)) {
-                            actionListener.onResponse(MLConnectorGetResponse.builder().mlConnector(mlConnector).build());
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    context.restore();
+                    log.debug("Completed Get Connector Request, id:{}", connectorId);
+                    if (throwable != null) {
+                        Throwable cause = throwable.getCause() == null ? throwable : throwable.getCause();
+                        if (cause instanceof IndexNotFoundException) {
```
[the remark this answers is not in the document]
dbwiddis: (Repeat comment on planning on aligning all exception handling with current user experience.)

plugin/src/main/java/org/opensearch/ml/action/connector/GetConnectorTransportAction.java
```diff
@@ -66,50 +66,60 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLConn
...
+                                log.error("Failed to parse ml connector" + r.id(), e);
+                                actionListener.onFailure(e);
+                            }
                         } else {
                             actionListener
                                 .onFailure(
                                     new OpenSearchStatusException(
-                                        "You don't have permission to access this connector",
-                                        RestStatus.FORBIDDEN
+                                        "Failed to find connector with the provided connector id: " + connectorId,
```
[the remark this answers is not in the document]
dbwiddis: (Repeat comment on planning on aligning all exception handling with current user experience.)

plugin/src/main/java/org/opensearch/ml/action/connector/TransportCreateConnectorAction.java
dbwiddis: I have not yet implemented any index creation/deletion code in the SDKClient.

This has to be developed incrementally. We're already at over 2000 lines of code being added.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
dbwiddis: Please see my very lenghty initial comment on this PR.  I discussed this.
> 1. ... These are eventually intended to be moved outside of this plugin to a common repository (probably in the OpenSearch repo as a new module) but remain here during initial development to ease development iteration.
> 2. The changes to the Plugin initialization (inject the local client and a client module) are probably going to stay in somewhat a similar state, but the SdkClientModule is currently temporary (creating an http client without TLS/SSL) pending a more properly thought out long term pluggable/injectable solution.

plugin/src/main/java/org/opensearch/ml/sdkclient/LocalClusterIndicesClient.java
dbwiddis: (Repeat comment on planning on aligning all exception handling with current user experience.)

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
dbwiddis: There is a comment in the `DeleteDataObjectResponse` javadocs.  Where else should this be commented?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
dbwiddis: No, because the injector will just see `SdkClient` in the transport action, so that has to be the key.  They won't know which instance it's bound to. That's the whole purpose of this module.  We had this discussion [here](https://github.com/opensearch-project/ml-commons/pull/2430#discussion_r1597261968) and this module was in response to your suggestion.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -0,0 +1,84 @@
...
+    private OpenSearchClient createOpenSearchClient() {
+        try {
+            // Basic http(not-s) client using RestClient.
+            RestClient restClient = RestClient
+                // This HttpHost syntax works with export REMOTE_METADATA_ENDPOINT=http://127.0.0.1:9200
+                .builder(HttpHost.create(remoteMetadataEndpoint))
+                .setStrictDeprecationMode(true)
+                .setHttpClientConfigCallback(httpClientBuilder -> {
+                    try {
+                        return httpClientBuilder.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE);
```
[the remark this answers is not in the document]
dbwiddis: Yes!  This is temporary as I explained in my original PR comment, bullet point 2.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
dbwiddis: It's what we use in the Java Client. What are your particular security concerns?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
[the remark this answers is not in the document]
dhrubo-os: Yes, I saw your answer earlier too. I think for DDB / other non OS cases, if we just share a random shardID, that could be confusing for customers. I would prefer sharing any static String, maybe like Not Applicable or N/A, which can convey the actual meaning rather than causing unnecessary confusion.

plugin/src/test/java/org/opensearch/ml/action/connector/GetConnectorTransportActionTests.java
```diff
@@ -87,82 +120,103 @@ public void setup() throws IOException {
...
 
         GetResponse getResponse = prepareConnector();
-        doAnswer(invocation -> {
-            ActionListener<GetResponse> listener = invocation.getArgument(1);
-            listener.onResponse(getResponse);
-            return null;
-        }).when(client).get(any(), any());
+        PlainActionFuture<GetResponse> future = PlainActionFuture.newFuture();
+        future.onResponse(getResponse);
+        when(client.get(any(GetRequest.class))).thenReturn(future);
```
dhrubo-os, line 141: Why `client.get`? Shouldn't we use `sdkClient`? Are you sure these tests are reflecting the actual behavior we have in the class?

plugin/src/test/java/org/opensearch/ml/action/connector/GetConnectorTransportActionTests.java
```diff
@@ -87,82 +120,103 @@ public void setup() throws IOException {
...
         }).when(connectorAccessControlHelper).validateConnectorAccess(any(), any(), any());
 
         GetResponse getResponse = prepareConnector();
-        doAnswer(invocation -> {
-            ActionListener<GetResponse> listener = invocation.getArgument(1);
-            listener.onResponse(getResponse);
-            return null;
-        }).when(client).get(any(), any());
+        PlainActionFuture<GetResponse> future = PlainActionFuture.newFuture();
+        future.onResponse(getResponse);
```
dhrubo-os, line 162: Shouldn't this be `CompletionStage<GetDataObjectResponse>`?

plugin/src/test/java/org/opensearch/ml/action/connector/GetConnectorTransportActionTests.java
dhrubo-os: Thanks for the explanation.

common/src/main/java/org/opensearch/sdk/GetDataObjectRequest.java
```diff
@@ -0,0 +1,108 @@
...
+            return this;
+        }
+
+        /**
+         * Add a fetchSourceContext to this builder
+         * @param fetchSourceContext the fetchSourceContext
+         * @return the updated builder
+         */
+        public Builder fetchSourceContext(FetchSourceContext fetchSourceContext) {
+            this.fetchSourceContext = fetchSourceContext;
```
dhrubo-os, line 96: what happens if the `fetchSourceContext` is null?

common/src/main/java/org/opensearch/sdk/GetDataObjectResponse.java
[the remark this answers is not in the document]
dhrubo-os: >The Optional not being present means the document/item/data was not found.

Should we add a comment regarding this in the code?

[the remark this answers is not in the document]
dhrubo-os: Sometime before parsing, we need to get information from the source like this: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/GetModelTransportAction.java#L94-L96

How can we address this?

common/src/main/java/org/opensearch/sdk/DeleteDataObjectResponse.java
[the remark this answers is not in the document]
dbwiddis: > I would prefer sharing any static String, maybe like `Not Applicable` or `N/A`, which can convey the actual meaning rather than causing unnecessary confusion.

The `Strings.UNKNOWN_UUID_VALUE` is `_na_`.

common/src/main/java/org/opensearch/sdk/GetDataObjectRequest.java
dbwiddis: That's the default value for the `GetRequest` if it's not provided.  Although I did notice that I didn't pass this on to the GetRequest so I'll fix that.

common/src/main/java/org/opensearch/sdk/GetDataObjectResponse.java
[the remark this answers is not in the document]
dbwiddis: Addressed in #2489.  Please comment there if you think the javadoc needs more amplification on the fact that a non-present optional and empty source map mean that the Get call returned that the document didn't exist in the index.

PR #2462, opened 2024-05-20
Implementing retry for remote connector to mitigate throttling issue

### LINKED ISSUE #2438: [RFC] Add retry with backoff for SageMaker throttling exception to mitigate the data lost problem

zhichao-aws:
### Description
Please check #2438 for more details and context
What does this PR change:
1. Use core class RetryableAction to implement a retry policy for the RemoteConnectorExecutor.invokeRemoteModel method to retry for SageMaker throttling exception.
  a. New class RetryableException is added. And we can extend it once we found other retryable cases.
2. For sub-requests cases, we need to retry for failed sub-request, instead of all sub-requests. So we refactor the RemoteConnectorExecutor.executePredict and MLSdkAsyncHttpResponseHandler, to use GroupedActionListener instead of CoundDownLatch for sub-requests case. 
With previous code we call onResponse/onFailure only after all sub-requests have response. With the refactor, we call onResponse/onFailure for all sub-requests, and the onFailure will further trigger the retry logic for sub-request in RetryableAction. 
3. Add ConnectorRetryOption to control the behavior of retry. Per @zane-neo and @ylwu-amzn 's suggestion, we'll move the retry settings from cluster settings to ConnectorClientConfig.
 
Note: the retry is disabled by default to make the behavior consistent with previous version.
 
### Issues Resolved
#2438
 
### Check List
  - [ ] All tests pass


zhichao-aws:
This PR is ready for review

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -197,27 +136,46 @@ public void onError(Throwable t) {
...
+
+        if (statusCode < HttpStatus.SC_OK || statusCode > HttpStatus.SC_MULTIPLE_CHOICES) {
+            if (containsThrottlingException) {
+                actionListener
+                    .onFailure(
+                        new RetryableException(
+                            REMOTE_SERVICE_ERROR + "The request was denied due to remote server throttling.",
+                            RestStatus.fromCode(statusCode)
+                        )
+                    );
```
Zhangxunmt, line 164: Is there any other benefits of removing exceptionHolder? I feel like this is adding a lot of manual checkups for different exceptions. For handling more other exceptions, this would needs to add more code and flags? I think this is less flexible than using exceptionHolder, even we don't need to countDown any more in the retry.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -117,6 +117,27 @@ private MLCommonsSettings() {}
...
         .boolSetting("plugins.ml_commons.remote_inference.enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
 
+    public static final Setting<Boolean> ML_COMMONS_REMOTE_INFERENCE_RETRY_ENABLED = Setting
+        .boolSetting("plugins.ml_commons.remote_inference.retry_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
+    public static final Setting<Integer> ML_COMMONS_REMOTE_INFERENCE_RETRY_BACKOFF_MILLIS = Setting
+        .intSetting(
+            "plugins.ml_commons.remote_inference.retry_backoff_millis",
+            100,
+            1,
```
Zhangxunmt, line 127: Why skipping the max value? I think it makes sense to put a max allowed value here. Same for ML_COMMONS_REMOTE_INFERENCE_RETRY_TIMEOUT_SECONDS

plugin/src/main/java/org/opensearch/ml/utils/RemoteInferenceRetryUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_REMOTE_INFERENCE_RETRY_BACKOFF_MILLIS, it -> retryBackoffMillis = it);
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(ML_COMMONS_REMOTE_INFERENCE_RETRY_TIMEOUT_SECONDS, it -> retryTimeoutSeconds = it);
+    }
+
+    public static ConnectorRetryOption getRetryOptionFromClusterSettings() {
+        ConnectorRetryOption connectorRetryOption = new ConnectorRetryOption();
```
Zhangxunmt, line 36: You have the all arguments constructor, so using that constructor can do a single line instantiation.

Zhangxunmt (commented):
Did you verify to what extend this retry can mitigate the throttling? For heavy throttling that happens for a long time window that is longer than the time_out, I assume it would still likely report throttling errors after retries?

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -117,6 +117,27 @@ private MLCommonsSettings() {}
...
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
+        );
+
+    public static final Setting<Integer> ML_COMMONS_REMOTE_INFERENCE_RETRY_TIMEOUT_SECONDS = Setting
+        .intSetting(
+            "plugins.ml_commons.remote_inference.retry_timeout_seconds",
+            30,
```
Zhangxunmt, line 135: Was there any end-to-end tests using the _Bulk API to verify the impact to performance using the 30s retry? For example, having a sageMaker endpoint with t2.medium instance type and invoking it with 200 docs with 3 nodes, etc.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +199,61 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+        final RetryableAction<Tuple<Integer, ModelTensors>> invokeRemoteModelAction = new RetryableAction<>(
+            getLogger(),
+            getClient().threadPool(),
+            TimeValue.timeValueMillis(executionContext.getConnectorRetryOption().getRetryBackoffMillis()),
+            TimeValue.timeValueSeconds(executionContext.getConnectorRetryOption().getRetryTimeoutSeconds()),
+            actionListener,
+            BackoffPolicy
+                .constantBackoff(
+                    TimeValue.timeValueMillis(executionContext.getConnectorRetryOption().getRetryBackoffMillis()),
+                    Integer.MAX_VALUE
```
Zhangxunmt, line 226: I think we should have a max number of retries to not always time_out in edge cases. Imagine for a fully throttled service and we need to time out every request causing strange performance behaviors.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -918,6 +921,9 @@ public List<Setting<?>> getSettings() {
                 MLCommonsSettings.ML_COMMONS_REMOTE_MODEL_ELIGIBLE_NODE_ROLES,
                 MLCommonsSettings.ML_COMMONS_LOCAL_MODEL_ELIGIBLE_NODE_ROLES,
                 MLCommonsSettings.ML_COMMONS_REMOTE_INFERENCE_ENABLED,
+                MLCommonsSettings.ML_COMMONS_REMOTE_INFERENCE_RETRY_ENABLED,
```
b4sjoo, line 924: Why we need to add a setting flag for this feature?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -197,27 +136,46 @@ public void onError(Throwable t) {
...
+    private void response() {
+        String body = responseBody.toString();
+        if (Strings.isBlank(body)) {
+            log.error("Remote model response body is empty!");
+            actionListener.onFailure(new OpenSearchStatusException("No response from model", RestStatus.BAD_REQUEST));
+            return;
+        }
+
+        if (statusCode < HttpStatus.SC_OK || statusCode > HttpStatus.SC_MULTIPLE_CHOICES) {
+            if (containsThrottlingException) {
```
xinyual, line 157: Do we really need to make retry action specific for throttling? Can we make it general for all exceptions?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: Good point, ack

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zhichao-aws: ack, will set to 5s and 60s

plugin/src/main/java/org/opensearch/ml/utils/RemoteInferenceRetryUtils.java
zhichao-aws: I think the current implementation is more robust to future extension. When we add more parameter to retry option in the future, it won't fail the compile at all places where we create an object

zhichao-aws:
> Did you verify to what extend this retry can mitigate the throttling? For heavy throttling that happens for a long time window that is longer than the time_out, I assume it would still likely report throttling errors after retries?

Please check this [benchmark result](https://github.com/opensearch-project/ml-commons/issues/2438#issuecomment-2117001088) and the context description in the RFC issue. 
If there is heavy throttling that happens for a long time window that is longer than the time_out, 
1. we will get many `Acquire operation took longer than the configured maximum time` errors because some requests are queued longer than 10s at the async client.
2. For this case, my understanding is the actual workload exceeds the capacity of the endpoint. And users should upgrade to a better instance instead of just retry. Otherwise new requests will be queued, and the response time will get very large and harms the user experience

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zhichao-aws: I tried with g4dn.xlarge + 3 nodes and multiple bulk settings, including 12 clients with size 10 and 1 clients with size 2000. ([ref](https://github.com/opensearch-project/ml-commons/issues/2438#issuecomment-2117001088))
If the workload within the capacity of the endpoint, i.e. the endpoint can process all requests within the connectionAcquisitionTimeout, I didn't see any throttling exception in the benchmark

zhichao-aws: And I think for `t2.medium` it works the same, the only difference is the endpoint throughput capacity

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
zhichao-aws: I want to follow the best practice for contributing in ml-commons proposed by Yaliang
> For new features, suggest add flag setting to control enable/disable it. 

From my point, one of the use case for disable is to adjust the `max_connections` of the connectors

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: Not all exceptions should be retried, e.g. GPU OOM exception, malformed parameters etc. Retry for these exceptions will only increase the load of remote endpoint, but won't get it run successfully.
For now we only find the throttling exception is suitable for retry. In the future we can extend it to more cases based on our new findings

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: For the cases you mentioned this settings make sense. But it is hard to set a value proper for all cases. The value is infected by input doc length, remote endpoint type, retry backoff value, and even the workload amount from other services if they share the same SageMaker endpoint with OpenSearch.
I think we can implement if as @xinyual suggest: https://github.com/opensearch-project/ml-commons/pull/2462#discussion_r1609139231
We add this setting, but disable this setting by default

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +199,61 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+            public void tryAction(ActionListener<Tuple<Integer, ModelTensors>> listener) {
+                // the listener here is RetryingListener
+                // If the request success, or can not retry, will call delegate listener
+                invokeRemoteModel(mlInput, parameters, payload, executionContext, listener);
+            }
+
+            @Override
+            public boolean shouldRetry(Exception e) {
+                final Throwable cause = ExceptionsHelper.unwrapCause(e);
+                boolean shouldRetry = cause instanceof RetryableException;
```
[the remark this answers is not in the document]
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorRetryOption.java
```diff
@@ -0,0 +1,23 @@
...
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class ConnectorRetryOption {
+    private boolean retryEnabled = false;
+    private Integer retryBackoffMillis;
+    private Integer retryTimeoutSeconds;
+    private Integer maxRetryTimes;
+    private String retyExecutor;
```
b4sjoo, line 22: Typo here.

zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -69,15 +70,19 @@ public AwsConnectorExecutor(Connector connector) {
         this.httpClient = MLHttpClientFactory.getAsyncHttpClient(connectionTimeout, readTimeout, maxConnection);
     }
 
+    @Override
+    public Logger getLogger() {
```
zane-neo, line 74: Why do we need to expose this log?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +199,65 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+                invokeRemoteModel(mlInput, parameters, payload, executionContext, listener);
+            }
+
+            @Override
+            public boolean shouldRetry(Exception e) {
+                Throwable cause = ExceptionsHelper.unwrapCause(e);
+                Integer maxRetryTimes = executionContext.getConnectorRetryOption().getMaxRetryTimes();
+                boolean shouldRetry = cause instanceof RetryableException;
+                if (shouldRetry) {
+                    getLogger().debug(String.format(Locale.ROOT, "The %d-th retry for invoke remote model", ++retryTimes));
```
zane-neo, line 245: Logging here is misleading, consider the first time: the log says `The 1th retry for invoke remote model`, but if the maxRetryTimes is -1, the retry doesn't happen.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
zhichao-aws: Logger is a required parameter to init a RetryableAction, where it's used to log debug info https://github.com/opensearch-project/OpenSearch/blob/27003270adedebd43eedbd0186c0825f2c0e8613/server/src/main/java/org/opensearch/action/support/RetryableAction.java#L192

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
zane-neo: We should consider if it's better to configure these in connectors, since different endpoint might need different strategies.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: Fixed in the new commit.
BTW if the maxRetryTimes is -1, we don't restrict it, we only restrict by time out value. But the log here does mislead users and I just fixed it

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zane-neo: Consider changing this to milliSeconds, the inference could be in reading path which is sensitive to latency.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
zhichao-aws: And we use the exposed logger at RemoteConnectorExecutor, as we init he RetryableAction here

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zhichao-aws: But change to millis is only about metric, it doesn't change the actual wait time, right?

Since default hyperparameters should work for general scenerios, I believe the top priority is to make the connector run stably under reasonable workload, don't throw throttling exceptions that makes data lost. That's why I set it a rather long timeout value of 30s.
For latency sensitive use case, they can configure the rest request level timeout to achieve the same effect

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
zhichao-aws: Makes sense.

ylwu-amzn: Seems we can add to ConnectorClientConfig class

ylwu-amzn:
Can you add more details description to explain what this PR does to fix the issue ?

plugin/src/main/java/org/opensearch/ml/utils/RemoteInferenceRetryUtils.java
ylwu-amzn: Add retry configurations to ConnectorClientConfig?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
```diff
@@ -0,0 +1,21 @@
+package org.opensearch.ml.engine.algorithms.remote;
```
ylwu-amzn, line 1: no license header

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -112,73 +112,24 @@ private void handleThrottlingInHeader(SdkHttpFullResponse sdkResponse) {
...
-                            .getExceptionHolder()
-                            .compareAndSet(null, new MLException("Fail to execute predict in aws connector", e));
-                    }
-                }
-            }
-        }
-    }
-
-    // Only all requests successful case will be processed here.
-    private void reOrderTensorResponses(Map<Integer, ModelTensors> tensorOutputs) {
```
ylwu-amzn, line 168: Why remove this method ? Are you trying to refactor Zan's code ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: Yes, we need to refactor it to use GroupedActionListener. With previous code we call onResponse/onFailure only after all sub-requests have response. With the refactor, we call onResponse/onFailure for all sub-requests, and the onFailure will further trigger the retry logic for sub-request in RetryableAction.
The sort logic is moved to another ActionListener which wraps the GroupedActionListener. After all sub-requests succeed, the GroupedActionListener.onResponse will trigger the sort logic. After that the sorted results will be sent to the final delegate listener.

zhichao-aws:
> Can you add more details description to explain what this PR does to fix the issue ?

Sure. Updated in the PR description

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +199,65 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+        ActionListener<Tuple<Integer, ModelTensors>> actionListener
+    ) {
+        final RetryableAction<Tuple<Integer, ModelTensors>> invokeRemoteModelAction = new RetryableAction<>(
+            getLogger(),
+            getClient().threadPool(),
+            TimeValue.timeValueMillis(executionContext.getConnectorRetryOption().getRetryBackoffMillis()),
+            TimeValue.timeValueSeconds(executionContext.getConnectorRetryOption().getRetryTimeoutSeconds()),
+            actionListener,
+            BackoffPolicy
+                .constantBackoff(
```
zane-neo, line 224: Is there jitter involved in this constant backoff policy? We should consider making this a configuration as well, different endpoint might need different policy.

zhichao-aws: The jitter is not involved. As mentioned in #2438 , constant backoff is good enough from performance perspective. And since there is still a queue maintained at async client,  the retry request is just put back at the tail of the queue, and not guaranteed to get executed. So I think the constant backoff is good enough.
Can you think of any use cases that we need exponential backoff? I think we can extend current retry options once we find an actual use case or customers call out about it. We already have multiple settings of retry, and I don't want it too complicated. (From my initial thought, the default settings are good enough and it will be rare to tune these settings)

plugin/src/main/java/org/opensearch/ml/utils/RemoteInferenceRetryUtils.java
zhichao-aws: I moved the settings to ConnectorClientConfig in the latest commit

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
zhichao-aws: Moved to ConnectorClientConfig in the latest commit.

zhichao-aws:
The settings are moved to ConnectorClientConfig, and I've checked that connector upgraded from older version has the default retry settings.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zane-neo: LGTM, the reOrder method is basically moved to a different place which doesn't seem to break the functionality.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
zane-neo: We don't have a mechanism that to identify the request is from query or ingestion, and since we're targeting to solve ingestion issue, I think it's fine to leave it what it is now.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zane-neo: It's better to add jitter to avoid all the failure requests are been retried at the same time which can highly possible causing the retry failing. For exponential backoff:
1. Choosing constant or exponential is based on different cases, e.g. if a service doesn't have peak traffic then it's fine to use constant backoff, otherwise an exponential might be more useful since during the peak time, constant can keep failing.
2. I don't think adding an optional configuration to user is adding complication, we can still set default backoff policy to constant.
3. When adding a new feature it's better to think of it a complete one, which means if user wants to use exponential policy, they can simply configure it instead of call out to community. Also, though we're addressing the sagemaker issue initially but we should consider all other cases since this solution is a generic one.

zhichao-aws: OK it makes sense to add a settings for determine here. I'll add a settings to support ExponentialEqualJitterBackoffIterator and ExponentialFullJitterBackoff here

zhichao-aws: For ConstantBackoff, In OpenSearch core we don't have any existing interface to add jitter to it. But I don't think this is a blocking issue. We can make contributions to core and extend the settings here in the future

zane-neo: Sure, it's possible that constant backoff was designed to not have jitter, I think the currently policies are enough to cover different cases, if a different policy is needed, we can add it in the future.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -107,15 +164,34 @@ public static ConnectorClientConfig parse(XContentParser parser) throws IOExcept
...
+                    retryBackoffMillis = parser.intValue();
+                    break;
+                case RETRY_TIMEOUT_SECONDS_FIELD:
+                    retryTimeoutSeconds = parser.intValue();
+                    break;
+                case MAX_RETRY_TIMES_FIELD:
+                    maxRetryTimes = parser.intValue();
+                    break;
+                case RETRY_BACKOFF_POLICY_FIELD:
+                    retryBackoffPolicy = parser.text();
```
zane-neo, line 177: Add break here to reduce the potential risk.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -197,27 +148,44 @@ public void onError(Throwable t) {
...
 
-    private void response(Map<Integer, ModelTensors> tensors) {
-        processResponse(statusCode, responseBody.toString(), parameters, tensorOutputs);
-        executionContext.getCountDownLatch().countDown();
-        // when countdown's count equals to 0 means all responses are received.
-        if (executionContext.getCountDownLatch().getCount() == 0) {
-            if (executionContext.getExceptionHolder().get() != null) {
-                actionListener.onFailure(executionContext.getExceptionHolder().get());
-                return;
+    private void response() {
```
zane-neo, line 160: In this method there're three places are dealing with exception with similar codes, and the `listener.onFailure` is not near them, suggest adding a new method to deal with the exceptions like this:
```
void handleException(Exception e) {
  if (exceptionHolder.get() != null) {
    actionListener.onFailure(exceptionHolder.get());
  } else {
    actionListener.onFailure(e);
  }
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +200,78 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+        final RetryableAction<Tuple<Integer, ModelTensors>> invokeRemoteModelAction = new RetryableAction<>(
+            getLogger(),
+            getClient().threadPool(),
+            TimeValue.timeValueMillis(getConnectorClientConfig().getRetryBackoffMillis()),
+            TimeValue.timeValueSeconds(getConnectorClientConfig().getRetryTimeoutSeconds()),
+            actionListener,
+            getRetryBackoffPolicy(getConnectorClientConfig()),
+            RETRY_EXECUTOR
+        ) {
+            Integer retryTimes = 0;
```
zane-neo, line 244: Seems `int` is better here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +200,78 @@ && getUserRateLimiterMap().get(user.getName()) != null
             if (getMlGuard() != null && !getMlGuard().validate(payload, MLGuard.Type.INPUT)) {
                 throw new IllegalArgumentException("guardrails triggered for user input");
             }
-            invokeRemoteModel(mlInput, parameters, payload, tensorOutputs, countDownLatch, actionListener);
+            if (getConnectorClientConfig().getMaxRetryTimes() != 0) {
```
zane-neo, line 203: Why checking max retry times instead of a switch? We should consider to add a switch to control retry since some users might not want to enable this feature.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +200,78 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+                return BackoffPolicy
+                    .exponentialEqualJitterBackoff(
+                        connectorClientConfig.getRetryBackoffMillis(),
+                        connectorClientConfig.getRetryTimeoutSeconds()
+                    );
+            case "exponential_full_jitter":
+                return BackoffPolicy.exponentialFullJitterBackoff(connectorClientConfig.getRetryBackoffMillis());
+            default:
+                return BackoffPolicy
+                    .constantBackoff(TimeValue.timeValueMillis(connectorClientConfig.getRetryBackoffMillis()), Integer.MAX_VALUE);
```
zane-neo, line 224: What's the second parameter and why using Integer.MAX_VALUE, please add some comment.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.opensearch.ml.engine.algorithms.remote;
+
+import java.io.IOException;
+
+import org.opensearch.OpenSearchStatusException;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.rest.RestStatus;
+
+public class RetryableException extends OpenSearchStatusException {
```
zane-neo, line 14: Is it possible that user can see `retry...` anywhere in the response if return this exception?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: Set the retry times to zero is same as switch off the feature. We can point out this in our documents. 

We already have 4 settings for retry and don't want it over designed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: The second parameter is the maxNumberOfRetries for ConstantBackoff. However, we can't reuse it, because the ConstantBackoffIterator.next() throws exception when it reaches the limit, but the RetryableAction doesn't handle the exception throwed from iterator, and will make the request hanging. 

Setting it to  Integer.MAX_VALUE to avoid throwing this exception. Instead we handle the max retry numbers at `shouldRetry`.

zhichao-aws: Comments added

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
zhichao-aws: If the retry succeed, then users won't see it in the response. If the request fails after timeout or exceed max retry times, users can see all exceptions during the retries.
By setting the logger to debug level, users can see the retry log and exception information. (logged at should retry)

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -197,27 +157,39 @@ public void onError(Throwable t) {
...
+        String body = responseBody.toString();
+        if (Strings.isBlank(body)) {
+            log.error("Remote model response body is empty!");
+            handleException(new OpenSearchStatusException("No response from model", RestStatus.BAD_REQUEST), null);
         } else {
-            log.debug("Not all responses received, left response count is: " + executionContext.getCountDownLatch().getCount());
+            if (statusCode < HttpStatus.SC_OK || statusCode > HttpStatus.SC_MULTIPLE_CHOICES) {
+                handleException(
+                    new OpenSearchStatusException(REMOTE_SERVICE_ERROR + body, RestStatus.fromCode(statusCode)),
+                    String.format(Locale.ROOT, "Remote server returned error code: %d", statusCode)
```
zane-neo, line 178: It seems only here is passing the errorMsg, can add log here and change the handleException method to accept only one parameter, cause a lot of invocation passing `null` doesn't seem neat. Also I prefer to move the actionListener.onFailure to handleException method, so this method will be more clean.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
zane-neo: I mean when OpenSearchStatusException is returned to user, the response is similar to below:
```
{
  "error": {
    "root_cause": [
      {
        "type": "status_exception",
        "reason": "Model is disabled."
      }
    ],
    "type": "status_exception",
    "reason": "error message"
  },
  "status": 403
}
```
What if the retryableException is returned to user? Will it show `retry_exception` or still `status_exception`?

zhichao-aws: Here is a real response body fails with max_retry_times=2
Do you think we need to change it?
```
[{'index': {'_index': 'scifact',
   '_id': '967',
   'status': 400,
   'error': {'type': 'retryable_exception',
    'reason': 'retryable_exception: Error from remote service: The request was denied due to remote server throttling.',
    'suppressed': [{'type': 'retryable_exception',
      'reason': 'retryable_exception: Error from remote service: The request was denied due to remote server throttling.'},
     {'type': 'retryable_exception',
      'reason': 'retryable_exception: Error from remote service: The request was denied due to remote server throttling.'}]}}}]
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: We use the exceptionHolder because @Zhangxunmt call out that we may parse more errors except throttling from onHeaders. Since we need to cache the exceptions and call onFailure in response() method, using exceptionHolder can reduces boolean flags.
I just realized that we can only use the exceptionHolder in onHeaders, and directly call onFailure at response(). This is a more neat way

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RetryableException.java
zhichao-aws: Changed it to SageMakerThrottlingException, now we'll get this:
```
[{'index': {'_index': 'scifact',
   '_id': '897',
   'status': 400,
   'error': {'type': 'sage_maker_throttling_exception',
    'reason': 'sage_maker_throttling_exception: Error from remote service: The request was denied due to remote server throttling.',
    'suppressed': [{'type': 'sage_maker_throttling_exception',
      'reason': 'sage_maker_throttling_exception: Error from remote service: The request was denied due to remote server throttling.'},
     {'type': 'sage_maker_throttling_exception',
      'reason': 'sage_maker_throttling_exception: Error from remote service: The request was denied due to remote server throttling.'}]}}}]
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: updated

zane-neo (approved):

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -26,45 +29,83 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
...
+            this.maxRetryTimes = input.readOptionalInt();
+            this.retryBackoffPolicy = input.readOptionalString();
+        }
     }
 
     public ConnectorClientConfig() {
         this.maxConnections = MAX_CONNECTION_DEFAULT_VALUE;
         this.connectionTimeout = CONNECTION_TIMEOUT_DEFAULT_VALUE;
         this.readTimeout = READ_TIMEOUT_DEFAULT_VALUE;
+        this.retryBackoffMillis = RETRY_BACKOFF_MILLIS_DEFAULT_VALUE;
```
ylwu-amzn, line 91: So no matter user configure it or not, we will always add retry for all connectors? That's a breaking change for the old versions.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -107,15 +164,35 @@ public static ConnectorClientConfig parse(XContentParser parser) throws IOExcept
...
+                    break;
+                case RETRY_BACKOFF_POLICY_FIELD:
+                    retryBackoffPolicy = parser.text();
+                    break;
                 default:
                     parser.skipChildren();
                     break;
             }
         }
+        if(!ALL_SUPPORTED_BACKOFF_POLICY.contains(retryBackoffPolicy)){
```
ylwu-amzn, line 184: No need to add this

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ExecutionContext.java
```diff
@@ -18,15 +15,10 @@
...
  * sequence is used to identify the index of the split request.
- * countDownLatch is used to wait for all the split requests to finish.
- * exceptionHolder is used to hold any exception thrown in a split-batch request.
  */
 @Data
 @AllArgsConstructor
 public class ExecutionContext {
     // Should never be null
     private int sequence;
-    private CountDownLatch countDownLatch;
```
ylwu-amzn, line 29: Why remove this part ? Is it ok to remove this @zane-neo, can you confirm ?

ylwu-amzn: This PR is not purely adding retry, it also contains code refactoring. 

I see several times that people want to fix other issues/add some other minor changes in same PR, but it turns out the extra change brings new bugs. Please work with Zan to make sure it will not break BWC.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -188,16 +200,84 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+            } else {
+                invokeRemoteModel(mlInput, parameters, payload, executionContext, actionListener);
+            }
+        }
+    }
+
+    default BackoffPolicy getRetryBackoffPolicy(ConnectorClientConfig connectorClientConfig) {
+        // we've validated the value of retryBackOffPolicy at ConnectorClientConfig.parse
+        switch (connectorClientConfig.getRetryBackoffPolicy()) {
+            case "exponential_equal_jitter":
```
ylwu-amzn, line 214: Suggest create ENUM or constant

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -112,73 +119,21 @@ private void handleThrottlingInHeader(SdkHttpFullResponse sdkResponse) {
...
-
-    // Only all requests successful case will be processed here.
-    private void reOrderTensorResponses(Map<Integer, ModelTensors> tensorOutputs) {
-        ModelTensors[] modelTensors = new ModelTensors[tensorOutputs.size()];
-        log.debug("Reordered tensor outputs size is {}", tensorOutputs.size());
-        for (Map.Entry<Integer, ModelTensors> entry : tensorOutputs.entrySet()) {
-            modelTensors[entry.getKey()] = entry.getValue();
+        if (containsThrottlingException) {
+            handleException(
+                new SageMakerThrottlingException(
```
ylwu-amzn, line 124: The class name `SageMakerThrottlingException` is limited to Sagemaker, let's use a more general name

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ExecutionContext.java
zhichao-aws: This is a neccessary refactor to support retry for split-batch.

zhichao-aws: Sure, I've gone through several meetings with Zan about this changes. And also conducted experiments at my environment to test these code.

I'll follow up if @zane-neo has any concerns on this

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
zhichao-aws: How about RemoteConnectorThrottlingException?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zhichao-aws: ack

zhichao-aws: updated to enum

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ExecutionContext.java
zane-neo: Since using GroupActionListener and this listener internally handles the multiple split request's response so no need to handle this with CountDownLatch, this change LGTM.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
Zhangxunmt: +1, can we add a flag here to disable the retry by default so BWC will not break?

zhichao-aws: This is an enhancements to improve the user experience, so I believe we should open it by default. For bwc, I tested restart upgrade and rolling upgrade, this PR doesn't break anything

zane-neo: Agree, this brings benefit to customers, we should enable this by default.

Zhangxunmt: This is the lowest level function that many other application depends on like neural search. They have implemented retry 3 times by default. This PR enforces every dependency to retry at least 3 times by default. In the neural search they would retry 9 times when throttled. It's breaking the BWC behavior at least for retry times.

Zhangxunmt: In throttling, does customer really wants retry 3 times more than their setups?

ylwu-amzn: Think about this, user has some code which on top of ml-commons, they have implemented some retry , like retry 5 times. Now we add another retry in ml-commons , like 3 times, so when this feature out, the  user will see 15 times retry , right ? That will be too much for them. Let's have a setting to control enable/disable this retry, or retry times.

Zhangxunmt (approved):

zane-neo (approved):

Issue #2419, opened 2024-05-08
[BUG] Failed to delete model 


ylwu-amzn:
**What is the bug?**

Undeploy Response is successful with state as "undeployed" and when we try to delete the model, It is throwing an error saying model cannot be deleted in deploying or deployed state. The calls were within milliseconds.

**How can one reproduce the bug?**

**What is the expected behavior?**
If undeploy API return successfully, it should mean the model deployed and model state flushed to index correctly.

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

ylwu-amzn:
@b4sjoo Can you help take a look ?

b4sjoo:
Ok, let me try if I can reproduce it. I think it is a little bit flaky.

b4sjoo:
According to our investigation, this is because when we return the undeploy response the cluster might still be executing `syncUpUndeployedModels` task. The above PR should fix the issue by wrapping it with the listener.

zhichao-aws:
Seems #1986 has the same root cause

dblock:
[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/krisfreedain)]

PR #2471, opened 2024-05-23
Add retry to model delete API

b4sjoo:
### Description
Adding a retry parameter in model delete API, with retry times and delay time interval
 
### Issues Resolved
This PR should resolve https://github.com/opensearch-project/ml-commons/issues/2419
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/model/MLModelDeleteRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/model/MLModelDeleteRequest.java
@@ -9,6 +9,7 @@ import lombok.Builder;
 import lombok.Getter;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.core.common.io.stream.InputStreamStreamInput;
 import org.opensearch.core.common.io.stream.OutputStreamStreamOutput;
 import org.opensearch.core.common.io.stream.StreamInput;
@@ -21,24 +22,40 @@ import java.io.UncheckedIOException;
 
 import static org.opensearch.action.ValidateActions.addValidationError;
 
+@Getter
 public class MLModelDeleteRequest extends ActionRequest {
-    @Getter
     String modelId;
+    Boolean retry;
+    Integer maxRetry;
+    TimeValue retryDelay;
+    TimeValue retryTimeout;
 
     @Builder
-    public MLModelDeleteRequest(String modelId) {
+    public MLModelDeleteRequest(String modelId, Boolean retry, Integer maxRetry, TimeValue retryDelay, TimeValue retryTimeout) {
         this.modelId = modelId;
+        this.retry = retry;
+        this.maxRetry = maxRetry;
+        this.retryDelay = retryDelay;
+        this.retryTimeout = retryTimeout;
     }
 
     public MLModelDeleteRequest(StreamInput input) throws IOException {
         super(input);
         this.modelId = input.readString();
+        this.retry = input.readOptionalBoolean();
+        this.maxRetry = input.readOptionalVInt();
+        this.retryDelay = input.readOptionalTimeValue();
+        this.retryTimeout = input.readOptionalTimeValue();
     }
 
     @Override
     public void writeTo(StreamOutput output) throws IOException {
         super.writeTo(output);
         output.writeString(modelId);
+        output.writeOptionalBoolean(retry);
+        output.writeOptionalVInt(maxRetry);
+        output.writeOptionalTimeValue(retryDelay);
+        output.writeOptionalTimeValue(retryTimeout);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -181,6 +164,31 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
 
+    @VisibleForTesting
+    void retryDeleteModel(
+        String modelId,
+        MLModelState mlModelState,
+        Boolean isHidden,
+        MLModelDeleteRequest mlModelDeleteRequest,
+        ActionListener<DeleteResponse> actionListener
+    ) {
+        Boolean retry = mlModelDeleteRequest.getRetry();
```
rbhavna, line 175: where exactly are these retry fields being used?

b4sjoo: At that time the PR is not ready to be reviewed yet. Now is ok.

b4sjoo:
All checking passed the previously failed IT, feel free to search `VisualizationsToolIT` to confirm: https://github.com/opensearch-project/ml-commons/actions/runs/9230239830/job/25398062518

common/src/main/java/org/opensearch/ml/common/transport/model/MLModelDeleteRequest.java
```diff
@@ -21,24 +22,40 @@
...
+        this.retry = retry;
+        this.maxRetry = maxRetry;
+        this.retryDelay = retryDelay;
+        this.retryTimeout = retryTimeout;
     }
 
     public MLModelDeleteRequest(StreamInput input) throws IOException {
         super(input);
         this.modelId = input.readString();
+        this.retry = input.readOptionalBoolean();
```
b4sjoo, line 45: Tested with old master - new data & new master - old data, no need to add version filter here to enable bwc.

common/src/main/java/org/opensearch/ml/common/transport/model/MLModelDeleteRequest.java
```diff
@@ -21,24 +22,40 @@
...
+    TimeValue retryTimeout;
 
     @Builder
-    public MLModelDeleteRequest(String modelId) {
+    public MLModelDeleteRequest(String modelId, Boolean retry, Integer maxRetry, TimeValue retryDelay, TimeValue retryTimeout) {
         this.modelId = modelId;
+        this.retry = retry;
+        this.maxRetry = maxRetry;
+        this.retryDelay = retryDelay;
+        this.retryTimeout = retryTimeout;
```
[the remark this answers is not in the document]
b4sjoo: See https://github.com/opensearch-project/ml-commons/pull/2471/files#diff-a13464eb2560d8f4b9ba6dc62461c02026c1d4681e18a6050c7e8fdd1b316c9fR51-R54, they all got default values.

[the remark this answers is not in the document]
Zhangxunmt: What if this delete API is called at the transport layer, do the same defaults values apply?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -181,6 +142,70 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+    @VisibleForTesting
+    void retryDeleteModel(
+        String modelId,
+        Boolean isHidden,
+        MLModelDeleteRequest mlModelDeleteRequest,
+        ActionListener<DeleteResponse> actionListener
+    ) {
+        Boolean isRetry = mlModelDeleteRequest.getRetry() == null || mlModelDeleteRequest.getRetry();
+        TimeValue retryDelay = mlModelDeleteRequest.getRetryDelay();
+        TimeValue retryTimeout = mlModelDeleteRequest.getRetryTimeout();
```
rbhavna, line 154: Where are these two variables being used?

rbhavna:
> I'm not sure if we are addressing the right problem here. Ideally if the model is un-deployed correctly (from index) then we should not have this problem.
> 
> May be we should update the undeploy api where we will send the final response (may be we need to wait there?) to the customer when undeploy status is written in the index not just from the cache?

Yes agreed. The other APIs like predict/get might be having similar issue. Like if user sends predict request immediately after getting undeployed response, ideally it should have failed. But the request goes through because the index is still not refreshed with undeployed status. Similarly get model API will still show the status as deployed when sent immediately after undeploying and could be confusing to user.

dbwiddis:
> I'm not sure if we are addressing the right problem here. Ideally if the model is un-deployed correctly (from index) then we should not have this problem.
> 
> May be we should update the undeploy api where we will send the final response (may be we need to wait there?) to the customer when undeploy status is written in the index not just from the cache?

I had to hack around this problem in flow framework during our deprovisioning using time delays. (100ms was sufficient in most cases.) I'm very in favor of a slightly shorter delay in undeploy response as a preference to retries.

A similar situation may occur when attempting to delete a connector used in a model, that this may solve.

PR #2485, opened 2024-05-30
[Feature/multi_tenancy] Remove strict version dependency to compile minimum compatible version

dbwiddis:
### Description

Changes the Version compatibility constants to calculate independently of constants in those OpenSearch versions.
 
### Issues Resolved

Relates to #2483 (which can forward-port this commit to `main`)
 
### Check List
- ~[ ] New functionality includes testing.~
  - ~[ ] All tests pass~
- ~[ ] New functionality has been documented.~
  - ~[ ] New functionality has javadoc added~
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2486, opened 2024-05-30
[Backport main] Remove strict version dependency to compile minimum compatible version

### LINKED ISSUE #2483: [BUG] Minimum supported version check requires that minimum version to compile
**What is the bug?**

In the 2.14 release, two constants were added to gate features based on versions below 2.14:
```java
    public static final Version MINIMAL_SUPPORTED_VERSION_FOR_INTERFACE = Version.V_2_14_0;
    public static final Version MINIMAL_SUPPORTED_VERSION_FOR_MODEL_TTL = Version.V_2_14_0;
```

Unfortunately, those version constants are only available in the OpenSearch 2.14.0 (and later) dependency, which means that we must be on 2.14 (or later) to compile the plugin, and these gating checks are essentially useless.

There are similar constants for other minimal supported versions going back to 2.11.0 which are all essentially useless.

1. On the 2.x branch, change the version in `build.gradle` to a version before 2.14.0
2. Run `./gradlew assemble` (or any command that compiles the source)
3. Observe compile failures on the above constants

Code below the minimal supported versions can use those constants as feature support checks.

This impacts support using current 2.x-based feature branches on other environments still using 2.13 as a fixed version.

The code can easily be replaced by a calculation of the `Version` from a string:
```java
Version.fromString("2.14.0")
```

b4sjoo:
### Description
Forward port #2485 to main branch
 
### Issues Resolved
This should close #2483, thanks to @dbwiddis 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.ml.common;
 
+import org.opensearch.Version;
 import org.opensearch.ml.common.agent.MLAgent;
 import org.opensearch.ml.common.connector.AbstractConnector;
 import org.opensearch.ml.common.controller.MLController;
@@ -527,4 +528,9 @@ public class CommonValue {
                         + "\": {\"type\": \"long\"}\n"
                         + "    }\n"
                         + "}";
+        // Calculate Versions independently of OpenSearch core version
+        public static final Version VERSION_2_11_0 = Version.fromString("2.11.0");
+        public static final Version VERSION_2_12_0 = Version.fromString("2.12.0");
+        public static final Version VERSION_2_13_0 = Version.fromString("2.13.0");
+        public static final Version VERSION_2_14_0 = Version.fromString("2.14.0");
 }
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -62,6 +62,15 @@ dependencies {
     testImplementation group: 'commons-io', name: 'commons-io', version: '2.15.1'
     implementation group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
     implementation 'com.jayway.jsonpath:json-path:2.9.0'
+
+    implementation "org.opensearch.client:opensearch-java:2.10.2"
+
+    configurations.all {
+        resolutionStrategy.force 'org.apache.httpcomponents.core5:httpcore5:5.2.4'
+        resolutionStrategy.force 'org.apache.httpcomponents.core5:httpcore5-h2:5.2.4'
+        resolutionStrategy.force 'jakarta.json:jakarta.json-api:2.1.3'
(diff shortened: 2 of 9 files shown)


plugin/build.gradle
```diff
@@ -62,6 +62,15 @@ dependencies {
...
     implementation 'com.jayway.jsonpath:json-path:2.9.0'
+
+    implementation "org.opensearch.client:opensearch-java:2.10.2"
+
+    configurations.all {
+        resolutionStrategy.force 'org.apache.httpcomponents.core5:httpcore5:5.2.4'
+        resolutionStrategy.force 'org.apache.httpcomponents.core5:httpcore5-h2:5.2.4'
+        resolutionStrategy.force 'jakarta.json:jakarta.json-api:2.1.3'
+        resolutionStrategy.force "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
+    }
```
dbwiddis, line 73: I'm not sure you need all of these dependencies, they're specific to the feature branch.

plugin/build.gradle
```diff
@@ -285,7 +294,7 @@ List<String> jacocoExclusions = [
         'org.opensearch.ml.profile.MLPredictRequestStats',
         'org.opensearch.ml.action.deploy.TransportDeployModelAction',
         'org.opensearch.ml.action.deploy.TransportDeployModelOnNodeAction',
-	'org.opensearch.ml.action.undeploy.TransportUndeployModelsAction',
+	    'org.opensearch.ml.action.undeploy.TransportUndeployModelsAction',
```
dbwiddis, line 297: unneeded whitespace.

plugin/build.gradle
b4sjoo: Removed

plugin/build.gradle
b4sjoo: Removed, though idk why it still showed as a change

dbwiddis: Looks like an 8-space tab got switched by someone somewhere to an assumed 4-space tab + 4 spaces, but GitHub does bigger tabs.  Which is why tabs are evil. :)

dbwiddis (approved):

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -527,4 +528,9 @@ public class CommonValue {
                         + "\": {\"type\": \"long\"}\n"
                         + "    }\n"
                         + "}";
+        // Calculate Versions independently of OpenSearch core version
+        public static final Version VERSION_2_11_0 = Version.fromString("2.11.0");
```
Hailong-am, line 532: can we use `Version.V_2_11_0` directly from https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/Version.java#L98

jngz-es (approved):

mingshl (approved):

Issue #2364, opened 2024-04-26
[Enhancement] Add default multi-modal process function in ml-commons


zane-neo:
**Is your feature request related to a problem?**
multi-modal is an already supported feature in neural-search and ml-commons, but now in ml-commons we need to write painless scripts to pre/post-process functions the inputs/outputs of multi-modal requests which is inconvenient for users.

**What solution would you like?**
We can add default pre/post process functions for multi-modal case so user can use very simple string to configure their connectors. 

**What alternatives have you considered?**

**Do you have any additional context?**

dhrubo-os:
@zane-neo are you planning to release this enhancement in 2.15? What's the plan?

zane-neo:
> @zane-neo are you planning to release this enhancement in 2.15? What's the plan?

This will be released in 2.16.

PR #2500, opened 2024-06-05
Add multi modal default preprocess function

zane-neo:
### Description
Add multi modal default preprocess function
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2364
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.common.connector;
 import org.opensearch.ml.common.connector.functions.preprocess.BedrockEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.CohereEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.CohereRerankPreProcessFunction;
+import org.opensearch.ml.common.connector.functions.preprocess.MultiModalEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.OpenAIEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
@@ -22,6 +23,7 @@ public class MLPreProcessFunction {
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
     public static final String TEXT_DOCS_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.bedrock.embedding";
+    public static final String TEXT_IMAGE_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.multimodal.embedding";
     public static final String TEXT_DOCS_TO_DEFAULT_EMBEDDING_INPUT = "connector.pre_process.default.embedding";
     public static final String TEXT_SIMILARITY_TO_COHERE_RERANK_INPUT = "connector.pre_process.cohere.rerank";
     public static final String TEXT_SIMILARITY_TO_DEFAULT_INPUT = "connector.pre_process.default.rerank";
@@ -34,7 +36,9 @@ public class MLPreProcessFunction {
         OpenAIEmbeddingPreProcessFunction openAIEmbeddingPreProcessFunction = new OpenAIEmbeddingPreProcessFunction();
         BedrockEmbeddingPreProcessFunction bedrockEmbeddingPreProcessFunction = new BedrockEmbeddingPreProcessFunction();
         CohereRerankPreProcessFunction cohereRerankPreProcessFunction = new CohereRerankPreProcessFunction();
+        MultiModalEmbeddingPreProcessFunction multiModalEmbeddingPreProcessFunction = new MultiModalEmbeddingPreProcessFunction();
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT, cohereEmbeddingPreProcessFunction);
+        PRE_PROCESS_FUNCTIONS.put(TEXT_IMAGE_TO_BEDROCK_EMBEDDING_INPUT, multiModalEmbeddingPreProcessFunction);
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT, openAIEmbeddingPreProcessFunction);
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_DEFAULT_EMBEDDING_INPUT, openAIEmbeddingPreProcessFunction);
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_BEDROCK_EMBEDDING_INPUT, bedrockEmbeddingPreProcessFunction);
--- a/common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
@@ -8,7 +8,7 @@ package org.opensearch.ml.common.connector;
 import org.opensearch.ml.common.connector.functions.preprocess.BedrockEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.CohereEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.CohereRerankPreProcessFunction;
-import org.opensearch.ml.common.connector.functions.preprocess.MultiModalEmbeddingPreProcessFunction;
+import org.opensearch.ml.common.connector.functions.preprocess.MultiModalConnectorPreProcessFunction;
 import org.opensearch.ml.common.connector.functions.preprocess.OpenAIEmbeddingPreProcessFunction;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
@@ -36,7 +36,7 @@ public class MLPreProcessFunction {
         OpenAIEmbeddingPreProcessFunction openAIEmbeddingPreProcessFunction = new OpenAIEmbeddingPreProcessFunction();
         BedrockEmbeddingPreProcessFunction bedrockEmbeddingPreProcessFunction = new BedrockEmbeddingPreProcessFunction();
         CohereRerankPreProcessFunction cohereRerankPreProcessFunction = new CohereRerankPreProcessFunction();
-        MultiModalEmbeddingPreProcessFunction multiModalEmbeddingPreProcessFunction = new MultiModalEmbeddingPreProcessFunction();
+        MultiModalConnectorPreProcessFunction multiModalEmbeddingPreProcessFunction = new MultiModalConnectorPreProcessFunction();
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT, cohereEmbeddingPreProcessFunction);
         PRE_PROCESS_FUNCTIONS.put(TEXT_IMAGE_TO_BEDROCK_EMBEDDING_INPUT, multiModalEmbeddingPreProcessFunction);
         PRE_PROCESS_FUNCTIONS.put(TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT, openAIEmbeddingPreProcessFunction);
--- a/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
@@ -15,6 +15,7 @@ import org.opensearch.script.ScriptType;
 import org.opensearch.script.TemplateScript;
 
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -45,6 +46,10 @@ public abstract class ConnectorPreProcessFunction implements Function<MLInput, R
     public void validateTextDocsInput(MLInput mlInput) {
         if (!(mlInput.getInputDataset() instanceof TextDocsInputDataSet)) {
             throw new IllegalArgumentException("This pre_process_function can only support TextDocsInputDataSet");
+        }
+        List<String> docs = ((TextDocsInputDataSet) mlInput.getInputDataset()).getDocs();
+        if (docs.size() == 1 && docs.get(0) == null) {
+            throw new IllegalArgumentException("No input text or image provided");
         }
     }
 
--- a/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
@@ -16,11 +16,18 @@ import org.opensearch.script.TemplateScript;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


ylwu-amzn:
Add IT to cover this ?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
```diff
@@ -46,6 +47,10 @@ public void validateTextDocsInput(MLInput mlInput) {
         if (!(mlInput.getInputDataset() instanceof TextDocsInputDataSet)) {
             throw new IllegalArgumentException("This pre_process_function can only support TextDocsInputDataSet");
```
martin-gaievski, line 48: can you make error message more meaningful to the user, if possible please add details of how/what they need to change to fix it.
in addition to that we can log more details at `warn` or `info` level, like what's the actual class of input dataset object

martin-gaievski, line 48: If you have context can you please add class and method level comments, somehow that is missing for this class.

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
+import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
+import org.opensearch.ml.common.input.MLInput;
+
+import java.util.Map;
+
+import static org.opensearch.ml.common.utils.StringUtils.convertScriptStringToJsonString;
+
+public class MultiModalEmbeddingPreProcessFunction extends ConnectorPreProcessFunction {
```
martin-gaievski, line 18: please add class level comments

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
```diff
@@ -0,0 +1,39 @@
...
+    public MultiModalEmbeddingPreProcessFunction() {
+        this.returnDirectlyForRemoteInferenceInput = true;
+    }
+
+    @Override
+    public void validate(MLInput mlInput) {
+        validateTextDocsInput(mlInput);
+    }
+
+    // The input will must have inputText even it's null, input image is optional.
```
martin-gaievski, line 29: please correct wording for comment and format method level comment as per java conventions

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public RemoteInferenceInputDataSet process(MLInput mlInput) {
+        TextDocsInputDataSet inputData = (TextDocsInputDataSet) mlInput.getInputDataset();
```
martin-gaievski, line 32: should be check for type here before we cast to TextDocsInputDataSet?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
```diff
@@ -0,0 +1,39 @@
...
+        if (inputData.getDocs().size() == 1) {
+            return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(Map.of("parameters", Map.of("inputText", inputData.getDocs().get(0))))).build();
+        } else {
```
martin-gaievski, line 35: we need to check if docs collection is not empty

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public void validate(MLInput mlInput) {
+        validateTextDocsInput(mlInput);
+    }
+
+    // The input will must have inputText even it's null, input image is optional.
+    @Override
+    public RemoteInferenceInputDataSet process(MLInput mlInput) {
```
martin-gaievski, line 31:
```suggestion
    public RemoteInferenceInputDataSet process(final MLInput mlInput) {
```

common/src/test/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunctionTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.rmi.Remote;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class MultiModalEmbeddingPreProcessFunctionTest {
```
martin-gaievski, line 27: for this repo all unit tests are derived from OpenSearchTestCase or one of it children, any reason why it's different here?

common/src/test/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunctionTest.java
```diff
@@ -0,0 +1,101 @@
...
+        textDocsInputDataSet = TextDocsInputDataSet.builder().docs(Arrays.asList("hello", "world")).build();
+        remoteInferenceInputDataSet = RemoteInferenceInputDataSet.builder().parameters(Map.of("inputText", "value1", "inputImage", "value2")).build();
+
+        textEmbeddingInput = MLInput.builder().algorithm(FunctionName.TEXT_EMBEDDING).inputDataset(textDocsInputDataSet).build();
+        textSimilarityInput = MLInput.builder().algorithm(FunctionName.TEXT_SIMILARITY).inputDataset(textSimilarityInputDataSet).build();
+        remoteInferenceInput = MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(remoteInferenceInputDataSet).build();
+    }
+
+    @Test
+    public void process_NullInput() {
```
martin-gaievski, line 54: please use camelcase for method name, also please use following format when naming method: "itemBeingTested_scenario_expectedOutcome".

martin-gaievski: same comment for all methods in this test class

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
martin-gaievski: Also can you think of a better name for this class? We're only work with text and image embeddings in a certain format, plus parent class is about connector while this child class is for embedding processor. Maybe `MultiModalModelConnectorPreProcessFunction`, or something similar.

zane-neo: The parent class name connector is a little bit redundant here since the pre process functions are specifically for connector use only. Also it's possible that in the future there'll be a pre process function of multi-modal connector not for embedding, but since the chance is low, I'll change this to the name you suggested.

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
zane-neo: It make sense we log details of input dataset type, but for changing the error message giving an TextDocsInputDataSet example doesn't seem make sense because the format of the error message will be odd. I prefer we give a link to the source code of TextDocsInputDataSet and user can learn the data structure in the code. WDYT?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
zane-neo: The validate method already validated that this should be TextDocsInputDataSet type, so no need to check here.

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
zane-neo: From neural search code: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L287, the collection will never be empty. But to make code more adapt to change, will add check for empty docs.

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
zane-neo: Done

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalEmbeddingPreProcessFunction.java
zane-neo: Done

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final String GITHUB_CI_AWS_REGION = "us-west-2";
+
+    @SneakyThrows
+    @Before
+    public void setup() throws IOException, InterruptedException {
+        RestMLRemoteInferenceIT.disableClusterConnectorAccessControl();
+        Thread.sleep(20000);
+    }
+
+    public void test_bedrock_multimodal_model() throws Exception {
```
ylwu-amzn, line 40: Can you add IT for text or image is null?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalConnectorPreProcessFunction.java
```diff
@@ -0,0 +1,56 @@
...
+     *  The inputText will always show up in the first document, even it's null.
+     */
+    @Override
+    public RemoteInferenceInputDataSet process(MLInput mlInput) {
+        TextDocsInputDataSet inputData = (TextDocsInputDataSet) mlInput.getInputDataset();
+        if (inputData.getDocs().size() == 1) {
+            return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(Map.of("parameters", Map.of("inputText", inputData.getDocs().get(0))))).build();
+        } else {
+            return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(Map.of("parameters", Map.of("inputText", inputData.getDocs().get(0), "inputImage", inputData.getDocs().get(1))))).build();
+        }
```
ylwu-amzn, line 54:
```suggestion
        TextDocsInputDataSet inputData = (TextDocsInputDataSet) mlInput.getInputDataset();
        Map parametersMap = new HashMap<>();
        parametersMap.put("inputText", inputData.getDocs().get(0));
        if (inputData.getDocs().size() > 1) {
            parametersMap.put("inputImage", inputData.getDocs().get(1));
        }
        return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(Map.of("parameters", parametersMap))).build();
    
```

ylwu-amzn (approved):

jngz-es (dismissed):

ylwu-amzn:
@zane-neo resolve conflicts ?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
dhrubo-os: Is there any tutorial / website link we can refer to instead of source code link?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
```diff
@@ -15,16 +15,31 @@
...
+/**
+ * This abstract class represents a pre-processing function for a connector.
+ * It takes an instance of {@link MLInput} as input and returns an instance of {@link RemoteInferenceInputDataSet}.
+ * The input data is expected to be of type {@link MLInput}, and the pre-processing function can be customized by implementing the {@link #validate(MLInput)} and {@link #process(MLInput)} methods.
+ * If the input data is already of type {@link RemoteInferenceInputDataSet}, it can be returned directly by setting the {@link #returnDirectlyForRemoteInferenceInput} flag to true.
+ */
 @Log4j2
 public abstract class ConnectorPreProcessFunction implements Function<MLInput, RemoteInferenceInputDataSet> {
 
     protected boolean returnDirectlyForRemoteInferenceInput;
```
dhrubo-os, line 34: can we add comment for this variable too? What does return directly mean here?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalConnectorPreProcessFunction.java
```diff
@@ -0,0 +1,59 @@
...
+     *  This method validates the input data and then processes it to create a {@link RemoteInferenceInputDataSet} object.
+     *  If the input data is already of type {@link RemoteInferenceInputDataSet}, it is returned directly.
+     *  The inputText will always show up in the first document, even it's null.
+     */
+    @Override
+    public RemoteInferenceInputDataSet process(MLInput mlInput) {
+        TextDocsInputDataSet inputData = (TextDocsInputDataSet) mlInput.getInputDataset();
+        Map<String, String> parametersMap = new HashMap<>();
+        parametersMap.put("inputText", inputData.getDocs().get(0));
+        if (inputData.getDocs().size() > 1) {
```
dhrubo-os, line 53: We want size to be only 2, right? May be check that?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/ConnectorPreProcessFunction.java
zane-neo: We don't have a good document describing this, I think it's fine to add the key data structure of this request to the error message. Will make the change soon.

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/MultiModalConnectorPreProcessFunction.java
zane-neo: It's fine not to check this as we're using fixed index number to fetch the input docs.

xinyual (approved):

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/connector/MLPreProcessFunction.java
```diff
@@ -22,6 +23,7 @@ public class MLPreProcessFunction {
     public static final String TEXT_DOCS_TO_COHERE_EMBEDDING_INPUT = "connector.pre_process.cohere.embedding";
     public static final String TEXT_DOCS_TO_OPENAI_EMBEDDING_INPUT = "connector.pre_process.openai.embedding";
     public static final String TEXT_DOCS_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.bedrock.embedding";
+    public static final String TEXT_IMAGE_TO_BEDROCK_EMBEDDING_INPUT = "connector.pre_process.multimodal.embedding";
```
ylwu-amzn, line 26: How about adding bedrock to the name "connector.pre_process.bedrock.multimodal_embedding" ? That can keep consistent with `"connector.pre_process.bedrock.embedding"`

PR #2508, opened 2024-06-05
ml inference ingest processor support for local models

rbhavna:
### Description
ml inference ingest processor support for local models
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -1000,7 +1000,10 @@ public void loadExtensions(ExtensionLoader loader) {
     public Map<String, org.opensearch.ingest.Processor.Factory> getProcessors(org.opensearch.ingest.Processor.Parameters parameters) {
         Map<String, org.opensearch.ingest.Processor.Factory> processors = new HashMap<>();
         processors
-            .put(MLInferenceIngestProcessor.TYPE, new MLInferenceIngestProcessor.Factory(parameters.scriptService, parameters.client));
+            .put(
+                MLInferenceIngestProcessor.TYPE,
+                new MLInferenceIngestProcessor.Factory(parameters.scriptService, parameters.client, xContentRegistry)
```
mingshl, line 1005: why we need the xContentRegistry passed from the plugin?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
```diff
@@ -162,10 +186,44 @@ private void processPredictions(
...
+            String newDocumentFieldName = entry.getKey();
+            List<String> dotPaths = newOutputMapping.get(newDocumentFieldName);
+
+            int existingFields = 0;
+            for (String path : dotPaths) {
+                if (ingestDocument.hasField(path)) {
+                    existingFields++;
+                }
+            }
+            if (!override && existingFields == dotPaths.size()) {
```
mingshl, line 220: if override is false and existing Fields is the same as the dothPath size, we skip adding the output mapping in silent ?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
```diff
@@ -162,10 +186,44 @@ private void processPredictions(
...
+        }
+
+        for (Map.Entry<String, String> entry : outputMapping.entrySet()) {
+            String newDocumentFieldName = entry.getKey();
+            List<String> dotPaths = newOutputMapping.get(newDocumentFieldName);
+
+            int existingFields = 0;
+            for (String path : dotPaths) {
+                if (ingestDocument.hasField(path)) {
+                    existingFields++;
```
mingshl, line 217: when the document doesn't have the new field, will it add the newField to output mapping?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
```diff
@@ -202,14 +275,10 @@ public void onResponse(MLTaskResponse mlTaskResponse) {
                         // document field as key, model field as value
                         String newDocumentFieldName = entry.getKey();
                         String modelOutputFieldName = entry.getValue();
-                        if (ingestDocument.hasField(newDocumentFieldName)) {
-                            throw new IllegalArgumentException(
-                                "document already has field name "
-                                    + newDocumentFieldName
-                                    + ". Not allow to overwrite the same field name, please check output_map."
-                            );
+                        if (!newOutputMapping.containsKey(newDocumentFieldName)) {
```
mingshl, line 278: please write unit test to test onFailure  and rest the failure

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -45,17 +57,57 @@ public interface ModelExecutor {
...
+        payload = inputMapSubstitutor.replace(payload);
+        StringSubstitutor parametersSubstitutor = new StringSubstitutor(inputParams, "${ml_inference.", "}");
+        payload = parametersSubstitutor.replace(payload);
 
-        return request;
+        if (!isJson(payload)) {
+            throw new IllegalArgumentException("Invalid payload: " + payload);
+        }
+
+        // String jsonStr;
```
mingshl, line 98: please remove the comments

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -45,17 +57,57 @@ public interface ModelExecutor {
...
+        // try {
+        // jsonStr = AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> gson.toJson(inputParams));
+        // } catch (PrivilegedActionException e) {
+        // throw new IllegalArgumentException("wrong connector");
+        // }
+        XContentParser parser = XContentType.JSON.xContent().createParser(xContentRegistry, null, payload);
+
+        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+        MLInput mlInput = MLInput.parse(parser, functionName.name());
+        // MLInput mlInput = MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(inputDataSet).build();
```
mingshl, line 108: remove comments as well

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
```diff
@@ -137,68 +185,69 @@ public void testExecute_nestedObjectStringDocumentSuccess() {
...
+//        List<Map<String, String>> inputMap = getInputMapsForNestedObjectChunks("chunks.*.chunk.text");
+//
+//        MLInferenceIngestProcessor processor = createMLInferenceProcessor("model1", null, inputMap, null, true, false);
+//        ModelTensor modelTensor = ModelTensor.builder().dataAsMap(ImmutableMap.of("response", Arrays.asList(1, 2, 3))).build();
+//        ModelTensors modelTensors = ModelTensors.builder().mlModelTensors(Arrays.asList(modelTensor)).build();
+//        ModelTensorOutput mlModelTensorOutput = ModelTensorOutput.builder().mlModelOutputs(Arrays.asList(modelTensors)).build();
+//
+//        doAnswer(invocation -> {
+//            ActionListener<MLTaskResponse> actionListener = invocation.getArgument(2);
+//            actionListener.onResponse(MLTaskResponse.builder().output(mlModelTensorOutput).build());
```
mingshl, line 199: please remove comments

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
rbhavna: working on unit tests now

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
rbhavna: Yes, it means if the ingest document had a field that already has the processor out field, e.g. text_embedding field, we can skip it in current running processor. User can explicitly set override to true to rewrite the output field though

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
```diff
@@ -184,15 +244,30 @@ private void processPredictions(
             }
         }
 
-        ActionRequest request = getRemoteModelInferenceRequest(modelParameters, inferenceProcessorAttributes.getModelId());
+        Set<String> inputMapKeys = new HashSet<>(modelParameters.keySet());
+        inputMapKeys.removeAll(modelConfigs.keySet());
```
jngz-es, line 248: I don't get it about this logic. Looks like `modelParameter` is same as `modelConfigs` which both are initialized by `inferenceProcessorAttributes.getModelConfigMaps()`.

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -130,11 +177,35 @@ default Object getModelOutputValue(ModelTensorOutput modelTensorOutput, String m
...
+                } catch (Exception e) {
+                    if (ignoreMissing) {
+                        return modelTensorOutputMap;
+                    } else {
+                        throw new IllegalArgumentException("model inference output cannot find such json path: " + modelOutputFieldName, e);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            throw new RuntimeException("An unexpected error occurred: " + e.getMessage());
```
jngz-es, line 205: Shall we also add `ignoreMissing` checking here? If true, return a result?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
rbhavna: `modelParameters` basically consists of all the input fields and values that are sent as part of model inference. It is updated [here](https://github.com/rbhavna/ml-commons/blob/ml_ingest_processor/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java#L357) to include the input fields from ingest document along with model_configs defined in the processor.

What we are doing from line 247 to 252 is separating the inputs that come from ingest document such as `text_docs` and adding them to different map. 

This is because in remote model, we send all inputs and config fields as part of one map called `parameters` But local models need them to be sent as separate fields

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
rbhavna: good catch. But we are adding it [here](https://github.com/rbhavna/ml-commons/blob/ml_ingest_processor/plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java#L262) that will check the ignoreMissing when we call `getModelOutputValue` in line line 190. It should probably cover all the other cases.

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
mingshl: how does the user know if the field is skipped? Maybe we can add some logging.

rbhavna: sure will do that

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
```diff
@@ -350,6 +361,108 @@ public void testMLInferenceProcessorWithForEachProcessor() throws Exception {
...
+        uploadDocument(indexName, "1", uploadDocumentRequestBody);
+        Map document = getDocument(indexName, "1");
+
+        List embeddingList = JsonPath.parse(document).read("_source.books[*].title_embedding");
+        Assert.assertEquals(2, embeddingList.size());
+
+        List embedding1 = JsonPath.parse(document).read("_source.books[0].title_embedding");
+        Assert.assertEquals(1536, embedding1.size());
+        List embedding2 = JsonPath.parse(document).read("_source.books[1].title_embedding");
+        Assert.assertEquals(1536, embedding2.size());
```
mingshl, line 463: for the IT, this test for each processor with nested documents, can you also add the test that do not use for each processor? does it still work fine?

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
```diff
@@ -390,4 +502,24 @@ protected Map getDocument(final String index, final String docId) throws Excepti
         return parseResponseToMap(docResponse);
     }
 
+    protected MLRegisterModelInput registerModelInput() {
+        MLModelConfig modelConfig = TextEmbeddingModelConfig
+            .builder()
+            .modelType("bert")
+            .frameworkType(TextEmbeddingModelConfig.FrameworkType.SENTENCE_TRANSFORMERS)
```
mingshl, line 509: is the local model only supporting for sentence transformers? have you test other type of local models?

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -45,17 +57,50 @@ public interface ModelExecutor {
...
+    ) throws IOException {
         if (parameters == null) {
             throw new IllegalArgumentException("wrong input. The model input cannot be empty.");
         }
-        RemoteInferenceInputDataSet inputDataSet = RemoteInferenceInputDataSet.builder().parameters(parameters).build();
+        FunctionName functionName = FunctionName.REMOTE;
+        if (functionNameStr != null) {
+            functionName = FunctionName.from(functionNameStr);
+        }
+        // RemoteInferenceInputDataSet inputDataSet = RemoteInferenceInputDataSet.builder().parameters(parameters).build();
```
mingshl, line 76: remove comments here

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -45,17 +57,50 @@ public interface ModelExecutor {
...
+        Map<String, Object> inputParams = new HashMap<>();
+        if (FunctionName.REMOTE == functionName) {
+            inputParams.put("parameters", StringUtils.toJson(parameters));
+        } else {
+            inputParams.putAll(parameters);
+        }
 
-        MLInput mlInput = MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(inputDataSet).build();
+        String payload = modelInput;
+        // payload = fillNullParameters(parameters, payload);
```
mingshl, line 86: remove comments here as well

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -45,17 +57,50 @@ public interface ModelExecutor {
...
+        StringSubstitutor modelConfigSubstitutor = new StringSubstitutor(modelConfigs, "${model_config.", "}");
+        payload = modelConfigSubstitutor.replace(payload);
+        StringSubstitutor inputMapSubstitutor = new StringSubstitutor(inputMappings, "${input_map.", "}");
+        payload = inputMapSubstitutor.replace(payload);
```
mingshl, line 90: please add unit tests when the model config is changed.

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
rbhavna: It is also working on sparse_encoders and cross encoding models. Will add more unit tests

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
rbhavna: Yes it works fine. Will add few more tests for local models

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
rbhavna: Found it difficult to add ITs for other models. Added them in UTs. Currently we dont have predict IT tests for other models. Its not letting me use pre-trained models because of size and timing-out issue. I can add them by first adding a few small model URLs to the test data first. I added a TODO within the test class

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -186,6 +183,7 @@ default Object getModelOutputValue(MLOutput mlOutput, String modelOutputFieldNam
         try (XContentBuilder builder = XContentFactory.jsonBuilder()) {
             String modelOutputJsonStr = mlOutput.toXContent(builder, ToXContent.EMPTY_PARAMS).toString();
             Map<String, Object> modelTensorOutputMap = gson.fromJson(modelOutputJsonStr, Map.class);
+            System.out.println("output value" + modelOutputJsonStr);
```
ylwu-amzn, line 186: Remote this line ?

jngz-es (dismissed):

ylwu-amzn (dismissed):

mingshl (approved):

ylwu-amzn (approved):

b4sjoo (approved):

PR #2554, opened 2024-06-14
Fix master key2

ylwu-amzn:
### Description
1. reduce latch wait timeout from 5 seconds to 1 seconds when init encryption master key
2. create new encryption master key if not found
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
@@ -132,22 +132,25 @@ public class EncryptorImpl implements Encryptor {
                             if (ExceptionUtils.getRootCause(e) instanceof VersionConflictEngineException) {
                                 GetRequest getMasterKeyRequest = new GetRequest(ML_CONFIG_INDEX).id(MASTER_KEY);
                                 try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
-                                    client.get(getMasterKeyRequest, ActionListener.wrap(getMasterKey -> {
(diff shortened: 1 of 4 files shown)


rbhavna (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -104,28 +113,67 @@ private void initMasterKey() {
...
+        }, e -> {
+            log.debug("Failed to init ML config index", e);
+            exceptionRef.set(e);
             latch.countDown();
-        }
+        }));
 
         try {
-            latch.await(5, SECONDS);
+            latch.await(1, SECONDS);
```
zane-neo, line 176: Can we make sure the blocking here is safe? It won't have impact on other threads like event loop?

ylwu-amzn: This is not a perfect fix. Just for quickly unblock query assistant. Did testing on single node and multi-node cluster, looks good.

ylwu-amzn: @rbhavna  will have a long-term fix which will remove such await

rbhavna (approved):

zane-neo (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
zane-neo: @rbhavna please create an issue to track this.

PR #2553, opened 2024-06-13
AWS DDB SDK client support for remote data store

arjunkumargiri:
### Description
Add AWS DDB implementation of SDK client. This change would allow plugin metadata to be written to AWS DDB tables. Users need to set environment variable `REMOTE_METADATA_TYPE` as `AWSDynamoDB` to use DDB as remote plugin metadata store.
 
 
### Check List
- [ ] New functionality includes testing.
  - [x ] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2559, opened 2024-06-16
[Feature/multi_tenancy] Add SearchDataObject interface, Client, and Connector Implementations 

dbwiddis:
### Description

 - Adds search method to SdkClient interface
 - Implements search method on Local and Remote clients
 - Migrates Connector Search action to use SdkClient for the searching
 - Updates Delete and Update connector actions which search model index to use new API
 
### Issues Resolved

Continuation of PR https://github.com/opensearch-project/ml-commons/pull/2459
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
+++ b/common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
@@ -16,11 +16,12 @@ public class SearchDataObjectRequest {
     private final SearchSourceBuilder searchSourceBuilder;
 
     /**
-     * Instantiate this request with an index and id
+     * Instantiate this request with an optional list of indices and search source
      * <p>
-     * For data storage implementations other than OpenSearch, an index may be referred to as a table and the id may be referred to as a primary key.
-     * @param indices the indices to search for the object
-     * @param searchSourceBuilder the context to use when fetching _source
+     * For data storage implementations other than OpenSearch, an index may be referred to as a table
+     *
+     * @param indices             the indices to search for the object
+     * @param searchSourceBuilder the search body containing the query
      */
     public SearchDataObjectRequest(String[] indices, SearchSourceBuilder searchSourceBuilder) {
         this.indices = indices;
@@ -72,7 +73,7 @@ public class SearchDataObjectRequest {
          */
         public Builder searchSourceBuilder(SearchSourceBuilder searchSourceBuilder) {
             this.searchSourceBuilder = searchSourceBuilder;
-            
+
             return this;
         }
 
--- a/common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
+++ b/common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+public class SearchDataObjectRequest {
+
+    private final String[] indices;
+    private final SearchSourceBuilder searchSourceBuilder;
+
+    /**
+     * Instantiate this request with an index and id
+     * <p>
+     * For data storage implementations other than OpenSearch, an index may be referred to as a table and the id may be referred to as a primary key.
+     * @param indices the indices to search for the object
+     * @param searchSourceBuilder the context to use when fetching _source
+     */
+    public SearchDataObjectRequest(String[] indices, SearchSourceBuilder searchSourceBuilder) {
+        this.indices = indices;
+        this.searchSourceBuilder = searchSourceBuilder;
+    }
+
+    /**
+     * Returns the indices
+     * @return the indices
+     */
+    public String[] indices() {
+        return this.indices;
+    }
+
+    /**
+     * Returns the builder for searching
+     * @return the SearchSourceBuilder
+     */
+    public SearchSourceBuilder searchSourceBuilder() {
+        return this.searchSourceBuilder;
+    }
+
+    /**
+     * Class for constructing a Builder for this Request Object
+     */
+    public static class Builder {
+        private String[] indices = null;
+        private SearchSourceBuilder searchSourceBuilder;
+
+        /**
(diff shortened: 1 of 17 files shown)


common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private final String[] indices;
+    private final SearchSourceBuilder searchSourceBuilder;
+
+    /**
+     * Instantiate this request with an optional list of indices and search source
+     * <p>
+     * For data storage implementations other than OpenSearch, an index may be referred to as a table
+     *
+     * @param indices             the indices to search for the object
```
dhrubo-os, line 23: How this will work? After searching in multiple indices are we going to combine the results and then send it back through `SearchDataObjectResponse`?

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
```diff
@@ -117,22 +118,41 @@ private void handleConnectorAccessValidationFailure(String connectorId, Exceptio
...
-                if (searchHits.length == 0) {
-                    deleteConnector(connectorId, actionListener);
-                } else {
-                    handleModelsUsingConnector(searchHits, connectorId, actionListener);
-                }
-            }, e -> handleSearchFailure(connectorId, e, actionListener)), context::restore));
+
+            SearchDataObjectRequest searchDataObjectRequest = new SearchDataObjectRequest.Builder()
+                .indices(ML_MODEL_INDEX)
+                .searchSourceBuilder(sourceBuilder)
```
dhrubo-os, line 129: I think we should also add tenant id in the sourceBuilder?

common/src/main/java/org/opensearch/sdk/SearchDataObjectRequest.java
dbwiddis: In practice in ML commons, there's a single index (the system index for connectors, models, etc.).

In general it can be an empty list (search all indices) or multiple indices.   The search request really just packages up the entire JSON of the search query and the full response as well, parsing it back out from the XContent.

So the search response will be identical (some metadata/stats and then a list of hits).

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
dbwiddis: Yes, have you already added it to all the others?  I lost track with the multiple PRs at the same time.

dhrubo-os: No, I haven't made any changes with search functionalities yet.

dbwiddis: I think @arjunkumargiri is adding tenant id in https://github.com/opensearch-project/ml-commons/pull/2553/files

It's not currently in of the other requests; probably needs to be its own standalone field.

dhrubo-os: yeah, I'll address that accordingly after that PR is being merged.

plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java
```diff
@@ -90,13 +99,39 @@ private void search(SearchRequest request, ActionListener<SearchResponse> action
             final ActionListener<SearchResponse> doubleWrappedListener = ActionListener
                 .wrap(wrappedListener::onResponse, e -> wrapListenerToHandleSearchIndexNotFound(e, wrappedListener));
 
-            if (connectorAccessControlHelper.skipConnectorAccessControl(user)) {
-                client.search(request, doubleWrappedListener);
-            } else {
+            if (!connectorAccessControlHelper.skipConnectorAccessControl(user)) {
```
dhrubo-os, line 102: why did we add `!` here

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
```diff
@@ -117,22 +118,41 @@ private void handleConnectorAccessValidationFailure(String connectorId, Exceptio
...
+                            }
+                        } catch (Exception e) {
+                            log.error("Failed to parse search response", e);
+                            actionListener
+                                .onFailure(
+                                    new OpenSearchStatusException("Failed to parse search response", RestStatus.INTERNAL_SERVER_ERROR)
+                                );
+                        }
+                    } else {
+                        Throwable cause = st.getCause() == null ? st : st.getCause();
```
dhrubo-os, line 152: May be we should have a recursive method to identify the root cause as we know the exception can be wrapped with other exception. What do you think about this method:

```
private Throwable getRootCause(Throwable throwable) {
    Throwable cause = throwable;
    while (cause.getCause() != null && cause != cause.getCause()) {
        cause = cause.getCause();
    }
    return cause;
}
```

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java
dbwiddis: See `unwrapCuase()` here.  This seems to do what you're asking for.  https://github.com/opensearch-project/OpenSearch/blob/802f2e6e4b21f27ddc6c01e7fc6f6cdcd69138d3/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java#L127-L145

dhrubo-os: Yes! Should we refactor the code accordingly then?

dbwiddis: Possibly?  But possibly not.   I intentionally only unwrapped one level.  Tracing the code, we know we're using `actionGet` so looking at the code we either throw the original exception (if it's a RTE) or it's a wrapped UncategorizedExecutionException containing the cause, which is what we're extracting here.    So that built in recursive method is a bit too deep.

While a util method is helpful here, a ternary one-liner is also pretty brief so I'm struggling to see the huge benefit of a refactor.  

So I'm open to it, but not thinking it's a high priority.

PR #2575, opened 2024-06-19
multi-tenancy for model group CRUD

dhrubo-os:
### Description
[Implemented multi-tenancy feature with sdkclient for model group apis (CRUD)]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
@@ -24,6 +24,7 @@ import org.opensearch.core.xcontent.XContentParser;
 import java.util.Objects;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID;
 
 @Getter
 public class MLModelGroup implements ToXContentObject {
@@ -52,6 +53,7 @@ public class MLModelGroup implements ToXContentObject {
 
     private Instant createdTime;
     private Instant lastUpdatedTime;
+    private String tenantId;
 
 
     @Builder(toBuilder = true)
@@ -59,7 +61,7 @@ public class MLModelGroup implements ToXContentObject {
                         List<String> backendRoles, User owner, String access,
                         String modelGroupId,
                         Instant createdTime,
-                        Instant lastUpdatedTime) {
+                        Instant lastUpdatedTime, String tenantId) {
         this.name = Objects.requireNonNull(name, "model group name must not be null");
         this.description = description;
         this.latestVersion = latestVersion;
@@ -69,6 +71,7 @@ public class MLModelGroup implements ToXContentObject {
         this.modelGroupId = modelGroupId;
         this.createdTime = createdTime;
         this.lastUpdatedTime = lastUpdatedTime;
+        this.tenantId = tenantId;
     }
 
 
@@ -88,6 +91,8 @@ public class MLModelGroup implements ToXContentObject {
         modelGroupId = input.readOptionalString();
         createdTime = input.readOptionalInstant();
         lastUpdatedTime = input.readOptionalInstant();
+        // TODO Add BWC check
+        tenantId = input.readOptionalString();
     }
 
     public void writeTo(StreamOutput out) throws IOException {
@@ -110,6 +115,8 @@ public class MLModelGroup implements ToXContentObject {
         out.writeOptionalString(modelGroupId);
         out.writeOptionalInstant(createdTime);
         out.writeOptionalInstant(lastUpdatedTime);
+        // TODO check BWC
+        out.writeOptionalString(tenantId);
     }
 
     @Override
@@ -138,6 +145,9 @@ public class MLModelGroup implements ToXContentObject {
         if (lastUpdatedTime != null) {
             builder.field(LAST_UPDATED_TIME_FIELD, lastUpdatedTime.toEpochMilli());
         }
+        if (tenantId != null) {
+            builder.field(TENANT_ID, tenantId);
+        }
         builder.endObject();
         return builder;
     }
@@ -152,6 +162,7 @@ public class MLModelGroup implements ToXContentObject {
         String modelGroupId = null;
         Instant createdTime = null;
         Instant lastUpdateTime = null;
+        String tenantId = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -190,6 +201,8 @@ public class MLModelGroup implements ToXContentObject {
                 case LAST_UPDATED_TIME_FIELD:
                     lastUpdateTime = Instant.ofEpochMilli(parser.longValue());
                     break;
+                case TENANT_ID:
+                    tenantId = parser.textOrNull();
                 default:
                     parser.skipChildren();
                     break;
@@ -205,6 +218,7 @@ public class MLModelGroup implements ToXContentObject {
                 .modelGroupId(modelGroupId)
                 .createdTime(createdTime)
                 .lastUpdatedTime(lastUpdateTime)
+                .tenantId(tenantId)
                 .build();
     }
 
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
@@ -105,7 +105,7 @@ public class MLIndicesHandler {
                         log.info("Skip creating the Index:{} that is already created by another parallel request", indexName);
                         internalListener.onResponse(true);
                     } else {
-                        log.error("Failed to create index " + indexName, e);
+                        log.error("Failed to create index {}", indexName, e);
                         internalListener.onFailure(e);
                     }
                 });
@@ -165,7 +165,7 @@ public class MLIndicesHandler {
                 }
             }
         } catch (Exception e) {
-            log.error("Failed to init index " + indexName, e);
+            log.error("Failed to init index {}", indexName, e);
             listener.onFailure(e);
         }
     }
@@ -187,7 +187,7 @@ public class MLIndicesHandler {
         Integer oldVersion = CommonValue.NO_SCHEMA_VERSION;
         Map<String, Object> indexMapping = indexMetaData.mapping().getSourceAsMap();
         Object meta = indexMapping.get(META);
-        if (meta != null && meta instanceof Map) {
+        if (meta instanceof Map) {
             @SuppressWarnings("unchecked")
             Map<String, Object> metaMapping = (Map<String, Object>) meta;
             Object schemaVersion = metaMapping.get(SCHEMA_VERSION_FIELD);
--- a/plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
@@ -10,7 +10,6 @@ import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.GENERAL_THREAD_POOL;
 import static org.opensearch.ml.utils.RestActionUtils.PARAMETER_MODEL_GROUP_ID;
 
-import org.opensearch.OpenSearchException;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.delete.DeleteRequest;
@@ -35,12 +34,12 @@ import org.opensearch.ml.common.transport.model_group.MLModelGroupDeleteAction;
 import org.opensearch.ml.common.transport.model_group.MLModelGroupDeleteRequest;
 import org.opensearch.ml.helper.ModelAccessControlHelper;
 import org.opensearch.ml.settings.MLFeatureEnabledSetting;
-import org.opensearch.ml.utils.MLExceptionUtils;
 import org.opensearch.ml.utils.RestActionUtils;
 import org.opensearch.ml.utils.TenantAwareHelper;
 import org.opensearch.sdk.DeleteDataObjectRequest;
 import org.opensearch.sdk.DeleteDataObjectResponse;
 import org.opensearch.sdk.SdkClient;
+import org.opensearch.sdk.SdkClientUtils;
 import org.opensearch.sdk.SearchDataObjectRequest;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
@@ -139,7 +138,7 @@ public class DeleteModelGroupTransportAction extends HandledTransportAction<Acti
                                         );
                                 }
                             } else {
-                                Throwable cause = MLExceptionUtils.getRootCause(st);
+                                Exception cause = SdkClientUtils.unwrapAndConvertToException(st);
                                 handleModelSearchFailure(modelGroupId, tenantId, cause, actionListener);
                             }
                         });
@@ -173,13 +172,9 @@ public class DeleteModelGroupTransportAction extends HandledTransportAction<Acti
         ActionListener<DeleteResponse> actionListener
     ) {
         if (throwable != null) {
-            Throwable cause = MLExceptionUtils.getRootCause(throwable);
+            Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
             log.error("Failed to delete ML Model Group {}", modelGroupId, cause);
-            if (cause instanceof Exception) {
-                actionListener.onFailure((Exception) cause);
-            } else {
-                actionListener.onFailure(new OpenSearchException(cause));
-            }
+            actionListener.onFailure(cause);
         } else {
             log.debug("Completed Delete Model Group Request, model group id:{} deleted", response.id());
             DeleteResponse deleteResponse = new DeleteResponse(response.shardId(), response.id(), 0, 0, 0, response.deleted());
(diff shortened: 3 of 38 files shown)


common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
```diff
@@ -52,14 +53,15 @@ public class MLModelGroup implements ToXContentObject {
...
+    private String tenantId;
 
 
     @Builder(toBuilder = true)
     public MLModelGroup(String name, String description, int latestVersion,
                         List<String> backendRoles, User owner, String access,
                         String modelGroupId,
                         Instant createdTime,
-                        Instant lastUpdatedTime) {
+                        Instant lastUpdatedTime, String tenantId) {
```
dbwiddis, line 64: Nit: spotless should have wrapped this on a new line?

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
```diff
@@ -59,4 +59,12 @@ public static void logException(String errorMessage, Exception e, Logger log) {
...
         }
     }
+
+    public static Throwable getRootCause(Throwable throwable) {
+        Throwable cause = throwable;
+        while (cause.getCause() != null && cause != cause.getCause()) {
+            cause = cause.getCause();
+        }
+        return cause;
+    }
```
dbwiddis, line 69: Recommend removing this as the handling has been implemented in #2577

plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
```diff
@@ -81,42 +104,107 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+                                    actionListener
+                                        .onFailure(
+                                            new OpenSearchStatusException(
+                                                "Failed to parse search response",
+                                                RestStatus.INTERNAL_SERVER_ERROR
+                                            )
+                                        );
+                                }
+                            } else {
+                                Throwable cause = MLExceptionUtils.getRootCause(st);
```
dbwiddis, line 142:
```suggestion
                                RuntimeException cause = RuntimeException cause = SdkClientUtils.unwrapAndConvertToRuntime(st);
```

plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
```diff
@@ -81,42 +104,107 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+            return;
+        }
+
+        log.error("Failed to search for models using model group id: {}", modelGroupId, cause);
+        if (cause instanceof Exception) {
+            actionListener.onFailure((Exception) cause);
+        } else {
+            actionListener.onFailure(new OpenSearchException(cause));
+        }
     }
```
dbwiddis, line 209:
```suggestion
    private void handleModelSearchFailure(
        String modelGroupId,
        String tenantId,
        RuntimeException cause,
        ActionListener<DeleteResponse> actionListener
    ) {
        if (cause instanceof IndexNotFoundException) {
            DeleteRequest deleteRequest = new DeleteRequest(ML_MODEL_GROUP_INDEX, modelGroupId);
            deleteModelGroup(deleteRequest, tenantId, actionListener);
            return;
        }

        log.error("Failed to search for models using model group id: {}", modelGroupId, cause);
        actionListener.onFailure(cause);
    }
```

plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
```diff
@@ -81,42 +104,107 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+        ActionListener<DeleteResponse> actionListener
+    ) {
+        if (throwable != null) {
+            Throwable cause = MLExceptionUtils.getRootCause(throwable);
+            log.error("Failed to delete ML Model Group {}", modelGroupId, cause);
+            if (cause instanceof Exception) {
+                actionListener.onFailure((Exception) cause);
+            } else {
+                actionListener.onFailure(new OpenSearchException(cause));
             }
```
dbwiddis, line 182:
```suggestion
            RuntimeException cause = SdkClientUtils.unwrapAndConvertToRuntime(throwable);
            log.error("Failed to delete ML Model Group {}", modelGroupId, cause);
            actionListener.onFailure(cause);
```

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
```diff
@@ -42,83 +47,124 @@
...
-                                            "User doesn't have privilege to perform this operation on this model group",
-                                            RestStatus.FORBIDDEN
-                                        )
-                                    );
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    log.debug("Completed Get Model group Request, id:{}", modelGroupId);
+                    if (throwable != null) {
+                        Throwable cause = throwable.getCause() == null ? throwable : throwable.getCause();
```
dbwiddis, line 105: Replace with new util and simplify onFailure calls as in above examples after merging #2577

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -57,70 +63,110 @@ public class TransportUpdateModelGroupAction extends HandledTransportAction<Acti
...
-                        if (modelAccessControlHelper.isSecurityEnabledAndModelAccessControlEnabled(user)) {
-                            validateRequestForAccessControl(updateModelGroupInput, user, mlModelGroup);
+            ActionListener<MLUpdateModelGroupResponse> wrappedListener = ActionListener.runBefore(listener, context::restore);
+
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    log.debug("Completed Get Model group Request, id:{}", modelGroupId);
+                    if (throwable != null) {
+                        Throwable cause = MLExceptionUtils.getRootCause(throwable);
```
dbwiddis, line 123: Similar comments above; replace with new util and simplify onFailure calls below

plugin/src/main/java/org/opensearch/ml/action/tasks/DeleteTaskTransportAction.java
```diff
@@ -88,7 +89,7 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
                 .whenComplete((r, throwable) -> {
                     log.debug("Completed Get task Request, id:{}", taskId);
                     if (throwable != null) {
-                        Throwable rootCause = getRootCause(throwable);
+                        Throwable rootCause = MLExceptionUtils.getRootCause(throwable);
```
dbwiddis, line 92: Another case for new util

plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
```diff
@@ -108,24 +117,31 @@ public void createModelGroup(MLRegisterModelGroupInput input, ActionListener<Str
...
-                                wrappedListener.onFailure(e);
-                            }));
+                            sdkClient
+                                .putDataObjectAsync(
+                                    new PutDataObjectRequest.Builder().index(ML_MODEL_GROUP_INDEX).dataObject(mlModelGroup).build(),
+                                    client.threadPool().executor(GENERAL_THREAD_POOL)
+                                )
+                                .whenComplete((r, throwable) -> {
+                                    if (throwable != null) {
+                                        Throwable cause = MLExceptionUtils.getRootCause(throwable);
```
dbwiddis, line 132: Another case for new util

plugin/src/main/java/org/opensearch/ml/model/MLModelGroupManager.java
```diff
@@ -192,18 +208,38 @@ public void validateUniqueModelGroupName(String name, ActionListener<SearchRespo
...
+            SearchDataObjectRequest searchDataObjectRequest = new SearchDataObjectRequest.Builder()
+                .indices(searchRequest.indices())
+                .searchSourceBuilder(searchRequest.source())
+                .build();
+
+            sdkClient
+                .searchDataObjectAsync(searchDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    if (throwable != null) {
+                        Throwable cause = MLExceptionUtils.getRootCause(throwable);
```
dbwiddis, line 220: And another

dbwiddis (commented):
LGTM with a formatting nit and our overlapping attempt to handle checked exceptions thrown from inside SdkClient calls.   I think my solution in #2577 is better.

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
```diff
@@ -42,83 +47,124 @@
...
-                                        new OpenSearchStatusException(
-                                            "User doesn't have privilege to perform this operation on this model group",
-                                            RestStatus.FORBIDDEN
-                                        )
-                                    );
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    log.debug("Completed Get Model group Request, id:{}", modelGroupId);
+                    if (throwable != null) {
```
arjunkumargiri, line 104: Generally it is better to handle exception using try catch blocks. Can we follow a pattern similar to https://github.com/opensearch-project/ml-commons/blob/c15ca42385312a92e1f0e4829f01f4b1b31c1008/common/src/main/java/org/opensearch/sdk/SdkClient.java#L76-L79

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
```diff
@@ -42,83 +47,124 @@
...
-                        modelAccessControlHelper.validateModelGroupAccess(user, modelGroupId, client, ActionListener.wrap(access -> {
-                            if (!access) {
-                                wrappedListener
-                                    .onFailure(
-                                        new OpenSearchStatusException(
-                                            "User doesn't have privilege to perform this operation on this model group",
-                                            RestStatus.FORBIDDEN
-                                        )
-                                    );
+            sdkClient
```
arjunkumargiri, line 100: High level comment: This code block have multiple nested if/else block. Better to refactor and break into smaller functions.

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
dhrubo-os: not really I think. I applied `./gradlew spotlessApply`, but that didn't change anything. You can see old code `List<String> backendRoles, User owner, String access,` in the same way.

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
```diff
@@ -88,6 +91,8 @@ public MLModelGroup(StreamInput input) throws IOException{
         modelGroupId = input.readOptionalString();
         createdTime = input.readOptionalInstant();
         lastUpdatedTime = input.readOptionalInstant();
+        // TODO Add BWC check
```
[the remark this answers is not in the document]
dhrubo-os: in a cluster if there's node which is running the old version of OS and if the request is coming from the node with new version, we might see some mismatch.

plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
```diff
@@ -81,42 +104,107 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+                                    SearchResponse searchResponse = SearchResponse.fromXContent(sr.parser());
+                                    SearchHit[] searchHits = searchResponse.getHits().getHits();
+                                    if (searchHits.length == 0) {
+                                        deleteModelGroup(deleteRequest, tenantId, wrappedListener);
+                                    } else {
+                                        actionListener
+                                            .onFailure(
+                                                new OpenSearchStatusException(
+                                                    "Cannot delete the model group when it has associated model versions",
+                                                    RestStatus.CONFLICT
```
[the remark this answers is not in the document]
dhrubo-os: It doesn't need to be. I feel like this is more relevant exception than we had before. Thoughts?

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
dbwiddis: @arjunkumargiri in general, true.

In practice, as I've been migrating `client.foo()` calls to `sdkClient.fooDataObject()` calls I have attempted to preserve the existing content, messy nesting and all.  My goal is migrating, not refactoring for efficiency, even if that's generally a good thing.   

My philosopy here is that the diff is much more easily reviewed by reviewers if the only change is between ActionListener and CompletionStage.  More precisely, the pattern:
```java
client.crud(crudRequest, ActionListener.wrap(
    r -> {
        // foo
    }, e -> {
        // bar
    }));
``` 
can be replaced by
```java
sdkClient.crudAsync(crudDataObjectRequest, executor).whenComplete({
    if (throwable == null) {
        // foo
    } else {
        // extract Exception from throwable
        // bar
    }
});
```

dhrubo-os:
> LGTM with a formatting nit and our overlapping attempt to handle checked exceptions thrown from inside SdkClient calls. I think my solution in #2577 is better.

refactored accordingly.

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
dbwiddis: See for example [this commit](https://github.com/opensearch-project/ml-commons/commit/a4ca16bc705e9d90f72debd203fe12ca60ace392).  This is really just applying the above pattern and updating the tests.

plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
dbwiddis: @arjunkumargiri see my reply to the high level comment earlier.

If we were writing code from scratch I'd be completely on board with this pattern.   New code probably _should_ follow this pattern.

But we're not writing new code. We're migrating existing code that is designed to handle a `void` return method and deal with response or exception in an `ActionListener` and using this `throwable == null` or `response == null` (or their opposites) pattern minimizes the effort involved in migrating and speeds the process.

common/src/main/java/org/opensearch/ml/common/MLModelGroup.java
dbwiddis: Huh. Weird. Spotless hasn't been doing things on this repo that I expect (I find trailing whitespace often).

Seems weird that it would put modelGroupId and createdTime on separate lines but then combine the next two arguments on one line. I think Spotless is broken :-|

ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
```diff
@@ -187,7 +187,7 @@ public void shouldUpdateIndex(String indexName, Integer newVersion, ActionListen
         Integer oldVersion = CommonValue.NO_SCHEMA_VERSION;
         Map<String, Object> indexMapping = indexMetaData.mapping().getSourceAsMap();
         Object meta = indexMapping.get(META);
-        if (meta != null && meta instanceof Map) {
+        if (meta instanceof Map) {
```
dbwiddis, line 190: +1, the day I discovered `instanceof` includes null checks changed my life ;)

plugin/src/main/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportAction.java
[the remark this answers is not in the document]
dbwiddis: Per my comments elsewhere in review, I'd personally prefer to keep existing exceptions/code logic as-is if we're just migrating the wrappers.  I agree it may be a better exception type but would prefer that in a separate PR (perhaps on the main branch backported here).

plugin/src/main/java/org/opensearch/ml/action/model_group/TransportUpdateModelGroupAction.java
```diff
@@ -57,70 +62,106 @@ public class TransportUpdateModelGroupAction extends HandledTransportAction<Acti
...
+                            wrappedListener.onFailure(new OpenSearchStatusException("Failed to find model group", RestStatus.NOT_FOUND));
                         } else {
-                            validateSecurityDisabledOrModelAccessControlDisabled(updateModelGroupInput);
+                            log.error("Failed to get ML group {}", modelGroupId, cause);
+                            wrappedListener.onFailure(cause);
+                        }
+                    } else {
+                        if (r != null && r.parser().isPresent()) {
+                            try {
+                                XContentParser parser = r.parser().get();
```
dbwiddis, line 133:
```suggestion
                            try (XContentParser parser = r.parser().get()) {                               
```

dbwiddis (approved):
LGTM generally.

I agree with @arjunkumargiri that we should avoid making refactorings that should really belong in main/2.x and get backported here, such as better exceptions.

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
dhrubo-os: Yup, I removed it.

PR #2576, opened 2024-06-19
[Feature/multi_tenancy] Add Tenant ID to Search API, refactor and add util methods

dbwiddis:
### Description

1. Adds tenant ID to SearchDataObjectRequest.  It was already added to the other requests in #2553 
2. Refactors `unwrapAndConvertToRuntime` static method from `SdkClient` interface to new `SdkClientUtils` class
3. Adds new `getRethrownExecutionExceptionRootCause` static method to `SdkClientUtils` class to handle common pattern of throwables from `actionGet()`.   (Have not yet migrated implementations; will do those as part of future PRs.)
 
### Issues Resolved

PR comments on multiple previous PRs.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -9,10 +9,15 @@
 package org.opensearch.sdk;
 
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+
 import org.opensearch.OpenSearchException;
+import org.opensearch.common.action.ActionFuture;
+import org.opensearch.common.util.concurrent.FutureUtils;
+import org.opensearch.common.util.concurrent.UncategorizedExecutionException;
 
 public class SdkClientUtils {
-    
+
     /**
      * Unwraps the cause of a {@link CompletionException}. If the cause is a subclass of {@link RuntimeException}, rethrows the exception.
      * Otherwise wraps it in an {@link OpenSearchException}. Properly re-interrupts the thread on {@link InterruptedException}.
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
```diff
@@ -0,0 +1,51 @@
...
+     * Get the original exception of an {@link UncategorizedExecutionException} with two levels of cause nesting.
+     * Intended to recreate the root cause of an exception thrown by {@link ActionFuture#actionGet}, which was handled by {@link FutureUtils#rethrowExecutionException}.
+     * @param throwable a throwable with possibly nested causes
+     * @return the root cause of an ExecutionException if it was not a RuntimeException, otherwise the original exception 
+     */
+    public static Throwable getRethrownExecutionExceptionRootCause(Throwable throwable) {
+        if (throwable instanceof UncategorizedExecutionException && throwable.getCause() instanceof ExecutionException) {
+            return throwable.getCause().getCause();
+        }
+        return throwable;
```
dhrubo-os, line 49: Shouldn't this be `throwable.getCause()`?

dbwiddis: Nope, `throwable` is an `UncategorizedExecutionException` and `throwable.getCause()` is an `ExecutionException`.  So the original exception is in two levels of nesting.

Probably not the original intent but that's how the exception wrapping in `actionGet()` works.

The test class makes it clear/simple with one RuntimeException and one not (IOException).

dbwiddis: The double-wrapped causes is what you implemented [here](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/tasks/DeleteTaskTransportAction.java#L156-L162).  The difference is that I'm very specifically targeting the 2-layer nesting from actionGet vs. a general arbitrarily deep nesting.

dbwiddis: Also coverage report shows both branches are covered by the two tests:
image

dhrubo-os (approved):

PR #2577, opened 2024-06-19
[Feature/multi_tenancy] Improve exception unwrapping to handle all cases

dbwiddis:
### Description

Improves the `SdkClientUtils` class implementations to handle all of the unwrapping of CompletionException and possibly double-wrapped ExecutionExceptions from checked root causes.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -8,6 +8,7 @@
  */
 package org.opensearch.sdk;
 
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 
@@ -21,11 +22,14 @@ public class SdkClientUtils {
     /**
      * Unwraps the cause of a {@link CompletionException}. If the cause is a subclass of {@link RuntimeException}, rethrows the exception.
      * Otherwise wraps it in an {@link OpenSearchException}. Properly re-interrupts the thread on {@link InterruptedException}.
-     * @param e the CompletionException
-     * @return The wrapped cause of the completion exception if unchecked, otherwise an OpenSearchException wrapping it.
+     * @param throwable a throwable, expected to be a {@link CompletionException} or {@link CancellationException}.
+     * @return the cause of the completion exception or the throwable, directly if a {@link RuntimeException} or wrapped in an OpenSearchException otherwise.
      */
-    public static RuntimeException unwrapAndConvertToRuntime(CompletionException e) {
-        Throwable cause = e.getCause();
+    public static RuntimeException unwrapAndConvertToRuntime(Throwable throwable) {
+        // Unwrap completion exception or pass through other exceptions
+        Throwable cause = throwable instanceof CompletionException ? throwable.getCause() : throwable;
+        // Double-unwrap checked exceptions wrapped in ExecutionException
+        cause = getRethrownExecutionExceptionRootCause(cause);
         if (cause instanceof InterruptedException) {
             Thread.currentThread().interrupt();
         }
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -20,12 +20,12 @@ import org.opensearch.common.util.concurrent.UncategorizedExecutionException;
 public class SdkClientUtils {
 
     /**
-     * Unwraps the cause of a {@link CompletionException}. If the cause is a subclass of {@link RuntimeException}, rethrows the exception.
+     * Unwraps the cause of a {@link CompletionException}. If the cause is an {@link Exception}, rethrows the exception.
      * Otherwise wraps it in an {@link OpenSearchException}. Properly re-interrupts the thread on {@link InterruptedException}.
      * @param throwable a throwable, expected to be a {@link CompletionException} or {@link CancellationException}.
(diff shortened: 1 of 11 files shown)


common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
```diff
@@ -21,11 +22,14 @@ public class SdkClientUtils {
...
+     * @param throwable a throwable, expected to be a {@link CompletionException} or {@link CancellationException}.
+     * @return the cause of the completion exception or the throwable, directly if a {@link RuntimeException} or wrapped in an OpenSearchException otherwise.
      */
-    public static RuntimeException unwrapAndConvertToRuntime(CompletionException e) {
-        Throwable cause = e.getCause();
+    public static RuntimeException unwrapAndConvertToRuntime(Throwable throwable) {
+        // Unwrap completion exception or pass through other exceptions
+        Throwable cause = throwable instanceof CompletionException ? throwable.getCause() : throwable;
+        // Double-unwrap checked exceptions wrapped in ExecutionException
+        cause = getRethrownExecutionExceptionRootCause(cause);
```
dhrubo-os, line 32: I could be wrong, but I feel like it is tailored for specific exceptions and may not handle other custom wrapping scenarios well.

dhrubo-os: What do you think about combining all pieces together? What problem do you see here?

```
/**
     * Unwraps the cause of a {@link CompletionException}. If the cause is a subclass of {@link RuntimeException}, rethrows the exception.
     * Otherwise wraps it in an {@link OpenSearchException}. Properly re-interrupts the thread on {@link InterruptedException}.
     * @param throwable a throwable, expected to be a {@link CompletionException} or {@link CancellationException}.
     * @return the cause of the completion exception or the throwable, directly if a {@link RuntimeException} or wrapped in an OpenSearchException otherwise.
     */
    public static RuntimeException unwrapAndConvertToRuntime(Throwable throwable) {
        Throwable cause = getRootCause(throwable);
        if (cause instanceof InterruptedException) {
            Thread.currentThread().interrupt();
        }
        if (cause instanceof RuntimeException) {
            return (RuntimeException) cause;
        }
        return new OpenSearchException(cause);
    }

    /**
     * Gets the root cause of a throwable. This unwraps all nested causes.
     * @param throwable a throwable with possibly nested causes
     * @return the root cause of the throwable
     */
    public static Throwable getRootCause(Throwable throwable) {
        Throwable cause = throwable;
        while (cause.getCause() != null && cause != cause.getCause()) {
            cause = cause.getCause();
        }
        return handleSpecificWrappers(cause);
    }

    /**
     * Handle specific known exception wrappers to get to the original root cause.
     * @param throwable a throwable, potentially wrapped in known wrappers
     * @return the unwrapped root cause
     */
    public static Throwable handleSpecificWrappers(Throwable throwable) {
        if (throwable instanceof CompletionException || throwable instanceof ExecutionException) {
            return throwable.getCause();
        }
        if (throwable instanceof UncategorizedExecutionException && throwable.getCause() instanceof ExecutionException) {
            return throwable.getCause().getCause();
        }
        return throwable;
    }
    ```

dbwiddis: > I could be wrong, but I feel like it is tailored for specific exceptions and may not handle other custom wrapping scenarios well.

It's tailored specifically for the `throwable` in the `whenComplete()` handling for the `SdkClient` and the `actionGet()` implementations. 

My concern with your method is arbitrarily deep getCause fetching that ultimately results in a `Throwable` but we need at least an `Exception` for the `ActionListener.onFailure()` parameter.

I think I can improve the client handling to do one more level of unwrapping of a cause as long as it's an `Exception`.

dbwiddis: I'm really happy with the current state of this PR.  It turns the `throwable` from `whenComplete()` into an Exception, accounting for unwrapping both a `CompletionException` and possible nested wrapping of checked exceptions in an `ExecutionException`.    It will almost always return the original `Exception` unless someone decides to start throwing `Error` in code somewhere, which will get wrapped.

It preserves all the original test class tests.

But it allows for directly failing the ActionListeners without having to check the class type.

dhrubo-os: >someone decides to start throwing Error in code somewhere, which will get wrapped.

You mean something like [this](https://github.com/search?q=repo%3Aopensearch-project%2Fml-commons+%22throw+new%22&type=code)?

dbwiddis: > You mean something like [this](https://github.com/search?q=repo%3Aopensearch-project%2Fml-commons+%22throw+new%22&type=code)?

Nope, all those are exceptions. I'm literally talking about the Java class [`Error`](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/lang/Error.html) which is the only `Throwable` subclass that is not an `Exception`.  

So if someone decides to `throw new IOError(blah)` then that will get wrapped as the cause of an `OpenSearchException`.  

There are [a few of those](https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch%20IOError&type=code) but I don't think we'll be seeing these in metadata handling, given the definition is "An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch."

dhrubo-os: I see, thanks for the clarification. Approving the PR now.

dhrubo-os (approved):

Issue #2581, opened 2024-06-20
[FEATURE] Re-enable tests with mockStatic in MLEngineTest to increase test coverage


chishui:
**Is your feature request related to a problem?**
Some tests in MLEngineTest are marked as `@Ignore` because they contain mockStatic which would throw exception due to the reason described here: https://github.com/opensearch-project/OpenSearch/issues/14420

**What solution would you like?**
Explicitly set inline mock maker when use mockStatic to mock a static function, then remove `@Ignore` and `TODO` to reenable those tests.

**What alternatives have you considered?**
A fix from OpenSearch core like described here https://github.com/opensearch-project/OpenSearch/issues/14420 could work for all OpenSearch repos which depend on OpenSearch core, but core maintainer proposed a refactoring of the static class which can be risky for the code here.

**Do you have any additional context?**
Other tests which try to mock static functions can do the same, but try to think about refactor code whenever possible to avoid mocking static functions.

chishui:
close as it's completed

PR #2582, opened 2024-06-20
Enable tests with mockStatic in MLEngineTest

chishui:
### Description
Enable tests with mockStatic in MLEngineTest
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2581
 
### Check List
  - [x] All tests pass

### DIFF
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
@@ -48,6 +48,7 @@ import org.opensearch.ml.engine.algorithms.regression.LinearRegression;
 import org.opensearch.ml.engine.encryptor.Encryptor;
 import org.opensearch.ml.engine.encryptor.EncryptorImpl;
 
+// TODO: refactor MLEngineClassLoader's static functions to avoid mockStatic
 public class MLEngineTest extends MLStaticMockBase {
     @Rule
     public ExpectedException exceptionRule = ExpectedException.none();
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
@@ -23,10 +23,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
-import org.mockito.MockSettings;
 import org.mockito.MockedStatic;
-import org.mockito.Mockito;
-import org.mockito.internal.creation.MockSettingsImpl;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.xcontent.XContentBuilder;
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
```diff
@@ -55,11 +56,14 @@ public class MLEngineTest {
     public ExpectedException exceptionRule = ExpectedException.none();
 
     private MLEngine mlEngine;
+    private MockSettings mockSettingsWithInlineMockMaker;
 
     @Before
     public void setUp() {
         Encryptor encryptor = new EncryptorImpl("m+dWmfmnNRiNlOdej/QelEkvMTyH//frS2TBeS2BP4w=");
         mlEngine = new MLEngine(Path.of("/tmp/test" + UUID.randomUUID()), encryptor);
+        mockSettingsWithInlineMockMaker = new MockSettingsImpl()
```
zane-neo, line 65: Is it possible to put this in a configuration file and all test files can use mockStatic method without the boilerplate code?

dhrubo-os:
According to [this conversation](https://github.com/opensearch-project/OpenSearch/issues/14420), seems like mocking static method is not encouraged. Why are we still mocking the static method then?

chishui:
> According to [this conversation](https://github.com/opensearch-project/OpenSearch/issues/14420), seems like mocking static method is not encouraged. Why are we still mocking the static method then?

It's not about mocking static, it's actually about static functions themself. Using static functions are not following best practices. However, there are many static functions already out there lacking of test coverages. It's not feasible to require a refactoring of code whenever we run into them. I feel it's much more practical to increase test coverage and increase the reliability of our code than refactoring static functions. But if someone is trying to introduce new static functions, let's try to require to see if we have other way of implementation.

ml-algorithms/src/test/java/org/opensearch/ml/engine/MLEngineTest.java
chishui: OpenSearch-core overrides the default mock maker through mockito-extension which limits the use of mockStatic. We can provide our own mockito-extension, but it'll impact both mock and mockStatic.

zane-neo: What impact will it be for using our own mockito-extension?

chishui: > if several mockito-extensions/org.mockito.plugins.MockMaker files exists in the classpath Mockito will only use the first returned by the standard [ClassLoader.getResource(java.lang.String)](http://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html?is-external=true#getResource(java.lang.String)) mechanism.

https://site.mockito.org/javadoc/current/org/mockito/plugins/MockMaker.html

So technically, if a library depends on both ml-commons and opensearch-core (might not be a case for now), could possibly start to use ml-commons's mockito-extension which changes their behavior implicitly. I still feel ideally, the fix should be done in opensearch-core to support both normal mock maker and inline mock maker, but since maintainers don't want to do in core, and mockStatic is still rare case, I feel we can just make mockStatic work in this way.

zane-neo: Ok, is it true that creating an abstract class like `StaticMockSupportTest` and a method like `myMockStatic`, then all the sub-classes of it can use myMockStatic method directly?

chishui: Made changes

zane-neo (dismissed):

chishui:
@dhrubo-os could you take a look again? Thanks

dhrubo-os:
Yeah, I do have concerns.  I also agree to the conversation with the other thread: `Credentials instance injectable`. 
Similarly we need to improve code which is under test injectable.
So I would suggest to look for the long term solution, rather than short team solution.
Two cons I see:
Introducing mockStatic in the code can encourage other developers to use more static methods in their design.
Tests that rely on static mocks can become brittle. Any change in the static method's behavior or signature can break multiple tests, leading to higher maintenance costs.

As that test was ignored for a while, I don't see enough urgency to solve this with a short term fix.

dhrubo-os (approved):

zane-neo (approved):

zane-neo:
The failure IT seems not caused by the PR, merging it.

PR #2585, opened 2024-06-20
[Feature/multi_tenancy] Migrate Model APIs using SDKClient

dbwiddis:
### Description

Migrates Model APIs

 - [X] Get
 - [x] Update
 - [x] Delete
 - [x] Search
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
@@ -9,6 +9,7 @@ import static org.opensearch.ml.common.CommonValue.ML_MODEL_GROUP_INDEX;
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.common.FunctionName.REMOTE;
 import static org.opensearch.ml.common.FunctionName.TEXT_EMBEDDING;
+import static org.opensearch.ml.plugin.MachineLearningPlugin.GENERAL_THREAD_POOL;
 import static org.opensearch.ml.settings.MLCommonsSettings.ML_COMMONS_TRUSTED_CONNECTOR_ENDPOINTS_REGEX;
 
 import java.io.IOException;
@@ -23,6 +24,7 @@ import java.util.Objects;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.DocWriteResponse;
+import org.opensearch.action.DocWriteResponse.Result;
 import org.opensearch.action.FailedNodeException;
 import org.opensearch.action.get.GetResponse;
 import org.opensearch.action.support.ActionFilters;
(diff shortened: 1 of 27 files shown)


plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
```diff
@@ -593,17 +613,27 @@ private UpdateRequest createUpdateModelGroupRequest(
     ) {
         modelGroupSourceMap.put(MLModelGroup.LATEST_VERSION_FIELD, updatedVersion);
         modelGroupSourceMap.put(MLModelGroup.LAST_UPDATED_TIME_FIELD, Instant.now().toEpochMilli());
-        UpdateRequest updateModelGroupRequest = new UpdateRequest();
-
+        /* Old code here. TODO investigate if we need to add seqNo and primaryTerm to data object request
+        UpdateRequest updateModelGroupRequest = new UpdateRequest();
```
dhrubo-os, line 617: I think we added this for VersionConflictEngineException. This occurs when maybe there are two updates happening at once. So I assume you are planning to have a discussion on this tomorrow, right?

plugin/src/main/java/org/opensearch/ml/helper/ModelAccessControlHelper.java
```diff
@@ -143,6 +151,90 @@ public void checkModelGroupPermission(MLModelGroup mlModelGroup, User user, Acti
         }
     }
 
+    public void validateModelGroupAccess(
```
dhrubo-os, line 154: I believe I worked on this method before. But somehow I don't see it in the feature branch. Not sure what happened!

plugin/src/main/java/org/opensearch/ml/sdkclient/LocalClusterIndicesClient.java
```diff
@@ -158,6 +136,10 @@ public CompletionStage<UpdateDataObjectResponse> updateDataObjectAsync(UpdateDat
                         new UpdateRequest(request.index(), request.id()).doc(request.dataObject().toXContent(sourceBuilder, EMPTY_PARAMS))
                     )
                     .actionGet();
+                if (updateResponse == null) {
+                    log.info("Null UpdateResponse");
+                    return new UpdateDataObjectResponse.Builder().id(request.id()).parser(null).build();
```
dhrubo-os, line 141: Is this expected? Should we throw error?

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/action/models/UpdateModelTransportAction.java
dbwiddis: Yes, we need some version of this and will address this in another PR.

plugin/src/main/java/org/opensearch/ml/sdkclient/LocalClusterIndicesClient.java
dbwiddis: This (and a similar one for Get) I kept to keep existing behavior/tests.  I spent nearly an hour digging through the NodeClient code to see when this might happen and do not see how we would ever get a null response however... other than as generated in tests, but the code passes along the null faithfully.

PR #2587, opened 2024-06-21
[Feature/multi_tenancy] Update GetDataObjectResponse to include full GetResponse parser

dbwiddis:
### Description

Some APIs (get model group) require the full `GetResponse` object from the API.

This changes the GetDataObjectResponse to use the full `GetResponse` as the `parser`.  Since there will always be a response, there is no longer a need for `Optional`, and the `isExists()` checks are available again.

It retains the source map as-is.

This parallels the implementation in the SearchDataObjectResponse (and probably would have done it that way if I'd done search first).

This allows for even closer-to-original migration of code expecting a GetResponse.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
@@ -5,18 +5,21 @@
 
 package org.opensearch.ml.action.model_group;
 
+import static org.opensearch.common.xcontent.json.JsonXContent.jsonXContent;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_GROUP_INDEX;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.GENERAL_THREAD_POOL;
 
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
+import org.opensearch.action.get.GetResponse;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.Strings;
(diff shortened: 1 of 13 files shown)


plugin/src/main/java/org/opensearch/ml/action/model_group/GetModelGroupTransportAction.java
```diff
@@ -134,35 +137,74 @@ private void handleThrowable(Throwable throwable, String modelGroupId, ActionLis
         }
     }
 
+    /*
```
dbwiddis, line 140: oops, had this in for reference, will remove in push shortly

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -110,14 +110,34 @@ public CompletionStage<GetDataObjectResponse> getDataObjectAsync(GetDataObjectRe
...
+                    found = true;
+                    source = getItemResponse.item().get(SOURCE).s();
                 }
-
-                String source = getItemResponse.item().get(SOURCE).s();
+                String simulatedGetResponse = "{\"_index\":\""
+                    + request.index()
+                    + "\",\"_id\":\""
+                    + request.id()
+                    + "\",\"_version\":1,\"_seq_no\":-2,\"_primary_term\":0,\"found\":"
```
dhrubo-os, line 126: Are all these values constant for every cases? Like version will be 1 always?

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
```diff
@@ -73,17 +68,19 @@
...
+    private OpenSearchClient openSearchClient;
+    private JsonpMapper mapper;
+
     /**
      * Instantiate this object with an OpenSearch Java client.
      * @param openSearchClient The client to wrap
      */
     public RemoteClusterIndicesClient(OpenSearchClient openSearchClient) {
         this.openSearchClient = openSearchClient;
+        this.mapper = openSearchClient._transport().jsonpMapper();
```
arjunkumargiri, line 83: is `_trasport()` internal method? JsonMapper should rather be passed in as part of constructor

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
dbwiddis: Probably need @arjunkumargiri to chime in here; this falls into the same category as "shard" information in DDB which is irrelevant, but I haven't been interacting with the DDB implementation enough to know what (if any) is relevant.

Otherwise best I can do is add a TODO to revisit this later.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -110,14 +110,34 @@ public CompletionStage<GetDataObjectResponse> getDataObjectAsync(GetDataObjectRe
...
-                    return new GetDataObjectResponse.Builder().id(request.id()).parser(Optional.empty()).build();
+                    found = false;
+                    source = null;
+                } else {
+                    found = true;
+                    source = getItemResponse.item().get(SOURCE).s();
                 }
-
-                String source = getItemResponse.item().get(SOURCE).s();
+                String simulatedGetResponse = "{\"_index\":\""
```
arjunkumargiri, line 122: This approach tightly couples SDKClient to the opensearch specific schema. This is resulting in non-opensearch clients building mimicking opensearch response which will not be flexible. 

Rather than modifying SDKClient can model group action fetch required field from parser?

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
arjunkumargiri: This would be a concern if plugins depends upon versions for any operation. It is preferred if the SDKClient interface hides all opensearch specific details and provides a generic CRUD+search interface.

dbwiddis: > It is preferred if the SDKClient interface hides all opensearch specific details

What do we do with OpenSearch customers who are expecting this information and may already have automation relying on the GetResponse API behavior that has been there for years?

I need a really strong argument for taking away information here.

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
dbwiddis: It's a `public` method on the `ApiClient` class:

https://github.com/opensearch-project/opensearch-java/blob/3af2d776903123c336686ffbfaafc5503f949b5b/java-client/src/main/java/org/opensearch/client/ApiClient.java#L66-L68

It's specifically listed exactly like this in the opensearch-java documentation:
https://github.com/opensearch-project/opensearch-java/blob/3af2d776903123c336686ffbfaafc5503f949b5b/guides/generic.md?plain=1#L81

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
dbwiddis: > Are all these values constant for every cases? Like version will be 1 always?

Looking more deeply I can just omit these from the strings and it will go with the defaults, which match what was there except version is -1.

dhrubo-os (approved):

PR #2590, opened 2024-06-24
SDK DDB client update and search interface implementation

arjunkumargiri:
### Description
SDK DDB client update and search interface implementation. 
* Added implementation of DDB update
* Search requests are routed to search SDK client
* Update create and search to write as json attribute instead of a string under source attribute.
 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
+++ b/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
@@ -95,7 +95,7 @@ public class DDBOpenSearchClient implements SdkClient {
             String source = Strings.toString(MediaTypeRegistry.JSON, request.dataObject());
             try {
                 JsonNode jsonNode = OBJECT_MAPPER.readTree(source);
-                Map<String, AttributeValue> item = JsonTransformer.convertJsonObjectToItem(jsonNode);
+                Map<String, AttributeValue> item = JsonTransformer.convertJsonObjectToDDBAttributeMap(jsonNode);
                 item.put(HASH_KEY, AttributeValue.builder().s(tenantId).build());
                 item.put(RANGE_KEY, AttributeValue.builder().s(id).build());
                 final PutItemRequest putItemRequest = PutItemRequest.builder().tableName(tableName).item(item).build();
@@ -136,7 +136,7 @@ public class DDBOpenSearchClient implements SdkClient {
                     sourceObject = null;
                 } else {
                     found = true;
-                    sourceObject = JsonTransformer.convertToObjectNode((getItemResponse.item()));
+                    sourceObject = JsonTransformer.convertDDBAttributeValueMapToObjectNode((getItemResponse.item()));
                 }
                 final String source = OBJECT_MAPPER.writeValueAsString(sourceObject);
                 String simulatedGetResponse = "{\"_index\":\""
@@ -177,7 +177,7 @@ public class DDBOpenSearchClient implements SdkClient {
             try {
                 String source = Strings.toString(MediaTypeRegistry.JSON, request.dataObject());
                 JsonNode jsonNode = OBJECT_MAPPER.readTree(source);
-                Map<String, AttributeValue> updateItem = JsonTransformer.convertJsonObjectToItem(jsonNode);
+                Map<String, AttributeValue> updateItem = JsonTransformer.convertJsonObjectToDDBAttributeMap(jsonNode);
                 updateItem.put(HASH_KEY, AttributeValue.builder().s(tenantId).build());
                 updateItem.put(RANGE_KEY, AttributeValue.builder().s(request.id()).build());
                 UpdateItemRequest updateItemRequest = UpdateItemRequest
--- a/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
+++ b/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
@@ -11,10 +11,6 @@ package org.opensearch.ml.sdkclient;
 import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -30,6 +26,7 @@ import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.ml.sdkclient.util.JsonTransformer;
 import org.opensearch.sdk.DeleteDataObjectRequest;
 import org.opensearch.sdk.DeleteDataObjectResponse;
 import org.opensearch.sdk.GetDataObjectRequest;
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -155,16 +200,140 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
         }), executor);
     }
 
+    /**
+     * DDB data needs to be synced with opensearch cluster. {@link RemoteClusterIndicesClient} will then be used to
+     * search data in opensearch cluster.
```
dbwiddis, line 205: For my education, what's happening here?  

Are we writing to DDB and reading from OS for all "fetch" operations or are we only syncing up for search?

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -155,16 +200,140 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
...
-        return null;
+        return this.remoteClusterIndicesClient.searchDataObjectAsync(request, executor);
     }
 
     private String getTableName(String index) {
         // Table name will be same as index name. As DDB table name does not support dot(.)
         // it will be removed from name.
         return index.replaceAll("\\.", "");
     }
+
```
dbwiddis, line 221: All these static methods should probably be in a util class, in which case they might be public anyway.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -62,7 +62,7 @@ public SdkClientModule() {
     SdkClientModule(String remoteMetadataType, String remoteMetadataEndpoint, String region) {
         this.remoteMetadataType = remoteMetadataType;
         this.remoteMetadataEndpoint = remoteMetadataEndpoint;
-        this.region = region;
+        this.region = region == null ? "us-west-2" : region;
```
dbwiddis, line 65: Why us-west-2?  Should this be a configuration setting we can read from?

dbwiddis (commented):
LGTM with some comments.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
arjunkumargiri: Good suggestion. Will update.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
arjunkumargiri: DDBOpenSearchClient will make use of DDB for CRUD operations. Data written to DDB will be synced with OpenSearch using offline integration between DDB and opensearch. SDKClient will then rely on opensearch only for search operations.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -38,30 +42,41 @@
...
     private static final String DEFAULT_TENANT = "DEFAULT_TENANT";
 
     private static final String HASH_KEY = "tenant_id";
     private static final String RANGE_KEY = "id";
-    private static final String SOURCE = "source";
 
     private DynamoDbClient dynamoDbClient;
+    private RemoteClusterIndicesClient remoteClusterIndicesClient;
+
+    public DDBOpenSearchClient(DynamoDbClient dynamoDbClient, RemoteClusterIndicesClient remoteClusterIndicesClient) {
```
dhrubo-os, line 76: Can we add java doc? Seems like `remoteClusterIndicesClient` will only be used for search operation? So if customer doesn't need the search functionality then they can just send null for `remoteClusterIndicesClient`?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientModule.java
```diff
@@ -80,7 +80,8 @@ protected void configure() {
                 return;
             case AWS_DYNAMO_DB:
                 log.info("Using dynamo DB as metadata store");
-                bind(SdkClient.class).toInstance(new DDBOpenSearchClient(createDynamoDbClient()));
+                bind(SdkClient.class)
+                    .toInstance(new DDBOpenSearchClient(createDynamoDbClient(), new RemoteClusterIndicesClient(createOpenSearchClient())));
```
Zhangxunmt, line 84: When you create a DDB client, do you see any sdk credential loading failures due to security issues? Last year when we tried to use aws sdk, the credential chain provider needs to read a file on the disk that is restricted access by OS so the SDK cannot be initialized. Wondering if this is a bug already fixed this year,

dbwiddis: @Zhangxunmt I was struggling with exactly that when trying to configure this client.

I was able to do a POC using AWS creds [here](https://github.com/opensearch-project/flow-framework/pull/631) and there were quite a few settings I had to add to [plugin-security.policy](https://github.com/opensearch-project/flow-framework/pull/631/files#diff-214b5d6612e51e2474aafa928918b67b571bfa31d6f4e398248840e5437714e9) and even that I needed to hard code a root path that I had permission to.

plugin/src/main/java/org/opensearch/ml/sdkclient/util/JsonTransformer.java
```diff
@@ -0,0 +1,143 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.annotations.VisibleForTesting;
+
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+public class JsonTransformer {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    @VisibleForTesting
+    public static Map<String, AttributeValue> convertJsonObjectToItem(JsonNode jsonNode) {
```
dhrubo-os, line 30: `convertJsonObjectToDDbAttributeMap`? as this method is only taking care of Dynamodb? Same goes for the other method name.

dhrubo-os: Rename the class too? `JsonTransformer` sounds too generic in this case.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -76,16 +93,18 @@ public CompletionStage<PutDataObjectResponse> putDataObjectAsync(PutDataObjectRe
...
+                item.put(HASH_KEY, AttributeValue.builder().s(tenantId).build());
+                item.put(RANGE_KEY, AttributeValue.builder().s(id).build());
+                final PutItemRequest putItemRequest = PutItemRequest.builder().tableName(tableName).item(item).build();
 
-            dynamoDbClient.putItem(putItemRequest);
-            return new PutDataObjectResponse.Builder().id(id).created(true).build();
+                dynamoDbClient.putItem(putItemRequest);
+                return new PutDataObjectResponse.Builder().id(id).created(true).build();
+            } catch (IOException e) {
+                throw new OpenSearchStatusException("Failed to parse data object  " + request.id(), RestStatus.BAD_REQUEST);
```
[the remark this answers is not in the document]
arjunkumargiri: DDB would through runtime exception which will not be wrapped.

plugin/src/main/java/org/opensearch/ml/sdkclient/util/JsonTransformer.java
arjunkumargiri: Will update method name, however would like to keep the name generic to support other json transformation functions in future

dhrubo-os (approved):

PR #2591, opened 2024-06-26
Agent Create + Delete + Get for multi-tenancy including sdkclient interface implementation

dhrubo-os:
### Description
[Agent Create + Delete + Get for multi-tenancy including sdkclient interface implementation]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
+++ b/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
@@ -460,7 +460,23 @@ public interface MachineLearningClient {
         return actionFuture;
     }
 
-    void deleteAgent(String agentId, ActionListener<DeleteResponse> listener);
+    /**
+     * Delete agent
+     * @param agentId The id of the agent to delete
+     * @param listener a listener to be notified of the result
+     */
+    default void deleteAgent(String agentId, ActionListener<DeleteResponse> listener) {
+        PlainActionFuture<DeleteResponse> actionFuture = PlainActionFuture.newFuture();
+        deleteAgent(agentId, null, actionFuture);
+    }
+
+    /**
+     * Delete agent
+     * @param agentId The id of the agent to delete
+     * @param tenantId the tenant id. This is necessary for multi-tenancy.
+     * @param listener a listener to be notified of the result
+     */
+    void deleteAgent(String agentId, String tenantId, ActionListener<DeleteResponse> listener);
 
     /**
      * Get a list of ToolMetadata and return ActionFuture.
--- a/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
@@ -9,6 +9,7 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.Setter;
 import org.opensearch.Version;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -35,7 +36,7 @@ import static org.opensearch.ml.common.CommonValue.TENANT_ID;
 import static org.opensearch.ml.common.utils.StringUtils.getParameterMap;
 
 @EqualsAndHashCode
-@Data
+@Getter
 public class MLAgent implements ToXContentObject, Writeable {
     public static final String AGENT_NAME_FIELD = "name";
     public static final String AGENT_TYPE_FIELD = "type";
@@ -52,19 +53,19 @@ public class MLAgent implements ToXContentObject, Writeable {
 
     private static final Version MINIMAL_SUPPORTED_VERSION_FOR_HIDDEN_AGENT = CommonValue.VERSION_2_13_0;
 
-    private String name;
-    private String type;
-    private String description;
+    private final String name;
+    private final String type;
+    private final String description;
     private LLMSpec llm;
     private List<MLToolSpec> tools;
     private Map<String, String> parameters;
     private MLMemorySpec memory;
 
-    private Instant createdTime;
-    private Instant lastUpdateTime;
-    private String appType;
+    private final Instant createdTime;
+    private final Instant lastUpdateTime;
+    private final String appType;
     private Boolean isHidden;
-    private String tenantId;
+    private final String tenantId;
 
     @Builder(toBuilder = true)
     public MLAgent(String name,
--- a/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.common.agent;
 
 import lombok.Builder;
+import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.opensearch.Version;
@@ -30,10 +31,11 @@ import java.util.Optional;
 import java.util.Set;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID;
 import static org.opensearch.ml.common.utils.StringUtils.getParameterMap;
 
 @EqualsAndHashCode
-@Getter
+@Data
 public class MLAgent implements ToXContentObject, Writeable {
     public static final String AGENT_NAME_FIELD = "name";
     public static final String AGENT_TYPE_FIELD = "type";
@@ -62,6 +64,7 @@ public class MLAgent implements ToXContentObject, Writeable {
     private Instant lastUpdateTime;
     private String appType;
     private Boolean isHidden;
+    private String tenantId;
 
     @Builder(toBuilder = true)
     public MLAgent(String name,
@@ -74,7 +77,8 @@ public class MLAgent implements ToXContentObject, Writeable {
                    Instant createdTime,
                    Instant lastUpdateTime,
                    String appType,
-                   Boolean isHidden) {
+                   Boolean isHidden,
+                   String tenantId) {
         this.name = name;
         this.type = type;
         this.description = description;
@@ -87,6 +91,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         this.appType = appType;
         // is_hidden field isn't going to be set by user. It will be set by the code.
         this.isHidden = isHidden;
+        this.tenantId = tenantId;
         validate();
     }
 
@@ -152,6 +157,8 @@ public class MLAgent implements ToXContentObject, Writeable {
         if (streamInputVersion.onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_HIDDEN_AGENT)) {
             isHidden = input.readOptionalBoolean();
         }
+        // TODO: check BWC later
+        tenantId = input.readOptionalString();
         validate();
     }
 
@@ -175,7 +182,7 @@ public class MLAgent implements ToXContentObject, Writeable {
         } else {
             out.writeBoolean(false);
         }
-        if (parameters != null && parameters.size() > 0) {
+        if (parameters != null && !parameters.isEmpty()) {
             out.writeBoolean(true);
             out.writeMap(parameters, StreamOutput::writeString, StreamOutput::writeOptionalString);
         } else {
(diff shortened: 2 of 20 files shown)


client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -460,7 +460,23 @@ default ActionFuture<DeleteResponse> deleteAgent(String agentId) {
...
+        deleteAgent(agentId, null, actionFuture);
+    }
+
+    /**
+     * Delete agent
+     * @param agentId The id of the agent to delete
+     * @param tenantId the tenant id. This is necessary for multi-tenancy.
+     * @param listener a listener to be notified of the result
+     */
+    void deleteAgent(String agentId, String tenantId, ActionListener<DeleteResponse> listener);
```
ylwu-amzn, line 479: Do we need to expose transport API for delete agent for multi-tenancy ?

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -30,10 +31,11 @@
 import java.util.Set;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID;
 import static org.opensearch.ml.common.utils.StringUtils.getParameterMap;
 
 @EqualsAndHashCode
-@Getter
+@Data
```
ylwu-amzn, line 38: Add this will expose all parameters' set method, that seems not necessary/safe. For example, we don't want anyone can update agent type/appType once agent created

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentDeleteRequest.java
```diff
@@ -21,24 +22,31 @@
 
 import static org.opensearch.action.ValidateActions.addValidationError;
 
+@Data
```
ylwu-amzn, line 25: Seems not necessary to expose set method

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -39,75 +52,123 @@
...
+
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
+                    context.restore();
+                    log.debug("Completed Get Agent Request, Agent id:{}", agentId);
+                    if (throwable != null) {
+                        Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
+                        if (cause instanceof IndexNotFoundException) {
+                            log.error("Failed to get Agent index", cause);
```
ylwu-amzn, line 105: This error log seems not helpful, because this is expected and not server side error

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -39,75 +52,123 @@
...
-                                .onFailure(
-                                    new OpenSearchStatusException(
-                                        "User doesn't have privilege to perform this operation on this hidden agent",
-                                        RestStatus.FORBIDDEN
-                                    )
-                                );
+
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
+                .whenComplete((r, throwable) -> {
```
ylwu-amzn, line 99: `CompletionStage` whenComplete method is non-blocking, any difference with ActionListener non-blocking ?

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -39,75 +52,123 @@
...
+                                                actionListener.onFailure(e);
+                                            }
+                                        }
+                                    }
+                                } catch (Exception e) {
+                                    log.error("Failed to parse ml agent {}", agentId, e);
+                                    actionListener.onFailure(e);
+                                }
+                            } else {
+                                actionListener.onFailure(new OpenSearchStatusException("Fail to find task", RestStatus.NOT_FOUND));
```
ylwu-amzn, line 162: task -> agent ?

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -39,75 +52,123 @@
...
+                            }
+                        } catch (Exception e) {
+                            actionListener.onFailure(e);
```
ylwu-amzn, line 165: print error log?

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -39,75 +52,123 @@
...
+                            GetResponse gr = GetResponse.fromXContent(r.parser());
+                            if (gr.isExists()) {
+                                try (
+                                    XContentParser parser = jsonXContent
+                                        .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, gr.getSourceAsString())
+                                ) {
+                                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                                    MLAgent mlAgent = MLAgent.parse(parser);
+                                    if (TenantAwareHelper
+                                        .validateTenantResource(mlFeatureEnabledSetting, tenantId, mlAgent.getTenantId(), actionListener)) {
```
ylwu-amzn, line 122: `validateTenantResource` method responsibility looks not clear, it should validate tenant , but actually it also covers send back the validation failure through action listener. 
It looks ok for now, but such combined/unclear responsibility make it hard to use this method if there is not actionlistener.

ylwu-amzn: It also add some effort for code review. When I review this part, I doubt why there is no `else` block matching this `if`. So I have to read the code of `validateTenantResource` method to learn why. This impacts code readability.

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -116,4 +177,22 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+        String agentId,
+        ActionListener<DeleteResponse> actionListener
+    ) {
+        if (throwable != null) {
+            Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
+            log.error("Failed to delete ML Agent : {}", agentId, cause);
+            actionListener.onFailure(cause);
+        } else {
+            log.info("Agent deletion result: {}, agent id: {}", response.deleted(), response.id());
+            DeleteResponse deleteResponse = new DeleteResponse(response.shardId(), response.id(), 0, 0, 0, response.deleted());
```
ylwu-amzn, line 193: This seems a breaking change by harding code `0, 0, 0`. Is it possible to record real values in `DeleteDataObjectResponse`, so we can get the real values just like `response.shardInfo()`

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
```diff
@@ -42,74 +51,106 @@
...
+                            actionListener.onFailure(new OpenSearchStatusException("Failed to get agent index", RestStatus.NOT_FOUND));
                         } else {
-                            actionListener.onResponse(MLAgentGetResponse.builder().mlAgent(mlAgent).build());
+                            log.error("Failed to get ML Agent {}", agentId, cause);
+                            actionListener.onFailure(cause);
+                        }
+                    } else {
+                        try {
+                            GetResponse gr = GetResponse.fromXContent(r.parser());
+                            if (gr.isExists()) {
```
arjunkumargiri, line 111: Exception handling and parsing `GetResponse` has duplicate code across `DeleteAgentTransportAction` and `FetchAgentTransportAction` class. Move it a to common utility.

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
dhrubo-os: @dbwiddis had also similar comment [here](https://github.com/opensearch-project/ml-commons/pull/2588#discussion_r1649758398).  Do you think renaming the method to `validateTenantIdandFailListenerIfInvalid` can help?

In ml-commons we have similar methods like this : [validateModelGroupAccess](https://github.com/opensearch-project/ml-commons/blob/f61ab171f7e36eee727b678ca0f26f46204ae891/plugin/src/main/java/org/opensearch/ml/helper/ModelAccessControlHelper.java#L77C17-L77C41), [validateConnectorAccess](https://github.com/opensearch-project/ml-commons/blob/f61ab171f7e36eee727b678ca0f26f46204ae891/plugin/src/main/java/org/opensearch/ml/helper/ConnectorAccessControlHelper.java#L62C17-L78)

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
dhrubo-os: Ideally request should land on Flow framework and flow framework will invoke this transport api.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
dhrubo-os: ok I'll add setter for tenantId, then?

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
dbwiddis: @ylwu-amzn yes, I've noted this and had discussion with @dhrubo-os and @arjunkumargiri ... will fix Delete, Update, and Get to process full response similar to #2587. 

More discussion here: https://github.com/opensearch-project/ml-commons/pull/2587/files#r1651490825

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
dbwiddis: > Move it a to common utility.

Not sure what is going to be most common here, as there's a lot of variation inside these blocks.

It's also much easier for this initial migration for review to literally copy/paste the existing (duplicate) contents.

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
dbwiddis: > is there anything else you want to add here?

That sums it up.  The async processing corresponding to the ActionListener is in the <foo>DataObjectAsync() calls, which internally complete the future.  The `whenComplete()` triggers only when that future has been completed so is essentially the same as an action listener callback and can process synchronously ( and in many cases just invokes the action listener.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
ylwu-amzn: Is it safe/necessary to set tenantId after agent created ? Can we add it to constructor method ?

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
dhrubo-os: Ok, I'll address this in another PR then. I have to touch multiple files in that case.

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
dhrubo-os: Yea, not only for agents, I saw there any other apis where get functionalities could be put into a common utility. I was thinking to address those in separate PRs.

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -102,7 +102,7 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
                     if (throwable != null) {
                         Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
                         if (cause instanceof IndexNotFoundException) {
-                            log.error("Failed to get Agent index", cause);
+                            log.info("Failed to get Agent index", cause);
```
ylwu-amzn, line 105: `log.info("Agent index not found");`

ylwu-amzn (approved):

PR #2595, opened 2024-06-27
Use full DocWriteResponse json for Put/Update/Delete responses

dbwiddis:
### Description

Updates the SDKClient to use the full IndexResponse, UpdateResponse, and DeleteResponse objects.

Parallel of similar work on GetResponse in #2587 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/model/MLModelGroupManagerTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/model/MLModelGroupManagerTests.java
@@ -10,7 +10,7 @@ import static org.mockito.ArgumentMatchers.isA;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-import static org.opensearch.action.DocWriteResponse.Result.CREATED;
+import static org.opensearch.ml.common.CommonValue.ML_MODEL_GROUP_INDEX;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.GENERAL_THREAD_POOL;
 import static org.opensearch.ml.plugin.MachineLearningPlugin.ML_THREAD_POOL_PREFIX;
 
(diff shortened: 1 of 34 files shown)


plugin/src/test/java/org/opensearch/ml/model/MLModelGroupManagerTests.java
```diff
@@ -198,7 +196,7 @@ public void test_ModelGroupNameNotUnique() throws IOException, InterruptedExcept
         CountDownLatch latch = new CountDownLatch(1);
         LatchedActionListener<String> latchedActionListener = new LatchedActionListener<>(actionListener, latch);
         mlModelGroupManager.createModelGroup(mlRegisterModelGroupInput, latchedActionListener);
-        latch.await();
+        latch.await(500, TimeUnit.MILLISECONDS);
```
dhrubo-os, line 199: Why do we need 500 now?

dhrubo-os: What happens after that? Will there by any chance for these tests to be flaky if timeout happens?

dbwiddis: The purpose for the latch is a race condition as the async stuff processes in multiple threads.  

The previous tests returned immediately in the same thread with a mock `.when()` response.  Now there is an async execution layer in between the call and the mock so there are two thread involved.

It should complete in sub-millisecond time in most cases, no more than a few milliseconds in nearly all cases.

Leaving everything as `latch.await()` with no timeout would frequently stall tests and have them not complete until the test suite itself killed the threads on timeout -- this worked on CI but was frustrating debugging local test failures when a test just kept running and running.

TLDR: we'll only hit the timeout if there's a bug in the code, and it will help get a stack trace faster to make debugging easier and save developers like me a ton of debugging time.  Worst case scenario if CI is running on a fully loaded machine and multithreaded tests take more than half a second to run, they'll fail flakily, but never pass when they should have failed.

dhrubo-os (approved):

PR #2602, opened 2024-07-01
Implement SdkClient in TransportRegisterModelAction

dbwiddis:
### Description

Implements needed methods in MLModelManager for the TransportRegisterModelAction when the model is remote.

Migrates all code calling non-sdk MLModelGroupManager methods and removes old method/test.

Also updates SDK request tests to include Tenant ID and creates the missing UpdateRequest test.
 
Still TODO:  Concurrency guarantee on update request.  Still considering the best way to do that, will follow up with another PR for that, probably tomorrow.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -140,8 +140,10 @@ import org.opensearch.ml.utils.MLExceptionUtils;
 import org.opensearch.ml.utils.MLNodeUtils;
 import org.opensearch.script.ScriptService;
 import org.opensearch.sdk.GetDataObjectRequest;
+import org.opensearch.sdk.PutDataObjectRequest;
 import org.opensearch.sdk.SdkClient;
 import org.opensearch.sdk.SdkClientUtils;
+import org.opensearch.sdk.UpdateDataObjectRequest;
 import org.opensearch.search.fetch.subphase.FetchSourceContext;
 import org.opensearch.threadpool.ThreadPool;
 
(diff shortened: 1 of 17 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -550,12 +582,11 @@ private void indexRemoteModel(
...
 
-                IndexRequest indexModelMetaRequest = new IndexRequest(ML_MODEL_INDEX);
-                if (registerModelInput.getIsHidden() != null && registerModelInput.getIsHidden()) {
-                    indexModelMetaRequest.id(modelName);
-                }
-                indexModelMetaRequest.source(mlModelMeta.toXContent(XContentBuilder.builder(JSON.xContent()), EMPTY_PARAMS));
-                indexModelMetaRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+                PutDataObjectRequest putModelMetaRequest = new PutDataObjectRequest.Builder()
+                    .index(ML_MODEL_INDEX)
+                    .id(Boolean.TRUE.equals(registerModelInput.getIsHidden()) ? modelName : null)
```
dhrubo-os, line 587: in this case model id will be null or auto generated in the index?

dbwiddis: Yes, if `id` is omitted in an index request it auto-generates a UUID.

This was an attempt to faithfully reproduce this.  Yoda expressions avoid the need for the null check and are much more readable!
```java
                if (registerModelInput.getIsHidden() != null && registerModelInput.getIsHidden()) {
                    indexModelMetaRequest.id(modelName);
                }
```

dhrubo-os (approved):

PR #2605, opened 2024-07-02
[Feature/multi_tenancy] Add ifSeqNo and ifPrimaryTerm for update concurrency

dbwiddis:
### Description

Tracks SeqNo and PrimaryTerm used for strong concurrency for updates.
 - Adds to update requests
 - Implements tracking of this version in DynamoDB

Still TODO (@arjunkumargiri  will do in another PR):
 - DDB: Initialize SeqNo to 0 for new documents, but increment on overwrite
 - DDB: Increment on successful update
 - DDB: Return -2 on delete
 
### Check List
  - [x] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
+++ b/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
@@ -55,6 +55,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import lombok.extern.log4j.Log4j2;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.services.dynamodb.model.ConditionalCheckFailedException;
 import software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
@@ -71,8 +72,9 @@ public class DDBOpenSearchClient implements SdkClient {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final String DEFAULT_TENANT = "DEFAULT_TENANT";
 
-    private static final String HASH_KEY = "tenant_id";
-    private static final String RANGE_KEY = "id";
+    private static final String HASH_KEY = "_tenant_id";
+    private static final String RANGE_KEY = "_id";
+    private static final String SEQ_NO_KEY = "_seq_no";
 
     private DynamoDbClient dynamoDbClient;
     private RemoteClusterIndicesClient remoteClusterIndicesClient;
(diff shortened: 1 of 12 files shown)


plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -227,7 +252,12 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
             .build();
         return CompletableFuture.supplyAsync(() -> AccessController.doPrivileged((PrivilegedAction<DeleteDataObjectResponse>) () -> {
             try {
+                // TODO need to return SEQ_NO here
+                // If document doesn't exist, increment and return highest seq no ever seen, but we would have to track seqNo here
+                // If document never existed, return -2 (unassigned) for seq no (probably what we have to do here)
+                // If document exists, increment and return SEQ_NO
```
arjunkumargiri, line 258: In case of delete, why does seq no be updated? can we safely delete document as long as seq no matches?

dbwiddis: @arjunkumargiri This is somewhat of an artifact of how "soft deletes" work in OpenSearch. 

When you call the Delete API, it only marks the document as deleted in the metadata but it doesn't actually remove it until the segment is merged.   It's actually still available in search results [unless you use the `refresh` parameter](https://opensearch.org/docs/latest/api-reference/document-apis/delete-document/).

So it still tracks the `seq_no` as a change to the document for replication/tracking purposes until the document actually goes away on a merge, at which point I assume that tracking disappears.

dhrubo-os:
May be not a part of this PR, but we need to address this [use case](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java#L126). @dbwiddis could you please track this action item in your end? Thanks.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
dhrubo-os: How can we distinguish between 'If the document doesn't exist' and 'If the document never existed'?

dbwiddis:
> May be not a part of this PR, but we need to address this [use case](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java#L126). @dbwiddis could you please track this action item in your end? Thanks.

I'm not clear what the action item is?  The code currently throws an exception and returns an error to the user; wouldn't it be exactly the same error?  Are you referring to per-tenant keys or something else?

dhrubo-os:
> > May be not a part of this PR, but we need to address this [use case](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java#L126). @dbwiddis could you please track this action item in your end? Thanks.
> 
> I'm not clear what the action item is? The code currently throws an exception and returns an error to the user; wouldn't it be exactly the same error? Are you referring to per-tenant keys or something else?

So currently [Optype is Index](https://github.com/opensearch-project/OpenSearch/blob/41fa0855faf1913b0f5334398578cd6e2307d853/server/src/main/java/org/opensearch/action/index/IndexRequest.java#L116) by default. But for config key creation, we want `OpType.CREATE` specifically. From current sdkClient interface we can't achieve this.

dbwiddis:
> we want `OpType.CREATE` specifically. From current sdkClient interface we can't achieve this.

Got it... I'm not sure we want to use "OpType" generaly.  "Upsert" is the standard language here.  We probably want to implement conditional writes on the DynamoDB version.

Let's align on a name for it... will track separately.

dhrubo-os (approved):

Issue #2560, opened 2024-06-17
[BUG] Flaky IT in ml-commons


zane-neo:
**What is the bug?**
Below is a flaky IT which fails time to time, we need to fix this: 
https://github.com/opensearch-project/ml-commons/actions/runs/9540553655/job/26292459753?pr=2500#step:6:3006

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

rbhavna:
Below is the exception we are seeing:
```
Suite: Test class org.opensearch.ml.tools.VisualizationsToolIT
  2> REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.tools.VisualizationsToolIT.testVisualizationNotFound" -Dtests.seed=B77447D33F3C293E -Dtests.security.manager=false -Dtests.locale=is-IS -Dtests.timezone=America/Anguilla -Druntime.java=11
  2> org.opensearch.client.ResponseException: method [DELETE], host [http://127.0.0.1:43309/], URI [/_plugins/_ml/models/29kFJJABf_wnGiLB9K81], status line [HTTP/1.1 400 Bad Request]
    {"error":{"root_cause":[{"type":"status_exception","reason":"Model cannot be deleted in deploying or deployed state. Try undeploy model first then delete"}],"type":"status_exception","reason":"Model cannot be deleted in deploying or deployed state. Try undeploy model first then delete"},"status":400}
        at __randomizedtesting.SeedInfo.seed([B77447D33F3C293E:7817E34E380F133E]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:385)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:355)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:330)
        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:179)

        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:152)
        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:141)
        at app//org.opensearch.ml.rest.MLCommonsRestTestCase.deleteModel(MLCommonsRestTestCase.java:653)
[...]

PR #2613, opened 2024-07-05
Fix model delete failed after model undeployed

zane-neo:
### Description
When deleting a model, we need to check the model status by querying the model index: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L108, which means when undepdeloying model response returns, we have to make sure the model status in index is updated to `DELETED`. 

But in UndeployModelTransportAction there's no such guarantee, when updating the model index: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java#L192, the response returns almost at the same time when the each node's undeploy finished: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/nodes/TransportNodesAction.java#L315. Which means updating index and response to client run in parallel, so this issue happens occasionally.

So we need a mechanism to guarantee only when the update model index is completed, the response to client get run by adding a countdownlatch to block the response client action when updating index. This is safe to block since this `newResponse` method runs in a threadpool instead of the event loop threads. I also changed the threadpool to model deploy threadpool which is dedicated for deploy/undeploy model instead of management threadpool.
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2419
https://github.com/opensearch-project/ml-commons/issues/2560
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
@@ -113,7 +113,7 @@ public class TransportUndeployModelAction extends
                     // when undeploy a undeployed model, the entry.getvalue() is null
                     if (entry.getValue() != null
                         && (!modelWorkNodesBeforeRemoval.containsKey(entry.getKey())
-                        || modelWorkNodesBeforeRemoval.get(entry.getKey()).length < entry.getValue().length)) {
+                            || modelWorkNodesBeforeRemoval.get(entry.getKey()).length < entry.getValue().length)) {
                         modelWorkNodesBeforeRemoval.put(entry.getKey(), entry.getValue());
                     }
                 }
@@ -202,7 +202,10 @@ public class TransportUndeployModelAction extends
             try {
                 boolean success = countDownLatch.await(1000, TimeUnit.MILLISECONDS);
                 if (!success) {
-                    log.error("Failed to update model state as undeployed in model index after waiting for 1 second, please check model status manually");
+                    log
+                        .error(
+                            "Failed to update model state as undeployed in model index after waiting for 1 second, please check model status manually"
+                        );
                 }
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
@@ -7,6 +7,7 @@ package org.opensearch.ml.action.undeploy;
 
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.common.CommonValue.UNDEPLOYED;
+import static org.opensearch.ml.plugin.MachineLearningPlugin.DEPLOY_THREAD_POOL;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -14,6 +15,8 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.opensearch.action.FailedNodeException;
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
Please rebase to take changes from #2627. Thanks.

Hailong-am:
what's the root cause?

xinyual (approved):

zane-neo:
> what's the root cause?

Added more details in the description.

Zhangxunmt:
Which flaky IT is fixed from this change?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -7,13 +7,16 @@
 
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 import static org.opensearch.ml.common.CommonValue.UNDEPLOYED;
+import static org.opensearch.ml.plugin.MachineLearningPlugin.DEPLOY_THREAD_POOL;
```
Zhangxunmt, line 10: Is it possible that multiple Undeploy actions blocking the Deploy_Thread_Pool will cause "Deploy" to fail?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -188,24 +183,35 @@ void processUndeployModelResponseAndUpdate(
...
             } else {
                 syncUpUndeployedModels(syncUpRequest);
-                listener.onResponse(undeployModelNodesResponse);
+                context.restore();
+                countDownLatch.countDown();
+            }
+        }
+        if (countDownLatch.getCount() != 0) {
+            try {
+                boolean success = countDownLatch.await(1000, TimeUnit.MILLISECONDS);
```
Zhangxunmt, line 203: Is this 1 second too large? In most cases the update index happen pretty fast.

zane-neo: This is a `at most` value, if the update index happens within 1s, the await will break immediately after it's done. If update takes more than 1s, then await waits at most 1s to break.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
zane-neo: Most of cases, the update index completes immediately which means there's no actual blocking here. In edge cases, if the update completes relatively slow, the requests will be queued. In a more edge case, the queue is full, then the deploy model will fail. But this usually mean the system is not in a correct status(updating index taking too long time). Also this failure is recoverable which doesn't seem a critical issue.

zane-neo:
> Which flaky IT is fixed from this change?

Check this issue: https://github.com/opensearch-project/ml-commons/issues/2560

ylwu-amzn:
@zane-neo I remember we have add immediate refresh policy . Why that can't work?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
ylwu-amzn: This will block current thread, which not recommended. We saw issues several times for such blocking thread code

zane-neo: If you're not blocking event loop thread, it's safe.

zane-neo:
> @zane-neo I remember we have add immediate refresh policy . Why that can't work?

Please check the description, simply put, this is a race condition issue that updating document and responding to user happens at the same time, no sequence guarantee.

zane-neo:
This issue still exists now since this PR not merged: https://github.com/opensearch-project/ml-commons/actions/runs/10245403756/job/28381967022#step:6:3426

zane-neo:
Checked again on the issue, this fix is not fixing the root cause. When using a script like below to reproduce the issue:
```
reg_res = client.post('localhost:9200/_plugins/_ml/models/_register', register_model_payload)
model_id = extract_model_id(reg_res)
deploy_res = client.post('localhost:9200/_plugins/_ml/models/{model_id}/_load')
undeploy_res = client.post('localhost:9200/_plugins/_ml/models/{model_id}/_unload')
delete_res = client.post('localhost:9200/_plugins/_ml/models/{model_id}')
```
This issue can be reproduce occasionally, but if adding `time.sleep(1)` after `deploy_res` returned, the issue can't be reproduced anymore.
The reason is deploy model is in an async approach and when client received the response, the model status in index is not guaranteed to `deployed`, this doesn't have a simple fix since a lot of files are involved in deploy flow.

This issue is not able to reproduced easily on local, so will add logs in the delete model action to clearly show the model status when it's not in a deletable status, and closing this PR now.

PR #2634, opened 2024-07-10
change disk circuit breaker to cluster settings

zane-neo:
### Description
change disk circuit breaker to cluster settings
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2639
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/breaker/ThresholdCircuitBreaker.java
+++ b/plugin/src/main/java/org/opensearch/ml/breaker/ThresholdCircuitBreaker.java
@@ -5,20 +5,19 @@
 
 package org.opensearch.ml.breaker;
 
+import lombok.Data;
+
 /**
  * An abstract class for all breakers with threshold.
  * @param <T> data type of threshold
  */
+@Data
 public abstract class ThresholdCircuitBreaker<T> implements CircuitBreaker {
 
     private T threshold;
 
     public ThresholdCircuitBreaker(T threshold) {
         this.threshold = threshold;
-    }
-
-    public T getThreshold() {
-        return threshold;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -77,6 +77,16 @@ private MLCommonsSettings() {}
...
 
+    public static final Setting<Integer> ML_COMMONS_DISK_FREE_SPACE_THRESHOLD = Setting
+        .intSetting(
+            "plugins.ml_commons.disk_free_space_threshold",
+            5,
+            0,
+            Integer.MAX_VALUE,
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
+        );
```
Hailong-am, line 88: nit: another alternative could be `ByteSizeValue` instead of int. it has more fine-grained control.

plugin/src/main/java/org/opensearch/ml/breaker/ThresholdCircuitBreaker.java
```diff
@@ -5,10 +5,13 @@
...
+import lombok.Data;
+
 /**
  * An abstract class for all breakers with threshold.
  * @param <T> data type of threshold
  */
+@Data
 public abstract class ThresholdCircuitBreaker<T> implements CircuitBreaker {
 
     private T threshold;
```
Hailong-am, line 17: should we add `volatile` modifier for this field?

zane-neo: It's better to add volatile so that each thread can read the up-to-date value of this.

Hailong-am (approved):

b4sjoo (approved):

PR #2635, opened 2024-07-10
Add application_type into get ConversationMeta API response

### LINKED ISSUE #2632: [BUG] Get conversation meta don't return application_type

current get memory api https://opensearch.org/docs/latest/ml-commons-plugin/api/memory-apis/get-memory/ will return response like below, it miss [application_type](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/conversation/ConversationalIndexConstants.java#L59) field.
```
{
  "memory_id": "gW8Aa40BfUsSoeNTvOKI",
  "create_time": "2024-02-02T18:07:06.887061463Z",
  "updated_time": "2024-02-02T19:01:32.121444968Z",
  "name": "Conversation for a RAG pipeline",
  "user": "admin"
}
```

mapping of system index `.plugins-ml-memory-meta`
```
{
  ".plugins-ml-memory-meta": {
    "mappings": {
      "_meta": {
        "schema_version": 1
      },
      "properties": {
        "application_type": {
          "type": "keyword"
        },
        "create_time": {
          "type": "date",
          "format": "strict_date_time||epoch_millis"
        },
        "name": {
          "type": "text"
        },
        "updated_time": {
          "type": "date",
          "format": "strict_date_time||epoch_millis"
        },
        "user": {
          "type": "keyword"
        }
      }
    }
  }
}
```

Hailong-am:
### Description
Add application_type into get ConversationMeta

An example
```
GET /_plugins/_ml/memory/xJkym5ABZevb6ngQphzz


{
  "memory_id": "xJkym5ABZevb6ngQphzz",
  "create_time": "2024-07-10T05:51:46.418785Z",
  "updated_time": "2024-07-10T05:52:06.657124Z",
  "name": "Summarise the alert 'Revenue is low'",
  "user": "admin",
  "application_type": "alert_analysis"
}

```
 
### Issues Resolved
#2632 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/test/java/org/opensearch/ml/common/conversation/ConversationMetaTests.java
+++ b/common/src/test/java/org/opensearch/ml/common/conversation/ConversationMetaTests.java
@@ -30,7 +30,7 @@ public class ConversationMetaTests {
     @Before
     public void setUp() {
         time = Instant.now();
-        conversationMeta = new ConversationMeta("test_id", time, time, "test_name", "admin");
+        conversationMeta = new ConversationMeta("test_id", time, time, "test_name", "admin", "testApp");
     }
 
     @Test
@@ -41,6 +41,7 @@ public class ConversationMetaTests {
         content.field(ConversationalIndexConstants.META_UPDATED_TIME_FIELD, time);
         content.field(ConversationalIndexConstants.META_NAME_FIELD, "meta name");
         content.field(ConversationalIndexConstants.USER_FIELD, "admin");
+        content.field(ConversationalIndexConstants.APPLICATION_TYPE_FIELD, "testApp");
         content.endObject();
 
         SearchHit[] hits = new SearchHit[1];
@@ -50,6 +51,7 @@ public class ConversationMetaTests {
         assertEquals(conversationMeta.getId(), "cId");
         assertEquals(conversationMeta.getName(), "meta name");
         assertEquals(conversationMeta.getUser(), "admin");
+        assertEquals(conversationMeta.getApplicationType(), "testApp");
     }
 
     @Test
@@ -63,12 +65,15 @@ public class ConversationMetaTests {
                         ConversationalIndexConstants.META_NAME_FIELD,
                         "meta name",
                         ConversationalIndexConstants.USER_FIELD,
-                        "admin"
+                        "admin",
+                        ConversationalIndexConstants.APPLICATION_TYPE_FIELD,
+                        "testApp"
                 );
         ConversationMeta conversationMeta = ConversationMeta.fromMap("test-conversation-meta", params);
         assertEquals(conversationMeta.getId(), "test-conversation-meta");
         assertEquals(conversationMeta.getName(), "meta name");
         assertEquals(conversationMeta.getUser(), "admin");
+        assertEquals(conversationMeta.getApplicationType(), "testApp");
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


common/src/test/java/org/opensearch/ml/common/conversation/ConversationMetaTests.java
```diff
@@ -30,7 +30,7 @@ public class ConversationMetaTests {
     @Before
     public void setUp() {
         time = Instant.now();
-        conversationMeta = new ConversationMeta("test_id", time, time, "test_name", "admin");
+        conversationMeta = new ConversationMeta("test_id", time, time, "test_name", "admin", "testApp");
```
chishui, line 33: Could you add a test case covering the situation when applicationType is null?

Zhangxunmt (commented):
Please ensure this is BWC with the Conversational Search, which doesn't require to input or return any application_type. This applicationType was not returned intentionally in the first place for 1)BWC and 2) disallow any end user to update it (it only saves into index and updated by Agents.)

common/src/test/java/org/opensearch/ml/common/conversation/ConversationMetaTests.java
Hailong-am: sure, applicationType null test case added

Hailong-am:
> Please ensure this is BWC with the Conversational Search, which doesn't require to input or return any application_type. This applicationType was not returned intentionally in the first place for 1)BWC and 2) disallow any end user to update it (it only saves into index and updated by Agents.)

1) BWC related logic added
2) current update conversation api don't have any restriction, do you think we should add a validation for update application type field? @Zhangxunmt

Hailong-am:
@HenryL27 @austintlee would you mind to take a look at this? does this impact to conversational search?

Hailong-am:
close this, we are going to add a new field

ylwu-amzn:
> current update conversation api don't have any restriction, do you think we should add a validation for update application type field? @Zhangxunmt

This seems a valid concern, @Zhangxunmt Can you enhance this part ?

> close this, we are going to add a new field

@Zhangxunmt has a proposal to add a general field to extension, please take a look if your proposal can solve this

Zhangxunmt:
@Hailong-am Currently only the "name" field in the Memory is allowed for updates so there is restrictions https://github.com/Zhangxunmt/ml-commons/blob/2.x/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java#L38. 

Do you have the PR open for adding a new field? When you add the new field, can you please use a general name with "flat_object" date type in the Mapping like the "additional_info"? Let's add a new field that can be reused by future use cases rather than a one-off thing. Also, please increase the Index Mapping version when you add new field.

Hailong-am:
> @Hailong-am Currently only the "name" field in the Memory is allowed for updates so there is restrictions https://github.com/Zhangxunmt/ml-commons/blob/2.x/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java#L38.
> 
> Do you have the PR open for adding a new field? When you add the new field, can you please use a general name with "flat_object" date type in the Mapping like the "additional_info"? Let's add a new field that can be reused by future use cases rather than a one-off thing. Also, please increase the Index Mapping version when you add new field.

@Zhangxunmt thanks for suggestion, a generic name would be nice for reuse and i will use `flat_object` type.  I have not raise the PR yet, will have it shortly today.

PR #2638, opened 2024-07-10
Refactor SdkClient to a concrete class with delegate implementation

dbwiddis:
### Description

Refactors `SdkClient` interface to a concrete class containing a delegate implementation.  This allows it to be injected in the `createComponents()` extension point and available to `@Inject` constructors.

Uses Cluster Settings to determine which implementation is generated by the SdkClientFactory.

Adds SSL/TLS and Basic Auth to the remote client.

Adds an AWS OpenSearch Service client.
  
### Check List
  - [x] All tests pass


plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * A class to create a {@link SdkClient} with implementation based on settings
+ */
+@Log4j2
+public class SdkClientFactory {
+
+    // TODO Move these settings keys to plugin getSettings implementation with proper prefixes
+    public static final String REMOTE_METADATA_TYPE = "REMOTE_METADATA_TYPE";
+    public static final String REMOTE_METADATA_ENDPOINT = "REMOTE_METADATA_ENDPOINT";
```
arjunkumargiri, line 51: Along with prefix, lets ensure that it follows settings naming convention.

common/src/main/java/org/opensearch/sdk/SdkClientImpl.java
```diff
@@ -0,0 +1,65 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+
+public abstract class SdkClientImpl {
```
arjunkumargiri, line 14: This should be an interface instead of abstract class

arjunkumargiri: Rename to SDKClientInterface?

common/src/main/java/org/opensearch/sdk/SdkClientImpl.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param executor the executor to use for asynchronous execution
+     * @return A completion stage encapsulating the response or exception
+     */
+    public abstract CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(SearchDataObjectRequest request, Executor executor);
+    
+    /**
+     * Get the underlying client implementation
+     * @return the client implementation
+     */
+    public abstract Object getImplementation();
```
arjunkumargiri, line 64: Why is this required? underlying implementation need not be exposed

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
dbwiddis: The TODO commend does include "with proper prefixes" :)

common/src/main/java/org/opensearch/sdk/SdkClientImpl.java
dbwiddis: At one point I thought I needed to add it for tests but then later added the "wrap" constructor for tests so it became obsolete.  It can go, I guess. I thought it might be useful.

common/src/main/java/org/opensearch/sdk/SdkClientImpl.java
dbwiddis: SdkClientInterface comes from the Department of Redundancy Department. :)

I did SdkClientDelegate.

common/src/test/java/org/opensearch/sdk/SdkClientTests.java
```diff
@@ -86,162 +87,163 @@ public CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(SearchDat
                 return CompletableFuture.completedFuture(searchResponse);
             }
         });
+        sdkClient = new SdkClient(sdkClientImpl);
```
dhrubo-os, line 90: I assume we can use `sdkClient` like this way in algorithms module as we can't import SDKClientFactory class to algorithm module?

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -654,7 +649,7 @@ public Collection<Object> createComponents(
             .getClusterSettings()
             .addSettingsUpdateConsumer(MLCommonsSettings.ML_COMMONS_RAG_PIPELINE_FEATURE_ENABLED, it -> ragSearchPipelineEnabled = it);
 
-        LocalClusterIndicesClient localClusterIndicesClient = new LocalClusterIndicesClient(client, xContentRegistry);
+        SdkClient sdkClient = SdkClientFactory.createSdkClient(client, xContentRegistry, settings);
```
dhrubo-os, line 652: Cool, so that means we don't need to do ?

```
ModulesBuilder modules = new ModulesBuilder();
modules.add(new SdkClientModule(client, xContentRegistry));

 injector = modules.createInjector();

 // Get the injected SdkClient instance from the injector
 SdkClient sdkClient = injector.getInstance(SdkClient.class);

File: common/src/main/java/org/opensearch/sdk/SdkClientSettings.java [hunk 6]
```diff
@@ -0,0 +1,36 @@
+/*
```
dhrubo-os, line 1: I assume these settings will eventually move to core, right?

Otherwise we usually keep our settings in one place: https://github.com/opensearch-project/ml-commons/blob/0b9708b831f87ac6eb35b5c3039c61afe9425d26/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java#L17

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
dbwiddis: Correct.  That created an injector that was different than the one OpenSearch processes with the list of objects from `createComponents()`.  But the clients still had their runtime class so those would still have been bound to the subclasses, and this really just made a new/different SdkClient available to the various other classes in this method.   We could have just instantiated an sdkClient without jumping through these hoops and gotten the same thing.

With the new code we have a concrete class `SdkClient` that is injected and available by its class key (which is what the separate module was doing) and hides the implementation as an internal field.   Not the most object-oriented way to do things, but it works around the injection headaches.

common/src/main/java/org/opensearch/sdk/SdkClientSettings.java
dbwiddis: Yes, these are intended to be moved to wherever this whole package will eventually move (an OpenSearch module?  Another plugin?  The SDK for Java extensions repo? But it will not remain in ML Commons long term and those settings names will probably change when we relocate it.

common/src/test/java/org/opensearch/sdk/SdkClientTests.java
dbwiddis: Yes but you'd still need the LocalClusterIndicesClient.  One possibility is moving that default client to the `common` module since it doesn't need the Jackson stuff that the DDB Client does.

arjunkumargiri (approved):

common/src/test/java/org/opensearch/sdk/SdkClientTests.java
dhrubo-os: That could be a easy workaround. Do you want to address that in this PR? Or in another PR?

dbwiddis: Probably another PR. I may make another attempt to solve the jackson problems.

dhrubo-os (approved):

PR #2640, opened 2024-07-11
[Feature/multi_tenancy] Support forbidding upserts in PutDataObjectRequest

dbwiddis:
### Description

Adds an `overwriteIfExists` boolean to the `PutDataObjectRequest`, defaulting to `true`.  This replicates current default `IndexRequest` behavior using `OpType.INDEX`.   If set false, uses `OpType.CREATE`.
 
### Issues Resolved

https://github.com/opensearch-project/ml-commons/pull/2605#issuecomment-2214777743
 
### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2644, opened 2024-07-12
applying multi-tenancy and sdkclient implementation for config api + master key

dhrubo-os:
### Description
[applying multi-tenancy and sdkclient implementation for config api + master key]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLConfig.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.time.Instant;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID;
 
 @Getter
 @EqualsAndHashCode
@@ -37,19 +38,22 @@ public class MLConfig implements ToXContentObject, Writeable {
 
     private Configuration configuration;
     private final Instant createTime;
-    private Instant lastUpdateTime;
+    private final Instant lastUpdateTime;
+    private final String tenantId;
 
     @Builder(toBuilder = true)
     public MLConfig(
             String type,
             Configuration configuration,
             Instant createTime,
-            Instant lastUpdateTime
+            Instant lastUpdateTime,
+            String tenantId
     ) {
         this.type = type;
         this.configuration = configuration;
         this.createTime = createTime;
         this.lastUpdateTime = lastUpdateTime;
+        this.tenantId = tenantId;
     }
 
     public MLConfig(StreamInput input) throws IOException {
@@ -59,6 +63,8 @@ public class MLConfig implements ToXContentObject, Writeable {
         }
         createTime = input.readOptionalInstant();
         lastUpdateTime = input.readOptionalInstant();
+        //Check BWC later
+        tenantId = input.readOptionalString();
     }
 
     @Override
@@ -72,6 +78,8 @@ public class MLConfig implements ToXContentObject, Writeable {
         }
         out.writeOptionalInstant(createTime);
         out.writeOptionalInstant(lastUpdateTime);
+        // check BWC later
+        out.writeOptionalString(tenantId);
     }
 
     @Override
@@ -89,12 +97,14 @@ public class MLConfig implements ToXContentObject, Writeable {
         if (lastUpdateTime != null) {
             builder.field(LAST_UPDATE_TIME_FIELD, lastUpdateTime.toEpochMilli());
         }
+        if (tenantId != null) {
+            builder.field(TENANT_ID, tenantId);
+        }
         return builder.endObject();
     }
 
     public static MLConfig fromStream(StreamInput in) throws IOException {
-        MLConfig mlConfig = new MLConfig(in);
-        return mlConfig;
+        return new MLConfig(in);
     }
 
     public static MLConfig parse(XContentParser parser) throws IOException {
@@ -102,6 +112,7 @@ public class MLConfig implements ToXContentObject, Writeable {
         Configuration configuration = null;
         Instant createTime = null;
         Instant lastUpdateTime = null;
+        String tenantId = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -121,6 +132,8 @@ public class MLConfig implements ToXContentObject, Writeable {
                 case LAST_UPDATE_TIME_FIELD:
                     lastUpdateTime = Instant.ofEpochMilli(parser.longValue());
                     break;
+                case TENANT_ID:
+                    tenantId = parser.textOrNull();
                 default:
                     parser.skipChildren();
                     break;
@@ -131,6 +144,7 @@ public class MLConfig implements ToXContentObject, Writeable {
                 .configuration(configuration)
                 .createTime(createTime)
                 .lastUpdateTime(lastUpdateTime)
+                .tenantId(tenantId)
                 .build();
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -13,6 +13,7 @@ import java.security.PrivilegedExceptionAction;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -66,8 +67,8 @@ public interface Connector extends ToXContentObject, Writeable {
 
     <T> T createPredictPayload(Map<String, String> parameters);
 
-    void decrypt(Function<String, String> function);
-    void encrypt(Function<String, String> function);
+    void decrypt(BiFunction<String, String, String> function, String tenantId);
+    void encrypt(BiFunction<String, String, String> function, String tenantId);
 
     Connector cloneConnector();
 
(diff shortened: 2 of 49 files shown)


arjunkumargiri (commented):
High level comments: 
* Is master key required per tenant? 
* can a single master key be reused across all tenants? 
* Does the master key details needs to be exposed through getConfig API?
* Can a different encryptor implementation be used if master key need not be stored in config index?

dhrubo-os:
>* Is master key required per tenant?
>can a single master key be reused across all tenants?

Master key per tenant will reduce the blast radius from the security POV. IF we want to use single master key, that master key will be static which doesn't add much value IMO.

> Does the master key details needs to be exposed through getConfig API?

2.  Not necessarily. But this config index also being used to store the root agent ID.

> Can a different encryptor implementation be used if master key need not be stored in config index?

3. Could you please give an example? I was mostly focusing to keep the code consistent here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -102,94 +111,246 @@ public String decrypt(String encryptedText) {
...
+            );
+    }
+
+    private PutDataObjectRequest createPutDataObjectRequest(String tenantId, String generatedMasterKey) {
+        return PutDataObjectRequest
+            .builder()
+            .tenantId(tenantId)
+            .index(ML_CONFIG_INDEX)
+            .id(MASTER_KEY)
+            .overwriteIfExists(false)
```
ylwu-amzn, line 256: make sure this can work, otherwise will have race condition issue

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -223,7 +227,7 @@ private void undeployExpiredModels(
 
     @VisibleForTesting
     void initMLConfig() {
-        if (mlConfigInited) {
+        if (mlConfigInited || mlFeatureEnabledSetting.isMultiTenancyEnabled()) {
             return;
         }
         mlIndicesHandler.initMLConfigIndex(ActionListener.wrap(r -> {
```
ylwu-amzn, line 233: Change this part like EncryptorImpl.java?

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -102,94 +111,246 @@ public String decrypt(String encryptedText) {
...
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            sdkClient
+                .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor("opensearch_ml_general"))
+                .whenComplete(
+                    (response, throwable) -> handleGetDataObjectResponse(tenantId, context, response, throwable, exceptionRef, latch)
+                );
+        }
+    }
+
+    private GetDataObjectRequest createGetDataObjectRequest(String tenantId, FetchSourceContext fetchSourceContext) {
```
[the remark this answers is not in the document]
dhrubo-os: Agree but here this method is being re-used couple of times. So I think that's reducing unnecessary code lines?

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -102,94 +111,246 @@ public String decrypt(String encryptedText) {
...
+
+    private void handleGetDataObjectSuccess(
+        GetDataObjectResponse response,
+        String tenantId,
+        AtomicReference<Exception> exceptionRef,
+        CountDownLatch latch,
+        ThreadContext.StoredContext context
+    ) {
+        try {
+            GetResponse gr = response.parser() == null ? null : GetResponse.fromXContent(response.parser());
```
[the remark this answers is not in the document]
dhrubo-os: Sure.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
dhrubo-os: For multi-tenancy, init master key will happen only when the first request comes with tenant id. The cron job doesn't know tenant id, so we have to skip it directly for Multi-tenancy use case.

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
dhrubo-os: [This PR](https://github.com/opensearch-project/ml-commons/pull/2640) added this functionality. Currently we have unit tests, but we'll also integration tests on this as well. @dbwiddis may be we can have an action item on this as a part of [this issue](https://github.com/opensearch-project/ml-commons/issues/2650)?

dbwiddis: > [This PR](https://github.com/opensearch-project/ml-commons/pull/2640) added this functionality. Currently we have unit tests, but we'll also integration tests on this as well. @dbwiddis may be we can have an action item on this as a part of [this issue](https://github.com/opensearch-project/ml-commons/issues/2650)?

The PR didn't so much "add" a capability as align the API to the existing capability.  The Unit tests validates that this code produces the same Index Request as the original with `OpType.CREATE`.

If an integ test is missing, it's currently missing on `main` for this class and should be put there separate from this migration.  

is this a gap in current ML Commons integ testing?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
```diff
@@ -122,7 +122,7 @@ public void executePredict_RemoteInferenceInput_EmptyIpAddress() {
             .credential(credential)
             .actions(Arrays.asList(predictAction))
             .build();
-        connector.decrypt((c) -> encryptor.decrypt(c));
+        connector.decrypt((c, tenantId) -> encryptor.decrypt(c, null), null);
```
dbwiddis, line 125: This is starting to strain readability.  

Should the last part of that be `encryptor.decrypt(c, tenantId), null` ?

If so I think this is the same thing and more readable: `encryptor::decrypt, null`

If not, deserves a comment explaining why we're throwing away tenant ID, probably best put above a constant bifunction you can throw into this method to replace `(c, tenantId) -> encryptor.decrypt(c, null)`

dbwiddis (approved):
LGTM Generally. Left a few suggestions/comments.

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
```diff
@@ -66,8 +67,8 @@ public interface Connector extends ToXContentObject, Writeable {
 
     <T> T createPredictPayload(Map<String, String> parameters);
 
-    void decrypt(Function<String, String> function);
-    void encrypt(Function<String, String> function);
+    void decrypt(BiFunction<String, String, String> function, String tenantId);
+    void encrypt(BiFunction<String, String, String> function, String tenantId);
```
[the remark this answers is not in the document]
dhrubo-os: Agree, I'll add that in the next follow up PR.

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -59,6 +63,8 @@ public MLConfig(StreamInput input) throws IOException {
         }
         createTime = input.readOptionalInstant();
         lastUpdateTime = input.readOptionalInstant();
+        //Check BWC later
```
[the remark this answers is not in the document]
dhrubo-os: Agree, I will address that in the next follow up PR.

common/src/main/java/org/opensearch/sdk/PutDataObjectRequest.java
```diff
@@ -145,6 +149,16 @@ public Builder dataObject(ToXContentObject dataObject) {
...
 
+        /**
+         * Add a data object as a map to this builder
+         * @param dataObjectMap the data object as a map of fields
+         * @return the updated builder
+         */
+        public Builder dataObject(Map<String, Object> dataObjectMap) {
+            this.dataObject = (builder, params) -> builder.map(dataObjectMap);
+            return this;
+        }
```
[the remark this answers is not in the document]
dhrubo-os: Agree, I'll add that in the next follow up PR.

common/src/main/java/org/opensearch/sdk/UpdateDataObjectRequest.java
```diff
@@ -67,7 +67,7 @@ public String id() {
     public String tenantId() {
         return this.tenantId;
     }
-    
+
```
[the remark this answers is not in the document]
dhrubo-os: That's a great callout. I just raised a PR to address that in main: https://github.com/opensearch-project/ml-commons/pull/2658

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutorTest.java
dhrubo-os: Here,  the `decrypt` method of the connector is called with a lambda function and null as the arguments.

1. This lambda function is an implementation of the BiFunction<String, String, String> interface.
2. It takes two parameters: c (which represents the encrypted text) and tenantId (which is not used in this context).
3. Inside the lambda function, encryptor.decrypt(c, null) is called.c is passed as the encryptedText.
null is passed as the tenantId to the encryptor.decrypt method.

Issue #2645, opened 2024-07-12
[FEATURE] ChatAgentRunner supports passing CHAT_HISTORY to tools if needed


qianheng-aws:
**Is your feature request related to a problem?**
In the executing process of ChatAgentRunner, LLM generates `action` and `actionInput` to choose a tool and its input parameters. 
However, some tools may need input parameters like CHAT_HISTORY which doesn't comes from LLM response but the in-process parameters of ChatAgentRunner.

We wants to make a CreateAlertTool to create alerting monitor by LLM. The configuration of alert if somehow complex, to make the response solid, this tool should has its own prompt and access LLM itself. And to make it context-aware, it needs CHAT_HISTORY as its input as well. There is the related issue in skills repo: https://github.com/opensearch-project/skills/issues/338

**What solution would you like?**
Add a new method of `needHistory` on the `Tool` interface, so ChatAgentRunner can know whether to pass CHAT_HISTORY to this tool when calling `constructToolParams` method. 

**What alternatives have you considered?**
I have considered asking LLM to put the CHAT_HISTORY in `actionInput` by adding explicit instructions in the description of CreateAlertTool, so we don't need to make a change in this repo. 
But the results were not good or accurate. Since the length of CHAT_HISTORY is very long most times, the CHAT_HISTORY in `actionInput` returned by LLM are mostly wrong or incomplete. And it's somehow a waste of LLM tokens. 
Since we have CHAT_HISTORY stored in the parameters of ChatAgentRunner, we better to pass it directly to the tools if they need.

**Do you have any additional context?**
There is a related issue in skills repo has **dependency** on this one: https://github.com/opensearch-project/skills/issues/338

qianheng-aws:
resolved by this PR: https://github.com/opensearch-project/ml-commons/pull/2714

PR #2646, opened 2024-07-12
chat agent runner supports passing chat history to tool if needed

qianheng-aws:
### Description
Some tools may need CHAT_HISTORY as its input parameters, this PR make ChatAgentRunner supports passing CHAT_HISTORY to tools if needed.

### Related Issues
original issue: https://github.com/opensearch-project/ml-commons/issues/2645
issue has dependency on this: https://github.com/opensearch-project/skills/issues/338

### Check List
- [ ] All tests pass

### DIFF
--- a/spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
+++ b/spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
@@ -98,6 +98,14 @@ public interface Tool {
     }
 
     /**
+     * Whether the tool needs the history of the conversation.
+     * @return if true, agent runner should pass history information as input parameter to this tool
+     */
+    default boolean needHistory() {
+        return false;
+    }
+
+    /**
      * Tool factory which can create instance of {@link Tool}.
      * @param <T> The subclass this factory produces
      */
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -461,7 +461,8 @@ public class AgentUtils {
         String question,
         AtomicReference<String> lastActionInput,
         String action,
-        String actionInput
+        String actionInput,
+        Map<String, String> parameters
     ) {
         Map<String, String> toolParams = new HashMap<>();
         Map<String, String> toolSpecParams = toolSpecMap.get(action).getParameters();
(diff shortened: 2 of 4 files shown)


xinyual:
Why we need to pass history to tool?

qianheng-aws:
> Why we need to pass history to tool? Can LLM generate parameters according to history instead of pass history directly to the tool?

To make this tool context aware since it needs to request LLM itself. 

The LLM in ChatAgentRunner cannot generate the right parameters out of this tool, because it doesn't know the mapping info(though it can select IndexMappingTool sometimes if you told it todo so, it's very unstable), and the results are mostly inaccurate due to its complex and unclear prompt.

xinyual (approved):

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
```diff
@@ -97,6 +97,14 @@ default boolean useOriginalInput() {
         return false;
     }
 
+    /**
+     * Whether the tool needs the history of the conversation.
+     * @return if true, agent runner should pass history information as input parameter to this tool
+     */
+    default boolean needHistory() {
```
ylwu-amzn, line 104: Adding new method to Tool interface is a concern to tool developers. They need to go through all methods to check if they need to implement the method or not. With more and more methods added to Tool interface, will take a lot of time to learn what's each method for, it also brings confusion. Let's discuss with @jngz-es to check if there are any other options ?

ylwu-amzn: @qianheng-aws Can you check if you can get chat history in tool by using `parameters.get("chat_history")`

qianheng-aws: In CreatAlertTool, parameters don't have entry "chat_history" unless we pass the history to it. That's why we need this method to control whether passing "chat_history"

zane-neo: Is this a necessary to add a new method to determine if the chat_history should be passed or not? Can we always pass chat_history to all tools as parameter so tool can access this based on their needs.

qianheng-aws: @zane-neo  Incline with that if it's acceptable. What's your opinion? @ylwu-amzn

qianheng-aws:
resolved by this PR: https://github.com/opensearch-project/ml-commons/pull/2714

PR #2651, opened 2024-07-16
feat: add bedrock runtime agent for knowledge base

yuye-aws:
### Description
Enable users to register bedrock runtime agent for calling bedrock knowledge base.
 
### Issues Resolved

```
{
  "name": "Amazon Bedrock Connector: knowledge",
  "description": "The connector to the Bedrock knowledge base",
  "version": 1,
  "protocol": "aws_sigv4",
  "parameters": {
    "region": "...",
    "service_name": "bedrock"
  },
  "credential": {
    "access_key": "...",
    "secret_key": "..."
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "url": "...",
      "headers": {
        "x-amz-content-sha256": "required"
      },
      "request_body": "{\"retrievalQuery\": {\"text\": \"${parameters.text}\"}}"
    }
  ]
}

POST /_plugins/_ml/models/_register?deploy=true
{
  "name": "bedrock: knowledge base",
  "function_name": "remote",
  "description": "Test connector for bedrock knowledge base",
  "connector_id": <connector_id>
}

POST /_plugins/_ml/models/<model_id>/_predict/
{
  "parameters": {
    "text": "aws 2021"
  }
}
```
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
+++ b/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
@@ -144,7 +144,8 @@ public final class MLCommonsSettings {
                     "^https://runtime\\.sagemaker\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
                     "^https://api\\.openai\\.com/.*$",
                     "^https://api\\.cohere\\.ai/.*$",
-                    "^https://bedrock-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$"
+                    "^https://bedrock-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
+                    "^https://bedrock-agent-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$"
                 ),
             Function.identity(),
             Setting.Property.NodeScope,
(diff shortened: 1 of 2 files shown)


xinyual (approved):

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -144,7 +144,8 @@ private MLCommonsSettings() {}
                     "^https://runtime\\.sagemaker\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
                     "^https://api\\.openai\\.com/.*$",
                     "^https://api\\.cohere\\.ai/.*$",
-                    "^https://bedrock-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$"
+                    "^https://bedrock-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
+                    "^https://bedrock-agent-runtime\\..*[a-z0-9-]\\.amazonaws\\.com/.*$"
```
Hailong-am, line 148: should we combine these two bedrock endpoint into one?

Hailong-am (approved):

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
yuye-aws: Do you mean regex like the following?
`^https://(?:bedrock-runtime|bedrock-agent-runtime)\\..*[a-z0-9-]\\.amazonaws\\.com/.*$
`

But I think the current code is more readable, so I prefer remain unchanged.

xinyual (approved):

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
Hailong-am: I mean bedrock.* to include all bedrock related, so we don't need to add more if there has another bedrock-xyz. I think it's ok to keep as two for now and merge into one if we have more bedrock endpoints comes after.

ylwu-amzn (approved):

PR #2652, opened 2024-07-16
Add feature enable setting for controller index

b4sjoo:
### Description
This PR will add a feature enable setting for controller index, to support multi-tenancy
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
@@ -9,6 +9,7 @@ import static org.opensearch.ml.common.CommonValue.ML_CONTROLLER_INDEX;
 import static org.opensearch.ml.common.FunctionName.REMOTE;
 import static org.opensearch.ml.common.FunctionName.TEXT_EMBEDDING;
 import static org.opensearch.ml.common.utils.StringUtils.getErrorMessage;
+import static org.opensearch.ml.utils.MLExceptionUtils.CONTROLLER_DISABLED_ERR_MSG;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -51,6 +52,7 @@ import org.opensearch.ml.engine.indices.MLIndicesHandler;
 import org.opensearch.ml.helper.ModelAccessControlHelper;
 import org.opensearch.ml.model.MLModelCacheHelper;
 import org.opensearch.ml.model.MLModelManager;
+import org.opensearch.ml.settings.MLFeatureEnabledSetting;
 import org.opensearch.ml.utils.RestActionUtils;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
@@ -68,6 +70,7 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
     ClusterService clusterService;
     MLModelCacheHelper mlModelCacheHelper;
     ModelAccessControlHelper modelAccessControlHelper;
+    private MLFeatureEnabledSetting mlFeatureEnabledSetting;
 
     @Inject
     public CreateControllerTransportAction(
@@ -78,7 +81,8 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
         ClusterService clusterService,
         ModelAccessControlHelper modelAccessControlHelper,
         MLModelCacheHelper mlModelCacheHelper,
-        MLModelManager mlModelManager
+        MLModelManager mlModelManager,
+        MLFeatureEnabledSetting mlFeatureEnabledSetting
     ) {
         super(MLCreateControllerAction.NAME, transportService, actionFilters, MLCreateControllerRequest::new);
         this.mlIndicesHandler = mlIndicesHandler;
@@ -87,6 +91,7 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
         this.clusterService = clusterService;
         this.mlModelCacheHelper = mlModelCacheHelper;
         this.modelAccessControlHelper = modelAccessControlHelper;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
     }
 
     @Override
@@ -98,6 +103,9 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
+                throw new IllegalStateException(CONTROLLER_DISABLED_ERR_MSG);
+            }
             ActionListener<MLCreateControllerResponse> wrappedListener = ActionListener.runBefore(actionListener, context::restore);
             mlModelManager.getModel(modelId, null, excludes, ActionListener.wrap(mlModel -> {
                 FunctionName functionName = mlModel.getAlgorithm();
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/DeleteControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/DeleteControllerTransportAction.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.action.controller;
 import static org.opensearch.action.support.WriteRequest.RefreshPolicy.IMMEDIATE;
 import static org.opensearch.ml.common.CommonValue.ML_CONTROLLER_INDEX;
 import static org.opensearch.ml.common.utils.StringUtils.getErrorMessage;
+import static org.opensearch.ml.utils.MLExceptionUtils.CONTROLLER_DISABLED_ERR_MSG;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -41,6 +42,7 @@ import org.opensearch.ml.common.transport.controller.MLUndeployControllerNodesRe
 import org.opensearch.ml.helper.ModelAccessControlHelper;
 import org.opensearch.ml.model.MLModelCacheHelper;
 import org.opensearch.ml.model.MLModelManager;
+import org.opensearch.ml.settings.MLFeatureEnabledSetting;
 import org.opensearch.ml.utils.RestActionUtils;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
@@ -58,6 +60,7 @@ public class DeleteControllerTransportAction extends HandledTransportAction<Acti
     MLModelManager mlModelManager;
     MLModelCacheHelper mlModelCacheHelper;
     ModelAccessControlHelper modelAccessControlHelper;
+    private MLFeatureEnabledSetting mlFeatureEnabledSetting;
 
     @Inject
     public DeleteControllerTransportAction(
@@ -68,7 +71,8 @@ public class DeleteControllerTransportAction extends HandledTransportAction<Acti
         ClusterService clusterService,
         MLModelManager mlModelManager,
         MLModelCacheHelper mlModelCacheHelper,
-        ModelAccessControlHelper modelAccessControlHelper
+        ModelAccessControlHelper modelAccessControlHelper,
+        MLFeatureEnabledSetting mlFeatureEnabledSetting
     ) {
         super(MLControllerDeleteAction.NAME, transportService, actionFilters, MLControllerDeleteRequest::new);
         this.client = client;
@@ -77,6 +81,7 @@ public class DeleteControllerTransportAction extends HandledTransportAction<Acti
         this.mlModelManager = mlModelManager;
         this.mlModelCacheHelper = mlModelCacheHelper;
         this.modelAccessControlHelper = modelAccessControlHelper;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
     }
 
     @Override
@@ -86,6 +91,9 @@ public class DeleteControllerTransportAction extends HandledTransportAction<Acti
         User user = RestActionUtils.getUserContext(client);
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
+                throw new IllegalStateException(CONTROLLER_DISABLED_ERR_MSG);
+            }
             ActionListener<DeleteResponse> wrappedListener = ActionListener.runBefore(actionListener, context::restore);
             mlModelManager.getModel(modelId, null, excludes, ActionListener.wrap(mlModel -> {
                 Boolean isHidden = mlModel.getIsHidden();
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/GetControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/GetControllerTransportAction.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.action.controller;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_CONTROLLER_INDEX;
 import static org.opensearch.ml.common.utils.StringUtils.getErrorMessage;
+import static org.opensearch.ml.utils.MLExceptionUtils.CONTROLLER_DISABLED_ERR_MSG;
 import static org.opensearch.ml.utils.MLNodeUtils.createXContentParserFromRegistry;
 import static org.opensearch.ml.utils.RestActionUtils.getFetchSourceContext;
 
@@ -33,6 +34,7 @@ import org.opensearch.ml.common.transport.controller.MLControllerGetRequest;
 import org.opensearch.ml.common.transport.controller.MLControllerGetResponse;
 import org.opensearch.ml.helper.ModelAccessControlHelper;
 import org.opensearch.ml.model.MLModelManager;
+import org.opensearch.ml.settings.MLFeatureEnabledSetting;
 import org.opensearch.ml.utils.RestActionUtils;
 import org.opensearch.search.fetch.subphase.FetchSourceContext;
 import org.opensearch.tasks.Task;
@@ -50,6 +52,7 @@ public class GetControllerTransportAction extends HandledTransportAction<ActionR
     ClusterService clusterService;
     MLModelManager mlModelManager;
     ModelAccessControlHelper modelAccessControlHelper;
+    private MLFeatureEnabledSetting mlFeatureEnabledSetting;
 
     @Inject
     public GetControllerTransportAction(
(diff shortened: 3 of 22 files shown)


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -736,9 +735,9 @@ public List<RestHandler> getRestHandlers(
         RestMemorySearchInteractionsAction restSearchInteractionsAction = new RestMemorySearchInteractionsAction();
         RestMemoryGetConversationAction restGetConversationAction = new RestMemoryGetConversationAction();
         RestMemoryGetInteractionAction restGetInteractionAction = new RestMemoryGetInteractionAction();
-        RestMLCreateControllerAction restMLCreateControllerAction = new RestMLCreateControllerAction();
+        RestMLCreateControllerAction restMLCreateControllerAction = new RestMLCreateControllerAction(mlFeatureEnabledSetting);
         RestMLGetControllerAction restMLGetControllerAction = new RestMLGetControllerAction();
-        RestMLUpdateControllerAction restMLUpdateControllerAction = new RestMLUpdateControllerAction();
+        RestMLUpdateControllerAction restMLUpdateControllerAction = new RestMLUpdateControllerAction(mlFeatureEnabledSetting);
         RestMLDeleteControllerAction restMLDeleteControllerAction = new RestMLDeleteControllerAction();
```
dhrubo-os, line 741: What about for delete controller? Do we need to check there too?

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
```diff
@@ -24,6 +24,8 @@ public class MLExceptionUtils {
         "Local Model is currently disabled. To enable it, update the setting \"plugins.ml_commons.local_model.enabled\" to true.";
     public static final String AGENT_FRAMEWORK_DISABLED_ERR_MSG =
         "Agent Framework is currently disabled. To enable it, update the setting \"plugins.ml_commons.agent_framework_enabled\" to true.";
+    public static final String CONTROLLER_DISABLED_ERR_MSG =
+        "Controller is currently disabled. To enable it, update the setting \"plugins.ml_commons.controller_enabled\" to true.";
```
dhrubo-os, line 28: Can we have more explicit message like "User level throttling controller" or something like that? Otherwise I'm not sure if customer knows what is Controller in general.

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
```diff
@@ -98,6 +103,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLCrea
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
dhrubo-os, line 106: Do we also need this for getController and delete controller?

dhrubo-os:
Shouldn't we check this kind of places too: https://github.com/opensearch-project/ml-commons/blob/b6618b2bcd18ab7070a8b8923074d365efe89dd9/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java#L1044

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
b4sjoo: The user must know what is controller in general because when a user calls the API it's `_controller` API. And in the future we will include fields more than user-level throttling in controller index, so, have message with user level throttling is not a maintainable way.

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
b4sjoo: Added

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
b4sjoo: Added

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
```diff
@@ -25,11 +26,15 @@ public class MLFeatureEnabledSetting {
     private volatile Boolean isLocalModelEnabled;
     private volatile AtomicBoolean isConnectorPrivateIpEnabled;
 
+    private volatile Boolean isModelRateLimiterEnabled;
```
dhrubo-os, line 29: what is this for?

b4sjoo: It's redundant, removed.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -746,10 +745,10 @@ public List<RestHandler> getRestHandlers(
         RestMemorySearchInteractionsAction restSearchInteractionsAction = new RestMemorySearchInteractionsAction();
         RestMemoryGetConversationAction restGetConversationAction = new RestMemoryGetConversationAction();
         RestMemoryGetInteractionAction restGetInteractionAction = new RestMemoryGetInteractionAction();
-        RestMLCreateControllerAction restMLCreateControllerAction = new RestMLCreateControllerAction();
-        RestMLGetControllerAction restMLGetControllerAction = new RestMLGetControllerAction();
-        RestMLUpdateControllerAction restMLUpdateControllerAction = new RestMLUpdateControllerAction();
-        RestMLDeleteControllerAction restMLDeleteControllerAction = new RestMLDeleteControllerAction();
+        RestMLCreateControllerAction restMLCreateControllerAction = new RestMLCreateControllerAction(mlFeatureEnabledSetting);
```
zane-neo, line 748: Why this injected to RestActions?

zane-neo (commented):
Any ITs to cover this?

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -380,7 +379,7 @@ public MachineLearningPlugin(Settings settings) {
 
     @Override
     public List<ActionHandler<? extends ActionRequest, ? extends ActionResponse>> getActions() {
-        return ImmutableList
+        return List
```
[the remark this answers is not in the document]
b4sjoo: According to PE we should avoid using class like `ImmutableList` `ImmutableMap`, instead if we want to create an immutable type we should use java.utils directly

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
b4sjoo: Because we want to disable it in rest layer

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -199,4 +199,7 @@ private MLCommonsSettings() {}
 
     public static final Setting<Boolean> ML_COMMONS_CONNECTOR_PRIVATE_IP_ENABLED = Setting
         .boolSetting("plugins.ml_commons.connector.private_ip_enabled", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
+    public static final Setting<Boolean> ML_COMMONS_CONTROLLER_ENABLED = Setting
+        .boolSetting("plugins.ml_commons.controller_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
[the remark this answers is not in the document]
b4sjoo: It is Controller and not Model Controller, so I would insist this name.

plugin/src/main/java/org/opensearch/ml/utils/MLExceptionUtils.java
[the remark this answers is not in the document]
b4sjoo: Same here.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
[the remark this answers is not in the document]
dhrubo-os: `ImmutableList` is inherently thread-safe. Multiple threads can access it without requiring synchronization. A standard List is not thread-safe by default. Did you check if it can arise any issue? Let's not change this in this PR if we are unsure about the further consequences. If we need to change this, we can raise a separate PR.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
dhrubo-os: Controller for what? Model or Connector or ML-Commons?

plugin/src/main/java/org/opensearch/ml/rest/RestMLCreateControllerAction.java
```diff
@@ -61,6 +66,10 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
      * @return MLCreateControllerRequest
      */
     private MLCreateControllerRequest getRequest(RestRequest request) throws IOException {
+        if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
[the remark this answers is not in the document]
dhrubo-os: let's say, other plugins are also sending a transport request to ml-commons. Transport validation will be necessary in that case. We can treat this as a front-end and back-end validation.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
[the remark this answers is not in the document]
b4sjoo: Ok will revert

plugin/src/main/java/org/opensearch/ml/rest/RestMLCreateControllerAction.java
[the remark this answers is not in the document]
b4sjoo: Correct, and for security reason we'd better add the rest layer check so that when DDOS attack come we can minimize the node pressure.

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
```diff
@@ -71,4 +76,12 @@ public AtomicBoolean isConnectorPrivateIpEnabled() {
         return isConnectorPrivateIpEnabled;
     }
 
+    /**
+     * Whether the controller feature is enabled. If disabled, APIs in ml-commons will block controller.
+     * @return whether the controller is enabled.
+     */
+    public Boolean isControllerEnabled() {
```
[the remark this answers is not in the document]
b4sjoo: Will remove

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
b4sjoo: Addressed

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
b4sjoo: Eventually we want to generalize it for whole ML plugin

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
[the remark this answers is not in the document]
zane-neo: Change `Boolean` to `boolean`.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
zane-neo: You can generalize the feature, but the name needs to be specific so that users know what controller it's for. Using a general name doesn't seem a good idea to me.

plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java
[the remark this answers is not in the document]
dhrubo-os: I think that's fine as other variables are also `Boolean`. And Boolean could be helpful to represent nullability to maintain backward compatibility.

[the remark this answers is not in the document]
zane-neo: This new setting has a default value, so the returned value will never be null. I didn't see the ability to maintain backward compatibility. Using raw type can increase performance a little bit but it's definitely not a blocking issue.

dhrubo-os:
@b4sjoo Any update on this?

b4sjoo:
I don't think there are any blocking comments left, can we approve and merge it?

b4sjoo:
> @b4sjoo Any update on this?

@dhrubo-os I don't think there are any blocking comments left, can we approve and merge it?

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
dhrubo-os, line 106: I just realized why do we need to throw the exception after stashing the context? We can throw this exception before stashing the context, right?

plugin/src/main/java/org/opensearch/ml/action/controller/DeleteControllerTransportAction.java
```diff
@@ -86,6 +91,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
         User user = RestActionUtils.getUserContext(client);
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
dhrubo-os, line 94: Same here.

plugin/src/main/java/org/opensearch/ml/action/controller/GetControllerTransportAction.java
```diff
@@ -79,6 +84,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLCont
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
dhrubo-os, line 87: Same here

plugin/src/main/java/org/opensearch/ml/action/controller/UpdateControllerTransportAction.java
```diff
@@ -90,6 +95,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
dhrubo-os, line 98: Same here.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -1254,6 +1255,9 @@ public synchronized void updateModelCache(String modelId, ActionListener<String>
      */
     public synchronized void deployControllerWithDeployedModel(String modelId, ActionListener<String> listener) {
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            if (!mlFeatureEnabledSetting.isControllerEnabled()) {
```
dhrubo-os, line 1258: Same here

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
dhrubo-os: @zane-neo do you have any suggestion? 

We already published this feature as controller: https://opensearch.org/docs/latest/ml-commons-plugin/api/controller-apis/index/

dhrubo-os:
I haven't see any unit tests when controller is disabled for the transport actions. Can we add some tests accordingly?

dhrubo-os:
Integration tests are failing. I retried now. I left few comments @b4sjoo

b4sjoo:
> I haven't see any unit tests when controller is disabled for the transport actions. Can we add some tests accordingly?

Fine I can add them when I get back. Not a hard job

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
b4sjoo: Yes you are right. I just followed other's patterns.

dhrubo-os:
I think we should also check [here](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L290) before performing deletion?

b4sjoo:
> I think we should also check [here](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L290) before performing deletion?

No need because it calls `DeleteControllerTransportAction`, which already have applied feature flag. No need to do the duplicated call.

plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
b4sjoo: Just verified with UT that we cannot do this movement. This is because if we move the exception outside it will throw the exception directly, instead of catching by `actionlistener`.

dhrubo-os: What's the difference? how stashing context help to catch by `actionlistener`?

b4sjoo: Actually I am not so sure about why, but UT's behaved in such way

b4sjoo: Maybe we can see if @ylwu-amzn has any output here?

dhrubo-os: Because in the catch block we perform `actionListener.onFailure(e);` to catch the exception. Let's keep the code as it is then.

dhrubo-os (approved):

Zhangxunmt (approved):

Issue #2654, opened 2024-07-16
[BUG] MLModelTool returns null if the response of LLM is a pure json object


qianheng-aws:
**What is the bug?**
MLModelTool returns null if the response of LLM is a pure json object

**How can one reproduce the bug?**
This bug happens when developing the feature of "text2visualization", it's hard to build the environment since it has not been released. But we can simplify the case with below several steps:
1. register a flow agent with MLModelTool whose prompt should be like "\n\nHuman:You should generate a json object directly with format {name: ${NAME}, age: ${AGE}}. \n\nHuman:${parameters.question}\n\nAssistant:"
2. request flow agent with question to generate a json object directly, e.g. "Generate a json object for Jack who is 18"
3. it will occasionally produce error message like below
```
"{\"error\":{\"reason\":\"System Error\",\"details\":\"Cannot invoke \\\"org.opensearch.ml.common.output.Output.toXContent(org.opensearch.core.xcontent.XContentBuilder, org.opensearch.core.xcontent.ToXContent$Params)\\\" because \\\"this.output\\\" is null\",\"type\":\"NullPointerException\"},\"status\":500}"
```

**What is the expected behavior?**
It should return the right json object `{name: Jack, age: 18}`

**What is your host/environment?**
 - OS: [e.g. macOS]
 - Version [e.g. 3.0.0-SNAPSHOT]
 - Plugin

**Do you have any screenshots?**
The error message in dev-tools:
![image](https://github.com/user-attachments/assets/dfac4c87-1a9d-4b05-90d8-8b612da9acd0)

The debug result of ModelTensors:
![image](https://github.com/user-attachments/assets/ec91e303-743d-487f-94fc-9763ef143f15)

**Do you have any additional context?**
It's caused by the fragile implementation of MLModelTool which misses the case that LLM can return a json object directly. See this line:
https://github.com/opensearch-project/ml-commons/blob/b6618b2bcd18ab7070a8b8923074d365efe89dd9/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java#L69

PR #2655, opened 2024-07-16
Fix MLModelTool returns null if the response of LLM is a pure json object

qianheng-aws:
### Description
Fix MLModelTool returns null if the response of LLM is a pure json object. It will detect wether the LLM response is a pure json object or a response string, and parse in different way.

There is similar logic in [AgentUtils.java](https://github.com/opensearch-project/ml-commons/blob/b6618b2bcd18ab7070a8b8923074d365efe89dd9/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java#L193) for MLChatAgentRunner

https://github.com/opensearch-project/ml-commons/blob/b6618b2bcd18ab7070a8b8923074d365efe89dd9/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java#L193
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2654
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
@@ -8,7 +8,6 @@ package org.opensearch.ml.engine.tools;
 import java.util.List;
 import java.util.Map;
 
-import java.util.NoSuchElementException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.client.Client;
 import org.opensearch.core.action.ActionListener;
@@ -70,8 +69,7 @@ public class MLModelTool implements Tool {
         outputParser = o -> {
             try {
                 List<ModelTensors> mlModelOutputs = (List<ModelTensors>) o;
-                Map<String, ?> dataAsMap = mlModelOutputs.getFirst().getMlModelTensors().getFirst()
-                    .getDataAsMap();
+                Map<String, ?> dataAsMap = mlModelOutputs.getFirst().getMlModelTensors().getFirst().getDataAsMap();
                 // Return the response field if it exists, otherwise return the whole response as json string.
                 if (dataAsMap.containsKey(responseField)) {
                     return dataAsMap.get(responseField);
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
@@ -69,7 +69,8 @@ public class MLModelTool implements Tool {
         outputParser = o -> {
             List<ModelTensors> mlModelOutputs = (List<ModelTensors>) o;
             Map<String, ?> dataAsMap = mlModelOutputs.get(0).getMlModelTensors().get(0).getDataAsMap();
-            if (dataAsMap.size() == 1 && dataAsMap.containsKey(responseField)) {
+            // Return the response field if it exists, otherwise return the whole response a json string.
+            if (dataAsMap.containsKey(responseField)) {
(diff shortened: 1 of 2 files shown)


xinyual (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
```diff
@@ -66,7 +68,13 @@ public MLModelTool(Client client, String modelId, String responseField) {
 
         outputParser = o -> {
             List<ModelTensors> mlModelOutputs = (List<ModelTensors>) o;
-            return mlModelOutputs.get(0).getMlModelTensors().get(0).getDataAsMap().get(responseField);
+            Map<String, ?> dataAsMap = mlModelOutputs.get(0).getMlModelTensors().get(0).getDataAsMap();
+            // Return the response field if it exists, otherwise return the whole response a json string.
+            if (dataAsMap.containsKey(responseField)) {
```
zane-neo, line 73: Original code doesn't have check on dataAsMap which might cause NPE in edge case, it would be good to add a check here.

qianheng-aws: I see many places that access dataAsMap directly without any index checking, so I follow that rule. But I can add array index checking.

zane-neo (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
```diff
@@ -54,6 +55,7 @@ public class MLModelTool implements Tool {
     private Parser inputParser;
     @Setter
     @Getter
+    @VisibleForTesting
```
dhrubo-os, line 58: Why do we need this?

qianheng-aws: Make UT be able to access the private field `outputParser` and test whether its results are as expected.

xinyual (approved):

b4sjoo:
@qianheng-aws This PR will break 2.x compile if backport, you need to fix and raise a backport PR manually, please check

qianheng-aws:
> @qianheng-aws This PR will break 2.x compile if backport, you need to fix and raise a backport PR manually, please check

Thanks for reminding. It's because I use a java21 API in this change while branch 2.x is built under java17. I've raised a new PR to backport it, https://github.com/opensearch-project/ml-commons/pull/2675

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
dhrubo-os: But in the test you already declared: `Parser outputParser = new MLModelTool(client, "modelId", "response").getOutputParser();` How are you using `outputParser` from the MLModelTool class then?

dhrubo-os:
https://github.com/opensearch-project/ml-commons/actions/runs/9997362810/job/27633752977?pr=2679

this workflow is failing. May be we should use method which is compatible with java 11 too. Could you please refactor this change? Thanks

qianheng-aws:
> https://github.com/opensearch-project/ml-commons/actions/runs/9997362810/job/27633752977?pr=2679
> 
> this workflow is failing. May be we should use method which is compatible with java 11 too. Could you please refactor this change? Thanks

Raised this PR https://github.com/opensearch-project/ml-commons/pull/2682 to avoid diverge on main branch. It will be compatible with java11 as well.

PR #2666, opened 2024-07-17
agent execution with sdk client and multi-tenancy functionality

dhrubo-os:
### Description
[agent execution with sdk client and multi-tenancy functionality]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java
```diff
@@ -260,6 +265,7 @@ Tool createTool(MLToolSpec toolSpec) {
         if (toolSpec.getName() != null) {
             tool.setName(toolSpec.getName());
         }
+        tool.setTenantId(toolSpec.getTenantId());
```
ylwu-amzn, line 268: Why need to pass tenant id to Tool?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
```diff
@@ -8,6 +8,8 @@
 import java.util.List;
 import java.util.Map;
 
+import javax.swing.*;
+
```
arjunkumargiri, line 12: Why is this import required?

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
```diff
@@ -80,6 +80,9 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLAgen
         MLAgentGetRequest mlAgentGetRequest = MLAgentGetRequest.fromActionRequest(request);
         String agentId = mlAgentGetRequest.getAgentId();
         String tenantId = mlAgentGetRequest.getTenantId();
+        if (!TenantAwareHelper.validateTenantId(mlFeatureEnabledSetting, tenantId, actionListener)) {
+            return;
```
arjunkumargiri, line 84: We will need a similar validation for agent execute action.

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
```diff
@@ -97,6 +97,18 @@ default boolean useOriginalInput() {
         return false;
     }
 
+    /**
+     * Get the tenant ID.
+     * @return the tenant ID
+     */
+    String getTenantId();
```
arjunkumargiri, line 104: Tenant ID will be a runtime variable, if a single instance of tool is shared between threads then this approach could lead to race conditions.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java
ylwu-amzn: Adding new method to pass parameters between agent and tools , tools and tools is not extensible/scalable. If we are going to add 100s new parameters, should we add 100s new set/get methods ?

I personally prefer to pass tenant id in `Tool.run(...)` method. For now we only pass Map of parameters which is mutable. For such immutable parameters, we can add a new Immutable map like 
```
default <T> void run(Map<String, String> parameters, ImmutableMap<String, String> configurations, ActionListener<T> listener) {};
```

arjunkumargiri: If there is a requirement for multiple configuration to be added to a tool, we could consider creating a class for configuration and set that to abstract class. 

However, given that tenant ID is a runtime parameter which could change based on API invocation it makes sense to add it to `run` method instead of adding to abstract class.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -135,11 +135,11 @@ public void removeCredential() {
     @Override
     public String getPredictEndpoint(Map<String, String> parameters) {
         Optional<ConnectorAction> predictAction = findPredictAction();
-        if (!predictAction.isPresent()) {
+        if (predictAction.isEmpty()) {
```
dbwiddis, line 138: Huh. I didn't even know this was a thing!

common/src/main/java/org/opensearch/ml/common/dataset/remote/RemoteInferenceInputDataSet.java
```diff
@@ -32,7 +32,7 @@ public RemoteInferenceInputDataSet(Map<String, String> parameters) {
     public RemoteInferenceInputDataSet(StreamInput streamInput) throws IOException {
         super(MLInputDataType.REMOTE);
         if (streamInput.readBoolean()) {
-            parameters = streamInput.readMap(s -> s.readString(), s-> s.readString());
+            parameters = streamInput.readMap(StreamInput::readString, StreamInput::readString);
```
dbwiddis, line 35: +1

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -107,70 +121,137 @@ public void execute(Input input, ActionListener<Output> listener) {
...
-                            String memoryId = inputDataSet.getParameters().get(MEMORY_ID);
-                            String parentInteractionId = inputDataSet.getParameters().get(PARENT_INTERACTION_ID);
-                            String regenerateInteractionId = inputDataSet.getParameters().get(REGENERATE_INTERACTION_ID);
-                            String appType = mlAgent.getAppType();
-                            String question = inputDataSet.getParameters().get(QUESTION);
-
-                            if (memoryId == null && regenerateInteractionId != null) {
-                                throw new IllegalArgumentException("A memory ID must be provided to regenerate.");
+                sdkClient
+                    .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor("opensearch_ml_general"))
```
dbwiddis, line 136: the executor should probably be a constant somewhere?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java
dbwiddis: A more scalable method of passing parameters like this would be a key-value map that we could add to.

That said,

> tenant ID is a runtime parameter which could change based on API invocation

I'm a bit confused here. Isn't tenant ID always the same for a given customer? We're using it as a key in our metadata store.  When does it change?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java
dbwiddis: And why is it a `*`?  Wildcard imports are generally discouraged.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
dhrubo-os: I'm thinking to move `GENERAL_THREAD_POOL` (along with other POOLs static variable) to Common module so that algorithm & plugin both can have access. But that demands a separate PR. So it'll here for a while

ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
```diff
@@ -17,5 +17,14 @@ public interface Executable {
...
      * @return execution result
      */
-    void execute(Input input, ActionListener<Output> listener) throws ExecuteException;
+    default void execute(Input input, ActionListener<Output> listener) throws ExecuteException {
+        execute(input, null, listener);
+    }
+
+    /**
+     * Execute algorithm with given input data.
+     * @param input input data
```
arjunkumargiri, line 26: Add java doc for tenant ID

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -69,108 +75,187 @@ public class MLAgentExecutor implements Executable {
     public static final String MESSAGE_HISTORY_LIMIT = "message_history_limit";
 
     private Client client;
+    private SdkClient sdkClient;
     private Settings settings;
     private ClusterService clusterService;
     private NamedXContentRegistry xContentRegistry;
     private Map<String, Tool.Factory> toolFactories;
     private Map<String, Memory.Factory> memoryFactoryMap;
+    private volatile Boolean isMultiTenancyEnabled;
```
arjunkumargiri, line 84: Why does this need to be volatile?

plugin/src/main/java/org/opensearch/ml/action/agents/GetAgentTransportAction.java
dhrubo-os: Addressed : https://github.com/opensearch-project/ml-commons/blob/ee1181c93de55a7c0c1509586e07c0aeebcbfe20/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java#L120-L122

arjunkumargiri (approved):

Issue #2649, opened 2024-07-15
[BUG] ChatAgentRunner extract final_answer incorrectly when having wrapping lines


qianheng-aws:
**What is the bug?**
When the response of LLM is a string of wrapping lines, it extract the final answer incorrectly with redundant characters **`" }`**

```
Response of LLM:

The alert has been successfully created based on the input provided. The key details are:
```json
{
    "thought": "Now I know the final answer",
    "final_answer": "The alert ‘Error Response Alert’ has been created with the following configuration:
- Indices: opensearch_dashboards_sample_data_logs
- Time field: timestamp
- Bucket interval: 60 minutes
- Filters: response is not 200
- Aggregation: Count of bytes field
- Trigger: ‘Error Response Count Above 50’ - severity 1, threshold 50 (above)
This alert will trigger when the error response count exceeds 50 per 60 minute interval."
}
```. 

Final Answer:

The alert ‘Error Response Alert’ has been created with the following configuration:
- Indices: opensearch_dashboards_sample_data_logs
- Time field: timestamp
- Bucket interval: 60 minutes
- Filters: response is not 200
- Aggregation: Count of bytes field
- Trigger: ‘Error Response Count Above 50’ - severity 1, threshold 50 (above)
This alert will trigger when the error response count exceeds 50 per 60 minute interval.”
}

```

**How can one reproduce the bug?**
It's flaky and depends on whether the response string is compatible with  `org.opensearch.ml.common.utils.StringUtils.isJson`. In my response above, the method `isJson` returns false when calling `AgentUtils::parseThoughtResponse` as it has wrapping lines. Then it fallback to `extractFinalAnswer` to get final answer, but the match pattern is 
``` String pattern = "\"final_answer\"\\s*:\\s*\"(.*?)$";``` which leads to match `" }` as well in the result.

**What is the expected behavior?**
The final answer should be:
```
The alert ‘Error Response Alert’ has been created with the following configuration:
- Indices: opensearch_dashboards_sample_data_logs
- Time field: timestamp
- Bucket interval: 60 minutes
- Filters: response is not 200
- Aggregation: Count of bytes field
- Trigger: ‘Error Response Count Above 50’ - severity 1, threshold 50 (above)
This alert will trigger when the error response count exceeds 50 per 60 minute interval.
```

**What is your host/environment?**
 - OS: [e.g. macos]
 - Version [e.g. 3.0.0-SNAPSHOT]
 - Plugins: alert, ml-commons, skills

[...]

PR #2676, opened 2024-07-18
Fix final answer with extra meaningless symbol

qianheng-aws:
### Description
Fix final answer with extra meaningless symbol like `"}````
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2649
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -295,7 +295,7 @@ public class AgentUtils {
     public static String extractFinalAnswer(String text) {
         String result = null;
         if (text.contains("\"final_answer\"")) {
-            String pattern = "\"final_answer\"\\s*:\\s*\"(.*?)$";
+            String pattern = "\"final_answer\"\\s*:\\s*\"(.*)\"";
             Pattern jsonBlockPattern = Pattern.compile(pattern, Pattern.DOTALL);
             Matcher jsonBlockMatcher = jsonBlockPattern.matcher(text);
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -295,7 +295,7 @@ public static void parseThoughtResponse(Map<String, String> modelOutput, String
     public static String extractFinalAnswer(String text) {
         String result = null;
         if (text.contains("\"final_answer\"")) {
-            String pattern = "\"final_answer\"\\s*:\\s*\"(.*?)$";
+            String pattern = "\"final_answer\"\\s*:\\s*\"(.*)\"";
```
ylwu-amzn, line 298: What if the final answer has some " inside it ,for example
```
"final_answer": "This is a \"test answer\""
```

qianheng-aws: It will match `This is a \"test answer\"` since it use `.*` which is a greedy mode instead of `.*?`.

I'll add a new unit test to verify this case.

ylwu-amzn (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
ylwu-amzn: please test os assistant and chat experience, make sure this change will not break it

xinyual (approved):

yuye-aws:
Backport to 2.x?

PR #2686, opened 2024-07-19
applied multi tenancy + sdk client for search actions

dhrubo-os:
### Description
[applied multi tenancy + sdk client for search actions]
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
@@ -1,0 +1,65 @@
+package org.opensearch.ml.common.transport.search;
+
+import lombok.Builder;
+import lombok.Getter;
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.core.common.io.stream.InputStreamStreamInput;
+import org.opensearch.core.common.io.stream.OutputStreamStreamOutput;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+@Getter
+public class MLSearchActionRequest extends SearchRequest {
+    SearchRequest searchRequest;
+    String tenantId;
+
+    @Builder
+    public MLSearchActionRequest(SearchRequest searchRequest, String tenantId) {
+        this.searchRequest = searchRequest;
+        this.tenantId = tenantId;
+    }
+
+    public MLSearchActionRequest(StreamInput input) throws IOException {
+        super(input);
+        if (input.readBoolean()) {
+            searchRequest = new SearchRequest(input);
+        }
+        this.tenantId = input.readOptionalString();
+    }
+
+    @Override
+    public void writeTo(StreamOutput output) throws IOException {
+        super.writeTo(output);
+
+        if (searchRequest != null) {
+            output.writeBoolean(true); // user exists
+            searchRequest.writeTo(output);
+        } else {
+            output.writeBoolean(false); // user does not exist
+        }
+        output.writeOptionalString(tenantId);
+    }
+
+
+    public static MLSearchActionRequest fromActionRequest(ActionRequest actionRequest) {
+        if (actionRequest instanceof MLSearchActionRequest) {
+            return (MLSearchActionRequest)actionRequest;
+        }
+
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+             OutputStreamStreamOutput osso = new OutputStreamStreamOutput(baos)) {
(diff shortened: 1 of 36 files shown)


plugin/src/main/java/org/opensearch/ml/rest/AbstractMLSearchAction.java
```diff
@@ -46,12 +58,17 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
         searchSourceBuilder.parseXContent(request.contentOrSourceParamParser());
         searchSourceBuilder.fetchSource(getSourceContext(request, searchSourceBuilder));
         searchSourceBuilder.seqNoAndPrimaryTerm(true).version(true);
+        if (mlFeatureEnabledSetting.isMultiTenancyEnabled()) {
+            // Add tenant ID filter
+            String tenantId = getTenantID(mlFeatureEnabledSetting.isMultiTenancyEnabled(), request);
+            searchSourceBuilder.query(QueryBuilders.termQuery(TENANT_ID, tenantId));
+        }
         SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(index);
         return channel -> client.execute(actionType, searchRequest, search(channel));
```
dbwiddis, line 67: I'm really not fond of putting logic in the Rest Handlers, as the `MachineLearningClient` doesn't call them; it directly invokes the relevant transport action.

In particular, while this works for the ML Commons REST API, we won't be able to search for an agent from Flow Framework or any other external plugin that consumes the MLClient; we'd essentially have to duplicate all this code on our side to use the search API.

Since Rest Handlers don't benefit from injection, this also means passing more parameters around in constructors.

We're already a ways down this road with all the other logic building the query here, so for this PR I guess it's OK, but this really should be refactored into a transport action where the feature enabled settings can be injected and it can be called from a client.

plugin/src/main/java/org/opensearch/ml/rest/RestMLSearchConnectorAction.java
```diff
@@ -10,15 +10,22 @@
...
 
 public class RestMLSearchConnectorAction extends AbstractMLSearchAction<Connector> {
     private static final String ML_SEARCH_CONNECTOR_ACTION = "ml_search_connector_action";
     private static final String SEARCH_CONNECTOR_PATH = ML_BASE_URI + "/connectors/_search";
 
-    public RestMLSearchConnectorAction() {
-        super(ImmutableList.of(SEARCH_CONNECTOR_PATH), ML_CONNECTOR_INDEX, Connector.class, MLConnectorSearchAction.INSTANCE);
+    public RestMLSearchConnectorAction(MLFeatureEnabledSetting mlFeatureEnabledSetting) {
+        super(
+            ImmutableList.of(SEARCH_CONNECTOR_PATH),
```
dbwiddis, line 23: (Observation) Java has had `List.of()` since JDK9.  I wonder if you can get rid of the whole Guava dependency with a global search and replace.

dbwiddis (approved):
Overall LGTM.

A few comments that might want to be addressed in a future PR.

plugin/src/main/java/org/opensearch/ml/rest/AbstractMLSearchAction.java
dhrubo-os: that's a good point. In the transport layer, we can check again, if `mlFeatureEnabledSetting.isMultiTenancyEnabled()` is true, we should expect tenant filtering in the search request. Otherwise throw exception?

dbwiddis: Yes.  Problem is such an action doesn’t yet exist.   I’m ok with this as is with an open issue to fix it long term.

dhrubo-os: I added the check in the transport search layer. Could you please review again?

arjunkumargiri: Rather than adding the term query check in rest/transport actions, can it be added as part of sdkclient? this will prevent code bugs by ensuring new code flows does not miss out on adding the check by implementing the term condition at the lowest layer.

dbwiddis (approved):
(Still) LGTM, minor (some might say trivial) style notes.

common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            output.writeBoolean(false); // user does not exist
+        }
+        output.writeOptionalString(tenantId);
+    }
+
+
+    public static MLSearchActionRequest fromActionRequest(ActionRequest actionRequest) {
+        if (actionRequest instanceof MLSearchActionRequest) {
+            return (MLSearchActionRequest)actionRequest;
```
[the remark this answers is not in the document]
dhrubo-os: address that in the next PR.

PR #2394, opened 2024-05-01
Add IT for interface

b4sjoo:
### Description
Add IT for interface

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2689, opened 2024-07-20
Automated model interface generation on aws llms

b4sjoo:
### Description
This PR will enable automated model interface generation on AWS managed llms (Bedrock models), to unblock Fractal team
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/utils/ModelInterfaceUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/ModelInterfaceUtils.java
@@ -556,16 +556,16 @@ public class ModelInterfaceUtils {
 
     private static Map<String, String> createPresetModelInterfaceByConnector(Connector connector) {
         if (connector.getParameters() != null) {
-            switch (connector.getParameters().get("service_name")) {
+            switch ((connector.getParameters().get("service_name") != null) ? connector.getParameters().get("service_name") : "null") {
                 case "bedrock":
                 log.info("Creating preset model interface for Amazon Bedrock model: {}", connector.getParameters().get("model"));
-                switch (connector.getParameters().get("model")) {
+                switch ((connector.getParameters().get("model") != null) ? connector.getParameters().get("model") : "null") {
                     case "ai21.j2-mid-v1":
                         return BEDROCK_AI21_LABS_JURASSIC2_MID_V1_MODEL_INTERFACE;
                     case "anthropic.claude-3-sonnet-20240229-v1:0":
                         return BEDROCK_ANTHROPIC_CLAUDE_V3_SONNET_MODEL_INTERFACE;
                     case "anthropic.claude-v2":
-                        return  BEDROCK_ANTHROPIC_CLAUDE_V2_MODEL_INTERFACE;
+                        return BEDROCK_ANTHROPIC_CLAUDE_V2_MODEL_INTERFACE;
                     case "cohere.embed.english-v3":
                         return BEDROCK_COHERE_EMBED_ENGLISH_V3_MODEL_INTERFACE;
                     case "cohere.embed.multilingual-v3":
@@ -579,7 +579,7 @@ public class ModelInterfaceUtils {
                 }
                 case "comprehend":
                     log.info("Creating preset model interface for Amazon Comprehend DetectDominantLanguage API");
-                    switch (connector.getParameters().get("api_name")) {
+                    switch ((connector.getParameters().get("api_name") != null) ? connector.getParameters().get("api_name") : "null"){
                         case "DetectDominantLanguage":
                             return AMAZON_COMPREHEND_DETECTDOMAINANTLANGUAGE_API_INTERFACE;
                         default:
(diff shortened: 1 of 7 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestMLGuardrailsIT.java
```diff
@@ -423,7 +427,11 @@ protected Response registerRemoteModelWithLocalRegexGuardrails(String name, Stri
             + "      ],\n"
             + "      \"regex\": [\"regex1\", \"regex2\"]\n"
             + "    }\n"
-            + "  }\n"
+            + "},\n"
```
mingshl, line 430: the IT only covers the general model general interface, please add the IT for aws llm interface later

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -239,7 +240,14 @@ private void doRegister(MLRegisterModelInput registerModelInput, ActionListener<
             if (Strings.isNotBlank(registerModelInput.getConnectorId())) {
                 connectorAccessControlHelper.validateConnectorAccess(client, registerModelInput.getConnectorId(), ActionListener.wrap(r -> {
                     if (Boolean.TRUE.equals(r)) {
-                        createModelGroup(registerModelInput, listener);
+                        if (registerModelInput.getModelInterface() == null) {
+                            mlModelManager.getConnector(registerModelInput.getConnectorId(), ActionListener.wrap(connector -> {
```
ylwu-amzn, line 244: Should we check if it's internal connector inside model ?

plugin/src/test/java/org/opensearch/ml/rest/RestMLGuardrailsIT.java
b4sjoo: This is to resolve IT failure, we have complete interface IT: https://github.com/opensearch-project/ml-commons/pull/2394

common/src/main/java/org/opensearch/ml/common/utils/ModelInterfaceUtils.java
```diff
@@ -0,0 +1,623 @@
...
+import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.connector.Connector;
+import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
+
+import java.util.Map;
+
+@Log4j2
+public class ModelInterfaceUtils {
+
+    private static final String GENERAL_CONVERSATIONAL_MODEL_INTERFACE_INPUT = "{\n" +
```
dhrubo-os, line 18: Should we maintain any kind of version for Model Interface? What happens if we want to add another field to the model interface later?

rbhavna (approved):

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -239,7 +240,14 @@ private void doRegister(MLRegisterModelInput registerModelInput, ActionListener<
             if (Strings.isNotBlank(registerModelInput.getConnectorId())) {
                 connectorAccessControlHelper.validateConnectorAccess(client, registerModelInput.getConnectorId(), ActionListener.wrap(r -> {
                     if (Boolean.TRUE.equals(r)) {
-                        createModelGroup(registerModelInput, listener);
+                        if (registerModelInput.getModelInterface() == null) {
```
dhrubo-os, line 243: are we expecting a new field in model index? I don't see any declaration of this new field? Or did you already do that in any other PR?

plugin/src/test/java/org/opensearch/ml/rest/RestMLGuardrailsIT.java
```diff
@@ -366,7 +366,11 @@ protected Response registerRemoteModel(String modelGroupName, String name, Strin
             + "  \"description\": \"test model\",\n"
             + "  \"connector_id\": \""
             + connectorId
-            + "\"\n"
+            + "\",\n"
+            + "  \"interface\": {\n"
```
dhrubo-os, line 370: Is that a new field in the connector index?

b4sjoo: See above, it's in model index

common/src/main/java/org/opensearch/ml/common/utils/ModelInterfaceUtils.java
b4sjoo: It's will be bounded by blueprints in the future. So if we want add another field we just directly introduce a new variable

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
dhrubo-os, line 244: can't you stash thread context here before invoking this method?

b4sjoo: Well yeah this is something we can try

Issue #2630, opened 2024-07-09
[BUG] bump ML config index schema version


ylwu-amzn:
**What is the bug?**
ML config index mapping updated, but schema version not bumped. This PR https://github.com/opensearch-project/ml-commons/pull/1995 updated

dhrubo-os:
Same for `.plugins-ml-config` index. 

In 2.9 index mapping definition is:

```
{
  ".plugins-ml-config" : {
    "mappings" : {
      "_meta" : {
        "schema_version" : 2
      },
      "properties" : {
        "create_time" : {
          "type" : "date",
          "format" : "strict_date_time||epoch_millis"
        },
        "master_key" : {
          "type" : "keyword"
        }
      }
    }
  }
}
```

and in 2.13 it is:

```
{
  ".plugins-ml-config" : {
    "mappings" : {
      "dynamic" : "strict",
      "_meta" : {
        "schema_version" : 2
      },
      "properties" : {
        "configuration" : {
          "type" : "flat_object"
        },
        "create_time" : {
          "type" : "date",
          "format" : "strict_date_time||epoch_millis"
        },
        "last_update_time" : {
          "type" : "date",
          "format" : "strict_date_time||epoch_millis"
        },
        "master_key" : {
          "type" : "keyword"
        },
        "type" : {
          "type" : "keyword"
        }
      }
    }
  }
}
```

ylwu-amzn:
Sorry my bad, it should be ML config index, not model group .

dhrubo-os:
I found another PR: https://github.com/opensearch-project/ml-commons/pull/2276/files where version didn't get bumped.

PR #2710, opened 2024-07-23
update config index mappings to use correct field types

rbhavna:
### Description
update config index mappings to use correct field types.

In 2.13, config index mapping was updated with three new fields `configuration`, `type`, and `last_update_time`but schema version was not bumped. Due to which, the domains which were upgraded to 2.13 from previous versions did not go through update index mapping as they did not detect a schema version bump. 

During indexing, these fields got created with wrong types not as specified in the mapping. Now when such domains are going to upgrade to 2.15, with schema_version bump, the index mapping tries to get updated but will throw exception since the field types for the three fields specified above are different. Hence the index initialization fails and indexing into the `config` index wouldn't be possible. 

Hence, the PR tries to add complete new fields(`ml_configuration`, `config_type`, `last_updated_time` while removing the old ones. So whenever mapping is getting updated, since there is no type-conflict for old fields, they get retained in the index as is while new fields get added. It is therefore suggested to start using these new fields from 2.15 while indexing into config index.
 
### Issues Resolved

https://github.com/opensearch-project/ml-commons/issues/2630
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLConfig.java
@@ -111,6 +111,7 @@ public class MLConfig implements ToXContentObject, Writeable {
         Configuration mlConfiguration = null;
         Instant createTime = null;
         Instant lastUpdateTime = null;
+        Instant lastUpdatedTime = null;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -136,6 +137,9 @@ public class MLConfig implements ToXContentObject, Writeable {
                 case LAST_UPDATE_TIME_FIELD:
                     lastUpdateTime = Instant.ofEpochMilli(parser.longValue());
                     break;
+                case LAST_UPDATED_TIME_FIELD:
+                    lastUpdatedTime = Instant.ofEpochMilli(parser.longValue());
+                    break;
                 default:
                     parser.skipChildren();
                     break;
@@ -145,7 +149,7 @@ public class MLConfig implements ToXContentObject, Writeable {
                 .type(configType == null ? type : configType)
                 .configuration(mlConfiguration == null ? configuration : mlConfiguration)
                 .createTime(createTime)
-                .lastUpdateTime(lastUpdateTime)
+                .lastUpdateTime(lastUpdatedTime == null ? lastUpdateTime : lastUpdatedTime)
                 .build();
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/MLConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLConfig.java
@@ -27,10 +27,15 @@ public class MLConfig implements ToXContentObject, Writeable {
 
     public static final String TYPE_FIELD = "type";
 
+    public static final String CONFIG_TYPE_FIELD = "config_type";
+
     public static final String CONFIGURATION_FIELD = "configuration";
+
+    public static final String ML_CONFIGURATION_FIELD = "ml_configuration";
 
     public static final String CREATE_TIME_FIELD = "create_time";
     public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
+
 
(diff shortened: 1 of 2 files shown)


jngz-es:
Hi @rbhavna , I don't quite get why we change the field names for the type and configuration. Could you explain a little bit?

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -112,9 +119,15 @@ public static MLConfig parse(XContentParser parser) throws IOException {
                 case TYPE_FIELD:
                     type = parser.text();
                     break;
+                case CONFIG_TYPE_FIELD:
+                    configType = parser.text();
+                    break;
                 case CONFIGURATION_FIELD:
                     configuration = Configuration.parse(parser);
                     break;
+                case ML_CONFIGURATION_FIELD:
```
ylwu-amzn, line 128: Another field added in 2.13, `LAST_UPDATE_TIME_FIELD`, let's use similar way to fix?

rbhavna:
> Hi @rbhavna , I don't quite get why we change the field names for the type and configuration. Could you explain a little bit?

added details to PR description

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -27,11 +27,18 @@ public class MLConfig implements ToXContentObject, Writeable {
...
+    public static final String CONFIG_TYPE_FIELD = "config_type";
+
     public static final String CONFIGURATION_FIELD = "configuration";
 
+    public static final String ML_CONFIGURATION_FIELD = "ml_configuration";
+
     public static final String CREATE_TIME_FIELD = "create_time";
     public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
 
+    public static final String LAST_UPDATED_TIME_FIELD = "last_updated_time";
```
Zhangxunmt, line 39: Is it possible to reuse the old Field names for CX experience BWC, rather than defining new ones?

Zhangxunmt (approved):

ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/MLConfig.java
rbhavna: Using old names and trying to update their types is failing. OpenSearch does not allow conversion between certain types and flat_object is one of them

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -78,10 +85,10 @@ public void writeTo(StreamOutput out) throws IOException {
     public XContentBuilder toXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
         XContentBuilder builder = xContentBuilder.startObject();
         if (type != null) {
-            builder.field(TYPE_FIELD, type);
+            builder.field(CONFIG_TYPE_FIELD, type);
```
ylwu-amzn, line 88: Before this PR
```
    "type" : "os_query_assist_ppl_agent",
    "configuration" : {
      "agent_id" : "qdFHuZEBJv7K-ogaOTOB"
    }
```
Now we have
```
    "config_type" : "os_query_assist_ppl_agent",
    "ml_configuration" : {
      "agent_id" : "qdFHuZEBJv7K-ogaOTOB"
    }
```
This is a breaking

ylwu-amzn: Same for line 91, that's also breaking

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -112,25 +122,34 @@ public static MLConfig parse(XContentParser parser) throws IOException {
...
+                    break;
                 default:
                     parser.skipChildren();
                     break;
             }
         }
         return MLConfig.builder()
-                .type(type)
-                .configuration(configuration)
+                .type(configType == null ? type : configType)
```
ylwu-amzn, line 149:
```
.type(type)
.configType(configType)
```

Issue #2712, opened 2024-07-23
[BUG] KnowledgeBaseTool may return parameter placeholder not filled error


yuye-aws:
**What is the bug?**
I am implementing the the knowledge base tool. Unfortunately, I meet with a tricky bug from the connector. I did some deep dive and found the error is originating from `validatePayload` function from `Connector.java`.

**How can one reproduce the bug?**
Prerequisites: configure LLM and bedrock knowledge base, and then obtain their `model_id`.

1. Register the agent with knowledge base tool.
```
POST /_plugins/_ml/agents/_register
{
  "name": "Test_Agent_For_ReAct_KB",
  "type": "conversational",
  "description": "this is a test agent for knowledge base",
  "llm": {
    "model_id": "rBZl3pABQHtbJkH1MFP3",
    "parameters": {
      "max_iteration": 10,
      "stop_when_no_tool_found": true,
      "response_filter": "$.completion"
    }
  },
  "memory": {
    "type": "conversation_index"
  },
  "tools": [
    {
      "type": "KnowledgeBaseTool",
      "name": "OpensearchKnowledgeBaseTool",
      "description": "A tool to search the Opensearch knowledge base, the knowledge base consists of segments of OpenSearch documents. You should always search data with this tool when encountering general questions about Opensearch. But for questions about current concerete cluster, use this tool can not help you. If this tool provides useful info, give the answer and also quote the origin doc. If this tool can not provide knowledge you need, give answer based on your own knowledge. Action Input: {\"text\": <natrual language keywords for question>}",
      "parameters": {
        "model_id": "phZk3pABQHtbJkH1fVPy"
      }
    },
    {
      "type": "CatIndexTool",
      "name": "RetrieveIndexMetaTool",
      "description": "Use this tool to get OpenSearch index information: (health, status, index, uuid, primary count, replica count, docs.count, docs.deleted, store.size, primary.store.size)."
    }
  ],
  "app_type": "my app"
}

```

2. Execute the agent to retrieve documents on "agent framework".
```
POST /_plugins/_ml/agents/_haq3pABQHtbJkH1eVNE/_execute
{
  "parameters": {
    "question": "Use OpensearchKnowledgeBaseTool to get information about agent framework in opensearch",
    "verbose": true
  }
}
```

3. Receive the following error:
```
[...]

PR #2724, opened 2024-07-24
enhance: support recursive_parameter_enabled in connector

yuye-aws:
### Description
When using LLM in ml-commons connector, the prompt may contains substring like `${parameters.xxxx} `. The PRs adds a parameter named `recursive_parameter_enabled` to enable users to configure whether to recursively replace the parameter in payload.

Here are an example to show why skipping parameter validation won't work. Suppose that the current parameter map is `{"prompt": "You are a ${parameters.role}", "role": "bot"}` the payload creation is ambiguous. The current connector creation consists of two steps:

`${parameters.prompt} ->  You are a ${parameters.role} ->  You are a bot.
`

The user might also wish to send the  `You are a ${parameters.role}` to the remote model. If we simply skip parameter validation, there would be no way for user to control the behavior.

We thus need a parameter `recursive_parameter_enabled` for users to configure the behavior of filling in parameters. The user can specify this parameter in the following manner:
```
{
  "name": "BedRock test connector",
  "description": "The connector to BedRock service for claude model",
  "version": 1,
  "protocol": "aws_sigv4",
  "parameters": {
      "region": "us-east-1",
      "service_name": "bedrock",
      "anthropic_version": "bedrock-2023-05-31",
      "endpoint": "bedrock-runtime.us-east-1.amazonaws.com",
      "auth": "Sig_V4",
      "content_type": "application/json",
      "max_tokens_to_sample": 8000,
      "temperature": 0.00001,
      "recursive_parameter_enabled": true
  },
  "credential": {
    "access_key": "xxxx",
    "secret_key": "xxxx"
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "url": "https://bedrock-runtime.us-east-1.amazonaws.com/model/anthropic.claude-instant-v1/invoke",
      "headers": { 
        "x-amz-content-sha256": "required"
      },
      "request_body": "{\"prompt\":\"${parameters.prompt}\", \"max_tokens_to_sample\":${parameters.max_tokens_to_sample}, \"temperature\":${parameters.temperature},  \"anthropic_version\":\"${parameters.anthropic_version}\" }"
    }
  ]
}
```

### Issues Resolved
Resolve https://github.com/opensearch-project/ml-commons/issues/2712
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -94,7 +94,7 @@ public interface Connector extends ToXContentObject, Writeable {
         return requiredParameters;
     }
 
-    default void validatePayload(Set<String> requiredParameters, Map<String, String> parameters) {
+    default void validateParameters(Set<String> requiredParameters, Map<String, String> parameters) {
         StringBuilder errorBuilder = new StringBuilder();
         for (String requiredParameter : requiredParameters) {
             if (!parameters.containsKey(requiredParameter)) {
@@ -103,6 +103,21 @@ public interface Connector extends ToXContentObject, Writeable {
         }
         if (!errorBuilder.isEmpty()) {
             String error = errorBuilder.substring(0, errorBuilder.length() - 2);
+            throw new IllegalArgumentException("Some parameter placeholder not filled in payload: " + error);
+        }
+    }
+
+    default void validatePayload(String payload) {
+        if (payload != null && payload.contains("${parameters")) {
+            Pattern pattern = Pattern.compile("\\$\\{parameters\\.([^}]+)}");
+            Matcher matcher = pattern.matcher(payload);
+
+            StringBuilder errorBuilder = new StringBuilder();
+            while (matcher.find()) {
+                String parameter = matcher.group(1);
+                errorBuilder.append(parameter).append(", ");
+            }
+            String error = errorBuilder.substring(0, errorBuilder.length() - 2).toString();
             throw new IllegalArgumentException("Some parameter placeholder not filled in payload: " + error);
         }
     }
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -10,9 +10,11 @@ import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -60,7 +62,9 @@ public interface Connector extends ToXContentObject, Writeable {
 
     String getActionHttpMethod(String action);
 
-    <T> T createPayload(String action, Map<String, String> parameters);
+    <T> T createRawPayload(String action);
+
+    <T> T fillInPayload(String payload, Map<String, String> parameters);
 
     void decrypt(String action, Function<String, String> function);
     void encrypt(Function<String, String> function);
@@ -77,17 +81,28 @@ public interface Connector extends ToXContentObject, Writeable {
 
     <T> void parseResponse(T orElse, List<ModelTensor> modelTensors, boolean b) throws IOException;
 
-    default void validatePayload(String payload) {
+    default Set<String> getRequiredParameters(String payload) {
+        Set<String> requiredParameters = new HashSet<>();
         if (payload != null && payload.contains("${parameters")) {
             Pattern pattern = Pattern.compile("\\$\\{parameters\\.([^}]+)}");
             Matcher matcher = pattern.matcher(payload);
-
-            StringBuilder errorBuilder = new StringBuilder();
             while (matcher.find()) {
                 String parameter = matcher.group(1);
-                errorBuilder.append(parameter).append(", ");
+                requiredParameters.add(parameter);
             }
-            String error = errorBuilder.substring(0, errorBuilder.length() - 2).toString();
+        }
+        return requiredParameters;
+    }
+
+    default void validatePayload(Set<String> requiredParameters, Map<String, String> parameters) {
+        StringBuilder errorBuilder = new StringBuilder();
+        for (String requiredParameter : requiredParameters) {
+            if (!parameters.containsKey(requiredParameter)) {
+                errorBuilder.append(requiredParameter).append(", ");
+            }
+        }
+        if (!errorBuilder.isEmpty()) {
+            String error = errorBuilder.substring(0, errorBuilder.length() - 2);
             throw new IllegalArgumentException("Some parameter placeholder not filled in payload: " + error);
         }
     }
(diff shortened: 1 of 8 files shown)


yuye-aws:
Please hold on this PR for a while. We are looking for a better solution

yuye-aws:
Reimplemented this PR with idea from RFC: https://github.com/opensearch-project/ml-commons/issues/2748

yuye-aws:
This PR still needs more discussion. I can fix UT later.

yuye-aws:
Can we add backport 2.x?

yuye-aws:
This PR is ready for review.

common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java
```diff
@@ -189,6 +189,15 @@ public void createPayload() {
         Assert.assertEquals("{\"input\": \"test input value\"}", predictPayload);
     }
 
+    @Test
+    public void createPayload_Skip_Parameter_Validation() {
+        exceptionRule.expect(IllegalArgumentException.class);
+        exceptionRule.expectMessage("Some parameter placeholder not filled in payload: input");
```
ylwu-amzn, line 195: Why we expect to see this error with `skip_validating_missing_parameters` as `true`

common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java
```diff
@@ -189,6 +189,15 @@ public void createPayload() {
...
+        HttpConnector connector = createHttpConnector();
+        String predictPayload = connector.createPayload(PREDICT.name(), Map.of("skip_validating_missing_parameters", "true"));
```
ylwu-amzn, line 197: Please add a test which not throw exception with `skip_validating_missing_parameters` as true.

For example, "promote": "hello ${parameters.foo}"

yuye-aws: Ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -189,7 +190,9 @@ default void preparePayloadAndInvoke(
         // override again to always prioritize the input parameter
         parameters.putAll(inputParameters);
         String payload = connector.createPayload(action, parameters);
-        connector.validatePayload(payload);
+        if (!Boolean.parseBoolean(parameters.getOrDefault(SKIP_VALIDATE_MISSING_PARAMETERS, "false"))) {
```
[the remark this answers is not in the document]
ylwu-amzn: I don't think this is a big concern. Just like this blueprint https://github.com/opensearch-project/ml-commons/blob/main/docs/remote_inference_blueprints/cohere_connector_embedding_blueprint.md , user should tune `input_type` and `truncate` based on their use case.

We are not setting the expectation that user can create one connector with a default set of parameters, then it can be used for all use cases.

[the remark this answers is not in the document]
dhrubo-os: but if connector can be used with dynamic parameters, that would be the best. Let's say for the same connector may be in one agent, `skip_validating_missing_parameters` can be true, but for other agent this can be false?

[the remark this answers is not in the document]
Zhangxunmt: +1, any parameter in the connector can be overwritten later in the "Predict" request. We can treat the parameters in the Connector as default values that are subject to change.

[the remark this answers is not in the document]
yuye-aws: Here is our conclusion today: https://github.com/opensearch-project/ml-commons/issues/2748#issuecomment-2259515610

common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java
yuye-aws: Added new UT in ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java

common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java
yuye-aws: Added new UT in ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
```diff
@@ -77,21 +81,37 @@ public interface Connector extends ToXContentObject, Writeable {
...
+                errorBuilder.append(requiredParameter).append(", ");
+            }
+        }
+        if (!errorBuilder.isEmpty()) {
+            String error = errorBuilder.substring(0, errorBuilder.length() - 2);
             throw new IllegalArgumentException("Some parameter placeholder not filled in payload: " + error);
         }
     }
 
+    default void validatePayload(String payload) {
```
zane-neo, line 110: This default method doesn't make sense, please change this to an abstract method.

zane-neo (commented):
Any IT to cover this?

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
yuye-aws: Sorry for not getting your point. This method is used when parameter RECURSIVE_PARAMETER_ENABLED is true, which is the default behavior.

yuye-aws:
> Any IT to cover this?

This PR already has an UT file to cover when RECURSIVE_PARAMETER_ENABLED is false, true or missing. What do you expect to test from the IT?

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
zane-neo: When adding a new default function, basically you assume that this function will be used in multiple sub-classes, and only a few overrides it, this doesn't seem the case since only one place using it.
Also, there's already the `validateParameters` method which is basically same function with this method, it's redundant to add a similar function in interface layer.

zane-neo:
> > Any IT to cover this?
> 
> This PR already has an UT file to cover when RECURSIVE_PARAMETER_ENABLED is false, true or missing. What do you expect to test from the IT?

I was expecting when the parameter is set to true/false/missing, the http response are as expected.

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
yuye-aws: Ack. Will update this.

yuye-aws:
> > > Any IT to cover this?
> > 
> > 
> > This PR already has an UT file to cover when RECURSIVE_PARAMETER_ENABLED is false, true or missing. What do you expect to test from the IT?
> 
> I was expecting when the parameter is set to true/false/missing, the http response are as expected.

IT added in `RestMLRemoteInferenceIT.java`. Please help review, thanks!

yuye-aws:
Closing this PR as I have implemented a new solution in https://github.com/opensearch-project/ml-commons/pull/2812

PR #2732, opened 2024-07-24
excluding remote models from max node per node setting

dhrubo-os:
### Description

Currently we have a settings `plugins.ml_commons.max_model_on_node` how many models we can load in a single node. But this shouldn't include Remote models. So excluding remote model from that consideration.

### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -990,7 +990,7 @@ public class MLModelManager {
         mlStats.createModelCounterStatIfAbsent(modelId, ActionName.DEPLOY, ML_ACTION_REQUEST_COUNT).increment();
         List<String> workerNodes = mlTask.getWorkerNodes();
         if (modelCacheHelper.isModelDeployed(modelId)) {
-            if (!autoDeployModel && workerNodes != null && workerNodes.size() > 0) {
+            if (!autoDeployModel && workerNodes != null && !workerNodes.isEmpty()) {
                 log.info("Set new target node ids {} for model {}", Arrays.toString(workerNodes.toArray(new String[0])), modelId);
                 modelCacheHelper.setDeployToAllNodes(modelId, deployToAllNodes);
                 modelCacheHelper.setTargetWorkerNodes(modelId, workerNodes);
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -999,7 +999,7 @@ public void deployModel(
             listener.onResponse("successful");
             return;
         }
-        if (modelCacheHelper.getLocalDeployedModels().length >= maxModelPerNode) {
+        if (functionName != FunctionName.REMOTE && modelCacheHelper.getLocalDeployedModels().length >= maxModelPerNode) {
```
ylwu-amzn, line 1002: Thanks, have you checked if any other settings have similar issues?

dhrubo-os: There are few task related settings like `plugins.ml_commons.max_register_model_tasks_per_node`, `plugins.ml_commons.max_deploy_model_tasks_per_node`, `"plugins.ml_commons.max_ml_task_per_nod`. But I don't think that should cause any issue.

ylwu-amzn: I don't think user will trigger a lot of register and depoy model tasks. But for `plugins.ml_commons.max_ml_task_per_node`, can you check if this limit how many predict tasks can run in parallel on one node ?

ylwu-amzn:
One minor suggestion: the commit message `excluding remote models` is too general , suggest make it more clear like `excluding remote models from max node per node setting`

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
dhrubo-os: From the [code](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/task/MLTaskDispatcher.java#L131), it looks like we check this only when dispatch policy is `least_load` , not `round_robin`. By default, task dispatch policy is `round_robin`. I think this should be fine for now.

dhrubo-os:
> One minor suggestion: the commit message `excluding remote models` is too general , suggest make it more clear like `excluding remote models from max node per node setting`

addressed.

ylwu-amzn (approved):

Zhangxunmt (approved):

ylwu-amzn:
@dhrubo-os I think we should publish PR to main branch first. Then backport to 2.x and release branch 2.16

PR #2759, opened 2024-07-26
Add XContentType to wrap the CreateIndexRequest mappings in _doc key to fix v1 templates issue

rbhavna:
### Description
Adds the _doc key wrapper around index mapping as required by CreateIndexRequest javadoc.

### Related Issues
See also https://github.com/opensearch-project/OpenSearch/issues/14984

### Testing Details
When the cluster has v1 templates with patterns matching the index, without the _doc wrapper, dynamic mappings are getting enabled for the index. From below, created_time is being created as long (dynamically picked) instead of `timestamp` type which is defined in the [mapping definition](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/CommonValue.java#L138).

Screenshot 2024-07-26 at 1 54 05 PM

With changes in this PR by adding `_doc` wrapper, even if cluster has v1 templates, indices are getting created with expected mappings.

Screenshot 2024-07-26 at 2 17 26 PM

### Check List

### DIFF
--- a/memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
@@ -87,7 +87,7 @@ public class ConversationMetaIndex {
             log.debug("No conversational meta index found. Adding it");
             CreateIndexRequest request = Requests
                 .createIndexRequest(META_INDEX_NAME)
-                .mapping(ConversationalIndexConstants.META_MAPPING)
+                .mapping("{\"_doc\":" + ConversationalIndexConstants.META_MAPPING + "}")
                 .settings(INDEX_SETTINGS);
             try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
                 ActionListener<Boolean> internalListener = ActionListener.runBefore(listener, () -> threadContext.restore());
--- a/memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
+++ b/memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
@@ -46,6 +46,7 @@ import org.opensearch.client.Client;
 import org.opensearch.client.Requests;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.ConfigConstants;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
@@ -87,7 +88,7 @@ public class ConversationMetaIndex {
             log.debug("No conversational meta index found. Adding it");
             CreateIndexRequest request = Requests
                 .createIndexRequest(META_INDEX_NAME)
-                .mapping("{\"_doc\":" + ConversationalIndexConstants.META_MAPPING + "}")
+                .mapping(ConversationalIndexConstants.META_MAPPING, XContentType.JSON)
                 .settings(INDEX_SETTINGS);
             try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
                 ActionListener<Boolean> internalListener = ActionListener.runBefore(listener, () -> threadContext.restore());
--- a/release-notes/opensearch-ml-common.release-notes-2.16.0.0.md
+++ b/release-notes/opensearch-ml-common.release-notes-2.16.0.0.md
@@ -26,7 +26,6 @@ Compatible with OpenSearch 2.16.0
 * Fix yaml test issue (#2700)[https://github.com/opensearch-project/ml-commons/pull/2700]
 * Fix MLModelTool returns null if the response of LLM is a pure json object (#2675)[https://github.com/opensearch-project/ml-commons/pull/2675]
 * Bump ml config index schema version (#2656)[https://github.com/opensearch-project/ml-commons/pull/2656]
-* Wrap CreateIndexRequest mappings in _doc key in ml-commons (#2759)[https://github.com/opensearch-project/ml-commons/pull/2759]
 
 ### Maintenance
 * Upgrade djl version to 0.28.0 (#2578)[https://github.com/opensearch-project/ml-commons/pull/2578]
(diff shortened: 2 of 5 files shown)


Zhangxunmt (dismissed):

ylwu-amzn:
Have you checked any difference by wrapping index mapping with `_doc` or not ?

dbwiddis:
@rbhavna please unlink the opensearch bug as there are other impacted plugins that need to track it

rbhavna:
> Have you checked any difference by wrapping index mapping with `_doc` or not ?

Yes when v1 template is added already, without _doc wrapper I see dynamic mappings are created. Will add details to PR description

dbwiddis:
> Have you checked any difference by wrapping index mapping with `_doc` or not ?

@ylwu-amzn I've done testing with all combinations of v1 and v2 templates, with and without the _doc.   Adding _doc does not cause any additional problems and is, in fact, added by the REST API.  

I've tested all combinations (with and without _doc) and with/without v1 and v2 templates. The only failure occurs with the v1 template and lack of _doc.

@owaiskazi19 and I walked through all code paths with both deprecated v1 and current v2 mappings.  In v2 case and with no v1 mapping, code automatically determines the _doc type at run time and thus these not impacted as they are processed in a list.  The only impact is with v1 (deprecated) index templates which do a map "merge" and because the template populates the _doc field, it is now "present" and the auto-populating that occurs when there's no _doc doesn't happen and the user mapping is lost.

owaiskazi19:
@rbhavna just to be sure, can you also test with v2 templates both the scenario.

memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
```diff
@@ -87,7 +87,7 @@ public void initConversationMetaIndexIfAbsent(ActionListener<Boolean> listener)
             log.debug("No conversational meta index found. Adding it");
             CreateIndexRequest request = Requests
                 .createIndexRequest(META_INDEX_NAME)
-                .mapping(ConversationalIndexConstants.META_MAPPING)
+                .mapping("{\"_doc\":" + ConversationalIndexConstants.META_MAPPING + "}")
```
dbwiddis, line 90: FYI after further research, another possible (and less controversial) fix is using a differnet method that includes the MediaType

```suggestion
                .mapping(ConversationalIndexConstants.META_MAPPING, XContentType.JSON)
```

rbhavna:
> @rbhavna just to be sure, can you also test with v2 templates both the scenario? Also, looks like you attached the same screenshot in the description

Thanks @owaiskazi19 Updated the description. And will do more testing with v2 templates

rbhavna:
> > Have you checked any difference by wrapping index mapping with `_doc` or not ?
> 
> @ylwu-amzn I've tested all combinations (with and without _doc) and with/without v1 and v2 templates. The only failure occurs with the v1 template and lack of _doc. Adding _doc does not cause any additional problems and is, in fact, added by the REST API.
> 
> @owaiskazi19 and I walked through all code paths with both deprecated v1 and current v2 mappings. In v2 case and with no v1 mapping, code automatically determines the _doc type at run time and thus these not impacted as they are processed in a list. The only impact is with v1 (deprecated) index templates which do a map "merge" and because the template populates the _doc field, it is now "present" and the auto-populating that occurs when there's no _doc doesn't happen and the user mapping is lost.

I am actually curious if we can have similar fix for update index as well. This fix will take care of any newly created domains from 2.15. For domains upgrading from previous versions with incorrect mappings to 2.15, do you think adding this _doc wrapper helps?

memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java
rbhavna: done

rbhavna:
> @rbhavna just to be sure, can you also test with v2 templates both the scenario? Also, looks like you attached the same screenshot in the description

@owaiskazi19 Tested the v2 template scenario. If cluster has v2 template that matches the index being created, irrespective of v1 template being present or not and with/without _doc wrapper in the code, user defined mappings are created. So the issue occurs on an index only when all the below three scenarios are met

- there is no matching v2 template 
- there is a matching v1 template
- there is no _doc wrapper explicitly added to create index request

Correct me if wrong @dbwiddis

owaiskazi19 (approved):
LGTM

ylwu-amzn (approved):

owaiskazi19:
> I am actually curious if we can have similar fix for update index as well. This fix will take care of any newly created domains from 2.15. For domains upgrading from previous versions with incorrect mappings to 2.15, do you think adding this _doc wrapper helps?

If the domain had an incorrect mapping, it would have failed in the first place when ingesting docs to those indices. @dbwiddis wdyt?

release-notes/opensearch-ml-common.release-notes-2.16.0.0.md
```diff
@@ -26,6 +26,7 @@ Compatible with OpenSearch 2.16.0
 * Fix yaml test issue (#2700)[https://github.com/opensearch-project/ml-commons/pull/2700]
 * Fix MLModelTool returns null if the response of LLM is a pure json object (#2675)[https://github.com/opensearch-project/ml-commons/pull/2675]
 * Bump ml config index schema version (#2656)[https://github.com/opensearch-project/ml-commons/pull/2656]
+* Wrap CreateIndexRequest mappings in _doc key in ml-commons (#2759)[https://github.com/opensearch-project/ml-commons/pull/2759]
```
dhrubo-os, line 29: let's not add release note in the same PR.

ylwu-amzn:
The PR content can't match this commit message `Wrap CreateIndexRequest mappings in _doc key in ml-commons`

release-notes/opensearch-ml-common.release-notes-2.16.0.0.md
rbhavna: sure will remove it

rbhavna:
> The PR content can't match this commit message `Wrap CreateIndexRequest mappings in _doc key in ml-commons`
> 
> how about `add XContentType to fix v1 index template issue` ?

let me change to a meaningful one

dbwiddis:
>  So the issue occurs on an index only when all the below three scenarios are met
> 
> * there is no matching v2 template
> * there is a matching v1 template
> * there is no _doc wrapper explicitly added to create index request
> 
> Correct me if wrong @dbwiddis

Generally yes.  Some caveats:
1. the `_doc` wrapper is only required for the single-arg String json mapping.  If it adds on `, XContentType.JSON` it is ok.
2. the matching v1 template is often the `template_default_settings` which I think is how default number of shards/replicas/etc. was handled in ES 7.7 or earlier.

dbwiddis (approved):

ylwu-amzn (approved):

b4sjoo (approved):

ylwu-amzn (approved):

dhrubo-os (approved):

Issue #2502, opened 2024-06-05
[BUG] connector update API to update parameter got illegal_argument_exception when "actions.url" has a reference on parameter


chishui:
**What is the bug?**
I created a connector with payload below
```
{
    "name": "OpenAI embedding Connector",
    "description": "The connector to public OpenAI model service for GPT 3.5",
    "version": 1,
    "protocol": "http",
    "parameters": {
        "endpoint": "api.openai.com",
        "auth": "API_Key",
        "content_type": "application/json",
        "max_tokens": 7,
        "temperature": 0,
        "model": "text-embedding-ada-002"
    },
    "actions": [
        {
            "action_type": "predict",
            "method": "POST",
            "url": "https://${parameters.endpoint}/v1/embeddings",
            "pre_process_function": "connector.pre_process.openai.embedding",
            "post_process_function": "connector.post_process.openai.embedding",
            "headers": {
                "Authorization": "Bearer ${credential.openAI_key}"
            },
            "request_body": "{ \"model\": \"${parameters.model}\", \"input\": ${parameters.input}}"
        }
    ]
}
```

Then I call connector update API below to update `temperature` field in parameters
```
{	
	"parameters": {
		"temperature": 3
	}
}
```
I get an exception:
```
{
	"error": {
		"root_cause": [
			{
				"type": "illegal_argument_exception",
				"reason": "Connector URL is not matching the trusted connector endpoint regex, URL is: https://${parameters.endpoint}/v1/embeddings"
			}
		],
		"type": "illegal_argument_exception",
		"reason": "Connector URL is not matching the trusted connector endpoint regex, URL is: https://${parameters.endpoint}/v1/embeddings"
	},
	"status": 400
}
```
It worked well if I created the connector with url directly configured not through parameters `"url": "https://api.openai.com/v1/embeddings",` 

I think it fails [this validation](https://github.com/opensearch-project/ml-commons/blob/9aea9a08ef2d9eba082bc34382a8a04c1cc5bffa/plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java#L94) because the parameters payload in the update API doesn't include endpoint.

**How can one reproduce the bug?**
Check bug description

**What is the expected behavior?**
I'm wondering what the expected behavior is when the update API only includes one attribute in parameters, should it overwrite the whole `parameters` field or it should merge with existing `parameters` configuration. The current behavior is merge.

[...]

PR #2784, opened 2024-07-29
Merge the existing parameters when updating connectors

Zhangxunmt:
### Description
This PR fixes the issue reported in https://github.com/opensearch-project/ml-commons/issues/2502. This is not a bug since the issue was an expected behavior based on the initial design of the Update Connector API. However, over time we think that, in the Update APIs, it's a better CX experience to merge the updated parameters with the existing ones. 

Also, some details of the sagemaker model for batch inference is added in this PR.

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -287,7 +287,7 @@ public class HttpConnector extends AbstractConnector {
             this.protocol = updateContent.getProtocol();
         }
         if (updateContent.getParameters() != null && updateContent.getParameters().size() > 0) {
-            this.parameters = updateContent.getParameters();
+            getParameters().putAll(updateContent.getParameters());
         }
         if (updateContent.getCredential() != null && updateContent.getCredential().size() > 0) {
             this.credential = updateContent.getCredential();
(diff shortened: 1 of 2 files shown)


dhrubo-os (approved):

b4sjoo (approved):

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -287,7 +287,7 @@ public void update(MLCreateConnectorInput updateContent, Function<String, String
             this.protocol = updateContent.getProtocol();
         }
         if (updateContent.getParameters() != null && updateContent.getParameters().size() > 0) {
-            this.parameters = updateContent.getParameters();
+            getParameters().putAll(updateContent.getParameters());
```
chishui, line 290: Add UT and integ test? Seems this logic is not tested, otherwise the change of the behavior should break existing tests

chishui: BTW, thanks for the fix

Zhangxunmt: Will do in the next PR.

PR #2688, opened 2024-07-20
add initial MLInferenceSearchResponseProcessor

mingshl:
### Description
add initial MLInferenceSearchResponseProcessor

this PR only support collecting many document field in a list to add in one prediction call. 

The one document to one prediction call will add incrementally in a next PR
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2173
 
### Check List
- [x] New functionality includes testing.
  - [ ] All tests pass
- [x] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #2444, opened 2024-05-14
[META] Generic ML inference processor for search pipeline


ylwu-amzn:
Related issue https://github.com/opensearch-project/ml-commons/issues/2414

We have RAG and rerank search pipeline today. Some user want to leverage Gen AI in other use case, for example, they can leverage lang-ident model to check what language user is using, then translate the search response to that language.  We will build generic ML inference search processor for search request and response, so customer can use any model in their search pipeline.

Issue #2173, opened 2024-03-04
[RFC] ML Inference Processors 


mingshl:
### Problem statement

Currently, there are different implementations of search processors/ingest processors that use a machine learning model, e.g, [TextEmbeddingProcessor](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java) for text-embedding models ,  [GenerativeQAResponseProcessor](https://github.com/opensearch-project/ml-commons/blob/main/search-processors/src/main/java/org/opensearch/searchpipelines/questionanswering/generative/GenerativeQAResponseProcessor.java) for large language models, [PersonalizeRankingResponseProcessor](https://github.com/opensearch-project/search-processor/blob/main/amazon-personalize-ranking/src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java) for reranking models housing in AWS personalized Service.  Looking forward, when each type of machine learning models has a separate type of processor, the number of processors will grow to be enormous. It will be in-convenient for users when configuring different processors. However, [ML Commons plugin](https://github.com/opensearch-project/ml-commons/) supports connecting to a foundation model hosted on an external platform and uploading your own pre-trained model to the OpenSearch cluster, users can utilize the model_id from ml-commons plugins to apply in search/ingest process. We can simplified the multiple implementations  of search processors/ingest processors that use a machine learning model. 


### Motivation: 

To Improve the ease of using machine learning models to process ingest/search requests, we are introducing a Machine Learning Inference Processor to OpenSearch-ml-common to uses a machine learning model to read from the data and add the prediction outcome to the data that is being ingested through the ingest pipeline, or return the prediction outcomes along with the search response that is returned through the search pipeline. 


### Scope: 

* Developers can use models connected/uploaded in ml-commons to generate inference and add to ingest document/search response 
* Developers can use the ML Inference processors to handle input mappings from fields in ingest documents/search hits to model input, and handle output mapping from model inference output to new fields in ingest documents/search hits. 
* Developers can apply different Inference configurations(e.g, embedding_type, knn_size) to ML Inference processors to fit different use cases.
* Developers can use the ML Inference processors to call a model id to conduct multiple inferences processes for different set of input and output. e.g, creating embedding for two fields. 

### Out of Scope:

[...]

PR #2801, opened 2024-08-02
Support one_to_one in ML Inference Search Response Processor

mingshl:
### Description
Many document to one prediction is the default config in ML Inference Search Response Processor. This PR adds one document to one prediction support following up with https://github.com/opensearch-project/ml-commons/pull/2688

in 2.16, ml inference search response processor support collect documents into a list of model input and make one prediction call.

for example, 

the search response returns as following, and we would like to use the `text` field in the document to send to model input. 
```
{
  "took": 100,
  "timed_out": false,
  "_shards": {
    "total": 22,
    "successful": 22,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 2,
      "relation": "gte"
    },
    "max_score": 1,
    "hits": [
      {
        "_index": "test_index",
        "_id": "1",
        "_score": 1,
        "_source": {
          "text": " this is document 1"
      },
     {
            "_index": "test_index",
            "_id": "2",
            "_score": 1,
            "_source": {
              "text": " this is document 2"
          } 
      ] 
}
```

one_to_one inference is false setting will combine the two document containing the field `text` into `["this is document 1", "this is document 2"]` and this is one round of prediction.

in 2.17, we support one_to_one inference is true setting, that will make two rounds of prediction, the first prediction, we send `"this is document 1"` and the second prediction we send  `"this is document 2"`.

when users want to use a model that accept a list of model input, for example, using openai embedding model, we can set one_to_one is false, 

when users want to use a model that accept a single string model input, for example, using bedrock embedding model, we can set one_to_one is true. The most common use case for one-to-one inference is rerank use case using xgboost, which usually take a single document and compare with the search string and return a single score. 

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/2173
https://github.com/opensearch-project/ml-commons/issues/2444
https://github.com/opensearch-project/ml-commons/issues/2839
### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
+++ b/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
@@ -24,6 +24,7 @@ import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
@@ -33,6 +34,7 @@ import org.opensearch.common.collect.Tuple;
 import org.opensearch.common.xcontent.XContentHelper;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.MediaType;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentBuilder;
@@ -176,7 +178,8 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
             if (ignoreFailure) {
                 responseListener.onResponse(response);
             } else {
-                responseListener.onFailure(new RuntimeException(e.getMessage()));
+                responseListener
+                    .onFailure(new OpenSearchStatusException("Failed to process response: " + e.getMessage(), RestStatus.BAD_REQUEST));
             }
         }
     }
--- a/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
+++ b/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
@@ -126,12 +126,11 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
     /**
      * Processes the search response asynchronously by rewriting the documents with the inference results.
      *
-     * At default, process many document in one prediction through rewriteResponseDocuments method.
-     * but when process inference one document for one inference,
-     * separate one N-hits searchResponse into N one-hit search response,
-     * execute the same rewriteResponseDocument method,
-     * after N one-hit search response with inference result gets back,
-     * combined N one-hit search response back into one N-hits searchResponse.
+     * By default, it processes multiple documents in a single prediction through the rewriteResponseDocuments method.
+     * However, when processing one document per inference, it separates the N-hits search response into N one-hit search responses,
+     * executes the same rewriteResponseDocument method for each one-hit search response,
+     * and after receiving N one-hit search responses with inference results,
+     * it combines them back into a single N-hits search response.
      *
      * @param request          the search request
      * @param response         the search response
--- a/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
+++ b/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
@@ -21,6 +21,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -164,13 +165,18 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
                     responseListener,
                     hits
                 );
-
+                AtomicBoolean isOneHitListenerFailed = new AtomicBoolean(false);
+                ;
                 for (SearchHit hit : hits) {
                     SearchHit[] newHits = new SearchHit[1];
                     newHits[0] = hit;
                     SearchResponse oneHitResponse = SearchResponseUtil.replaceHits(newHits, response);
-                    ActionListener<SearchResponse> oneHitListener = getOneHitListener(combineResponseListener);
+                    ActionListener<SearchResponse> oneHitListener = getOneHitListener(combineResponseListener, isOneHitListenerFailed);
                     rewriteResponseDocuments(oneHitResponse, oneHitListener);
+                    // if any OneHitListener failure, try stop the rest of the predictions
+                    if (isOneHitListenerFailed.get()) {
+                        break;
+                    }
                 }
             }
 
@@ -189,12 +195,16 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
      * onResponse and onFailure callbacks to a GroupedActionListener.
      *
      * @param combineResponseListener The GroupedActionListener to which the
-     *                                 onResponse and onFailure callbacks will be
-     *                                 delegated.
+     *                                onResponse and onFailure callbacks will be
+     *                                delegated.
+     * @param isOneHitListenerFailed
      * @return An ActionListener that delegates its callbacks to the provided
-     *         GroupedActionListener.
+     * GroupedActionListener.
      */
-    private static ActionListener<SearchResponse> getOneHitListener(GroupedActionListener<SearchResponse> combineResponseListener) {
+    private static ActionListener<SearchResponse> getOneHitListener(
+        GroupedActionListener<SearchResponse> combineResponseListener,
+        AtomicBoolean isOneHitListenerFailed
+    ) {
         ActionListener<SearchResponse> oneHitListener = new ActionListener<>() {
             @Override
             public void onResponse(SearchResponse response) {
@@ -203,6 +213,8 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
 
             @Override
             public void onFailure(Exception e) {
+                // if any OneHitListener failure, try stop the rest of the predictions and return
+                isOneHitListenerFailed.compareAndSet(false, true);
                 combineResponseListener.onFailure(e);
             }
         };
@@ -239,11 +251,11 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
 
             @Override
             public void onFailure(Exception e) {
-                // when one hit failed and ignoreFailure return original response
                 if (ignoreFailure) {
                     responseListener.onResponse(response);
                 } else {
-                    responseListener.onFailure(e);
+                    responseListener
+                        .onFailure(new OpenSearchStatusException("Failed to process response: " + e.getMessage(), RestStatus.BAD_REQUEST));
                 }
             }
         }, hits.length);
@@ -265,7 +277,7 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
         // hitCountInPredictions keeps track of the count of hit that have the required input fields for each round of prediction
         Map<Integer, Integer> hitCountInPredictions = new HashMap<>();
 
-        ActionListener<Map<Integer, MLOutput>> rewriteResponseListener = createRewriteResponseListenerManyToOne(
+        ActionListener<Map<Integer, MLOutput>> rewriteResponseListener = createRewriteResponseListener(
             response,
             responseListener,
             processInputMap,
@@ -273,13 +285,13 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
             hitCountInPredictions
         );
 
-        GroupedActionListener<Map<Integer, MLOutput>> batchPredictionListener = createBatchPredictionListenerManyToOne(
+        GroupedActionListener<Map<Integer, MLOutput>> batchPredictionListener = createBatchPredictionListener(
             rewriteResponseListener,
             inputMapSize
         );
         SearchHit[] hits = response.getHits().getHits();
         for (int inputMapIndex = 0; inputMapIndex < max(inputMapSize, 1); inputMapIndex++) {
-            processPredictionsManyToOne(hits, processInputMap, inputMapIndex, batchPredictionListener, hitCountInPredictions);
+            processPredictions(hits, processInputMap, inputMapIndex, batchPredictionListener, hitCountInPredictions);
         }
     }
 
@@ -293,7 +305,7 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
      * @param hitCountInPredictions   a map to keep track of the count of hits that have the required input fields for each round of prediction
      * @throws IOException if an I/O error occurs during the prediction process
      */
-    private void processPredictionsManyToOne(
+    private void processPredictions(
         SearchHit[] hits,
         List<Map<String, String>> processInputMap,
         int inputMapIndex,
@@ -356,7 +368,6 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
 
                     // when not existed in the map, add into the modelInputParameters map
                     updateModelInputParameters(modelInputParameters, modelInputFieldName, documentValue);
-
                 }
             }
         }
@@ -430,7 +441,7 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
      * @param inputMapSize            the size of the input map
      * @return a grouped action listener for batch predictions
      */
-    private GroupedActionListener<Map<Integer, MLOutput>> createBatchPredictionListenerManyToOne(
+    private GroupedActionListener<Map<Integer, MLOutput>> createBatchPredictionListener(
         ActionListener<Map<Integer, MLOutput>> rewriteResponseListener,
         int inputMapSize
     ) {
@@ -462,7 +473,7 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
      * @param hitCountInPredictions a map to keep track of the count of hits that have the required input fields for each round of prediction
      * @return an action listener for rewriting the response with the inference results
      */
-    private ActionListener<Map<Integer, MLOutput>> createRewriteResponseListenerManyToOne(
+    private ActionListener<Map<Integer, MLOutput>> createRewriteResponseListener(
         SearchResponse response,
         ActionListener<SearchResponse> responseListener,
         List<Map<String, String>> processInputMap,
@@ -494,7 +505,7 @@ public class MLInferenceSearchResponseProcessor extends AbstractProcessor implem
                                 Map<String, String> outputMapping = getDefaultOutputMapping(mappingIndex, processOutputMap);
 
                                 boolean isModelInputMissing = false;
-                                if (processInputMap != null) {
+                                if (processInputMap != null && !processInputMap.isEmpty()) {
                                     isModelInputMissing = checkIsModelInputMissing(document, inputMapping);
                                 }
                                 if (!isModelInputMissing) {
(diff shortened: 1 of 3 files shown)


mingshl:
We need to bump the version in the build.gradle file to let the CI runs. 
```
> Could not resolve all dependencies for configuration ':opensearch-ml-plugin:runtimeClasspath'.
   > Conflict found for the following module:
       - org.apache.httpcomponents.core5:httpcore5 between versions 5.2.5 and 5.2.2
       - 
```

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +152,106 @@ public void processResponseAsync(
...
+                    rewriteResponseDocuments(oneHitResponse, oneHitListener);
+                }
+            }
+
         } catch (Exception e) {
             if (ignoreFailure) {
                 responseListener.onResponse(response);
             } else {
-                responseListener.onFailure(e);
+                responseListener.onFailure(new RuntimeException(e.getMessage()));
```
zane-neo, line 180: Why encapsulate the original exception to RuntimeException?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -306,18 +396,28 @@ public void onFailure(Exception e) {
...
+     * create a new list and add the document value
+     * @param modelInputParameters The map containing the model input parameters.
+     * @param modelInputFieldName The name of the model input field.
+     * @param documentValue The value from the document that needs to be added to the model input parameters.
+     */
+    private void updateModelInputParameters(Map<String, Object> modelInputParameters, String modelInputFieldName, Object documentValue) {
+        if (!this.oneToOne) {
+            if (!modelInputParameters.containsKey(modelInputFieldName)) {
+                List<Object> documentValueList = new ArrayList<>();
+                documentValueList.add(documentValue);
```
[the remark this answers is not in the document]
mingshl: the comment is already added here 

https://github.com/opensearch-project/ml-commons/pull/2801/files#diff-f2182d1cfc06d79c1ef1b25c20174c3ada744a16784d80cfb32d697128526ceeR404

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
mingshl: to avoid 500x error

dhrubo-os:
Thanks for the explanation. IMO, `one_to_one` doesn't not seem intuitive. May be `single_document_inference`? 

>in 2.16, we support one_to_one inference is true setting,

So we already released this one_to_one inference in 2.16?

Where do we save this setting?

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessorTests.java
```diff
@@ -96,6 +98,17 @@ public void testProcessResponseException() throws Exception {
         }
     }
 
+    // public void testPrint(){
```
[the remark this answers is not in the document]
mingshl: right!! I need to remove this test

mingshl:
> Thanks for the explanation. IMO, `one_to_one` doesn't not seem intuitive. May be `single_document_inference`? Or `document_wise_inference`?
> 
> > in 2.16, we support one_to_one inference is true setting,
> 
> So we already released this one_to_one inference in 2.16?
> 
> Where do we save this setting?

aw ha, it should be 2.17, yeah this change is going to add to 2.17. just modified the descroiption.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -125,9 +126,16 @@ public SearchResponse processResponse(SearchRequest request, SearchResponse resp
     /**
      * Processes the search response asynchronously by rewriting the documents with the inference results.
      *
-     * @param request         the search request
-     * @param response        the search response
-     * @param responseContext the pipeline processing context
+     * At default, process many document in one prediction through rewriteResponseDocuments method.
```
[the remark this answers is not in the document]
mingshl: at default, when one_to_one is false,

we run rewriteResponseDocuments by send a response that has N hits, meaning N document, and it will compile a list of model input with N documents, for example, when N=2, the model input is `"input":["document1", "document2"]`

 when one_to_one is true, 
Step 1:  in processResponseAsync method, it decompose the ONE response with N hits into  N responses with one hit, https://github.com/opensearch-project/ml-commons/blob/b52053980615afe1167ac7c458978d26a2e70215/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java#L170 and

Step 2:  run the same rewriteResponseDocuments method, in this way, it sends N model input with one document/hit. For example, when N=2, the first model input `"input":"document1"`, the second model input `"input": "document2"`, after every response aded the model predictions into the hits. 

Step 3: The combineResponseListener will combine the N responses with one hit back into ONE response with N hits, and the return the ONE response.

mingshl:
> Thanks for the explanation. IMO, `one_to_one` doesn't not seem intuitive. May be `single_document_inference`? Or `document_wise_inference`?
> 
> > in 2.16, we support one_to_one inference is true setting,
> 
> So we already released this one_to_one inference in 2.16?
> 
> Where do we save this setting?

in the ml search response processor, we already had it in 2.16 but it's not supported to turn on to true yet.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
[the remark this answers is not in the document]
mingshl: summarized the java doc, is it better? @dhrubo-os 

```
    /**
     * Processes the search response asynchronously by rewriting the documents with the inference results.
     *
     * By default, it processes multiple documents in a single prediction through the rewriteResponseDocuments method.
     * However, when processing one document per inference, it separates the N-hits search response into N one-hit search responses,
     * executes the same rewriteResponseDocument method for each one-hit search response,
     * and after receiving N one-hit search responses with inference results,
     * it combines them back into a single N-hits search response.
     *
     * @param request          the search request
     * @param response         the search response
     * @param responseContext  the pipeline processing context
     * @param responseListener the listener to be notified when the response is processed
     */
```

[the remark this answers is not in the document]
dhrubo-os: That's better. Thanks.

dhrubo-os:
> in the ml search response processor, we already had it in 2.16 but it's not supported to turn on to true yet.

can we change the settings name? or we need to stick with `one_to_one`?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
Zhangxunmt: Suggest use OpenSearchStatusException to replace RuntimeException.

Zhangxunmt (dismissed):

mingshl:
> > in the ml search response processor, we already had it in 2.16 but it's not supported to turn on to true yet.
> 
> can we change the settings name? or we need to stick with `one_to_one`?

the name is added since 2.16, if we change the name in 2.17, it might case confusion.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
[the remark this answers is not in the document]
mingshl: changed in the new commit https://github.com/opensearch-project/ml-commons/pull/2801/commits/eca844b5850c502a8acd207c2f7a0d442839bd8f

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
mingshl: I don't see 5xx exception, it's for precautions. changed to OpenSearchStatusException similar to https://github.com/opensearch-project/ml-commons/blob/9663053d544b59ef238ae36bf49b774543552df5/plugin/src/main/java/org/opensearch/ml/utils/IndexUtils.java#L134


added in commit [use OpenSearchStatusException in error handling](https://github.com/opensearch-project/ml-commons/pull/2801/commits/0d5c6bf66f87f1819ae4554b55d278ea0fa7c5ec)

dhrubo-os:
> the name is added since 2.16, if we change the name in 2.17, it might case confusion.

Confusion to whom? Did we make any public documentation about this?

mingshl:
> > the name is added since 2.16, if we change the name in 2.17, it might case confusion.
> 
> Confusion to whom? Did we make any public documentation about this?



> > the name is added since 2.16, if we change the name in 2.17, it might case confusion.
> 
> Confusion to whom? Did we make any public documentation about this?

we did add it to the documentation already https://github.com/opensearch-project/documentation-website/blob/main/_search-plugins/search-pipelines/ml-inference-search-response.md

mingshl:
flaky CI test failed at ConversationalMemory tests, which is not related to this change. 

```
ConversationalMemoryHandlerITTests > testDifferentUsers_DifferentConversations FAILED
    java.lang.Exception at [11AB9FC15F2A1BE]:0

ConversationalMemoryHandlerITTests > classMethod FAILED
    java.lang.Exception at [11AB9FC15F2A1BE]:0

ConversationalMemoryHandlerITTests > classMethod FAILED
    java.lang.AssertionError at ConversationalMemoryHandlerITTests.java:537

ConversationalTests > classMethod FAILED
    java.lang.AssertionError at TestRuleTemporaryFilesCleanup.java:100

ConversationalTests > classMethod FAILED
    java.io.IOException at IOUtils.java:341

ConversationActionTests > classMethod FAILED
    java.lang.AssertionError at [515BEDD3C5B2DA24]:0

ConversationMetaIndexITTests > testCanQueryOverConversations FAILED
    java.lang.IllegalStateException at [8D3DC5953A847910:424C17DF546163FA]:0
        Caused by: java.nio.file.NoSuchFileException at WindowsException.java:85

258 tests completed, 7 failed, 1 skipped

```

Zhangxunmt (approved):

dhrubo-os:
> flaky CI test failed at ConversationalMemory tests, which is not related to this change.
> 
> ```
> ConversationalMemoryHandlerITTests > testDifferentUsers_DifferentConversations FAILED
>     java.lang.Exception at [11AB9FC15F2A1BE]:0
> 
> ConversationalMemoryHandlerITTests > classMethod FAILED
>     java.lang.Exception at [11AB9FC15F2A1BE]:0
> 
> ConversationalMemoryHandlerITTests > classMethod FAILED
>     java.lang.AssertionError at ConversationalMemoryHandlerITTests.java:537
> 
> ConversationalTests > classMethod FAILED
>     java.lang.AssertionError at TestRuleTemporaryFilesCleanup.java:100
> 
> ConversationalTests > classMethod FAILED
>     java.io.IOException at IOUtils.java:341
> 
> ConversationActionTests > classMethod FAILED
>     java.lang.AssertionError at [515BEDD3C5B2DA24]:0
> 
> ConversationMetaIndexITTests > testCanQueryOverConversations FAILED
>     java.lang.IllegalStateException at [8D3DC5953A847910:424C17DF546163FA]:0
>         Caused by: java.nio.file.NoSuchFileException at WindowsException.java:85
> 
> 258 tests completed, 7 failed, 1 skipped
> ```

https://github.com/opensearch-project/ml-commons/actions/runs/10377677651/job/28732383708?pr=2801

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +153,107 @@ public void processResponseAsync(
...
+        ActionListener<SearchResponse> responseListener,
+        SearchHit[] hits
+    ) {
+        GroupedActionListener<SearchResponse> combineResponseListener = new GroupedActionListener<>(new ActionListener<>() {
+            @Override
+            public void onResponse(Collection<SearchResponse> responseMapCollection) {
+                SearchHit[] combinedHits = new SearchHit[hits.length];
+                int i = 0;
+                for (SearchResponse OneHitResponseAfterInference : responseMapCollection) {
+                    SearchHit[] hitsAfterInference = OneHitResponseAfterInference.getHits().getHits();
```
dhrubo-os, line 232: This code assumes that `hitsAfterInference` always has at least one element (`hitsAfterInference[0]`). If any `SearchResponse` in `responseMapCollection` has no hits (i.e., `hitsAfterInference` is empty), this would throw an `ArrayIndexOutOfBoundsException`.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +153,107 @@ public void processResponseAsync(
...
+     * @param responseListener The ActionListener to be notified with the combined SearchResponse.
+     * @param hits             The array of SearchHits to be processed.
+     * @return A GroupedActionListener that combines the SearchResponses and constructs a new SearchResponse.
+     */
+    private GroupedActionListener<SearchResponse> getCombineResponseGroupedActionListener(
+        SearchResponse response,
+        ActionListener<SearchResponse> responseListener,
+        SearchHit[] hits
+    ) {
+        GroupedActionListener<SearchResponse> combineResponseListener = new GroupedActionListener<>(new ActionListener<>() {
```
dhrubo-os, line 226: Ensure that `GroupedActionListener` is thread-safe, as it appears to be aggregating responses from potentially asynchronous operations. If multiple threads call onResponse or onFailure concurrently, this could lead to race conditions.

mingshl: right. this groupedActionListener collects response from multiple oneHitListener but it would onResponse after all  multiple oneHitListener returns.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
mingshl: when we made the oneHitResponse, at least it will contain one hit.

dhrubo-os:
```
[ant:jacocoReport] Rule violated for class org.opensearch.ml.processor.MLInferenceSearchResponseProcessor.2: branches covered ratio is 0.5, but expected minimum is 0.7
[ant:jacocoReport] Rule violated for class org.opensearch.ml.processor.MLInferenceSearchResponseProcessor.2: lines covered ratio is 0.7, but expected minimum is 0.8
[ant:jacocoReport] Rule violated for class org.opensearch.ml.processor.MLInferenceSearchResponseProcessor.1: lines covered ratio is 0.6, but expected minimum is 0.8
```

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +154,118 @@ public void processResponseAsync(
...
+                    hits
+                );
+                AtomicBoolean isOneHitListenerFailed = new AtomicBoolean(false);
+                ;
+                for (SearchHit hit : hits) {
+                    SearchHit[] newHits = new SearchHit[1];
+                    newHits[0] = hit;
+                    SearchResponse oneHitResponse = SearchResponseUtil.replaceHits(newHits, response);
+                    ActionListener<SearchResponse> oneHitListener = getOneHitListener(combineResponseListener, isOneHitListenerFailed);
+                    rewriteResponseDocuments(oneHitResponse, oneHitListener);
```
zane-neo, line 175: This for loop is a performance bottleneck, let's say each prediction takes 300ms, and there're 10 documents needs prediction, that's 3s, which is too long for user client. Do you have plan to optimize this?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +154,118 @@ public void processResponseAsync(
...
+                    // if any OneHitListener failure, try stop the rest of the predictions
+                    if (isOneHitListenerFailed.get()) {
+                        break;
```
zane-neo, line 178: If breaks here, is the GroupedActionListener can still get enough calls(onResponse/onFailure) to complete?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
mingshl: I am using a group action listener, so all 10 prediction tasks are running asynchronous in your example. After 10 tasks all succeed, it will combine the prediction result. It's not going to take 3s.

And yes if we have batch remote inference tasks, I plan to adapt and enhance this.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
mingshl: If it breaks, it must go pass line 218, and the combine response listener will onFailure, and that also lead to the response listener onFailure. I also have one to one prediction failed and verify it will only run one prediction only and call onFailure.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
```diff
@@ -144,20 +154,118 @@ public void processResponseAsync(
...
+                }
+            }
+
         } catch (Exception e) {
             if (ignoreFailure) {
                 responseListener.onResponse(response);
             } else {
-                responseListener.onFailure(e);
+                responseListener
+                    .onFailure(new OpenSearchStatusException("Failed to process response: " + e.getMessage(), RestStatus.BAD_REQUEST));
```
zane-neo, line 188: BAD_REQUEST is a client error: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400 and not the case here. Here we should throw the original exception got from remote server, or at least a 500 error identifying this is a server side error.

mingshl: The most common failure here is client.execute prediction task failed, so it's client errors. And also referencing her is the reason why I am adapting OpenSearchStatusExeception.  

https://github.com/opensearch-project/ml-commons/pull/2801#discussion_r1715957789

zane-neo: Http 400 error stands for user side error, meaning that when user get this error, they'll try to figure out the query DSL from their side, the `client` in opensearch is still a server side client, isn't the same one in user side.

Check the exception handling [here](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java#L129) and [here](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java#L186), at least for remote model, you can get the actual exception instead of throwing 400 error.

mingshl:
All tests passed

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchResponseProcessor.java
dhrubo-os: I agree with @zane-neo. We should forward the actual error. Wrapping the client-side error as a server-side error, or vice versa, would be confusing for both the customer and the service team when trying to identify the root cause of any issues.

mingshl: Thanks @dhrubo-os , @zane-neo  and @Zhangxunmt, agree that we should handle different exception properly, and set the default exception is the actual error. 

Added specific handling for OpenSearchStatusException and MLResourceNotFoundException, and the default exception is the actual error in the new commit [here](https://github.com/opensearch-project/ml-commons/pull/2801/commits/6e16254b611f0508c6e64682fbb9f50bdce0a4b8)

Please help review again and approve.

Zhangxunmt (approved):

b4sjoo (approved):

PR #2792, opened 2024-07-31
fixing agent execution for multi-tenancy

dhrubo-os:
### Description
[ `MLFeatureEnabledSetting` class is in `plugin` module and `algorithm` module doesn't have dependency on `plugin` module. So when we do `mlFeatureEnabledSetting.isMultiTenancyEnabled()` to the `MLAgentExecutor` class, it doesn't update the dynamic value of `ML_COMMONS_MULTI_TENANCY_ENABLED` settings. So we need a listener in algorithm module to get the update status of this settings.]

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2803, opened 2024-08-03
[Feature/multi_tenancy] Implement tenant id in Local and Remote Cluster Search Clients

dbwiddis:
### Description

Implements filtering by tenant ID in `LocalClusterIndicesClient` and `RemoteClusterIndicesClient`.

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/CommonValue.java
+++ b/common/src/main/java/org/opensearch/ml/common/CommonValue.java
@@ -36,560 +36,515 @@ import static org.opensearch.ml.common.model.TextEmbeddingModelConfig.POOLING_MO
 
 public class CommonValue {
 
-    public static Integer NO_SCHEMA_VERSION = 0;
-    public static final String REMOTE_SERVICE_ERROR = "Error from remote service: ";
-    public static final String USER = "user";
-    public static final String META = "_meta";
-    public static final String SCHEMA_VERSION_FIELD = "schema_version";
-    public static final String UNDEPLOYED = "undeployed";
-    public static final String NOT_FOUND = "not_found";
+        public static Integer NO_SCHEMA_VERSION = 0;
+        public static final String REMOTE_SERVICE_ERROR = "Error from remote service: ";
+        public static final String USER = "user";
+        public static final String META = "_meta";
+        public static final String SCHEMA_VERSION_FIELD = "schema_version";
+        public static final String UNDEPLOYED = "undeployed";
+        public static final String NOT_FOUND = "not_found";
 
-    /**
-     * The field name to be inserted into a document when operating tenant aware.
-     * <p>
-     * The value for this field will be supplied by REST callers using {@code x-tenant-id} header. If no such header is provided, {@link #DEFAULT_TENANT} is used.
-     * <p>
-     * This value should be chosen to avoid conflicts with any existing or future fields.  In the future this may be replaced by a plugin setting.  
-     */
-    public static final String TENANT_ID = "_tenant_id";
-    /**
-     * The default tenant id to be used if the REST header {@code x-tenant-id} is not present.
-     */
-    public static final String DEFAULT_TENANT = "DEFAULT_TENANT";
+        public static final String TENANT_ID = "tenant_id";
 
-    public static final String MASTER_KEY = "master_key";
-    public static final String CREATE_TIME_FIELD = "create_time";
-    public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
+        public static final String MASTER_KEY = "master_key";
+        public static final String CREATE_TIME_FIELD = "create_time";
+        public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
 
-    public static final String BOX_TYPE_KEY = "box_type";
-    // hot node
-    public static String HOT_BOX_TYPE = "hot";
-    // warm node
-    public static String WARM_BOX_TYPE = "warm";
-    public static final String ML_MODEL_GROUP_INDEX = ".plugins-ml-model-group";
-    public static final String ML_MODEL_INDEX = ".plugins-ml-model";
-    public static final String ML_TASK_INDEX = ".plugins-ml-task";
-    public static final Integer ML_MODEL_GROUP_INDEX_SCHEMA_VERSION = 2;
-    public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 11;
-    public static final String ML_CONNECTOR_INDEX = ".plugins-ml-connector";
-    public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 2;
-    public static final Integer ML_CONNECTOR_SCHEMA_VERSION = 3;
-    public static final String ML_CONFIG_INDEX = ".plugins-ml-config";
-    public static final Integer ML_CONFIG_INDEX_SCHEMA_VERSION = 2;
-    public static final String ML_CONTROLLER_INDEX = ".plugins-ml-controller";
-    public static final Integer ML_CONTROLLER_INDEX_SCHEMA_VERSION = 1;
-    public static final String ML_MAP_RESPONSE_KEY = "response";
-    public static final String ML_AGENT_INDEX = ".plugins-ml-agent";
-    public static final Integer ML_AGENT_INDEX_SCHEMA_VERSION = 2;
-    public static final String ML_MEMORY_META_INDEX = ".plugins-ml-memory-meta";
-    public static final Integer ML_MEMORY_META_INDEX_SCHEMA_VERSION = 1;
-    public static final String ML_MEMORY_MESSAGE_INDEX = ".plugins-ml-memory-message";
-    public static final Integer ML_MEMORY_MESSAGE_INDEX_SCHEMA_VERSION = 1;
-    public static final String USER_FIELD_MAPPING = "      \""
-        + CommonValue.USER
-        + "\": {\n"
-        + "        \"type\": \"nested\",\n"
-        + "        \"properties\": {\n"
-        + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
-        + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-        + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-        + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
-        + "        }\n"
-        + "      }\n";
-    public static final String ML_MODEL_GROUP_INDEX_MAPPING = "{\n"
-        + "  \"_meta\": {\n"
-        + "    \"schema_version\": "
-        + ML_MODEL_GROUP_INDEX_SCHEMA_VERSION
-        + "\n"
-        + "  },\n"
-        + "  \"properties\": {\n"
-        + "    \""
-        + MLModelGroup.MODEL_GROUP_NAME_FIELD
-        + "\": {\n"
-        + "      \"type\": \"text\",\n"
-        + "      \"fields\": {\n"
-        + "        \"keyword\": {\n"
-        + "          \"type\": \"keyword\",\n"
-        + "          \"ignore_above\": 256\n"
-        + "        }\n"
-        + "      }\n"
-        + "    },\n"
-        + "    \""
-        + MLModelGroup.DESCRIPTION_FIELD
-        + "\": {\n"
-        + "      \"type\": \"text\"\n"
-        + "    },\n"
-        + "    \""
-        + MLModelGroup.LATEST_VERSION_FIELD
-        + "\": {\n"
-        + "      \"type\": \"integer\"\n"
-        + "    },\n"
-        + "   \""
-        + MLModelGroup.MODEL_GROUP_ID_FIELD
-        + "\": {\n"
-        + "      \"type\": \"keyword\"\n"
-        + "    },\n"
-        + "    \""
-        + MLModelGroup.BACKEND_ROLES_FIELD
-        + "\": {\n"
-        + "      \"type\": \"text\",\n"
-        + "      \"fields\": {\n"
-        + "        \"keyword\": {\n"
-        + "          \"type\": \"keyword\",\n"
-        + "          \"ignore_above\": 256\n"
-        + "        }\n"
-        + "      }\n"
-        + "    },\n"
-        + "   \""
-        + MLModelGroup.ACCESS
-        + "\": {\n"
-        + "      \"type\": \"keyword\"\n"
-        + "    },\n"
-        + "    \""
-        + MLModelGroup.OWNER
-        + "\": {\n"
-        + "      \"type\": \"nested\",\n"
-        + "        \"properties\": {\n"
-        + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
-        + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-        + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
-        + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
-        + "        }\n"
-        + "    },\n"
-        + "     \""
-        + MLModelGroup.CREATED_TIME_FIELD
-        + "\": {\n"
-        + "      \"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
-        + "    \""
-        + MLModelGroup.LAST_UPDATED_TIME_FIELD
-        + "\": {\n"
-        + "      \"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"}\n"
-        + "  }\n"
-        + "}";
+        public static final String BOX_TYPE_KEY = "box_type";
+        // hot node
+        public static String HOT_BOX_TYPE = "hot";
+        // warm node
+        public static String WARM_BOX_TYPE = "warm";
+        public static final String ML_MODEL_GROUP_INDEX = ".plugins-ml-model-group";
+        public static final String ML_MODEL_INDEX = ".plugins-ml-model";
+        public static final String ML_TASK_INDEX = ".plugins-ml-task";
+        public static final Integer ML_MODEL_GROUP_INDEX_SCHEMA_VERSION = 2;
+        public static final Integer ML_MODEL_INDEX_SCHEMA_VERSION = 11;
+        public static final String ML_CONNECTOR_INDEX = ".plugins-ml-connector";
+        public static final Integer ML_TASK_INDEX_SCHEMA_VERSION = 2;
+        public static final Integer ML_CONNECTOR_SCHEMA_VERSION = 3;
+        public static final String ML_CONFIG_INDEX = ".plugins-ml-config";
+        public static final Integer ML_CONFIG_INDEX_SCHEMA_VERSION = 2;
+        public static final String ML_CONTROLLER_INDEX = ".plugins-ml-controller";
+        public static final Integer ML_CONTROLLER_INDEX_SCHEMA_VERSION = 1;
+        public static final String ML_MAP_RESPONSE_KEY = "response";
+        public static final String ML_AGENT_INDEX = ".plugins-ml-agent";
+        public static final Integer ML_AGENT_INDEX_SCHEMA_VERSION = 2;
+        public static final String ML_MEMORY_META_INDEX = ".plugins-ml-memory-meta";
+        public static final Integer ML_MEMORY_META_INDEX_SCHEMA_VERSION = 1;
+        public static final String ML_MEMORY_MESSAGE_INDEX = ".plugins-ml-memory-message";
+        public static final Integer ML_MEMORY_MESSAGE_INDEX_SCHEMA_VERSION = 1;
+        public static final String USER_FIELD_MAPPING = "      \""
+                        + CommonValue.USER
+                        + "\": {\n"
+                        + "        \"type\": \"nested\",\n"
+                        + "        \"properties\": {\n"
+                        + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
+                        + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
+                        + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
+                        + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
+                        + "        }\n"
+                        + "      }\n";
+        public static final String ML_MODEL_GROUP_INDEX_MAPPING = "{\n"
+                        + "  \"_meta\": {\n"
+                        + "    \"schema_version\": " + ML_MODEL_GROUP_INDEX_SCHEMA_VERSION + "\n"
+                        + "  },\n"
+                        + "  \"properties\": {\n"
+                        + "    \"" + MLModelGroup.MODEL_GROUP_NAME_FIELD + "\": {\n"
+                        + "      \"type\": \"text\",\n"
+                        + "      \"fields\": {\n"
+                        + "        \"keyword\": {\n"
+                        + "          \"type\": \"keyword\",\n"
+                        + "          \"ignore_above\": 256\n"
+                        + "        }\n"
+                        + "      }\n"
+                        + "    },\n"
+                        + "    \"" + MLModelGroup.DESCRIPTION_FIELD + "\": {\n"
+                        + "      \"type\": \"text\"\n"
+                        + "    },\n"
+                        + "    \"" + MLModelGroup.LATEST_VERSION_FIELD + "\": {\n"
+                        + "      \"type\": \"integer\"\n"
+                        + "    },\n"
+                        + "   \"" + MLModelGroup.MODEL_GROUP_ID_FIELD + "\": {\n"
+                        + "      \"type\": \"keyword\"\n"
+                        + "    },\n"
+                        + "    \"" + MLModelGroup.BACKEND_ROLES_FIELD + "\": {\n"
+                        + "      \"type\": \"text\",\n"
+                        + "      \"fields\": {\n"
+                        + "        \"keyword\": {\n"
+                        + "          \"type\": \"keyword\",\n"
+                        + "          \"ignore_above\": 256\n"
+                        + "        }\n"
+                        + "      }\n"
+                        + "    },\n"
+                        + "   \"" + MLModelGroup.ACCESS + "\": {\n"
+                        + "      \"type\": \"keyword\"\n"
+                        + "    },\n"
+                        + "    \"" + MLModelGroup.OWNER + "\": {\n"
+                        + "      \"type\": \"nested\",\n"
+                        + "        \"properties\": {\n"
+                        + "          \"name\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\", \"ignore_above\":256}}},\n"
+                        + "          \"backend_roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
+                        + "          \"roles\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}},\n"
+                        + "          \"custom_attribute_names\": {\"type\":\"text\", \"fields\":{\"keyword\":{\"type\":\"keyword\"}}}\n"
+                        + "        }\n"
+                        + "    },\n"
+                        + "     \"" + MLModelGroup.CREATED_TIME_FIELD + "\": {\n"
+                        + "      \"type\": \"date\", \"format\": \"strict_date_time||epoch_millis\"},\n"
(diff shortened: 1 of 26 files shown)


common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -189,7 +197,13 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
     public CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(SearchDataObjectRequest request, Executor executor) {
         return CompletableFuture.supplyAsync(() -> AccessController.doPrivileged((PrivilegedAction<SearchDataObjectResponse>) () -> {
             log.info("Searching {}", Arrays.toString(request.indices()));
-            SearchResponse searchResponse = client.search(new SearchRequest(request.indices(), request.searchSourceBuilder())).actionGet();
+            SearchSourceBuilder searchSource = request.searchSourceBuilder();
+            QueryBuilder existingQuery = searchSource.query();
+            String tenantId = request.tenantId() != null ? request.tenantId() : DEFAULT_TENANT;
```
arjunkumargiri, line 202: In case multi tenancy is disabled tenant ID field will be empty. Searching for documents with `DEFAULT_TENANT` will return incorrect result.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -65,6 +70,9 @@
 @Log4j2
 public class LocalClusterIndicesClient implements SdkClientDelegate {
 
+    private static final String HASH_KEY = "_tenant_id";
```
arjunkumargiri, line 73: Rename to `tenant_id`? By default there would be no _tenant_id field

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
```diff
@@ -71,6 +76,9 @@
 @Log4j2
 public class RemoteClusterIndicesClient implements SdkClientDelegate {
 
+    private static final String HASH_KEY = "_tenant_id";
```
arjunkumargiri, line 79: Query field should be `tenant_id`

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: I literally copied and pasted this line of code from the DDBClient.  Granted, that was changed in a recent PR because I didn't realize it was defined elsewhere: we should not be duplicating important strings like this in code, this is a classic example of why; I missed that we have a `CommonValue` for it and would have changed it instead.

I still maintain `_tenant_id` is better and plan to change the `CommonValue` string to align.  Why do prefer a plain field that any regular user can accidentally overwrite with a sensible sounding field?  Are we taking any action to prevent users from not using these special field names?

We use leading underscores for other important metadata like document Id (`_id`), as well as `_seq_no` and `_primary_term` and `_source`.   It's a well known pattern.  What if the user wants to actually use their own "tenant ID" in a document? How will they know that using `tenant_id` will end up denying them access to their documents?  Are there any other secret keys that will totally break their documents if they use them?

Let's align on this and

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
dbwiddis: Query field should reference the CommonValue constant instead of copy/pasting the same String everywhere

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: Sorry, this is extremely poorly documented.  As you can probably see, Integ tests are failing precisely because of this assumption.

But how, in the `common` module, is it even possible to know if multi-tenancy is disabled?  I moved this client here by request, but that makes it impossible to know whether multi-tenancy is enabled.

Tell me what this line of code should read.   I know it's wrong.  I don't know how to make it right.

dbwiddis: To be more clear, please answer:
1. When is tenant Id `DEFAULT_TENANT_ID` and when is it null?  Nobody has bothered to document the conventions here.   Why do we need a specific non-null string?
2. Why are we choosing the string "DEFAULT_TENANT_ID" for this value?   Similar to my objection to using the field name/hash key "tenant_id" we should not use any common string that anyone might accidentally use.  Can a user overwrite the `tenant_id` field with a value of `"DEFAULT_TENANT_ID"` and gain access to all data in the cluster?  If not, where is that documented (see #1).  Let's align on these constants, I feel that I have made multiple comments on other PRs that have not led to any discussion or final conclusion.  
3. How is it possible in the `common` module for this code to know the value in the `plugin` module is; specifically whether multi-tenancy is enabled?  What happens when we move this client code to another repository?  How is this external dependency going to know whether multi-tenancy is enabled?    Should it be assumed that if you're using any of the SDK client implementations you have to use multi-tenancy?

I consider this PR and honestly a lot of work on this dev branch blocked until we align on question 3 at least.

Flipping this to draft mode.  @dhrubo-os @arjunkumargiri please comment.

dhrubo-os: @dbwiddis please let me know how can I improve documentation [here](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java#L61-L63)

During index searching, can't we search tenant_id =  null (if tenant id isn't provided)? Why do we need any default value here?


>How is it possible in the common module for this code to know the value in the plugin module is; specifically whether multi-tenancy is enabled? What happens when we move this client code to another repository? How is this external dependency going to know whether multi-tenancy is enabled? Should it be assumed that if you're using any of the SDK client implementations you have to use multi-tenancy?


Yeah, I was having the same issue, then I ended up doing this: https://github.com/opensearch-project/ml-commons/pull/2792/files

In `MLAgentExecutor.java`, which stays in the Algorithm module I needed to have the dynamic multi-tenancy value.

dbwiddis: > During index searching, can't we search tenant_id = null (if tenant id isn't provided)? 

The filter by term query doesn’t accept null.  We can easily skip adding the filter if we know we’re not multi tenant.  But how do we know that?

and how do we protect against accidentally leaving it null and exposing other tenant information?

> please let me know how can I improve documentation [here](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java#L61-L63)

that just explains why we substitute a string but not how we treat null throughout the end to end. 

Does null always mean multi tenancy is not enabled?  It should but that needs validation somewhere.

dhrubo-os: Yeah for multi-tenancy we are expecting `tenant_id` not to be null. 
For rest layer, check is [here](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/rest/AbstractMLSearchAction.java#L60)

For transport layers checks are:

1. [agent search](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/agents/TransportSearchAgentAction.java#L61-L63)
2. [connector search](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/connector/SearchConnectorTransportAction.java#L82-L84)
3. [model group search](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/model_group/SearchModelGroupTransportAction.java#L72-L74)
4. [model search](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/models/SearchModelTransportAction.java#L49-L51)
5. [task search](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/action/tasks/SearchTaskTransportAction.java#L57-L59)


Please let me know where else we should put this validation?


>The filter by term query doesn’t accept null. We can easily skip adding the filter if we know we’re not multi tenant. But how do we know that?

We already have validation in rest and transport layer for tenant id check. Do we still need to validate tenant ID in SDKClient level? IF that's not necessary, we can skip to add tenant id in the search if tenant Id is null?

dhrubo-os: >that just explains why we substitute a string but not how we treat null throughout the end to end.

And where exactly do you suggest to put this documentation?

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: > We use leading underscores for other important metadata like document Id (_id), as well as _seq_no and _primary_term and _source. It's a well known pattern. What if the user wants to actually use their own "tenant ID" in a document? 

This makes sense, we should mark the metadata fields in an document in an `internal` approach with leading underscore.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -189,7 +195,13 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
...
         return CompletableFuture.supplyAsync(() -> AccessController.doPrivileged((PrivilegedAction<SearchDataObjectResponse>) () -> {
             log.info("Searching {}", Arrays.toString(request.indices()));
-            SearchResponse searchResponse = client.search(new SearchRequest(request.indices(), request.searchSourceBuilder())).actionGet();
+            SearchSourceBuilder searchSource = request.searchSourceBuilder();
+            QueryBuilder existingQuery = searchSource.query();
+            String tenantId = request.tenantId() != null ? request.tenantId() : CommonValue.DEFAULT_TENANT;
+            BoolQueryBuilder boolQuery = QueryBuilders.boolQuery().must(existingQuery == null ? new MatchAllQueryBuilder() : existingQuery);
+            boolQuery.filter(QueryBuilders.termQuery(CommonValue.TENANT_ID, tenantId));
+            searchSource.query(boolQuery);
+            SearchResponse searchResponse = client.search(new SearchRequest(request.indices(), searchSource)).actionGet();
```
zane-neo, line 204: We should add timeout for `actionGet` since if the response pending too long, the threads in the pool can be exhausted.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dhrubo-os: The reason I put `tenant_id` as a index field because other fields in the ml-commons system indices don't have leading `_`.  So only putting `_` for the `tenant_id` does break the consistency with other fields.

Now I feel like `_id`, `_seq_no`, `_primary_term`, `_primary_term` all these are coming from core. What should be the rationale of adding `_` when plugin is introducing this field?

>What if the user wants to actually use their own "tenant ID" in a document?

This problem can happen for any fields we defined in the ml-commons indices, right? Not just for `tenant_id`

>we should mark the metadata fields in an document in an internal approach with leading underscore.

what does `internal` mean here? Internal to whom? Plugin or Core?

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > We already have validation in rest and transport layer for tenant id check.

Lots of individual checks on individual REST and Transport actions don't give me a warm fuzzy.

> And where exactly do you suggest to put this documentation?

I don't know.  As you know, this feature branch is a collaboration primarily of three individuals and theres no documentation anywhere in code/ javadoc/readme/RFC issue/etc. that tells me how to handle Tenant ID consistently.

I recognize we're rapidly iterating on a feature branch and that's totally fine until now when we're talking about migrating to main, and I'm raising my hand with a big red _"the worst possible thing we can do right now is expose one tenant's data to another tenant"_.  

I really do not feel comfortable that we have talked about what is expected here, and what safeguards we have in place to prevent it.  The fact that integ tests are currently failing on this PR is demonstration of the fact that I do not yet understand what the code expects, how we are going to remain compatible with BWC when existing indices do not have tenant ID in their documents, and how we are going to ensure going forward that tenant ID is always available and any attempt by a user to manipulate it should result in failure.   Currently it seems we're relying on a UUID but it seems that spamming a search with a brute force attack of random UUIDs will eventually uncover some random user's data.  Is UUID sufficient?

People (especially me) make mistakes, and failing to insert the REST or Transport validation checks that are apparently already there may happen without at least a test somewhere proving that they work (and perhaps that's my job to write such a test).  I honestly don't understand enough right now about what is expected to write that test and I'm calling that out.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > > we should mark the metadata fields in an document in an internal approach with leading underscore.
> 
> what does `internal` mean here? Internal to whom? Plugin or Core?

A leading underscore in a variable name is a long understood coding convention for internal variables.  
 - Python: https://realpython.com/python-double-underscore/
 - C / C++: https://devblogs.microsoft.com/oldnewthing/20230109-00/?p=107685

It's not enforced, it's not mandatory, but it's what people expect.  Style guides say to not use those types of variable names -- which is precisely why you use them when you don't want to conflict with existing user present or future variable names.

Tell me: of all OpenSearch users now and ever in the future, how will you guarantee that no user will ever use the field name `tenant_id` in a document.

I'd wager many already have; and many will in the future.

> The reason I put `tenant_id` as a index field because other fields in the ml-commons system indices don't have leading `_`. So only putting `_` for the `tenant_id` does break the consistency with other fields.

If it's ml plugin specific then give it a ml plugin specific name.  There just needs to be a common consistent way to tell users "hey, don't use this field name or you'll break things".   Leading underscores is already a well understood way.  If you want to do something else, have a common prefix that you tell users to avoid.    Particularly if you use a non-verbose, very common identifier such as "tenant ID". 

> This problem can happen for any fields we defined in the ml-commons indices, right? Not just for `tenant_id`

Yes.   How are you going to tell users not to overwrite it?  Options are:
 - tell users not to start fields with `_`. (That's already an expectation.)
 - tell users not to start fields with `ml_plugin_`.  Seems reasonable.
 - maintain a list somewhere of reserved names, and include all future names that we'll ever want to use so that we don't break existing indices that might already have those names.   Seems impossible.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: I appreciate the concern but I don't think a timeout is the correct answer.

Do you think these should all be rewritten with ActionListeners?

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > Currently it seems we're relying on a UUID but it seems that spamming a search with a brute force attack of random UUIDs will eventually uncover some random user's data. Is UUID sufficient?

We should not be relying solely on UUIDs as tenant ID to prevent unauthorized access to other tenants' data.

https://www.rfc-editor.org/rfc/rfc4122#section-6
> Do not assume that UUIDs are hard to guess; they should not be used as security capabilities (identifiers whose mere possession grants access), for example.  A predictable random number source will exacerbate the situation.

dbwiddis: > Yeah, I was having the same issue, then I ended up doing this: https://github.com/opensearch-project/ml-commons/pull/2792/files
> 
> In `MLAgentExecutor.java`, which stays in the Algorithm module I needed to have the dynamic multi-tenancy value.

That's a great workaround for local files but it doesn't help in the case of SdkClient when we are planning on moving it out of this repository eventually.  It won't have access to that interface.  

That interface belongs in the SDKClient package(s) somewhere.

zane-neo: > We should not be relying solely on UUIDs as tenant ID to prevent unauthorized access to other tenants' data.

@dbwiddis Do you mean: A cluster has authN/authZ and multi-tenancy is enabled, tenantA's can access the cluster with their credentials but they tempered the tenantId to tenantB's and accessed to tenantB's data?

zane-neo: > That interface belongs in the SDKClient package(s) somewhere.

Can you give an example why this belongs to different package would be an issue? My assumption is there'll be a `tenantId` parameter in the interface and the value will be passed from upper layer, if it's null skip the tenantId filter, otherwise add it.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: If this is metadata in system index, then we don't need to force the leading underscore, since they'll not stored in user's index(not mixed with user data). For `_id`, `_seq_no `, they're mixed with user data, so better use `_` as persevered fields.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: What's the concern on adding timeout? Rewritten to with actionListeners would be better but I think it needs more effort.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > Do you mean: A cluster has authN/authZ and multi-tenancy is enabled

@zane-neo I had assumed that we were producing the tenant ID similarly to how we were producing document ID.  We are not; it is being supplied by the tenant. I no longer have this concern.

dbwiddis: > Can you give an example why this belongs to different package would be an issue?

@zane-neo the SdkClient and its implementations are not meant to permanently reside in ML Commons.  Ideally they will be in another dependency entirely, possibly eventually core or a module, possibly another plugin (similar to how skills is separated).

There needs to be a way to have a plugin supply this client the needed settings updater from the plugin module without relying on this being present in ML Commons.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > If this is metadata in system index, then we don't need to force the leading underscore, since they'll not stored in user's index(not mixed with user data). 

For this particular implementation we are using system indices only in ML plugins.

There are other plugins who may want to use a generic term "tenant_id" for a completely different purpose.  The name should either be more specific, or configurable by a plugin.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: > What's the concern on adding timeout? 

What should the timeout be?  Where do we allow the user to configure it?

> Rewritten to with actionListeners would be better

It does use an action listener internally as part of how this call works.  The future completes when the action listener gets the callback.   Anything I do to wrap an action listener would essentially duplicate the existing code returning a future.

> but I think it needs more effort.

"it needs more effort" is an extremely vague code review comment and isn't really helpful letting me know where you think the problem is or what you think the solution should be.  Can you suggest a specific alternative here?

The entire code block is wrapped inside a `CompletableFuture.supplyAsync()`.  The "search" call returns the future immediately so the only delay in this entire code block is the `actionGet()` call and any additional effort to wrap it is redundant to the existing async methods.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: > Ideally they will be in another dependency entirely, possibly eventually core or a module, possibly another plugin (similar to how skills is separated).

So this is a highly uncertainty, currently it seems fine to implement `skip if null else filter` now, it's not a hard blocker. If there's any early stage decision how this client will be separated, then that'll be a better place to discuss this concern.

dbwiddis: > currently it seems fine to implement skip if null else filter now

That's essentially what I've done already, skipping the filter if multi-tenancy isn't enabled -- but given how tightly this particular setting is with the clients it just makes sense to eventually move it (a later PR is fine) alongside the clients implementing it.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: > What should the timeout be? Where do we allow the user to configure it?

This can be a cluster setting and allow user to dynamically configure it.

> "it needs more effort" is an extremely vague code review comment and isn't really helpful letting me know where you think the problem is or what you think the solution should be. Can you suggest a specific alternative here?

My comment was based on your `rewritten with ActionListeners` and my assumption from this `rewritten` is to change all the API in `SdkClientDelegate` which needs much more effort than adding timeout. E.g. change from
```
CompletionStage<PutDataObjectResponse> putDataObjectAsync(PutDataObjectRequest request, Executor executor);
``` 
to
```
void putDataObjectAsync(PutDataObjectRequest request, ActionListener listener);
```
If this assumption is not correct, please specify how the rewritten will be done, thanks.

dbwiddis: > This can be a cluster setting and allow user to dynamically configure it.

There's already a timeout setting available on the REST params that users can set per-query.

This client is just a wrapper around the `NodeClient` `search()` call.  The NodeClient doesn't have a timeout; users who have timeout concerns can wrap it around their call to this method.  In particular, the `CompletableFuture` returned from this method has a `get()` option with a timeout.  

> My comment was based on your rewritten with ActionListeners

I looked into that and realized that it added no value, so my comment was premature.

I had intentionally used the non-ActionListener version of the API, wrapped in an async execution where the caller had more control over which thread pool was used.

> If this assumption is not correct, please specify how the rewritten will be done, thanks.

At this stage I don't think I will rewrite with `ActionListener` at all.

The only thing I'm currently considering doing is wrapping the `actionGet()` call, and only that call, inside another `supplyAsync()`.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dbwiddis: @zane-neo I looked into wrapping the `actionGet()` in another async thread. Here's what it would look like:
```
    @Override
    public CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(
        SearchDataObjectRequest request,
        Executor executor,
        Boolean isMultiTenancyEnabled
    ) {
        log.info("Searching {}", Arrays.toString(request.indices()));
        SearchSourceBuilder searchSource = request.searchSourceBuilder();
        if (Boolean.TRUE.equals(isMultiTenancyEnabled)) {
            QueryBuilder existingQuery = searchSource.query();
            String tenantId = request.tenantId() != null ? request.tenantId() : CommonValue.DEFAULT_TENANT;
            BoolQueryBuilder boolQuery = QueryBuilders.boolQuery().must(existingQuery == null ? new MatchAllQueryBuilder() : existingQuery);
            boolQuery.filter(QueryBuilders.termQuery(CommonValue.TENANT_ID, tenantId));
            searchSource.query(boolQuery);
        }
        log.info("Searching {}", Arrays.toString(request.indices()));
        ActionFuture<SearchResponse> searchResponseFuture = AccessController
            .doPrivileged((PrivilegedAction<ActionFuture<SearchResponse>>) () -> {
                return client.search(new SearchRequest(request.indices(), searchSource));
            });
        return CompletableFuture
            .supplyAsync(searchResponseFuture::actionGet, executor)
            .thenApply(searchResponse -> {
                log.info("Search returned {} hits", searchResponse.getHits().getTotalHits());
                try {
                    return SearchDataObjectResponse.builder().parser(createParser(searchResponse)).build();
                } catch (IOException e) {
                    // Rethrow unchecked exception on XContent parsing error
                    throw new OpenSearchStatusException(
                        "Failed to search indices " + Arrays.toString(request.indices()),
                        RestStatus.INTERNAL_SERVER_ERROR
                    );
                }
            });
    }
```
I'm not sure it really adds anything here.

At the moment I'm still leaning toward leaving the code as is.  Let me know if you have additional concerns.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -36,515 +36,560 @@
...
 
-        public static final String TENANT_ID = "tenant_id";
+    /**
+     * The field name to be inserted into a document when operating tenant aware.
+     * <p>
+     * The value for this field will be supplied by REST callers using {@code x-tenant-id} header. If no such header is provided, {@link #DEFAULT_TENANT} is used.
+     * <p>
+     * This value should be chosen to avoid conflicts with any existing or future fields.  In the future this may be replaced by a plugin setting.  
+     */
+    public static final String TENANT_ID = "_tenant_id";
```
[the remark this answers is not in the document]
dbwiddis: I argued strongly against "tenant_id" and haven't heard any arguments against it. I certainly don't think we've "agreed" or aligned on anything.

I'm open to anything else that we're pretty sure won't conflict with other plugins that may use the same field name.  
 - What I've done here is put the constant in one place so we can easily change it.
 - I've articulated my view here if you want to oppose it: https://github.com/opensearch-project/ml-commons/pull/2803#discussion_r1703533523

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -186,10 +192,18 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(DeleteDat
...
     @Override
-    public CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(SearchDataObjectRequest request, Executor executor) {
+    public CompletionStage<SearchDataObjectResponse> searchDataObjectAsync(SearchDataObjectRequest request, Executor executor, Boolean isMultiTenancyEnabled) {
         return CompletableFuture.supplyAsync(() -> AccessController.doPrivileged((PrivilegedAction<SearchDataObjectResponse>) () -> {
             log.info("Searching {}", Arrays.toString(request.indices()));
-            SearchResponse searchResponse = client.search(new SearchRequest(request.indices(), request.searchSourceBuilder())).actionGet();
+            SearchSourceBuilder searchSource = request.searchSourceBuilder();
+            if (Boolean.TRUE.equals(isMultiTenancyEnabled)) {
+                QueryBuilder existingQuery = searchSource.query();
+                String tenantId = request.tenantId() != null ? request.tenantId() : CommonValue.DEFAULT_TENANT;
```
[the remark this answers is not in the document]
dbwiddis: Why do we have a default tenant ID if that's the expectation?

We automatically add the default when indexing and it's not really obvious without looking at implementation details what to use. This just keeps the treatment consistent with index/update code (from where I copy-pasted this line).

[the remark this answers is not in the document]
dhrubo-os: @arjunkumargiri as we are getting `multi_tenancy` information in the SDKClient layer, may be we should `final String tenantId = request.tenantId() != null ? request.tenantId() : DEFAULT_TENANT;` throw exception in these lines when tenantId is null for [DDBOpenSearchClient](https://github.com/opensearch-project/ml-commons/blob/feature/multi_tenancy/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java)?

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: This is not async because this line still block waiting: `.supplyAsync(searchResponseFuture::actionGet, executor)`. But if user can somehow set the timeout to this client, then it's fine because it won't exhausted the threads in thread pool.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
zane-neo: LGTM if user can set timeout to this client.

dbwiddis: 1. Search query can use `timeout` field in search query at shard level which will be passed through.
2. The surrounding `CompletableFuture` can use timeout fields to end that query early.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
dhrubo-os: If we don't rely on good intention, which means if user doesn't set the timeout to the client, what happens?

dbwiddis: The same thing that happens on OpenSearch NodeClient search request with no timeout.

The executor thread keeps waiting.

In OpenSearch it's the ActionListener thread.  Here it's a user-settable executor.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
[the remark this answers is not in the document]
dbwiddis: Set it back to "tenant_id"

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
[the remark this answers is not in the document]
dbwiddis: Changed this to throw an exception.  I didn't realize that Default tenant was narrowly scoped only to the hash key for DDB client and not used anywhere else.

PR #2818, opened 2024-08-09
[Feature/multi_tenancy] Tenant-aware integration tests for Connector, Model, Agent, Model Groups

dbwiddis:
### Description

Tests the entire Create-Get-Update-Search-Delete cycle with multi-tenancy (both enabled and not).

Validates expected results when tenant aware and when not (current status quo).

Looking for review on the approach before replicating to models and agents and other tenant-aware updates, although I won't wait before starting that work. If you want a simpler review, get started now. :)

To execute just these tests:
```
./gradlew ':opensearch-ml-plugin:integTest' -Dtests.rest.tenantaware=true
```
or
```
./gradlew ':opensearch-ml-plugin:integTest' -Dtests.rest.tenantaware=false
```

Because the classes end in `IT`, the "false" version executes with normal integ tests as well, a separate test needs to be done for the `true` case.

### Check List

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -192,13 +192,11 @@ integTest {
     if (System.getProperty("tests.rest.tenantaware") != null) {
         filter {
             includeTestsMatching "org.opensearch.ml.rest.*TenantAwareIT"
-            // mock LLM run in localhost, it will not reachable for docker or remote cluster
-            excludeTestsMatching "org.opensearch.ml.tools.VisualizationsToolIT"
         }
     }
 
     // Only rest case can run with remote cluster
-    if (System.getProperty("tests.rest.cluster") != null) {
+    if (System.getProperty("tests.rest.cluster") != null && System.getProperty("tests.rest.tenantaware") == null) {
         filter {
             includeTestsMatching "org.opensearch.ml.rest.*IT"
             excludeTestsMatching "org.opensearch.ml.rest.*TenantAwareIT"
@@ -230,7 +228,7 @@ integTest {
                 ymlFile.withWriterAppend { writer ->
                     writer.write("\n# Use a remote cluster\n")
                     writer.write("plugins.ml_commons.remote_metadata_type: RemoteOpenSearch\n")
-                    writer.write("plugins.ml_commons.remote_metadata_endpoint: http://127.0.0.1\n")
+                    writer.write("plugins.ml_commons.remote_metadata_endpoint: https://127.0.0.1:9200\n")
                 }
             } else {
                 throw new GradleException("opensearch.yml not found at: $ymlFile")
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -189,15 +189,11 @@ integTest {
     systemProperty "password", System.getProperty("password")
 
     // Only tenant aware test if set
-    if (System.getProperty("tests.rest.tenantaware") != null) {
+    if (System.getProperty("tests.rest.tenantaware") == "true") {
         filter {
             includeTestsMatching "org.opensearch.ml.rest.*TenantAwareIT"
         }
         systemProperty "plugins.ml_commons.multi_tenancy_enabled", "true"
-    }
-
-    if (System.getProperty("tests.rest.tenantaware") != null) {
-        environment "plugins.ml_commons.multi_tenancy_enabled", "true"
     }
 
     // Only rest case can run with remote cluster
@@ -226,7 +222,7 @@ integTest {
 
     // The 'doFirst' delays till execution time.
     doFirst {
-        if (System.getProperty("tests.rest.tenantaware") != null) {
+        if (System.getProperty("tests.rest.tenantaware") == "true") {
             def ymlFile = file("$buildDir/testclusters/integTest-0/config/opensearch.yml")
             if (ymlFile.exists()) {
                 ymlFile.withWriterAppend {
@@ -234,24 +230,20 @@ integTest {
                     writer.write("\n# Set multitenancy\n")
                     writer.write("plugins.ml_commons.multi_tenancy_enabled: true\n")
                 }
-            } else {
-                throw new GradleException("opensearch.yml not found at: $ymlFile")
-            }
-        }
-    
-        // TODO this properly uses the remote client factory but needs a remote cluster set up
-        if (System.getProperty("tests.rest.cluster") != null && System.getProperty("tests.rest.tenantaware") != null) {
-            def ymlFile = file("$buildDir/testclusters/integTest-0/config/opensearch.yml")
-            if (ymlFile.exists()) {
-                ymlFile.withWriterAppend { writer ->
-                    writer.write("\n# Use a remote cluster\n")
-                    writer.write("plugins.ml_commons.remote_metadata_type: RemoteOpenSearch\n")
-                    writer.write("plugins.ml_commons.remote_metadata_endpoint: https://127.0.0.1:9200\n")
+                // TODO this properly uses the remote client factory but needs a remote cluster set up
+                // TODO get the endpoint from a system property
+                if (System.getProperty("tests.rest.cluster") != null) {
+                    ymlFile.withWriterAppend { writer ->
+                        writer.write("\n# Use a remote cluster\n")
+                        writer.write("plugins.ml_commons.remote_metadata_type: RemoteOpenSearch\n")
+                        writer.write("plugins.ml_commons.remote_metadata_endpoint: https://127.0.0.1:9200\n")
+                    }
                 }
             } else {
                 throw new GradleException("opensearch.yml not found at: $ymlFile")
             }
         }
+    
         // Tell the test JVM if the cluster JVM is running under a debugger so that tests can
         // use longer timeouts for requests.
         def isDebuggingCluster = getDebug() || System.getProperty("test.debug") != null
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -222,7 +222,7 @@ integTest {
     // The 'doFirst' delays till execution time.
     doFirst {
         // TODO this properly uses the remote client factory but needs a remote cluster set up
-        if (System.getProperty("tests.rest.tenantaware.remote") != null) {
+        if (System.getProperty("tests.rest.cluster") != null && System.getProperty("tests.rest.tenantaware") != null) {
             def ymlFile = file("$buildDir/testclusters/integTest-0/config/opensearch.yml")
             if (ymlFile.exists()) {
                 ymlFile.withWriterAppend { writer ->
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -193,6 +193,11 @@ integTest {
         filter {
             includeTestsMatching "org.opensearch.ml.rest.*TenantAwareIT"
         }
+        systemProperty "plugins.ml_commons.multi_tenancy_enabled", "true"
+    }
+
+    if (System.getProperty("tests.rest.tenantaware") != null) {
+        environment "plugins.ml_commons.multi_tenancy_enabled", "true"
     }
 
     // Only rest case can run with remote cluster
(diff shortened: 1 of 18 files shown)


dhrubo-os:
Should to separate out all the bug related fixes in one PR and then may be a separate PR for integ tests?

dbwiddis:
> Should to separate out all the bug related fixes in one PR and then may be a separate PR for integ tests?

Why?

dhrubo-os:
> > Should to separate out all the bug related fixes in one PR and then may be a separate PR for integ tests?
> 
> Why?

easier review, clearer commit history, simplified debugging and rollback (if needed), maintaining PR scope.

dbwiddis:
`./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.rest.RestMLRAGSearchProcessorIT.testBM25WithCohereUsingLlmResponseField"` is currently failing, output from the model is 
```json
  "ext": {
    "retrieval_augmented_generation": {
      "error": "Unknown error or response."
    }
  }
```

dhrubo-os:
> `./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.rest.RestMLRAGSearchProcessorIT.testBM25WithCohereUsingLlmResponseField"` is currently failing, output from the model is
> 
> ```json
>   "ext": {
>     "retrieval_augmented_generation": {
>       "error": "Unknown error or response."
>     }
>   }
> ```

Let me retry again. Yesterday test was passing in my PRs.

dbwiddis:
> Yesterday test was passing in my PRs.

It passes without the cohere api key change.

Looks like `BM25_SEARCH_REQUEST_WITH_LLM_RESPONSE_FIELD_TEMPLATE` expects 5 strings before 3 ints, but we are sending 7 strings resulting in two nulls:
```json
 {
     "_source": ["text"],
     "query" : {
       "match": {"text": "president"}
     },
      "ext": {
         "generative_qa_parameters": {
           "llm_model": "command",
           "llm_question": "who is lincoln",
           "context_size": null,
           "message_size": null,
           "timeout": 5,
           "llm_response_field": "5"
         }
     }
   }
```

dbwiddis:
Looks like #2246, #2250, and #2257 are related but the tests was disabled at the time (and still lis with your PR).  There was a recent pull to main, probably should put that in in favor of my fix.

dbwiddis:
I just need this commit: https://github.com/opensearch-project/ml-commons/commit/659a836775047e1d8f88a09eced7828d92e65d41

dhrubo-os:
> I just need this commit: [659a836](https://github.com/opensearch-project/ml-commons/commit/659a836775047e1d8f88a09eced7828d92e65d41)

Backported and merged to this branch.

dbwiddis:
> > I just need this commit: [659a836](https://github.com/opensearch-project/ml-commons/commit/659a836775047e1d8f88a09eced7828d92e65d41)
> 
> Backported and merged to this branch.

Thanks! I was in the process of doing that on my PR branch.   I'll see if I can extract some of the bugfix commits from this PR without too much effort.

dbwiddis:
Tests are now passing on this PR but it's failing on coverage checks for the remote model.  Guess I'll add that to jacoco suppressions for now.
```
[ant:jacocoReport] Rule violated for class org.opensearch.ml.sdkclient.SdkClientFactory: lines covered ratio is 0.7, but expected minimum is 0.8
[ant:jacocoReport] Rule violated for class org.opensearch.ml.sdkclient.SdkClientFactory.SocketAccess: lines covered ratio is 0.0, but expected minimum is 0.8
```

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
```diff
@@ -276,7 +276,7 @@ protected void wipeAllODFEIndices() throws IOException {
 
             for (Map<String, Object> index : parserList) {
                 String indexName = (String) index.get("index");
-                if (indexName != null && !".opendistro_security".equals(indexName)) {
+                if (indexName != null && !".opendistro_security".equals(indexName) && !".plugins-ml-config".equals(indexName)) {
```
dhrubo-os, line 279: I don't think we can just do this here. Because for single tenancy we are testing if the index is being created through the code or not. May be only for multi-tenancy we can do this.

dbwiddis: Yes I know.  

Problem is it’s annotated @after so can’t be overridden.  Need a better long term fix. 

Possibly delegate to another overridable method.

dbwiddis: Actually, this fix probably isn't needed, it's just useful to isolate the problem.  This is actually the local index involved here.

What we have is a race condition:
1. When the first test starts it creates a (local) ML Config index (which isn't needed)
2. When the first test finishes it issues a rest call in the `@After` of the superclass to delete the (local) index. This is answered by the primary shard (giving a successful request) but the cluster state isn't instantly updated.
3. The second test starts and checks the cluster state and thinks the index is there, but then it has to check the index mapping and by that time it has been deleted and results in the error.

The error probably exists in all the other ML Commons integ tests, but the unique part about our tests is that several start out immediately creating a connector and calling `encrypt()`.  Let me try throwing a refresh call at the start of every test method to ensure cluster state is caught up and see if it fixes....

dbwiddis: And refreshing worked.  Going to file this as a bug on `main`.

plugin/build.gradle
```diff
@@ -184,10 +187,23 @@ integTest {
     systemProperty "user", System.getProperty("user")
     systemProperty "password", System.getProperty("password")
 
+    // Only tenant aware test if set
+    if (System.getProperty("tests.rest.tenantaware") != null) {
```
dhrubo-os, line 191: don't we need to check `true` too?

dbwiddis: Yay, you're looking at this PR! :)

Yes, we do.  Most of the `build.gradle` right now is a work in progress; we need to figure out a way to test on all the different remote stores but they also require a lot more setup (docker, ddb, etc.).   Still need to do some cleanup at the end of all the tests as well (since the "wipe all indices" doesn't work remotely).

common/src/main/java/org/opensearch/sdk/SdkClient.java
```diff
@@ -19,18 +19,14 @@
 
 import static org.opensearch.sdk.SdkClientUtils.unwrapAndConvertToException;
 
-public class SdkClient implements SettingsChangeListener {
+public class SdkClient {
```
dhrubo-os, line 22: Do we really need to change this?

I understand if we make the settings static, there's no need for the listener. But even if we keep the settings static this piece of code shouldn't break the functionality. 

IMO, it's even a two way door, like if we want to make the settings dynamic, corresponding listener solution is already in place. On top of that we haven't finalized the nature of this settings yet.

common/src/test/java/org/opensearch/sdk/client/LocalClusterIndicesClientTests.java
```diff
@@ -677,4 +674,26 @@ public void testSearchDataObject_NullTenantId() throws IOException {
         assertEquals(OpenSearchStatusException.class, cause.getClass());
         assertEquals("Tenant ID is required when multitenancy is enabled.", cause.getMessage());
     }
+    
+    public void testSearchDataObject_NullTenantNoMultitenancy() throws IOException {
```
dhrubo-os, line 678: add `@Test`

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -185,8 +185,20 @@ private MLCommonsSettings() {}
...
+    // export OPENSEARCH_JAVA_OPTS="-Dplugins.ml_commons.multi_tenancy_enabled=true"
+    // ./bin/opensearch
+    // Or inline when starting OpenSearch:
+    // OPENSEARCH_JAVA_OPTS="-Dplugins.ml_commons.multi_tenancy_enabled=true" ./bin/opensearch
+    // 3. In the opensearch.yml configuration file:
+    // plugins.ml_commons.multi_tenancy_enabled: true
+    // After setting it, a full cluster restart is required for the changes to take effect.
     public static final Setting<Boolean> ML_COMMONS_MULTI_TENANCY_ENABLED = Setting
-        .boolSetting("plugins.ml_commons.multi_tenancy_enabled", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
+        .boolSetting("plugins.ml_commons.multi_tenancy_enabled", false, Setting.Property.NodeScope);
```
dhrubo-os, line 201: Even if we need to make this settings static, do we need to do this now? 
Usually dynamic was helpful for manual testing :D

plugin/src/test/java/org/opensearch/ml/rest/RestMLConnectorTenantAwareIT.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.client.Response;
+import org.opensearch.client.ResponseException;
+import org.opensearch.rest.RestRequest;
+
+public class RestMLConnectorTenantAwareIT extends MLCommonsTenantAwareRestTestCase {
+
+    public void testConnectorCRUD() throws Exception {
```
dhrubo-os, line 21: add `@Test`

dbwiddis:
This is pretty consistently failing on these tests:
```
Tests with failures:
117 tests completed, 2 failed, 14 skipped
 - org.opensearch.ml.rest.RestMLRAGSearchProcessorIT.testOpenAITextEmbeddingModel_ISO8859_1
 - org.opensearch.ml.rest.RestMLRemoteInferenceIT.testOpenAITextEmbeddingModel_ISO8859_1
```
These are not associated with my commits, but possibly related to other recent merges on `feature/multi_tenancy`.

dhrubo-os:
> This is pretty consistently failing on these tests:
> 
> ```
> Tests with failures:
> 117 tests completed, 2 failed, 14 skipped
>  - org.opensearch.ml.rest.RestMLRAGSearchProcessorIT.testOpenAITextEmbeddingModel_ISO8859_1
>  - org.opensearch.ml.rest.RestMLRemoteInferenceIT.testOpenAITextEmbeddingModel_ISO8859_1
> ```
> 
> These are not associated with my commits, but possibly related to other recent merges on `feature/multi_tenancy`.

Yeah I noticed these tests are also failing in 2.x branch. So this is related to OpenAi api. I'll look into this tomorrow to fix.

plugin/src/test/java/org/opensearch/ml/rest/MLCommonsTenantAwareRestTestCase.java
```diff
@@ -0,0 +1,243 @@
...
+    @SuppressWarnings("unchecked")
+    protected static String getErrorReasonFromResponseMap(Map<String, Object> map) {
+        // Two possible cases:
+        String type = ((Map<String, String>) map.get("error")).get("type");
+
+        // {
+        // "error": {
+        // "root_cause": [
+        // {
+        // "type": "status_exception",
```
dhrubo-os, line 133: Why it's showing 2 times?

dhrubo-os: Do you know from where are we getting this `status_exception`? May be we can change that to `OpenSearchStatusException` too? I can make the changes.

dbwiddis: I think this is a difference between local vs. remote client exceptions.

dbwiddis:
Closing in favor of #3516 which squashed and merged these commits

PR #2837, opened 2024-08-20
[BugFix] Avoid overriding the input parameter of the following tools in flow agent

### LINKED ISSUE #2836: [BUG] In AgentTool the input of the first tool will override the manual input of following tools.
**What is the bug?**
If we create an AgentTool containing several tools and some tool may have their own input parameter, when ChatAgentRunner selects this AgentTool to execute, it will generates its input parameter passing to the first tool of AgentTool. However, this input parameter will also override the default manual input parameter of following tools. We should not let it to override since it's only for the input parameter of the first tool.

1. Register an flow agent with 2 tools like:
```
{
  "name": "Test_Agent_For_PPL",
  "type": "flow",
  "description": "this is a test agent for invoking PPLTool",
  "tools": [
    {
      "type": "MLModelTool",
      "parameters": {
        "model_id": "xxxxxxxx",
        "prompt": """You are given a context of alert definition and a nature language question, summarize a final question as the input to generate PPL based that."""
      }
    },
    {
      "type": "PPLTool",
      "parameters": {
        "model_id": "xxxxxxxx",
        "model_type": "FINETUNE",
        "execute": true,
        "input": {
          "question": "${parameters.MLModelTool.output}"
        }
        
      }
    }
  ]
}
```
2. register a chat agent with AgentTool
```
OST /_plugins/_ml/agents/_register
{
  "name": "Test_Agent_For_ReAct_ClaudeV2",
  "type": "conversational",
  "description": "this is an agent to help analysis alert, answering follow-up question of the current alert including search data, create new alert.",
  "llm": {
    "model_id": "xxxxxxxx",
    "parameters": {
      "max_iteration": 5,
      "stop_when_no_tool_found": true
    }
  },
  "memory": {
    "type": "conversation_index"
  },
  "tools": [
    {
      "type": "AgentTool",
      "name": "AgentTool",
      "parameters": {
        "agent_id": "xxxxxx",
         "prompt": """xxxxx"""
       }
    }
  ]
}
```
3. update the current root agent
```
PUT .plugins-ml-config/_doc/os_chat
{"type":"os_chat_root_agent","configuration":{"agent_id":"xxxxxxxxxx"}}
```
6.  In chat window, ask question for providing more data.
7.  In debug mode, you will see the input of PPLTool will be overridden by the input of MLModelTool.

 The input of PPLTool should be our manual set input `{ "question": "${parameters.MLModelTool.output}"}` with placeholder substituted.

 - OS: mac
 - Version: 3.0.0-SNAP
 - Plugins: ml-commons, skills, sql

qianheng-aws:
### Description
Remove the input parameters from `params` in flow agent, since this input should only applies to the current tool instead of overriding all following tools.

### Related Issues
Resolves #2836

### Check List


jngz-es:
1. I don't quite get the fix. The original logic is that if the current tool has `input` parameter, then use it for tool running. In your example, I saw PPLTool having its `input` parameter which should be used to run PPLTool. So I have some confusion here. It's a valid case that a following tool uses the original `input` parameter if they don't have it. Removing it totally doesn't make sense for me.
2. The CI failed due to flow agent test cases, did you try to fix them and add your test case covering your fix? I think you might have new findings when you fixing the existing test cases.

qianheng-aws:
> 1. I don't quite get the fix. The original logic is that if the current tool has `input` parameter, then use it for tool running. In your example, I saw PPLTool having its `input` parameter which should be used to run PPLTool. So I have some confusion here. It's a valid case that a following tool uses the original `input` parameter if they don't have it. Removing it totally doesn't make sense for me.
> 2. The CI failed due to flow agent test cases, did you try to fix them and add your test case covering your fix? I think you might have new findings when you fixing the existing test cases.

1. In my example, the `input` parameter of PPLTool is from `toolSpec.getParameters()`, 
https://github.com/opensearch-project/ml-commons/blob/b51f47f5c5263c93b24e65c22ba7eb71bd3e1ce0/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java#L274 

while what I remove is the `input` parameter from `params`, which is a parameter of `AgentTool`. They are different. My point is the original `input` parameter in `params` should be apply to the first/current tool, so I remove it once it has been put into the `executeParams` of the current tool.

If it also makes sense that `a following tool uses the original input parameter if they don't have it`, I think we should adjust the sequence of putting `toolSpec.getParameters()` and putting `params`, so toolSpec's parameter could override params instead. Do you think this fix is acceptable? @jngz-es 

2. It seems that `params` could be an immutable map, I think the above method of changing sequence won't break this test then.

yuye-aws:
Good to see this PR. Can you also fix the bug in `MLConversationalFlowAgentRunner`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java
```diff
@@ -293,6 +293,9 @@ Map<String, String> getToolExecuteParams(MLToolSpec toolSpec, Map<String, String
             StringSubstitutor substitutor = new StringSubstitutor(executeParams, "${parameters.", "}");
             input = substitutor.replace(input);
             executeParams.put("input", input);
+            // Remove the current input from params to avoid passing it to the following tools.
+            // Otherwise, it will override the default input of the following tools.
+            params.remove("input");
```
yuye-aws, line 298: Intuitively, the method getToolExecuteParams is getting tool input parameters from `params`. The method should be a read-only method on `params`

qianheng-aws:
> Good to see this PR. Can you also fix the bug in `MLConversationalFlowAgentRunner`?

Seems the method of `getToolExecuteParams` is duplicate between `MLConversationalFlowAgentRunner` and `MLFlowAgentRunner`, I think maybe we should deduplicate it by putting it in the super class.

yuye-aws (approved):
Looks good to me.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -480,4 +480,37 @@ public static Map<String, String> constructToolParams(
         }
         return toolParams;
     }
+
+    public static Map<String, String> getToolExecuteParams(MLToolSpec toolSpec, Map<String, String> params) {
+        Map<String, String> executeParams = new HashMap<>();
+        // tooSpec parameter may override the parameters in params.
```
yuye-aws, line 486: tooSpec -> toolSpec

yuye-aws:
@jngz-es Do you know why we do not have unit test file for conversational flow agent?

Issue #2829, opened 2024-08-14
[BUG] model interface input schema validation failed for cohere model


mingshl:
**What is the bug?**
model interface input schema validation failed for cohere model when there is pre processing function, suspect that the model interface validation happens before the pre-processing function, so it throw exceptions before it reaches pre-processing function

following the [blueprint](https://github.com/opensearch-project/ml-commons/blob/main/docs/remote_inference_blueprints/bedrock_connector_cohere_cohere.embed-multilingual-v3_blueprint.md) , got status_exception

**How can one reproduce the bug?**

step 1: create connector follow the blue print:

```
POST /_plugins/_ml/connectors/_create
{
    "name": "Amazon Bedrock Connector: Cohere embed-multilingual-v3",
    "description": "Test connector for Amazon Bedrock Cohere embed-multilingual-v3",
    "version": 1,
    "protocol": "aws_sigv4",
    "credential": {
    },
    "parameters": {
        "region": "us-west-2",
        "service_name": "bedrock",
        "truncate": "NONE",
        "input_type": "search_document",
        "model": "cohere.embed-multilingual-v3"
    },
    "actions": [
        {
            "action_type": "predict",
            "method": "POST",
            "headers": {
                "x-amz-content-sha256": "required",
                "content-type": "application/json"
            },
            "url": "https://bedrock-runtime.${parameters.region}.amazonaws.com/model/${parameters.model}/invoke",
            "request_body": "{ \"texts\": ${parameters.texts}, \"truncate\": \"${parameters.truncate}\", \"input_type\": \"${parameters.input_type}\" }",
            "pre_process_function": "connector.pre_process.cohere.embedding",
            "post_process_function": "connector.post_process.cohere.embedding"
        }
    ]
}
```

Step 2: register and deploy model

```
POST /_plugins/_ml/models/_register
{
  "name": "Bedrock   model ",
  "function_name": "remote",
  "description": "test model",
  "connector_id": "sDsGU5EBVjLDgS3RQdWR"}
  
POST /_plugins/_ml/models/uDsGU5EBVjLDgS3RWdWZ/_deploy

```
Step 3: call predict using the same example from blue print

```
POST /_plugins/_ml/models/uDsGU5EBVjLDgS3RWdWZ/_predict  
{
  "parameters": {
    "texts" : ["上海", "This is a test"]
  }
}


```

got exceptions:

```
{
  "error": {
    "root_cause": [
      {
        "type": "status_exception",
[...]

PR #2852, opened 2024-08-28
Add processed function for remote inference input dataset parameters to convert it back to its orignal datatype

b4sjoo:
### Description
This PR add a method to process the parameters field in the remote inference input dataset to convert it back to its original datatype, instead of a string.

### Related Issues
In #2829 we have identified a model interface issue, however we found the root cause is the parsing method of `RemoteInferenceInputDataset` will make all the value inside `Parameters` map into string. Thus in the model interface validation process we need to parse it back to its original datatype.

### Check List


plugin/src/main/java/org/opensearch/ml/utils/MLNodeUtils.java
```diff
@@ -89,6 +90,41 @@ public static void validateSchema(String schemaString, String instanceString) th
...
+    public static String processRemoteInferenceInputDataSetParametersValue(String inputJson) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode rootNode = mapper.readTree(inputJson);
+
+        if (rootNode.has("parameters") && rootNode.get("parameters").isObject()) {
+            ObjectNode parametersNode = (ObjectNode) rootNode.get("parameters");
+
+            parametersNode.fields().forEachRemaining(entry -> {
+                String key = entry.getKey();
+                System.out.println(key);
```
ylwu-amzn, line 107: remove this line?

dhrubo-os:
apply spotless.

plugin/src/main/java/org/opensearch/ml/utils/MLNodeUtils.java
b4sjoo: Removed

plugin/src/main/java/org/opensearch/ml/utils/MLNodeUtils.java
```diff
@@ -89,6 +90,41 @@ public static void validateSchema(String schemaString, String instanceString) th
...
+                JsonNode value = entry.getValue();
+
+                if (value.isTextual()) {
```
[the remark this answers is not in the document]
b4sjoo: Actually no, this is because if you construct a invalid Json, it will cause parsing exception and thrown earlier in rest/transport layer. If you construct a valid Json, if parameters is either exist, or empty, or with wrong type (array, string, etc.) it will skip. If it is the correct type, then it must have entries.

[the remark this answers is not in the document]
b4sjoo: Add more UTs to address your concern.

b4sjoo:
Tested all automated model interface on a test domain, looks good.

Zhangxunmt (approved):

mingshl (approved):

plugin/src/main/java/org/opensearch/ml/utils/MLNodeUtils.java
```diff
@@ -89,6 +90,39 @@ public static void validateSchema(String schemaString, String instanceString) th
...
+
+            parametersNode.fields().forEachRemaining(entry -> {
+                String key = entry.getKey();
+                JsonNode value = entry.getValue();
+
+                if (value.isTextual()) {
+                    String textValue = value.asText();
+                    try {
+                        // Try to parse the string as JSON
+                        JsonNode parsedValue = mapper.readTree(textValue);
```
ylwu-amzn, line 113: can it handle this case that a parameter is string, but it's a json string. For example 
`"content": "{\"title\": \"hello\"}"`

PR #2865, opened 2024-08-30
[Feature/multi_tenancy] Exclude the document just deleted from the model delete by query

dbwiddis:
### Description

Excludes the just-deleted document from the `DeleteByQuery` for models.  Deleted documents can still show up in search hits, and in edge cases can cause the delete by query to fail.

### Check List
- [x] New functionality includes testing. (Separately in #2818)

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
@@ -41,6 +41,8 @@ import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
 import org.opensearch.index.query.TermsQueryBuilder;
 import org.opensearch.index.reindex.BulkByScrollResponse;
 import org.opensearch.index.reindex.DeleteByQueryAction;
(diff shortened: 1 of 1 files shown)


dbwiddis:
Linux failing on
```
REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.rest.RestMLGetTaskActionIT.testGetTaskAPI_Success" -Dtests.seed=5CEEE75283B1C4E4 -Dtests.security.manager=false -Dtests.locale=ga-IE -Dtests.timezone=Asia/Damascus -Druntime.java=21
```
Same failure as seen on #2866

dbwiddis:
Still flaky on Get Task. Looks like a Memory circuit breaker.
```
org.opensearch.ml.rest.RestMLGetTaskActionIT > testGetTaskAPI_Success STANDARD_ERROR
    REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.rest.RestMLGetTaskActionIT.testGetTaskAPI_Success" -Dtests.seed=DBFC699CE409C27E -Dtests.security.manager=false -Dtests.locale=cs -Dtests.timezone=America/Indianapolis -Druntime.java=21

org.opensearch.ml.rest.RestMLGetTaskActionIT > testGetTaskAPI_Success FAILED
    org.opensearch.client.ResponseException: method [POST], host [http://127.0.0.1:41905/], URI [/_plugins/_ml/_train/sample_algo?async=true], status line [HTTP/1.1 500 Internal Server Error]
    {"error":{"root_cause":[{"type":"m_l_limit_exceeded_exception","reason":"m_l_limit_exceeded_exception: Memory Circuit Breaker is open, please check your resources!"}],"type":"m_l_limit_exceeded_exception","reason":"m_l_limit_exceeded_exception: Memory Circuit Breaker is open, please check your resources!"},"status":500}
```

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -244,8 +246,12 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
     @VisibleForTesting
     void deleteModelChunks(String modelId, Boolean isHidden, ActionListener<Boolean> actionListener) {
         DeleteByQueryRequest deleteModelsRequest = new DeleteByQueryRequest(ML_MODEL_INDEX);
-        deleteModelsRequest.setQuery(new TermsQueryBuilder(MODEL_ID_FIELD, modelId));
-
+        deleteModelsRequest
+            .setQuery(
+                new BoolQueryBuilder()
+                    .must(new TermsQueryBuilder(MODEL_ID_FIELD, modelId)) // Match documents with the same model_id
+                    .mustNot(new TermQueryBuilder("_id", modelId)) // exclude the document just deleted
```
dhrubo-os, line 253: Could we explain more in the comment why do we need to exclude the document just deleted? 
In addition, why is this affecting multi-tenancy use case as this `deleteModelChunks` is for local models.

dbwiddis: > Could we explain more in the comment why do we need to exclude the document just deleted?

Sure, I'll add a comment.

Search results aren't always up to date until a refresh.  In this case we were deleting an object and then immediately searching.  I think this bug was triggered by an edge case where there are no chunks, but one (deleted) document matched the search.

I don't think it's wrong to make this change; seems a marginal possible performance improvement with one less search hit.

> In addition, why is this affecting multi-tenancy use case as this deleteModelChunks is for local models.

This was fixed before you added code to skip deleteModelChunks in this case.  This fixed that bug before catching up with your change, which is still good for performance reasons.

But it's a correct/general fix that should generally apply.

dhrubo-os (approved):

PR #2866, opened 2024-08-30
[Feature/multi_tenancy] Delay context restoring for initializing master key and deleting agent

dbwiddis:
### Description

Fixes too-eager restoring of context in cases where it needed to still be stashed

### Check List
- [x] New functionality includes testing. (Separately in #2818)

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
@@ -95,20 +95,19 @@ public class DeleteAgentTransportAction extends HandledTransportAction<ActionReq
             .build();
 
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-
+            ActionListener<DeleteResponse> wrappedListener = ActionListener.runBefore(actionListener, context::restore);
             sdkClient
                 .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor(GENERAL_THREAD_POOL))
                 .whenComplete((r, throwable) -> {
                     log.debug("Completed Get Agent Request, Agent id:{}", agentId);
                     if (throwable != null) {
-                        context.restore();
                         Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
                         if (cause instanceof IndexNotFoundException) {
                             log.info("Failed to get Agent index", cause);
-                            actionListener.onFailure(new OpenSearchStatusException("Failed to get agent index", RestStatus.NOT_FOUND));
+                            wrappedListener.onFailure(new OpenSearchStatusException("Failed to get agent index", RestStatus.NOT_FOUND));
                         } else {
                             log.error("Failed to get ML Agent {}", agentId, cause);
-                            actionListener.onFailure(cause);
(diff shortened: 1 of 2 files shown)


dbwiddis:
Windows failing on
```
REPRODUCE WITH: gradlew ':opensearch-ml-plugin:test' --tests "org.opensearch.ml.action.model_group.SearchModelGroupITTests.test_term_search" -Dtests.seed=57E7F417B46750BC -Dtests.security.manager=false -Dtests.locale=lv -Dtests.timezone=Kwajalein -Druntime.java=17
```

Linux failing on
```
REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.rest.RestMLGetTaskActionIT.testGetTaskAPI_Success" -Dtests.seed=34FFF2F5CDF198BA -Dtests.security.manager=false -Dtests.locale=ar-DZ -Dtests.timezone=America/Curacao -Druntime.java=21
```

Neither is associated with this PR and pass on the alternate OS

plugin/src/main/java/org/opensearch/ml/action/agents/DeleteAgentTransportAction.java
```diff
@@ -124,6 +124,7 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
                                     if (TenantAwareHelper
                                         .validateTenantResource(mlFeatureEnabledSetting, tenantId, mlAgent.getTenantId(), actionListener)) {
                                         if (mlAgent.getIsHidden() && !isSuperAdmin) {
+                                            context.restore();
```
dhrubo-os, line 127: Or should we stash the context again before the deletion?

dbwiddis: This particular branch doesn't do the deletion, it's returning the 403 message because you are not superadmin and the model is hidden.  The else branch keeps the context stashed and does the deletion.

dhrubo-os: Yeah, I know but I wasn't able to comment there as that wasn't a modified code section. In summary, I was just wondering if we should add context restore in multiple places? Or should we just stash again before performing any DB call. That will keep the stash context scope smaller and more secured IMO.

dbwiddis: The previous code used `runBefore()` on the action listener which basically just delayed restoring it until the action lister was completed (with success or failure).   I was somewhat trying to restore earlier with the initial code but it was too early.  This delays it for when it's needed.

I can probably put it back closer to the `runBefore()` implementation which is the opposite direction of your wondering... or since it's used in try-with-resources, it actually will be automatically restored as part of that auto-closure, so manually doing anything to it isn't actually required here.

I'm happy to go either way, which do you prefer?

dhrubo-os: I agree with you. Let's put it back closer to the `runBefore()` and then we can make the changes in main to keep the context scope smaller.

dhrubo-os (approved):

Issue #2839, opened 2024-08-20
[META] ML Inference Processor Enhancements 


mingshl:
**Is your feature request related to a problem?**

**Search response processor:** 

- [x] support one to one inference in ml inference search response processor
- [x] support list in substring during prediction API, so it can support GenAI/RAG use case in ml inference search response with prompt defined in connector level 
- [x] support custom prompt in model config, currently it will cause escape problem, due to model Input change in this issue: https://github.com/opensearch-project/ml-commons/issues/2822 

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

mingshl:
@opensearch-project/admin -- Can you please move this to our roadmap https://github.com/orgs/opensearch-project/projects/1 under 2.17?

mingshl:
support one to one inference in ml inference search response processor in #2801 

support list in substring during prediction API, so it can support GenAI/RAG use case in ml inference search response with prompt defined in connector level in https://github.com/opensearch-project/ml-commons/pull/2871

adding a to_string() method in httpConnector to support custom prompt in https://github.com/opensearch-project/ml-commons/pull/2871

PR #2811, opened 2024-08-07
Support list in response body

mingshl:
### Description

When an array as a string in parameter, the string in list are not escaped. 

This PR will escape the string when it's a json array. 

for example, before this change: 
```
POST /_plugins/_ml/connectors/_create
{
  "name": "Amazon Bedrock Connector: Claude Instant V1",
  "version": "1",
  "description": "The connector to bedrock Claude model",
  "protocol": "aws_sigv4",
  "parameters": {
    "max_tokens_to_sample": "8000",
    "service_name": "bedrock",
    "temperature": "1.0E-4",
    "response_filter": "$.completion",
    "region": "us-west-2",
    "anthropic_version": "bedrock-2023-05-31",
    "inputs":"please summerize the documents"
  },
  "credential": {
    "access_key": " ",
        "secret_key": " ",
        "session_token": "" },
  "actions": [
    {
      "action_type": "PREDICT",
      "method": "POST",
      "url": "https://bedrock-runtime.us-west-2.amazonaws.com/model/anthropic.claude-instant-v1/invoke",
      "headers": {
        "x-amz-content-sha256": "required",
        "content-type": "application/json"
      },
      "request_body":  "{\"prompt\":\"${parameters.prompt}\",\"max_tokens_to_sample\":300,\"temperature\":0.5,\"top_k\":250,\"top_p\":1,\"stop_sequences\":[\"\\n\\nHuman:\"]}"
    }
  ]
}


POST /_plugins/_ml/models/_register
{
  "name": "Bedrock agent model with prompt",
  "function_name": "remote",
  "description": "test model",
  "connector_id": "0CwoD5EB6KAJXDLxYDvd"}
  
POST /_plugins/_ml/models/2SwoD5EB6KAJXDLxezto/_deploy

## success with string
POST /_plugins/_ml/models/2SwoD5EB6KAJXDLxezto/_predict  
{
  "parameters": {
    "prompt":"\n\nHuman: You are a professional data analysist. You will always answer question based on the given context first. If the answer is not directly shown in the context, you will analyze the data and find the answer. If you don't know the answer, just say I don't know. Context: ${parameters.context}. \n\n Human: please summarize the documents \n\n Assistant:",
[...]

PR #2871, opened 2024-08-31
Fix custom prompt substitute with List issue in ml inference search response processor 

mingshl:
### Description
fix custom prompt substitute with List issue in ml inference search response processor 

using toString() method in placeholder to convert list into string, for example, `${parameters.context.toString()}` in the following processors config: 

```
PUT /_search/pipeline/my_pipeline_request_review_llm
{
  "response_processors": [
    {
      "ml_inference": {
        "tag": "ml_inference",
        "description": "This processor is going to run llm",
        "model_id": "cf46K5EBoVpekzRp8x_3",
        "function_name": "REMOTE",
        "input_map": [
          {
            "context": "review"
          }
        ],
        "output_map": [
          {
            "llm_response": "response"
          }
        ],
        "model_config": {
          "prompt":"\n\nHuman: You are a professional data analysist. You will always answer question based on the given context first. If the answer is not directly shown in the context, you will analyze the data and find the answer. If you don't know the answer, just say I don't know. Context: ${parameters.context.toString()}. \n\n Human: please summarize the documents \n\n Assistant:"
        },
        "ignore_missing": false,
        "ignore_failure": false
      }
    }
  ]
}
```

###

reverting https://github.com/opensearch-project/ml-commons/pull/2811/files  as a more appropriate approach to fix the toString issue https://github.com/opensearch-project/ml-commons/issues/2839 

### Check List


common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -322,6 +323,7 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
         if (connectorAction.isPresent() && connectorAction.get().getRequestBody() != null) {
             String payload = connectorAction.get().getRequestBody();
             payload = fillNullParameters(parameters, payload);
+            parseParameters(parameters);
```
ylwu-amzn, line 326: With this PR, can we revert this https://github.com/opensearch-project/ml-commons/pull/2811/files ?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -233,4 +233,49 @@ public static String getErrorMessage(String errorMessage, String modelId, Boolea
...
+        return prefixes;
+    }
+
+    /**
+     * Parses the given parameters map and processes the values containing toString() method calls.
+     *
+     * @param parameters A map containing key-value pairs where the values may contain toString() method calls.
+     * @return A new map with the processed values for the toString() method calls.
+     */
+    public static Map<String, String> parseParameters(Map<String, String> parameters) {
```
ylwu-amzn, line 265: Add prefix to method parameter list ?  For connector, it will be `parameters`. For other usecase, it maybe different, like local model could be `model_config`
```suggestion
    public static Map<String, String> parseParameters(String prefix, Map<String, String> parameters) {
```

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -233,4 +233,49 @@ public static String getErrorMessage(String errorMessage, String modelId, Boolea
...
+     *
+     * @param map A map containing key-value pairs where the values may contain toString() method calls.
+     * @return A list of prefixes for the toString() method calls found in the map values.
+     */
+    public static List<String> collectToStringPrefixes(Map<String, String> map) {
+        List<String> prefixes = new ArrayList<>();
+        for (String key : map.keySet()) {
+            String value = map.get(key);
+            if (value != null) {
+                Pattern pattern = Pattern.compile("\\$\\{(\\w+\\.\\w+)\\.toString\\(\\)\\}");
```
ylwu-amzn, line 248: For connector, we should only check `${parameters.xxx}` , but this pattern will also check `${model_config.xxx}`.

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
```diff
@@ -218,4 +224,184 @@ public void testGetErrorMessageWhenHiddenNull() {
...
+            .put(
+                "prompt",
+                "\\n\\nHuman: You are a professional data analyst. You will always answer question based on the given context first. If the answer is not directly shown in the context, you will analyze the data and find the answer. If you don't know the answer, just say I don't know. Context: ${parameters.context.toString()}. \\n\\n Human: please summarize the documents \\n\\n Assistant:"
+            );
+        ArrayList<String> listOfDocuments = new ArrayList<>();
+        listOfDocuments.add("document1");
+        parameters.put("context", toJson(listOfDocuments));
+
+        parseParameters(parameters);
+        System.out.println(parameters.get("context" + TO_STRING_FUNCTION_NAME));
```
ylwu-amzn, line 295: Remove the test code using `System.out.println`?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
mingshl: sure, we can do that. I think this is a more appropriate fix. Will raise a new PR to revert it.

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
mingshl: removed. thanks!

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
mingshl: all parameters will be prefix with `parameters` in   here 

https://github.com/opensearch-project/ml-commons/blob/59b417be215290c9af19bb2e4e1a4bfe5d007ec7/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java#L327

I added the test to test when it's `parameters.model_config.context`, then it would be taking the prefix as `model_config.context` properly, please see this UT 

https://github.com/opensearch-project/ml-commons/blob/59b417be215290c9af19bb2e4e1a4bfe5d007ec7/common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java#L316-L340

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
mingshl: changed to regex pattern to `"\\$\\{parameters\\.(.+?)\\.toString\\(\\)\\}"`

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
mingshl: can you please review again? @ylwu-amzn

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
mingshl: I will raise the new PR after this PR get merged, because it will cause conflict.

ylwu-amzn (dismissed):

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
mingshl: reverting @mingshl
[Revert "Support list in response body (](https://github.com/opensearch-project/ml-commons/pull/2871/commits/247b955b5a014adf36dafb713cf6be15f8d53fcb)https://github.com/opensearch-project/ml-commons/pull/2811[)"](https://github.com/opensearch-project/ml-commons/pull/2871/commits/247b955b5a014adf36dafb713cf6be15f8d53fcb) in this PR

ylwu-amzn (approved):

Zhangxunmt (approved):

dhrubo-os:
Shouldn't we use model interface instead of adding `toString` to the parameter to identify if we need to convert that to string? cc @ylwu-amzn

PR #2872, opened 2024-09-01
[Feature/multi_tenancy] Fix remote client configuration and mapping issues

dbwiddis:
### Description

Updates remote client configuration:
 - Allows HTTP connection to a remote store to simplify testing without dealing with certificates
 - Fixes two mapping issues by creating a serializer that properly transforms any ToXContentObject to correct JSON-P that the remote client understands.
   - the `version` field on the `MLModel` object was incorrectly given a `String` type, due to internal naming of the new `model_version` field with an internal `version` name, and automatic type mapping via reflection. Worse, the master key JSON wasn't being populated because it had an anonymous ToXContentObject returning a map but having no fields.  
 - Fixes mapping date types on remote model default to nanosecond precision, turning `long` into `double` and breaking parsing.  Changed to millisecond precision so a proper `long` is returned.  (Fix is on the ObjectMapper setting nanosecond precision to false)
 - Worked around enum casing differences in server and client

### Related Issues

Cherry-picked from #2818 

### Check List
- [x] New functionality includes testing. (Part of #2818)

### DIFF
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -8,9 +8,12 @@
  */
 package org.opensearch.sdk;
 
+import java.util.Locale;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.opensearch.OpenSearchException;
 import org.opensearch.common.action.ActionFuture;
@@ -50,5 +53,33 @@ public class SdkClientUtils {
             return throwable.getCause().getCause();
         }
         return throwable;
+    }
+
+    /**
+     * If an internal variable name matches a field name not assigned to that field, Jackson's ObjectMapper can parse the wrong type. This method removes quotes around numeric fields to work around these cases.
+     * @param field The JSON field to remove the quotes from its value
+     * @param json The full JSON to process
+     * @return The JSON with the quotes removed from a numeric value in the specified field
+     */
+    public static String unwrapQuotedInteger(String field, String json) {
+        String regex = "(\"" + Pattern.quote(field) + "\"):\"(\\d+)\"";
+        return json.replaceAll(regex, "$1:$2");
+    }
+
+    /**
+     * If an internal variable is an enum represented by all upper case, the Remote client may have it mapped in lower case. This method lowercases these enum values 
+     * @param field The JSON field to lowercase the value
+     * @param json The full JSON to process
+     * @return The JSON with the value lowercased
+     */
+    public static String lowerCaseEnumValues(String field, String json) {
+        // Use a matcher to find and replace the field value in lowercase
+        Matcher matcher = Pattern.compile("(\"" + Pattern.quote(field) + "\"):(\"[A-Z_]+\")").matcher(json);
+        StringBuffer sb = new StringBuffer();        
+        while (matcher.find()) {
+            matcher.appendReplacement(sb, matcher.group(1) + ":" + matcher.group(2).toLowerCase(Locale.ROOT));
+        }
+        matcher.appendTail(sb);        
+        return sb.toString();
     }
 }
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -56,14 +56,14 @@ public class SdkClientUtils {
     }
 
     /**
-     * If an internal variable name matches a field name not assigned to that field, Jackson's ObjectMapper can parse the wrong type. This method removes quotes around numeric fields to work around these cases.
+     * If an internal variable name matches a field name not assigned to that field, Jackson's ObjectMapper can assign it to the wrong name. This will align with XContent parsing.
      * @param field The JSON field to remove the quotes from its value
+     * @param newField The JSON field that it's supposed to be
      * @param json The full JSON to process
-     * @return The JSON with the quotes removed from a numeric value in the specified field
+     * @return The JSON with the field renamed
      */
-    public static String unwrapQuotedInteger(String field, String json) {
-        String regex = "(\"" + Pattern.quote(field) + "\"):\"(\\d+)\"";
-        return json.replaceAll(regex, "$1:$2");
+    public static String renameField(String field, String newField, String json) {
+        return json.replaceAll("\"" + Pattern.quote(field) + "\"", "\"" + Pattern.quote(newField) + "\"");
     }
 
     /**
--- a/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
+++ b/common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
@@ -56,17 +56,6 @@ public class SdkClientUtils {
     }
 
     /**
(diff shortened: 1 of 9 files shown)


dbwiddis:
Flipping to draft: identified the root cause of the mapping bug and will update this PR to correct it.

dbwiddis:
Fixed!

Before:
```json
{
    "_index": ".plugins-ml-config",
    "_id": "master_key_smzth7yZSwhXLMVS5KKb7tt15cAjHYzyfvfZo074Wzg=",
    "_score": 1.0,
    "_source": {
        "fragment": false
    }
},
```
After:
```json
{
    "_index": ".plugins-ml-config",
    "_id": "master_key_Lpv4sSf4q3SgYRHvc3wRwiV_Z0vTGs3Y3sjQA7iaqaQ=",
    "_score": 1.0,
    "_source": {
        "create_time": 1725427226701,
        "tenant_id": "TNEEE",
        "master_key": "bonPRPstRvZ3Jq2/BRwgVWb1Ari4AszjLWbFnRIfzbA="
    }
},
```
and the version naming issue on ML Models is also corrected (it has `model_version` as it should and not `version`)
```json
{
    "_index": ".plugins-ml-model",
    "_id": "rYN6u5EB_yGwbcQaJRdI",
    "_score": 1.0,
    "_source": {
        "name": "remote model for connector_id qoN6u5EB_yGwbcQaGxfX",
        "model_group_id": "q4N6u5EB_yGwbcQaIRd9",
        "algorithm": "REMOTE",
        "model_version": "1",
        "description": "test model",
        "model_state": "REGISTERED",
        "created_time": 1725427229836,
        "last_updated_time": 1725427229836,
        "is_hidden": false,
        "connector_id": "qoN6u5EB_yGwbcQaGxfX",
        "tenant_id": "TNEEE"
    }
},
```

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
```diff
@@ -100,6 +118,21 @@ static SdkClient wrapSdkClientDelegate(SdkClientDelegate delegate) {
     private static DynamoDbClient createDynamoDbClient(String region) {
         if (region == null) {
             throw new IllegalStateException("REGION environment variable needs to be set!");
+        } else if (region.equals("local")) {
```
dhrubo-os, line 121: can we add more comment here like why we will give `local` as region? And local is tied to `us-west-2`? Is this for integration test?

plugin/build.gradle
```diff
@@ -348,7 +352,9 @@ List<String> jacocoExclusions = [
         'org.opensearch.ml.helper.ModelAccessControlHelper',
         'org.opensearch.ml.action.models.DeleteModelTransportAction.2',
         'org.opensearch.ml.model.MLModelCacheHelper',
-        'org.opensearch.ml.model.MLModelCacheHelper.1'
+        'org.opensearch.ml.model.MLModelCacheHelper.1',
+        'org.opensearch.ml.sdkclient.SdkClientFactory',
```
dhrubo-os, line 356: Is that a `TODO`?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
dbwiddis: I actually was considering just removing this, as it only really works for local DDB container which is useful for some testing.

This is part of my frustration in trying to split out parts of the integ test that aren't really parts of the test, like client configuration.  Some are temporary, and some build on each other as we iterate.

As for the region, it actually can be any region, and any credentials, they just need to match your local `.aws/config` and `.aws/credential` settings.  

If you want me to submit a commit removing it, I will.  It's getting to be more of a waste of time administratively keeping two separate PRs up to date with incremental changes as we learn more about client setup.

plugin/build.gradle
dbwiddis: It's probably the local config causing coverage to drop, one more reason to remove it until it is properly implemented.

dbwiddis: It's also possible that it's not the local config, it's just the accumulation of lines of code on the Remote Client code branches, which aren't covered in unit tests.

I am strongly opposed to arbitrary minimum code coverage percentages and this is one reason why. The only alternative to maintainers is just to turn off the metric, rather than seek for maximum coverage. 90% isn't good enough if you've missed coverable lines but a 70% metric implies it is.  Sometimes 66% is all you can get without integ tests. Which I've written which cover all of these cases.

common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
```diff
@@ -51,4 +54,21 @@ public static Throwable getRethrownExecutionExceptionRootCause(Throwable throwab
...
         return throwable;
     }
+
+    /**
+     * If an internal variable is an enum represented by all upper case, the Remote client may have it mapped in lower case. This method lowercases these enum values 
+     * @param field The JSON field to lowercase the value
+     * @param json The full JSON to process
+     * @return The JSON with the value lowercased
+     */
+    public static String lowerCaseEnumValues(String field, String json) {
```
dhrubo-os, line 64: Can't we use GSON to manipulate the JSON Structure? like:

```
public static String lowerCaseEnumValues(String field, String json) {
    // Parse the JSON string into a JsonObject
    JsonElement jsonElement = JsonParser.parseString(json);
    
    if (jsonElement.isJsonObject()) {
        JsonObject jsonObject = jsonElement.getAsJsonObject();
        JsonElement fieldElement = jsonObject.get(field);

        // Check if the field exists and is a string
        if (fieldElement != null && fieldElement.isJsonPrimitive() && fieldElement.getAsJsonPrimitive().isString()) {
            String value = fieldElement.getAsString();
            
            // If the value is an all-uppercase string, convert it to lowercase
            if (value.matches("[A-Z_]+")) {
                jsonObject.addProperty(field, value.toLowerCase(Locale.ROOT));
            }
        }
    }
    
    // Return the modified JSON as a string
    return new Gson().toJson(jsonElement);
}
```

plugin/build.gradle
dhrubo-os: Do you have any suggestion in mind?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
dhrubo-os: So I'm going to approve and merge this PR for your easy development, let's revisit this later.

dhrubo-os (approved):

common/src/main/java/org/opensearch/sdk/SdkClientUtils.java
dbwiddis: @dhrubo-os we could, but that brings the overhead of entirely parsing the String into a whole tree of Java objects, then re-creating the String, just to do a string replacement on a single value.  

I'm hoping this is temporary code anyway, as it's possible the java client will update to handle this.

plugin/build.gradle
dbwiddis: @dhrubo-os my general suggestion is to try to cover everything and evaluate missing coverage with an eye toward whether it is tested elsewhre (like integ tests).  I'm not a maintainer on this repo so it's not my place to overturn what you have in case; but the point here is that the more a class involves externalities (remote client etc.) the less you will be able to test it with unit tests, and the less meaning coverage will have.

Sure, you can write a bunch of mocks and "exercise" the method to artificially increase your coverage numbers, which doesn't mean the code actually works. :)

Ultimately my suggestion is to cover what you can and if it's not above an arbitrarily chosen metric, suppress it from failing like is done here.

PR #2882, opened 2024-09-03
fix breaking changes in config index fields

rbhavna:
### Description
fix breaking changes in config index fields

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLConfig.java
@@ -10,6 +10,7 @@ import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedTok
 import java.io.IOException;
 import java.time.Instant;
 
+import org.opensearch.Version;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
@@ -28,16 +29,22 @@ public class MLConfig implements ToXContentObject, Writeable {
 
     public static final String TYPE_FIELD = "type";
 
-    public static final String CONFIG_TYPE_FIELD = "config_type";
-
     public static final String CONFIGURATION_FIELD = "configuration";
-
-    public static final String ML_CONFIGURATION_FIELD = "ml_configuration";
 
     public static final String CREATE_TIME_FIELD = "create_time";
     public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
 
+    // Adding below three new fields since the original fields, type, configuration, and last_update_time
+    // are not created with correct data types in config index due to missing schema version bump.
+    // Starting 2.15, it is suggested that below fields be used for creating new documents in config index
+
+    public static final String CONFIG_TYPE_FIELD = "config_type";
+
+    public static final String ML_CONFIGURATION_FIELD = "ml_configuration";
+
     public static final String LAST_UPDATED_TIME_FIELD = "last_updated_time";
+
+    private static final Version MINIMAL_SUPPORTED_VERSION_FOR_NEW_CONFIG_FIELDS = CommonValue.VERSION_2_15_0;
 
     @Setter
     private String type;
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -42,55 +42,93 @@ public class MLConfig implements ToXContentObject, Writeable {
     @Setter
     private String type;
 
+    @Setter
+    private String configType;
```
ylwu-amzn, line 46: Add some comment like this is a new field to replace `type` field ?

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -42,55 +42,93 @@ public class MLConfig implements ToXContentObject, Writeable {
...
+        }
         createTime = input.readOptionalInstant();
         lastUpdateTime = input.readOptionalInstant();
+        lastUpdatedTime = input.readOptionalInstant();
     }
 
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         out.writeOptionalString(type);
+        out.writeOptionalString(configType);
```
ylwu-amzn, line 90: Should we consider OS versions ?

rbhavna: Yeah I think so. There are going to be like completely new fields now. Not replacing the old ones

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -74,23 +106,42 @@ public void writeTo(StreamOutput out) throws IOException {
...
+        if (streamOutputVersion.onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_NEW_CONFIG_FIELDS)) {
+            out.writeOptionalString(configType);
+            if (mlConfiguration != null) {
+                out.writeBoolean(true);
+                mlConfiguration.writeTo(out);
+            } else {
+                out.writeBoolean(false);
+            }
+        }
+        out.writeOptionalInstant(lastUpdatedTime);
```
ylwu-amzn, line 118: Why not add this new field to if block above?

rbhavna: sure, missed it somehow

ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -74,23 +106,42 @@ public void writeTo(StreamOutput out) throws IOException {
...
+        }
+        if (configType != null) {
+            builder.field(CONFIG_TYPE_FIELD, configType);
         }
         if (configuration != null) {
-            builder.field(ML_CONFIGURATION_FIELD, configuration);
+            builder.field(CONFIGURATION_FIELD, configuration);
+        }
+        if (mlConfiguration != null) {
+            builder.field(ML_CONFIGURATION_FIELD, mlConfiguration);
```
lezzago, line 134: What happens when `ml_configuration` and `configuration` are both populated. Same with `config_type` and `config`? Will this cause any issues in the plugin?

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -28,43 +29,74 @@ public class MLConfig implements ToXContentObject, Writeable {
...
+        this.configType = configType;
         this.configuration = configuration;
+        this.mlConfiguration = mlConfiguration;
         this.createTime = createTime;
         this.lastUpdateTime = lastUpdateTime;
+        this.lastUpdatedTime = lastUpdatedTime;
     }
 
     public MLConfig(StreamInput input) throws IOException {
+        Version streamInputVersion = input.getVersion();
```
lezzago, line 81: Shouldn't this be optional since it was not there in previous versions for upgrades?

lezzago, line 81: Shouldn't this be optional since it was not there in previous versions for upgrades?

common/src/main/java/org/opensearch/ml/common/MLConfig.java
rbhavna: No there wouldn't be any issue but both will be shown in the response.

common/src/main/java/org/opensearch/ml/common/MLConfig.java
rbhavna: I didn't quite get it. This is only to ensure these new fields can be used only on domains starting from 2.15

lezzago: Oh nevermind, I had made some confusion with the parse function. There should be no issues here.

lezzago (approved):

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/MLConfig.java
```diff
@@ -74,23 +106,42 @@ public void writeTo(StreamOutput out) throws IOException {
...
 
     @Override
     public XContentBuilder toXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
         XContentBuilder builder = xContentBuilder.startObject();
         if (type != null) {
-            builder.field(CONFIG_TYPE_FIELD, type);
+            builder.field(TYPE_FIELD, type);
+        }
+        if (configType != null) {
+            builder.field(CONFIG_TYPE_FIELD, configType);
```
ylwu-amzn, line 128:
```
builder.field(CONFIG_TYPE_FIELD, configType);
builder.field(TYPE_FIELD, type);
```

PR #2889, opened 2024-09-04
Ensure cluster state is up to date with index deletions after integ test

### LINKED ISSUE #2888: [BUG] Race Condition when running multiple integ tests invoking EncryptorImpl.encrypt early
**What is the bug?**

> What we have is a race condition:
> 1. When the first test starts it creates a (local) ML Config index (which isn't needed)
> 2. When the first test finishes it issues a rest call in the `@After` of the superclass to delete the (local) index. This is answered by the primary shard (giving a successful request) but the cluster state isn't instantly updated.
> 3. The second test starts and checks the cluster state and thinks the index is there, but then it has to check the index mapping and by that time it has been deleted and results in the error.
> 
> The error probably exists in all the other ML Commons integ tests, but the unique part about our tests is that several start out immediately creating a connector and calling `encrypt()`.  Let me try throwing a refresh call at the start of every test method to ensure cluster state is caught up and see if it fixes....
> 

_Originally posted by @dbwiddis in https://github.com/opensearch-project/ml-commons/pull/2818#discussion_r1744345431_

1. Write integration tests which extend `MLCommonsRestTestCase`
2. Have multiple tests, as their first step, create a connector (or anything else that invokes `EncryptorImpl.encrypt()`.)
3. Execute the tests filtered to ensure that tests creating the `.plugins-ml-config` index execute sequentially.  Ideally, have the cluster under high CPU load so cluster updates are slower.
4. Observe a `ResourceNotFound` exception similar to this:

> [2024-09-03 12](tel:2024090312):57:47 [2024-09-03T19:57:47,678][ERROR][o.o.m.a.c.TransportCreateConnectorAction] [790661ea3297] Failed to create connector Cohere Chat Model
> [2024-09-03 12](tel:2024090312):57:47 org.opensearch.ResourceNotFoundException: The ML encryption master key has not been initialized yet. Please retry after waiting for 10 seconds.
> [2024-09-03 12](tel:2024090312):57:47 at org.opensearch.ml.engine.encryptor.EncryptorImpl.checkMasterKeyInitialization(EncryptorImpl.java:387) ~[opensearch-ml-algorithms-multi-tenancy-2.16.0-SNAPSHOT.jar:?]
> [2024-09-03 12](tel:2024090312):57:47 at org.opensearch.ml.engine.encryptor.EncryptorImpl.initMasterKey(EncryptorImpl.java:134) ~[opensearch-ml-algorithms-multi-tenancy-2.16.0-SNAPSHOT.jar:?]

Tests start with no existing indices (except for security).

 - OS: Reproduced locally on macOS but also occurs on CI tests

The root of the problem is this method:
```java
@SuppressWarnings("unchecked")
@After
protected void wipeAllODFEIndices() throws IOException {
    Response response = adminClient().performRequest(new Request("GET", "/_cat/indices?format=json&expand_wildcards=all"));
// ... other code ...
        for (Map<String, Object> index : parserList) {
            String indexName = (String) index.get("index");
            if (indexName != null && !".opendistro_security".equals(indexName)) {
                adminClient().performRequest(new Request("DELETE", "/" + indexName));
            }
        }
    }
}
```

The `performRequest()` call doesn't wait for the cluster state to be updated before returning, so the test case finishes and the next one is allowed to start while the cluster state is still processing the deletions.

Suggested fix:  add a `refreshAllIndices()` call at the end of this method.

dbwiddis:
### Description

Ensures the requested index deletions are updated in the cluster state prior to ending an integ test, to avoid impacting follow-on tests.

### Related Issues

Fixes #2888 

### Check List
- [x] New functionality includes testing. (it's part of testing and causes flaky failures which have stopped!)

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
@@ -257,7 +257,7 @@ public abstract class MLCommonsRestTestCase extends OpenSearchRestTestCase {
 
     @SuppressWarnings("unchecked")
     @After
-    protected void wipeAllODFEIndices() throws IOException {
+    protected void wipeAllODFEIndices() throws Exception {
         Response response = adminClient().performRequest(new Request("GET", "/_cat/indices?format=json&expand_wildcards=all"));
         MediaType xContentType = MediaType.fromMediaType(response.getEntity().getContentType());
         try (
@@ -280,7 +280,15 @@ public abstract class MLCommonsRestTestCase extends OpenSearchRestTestCase {
             for (Map<String, Object> index : parserList) {
                 String indexName = (String) index.get("index");
                 if (indexName != null && !".opendistro_security".equals(indexName)) {
-                    adminClient().performRequest(new Request("DELETE", "/" + indexName));
+                    deleteIndexWithAdminClient(indexName);
+                }
+            }
+            // Deletions process asynchronously and could impact future tests
+            // Ensure cluster state is up to date with the deletions before ending
+            for (Map<String, Object> index : parserList) {
+                String indexName = (String) index.get("index");
+                if (indexName != null && !".opendistro_security".equals(indexName) && indexExistsWithAdminClient(indexName)) {
+                    assertBusy(() -> assertFalse(indexExistsWithAdminClient(indexName)), 10, TimeUnit.SECONDS);
                 }
             }
         }


dbwiddis:
So it seems this change is fine on Windows and Linux tests but breaks pretty much breaks every Docker IT...

dbwiddis:
Closing this PR, as I'm essentially forcing a failure for the oft-seen condition that "tasks are still executing which may impact future tests".   The issue still remains but I'm not sure how or if to fix it or just defend against it when it matters.

PR #2922, opened 2024-09-10
populate time fields for connectors on return

### LINKED ISSUE #2890: [BUG] Created Time and Last Updated Time are not implemented on Connector
**What is the bug?**

The `.plugin-ml-connector` index contains field mappings for `created_time` and `last_updated_time` but these are never populated.

1. Create a connector.
2. Inspect the created document using the REST API
3. Observe there are no `created_time` or `last_updated_time` fields.

 - the created and last updated time are both populated with the same `Instant.now()` upon instantiation, or at least set before index request
 - the last updated time is updated before an update request

OpenSearch :)

```json
http://localhost:9200/.plugins-ml-connector/_doc/44ymvpEBPIzofVXwNZHj
```
```json
{
    "_id": "44ymvpEBPIzofVXwNZHj",
    "_version": 1,
    "_seq_no": 0,
    "_primary_term": 1,
    "found": true,
    "_source": {
        "name": "Cohere Chat Model",
        "version": "1",
        "description": "The connector to Cohere's public chat API",
        "protocol": "http",
        "parameters": {
            "model": "command"
        },
        "credential": {
            "cohere_key": "REDACTED"
        },
        "actions": [
            {
                "action_type": "PREDICT",
                "method": "POST",
                "url": "https://api.cohere.ai/v1/chat",
                "headers": {
                    "Authorization": "Bearer ${credential.cohere_key}",
                    "Request-Source": "unspecified:opensearch"
                },
                "request_body": "{ \"message\": \"${parameters.inputs}\", \"model\": \"${parameters.model}\" }"
            }
        ]
    }
}
```

The `HttpConnector` class has fields for `createdTime` and `lastUpdatedTime` which are properly parsed from incoming transport or XContent, and properly sent to outgoing transport or XContent unless null.   The problem is that they are always null.
 - there are no setters for them
 - they are never given any values

Suggested fix(es):
 - in the AbstractConnector, assign them a value on instantiating.  (Use the same value for both.)
 - Add a setter for the updated time
 - Set the updated time when using the UpdateConnector API.

brianf-aws:
### Description
Currently any class that extends the AbstractConnector class has the fields createdTime and lastUpdatedTime set to null. **_The solution was creating setters on the connector interface and implementing on the abstract connector. Now on Indexing new connectors (Via the create API) the time fields will be instantiated_. Additionally on update a connector, lastUpdateTime has been set.** Two unit tests in `UpdateConnectorTransportActionTests` named `UpdateConnectorTransportActionTests` and `testUpdateConnectorUpdatesHttpConnectorTimeFields`
### Related Issues
fixes #2890 **Created Time and Last Updated Time are not implemented on Connector**

### Check List

### Changes

> GET CONNECTOR (NO UPDATE)

| Connector              | createdTime | lastUpdateTime |
| :---------------- | :------: | ----: |
| Old (was indexed prior to code change)       |   ❌    | ❌ |
| New  (With code change)       |   ✅    | ✅ |

> GET CONNECTOR (AFTER UPDATE)

| Connector              | createdTime | lastUpdateTime |
| :---------------- | :------: | ----: |
| Old (was indexed prior to code change)       |   ❌    | ✅ |
| New  (With code change)       |   ✅    | ✅ |

1. Observe that connector creations using the [Get Connector API](https://opensearch.org/docs/latest/ml-commons-plugin/api/connector-apis/get-connector/) now have a `created_time,last_updated_time`
   a. creating a connector

```
curl -X POST \
  -H "Content-Type: application/json" \
  -d '{
  "name": "Chat Gpt 3.5",
  "description": "Brians Test Chatgpt",
  "version": "1",
  "protocol": "http",
  "parameters": {
    "endpoint": "api.openai.com",
    "model": "gpt-3.5-turbo"
  },
  "credential": {
    "openAI_key": "'"$BRIAN_OPENAI"'"
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "headers": {
      },
      "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": ${parameters.messages} }"
    }
  ]
}' \
  localhost:9200/_plugins/_ml/connectors/_create
```
     b. Inspecting changes via a get request  `curl -XGET "http://localhost:9200/_plugins/_ml/connectors/lkf-3JEBz83ApvP3YOPy"`. Also note the followinf fields have been added `created_time,last_updated_time`
```
 {
    "name": "Chat Gpt 3.5",
    "version": "1",
    "description": "Brians Test Chatgpt",
    "protocol": "http",
    "parameters": {
        "endpoint": "api.openai.com",
        "model": "gpt-3.5-turbo"
    },
    "actions": [
        {
            "action_type": "PREDICT",
            "method": "POST",
            "headers": {
            },
            "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": ${parameters.messages} }"
        }
    ],
    "created_time": 1725989544074,
    "last_updated_time": 1725989544074
}
```

2. Note how when updating the connector using the [Update Connector API](https://opensearch.org/docs/latest/ml-commons-plugin/api/connector-apis/update-connector/) and then calling get it changes the the last updated time and leaves the created time the same
    a. Update the connector `curl -XPUT "http://localhost:9200/_plugins/_ml/connectors/lkf-3JEBz83ApvP3YOPy" -H "Content-Type: application/json' -d'
{
  "description": "The new description of Brians chatGpt model"
}"` 

   b. get the connector again and note the new time `curl -XGET "http://localhost:9200/_plugins/_ml/connectors/lkf-3JEBz83ApvP3YOPy"` which yields

```

{
    "name": "Chat Gpt 3.5",
    "version": "1",
    "description": "The new description of Brians chatGpt model",
    "protocol": "http",
    "parameters": {
        "endpoint": "api.openai.com",
        "model": "gpt-3.5-turbo"
    },
    "actions": [
        {
            "action_type": "PREDICT",
            "method": "POST",
            "headers": {
            },
            "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": ${parameters.messages} }"
        }
    ],
    "created_time": 1725989544074,
    "last_updated_time": 1725989603171
}
```
3. In summary the initial connector creation has the fields `"created_time": 1725989544074, "last_updated_time": 1725989544074` and after the update its `"created_time": 1725989544074, "last_updated_time": 1725989603171`. You can see how it changed but the created_time stayed the same.

### Testing
- Manual testing via using docker-compose (**Was done in 2.x to get around version incompatibility and cherry picked to main branch**)
1. Spin up a docker compose file without the code changes (i.e. using the ml-commons code thats packed with openSearch)
2. Create a connector and see it has been created via the get connectors API, (Observe no time fields)
3. shut down the container run `./gradlew assemble` to pack the code change into a zip file
4. add the following commands to the docker compose file `    command: bash -c "bin/opensearch-plugin remove opensearch-skills; bin/opensearch-plugin remove opensearch-ml; bin/opensearch-plugin install --batch file:///<Path to the zip>/opensearch-ml-X.X.X.X-SNAPSHOT.zip; ./opensearch-docker-entrypoint.sh opensearch"`
5. Get the connector that was done on step 2 observe no observed time
6. Create a NEW connector and observe it has the time fields update it to see the updated time stamp works
- Two new unit tests were onboarded to simulate these changes in `UpdateConnectorTransportActionTests` named 
1. `testUpdateConnectorDoesNotUpdateHttpConnectorTimeFields` checks that when creating a connector time fields arent populated we test this because we want to make sure that connectors dont get this field populated instead we want to make sure that the Transport layer sets this additionally when these fields are null we need to make sure the transport layer does not set the updated time on a update.
2. `testUpdateConnectorUpdatesHttpConnectorTimeFields` Checks that time fields are instantiated and that after a update the updated time is bigger than the created time.

---


dhrubo-os:
Let's add more details in the description like how did you test?

Let's create a connector and invoke the get connector api to get the connector details to show the create and update timestamp

After that update the connector and then show the change of updated timestamp.

brianf-aws:
Hey @dhrubo-os Thanks for the feedback I updated my post based on your feedback. I uploaded a video that you can click on to see a visual of the written steps (2 mins) Its on the heading **Changes**

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -65,11 +65,18 @@ public abstract class AbstractConnector implements Connector {
...
     protected AccessMode access;
+    @Setter
     protected Instant createdTime;
+    @Setter
     protected Instant lastUpdateTime;
     @Setter
     protected ConnectorClientConfig connectorClientConfig;
 
+    public AbstractConnector() {
+        this.createdTime = Instant.now();
```
ylwu-amzn, line 76: Will this method be used when retrieve an old connector ? The old connector has no create time and last update time. If this method used when retrieve old connector, it will assign `Instant.now()` to these two time fields, that's not expected.

brianf-aws: This constructor (AbstractConnector) will be invoked on the creation of (HTTPConnector) https://github.com/opensearch-project/ml-commons/blob/93d0429db5f7d32305362e985d0f9ffa908d5275/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java#L45

and (AWSConnector) https://github.com/opensearch-project/ml-commons/blob/93d0429db5f7d32305362e985d0f9ffa908d5275/common/src/main/java/org/opensearch/ml/common/connector/AwsConnector.java#L30

These were suggestions from the initial post
https://github.com/opensearch-project/ml-commons/issues/2890#issue-2506354115

I might be misunderstanding your concern so please let me know

dbwiddis: @brianf-aws Assume this fix goes into 2.18.0.   If there was a connector created in 2.13.0, it will have null for both of those time fields.   The concern is what happens when that connector is parsed.

I believe the XContent parsing and Transport (Writeable) implementations will maintain the null, but this is something you should confirm.

common/src/test/java/org/opensearch/ml/common/transport/model/MLUpdateModelInputTest.java
```diff
@@ -152,7 +154,14 @@ public void readInputStreamSuccessWithNullFields() throws IOException {
     @Test
     public void testToXContent() throws Exception {
         String jsonStr = serializationWithToXContent(updateModelInput);
-        assertEquals(expectedOutputStrForUpdateRequestDoc, jsonStr);
+
+        JSONObject connectorJsonObject = new JSONObject(jsonStr);
+        long connectorCreatedTime = connectorJsonObject.getJSONObject("connector").getLong("created_time");
+        long connectorLastUpdatedTime = connectorJsonObject.getJSONObject("connector").getLong("last_updated_time");
```
dbwiddis, line 160: This is relevant to @ylwu-amzn 's comment about old connectors.  If their value is null, an NPE will happen here: `getLong()` will probably return a null `Long` and then assigning it to a primitive `long` will cause an NPE on unboxing.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -65,11 +65,18 @@ public abstract class AbstractConnector implements Connector {
...
+        this.lastUpdateTime = Instant.now();
```
dbwiddis, line 77: `Instant.now()` may change by a few nanoseconds between these two lines.
```suggestion
        this.lastUpdateTime = this.createdTime;
```

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -65,11 +65,18 @@ public abstract class AbstractConnector implements Connector {
     protected User owner;
     @Setter
     protected AccessMode access;
+    @Setter
```
dbwiddis, line 68: Since you're only allowing setting the updated time remotely and setting created time internally you can probably omit this setter.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
brianf-aws: Hey @ylwu-amzn and @dbwiddis My newest commit c606ffadb77cb044881d9c2d97709ca6d13bce0b addresses this issue now upon indexing new connectors they will receive a time stamp while older connectors will have the same shape. Ive tested locally and via unit tests in [plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java](https://github.com/opensearch-project/ml-commons/commit/8c006de118ae26db6ae1b52f3697a3aac6d43912#diff-2be9eb714846960c89f2cc23fd91699d39eee15cbfe030ef8e06b81ccb41032cL204)

common/src/test/java/org/opensearch/ml/common/transport/model/MLUpdateModelInputTest.java
brianf-aws: Good catch, given the new code changes were only new **indexed** connectors will get timestamps in this latest commit c606ffadb77cb044881d9c2d97709ca6d13bce0b this test has been reverted to what it was before as now we aren't changing the the connectors on instantiation the previous test so tests on this previous commit 0d2693180bbd76aab7048490ce9eb4dc5eeb3fe1 (i.e. Before my code changes) will be the same with the addition of two UTs that I said in a previous comment :)

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
brianf-aws: Good catch, In my newest commit when connectors get indexed we will place the time and we wont touch the constructor of Abstract connector here is the logic that goes on in `TransportCreateConnectorAction.indexConnector(...)`
 ```
          Instant currentTime = Instant.now();
          connector.setCreatedTime(currentTime);
          connector.setLastUpdateTime(currentTime);
```

File: common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java (hunk 4 above)
Comment (brianf-aws) [reply common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java:68]: As stated in a previous comment regarding the abstract connector we onboarded  `void setCreatedTime(Instant createdTime); void setLastUpdateTime(Instant lastUpdateTime);` so Abstract connector will need to implement these using lombok setters. In this way we can `set the time on indexing`

File: plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java [hunk 5]
```diff
@@ -93,6 +95,11 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
                 if (Boolean.TRUE.equals(hasPermission)) {
                     connector.update(mlUpdateConnectorAction.getUpdateContent(), mlEngine::encrypt);
                     connector.validateConnectorURL(trustedConnectorEndpointsRegex);
+
+                    if (connector instanceof HttpConnector && ((HttpConnector) connector).getLastUpdateTime() != null) {
```
rbhavna, line 99: Why only HttpConnector? Can't we just check `connector.getLastUpdateTime() != null`

brianf-aws: I didn't want to bloat the [connector interface](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/Connector.java) with a getter as it already has two getters. Also AbstractConnector has the time fields not the connector interface

dhrubo-os: what if `getLastUpdateTime` is null? Don't we want to update those values also?

dhrubo-os: what's the concern if we just add `connector.setLastUpdateTime(Instant.now());` without any checks?

dbwiddis: IMO the updates should be built-in (but overridable) defaults on the AbstractConnector.  May not need a setter for it as it should generally done automatically.

brianf-aws: Hi @dhrubo-os, @dbwiddis  

This method will update connectors when it does this my theory is that deserializes the connector from an index and constructs the HTTPConnector object when this happens it is very hard to control the update fields (on instantiation).

If we decide to add the time fields in the abstract Connector constructor we have a problem between old connectors and new ones. This results in a unexpected bug when connectors that didn't have these fields get brought to life; every subsequent update on the connector makes the time fields a new number every time.

Now if we wanted old connectors to not have the time fields (what our users expect when they update a old connector) they wont see the time fields because this method checks if an update should work on time only if they have the field. When a connector thats been created with the code change, they will see time fields and will update accordingly well.

Please let me know if you are still confused I would be happy to explain more in depth :)

brianf-aws:
Hi Everyone thanks for your patience on this bug Ive updated this code with a final change that was discussed offline. Here is the final changes in a tabulated form. I would appreciate your review once more 😅 @dbwiddis, @rbhavna, @dhrubo-os, @ylwu-amzn 


> GET CONNECTOR (NO UPDATE)

| Connector              | createdTime | lastUpdateTime |
| :---------------- | :------: | ----: |
| Old (was indexed prior to code change)       |   ❌    | ❌ |
| New  (With code change)       |   ✅    | ✅ |

> GET CONNECTOR (AFTER UPDATE)

| Connector              | createdTime | lastUpdateTime |
| :---------------- | :------: | ----: |
| Old (was indexed prior to code change)       |   ❌    | ✅ |
| New  (With code change)       |   ✅    | ✅ |

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
```diff
@@ -14,10 +14,8 @@
 
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.time.Instant;
+import java.util.*;
```
dbwiddis, line 18: It's generally frowned on to use star imports.

dbwiddis (approved):
I like this version. Simple, readable.

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
brianf-aws: oops, yeah will fix that

ylwu-amzn (approved):
Thanks for fixing this issue!

dhrubo-os (approved):

dhrubo-os:
@brianf-aws Backport PR for 2.x is failing. Can you please check why it's failing. And also please raise a backport PR for 2.x branch.

Issue #2930, opened 2024-09-11
[BUG] Batch ingestion API bugs


ylwu-amzn:
Test with OS2.17 RC4 
```
POST /_plugins/_ml/_batch_ingestion
{
  "index_name": "my-test1",
  "field_map": {
    "_id": "$.recordId",
    "embedding": "source[0].$.modelOutput.embedding"
  },
  "credential": {
    "region": "us-east-1",
    "access_key": "xxx",
    "secret_key": "xxx",
    "session_token": "xxx"
  },
  "data_source": {
    "type": "s3",
    "source": ["s3://ylwu-model-test-output/cszce2bsex07/my_batch2.jsonl.out"]
  }
}
```
sample data of my_batch2.jsonl.out
```
{"modelInput":{"inputText":"hello word 1"},"modelOutput":{"embedding":[-0.034975495,0.072906666],"inputTextTokenCount":5},"recordId":"CALL0000001"}
```
It returns task id `xHk64pEBG9EkCQDLzc-I`

But this task stays on `CREATED` forever. Checked log , error happens

Remove `source[0].` from `embedding` field map can work

```
[2024-09-11T17:58:07,399][ERROR][o.o.m.e.i.S3DataIngestion] [client1] Missing property in path $['source']
[2024-09-11T17:58:07,400][ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [client1] uncaught exception in thread [opensearch[client1][opensearch_ml_train][T#2]]
org.opensearch.OpenSearchStatusException: Failed to batch ingest: Missing property in path $['source']
	at org.opensearch.ml.engine.ingest.S3DataIngestion.ingestSingleSource(S3DataIngestion.java:148) ~[?:?]
	at org.opensearch.ml.engine.ingest.S3DataIngestion.ingest(S3DataIngestion.java:66) ~[?:?]
	at org.opensearch.ml.action.batch.TransportBatchIngestionAction.lambda$doExecute$0(TransportBatchIngestionAction.java:96) ~[?:?]
	at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:946) ~[opensearch-2.17.0.jar:2.17.0]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
[...]

PR #2933, opened 2024-09-12
fix field mapping, add more error handling and remove checking jobId …

Zhangxunmt:
…filed in batch job response

### Description
This PR fix all problems in https://github.com/opensearch-project/ml-commons/issues/2930

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/ingest/AbstractIngestion.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/ingest/AbstractIngestion.java
@@ -77,6 +77,55 @@ public class AbstractIngestion implements Ingestable {
 
     /**
      * Filters fields in the map where the value contains the specified source index as a prefix.
+     * When there is only one source file, users can skip the source[] prefix
+     *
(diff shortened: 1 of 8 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/ingest/AbstractIngestion.java
```diff
@@ -75,6 +75,55 @@ protected double calculateSuccessRate(List<Double> successRates) {
             );
     }
 
+    /**
+     * Filters fields in the map where the value contains the specified source index as a prefix.
+     * When there is only one source file, users can skip the source[] prefix
```
ylwu-amzn, line 80: If there is only one source file, is it ok for cx to input `source[0].xxx` now ?

Zhangxunmt: Yes it works for both cases. this function makes sure it finds the fields with source[0] or any fields that does not prefixed with "source".

ylwu-amzn (approved):

PR #2934, opened 2024-09-12
[Feature/multi_tenancy] Fix SdkClient credential handling for AWS endpoints

dbwiddis:
### Description

Refactors SdkClient code as required to support (working) DDB/AOSS integration tests
 - removes the temporary "local" DDB implementation
 - fetches credentials from system environment (and updates gradle build to put them there)

### Check List
- [x] New functionality includes testing. (Tested by #2818)

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
+++ b/plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
@@ -8,7 +8,6 @@
  */
 package org.opensearch.ml.sdkclient;
 
-import java.net.URI;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Map;
@@ -50,15 +49,12 @@ import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 
 import lombok.extern.log4j.Log4j2;
-import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProviderChain;
 import software.amazon.awssdk.auth.credentials.ContainerCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.InstanceProfileCredentialsProvider;
-import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
-import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
 import software.amazon.awssdk.http.apache.ApacheHttpClient;
-import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 
@@ -92,16 +88,20 @@ public class SdkClientFactory {
                     throw new OpenSearchException("AWS Opensearch Service client requires a metadata endpoint and region.");
                 }
                 log.info("Using remote AWS Opensearch Service cluster as metadata store");
-                return new SdkClient(new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region)));
+                return new SdkClient(
+                    new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region, "es"))
+                );
             case SdkClientSettings.AWS_DYNAMO_DB:
                 if (Strings.isBlank(remoteMetadataEndpoint) || Strings.isBlank(region)) {
-                    throw new OpenSearchException("AWS Opensearch Service client requires a metadata endpoint and region.");
+                    throw new OpenSearchException(
+                        "AWS DynamoDB and Opensearch Serverless Service client requires a metadata endpoint and region."
+                    );
                 }
                 log.info("Using dynamo DB as metadata store");
                 return new SdkClient(
                     new DDBOpenSearchClient(
                         createDynamoDbClient(region),
-                        new RemoteClusterIndicesClient(createOpenSearchClient(remoteMetadataEndpoint))
+                        new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region, "aoss"))
                     )
                 );
             default:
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
```diff
@@ -92,16 +88,20 @@ public static SdkClient createSdkClient(Client client, NamedXContentRegistry xCo
...
                 log.info("Using remote AWS Opensearch Service cluster as metadata store");
-                return new SdkClient(new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region)));
+                return new SdkClient(
+                    new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region, "es"))
+                );
             case SdkClientSettings.AWS_DYNAMO_DB:
                 if (Strings.isBlank(remoteMetadataEndpoint) || Strings.isBlank(region)) {
-                    throw new OpenSearchException("AWS Opensearch Service client requires a metadata endpoint and region.");
+                    throw new OpenSearchException(
+                        "AWS DynamoDB and Opensearch Serverless Service client requires a metadata endpoint and region."
```
dhrubo-os, line 97: Do we need to mention about `Serverless`?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
```diff
@@ -92,16 +88,20 @@ public static SdkClient createSdkClient(Client client, NamedXContentRegistry xCo
...
+                    );
                 }
                 log.info("Using dynamo DB as metadata store");
                 return new SdkClient(
                     new DDBOpenSearchClient(
                         createDynamoDbClient(region),
-                        new RemoteClusterIndicesClient(createOpenSearchClient(remoteMetadataEndpoint))
+                        new RemoteClusterIndicesClient(createAwsOpenSearchServiceClient(remoteMetadataEndpoint, region, "aoss"))
```
dhrubo-os, line 104: `aoss` --> Shouldn't we get that from some configuration over than hard coded?

plugin/src/main/plugin-metadata/plugin-security.policy
```diff
@@ -24,4 +24,9 @@ grant {
 
     // Circuit Breaker
     permission java.lang.RuntimePermission "getFileSystemAttributes";
+    
+    // AWS credentials needed for clients
+    permission java.io.FilePermission "<<ALL FILES>>", "read";
+    
+    permission java.security.AllPermission;
```
dhrubo-os, line 31: This essentially bypasses all security checks in the Java Security Manager and gives the code unrestricted access to the system. Why do we want that?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
dbwiddis: "Need"? Probably not. But the "aoss" string in the code on line 104 means exactly that.

Should we make the signing service a configurable setting and vary the message based on that?

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
dbwiddis: Heh, I just replied to the other comment with this suggestion. Will do that and update PR.

plugin/src/main/plugin-metadata/plugin-security.policy
dbwiddis: one ring to rule them all

But seriously, I think that was leftover from some debugging and I need to remove it.

plugin/src/main/java/org/opensearch/ml/sdkclient/SdkClientFactory.java
```diff
@@ -110,6 +105,15 @@ public static SdkClient createSdkClient(Client client, NamedXContentRegistry xCo
         }
     }
 
+    private static void validateAwsParams(String clientType, String remoteMetadataEndpoint, String region, String serviceName) {
+        if (Strings.isBlank(remoteMetadataEndpoint) || Strings.isBlank(region)) {
+            throw new OpenSearchException(clientType + " client requires a metadata endpoint and region.");
+        }
+        if (!"es".equals(serviceName) && !"aoss".equals(serviceName)) {
```
dhrubo-os, line 112: What if an opensource community wants to use some other service other than `es` or `aoss`? I think this part needs to be generic? Isn't that the whole point we moved this service name to a settings `REMOTE_METADATA_SERVICE_NAME_KEY`?

dbwiddis: These are specific to `case SdkClientSettings.AWS_OPENSEARCH_SERVICE:` and `case SdkClientSettings.AWS_DYNAMO_DB` for the `remoteMetadata` type.  Any other service or any other provider can be added to that configuration.

These two particular clients use the AWS SDK for connection and "es" and "aoss" are the only documented options right now, see https://opensearch.org/docs/latest/clients/java/#connecting-to-amazon-opensearch-service

dbwiddis: Also relevant: https://github.com/opensearch-project/data-prepper/blob/627ae9ffdb2de5dda84566e7b6a4f130f5005210/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/OpenSearchClientFactory.java#L54-L55

dbwiddis: > What if an opensource community wants to use some other service other than `es` or `aoss`? I think this part needs to be generic? Isn't that the whole point we moved this service name to a settings `REMOTE_METADATA_SERVICE_NAME_KEY`?

For a concrete example, let's say OCI wants to have a client connect to their service through the [OCI OpenSearch SDK](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/javasdk.htm).   They would:
1. Define a new `SdkClientSettings.REMOTE_METADATA_TYPE`.  We already have "RemoteOpenSearch" (a cluster anywhere) and "AWSOpenSearchService" (AOS).  They'd add a new "OCISearchService" (or whatever name)
2. Instead of calling validation via `validateAWSParams()` which requires an endpoint, region, and one of two service names, they'd create a new `validateOCIParams()` checker method for whichever settings they needed to support methods on https://github.com/oracle/oci-java-sdk

dhrubo-os: Thanks for detailed explanation. Makes sense, sounds good to me.

dhrubo-os (approved):

PR #2941, opened 2024-09-12
[Feature/multi_tenancy] Implement retryOnConflict on UpdateDataObject

dbwiddis:
### Description

Implements `UpdateRequest.retryOnConflict()`
 - For local and remote OpenSearch client, simply passes through the setting
 - For DDB Client, implements the retry logic

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
+++ b/plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
@@ -233,29 +233,13 @@ public class DDBOpenSearchClient implements SdkClientDelegate {
                 String source = Strings.toString(MediaTypeRegistry.JSON, request.dataObject());
                 JsonNode jsonNode = OBJECT_MAPPER.readTree(source);
                 Map<String, AttributeValue> updateItem = JsonTransformer.convertJsonObjectToDDBAttributeMap(jsonNode);
-                updateItem.remove(HASH_KEY);
+                updateItem.remove(TENANT_ID);
                 updateItem.remove(RANGE_KEY);
                 Map<String, AttributeValue> updateKey = new HashMap<>();
                 updateKey.put(HASH_KEY, AttributeValue.builder().s(tenantId).build());
                 updateKey.put(RANGE_KEY, AttributeValue.builder().s(request.id()).build());
-                UpdateItemRequest.Builder updateItemRequestBuilder = UpdateItemRequest.builder().tableName(request.index()).key(updateKey);
-                Map<String, String> expressionAttributeNames = new HashMap<>();
-                expressionAttributeNames.put("#seqNo", SEQ_NO_KEY);
-                expressionAttributeNames.put("#source", SOURCE);
-                Map<String, AttributeValue> expressionAttributeValues = new HashMap<>();
-                expressionAttributeValues.put(":incr", AttributeValue.builder().n("1").build());
-                expressionAttributeValues.put(":source", AttributeValue.builder().m(updateItem).build());
-                updateItemRequestBuilder.updateExpression("SET #seqNo = #seqNo + :incr, #source = :source ");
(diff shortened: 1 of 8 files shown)


naradha (commented):
LGTM! 

We can have a default retry of 3 instead of 1.

naradha (approved):

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -288,6 +253,60 @@ public CompletionStage<UpdateDataObjectResponse> updateDataObjectAsync(
...
+                    .expressionAttributeNames(expressionAttributeNames)
+                    .expressionAttributeValues(expressionAttributeValues);
+                UpdateItemRequest updateItemRequest = updateItemRequestBuilder.build();
+                UpdateItemResponse updateItemResponse = dynamoDbClient.updateItem(updateItemRequest);
+                if (updateItemResponse != null
+                    && updateItemResponse.attributes() != null
+                    && updateItemResponse.attributes().containsKey(SEQ_NO_KEY)) {
+                    return Long.parseLong(updateItemResponse.attributes().get(SEQ_NO_KEY).n());
+                }
+            } catch (ConditionalCheckFailedException ccfe) {
```
dhrubo-os, line 298: what happens if we face other exception here?

dbwiddis: Any other exception will bubble up to line 244: an IOException will be caught and rethrown as a RuntimeException, any other exception _is_ a RuntimeException and will be passed along just like it would have on a non-retried update call.

> should we still try to update the document if the retriesRemaining > 0

No, the [OpenSearch Implementation](https://github.com/opensearch-project/OpenSearch/blob/c85ee6898b4e3e2f4a2037e42337a7eb8033de9e/server/src/main/java/org/opensearch/action/update/TransportUpdateAction.java#L355-L378) only retries on a `VersionConflictEngineException`, which the `ConditionalCheckFailedException` is the corresponding exception for DDB.

dhrubo-os: Can we have a test where we can see for `ConditionalCheckFailedException` case, retried 1+ times and for other cases it just tried 1 time?

dbwiddis: > Can we have a test where we can see for `ConditionalCheckFailedException` case, retried 1+ times and for other cases it just tried 1 time?

I'm not clear which case you want tested?  We currently test:
 - retries 0, success and exception cases
 - retries 1, failure and then success on retry
 - retries 1, failure and then another failure on retry

Although looking into it, I see a small problem... some tests are annotated with `@Test` and one begins with test.  MLCommons is inconsistent between which method is used in which module and it might be that one of the tests isn't running.

dbwiddis: @dhrubo-os Well it was running all the tests regardless of annotation or method naming.  So no issues there but I added both annotation and changed method names for consistency anyway.

Added a test of a generic RTE.   So the following are now tested:
 - retries 0, success
 - retries 0, conditional check exception
 - retries 0, runtime exception (just added this one)
 - retries 1, with success (happy case)
 - retires 1, with conflict, success, conflict (passes on first retry)
 - retries 1, with conflict, conflict, success (fails on first retry, never gets to success)

Any other cases you need?

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
dhrubo-os: Thanks for adding test. Approving.

PR #2959, opened 2024-09-16
[Feature/multi_tenancy] Add more missing tenant ids in requests

dbwiddis:
### Description

Adds tenant ID to a few requests that were missing them.

~NOTE: included a TODO temporarily working around model controller accidentally migrated to sdkClient in #2585. I'll follow up with another PR to remove it.~

### Check List
- [x] New functionality includes testing. (Tested by #2818)

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
@@ -373,9 +373,7 @@ public class DeleteModelTransportAction extends HandledTransportAction<ActionReq
             // for remote model we don't need to delete model chunks so reducing one latch countdown.
             countDownLatch.countDown();
         }
-        // TODO This uses remote client but we haven't implemented it anywhere else?!
-        // deleteController(modelId, isHidden, countDownActionListener);
-        countDownActionListener.onResponse(true);
+        deleteController(modelId, isHidden, countDownActionListener);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -355,7 +373,9 @@ private void deleteModelChunksAndController(
             // for remote model we don't need to delete model chunks so reducing one latch countdown.
             countDownLatch.countDown();
         }
-        deleteController(modelId, isHidden, countDownActionListener);
+        // TODO This uses remote client but we haven't implemented it anywhere else?!
+        // deleteController(modelId, isHidden, countDownActionListener);
```
dhrubo-os, line 377: I assume you will reverse the code for controller deletion and then enable this, right?

dbwiddis: > I assume you will reverse the code for controller deletion and then enable this, right?

Yep, as mentioned in the original comment.

I'll put it back to its pre #2585 state.  If there are any workarounds still needed, I understand a permanent fix is coming with feature-gating in #2652

dbwiddis: Except that commenting it out is causing tests to fail.  Let me pull it back out of this PR and address separately.

dhrubo-os (approved):

Issue #2918, opened 2024-09-09
[BUG] Conversational agent overwrites input from SearchIndexTool


reuschling:
It is straightforward and works of course when I define a SearchIndexTool inside a conversational flow agent, the question is inserted into the defined OpenSearch query, there will come search results.

Unfortunately, I can not manage this with a conversational agent.

This agent should always execute the same query:

```json
{
	"name": "Test Agent",
	"type": "conversational",
	"description": "Simple agent to test the agent framework",
	"llm": {
		"model_id": "{{ _.LlmModelId4agent }}",
		"parameters": {
			"max_iteration": 5,
			"stop_when_no_tool_found": true,
			"disable_trace": false
		}
	},
	"memory": {
		"type": "conversation_index"
	},
	"app_type": "chat_with_rag",
	"tools": [
		{
			"type": "SearchIndexTool",
			"description": "A tool to search opensearch index with natural language question. If you don't know answer for some question, you should always try to search data with this tool. Action Input: <natural language question>",
			"include_output_in_agent_response": true,
			"parameters": {
				"input": "{\"index\": \"scll_agendaitem\", \"query\": {\"query\": { \"match_all\": {}}} }"
			}
		}
	]
}
```

I get always this error message: `Failed to run the tool SearchIndexTool with the error message Expected a com.google.gson.JsonObject but was com.google.gson.JsonPrimitive; at path $.` Independent from the specified tool input. It seems that the conversational agent overwrites the `parameters.input` content with its generated, redefined query, wasting the query specification from the SearchIndexTool.

Expected behavior is that you define the OpenSearch query inside the tool description, and that the output from the former step of the conversational search should be inserted into this query, e.g. like this:

```json
{
	"type": "SearchIndexTool",
	"description": "A tool to search opensearch index with natural language question. If you don't know answer for some question, you should always try to search data with this tool. Action Input: <natural language question>",
	"parameters": {
		"input": "{\"index\": \"${parameters.index}\", \"query\": ${parameters.query} }",
		"query": {
			"query": {
				"match": {
					"tns_body_chunked768": {
						"query": "${parameters.FORMER_OUTPUT_OF_CONVERSATIONAL_AGENT}"
					}
				}
			},
			"size": "5",
			"_source": "tns_body_chunked768"
		}
	}
}
```

Maybe the 'input' parameter for the query specification can be renamed or configured.

yuye-aws:
[...]

PR #2977, opened 2024-09-23
add config field in MLToolSpec for static parameters

jngz-es:
### Description
Add config field in MLToolSpec to support static parameters in tool execution.

An example from the issue 2918.
```
{
	"name": "Test Agent",
	"type": "conversational",
	"description": "test agent",
	"llm": {
		"model_id": "<model id>",
		"parameters": {
      "max_iteration": 5,
      "stop_when_no_tool_found": true,
      "response_filter": "$.completion"
    }
	},
	"memory": {
		"type": "conversation_index"
	},
	"app_type": "chat",
	"tools": [
		{
			"type": "SearchIndexTool",
			"description": "A tool to search opensearch index with natural language question.",
			"include_output_in_agent_response": true,
			"config": {
				"input": "{\"index\": \"sample-index\", \"query\": {\"query\": { \"match_all\": {}}} }"
			}
		}
	]
}
```

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/2836
https://github.com/opensearch-project/ml-commons/issues/2918

### Check List


dbwiddis (commented):
Can you open an issue on Flow Framework repo amplifying what changed here? We will need to update the [`ToolStep`](https://github.com/opensearch-project/flow-framework/blob/43a9f871de50fc57758efd57c0f8899b22cecc3a/src/main/java/org/opensearch/flowframework/workflow/ToolStep.java#L51-L94) to account for a change in API.  Better yet, open a PR with the changes!

Also please make sure you update the [API specification](https://github.com/opensearch-project/opensearch-api-specification/blob/c6c0c417b19dd9b63c4e37379b7e94d412be7ea8/spec/schemas/ml._common.yaml#L271-L280) which Flow Framework will be using to validate these API parameters.   (It's a checklist item at the top of this PR.)

CC: @owaiskazi19 @amitgalitz @junweid62 @ylwu-amzn

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -69,6 +82,12 @@ public void writeTo(StreamOutput out) throws IOException {
             out.writeBoolean(false);
         }
         out.writeBoolean(includeOutputInAgentResponse);
+        if (configMap != null) {
```
ylwu-amzn, line 85: Add version checking for BWC

jngz-es:
The flow framework issue, https://github.com/opensearch-project/flow-framework/issues/878.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: Good catch, done.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -24,20 +26,31 @@
 @EqualsAndHashCode
 @Getter
 public class MLToolSpec implements ToXContentObject {
+    public static final Version MINIMAL_SUPPORTED_VERSION_FOR_TOOL_CONFIG = CommonValue.VERSION_2_17_0;
```
dbwiddis, line 29: Isn't 2.17.0 already released so this would be a 2.18.0 feature?

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -56,6 +70,9 @@ public MLToolSpec(StreamInput input) throws IOException {
             parameters = input.readMap(StreamInput::readString, StreamInput::readOptionalString);
         }
         includeOutputInAgentResponse = input.readBoolean();
+        if (input.getVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_TOOL_CONFIG) && input.readBoolean()) {
+            configMap = input.readMap(StreamInput::readOptionalString, StreamInput::readOptionalString);
```
dbwiddis, line 74: This seems odd here, it looks like we're allowing null keys and null values in the map which I doubt is the case (at least the keys shouldn't be optional, and I'm not sure of the benefit of null values vs. just omitting the key).  Same comment in the `writeTo()` below.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -87,6 +112,9 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field(PARAMETERS_FIELD, parameters);
         }
         builder.field(INCLUDE_OUTPUT_IN_AGENT_RESPONSE, includeOutputInAgentResponse);
+        if (configMap != null && configMap.size() > 0) {
```
dbwiddis, line 115:
```suggestion
        if (configMap != null && !configMap.isEmpty()) {
```

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -119,6 +148,9 @@ public static MLToolSpec parse(XContentParser parser) throws IOException {
                 case INCLUDE_OUTPUT_IN_AGENT_RESPONSE:
                     includeOutputInAgentResponse = parser.booleanValue();
                     break;
+                case CONFIG_FIELD:
+                    configMap = getParameterMap(parser.map());
```
dbwiddis, line 152: may want to rename `getParameterMap` to be more general `getStringToStringMap` or similar.

common/src/test/java/org/opensearch/ml/common/agent/MLAgentTest.java
```diff
@@ -46,7 +46,7 @@ public void constructor_NullName() {
             MLAgentType.CONVERSATIONAL.name(),
             "test",
             new LLMSpec("test_model", Map.of("test_key", "test_value")),
-            List.of(new MLToolSpec("test", "test", "test", Collections.EMPTY_MAP, false)),
+            List.of(new MLToolSpec("test", "test", "test", Collections.EMPTY_MAP, false, Collections.EMPTY_MAP)),
```
dbwiddis, line 49:
```suggestion
            List.of(new MLToolSpec("test", "test", "test", Collections.EMPTY_MAP, false, Collections.emptyMap())),
```
Use `emptyMap()` rather than `EMPTY_MAP` to get built in type validation.  This will probably create unchecked cast warnings.  (Same comment multiple times below.)

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
```diff
@@ -32,11 +32,70 @@ public void writeTo() throws IOException {
...
     public void toXContent() throws IOException {
-        MLToolSpec spec = new MLToolSpec("test", "test", "test", Map.of("test", "test"), false);
+        MLToolSpec spec = new MLToolSpec("test", "test", "test", Map.of("test", "test"), false, Map.of("configKey", "configValue"));
+        XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
+        spec.toXContent(builder, ToXContent.EMPTY_PARAMS);
+        String content = TestHelper.xContentBuilderToString(builder);
+
+        Assert
+            .assertEquals(
+                "{\"type\":\"test\",\"name\":\"test\",\"description\":\"test\",\"parameters\":{\"test\":\"test\"},\"include_output_in_agent_response\":false,\"config\":{\"configKey\":\"configValue\"}}",
```
dbwiddis, line 77: I don't think the order of fields in JSON is guaranteed unless you specify a sorting order.  Same comment in other assertions.

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
```diff
@@ -50,6 +109,28 @@ public void toXContent() throws IOException {
...
+                new NamedXContentRegistry(new SearchModule(Settings.EMPTY, Collections.emptyList()).getNamedXContents()),
+                null,
+                jsonStr
+            );
+        parser.nextToken();
+        MLToolSpec spec = MLToolSpec.parse(parser);
+
+        Assert.assertEquals(spec.getType(), "test");
+        Assert.assertEquals(spec.getName(), "test");
+        Assert.assertEquals(spec.getDescription(), "test");
```
dbwiddis, line 126: Nit: using the same string everywhere can let copy-paste-and-forget-to-change errors slip through.  Recommend using a different string for each field.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: For the key, changed from `readOptionalString/writeOptionalString` to `readString/writeString`.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: Done.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: It's an existing function, I prefer to keep it unchanged at this moment.

common/src/test/java/org/opensearch/ml/common/agent/MLAgentTest.java
jngz-es: Makes sense, changed.

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
jngz-es: Do you have any suggestions?

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: We want to catch up with 2.17 patch.

dbwiddis: 2.17.0 open source is already released.  If you want to change the version on the back port to that branch that gets synced elsewhere that’s a reasonable option but for anyone consuming the maven central artifact BWC would break with that version check.

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
dbwiddis: I usually do assertions that the string contains each key value pair.  Or just extract to a map with a util function and check keys.   I think there are some optional arguments with the Xcontent that let you specify an ordering.

dbwiddis: Looked into this.  Seems internal comments indicate the "XContent" preserves the same ordering, so since you're just parsing the same thing you built, this should be fine in this test case.  The concern would be if you're using any sort of external parsing using a map, where the ordering isn't guaranteed.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
dbwiddis: Actually looks like a 2.17.1 release is happening so it should be changed to that.

dbwiddis: Except they don't want any new features, just critical bugfixes. So nevermind :|

yuye-aws:
@reuschling Can you help review this PR?

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
yuye-aws: This PR is not a security fix. We can leave it to 2.18.

yuye-aws:
@jngz-es Can you please provide an example on how to use tool `config`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -470,12 +470,18 @@ private static void runTool(
...
+                    }
                     llmToolTmpParameters.put(MLAgentExecutor.QUESTION, actionInput);
                     tools.get(action).run(llmToolTmpParameters, toolListener); // run tool
                 } else {
                     Map<String, String> parameters = new HashMap<>();
                     parameters.putAll(tmpParameters);
                     parameters.putAll(toolParams);
+                    if (toolSpecMap.get(action).getConfigMap() != null) {
+                        parameters.putAll(toolSpecMap.get(action).getConfigMap());
+                    }
```
yuye-aws, line 484: I would suggest extract` parameters.putAll(toolSpecMap.get(action).getConfigMap());` outside of the if-else statement.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
```diff
@@ -434,6 +434,12 @@ Map<String, String> getToolExecuteParams(MLToolSpec toolSpec, Map<String, String
             input = substitutor.replace(input);
             executeParams.put("input", input);
         }
+
+        // Override all parameters in tool config to tool execution parameters as the config contains the static parameters.
+        if (toolSpec.getConfigMap() != null && !toolSpec.getConfigMap().isEmpty()) {
+            executeParams.putAll(toolSpec.getConfigMap());
+        }
+
```
yuye-aws, line 442: The function is the same as MLFlowAgentRunner. Please extract the function to classes like `AgentUtils`.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunnerTest.java
```diff
@@ -300,6 +300,26 @@ public void testGetToolExecuteParams() {
...
+        Map<String, String> params = Map
+            .of("toolType.param2", "value2", "toolName.param3", "value3", "param4", "value4", "toolName.tool_key", "dynamic value");
+
+        Map<String, String> result = mlFlowAgentRunner.getToolExecuteParams(toolSpec, params);
+
+        assertEquals("value1", result.get("param1"));
+        assertEquals("value3", result.get("param3"));
+        assertEquals("value4", result.get("param4"));
+        assertFalse(result.containsKey("toolType.param2"));
+        assertEquals("tool_config_value", result.get("tool_key"));
```
yuye-aws, line 320: Please also verify that non-config parameters can override

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
reuschling: As far as I see, this PR writes all parameters from the new config map into the parameter map that is relevant for execution, which means that config parameters will overwrite existing ones with same name, i.e. have prioritization. Thus, specifying an own static input parameter as suggested in #2836 and #2918 is possible, but crucial is also the possibility to specify placeholders inside the new config. I can not see some substitutions here but maybe this takes place somewhere else, for all parameters?

For #2836, the placeholder `"${parameters.MLModelTool.output}"` was asked for.
For #2918, SearchIndexTool inside conversational agent, there is the need of a placeholder for the **last action input** of the agent, that must inserted into the OpenSearch query defined statically inside the new config. If conversational agent just sets the input parameter for the next tool, I doubt this information is lost if I specify a static input inside the tool config. 

Last but not least, the documentation should be adjusted accordingly, e.g. [SearchIndexTool](https://opensearch.org/docs/latest/ml-commons-plugin/agents-tools/tools/search-index-tool/) and [tutorial example](https://opensearch.org/docs/latest/ml-commons-plugin/tutorials/rag-conversational-agent/#exposing-only-the-question-parameter). Also some list or hint to the common available placeholders would be nice, in order to know e.g. how the placeholder for the former conversational agent step action input looks like.

jngz-es:
> @jngz-es Can you please provide an example on how to use tool `config`?

Provided an example in the description section.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
jngz-es: Don't get it. I need to add the config parameters override immediately before tool run.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
jngz-es: Based on the discussion with Yuye, I will make the version 2.18. Thanks @yuye-aws

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
jngz-es: Good callout, I am thinking of supporting placeholders inside the config. Thanks @reuschling

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
jngz-es: done

common/src/test/java/org/opensearch/ml/common/agent/MLToolSpecTest.java
```diff
@@ -22,7 +22,7 @@ public class MLToolSpecTest {
 
     @Test
     public void writeTo() throws IOException {
-        MLToolSpec spec = new MLToolSpec("test", "test", "test", Map.of("test", "test"), false);
+        MLToolSpec spec = new MLToolSpec("test", "test", "test", Map.of("test", "test"), false, Map.of("configKey", "configValue"));
```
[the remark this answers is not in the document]
jngz-es: done

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -24,20 +26,31 @@
...
 
     @Builder(toBuilder = true)
-    public MLToolSpec(String type, String name, String description, Map<String, String> parameters, boolean includeOutputInAgentResponse) {
+    public MLToolSpec(
+        String type,
+        String name,
+        String description,
+        Map<String, String> parameters,
+        boolean includeOutputInAgentResponse,
+        Map<String, String> configMap
```
Zhangxunmt, line 52: Is it valid that all the configs are Strings? Is it better to use Map<String, Object> here?

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
dhrubo-os: +1 on this

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
```diff
@@ -56,6 +70,9 @@ public MLToolSpec(StreamInput input) throws IOException {
             parameters = input.readMap(StreamInput::readString, StreamInput::readOptionalString);
         }
         includeOutputInAgentResponse = input.readBoolean();
+        if (input.getVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_TOOL_CONFIG) && input.readBoolean()) {
+            configMap = input.readMap(StreamInput::readString, StreamInput::readOptionalString);
```
dhrubo-os, line 74: why `readOptionalString` ? so a key might not have a value?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -465,12 +465,21 @@ public static Map<String, String> constructToolParams(
...
         if (toolSpecParams != null) {
             toolParams.putAll(toolSpecParams);
         }
+        if (toolSpecConfigMap != null) {
+            toolParams.putAll(toolSpecConfigMap);
+        }
         if (tools.get(action).useOriginalInput()) {
             toolParams.put("input", question);
             lastActionInput.set(question);
+        } else if (toolSpecConfigMap != null && toolSpecConfigMap.containsKey("input")) {
```
dhrubo-os, line 478: Don't we have any test class for this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java
```diff
@@ -287,13 +287,18 @@ Map<String, String> getToolExecuteParams(MLToolSpec toolSpec, Map<String, String
                 executeParams.put(key, params.get(key));
             }
         }
+        // Override all parameters in tool config to tool execution parameters as the config contains the static parameters.
+        if (toolSpec.getConfigMap() != null && !toolSpec.getConfigMap().isEmpty()) {
```
dhrubo-os, line 291: rather than putting same piece of code, can we try to use a common method `getToolExecuteParams` for both MLFlowAgentRunner & MLConversationalFlowAgentRunner?

yuye-aws:
Pinging @reuschling to review the PR.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
dbwiddis: -1 on that.  The parameters map is already a `Map<String, String>`.  In the initial PR comment an example is given of a much more complex "object" but escaped as a String.

Trying to parse anything else from Flow Framework will get a lot more complex without typed objects to construct from the JSON.

common/src/main/java/org/opensearch/ml/common/agent/MLToolSpec.java
dbwiddis: Yes, this allows you to override a value with a non-value.

dbwiddis:
> Pinging @reuschling to review the PR.

Thanks @yuye-aws for prompting review on this. Thanks @Zhangxunmt for approval. Thanks @dhrubo-os for your initial review.

Code freeze / 2.18 RC generation is next Tuesday Oct 22.  Downstream plugin PRs (https://github.com/opensearch-project/flow-framework/pull/899) are dependent on this one being merged and need to merge before code freeze.  

Please do not wait until the last minute to get this merged; last-minute PR merges have occurred in at least 2 previous releases and this one's been open for a few weeks.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
```diff
@@ -428,12 +428,18 @@ Map<String, String> getToolExecuteParams(MLToolSpec toolSpec, Map<String, String
...
         if (executeParams.containsKey("input")) {
             String input = executeParams.get("input");
             StringSubstitutor substitutor = new StringSubstitutor(executeParams, "${parameters.", "}");
             input = substitutor.replace(input);
             executeParams.put("input", input);
         }
+
```
reuschling, line 442: I'm not sure if we have a chicken-and-egg problem here regarding the 'input' parameter. In #2918 'SearchIndexTool inside conversational agent', there is the need of a placeholder for the last output (i.e. action input) of the former agent step. This output/action input must be inserted into the OpenSearch query defined statically inside the new config, by placeholder.

There exists two possibilities for this:
1. There is an existing placeholder reference to the last output/action input that can be used, i.e. a dedicated parameter list entry. I'm not aware of such an entry, but if such a placeholder exists, the current solution would fix this scenario with a static config of e.g. ["input='"query": {"match": {"body": {"query": "${someUnknownPlaceholderToActionInput}"}}'"].
2. There is no such a placeholder, the last output/action input is just the 'input' entry in the current 'params' method parameter - so the placeholder inside the static config would be 'input'. This value inside 'params' is currently overwritten with 'input' from the static config, **before** substitution. So it is lost, right?

The final behavior should be as follows:
* method params: "input='What is Frodos age?'"  (comes from former agent step)
* static config params "input='"query": {"match": {"body": {"query": "${parameters.input}"}}'"

=> final params for tool execution should be then
"input='"query": {"match": {"body": {"query": "What is Frodos age?"}}'"

This could be achieved as follows:
```java
       // remember former input
       String formerInput = executeParams.get("input");

        if (toolSpec.getConfigMap() != null && !toolSpec.getConfigMap().isEmpty()) {
            executeParams.putAll(toolSpec.getConfigMap());
        }

        if (executeParams.containsKey("input")) {
            String input = executeParams.get("input");

            // set former input as substitution value
           if(formerInput != null)
               executeParams.put("input", formerInput); // add it as "formerInput" or "actionInput" to introduce a new placeholder

            StringSubstitutor substitutor = new StringSubstitutor(executeParams, "${parameters.", "}");
            input = substitutor.replace(input);
            executeParams.put("input", input);
```

The lack is that you can not reference to the input specified in the new static config anymore, which is maybe wanted behavior. To overcome this, we could introduce a new placeholder for the former input (aka the placeholder I mentioned in 1.). This would be a new parameter list entry, e.g. 'formerInput' (just add the formerInput with `executeParams.put("formerInput", formerInput);`).

The static config could be defined now as ["input='"query": {"match": {"body": {"query": "${parameters.formerInput}"}}'"].

Another name for 'formerInput' could be 'actionInput', maybe this is more intuitive.


The same is true for `MLFlowAgentRunner` of course, and for `AgentUtils`, whereby in AgentUtils the former input is defined in the method parameter 'actionInput'. So the only change here would be `executeParams.put("formerInput", actionInput);` or `executeParams.put("actionInput", actionInput);` respectively.

reuschling (changes requested):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
jngz-es: Thanks @reuschling for the detailed comment. Actually we are the first possibility you mentioned.
#### Flow Agent
We have a dedicated parameter parameter for the previous tool output here https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLFlowAgentRunner.java#L112. So you can reference it in the current tool input parameter as ${parameters.previous_tool_name.output}.
#### Conversational Agent
We take the action input for tools only from LLM response, i.e, the agent totally depends on LLM to do the ReAct. User can define a tool input to override the action input, but I don't see a use case for a placeholder of action input.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -465,12 +465,21 @@ public static Map<String, String> constructToolParams(
...
+            String input = toolSpecConfigMap.get("input");
+            StringSubstitutor substitutor = new StringSubstitutor(toolParams, "${parameters.", "}");
+            input = substitutor.replace(input);
+            toolParams.put("input", input);
```
ylwu-amzn, line 482: Add these
```
            if (isJson(actionInput)) {
                Map<String, String> params = getParameterMap(gson.fromJson(actionInput, Map.class));
                toolParams.putAll(params);
            }
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
reuschling: Good to hear that there is already a parameter for flow agent.
For Conversational agent, my example is exactly my use case. I want to predefine an Opensearch query, and the tool description says what's to insert. Thus, the previous llm step (the agent) doesn't have to generate the opensearch syntax, only a simple drop in, so I can guarantee this query will be executed, under all circumstances.
More specifically, I want to e. g. count documents with a query string inside.
Without a placeholder, the agent would have no possibility to adjust the query - only static queries independent from the previous steps in the thought chain would be possible.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
jngz-es: good catch, added

ylwu-amzn (approved):

ylwu-amzn (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLConversationalFlowAgentRunner.java
reuschling: @jngz-es any comment to my use case? This is my origin need and the reason behind #2918

jngz-es: Got your point. I created another issue to track it. https://github.com/opensearch-project/ml-commons/issues/3134

reuschling: Is there a chance left that it make it to 2.18?

yuye-aws: @reuschling The backport PR has been merged into the 2.x branch. The 2.18 branch will include this PR.

reuschling: I mean the conversational agent placeholder now tracked in #3134

jngz-es: Hi @reuschling , I raised a pr https://github.com/opensearch-project/ml-commons/pull/3200, could you take a look?

Issue #2974, opened 2024-09-19
[BUG] ML Inference Search Processors should have different model input format when input mapping has dollar symbols


mingshl:
**What is the bug?**
In ML Inference Processors, 

For ingest side, we consider the use case of  getting value from nested object and getting value from object, so we support both dot path notation to get field value, and support json path notation that can get nested value in a list. 

But for search processors, we support json path to read the object, the behavior is different, it always return the original data format. We want to have two ways of reading the object similar to ingest side. 

**How can one reproduce the bug?**

In 2.14, using ml_inference ingest processor: 

for nested document like this sample book index, 

```
{
  "book": [
    {
      "chunk": {
        "text": [
          {
            "chapter": "first chapter",
            "context": "this is the first part"
          },
          {
            "chapter": "first chapter",
            "context": "this is the second part"
          }
        ]
      }
    },
    {
      "chunk": {
        "text": [
          {
            "chapter": "second chapter",
            "context": "this is the third part"
          },
          {
            "chapter": "second chapter",
            "context": "this is the fourth part"
          }
        ]
      }
    }
  ]
} 
```
we can config the input maps as `{"input": "$.book.*.chunk.text.*.context"}`, which fetch the model input as 

```
{
  "input": [
    "this is the first part",
    "this is the second part",
    "this is the third part",
    "this is the fourth part"
  ]
}
```

for simple object like this sample `item` index, 
```

{
   "item_text":"red shoes"
}
```
if configuring input_map as `{"input": "item.text"}`

the model input will be in string representation. 

```
{
  "input": "red shoes"
}
```
if configuring input_map as `{"input": "$.item.text"}`

the model input will be list representation. 

```
{
  "input": ["red shoes"] 
}
```

But in search processors:

for the same `item` index, if configuring input_map as `{"input": "$.item.text"}`,

the model input will be in string representation. 

```
{
  "input": "red shoes"
}
```

**What is the expected behavior?**
we would like the similar logic as ingest processors, when using dot path notation without dollar symbol '$', it will get the original data format, but when using  dollar symbol '$', it should return a list of value representations. 

 in search processors:

for the same `item` index, if configuring input_map as `{"input": "$.item.text"}`,

the model input will be in list representation. 

```
{
  "input": ["red shoes"]
}
```
for the same `item` index, if configuring input_map as `{"input": "item.text"}`,

the model input will be in list representation. 

```
{
  "input": "red shoes"
}
```

**What is your host/environment?**
[...]

PR #2985, opened 2024-09-25
Fix ml inference ingest processor always return list using JsonPath

mingshl:
### Description

In ml inference ingest processor, when using jsonpath in input_maps, the json path configuration is set with `always_return_list`, however, in connectors and in ml inference search processors, the configuration is not always return list.

Trying to use the standard  json path configuration in modelExecutor class, so that ml inference search processors and ingest processors are using the same configuration. 

The default config now is returnning the original format of the object,

in the sample doc mentioned in the issue:
for simple object like this sample item index,

```
{
   "item_text":"red shoes"
}
```
, if configuring input_map as {"input": "$.item.text"},

the model input will be in original string format
```
{
  "input": "red shoes"
}
```

 if configuring input_map as {"input": "item.text"},

the model input will also remain the string format
```
{
  "input": "red shoes"
} 
```

### Related Issues

https://github.com/opensearch-project/ml-commons/issues/2974

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
+++ b/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
@@ -7,7 +7,6 @@ package org.opensearch.ml.processor;
 import static org.opensearch.ml.processor.InferenceProcessorAttributes.*;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
```diff
@@ -320,24 +312,29 @@ private void getMappedModelInputFromDocuments(
             Object documentFieldValue = ingestDocument.getFieldValue(originalFieldPath, Object.class);
             String documentFieldValueAsString = toString(documentFieldValue);
             updateModelParameters(modelInputFieldName, documentFieldValueAsString, modelParameters);
+            return;
```
ylwu-amzn, line 315: Seems like a bug , should we also backport this to old versions ?

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java
mingshl: of course, back-porting back till 2.14

rbhavna (approved):

PR #3080, opened 2024-10-08
Enhance batch job task management by adding default action types

rbhavna:
### Description
Enhance batch job task management by adding default action types

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -65,6 +65,8 @@ public interface Connector extends ToXContentObject, Writeable {
 
     List<ConnectorAction> getActions();
 
+    void setAction(ConnectorAction action);
+
     ConnectorClientConfig getConnectorClientConfig();
 
     String getActionEndpoint(String action, Map<String, String> parameters);
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -65,7 +65,7 @@ public interface Connector extends ToXContentObject, Writeable {
 
     List<ConnectorAction> getActions();
 
-    void setAction(ConnectorAction action);
+    void addAction(ConnectorAction action);
 
     ConnectorClientConfig getConnectorClientConfig();
 
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -23,6 +23,7 @@ import org.opensearch.core.xcontent.XContentParser;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.Setter;
 
 @Getter
 @EqualsAndHashCode
@@ -35,6 +36,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String REQUEST_BODY_FIELD = "request_body";
     public static final String ACTION_PRE_PROCESS_FUNCTION = "pre_process_function";
     public static final String ACTION_POST_PROCESS_FUNCTION = "post_process_function";
+
 
     private ActionType actionType;
     private String method;
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -23,7 +23,6 @@ import org.opensearch.core.xcontent.XContentParser;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import lombok.Setter;
 
 @Getter
 @EqualsAndHashCode
@@ -36,7 +35,6 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String REQUEST_BODY_FIELD = "request_body";
     public static final String ACTION_PRE_PROCESS_FUNCTION = "pre_process_function";
     public static final String ACTION_POST_PROCESS_FUNCTION = "post_process_function";
-
 
     private ActionType actionType;
     private String method;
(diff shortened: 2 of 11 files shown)


Zhangxunmt:
LGTM overall. Please apply spotLess and add UT coverages to pass CI. Also please define createConnectorAction() in the Util class so it can be shared across multiple classes.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -36,6 +37,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String ACTION_PRE_PROCESS_FUNCTION = "pre_process_function";
     public static final String ACTION_POST_PROCESS_FUNCTION = "post_process_function";
 
+
```
Zhangxunmt, line 40: No need to add this empty line.

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
```diff
@@ -65,6 +65,8 @@ public interface Connector extends ToXContentObject, Writeable {
 
     List<ConnectorAction> getActions();
 
+    void setAction(ConnectorAction action);
```
Zhangxunmt, line 68: Using addAction as the name is more accurate?

common/src/main/java/org/opensearch/ml/common/output/MLPredictionOutput.java
```diff
@@ -29,9 +30,11 @@ public class MLPredictionOutput extends MLOutput {
     public static final String TASK_ID_FIELD = "task_id";
     public static final String STATUS_FIELD = "status";
     public static final String PREDICTION_RESULT_FIELD = "prediction_result";
+    public static final String REMOTE_JOB_FIELD = "remote_job";
```
dhrubo-os, line 33: let's add a comment about this field.

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
```diff
@@ -245,4 +254,61 @@ private void processTaskResponse(MLTaskResponse taskResponse, ActionListener<MLC
...
+            .method(method)
+            .url(url)
+            .requestBody(requestBody)
+            .headers(headers)
+            .build();
+
+    }
+
+    private String getEndpointType(String url) {
+        if (url.contains("sagemaker"))
```
dhrubo-os, line 304: can we make static variables for all these values and then re-use?

plugin/src/main/java/org/opensearch/ml/action/tasks/GetTaskTransportAction.java
```diff
@@ -362,4 +371,62 @@ private boolean matchesPattern(Pattern pattern, String input) {
...
+            .actionType(BATCH_PREDICT_STATUS)
+            .method(method)
+            .url(url)
+            .requestBody(requestBody)
+            .headers(headers)
+            .build();
+
+    }
+
+    private String getEndpointType(String url) {
```
dhrubo-os, line 421: why do we need same method 2 times?

rbhavna: Yes will be moving the common methods to ConnectorUtils class

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -313,4 +318,58 @@ public static SdkHttpFullRequest buildSdkRequest(
...
+            case "cohere":
+                url = isCancelAction ? predictEndpoint + "/${parameters.id}/cancel" : predictEndpoint + "/${parameters.id}";
+                method = isCancelAction ? "POST" : "GET";
+                break;
+            case "bedrock":
+                url = isCancelAction
+                    ? predictEndpoint + "/${parameters.processedJobArn}/stop"
+                    : predictEndpoint + "/${parameters.processedJobArn}";
+                method = isCancelAction ? "POST" : "GET";
+                break;
```
Zhangxunmt, line 359: We should add the default branch for this switch statement to return null. In the GetTask and CancelTask, if the ConnectorAction is null, throw an exception with meaning logs like "please provide GetTask/CancelTask action in the connector".

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -61,6 +63,9 @@ public class ConnectorUtils {
     private static final Aws4Signer signer;
     public static final String SKIP_VALIDATE_MISSING_PARAMETERS = "skip_validating_missing_parameters";
 
+    public static final List<String> SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES = List
+        .of("sagemaker", "openai", "bedrock", "cohere");
+
```
Zhangxunmt, line 68: Ideally if a new platform is used but not listed here, CX should still be able to GetTask and CancelTask by manually adding the actions in the connector. But seems this is not the case in this PR?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
rbhavna: oh sorry I added it but I guess it got missed during refactoring. Let me add it

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
rbhavna: Yes if the connector already has the actions configured, then they can get/cancel task for any platform. Only if no action is provided, they we perform this check

Zhangxunmt (approved):

dhrubo-os (approved):

PR #3098, opened 2024-10-11
change to model group access for batch job task APIs

rbhavna:
### Description
change access control for batch job task APIs from connector to model group's access control

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
@@ -28,6 +28,7 @@ import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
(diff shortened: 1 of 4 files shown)


b4sjoo (approved):

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
```diff
@@ -177,25 +184,39 @@ private void processRemoteBatchPrediction(MLTask mlTask, ActionListener<MLCancel
...
+                            Connector connector = model.getConnector();
+                            executeConnector(connector, mlInput, actionListener);
+                        } else if (clusterService.state().metadata().hasIndex(ML_CONNECTOR_INDEX)) {
+                            ActionListener<Connector> listener = ActionListener
+                                .wrap(connector -> { executeConnector(connector, mlInput, actionListener); }, e -> {
+                                    log.error("Failed to get connector " + model.getConnectorId(), e);
+                                    actionListener.onFailure(e);
+                                });
+                            try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
+                                connectorAccessControlHelper
```
Zhangxunmt, line 205: Do we need to check connector access control again here? Are model/connector all access controlled through the model group so ideally any role would have the same permission to access both models and connectors?

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
rbhavna: I think its fine. batch_predict also only checks for only model access for now. Since predict API is controlled by model access, it makes sense to have it for batch job APIs too. If we are planning to have connector access too, then we should also include it into predict APIs for consistency.

And yes, I don't think connector access is controlled by model group access control. They were designed separately. I think we have to either unify them or remove one of them for remote models. Does not make sense to have both at connector and model level.

PR #2976, opened 2024-09-23
Filter out remote model auto redeployment

zane-neo:
### Description
Currently, ml-commons supports remote model auto deployment when first prediction request comes, so in MLModelAutoRedeplyer remote model auto redeployment can be  filtered out.

### Related Issues
NA

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3104, opened 2024-10-14
Fix PR #2976 bug due to missing adding function_name and algorithm in querying models

zane-neo:
### Description
Fix PR #2976 bug due to missing adding function_name and algorithm in querying models

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
+++ b/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
@@ -243,7 +243,7 @@ public class MLModelAutoReDeployer {
             MLModel.PLANNING_WORKER_NODES_FIELD,
             MLModel.DEPLOY_TO_ALL_NODES_FIELD,
             MLModel.FUNCTION_NAME_FIELD,
-            MLModel.ALGORITHM_FIELD};
+            MLModel.ALGORITHM_FIELD };
 
         String[] excludes = new String[] { MLModel.MODEL_CONTENT_FIELD, MLModel.OLD_MODEL_CONTENT_FIELD };
         FetchSourceContext fetchContext = new FetchSourceContext(true, includes, excludes);
@@ -266,7 +266,11 @@ public class MLModelAutoReDeployer {
             .ofNullable(sourceAsMap.get(MLModel.FUNCTION_NAME_FIELD))
             .orElse(sourceAsMap.get(MLModel.ALGORITHM_FIELD));
         if (functionName == null) {
-            log.error("Model function_name or algorithm is null, model is not in correct status, please check the model, model id is: {}", modelId);
+            log
+                .error(
+                    "Model function_name or algorithm is null, model is not in correct status, please check the model, model id is: {}",
+                    modelId
+                );
             return;
         }
         if (FunctionName.REMOTE == FunctionName.from(functionName)) {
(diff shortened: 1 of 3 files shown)


ylwu-amzn (dismissed):

plugin/src/test/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployerIT.java
```diff
@@ -0,0 +1,73 @@
...
+                });
+                waitForTask(deployFirstModelTaskId, MLTaskState.COMPLETED);
+            } catch (Exception e) {
+                logger.error(e.getMessage(), e);
+                fail(e.getMessage());
+            }
+        });
+    }
+
+    public void testModelAutoRedeploy() {
```
dhrubo-os, line 68: What is this method for?

plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
```diff
@@ -263,6 +265,14 @@ private void triggerModelRedeploy(ModelAutoRedeployArrangement modelAutoRedeploy
         String functionName = (String) Optional
             .ofNullable(sourceAsMap.get(MLModel.FUNCTION_NAME_FIELD))
             .orElse(sourceAsMap.get(MLModel.ALGORITHM_FIELD));
+        if (functionName == null) {
+            log
+                .error(
+                    "Model function_name or algorithm is null, model is not in correct status, please check the model, model id is: {}",
```
dhrubo-os, line 271: Does model deployment depends on model function name?

plugin/src/test/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployerIT.java
zane-neo: This is an empty method to trigger the IT's setup method, which is to immediately deploy 2 models when cluster spin up, and the model auto redeploy can also select them to auto redeploy, if the test method complete with error, it means some method in model auto redeploy is buggy. Please check the comments in the method body and the `setup` method.

plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
zane-neo: It doesn't at the beginning, but in this PR: #2976, we filtering out the remote model, that why we check function name here. And any model should have function name even auto redeploy doesn't depend on that: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java#L413

plugin/src/test/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployerIT.java
zane-neo: Moved the setup method body to the test method body, it can reproduce the failure case as well.

plugin/src/test/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployerIT.java
```diff
@@ -0,0 +1,65 @@
...
+import lombok.SneakyThrows;
+
+public class MLModelAutoReDeployerIT extends MLCommonsRestTestCase {
+
+    public void testModelAutoRedeploy() {
+        prepareModel();
+    }
+
+    @SneakyThrows
+    private void prepareModel() {
```
dhrubo-os, line 27: [nit] why do we need this method then?

dhrubo-os (approved):

plugin/src/test/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployerIT.java
zane-neo: It can be removed but it's not an urgent issue. We can optimize in future PR

b4sjoo (approved):

Issue #3093, opened 2024-10-11
[Enhancement] tune bedrock titan embedding model post process function to support v2


ylwu-amzn:
The current post process function of Bedrock Titan embedding model V2 can't support `embeddingTypes` parameter.

ylwu-amzn:
@zane-neo Can you take this issue? I see you added Bedrock text embedding model process function https://github.com/opensearch-project/ml-commons/pull/1554

PR #3136, opened 2024-10-22
Add dimensions parameter support for bedrock titan embedding v2 model

zane-neo:
### Description
1. Optimized the `Function` usage by adding a new interface `PreProcessFunction` as there isn't cases that the Function accepts different type of parameters. Currently it only accepts `MLInput` and `Map<String, String>`
2. Add support for `dimensions` parameter introduced in titan embedding v2 model.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3093

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/BedrockEmbeddingPreProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/BedrockEmbeddingPreProcessFunction.java
@@ -8,7 +8,9 @@ package org.opensearch.ml.common.connector.functions.preprocess;
 import static org.opensearch.ml.common.utils.StringUtils.convertScriptStringToJsonString;
 
 import java.util.Map;
+import java.util.Optional;
 
+import org.apache.commons.lang3.math.NumberUtils;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/BedrockEmbeddingPreProcessFunction.java
```diff
@@ -8,11 +8,16 @@
...
+import java.util.Optional;
 
+import org.apache.commons.lang3.math.NumberUtils;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
 
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
```
dhrubo-os, line 20: can we add @Log4j instead to keep the consistency?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/BedrockEmbeddingPreProcessFunction.java
```diff
@@ -24,10 +29,23 @@ public void validate(MLInput mlInput) {
...
         Map<String, Object> processedResult = Map.of("parameters", Map.of("inputText", inputData.getDocs().get(0)));
         return RemoteInferenceInputDataSet.builder().parameters(convertScriptStringToJsonString(processedResult)).build();
     }
+
+    @Override
+    public RemoteInferenceInputDataSet process(Map<String, String> connectorParams, MLInput mlInput) {
+        TextDocsInputDataSet inputData = (TextDocsInputDataSet) mlInput.getInputDataset();
+        // Amazon Titan Text Embeddings V2 model: https://docs.aws.amazon.com/bedrock/latest/userguide/titan-embedding-models.html
+        // Default dimension is 1024
+        int dimensions = Optional.ofNullable(connectorParams.get("dimensions")).map(x -> NumberUtils.toInt(x, 1024)).orElse(1024);
```
dhrubo-os, line 45: Is BedrockEmbeddingPreProcessFunction only for titan models? What about other bedrock models?

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
```diff
@@ -242,6 +244,108 @@ public void test_bedrock_multimodal_model_empty_imageInput_null_textInput() thro
...
+        RemoteInferenceInputDataSet inputDataSet = RemoteInferenceInputDataSet
+            .builder()
+            .parameters(Map.of("inputText", "Can you tell me a joke?", "dimensions", "512"))
+            .actionType(ConnectorAction.ActionType.PREDICT)
+            .build();
+        MLInput mlInput = MLInput.builder().inputDataset(inputDataSet).algorithm(FunctionName.REMOTE).build();
+        Map inferenceResult = predictTextEmbeddingModelIgnoreFunctionName(modelId, mlInput);
+        String errorMsg = String
+            .format(Locale.ROOT, "Failing test case name: %s, inference result: %s", testCaseName, gson.toJson(inferenceResult));
+        assertTrue(errorMsg, inferenceResult.containsKey("inference_results"));
```
dhrubo-os, line 337: Why it is showing error here?

common/src/main/java/org/opensearch/ml/common/connector/functions/preprocess/BedrockEmbeddingPreProcessFunction.java
xinyual: I also notice we have multimodal embedding model: https://docs.aws.amazon.com/bedrock/latest/userguide/titan-multiemb-models.html. Is the multimodal model using `MultiModalConnectorPreProcessFunction`, if so, can we also add dimension for this?

zane-neo: > Is BedrockEmbeddingPreProcessFunction only for titan models? What about other bedrock models?
currently this is a titan specific process function, but if in the future bedrock has another similar(similar means they have same request body, e.g. inputText & dimensions etc) text embedding model, then this can be reused to that one.
BedRock integrated with multiple models like anthropic, cohere, meta etc, currently we only have `titan, cohere` pre process functions in our code, from cohere API: https://docs.cohere.com/reference/embed I didn't see they support dimension parameter so it's fine.

dhrubo-os: In that case do we need to set the default dimension? If dimension is provided then we will add otherwise not?

zane-neo:
After adding more details of the issue, this PR is not necessary and please check the new PR: https://github.com/opensearch-project/ml-commons/pull/3262

PR #3158, opened 2024-10-23
applying multi-tenancy for forward request which is important for deploy request

dhrubo-os:
…loy request

### Description
[applying multi-tenancy for forward request which is important for deploy request

purpose of TransportForwardAction.java is to forward the request to other nodes in a multi-node cluster
Also at the same time, currently in multi-tenancy, if a model is deployed, the status is not being updated in the index. Because that update wasn't happening through sdkclient.
]

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
@@ -187,8 +187,7 @@ public class CreateControllerTransportAction extends HandledTransportAction<Acti
                     log.info(errorMessage, indexResponse.getResult());
                     if (indexResponse.getResult() == DocWriteResponse.Result.CREATED) {
                         // we aren't enabling controller feature for multi-tenancy. So tenant id is null by default.
-                        mlModelManager
-                            .updateModel(modelId, null, isHidden, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
+                        mlModelManager.updateModel(modelId, null, isHidden, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
                     }
                     if (!ArrayUtils.isEmpty(mlModelCacheHelper.getWorkerNodes(modelId))) {
                         log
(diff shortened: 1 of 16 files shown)


plugin/src/main/java/org/opensearch/ml/action/controller/CreateControllerTransportAction.java
```diff
@@ -186,7 +186,8 @@ private void indexAndCreateController(
                     String errorMessage = getErrorMessage("Model controller saved into index, result:{}", modelId, isHidden);
                     log.info(errorMessage, indexResponse.getResult());
                     if (indexResponse.getResult() == DocWriteResponse.Result.CREATED) {
-                        mlModelManager.updateModel(modelId, isHidden, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
+                        mlModelManager
+                            .updateModel(modelId, mlModel.getTenantId(), isHidden, Map.of(MLModel.IS_CONTROLLER_ENABLED_FIELD, true));
```
arjunkumargiri, line 190: Better to not add tenant ID to controller action. We are now adding multi tenancy to create controller but not to delete controller.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -285,10 +285,10 @@ private void deployModel(
...
             mlTask.setTaskId(taskId);
             if (algorithm == FunctionName.REMOTE) {
-                if (mlFeatureEnabledSetting.isMultiTenancyEnabled()) {
-                    listener.onResponse(new MLDeployModelResponse(taskId, MLTaskType.DEPLOY_MODEL, MLTaskState.CREATED.name()));
-                    return;
-                }
+                // if (mlFeatureEnabledSetting.isMultiTenancyEnabled()) {
+                // listener.onResponse(new MLDeployModelResponse(taskId, MLTaskType.DEPLOY_MODEL, MLTaskState.CREATED.name()));
+                // return;
+                // }
```
arjunkumargiri, line 291: Why are these commented? Remove them if they are no longer required.

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
```diff
@@ -302,6 +302,7 @@ private void deployModel(
                         () -> updateModelDeployStatusAndTriggerOnNodesAction(
                             modelId,
                             taskId,
+                            tenantId,
```
arjunkumargiri, line 305: In a multi node setting, how will the model be deployed across all nodes?

plugin/src/main/java/org/opensearch/ml/action/deploy/TransportDeployModelAction.java
dhrubo-os: Yeah, I meant to remove. I'll remove this in next rev

Issue #3119, opened 2024-10-16
[CI] Inconsistent test configuration between modules


dbwiddis:
**Is your feature request related to a problem?**

Tests run in the `:opensearch-ml-plugin` module behave differently than in `:opensearch-ml-common`.  
Specific pain points:
1. The JUnit `@Test` annotation is required in the `common` and many other modules.  However, it is not required anywhere in the `plugin` module since the `build.gradle` file has a `test { include '**/*Tests.class' }` configuration, which executes all methods in classes ending with `Tests` as tests.  This is confusing, leading to inadvertent disabling of tests when refactoring them from the plugin module (where the worked) to the common module (where they didn't even run).
2. Forbidden APIs checks only seem to be configured in the `plugin` module and [spam the build logs](https://github.com/opensearch-project/ml-commons/actions/runs/11369841586/job/31628242139#step:5:657) (only for the `plugin` module) saying `Forbidden annotation use: org.junit.Test [defaultMessage Just name your test method testFooBar]` 284 times.  This is actually inaccurate since there are no method name checks at all, it should really say what to name your class.  This is bad for two reasons:  it's confusing and leads to leaving off `@Test` annotation outside the `plugin` module for anyone who's looked at a lot of these logs; and it's extremely spammy adding 284*2=568 useless lines of text to the log.
3. Failing tests in the `plugin` module give the useful `REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:test'` summary with logs and a stack trace.  However, failing tests in the `common` module just tell me that a test method failed, giving its line number and no other details.  I can force them to show locally with gradle `--info` and `--stacktrace` switches, but that doesn't help when trying to diagnose CI build logs.

**What solution would you like?**

1. Align on a standard convention across the entire plugin
2. Remove the incorrect ForbiddenAPI check saying the annotation isn't required
3. Configure other modules to give useful info like plugin module currently does

And while I'm asking about tests, PLEASE don't cancel other tests when CI fails on one operating system.  Given how flaky tests currently are in this repo, it should not be assumed that a single failing test on one OS means all will fail.  This can easily be done by adding `fail-fast: false` under the `strategy:` entry for the build matrix.  See [example](https://github.com/opensearch-project/flow-framework/blob/389d680f6149255df54d44592c95f1cf8818f9c5/.github/workflows/CI.yml#L66), at least until the gradle check consistently passes.

[...]

PR #3192, opened 2024-10-30
[Feature/multi_tenancy] Implement client.bulk in SDK Client

dbwiddis:
### Description

Adds an SDKClient equivalent to `client.bulk()`.

Summary of changes (helpful for reviewers):
1. Added a new `DataObjectRequest` parent class. I originally started with a tagging interface just to help method signatures, but realized that four classes duplicated the same code for the index, id, and tenantId fields, so refactored those (and the associated builder code) into the superclass. Existing unit tests cover this change. https://github.com/opensearch-project/ml-commons/pull/3192/commits/caeea219cef12a70d344f2639f17f13ac9a62956 
2. Added a new `DataObjectResponse` parent class similar to above, refactoring the id and parser fields from the subclass. Existing unit tests cover this change. https://github.com/opensearch-project/ml-commons/pull/3192/commits/445936bcfa23410778cc024a485b69b307a44c62 (part 1)
3. Added new `BulkDataObjectRequest` and `BulkDataObjectResponse` classes and unit tests.  Added a `failed` field to the response superclass to track failures/errors. https://github.com/opensearch-project/ml-commons/pull/3192/commits/445936bcfa23410778cc024a485b69b307a44c62 (part 2)
4. Added the interface and stubbed implementations for the SdkClient `.bulk()` method, and unit tests https://github.com/opensearch-project/ml-commons/pull/3192/commits/22dccf4af0c41f07625ffe6857fd17df80b0bf79
5. Implemented the new interface method in all 3 clients.  The local and remote implementations use the OpenSearch bulk request natively.  The DynamoDB implementation simply iterates the bulk request and calls existing code for the individual requests; this is not the most efficient but necessary to preserve ordering.  If all the bulk requests are put and delete, it's possible to optimize this with DDB BulkWrite, which is a good future optimization; however since our primary use case uses update I'm deferring that complexity. https://github.com/opensearch-project/ml-commons/pull/3192/commits/ee30d210818a84b89027860e5c3a0ac7762a950b
6. Additional refactoring in response to code review: move the details of ingest_took to a lower level https://github.com/opensearch-project/ml-commons/pull/3192/commits/7f190785408ccbe60899eb6dc45fa78bcafd1033
7. More parameters in the sdkclient request/response to handle the needed implementation https://github.com/opensearch-project/ml-commons/pull/3192/commits/05a64d1553bc8b60be278d16b59bad488344a972
8. Implementation and tests https://github.com/opensearch-project/ml-commons/pull/3192/commits/e2a5ec638ef83265bc751bd788af77c78d72b746

In addition there are several whitespace / import reordering changes due to spotless not having been run on the common module. 

Will be handled integ tests in #2818 

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
+++ b/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.common.Nullable;
+import org.opensearch.core.common.Strings;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class BulkDataObjectRequest {
+
+    private final List<DataObjectRequest> requests = new ArrayList<>();
+    private final Set<String> indices = new HashSet<>();
+    private String globalIndex;
+
+    public BulkDataObjectRequest() {}
+
+    /**
+     * Instantiate this request with a global index.
+     * <p>
+     * For data storage implementations other than OpenSearch, an index may be referred to as a table and the id may be referred to as a primary key.
+     * @param index the index location for all the bulk requests
+     */
+    public BulkDataObjectRequest(@Nullable String globalIndex) {
+        this.globalIndex = globalIndex;
+    }
+
+    /**
+     * Returns the list of requests in this bulk request.
+     * @return the requests list
+     */
+    public List<DataObjectRequest> requests() {
+        return List.copyOf(this.requests);
+    }
+
+    /**
+     * Returns the indices being updated in this bulk request.
+     * @return the indices being updated
+     */
+    public Set<String> getIndices() {
+        return Collections.unmodifiableSet(indices);
+    }
+
+    /**
+     * Add the given request to the {@link BulkDataObjectRequest}
+     * @param request The request to add
+     * @return the updated request object
+     */
+    public BulkDataObjectRequest add(DataObjectRequest request) {
+        if (!request.isWriteRequest()) {
+            throw new IllegalArgumentException("No support for request [" + request.getClass().getName() + "]");
+        }
+        if (Strings.isNullOrEmpty(request.index())) {
+            if (Strings.isNullOrEmpty(globalIndex)) {
+                throw new IllegalArgumentException(
+                    "Either the request [" + request.getClass().getName() + "] or the bulk request must specify an index."
+                );
+            }
+            indices.add(globalIndex);
+            request.index(globalIndex);
+        } else {
+            indices.add(request.index());
+        }
+        requests.add(request);
+        return this;
+    }
+}
--- a/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
+++ b/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
@@ -14,6 +14,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.opensearch.action.support.WriteRequest.RefreshPolicy;
 import org.opensearch.common.Nullable;
 import org.opensearch.core.common.Strings;
 
@@ -21,6 +22,7 @@ public class BulkDataObjectRequest {
 
     private final List<DataObjectRequest> requests = new ArrayList<>();
     private final Set<String> indices = new HashSet<>();
+    private RefreshPolicy refreshPolicy = RefreshPolicy.NONE;
     private String globalIndex;
     private String globalTenantId;
 
@@ -87,6 +89,23 @@ public class BulkDataObjectRequest {
         return this;
     }
 
+    /**
+     * Should this request trigger a refresh ({@linkplain RefreshPolicy#IMMEDIATE}), wait for a refresh (
+     * {@linkplain RefreshPolicy#WAIT_UNTIL}), or proceed ignore refreshes entirely ({@linkplain RefreshPolicy#NONE}, the default).
+     */
+    public BulkDataObjectRequest setRefreshPolicy(RefreshPolicy refreshPolicy) {
+        this.refreshPolicy = refreshPolicy;
+        return this;
+    }
+
+    /**
+     * Should this request trigger a refresh ({@linkplain RefreshPolicy#IMMEDIATE}), wait for a refresh (
+     * {@linkplain RefreshPolicy#WAIT_UNTIL}), or proceed ignore refreshes entirely ({@linkplain RefreshPolicy#NONE}, the default).
+     */
+    public RefreshPolicy getRefreshPolicy() {
+        return refreshPolicy;
+    }
+    
     /**
      * Instantiate a builder for this object
      * @return a builder instance
--- a/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
+++ b/common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
@@ -24,18 +24,15 @@ public class BulkDataObjectRequest {
     private final Set<String> indices = new HashSet<>();
     private RefreshPolicy refreshPolicy = RefreshPolicy.NONE;
     private String globalIndex;
-    private String globalTenantId;
 
     /**
      * Instantiate this request with a global index.
      * <p>
      * For data storage implementations other than OpenSearch, an index may be referred to as a table and the id may be referred to as a primary key.
      * @param globalIndex the index location for all the bulk requests as a default if not already specified
-     * @param globalTenantId the tenantId for all the bulk requests, overwriting what's specified if not null
      */
-    public BulkDataObjectRequest(@Nullable String globalIndex, @Nullable String globalTenantId) {
+    public BulkDataObjectRequest(@Nullable String globalIndex) {
         this.globalIndex = globalIndex;
-        this.globalTenantId = globalTenantId;
     }
 
     /**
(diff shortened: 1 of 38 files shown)


dhrubo-os:
apply spotless

dbwiddis:
Will fix spotless in next commit.  Apparently my IDE import ordering is different than spotless, needed on common, not needed on plugin. See #3119 😁

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -352,6 +360,62 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(
...
+                        responses.add(new DeleteDataObjectResponse.Builder().id(dataObjectRequest.id()).failed(true).build());
+                    }
+                    log.error("Error in bulk operation for id {}: {}", dataObjectRequest.id(), e.getCause().getMessage(), e.getCause());
+                }
+            }
+            long endNanos = System.nanoTime();
+            long tookMillis = TimeUnit.NANOSECONDS.toMillis(endNanos - startNanos);
+
+            log.info("Bulk action complete for {} items, took {} ms", responses.size(), tookMillis);
+            return new BulkDataObjectResponse(responses.toArray(new DataObjectResponse[0]), tookMillis, NO_INGEST_TOOK);
```
dhrubo-os, line 415: Just curious to know what is `NO_INGEST_TOOK`?

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -214,6 +220,99 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(
...
+        Boolean isMultiTenancyEnabled
+    ) {
+        return CompletableFuture.supplyAsync(() -> AccessController.doPrivileged((PrivilegedAction<BulkDataObjectResponse>) () -> {
+            try {
+                log.info("Performing {} bulk actions on indices {}", request.requests().size(), request.getIndices());
+                BulkRequest bulkRequest = new BulkRequest();
+
+                for (DataObjectRequest dataObjectRequest : request.requests()) {
+                    if (dataObjectRequest instanceof PutDataObjectRequest) {
+                        try (XContentBuilder sourceBuilder = XContentFactory.jsonBuilder()) {
```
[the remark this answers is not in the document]
dbwiddis: > Should we add any retry mechanism for bulk requests?

In the case of the Local and Remote clients, we are just delegating exactly to the existing NodeClient version of this.  So we would only add retries on the individual requests if they already existed in the code.

In the case of the two places where we are implementing this for undeploying a model, we are only using the update request and we are not doing retries there.  If there's a retry on the update request it's handled here, but it doesn't look like that was actually done.  (When I implement this I'll make sure every option is handled.)

We can always add the capability in any of the requests later if it is needed; I have intentionally not tried to replicate every possible combination of parameters for the request objects until they are needed.

> during that time may be for a network glitches it failed to ingest one document. What happens that time?

1. That individual request would fail (the `failed` boolean would be set true in the `DataObjectResponse` subclass) and the JSON response could be decompiled into the full `BulkItemResponse` object with error information.
2. The caller would be able to test for any failures with a single check, and if there were, iterate the `BulkResponse` array to find the failed items.

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
dbwiddis: The `BulkResponse` has 3 fields, the array of `BulkItemResponse` objects, how long the operation took, and in some special cases how long ingestion took.  Since we're not ingesting that field stays at its default, which here is statically imported from here, basically indicating that data is not relevant:
https://github.com/opensearch-project/OpenSearch/blob/0363aa79c434343dc7f5b8df9f77b2d4652f4c7e/server/src/main/java/org/opensearch/action/bulk/BulkResponse.java#L70

dbwiddis: The `BulkResponse` has a 2-arg and 3-arg constructor, I could do the same and hide this implementation detail... probably a better option.

dbwiddis:
@dhrubo-os @arjunkumargiri This is ready for (re-) review.

common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.opensearch.common.Nullable;
+import org.opensearch.core.common.Strings;
+
+public class BulkDataObjectRequest {
+
+    private final List<DataObjectRequest> requests = new ArrayList<>();
+    private final Set<String> indices = new HashSet<>();
+    private RefreshPolicy refreshPolicy = RefreshPolicy.NONE;
+    private String globalIndex;
+    private String globalTenantId;
```
dhrubo-os, line 27: Add a comment: why it's global?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
```diff
@@ -178,7 +188,11 @@ void processUndeployModelResponseAndUpdate(
                         updateDocument.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, newPlanningWorkerNodes.size());
                         deployToAllNodes.put(modelId, false);
                     }
-                    updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(updateDocument);
+                    UpdateDataObjectRequest updateRequest = UpdateDataObjectRequest
+                        .builder()
+                        .id(modelId)
```
dhrubo-os, line 193: shouldn't we add `tenantId` here?

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
```diff
@@ -226,6 +237,127 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(
         }), executor);
     }
 
+    @Override
+    public CompletionStage<BulkDataObjectResponse> bulkDataObjectAsync(
```
dhrubo-os, line 241: This method seems too long. Can we break down this method?

common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
dbwiddis: This allows specifying it once when creating a bulk request without having to specify it individually for every sub-request.   Including this global value means we can validate the existence of a tenantid before executing, like we do for all the other requests.   We can remove this and then just iterate all the requests and confirm that they have non-null tenant IDs.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
dbwiddis: It's added globally on line 158 which overrides this.

common/src/main/java/org/opensearch/sdk/BulkDataObjectRequest.java
dbwiddis: Removed this and updated code to check all the individual requests.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java
dbwiddis: Removed global tenantID and added here.

plugin/src/main/java/org/opensearch/ml/sdkclient/RemoteClusterIndicesClient.java
dbwiddis: Done.

dhrubo-os (approved):

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
```diff
@@ -352,6 +371,138 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(
...
+
+            List<DataObjectResponse> responses = new ArrayList<>();
+
+            long startNanos = System.nanoTime();
+            for (DataObjectRequest dataObjectRequest : request.requests()) {
+                try {
+                    if (dataObjectRequest instanceof PutDataObjectRequest) {
+                        responses
+                            .add(
+                                putDataObjectAsync((PutDataObjectRequest) dataObjectRequest, executor, isMultiTenancyEnabled)
```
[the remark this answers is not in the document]
dbwiddis: > Dynamo DB supports bulk operations

Only for put and delete.  I had actually started trying to implement it that way with conditionals on whether we were updating, but it got rather complex very quickly with the proper handling of `seqNo`.  Since we are only using it for update we wouldn't even use that code.

I agree that in the long term we should use that API when there are no update requests.

common/src/main/java/org/opensearch/sdk/client/LocalClusterIndicesClient.java
```diff
@@ -214,6 +220,105 @@ public CompletionStage<DeleteDataObjectResponse> deleteDataObjectAsync(
...
+                            PutDataObjectRequest putRequest = (PutDataObjectRequest) dataObjectRequest;
+                            IndexRequest indexRequest = new IndexRequest(putRequest.index())
+                                .opType(putRequest.overwriteIfExists() ? OpType.INDEX : OpType.CREATE)
+                                .source(putRequest.dataObject().toXContent(sourceBuilder, EMPTY_PARAMS));
+                            if (!Strings.isNullOrEmpty(putRequest.id())) {
+                                indexRequest.id(putRequest.id());
```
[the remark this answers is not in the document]
dbwiddis: @arjunkumargiri see https://github.com/opensearch-project/ml-commons/pull/3192/commits/1355af0581194d964009c427f1d63da72d34d898

plugin/src/main/java/org/opensearch/ml/sdkclient/DDBOpenSearchClient.java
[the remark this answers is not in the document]
dbwiddis: (I also noted this in item 5 of the PR description.  Added a comment in the latest commit.)

arjunkumargiri (approved):

dhrubo-os (approved):

Issue #3191, opened 2024-10-30
[RFC] Check services before deleting MLModel


xinyual:
## Background 

Currently, models in ML-commons are widely used, particularly by pipelines and the agent framework. However, we lack a mechanism to detect if a model is being utilized by other services. To address this, we propose a new feature that will identify whether a model is in use by another service. If so, an error should be raised—similar to when attempting to delete a connector that is still in use (like [here](https://github.com/opensearch-project/ml-commons/blob/cc99c9dc3be8bfe2342d51cf21d29232cdc44a7d/plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java#L83C29-L83C31)).

## All related service
We list all service using MLModel.

- Pipeline 
   - rag-pipeline
   - neural-sparse search pipeline
   - neural-sparse ingest pipeline
   - neural-sparse two phase pipeline
   - ml inference pipeline
   - text-embedding
- Tools
   - MLModel Tool
   - PPLTool
   - RAGTool
   - .....

## Solution
We can apply the same solution used for [connectors and models](https://github.com/opensearch-project/ml-commons/blob/cc99c9dc3be8bfe2342d51cf21d29232cdc44a7d/plugin/src/main/java/org/opensearch/ml/action/connector/DeleteConnectorTransportAction.java#L76). Generally, when attempting to delete a specific model ID, we can search both services (pipelines and agents) to check if they are using the current model ID. We have separate solution for each service since they are different

### Pipeline:
For all the pipelines we developed, we found that they all use the key **model_id** to document their relevant model id. Therefore, we can use **GetPipelineAction** or **GetSearchPipelineAction** to retrieve all pipelines and then check the key “model_id” to detect whether they reference the current model ID.

### Agent and Tools
Tools are more complex than pipelines because they use different keys to store the model IDs they reference. Furthermore, some tools use more than one model ID. For example, **RAGTool** requires both **embedding_model_id** and **inference_model_id**, while **PPLTool** uses **model_id** to document its model. The target field we need to check varies by tool, making it challenging to detect relevant models with a single template. Therefore, we propose a few options to address this complexity.

#### Option 1 (prefer)
We add an abstract function **GETAllModelKeys** in Base Class Factory<Tool>. 
```
interface Factory<T extends Tool> {
     List<String> GETAllModelKeys;
}
```
[...]

Issue #3088, opened 2024-10-10
[Feature] Add warnings on model undeployment/deletion if associated in existing resources


ohltyler:
We can add helper checks via API (or at the least ML commons UI) to scan ingest pipelines, search pipelines, any other places where the models may be used in existing resources. At the least, provide some warning that undeploying/deleting a model may break some existing resources.

Issue #3087, opened 2024-10-10
[FEATURE] Check downstream services before deleting MLModel


yuye-aws:
**Is your feature request related to a problem?**
There are several types of downstream services dependent on **MLModel**. For example, several tools in agent framework (**MLModelTool**, **VectorDBTool** and **RAGTool**) uses MLModel. When deleting the MLModel, we need to first check whether any of the downstream service is using this MLModel.

**What solution would you like?**
Just like connector, it will return an error if the connector is being used by any model:

`1 models are still using this connector, please delete or update the models first: [L05NKJIBVJ7VbbiuxOUx]
`

**What alternatives have you considered?**

**Do you have any additional context?**

ylwu-amzn:
Thanks @yuye-aws, Do you have bandwidth to help on this ?

ylwu-amzn:
Assign to you first, let me know if you don't have enough bandwidth.

yuye-aws:
I'm working on the opensearch-ml-quickstart project in these weeks. @ylwu-amzn You can mark this issue as 2.19.

yuye-aws:
Plus one use case: Check ConnectorTool agent before deleting the connector

dhrubo-os:
Similar issue: https://github.com/opensearch-project/ml-commons/issues/3088

yuye-aws:
@dhrubo-os The issue mentions that `scan ingest pipelines, search pipelines, any other places where the models may be used in existing resources`. Can you at least provide a list of possible places where ml model is being used.

xinyual:
@ylwu-amzn Hi Yaliang, I can take this issue and come up with a RFC ASAP since yuye doesn't have enough bandwidth.

yuye-aws:
> @ylwu-amzn Hi Yaliang, I will take this issue and come up with a RFC ASAP since yuye doesn't have enough bandwidth.

Thanks Xinyuan! This issue can be assigned to you, right?

yuye-aws:
@xinyual Found that RAG tool also has two fields: embedding_model_id and inference_model_id  https://opensearch.org/docs/latest/ml-commons-plugin/agents-tools/tools/rag-tool/

yuye-aws:
Also the conversational agent needs `llm.model_id` https://opensearch.org/docs/latest/ml-commons-plugin/api/agent-apis/register-agent/

yuye-aws:
[...]

PR #3209, opened 2024-11-11
Check before delete

xinyual:
### Description
This pr is to check all downstream service before deleting ml model. Two downstream tasks are checking here:
1. Agent
2. Pipelines

For agent, we enforce this tool factory to implement a method to return the key field name of each tool. Then we create a should DSL query like 
```
{
  "query" : {
    "bool" : {
      "must" : [ {
        "bool" : {
          "should" : [ {
            "bool" : {
              "must_not" : [ {
                "exists" : {
                  "field" : "is_hidden",
                  "boost" : 1.0
                }
              } ],
              "adjust_pure_negative" : true,
              "boost" : 1.0
            }
          }, {
            "bool" : {
              "must" : [ {
                "terms" : {
                  "is_hidden" : [ false ],
                  "boost" : 1.0
                }
              }, {
                "exists" : {
                  "field" : "is_hidden",
                  "boost" : 1.0
                }
              } ],
              "adjust_pure_negative" : true,
              "boost" : 1.0
            }
          } ],
          "adjust_pure_negative" : true,
          "boost" : 1.0
        }
      }, {
        "bool" : {
          "should" : [ {
            "terms" : {
              "tools.parameters.<fieldName>" : [ "<modelId>" ],
              "boost" : 1.0
            }
          }],
          "adjust_pure_negative" : true,
          "boost" : 1.0
        }
      } ],
      "adjust_pure_negative" : true,
      "boost" : 1.0
    }
  }
}
```

For pipelines, we fetch all ingestion pipelines and search pipelines and check for each pipeline whether they contain the candidate model id.

### Related Issues
#3191 
#3087 
#3088

### Check List


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/RelatedModelIdHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+public class RelatedModelIdHelper {
```
zane-neo, line 19: A better name would be `AgentModelsSearcher`

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -84,25 +100,145 @@ public DeleteModelTransportAction(
...
         this.clusterService = clusterService;
         this.modelAccessControlHelper = modelAccessControlHelper;
+        this.relatedModelIdHelper = relatedModelIdHelper;
     }
 
     @Override
     protected void doExecute(Task task, ActionRequest request, ActionListener<DeleteResponse> actionListener) {
         MLModelDeleteRequest mlModelDeleteRequest = MLModelDeleteRequest.fromActionRequest(request);
         String modelId = mlModelDeleteRequest.getModelId();
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
zane-neo, line 118: Let's move the search agent models logic to a centralized place e.g. AgentModelsSearcher, and encapsulate them to return a boolean value to identify if anywhere using the model id.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -84,25 +100,145 @@ public DeleteModelTransportAction(
...
+            // check whether agent are using them
+            SearchRequest searchAgentRequest = relatedModelIdHelper.constructQueryRequest(modelId);
+            client.search(searchAgentRequest, ActionListener.runBefore(ActionListener.wrap(searchResponse -> {
```
zane-neo, line 121: Instead of querying agent, pipelines in a sequential manner, you can divide them to multiple single method responsible for single search and execute them parallel, each method can have a reference of CountDownLatch instance to coordinate them, refer: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java#L243

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +469,46 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
             && !mlModelState.equals(MLModelState.PARTIALLY_DEPLOYED);
     }
 
+    private <T> Boolean isPipelineContainsModel(
+        List<T> pipelineConfigurations,
+        String candidateModelId,
+        Function<T, Map<String, Object>> getConfigFunction
```
zane-neo, line 475: Is there a simple way to extract the configuration map from the responses, then change this method to accept Map<String, Object> as parameter?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -138,6 +139,11 @@ public String getDefaultType() {
         public String getDefaultVersion() {
             return null;
         }
+
+        @Override
+        public List<String> getRelatedModelIDKeyFields() {
```
zane-neo, line 144: A better name could be `getAllModelKeys`

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/RelatedModelIdHelper.java
```diff
@@ -0,0 +1,49 @@
...
+    private Map<String, List<String>> relatedModelIdMap;
+
+    public RelatedModelIdHelper(Map<String, Tool.Factory> ToolFactories) {
```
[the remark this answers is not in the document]
xinyual: Done

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Make it parallel.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
xinyual: Done

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Since now we're using parallel way to check agent/pipelines, do you think it's better to put them somewhere else? For me, these logic is tightly coupled with delete action so I prefer to keep them here.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: If we want to extract them before, we may need `instanceof` to check the type of input pipelineConfigurations.

yuye-aws:
@xinyual You can raise another PR in the skills repo. It's better we can review both PRs togeter.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java
```diff
@@ -144,5 +144,10 @@ public String getDefaultType() {
         public String getDefaultVersion() {
             return null;
         }
+
+        @Override
+        public List<String> getAllModelKeys() {
+            return List.of();
+        }
```
yuye-aws, line 151: ConnectorTool is using connector_id instead of model_id.

yuye-aws (commented):
Please update `DeleteConnectorTransportAction` with connector_id in ConnectorTool.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: @zane-neo Is there any other use case for AgentModelsSearcher? If not, I'm with @xinyual

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,49 @@
...
+    public SearchRequest constructQueryRequest(String candidateModelId) {
+        SearchRequest searchRequest = new SearchRequest(ML_AGENT_INDEX);
+        List<String> allKeyFields = collectAllKeys();
+        BoolQueryBuilder shouldQuery = QueryBuilders.boolQuery();
+        for (String keyField : allKeyFields) {
+            shouldQuery.should(QueryBuilders.termsQuery(TOOL_MODEL_RELATED_FIELD_PREFIX + keyField, candidateModelId));
+        }
+        searchRequest.source(new SearchSourceBuilder().query(shouldQuery));
+        return searchRequest;
+    }
```
yuye-aws, line 40: Have you investigated whether there is way to construct the query template beforehand? Then this function only needs to fill in the `candidateModelId`.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +262,130 @@ public void onFailure(Exception e) {
...
+            }
+        }, e -> {
+            countDownLatch.countDown();
+            noneBlocked.compareAndSet(true, false);
+            actionListener.onFailure(e);
+
+        });
+        checkAgentBeforeDeleteModel(modelId, countDownActionListener);
+        checkIngestPipelineBeforeDeleteModel(modelId, countDownActionListener);
+        checkSearchPipelineBeforeDeleteModel(modelId, countDownActionListener);
```
yuye-aws, line 386: Two questions:
1. Can you do these 3 operations in parallel?
2. If there is both an agent and an ingest pipeline depending on this model, will you only return the agents? It would better if you can return all the agents, ingest pipeline and search pipelines, so that the user would only meet with this error message once.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +262,130 @@ public void onFailure(Exception e) {
...
+            } else {
+                List<String> searchPipelineIds = getAllPipelineIds(
+                    searchPipelineResponse.pipelines(),
+                    org.opensearch.search.pipeline.PipelineConfiguration::getId
+                );
+                actionListener
+                    .onFailure(
+                        new OpenSearchStatusException(
+                            searchPipelineIds.size()
+                                + " search pipelines are still using this model, please delete or update the pipelines first: "
```
yuye-aws, line 351: nit: 1 search pipeline.
Also keep checkAgentBeforeDeleteModel and checkIngestPipelineBeforeDeleteModel consistent.

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
```diff
@@ -127,5 +129,7 @@ interface Factory<T extends Tool> {
          * @return the default tool version
          */
         String getDefaultVersion();
+
+        List<String> getAllModelKeys();
```
yuye-aws, line 133: Similarly, you can declare `List<String> getAllConnectorKeys();` for ConnectorTool.

xinyual:
> @xinyual You can raise another PR in the skills repo. It's better we can review both PRs togeter.

Sure.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java
xinyual: Yes. So we keep it empty.

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
yuye-aws, line 133: Since most of the tools will return an empty list, can you declare a default function?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: 1. Great work!
2. I mean, you can format the error message from 3 strings into 1. It's just low code effort but benefits many customers.

xinyual: For 2, the listener will multiply error message and I'm not sure whether we could queue them.

xinyual:
> Please update `DeleteConnectorTransportAction` with connector_id in ConnectorTool.

It's not related to this PR. I will check it and open another PR if it's need.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -84,25 +100,37 @@ public DeleteModelTransportAction(
...
+        this.agentModelsSearcher = agentModelsSearcher;
     }
 
     @Override
     protected void doExecute(Task task, ActionRequest request, ActionListener<DeleteResponse> actionListener) {
         MLModelDeleteRequest mlModelDeleteRequest = MLModelDeleteRequest.fromActionRequest(request);
         String modelId = mlModelDeleteRequest.getModelId();
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            ActionListener<DeleteResponse> wrappedListener = ActionListener.runBefore(actionListener, () -> context.restore());
+            checkDownstreamTaskBeforeDeleteModel(modelId, wrappedListener);
```
zane-neo, line 120: It's better to move this to right after the permission check passed as if a user doesn't have permission to delete the model, then we can save all these network calls to check usage.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: I see. You can use multiple listener. Once all of them has received a response or error, then concatenate the message.

xinyual: Done. Now concate the messages with comma, and it would be result like "**1 search pipelines are still using this model, please delete or update the pipelines first: [rag_pipeline],1 agents are still using this model, please delete or update the agents first: [EyoCQ5MBi0GstA7LbauT]"**

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Set;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+public class AgentModelsSearcher {
+    private Map<String, List<String>> relatedModelIdMap;
```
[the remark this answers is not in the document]
xinyual: Done. Change it to set.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Now I move it after we check whether user can access it. Please check it

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +483,46 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+        for (T pipelineConfiguration : pipelineConfigurations) {
+            Map<String, Object> config = getConfigFunction.apply(pipelineConfiguration);
+            if (searchThroughConfig(config, candidateModelId, "")) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private Boolean searchThroughConfig(Object searchCandidate, String candidateId, String prefixKey) {
```
zane-neo, line 500: Is that correct the prefixKey is the leaf target key looking for? And here it's a constant value `model_id`? If so, two suggestions are:
1. Make "model_id" a private static value with a meaningful name like `TARGET_MODEL_KEY`.
2. Change the prefixKey to a self explanation name like `targetModelKey`.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: Great! Please add 1 unit test.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -151,7 +169,8 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
                                                 )
                                             );
                                     } else if (isModelNotDeployed(mlModelState)) {
-                                        deleteModel(modelId, isHidden, actionListener);
+                                        checkDownstreamTaskBeforeDeleteModel(modelId, isHidden, actionListener);
+                                        ;
```
yuye-aws, line 174: Plz delete this extra ;

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +484,48 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
             && !mlModelState.equals(MLModelState.PARTIALLY_DEPLOYED);
     }
 
+    private <T> Boolean isPipelineContainsModel(
```
yuye-aws, line 487: Is there any need for the generic type? Seems that the only return type should be PipelineConfiguration by https://github.com/opensearch-project/OpenSearch/blob/ac443003bad13aeb6373f80439e154dcf25023a2/server/src/main/java/org/opensearch/action/ingest/GetPipelineResponse.java#L86-L88

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +484,48 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+
+    private Boolean searchThroughConfig(Object searchCandidate, String candidateId, String targetModelKey) {
+        Boolean flag = false;
+        if (searchCandidate instanceof String
+            && Objects.equals(targetModelKey, PIPELINE_TARGET_MODEL_KEY)
+            && Objects.equals(candidateId, searchCandidate)) {
+            return true;
+        } else if (searchCandidate instanceof List<?>) {
+            for (Object v : (List<?>) searchCandidate) {
+                flag = flag || searchThroughConfig(v, candidateId, targetModelKey);
```
yuye-aws, line 509: Can you early return false?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +484,48 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+            }
+        } else if (searchCandidate instanceof Map<?, ?>) {
+            for (Map.Entry<String, Object> entry : ((Map<String, Object>) searchCandidate).entrySet()) {
+                String key = entry.getKey();
+                Object value = entry.getValue();
+                flag = flag || searchThroughConfig(value, candidateId, key);
```
yuye-aws, line 515: Can you early return false?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +254,137 @@ public void onFailure(Exception e) {
...
+        }));
+    }
+
+    private void checkIngestPipelineBeforeDeleteModel(String modelId, ActionListener<Boolean> actionListener) {
+        GetPipelineRequest getPipelineRequest = new GetPipelineRequest();
+        client.execute(GetPipelineAction.INSTANCE, getPipelineRequest, ActionListener.wrap(ingestPipelineResponse -> {
+            if (!isPipelineContainsModel(
+                ingestPipelineResponse.pipelines(),
+                modelId,
+                org.opensearch.ingest.PipelineConfiguration::getConfigAsMap
```
yuye-aws, line 297: Since the return type of ingestPipelineResponse.pipelines() is List<PipelineConfiguration>, why not simply call pipelineConfiguration.getConfigAsMap()?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +254,137 @@ public void onFailure(Exception e) {
...
+            )) {
+                actionListener.onResponse(true);
+            } else {
+                List<String> searchPipelineIds = getAllPipelineIds(
+                    ingestPipelineResponse.pipelines(),
+                    org.opensearch.ingest.PipelineConfiguration::getId
```
yuye-aws, line 303: Same above, why not pipelineConfiguration.getId()?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +254,137 @@ public void onFailure(Exception e) {
...
+                );
```
yuye-aws, line 304: Seems that you are getting all the pipelines. How are you checking that not all the pipeline ids are included in the below error message?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +254,137 @@ public void onFailure(Exception e) {
...
+                searchPipelineResponse.pipelines(),
+                modelId,
+                org.opensearch.search.pipeline.PipelineConfiguration::getConfigAsMap
+            )) {
+                actionListener.onResponse(true);
+            } else {
+                List<String> searchPipelineIds = getAllPipelineIds(
+                    searchPipelineResponse.pipelines(),
+                    org.opensearch.search.pipeline.PipelineConfiguration::getId
+                );
```
yuye-aws, line 338: Same as the `checkIngestPipelineBeforeDeleteModel`. Why getting all the pipeline ids?

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +188,71 @@ public void testDeleteModel_Success() throws IOException {
...
+            ActionListener<GetSearchPipelineResponse> listener = invocation.getArgument(2);
+            listener.onResponse(getSearchPipelineResponse);
+            return null;
+        }).when(client).execute(eq(GetSearchPipelineAction.INSTANCE), any(), any());
+
+        deleteModelTransportAction.doExecute(null, mlModelDeleteRequest, actionListener);
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(argumentCaptor.capture());
+        assertEquals("1 search pipelines are still using this model, please delete or update the pipelines first: [1]", argumentCaptor.getValue().getMessage());
+    }
```
yuye-aws, line 206: In this unit test, can you create another search pipeline which is not using the model id. Also please update testDeleteModel_BlockedByIngestPipeline and testDeleteModel_BlockedByAgent accordingly

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +484,48 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+            Map<String, Object> config = getConfigFunction.apply(pipelineConfiguration);
+            if (searchThroughConfig(config, candidateModelId, "")) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private Boolean searchThroughConfig(Object searchCandidate, String candidateId, String targetModelKey) {
+        Boolean flag = false;
```
yuye-aws, line 502: Can you use primitive type?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +484,48 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+        if (searchCandidate instanceof String
+            && Objects.equals(targetModelKey, PIPELINE_TARGET_MODEL_KEY)
+            && Objects.equals(candidateId, searchCandidate)) {
+            return true;
+        } else if (searchCandidate instanceof List<?>) {
+            for (Object v : (List<?>) searchCandidate) {
+                flag = flag || searchThroughConfig(v, candidateId, targetModelKey);
+            }
+        } else if (searchCandidate instanceof Map<?, ?>) {
+            for (Map.Entry<String, Object> entry : ((Map<String, Object>) searchCandidate).entrySet()) {
```
yuye-aws, line 512: Please suppress warning

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Set;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+public class AgentModelsSearcher {
+    private Set<String> relatedModelIdSet;
```
yuye-aws, line 17: Plz make it final.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: No. ingestion Pipeline configure and search Pipeline configure is different type. https://github.com/opensearch-project/OpenSearch/blob/ac443003bad13aeb6373f80439e154dcf25023a2/server/src/main/java/org/opensearch/action/search/GetSearchPipelineResponse.java#L22

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: No. Ingestion and search pipeline are two different types.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Same above.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Good point. That's a bug here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AgentModelsSearcher(Map<String, Tool.Factory> toolFactories) {
```
yuye-aws, line 19: Raw use of factory type is not the best practice. Can you redeclare the type as `Tool.Factory<? extends Tool>`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
xinyual: Done.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +253,129 @@ public void onFailure(Exception e) {
...
+            log.error("Failed to delete ML Model: " + modelId, e);
+            actionListener.onFailure(e);
+
+        }));
+    }
+
+    private void checkIngestPipelineBeforeDeleteModel(String modelId, ActionListener<Boolean> actionListener) {
+        GetPipelineRequest getPipelineRequest = new GetPipelineRequest();
+        client.execute(GetPipelineAction.INSTANCE, getPipelineRequest, ActionListener.wrap(ingestPipelineResponse -> {
+            List<String> allRelevantPipelineIds = findRelevantPipelines(
```
yuye-aws, line 293: The word relevant is too ambiguous. I prefer dependent.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: I understand there're two classes sharing the same name. Can you take a look at the following example:
```
    // Method that accepts Object and checks the type at runtime
    public void processDate(Object date) {
        if (date instanceof java.util.Date) {
            System.out.println("Processing java.util.Date: " + date);
            // Additional processing for java.util.Date
        } else if (date instanceof java.sql.Date) {
            System.out.println("Processing java.sql.Date: " + date);
            // Additional processing for java.sql.Date
        } else {
            throw new IllegalArgumentException("Unsupported type: " + date.getClass().getName());
        }
    }
```

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +253,129 @@ public void onFailure(Exception e) {
         });
     }
 
+    private void checkAgentBeforeDeleteModel(String modelId, ActionListener<Boolean> actionListener) {
```
yuye-aws, line 256: We don't want to expose hidden agent ids to the user. Have you tested? What's your opinion @ylwu-amzn @dhrubo-os  @zane-neo ?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
yuye-aws: @zane-neo If so, do you think it would be better to refactor `findRelevantPipelines` into a separate class?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Done

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,47 @@
...
+        relatedModelIdSet = new HashSet<>();
+        for (Map.Entry<String, Tool.Factory> entry : toolFactories.entrySet()) {
+            Tool.Factory toolFactory = entry.getValue();
+            relatedModelIdSet.addAll(toolFactory.getAllModelKeys());
+        }
+    }
+
+    public SearchRequest constructQueryRequest(String candidateModelId) {
+        SearchRequest searchRequest = new SearchRequest(ML_AGENT_INDEX);
+        List<String> allKeyFields = collectAllKeys();
```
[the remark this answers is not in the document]
xinyual: Done

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Done

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Done.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Already add comment and change the logic

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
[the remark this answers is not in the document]
xinyual: Done.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Done.

brianf-aws (changes requested):
I like the direction this is going I guess mainly more javaDocs would be nice. Please see below some suggestions I gave.

I had some remarks about testing hopefully they make sense

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +189,113 @@ public void testDeleteModel_Success() throws IOException {
...
+        //org.opensearch.search.pipeline.PipelineConfiguration pipelineConfiguration = new PipelineConfiguration();
+        when(searchPipelineConfiguration.getId()).thenReturn("1");
+        when(searchPipelineConfiguration.getConfigAsMap()).thenReturn(configDataMap);
+
+        org.opensearch.search.pipeline.PipelineConfiguration irrelevantSearchPipelineConfiguration = mock(org.opensearch.search.pipeline.PipelineConfiguration.class);
+        Map<String, Object> irrelevantConfigMap = new HashMap<>();
+        irrelevantConfigMap.put("nothing", "nothing");
+        when(irrelevantSearchPipelineConfiguration.getConfigAsMap()).thenReturn(irrelevantConfigMap);
+        when(irrelevantSearchPipelineConfiguration.getId()).thenReturn("2");
+        when(getSearchPipelineResponse.pipelines()).thenReturn(List.of(searchPipelineConfiguration, irrelevantSearchPipelineConfiguration));
```
yuye-aws, line 230: Following the non-test code, change relevant to dependent

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +478,59 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+            if (searchThroughConfig(config, candidateModelId, "")) {
+                dependentPipelineConfigurations.add(getIdFunction.apply(pipelineConfiguration));
+            }
+        }
+        return dependentPipelineConfigurations;
+    }
+
+    // This method is to go through the pipeline configs and only when the key is model id and value is
+    // 1. String and equal to candidate id 2. A list of String containing candidate id We will return True. Otherwise False
+    private Boolean searchThroughConfig(Object searchCandidate, String candidateId, String targetModelKey) {
```
yuye-aws, line 499: targetModelKey is always an empty string. I do not think there is any need for this parameter

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +478,59 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+                }
+            } else if (currentCandidate instanceof List<?>) {
+                // Push all elements in the list onto the stack
+                for (Object v : (List<?>) currentCandidate) {
+                    stack.push(Pair.of(currentKey, v));
+                }
+            } else if (currentCandidate instanceof Map<?, ?>) {
+                // Push all values in the map onto the stack
+                for (Map.Entry<?, ?> entry : ((Map<?, ?>) currentCandidate).entrySet()) {
+                    String key = (String) entry.getKey();
```
yuye-aws, line 523: entry.getKey().toString() is safer

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
yuye-aws: Oh, I found that this change impacts `MLAgentExecutor` and other downstream classees. I don't want to make the scope too large for this PR. Maybe you can leave it to other PRs.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: My suggestion is to move all the searches to a separate class including agent search, pipeline search, but it's not mandatory as my explanation above.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: As it's hidden it could be an issue to expose the id, better filter out this.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +256,129 @@ public void onFailure(Exception e) {
...
+            actionListener.onFailure(e);
+
+        }));
+
+    }
+
+    private void checkDownstreamTaskBeforeDeleteModel(String modelId, Boolean isHidden, ActionListener<DeleteResponse> actionListener) {
+        CountDownLatch countDownLatch = new CountDownLatch(3);
+        AtomicBoolean noneBlocked = new AtomicBoolean(true);
+        List<String> errorMessages = new ArrayList<>();
```
zane-neo, line 356: Let's change this ArrayList to a thread safe implementation as three requests could complete at same time and writing error messages concurrently, which might cause concurrent issue.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +256,129 @@ public void onFailure(Exception e) {
...
+        ActionListener<Boolean> countDownActionListener = ActionListener.wrap(b -> {
```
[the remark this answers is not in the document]
zane-neo: Not quite following, what's the case that can't adapt to? Do you have an example?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
zane-neo: It's fine because we can assume the keys are all string, and if in edge case the key is null, (String) entry.getKey() won't throw exception but entry.getKey().toString() will, so let's keep it as is.

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +189,113 @@ public void testDeleteModel_Success() throws IOException {
...
+        deleteModelTransportAction.doExecute(null, mlModelDeleteRequest, actionListener);
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(argumentCaptor.capture());
+        assertEquals(
+            "1 ingest pipelines are still using this model, please delete or update the pipelines first: [1]",
+            argumentCaptor.getValue().getMessage()
+        );
+    }
+
+    public void testDeleteModel_BlockedByAgent() throws IOException {
```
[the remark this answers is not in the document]
xinyual: In each agent's request body, they will have different key field name. That's why we need AgentModelSearcher

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -656,4 +717,46 @@ private GetResponse buildResponse(MLModel mlModel) throws IOException {
...
+        }).when(client).execute(eq(GetPipelineAction.INSTANCE), any(), any());
+
+        when(getSearchPipelineResponse.pipelines()).thenReturn(List.of());
+        doAnswer(invocation -> {
+            ActionListener<GetSearchPipelineResponse> listener = invocation.getArgument(2);
+            listener.onResponse(getSearchPipelineResponse);
+            return null;
+        }).when(client).execute(eq(GetSearchPipelineAction.INSTANCE), any(), any());
+        configDataMap = Map
+            .of("single_model_id", "test_id", "list_model_id", List.of("test_id"), "test_map_id", Map.of("model_id", "test_id"));
```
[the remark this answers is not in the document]
xinyual: I add  two UTs, now list/single value/map will have their own UT.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Done.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +259,126 @@ public void onFailure(Exception e) {
...
+            actionListener.onFailure(e);
+
+        }));
+
+    }
+
+    private void checkDownstreamTaskBeforeDeleteModel(String modelId, Boolean isHidden, ActionListener<DeleteResponse> actionListener) {
+        CountDownLatch countDownLatch = new CountDownLatch(3);
+        AtomicBoolean noneBlocked = new AtomicBoolean(true);
+        CopyOnWriteArrayList<String> errorMessages = new CopyOnWriteArrayList<>();
```
zane-neo, line 357: It's good to use a thread safe vector here, but a suggestion is to change to ConcurrentLinkedQueue, this has better performance than CopyOnWriteArrayList as it doesn't need to copy the whole list when inserting new elements, and internally it's based on CAS so I believe it has best performance among all other thread safe list/queue.

xinyual: Great. Already change it.

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +189,113 @@ public void testDeleteModel_Success() throws IOException {
...
+        doAnswer(invocation -> {
+            ActionListener<GetPipelineResponse> listener = invocation.getArgument(2);
+            listener.onResponse(getIngestionPipelineResponse);
+            return null;
+        }).when(client).execute(eq(GetPipelineAction.INSTANCE), any(), any());
+
+        deleteModelTransportAction.doExecute(null, mlModelDeleteRequest, actionListener);
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(argumentCaptor.capture());
+        assertEquals("1 ingest pipelines are still using this model, please delete or update the pipelines first: [1],1 search pipelines are still using this model, please delete or update the pipelines first: [1]", argumentCaptor.getValue().getMessage());
```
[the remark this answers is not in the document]
xinyual: That's a point. Now I split it.

zane-neo (approved):

yuye-aws (approved):

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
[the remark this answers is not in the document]
brianf-aws: Hey @zane-neo dont have an example of the top of my head but this looks like it will probably be reused for a different resource later. Currently it does searching for agent, search and ingest pipeline; my thought process is that it could possibly do this for a different resource (when a new process is involved that uses model_id). Or if a different resource we have has a dependency like a connector id we could use this method similarly just a suggestion to think about.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +256,129 @@ public void onFailure(Exception e) {
...
+            if (countDownLatch.getCount() == 0) {
+                if (noneBlocked.get()) {
+                    deleteModel(modelId, isHidden, actionListener);
+                } else {
+                    actionListener.onFailure(new OpenSearchStatusException(String.join(",", errorMessages), RestStatus.CONFLICT));
+                }
+            }
+        }, e -> {
+            countDownLatch.countDown();
+            noneBlocked.compareAndSet(true, false);
```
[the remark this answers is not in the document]
xinyual: Yes. Change it since they're almost same performance but `set` is more readable.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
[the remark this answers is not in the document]
xinyual: Currently I prefer to keep it in this way since we don't have another example and we're not really sure about the generic scenario. If in the future we have another one, we could design a method for that.

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +193,205 @@ public void testDeleteModel_Success() throws IOException {
...
+        assertEquals("1 search pipelines are still using this model, please delete or update the pipelines first: [1]", argumentCaptor.getValue().getMessage());
+    }
+
+    public void testDeleteModel_BlockedBySearchPipelineSingleModelId() throws IOException {
+        Map<String, Object> configDataMapWithSingleModelId = Map
+            .of("model_id", "test_id", "list_model_id", List.of("test_id"), "test_map_id", Map.of("map_model_id", "test_id"));
+        when(searchPipelineConfiguration.getId()).thenReturn("search_1");
+        when(searchPipelineConfiguration.getConfigAsMap()).thenReturn(configDataMapWithSingleModelId);
+
+        org.opensearch.search.pipeline.PipelineConfiguration independentSearchPipelineConfiguration = mock(
```
brianf-aws, line 257: Is there a reason why we have to specify the full path?

plugin/src/test/java/org/opensearch/ml/action/models/DeleteModelTransportActionTests.java
```diff
@@ -159,20 +193,205 @@ public void testDeleteModel_Success() throws IOException {
...
+        deleteModelTransportAction.doExecute(null, mlModelDeleteRequest, actionListener);
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(argumentCaptor.capture());
+        assertEquals(
+            "1 ingest pipelines are still using this model, please delete or update the pipelines first: [ingest_1]",
+            argumentCaptor.getValue().getMessage()
+        );
+    }
+
+    public void testDeleteModel_BlockedByAgent() throws IOException {
```
brianf-aws, line 342: nit: Some spacing to break up the code would be good but not mandatory

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
```diff
@@ -127,5 +129,11 @@ interface Factory<T extends Tool> {
          * @return the default tool version
          */
         String getDefaultVersion();
+
+        /**
+         * Get model id related field names
+         * @return the list of all model id related field names
+         */
```
brianf-aws, line 136: This is my understanding of the interface
```
Retrieves Model Ids associated with the tool.
```

But I dont really know what you mean by @ return the list of all model id related field names

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +259,127 @@ public void onFailure(Exception e) {
...
+        }, e -> {
+            log.error("Failed to delete ML Model: " + modelId, e);
+            actionListener.onFailure(e);
+
+        }));
+
+    }
+
+    private void checkDownstreamTaskBeforeDeleteModel(String modelId, Boolean isHidden, ActionListener<DeleteResponse> actionListener) {
+        // Now 3 is: 1. Agent 2. Search pipeline 3. ingest pipeline
```
brianf-aws, line 355: How about something along the lines of checks 3 resources associated with with the model id 1. Agent 2. Search pipeline 3. ingest pipeline

plugin/src/test/java/org/opensearch/ml/plugin/DummyWrongTool.java
brianf-aws, line 1: What is this used for? I did a ctr f and couldnt find it being used

brianf-aws (commented):
Hey @xinyual I can tell you put a lot of effort into this 😄  , I think I understand what your code does. I think the icing on the cake would be a IT? Since that really checks how your code helps alleviate the issue you are solving.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+public class AgentModelsSearcher {
```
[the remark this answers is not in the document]
xinyual: It's not a Tool. Where do you want to put it? How about in ml-algorithms/src/main/java/org/opensearch/ml/engine/utlis?

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
xinyual: Otherwise, when a new Tool added, we cannot make sure all related model ID fields are detected. And that may skip the logic checking agent before deleting the model.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -81,6 +81,7 @@ public class CommonValue {
     public static final String ML_STOP_WORDS_INDEX = ".plugins-ml-stop-words";
     public static final Set<String> stopWordsIndices = ImmutableSet.of(".plugins-ml-stop-words");
     public static final Integer ML_MEMORY_MESSAGE_INDEX_SCHEMA_VERSION = 1;
+    public static final String TOOL_MODEL_RELATED_FIELD_PREFIX = "tools.parameters.";
```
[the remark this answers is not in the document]
xinyual: Already change it.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +259,127 @@ public void onFailure(Exception e) {
...
+                actionListener.onResponse(true);
+                return;
+            }
+            log.error("Failed to delete ML Model: " + modelId, e);
+            actionListener.onFailure(e);
+
+        }));
+    }
+
+    private void checkIngestPipelineBeforeDeleteModel(String modelId, ActionListener<Boolean> actionListener) {
```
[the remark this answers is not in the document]
xinyual: Already extract a common function for it. Please check it.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentModelsSearcher.java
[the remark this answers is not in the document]
xinyual: Already move to utlis

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +263,106 @@ public void onFailure(Exception e) {
...
+            log.error("Failed to delete ML Model: " + modelId, e);
+            actionListener.onFailure(e);
+
+        }));
+
+    }
+
+    private void checkDownstreamTaskBeforeDeleteModel(String modelId, Boolean isHidden, ActionListener<DeleteResponse> actionListener) {
+        // Now 3 is: 1. Agent 2. Search pipeline 3. ingest pipeline
+        CountDownLatch countDownLatch = new CountDownLatch(3);
```
dhrubo-os, line 339: Do we need to block the thread?

Can we use `CompletableFuture` instead of `CountDownLatch` to unblock the thread? 

A code example could be:

```
private void checkDownstreamTaskBeforeDeleteModel(String modelId, Boolean isHidden, ActionListener<DeleteResponse> actionListener) {
    AtomicBoolean failed = new AtomicBoolean(false);
    ConcurrentLinkedQueue<String> errorMessages = new ConcurrentLinkedQueue<>();

    CompletableFuture<Void> allChecks = CompletableFuture.allOf(
        runCheck(modelId, this::checkAgentBeforeDeleteModel, failed, errorMessages),
        runCheck(modelId, this::checkIngestPipelineBeforeDeleteModel, failed, errorMessages),
        runCheck(modelId, this::checkSearchPipelineBeforeDeleteModel, failed, errorMessages)
    );

    allChecks.thenRun(() -> {
        if (failed.get()) {
            actionListener.onFailure(new OpenSearchStatusException(String.join(". ", errorMessages), RestStatus.CONFLICT));
        } else {
            deleteModel(modelId, isHidden, actionListener);
        }
    });
}

private CompletableFuture<Void> runCheck(String modelId, BiConsumer<String, ActionListener<Boolean>> checkMethod, 
                                         AtomicBoolean failed, ConcurrentLinkedQueue<String> errorMessages) {
    CompletableFuture<Void> future = new CompletableFuture<>();
    if (failed.get()) {
        future.complete(null);
        return future;
    }

    checkMethod.accept(modelId, ActionListener.wrap(
        success -> {
            if (!success) {
                failed.set(true);
                errorMessages.add("Check failed for " + checkMethod.getClass().getSimpleName());
            }
            future.complete(null);
        },
        error -> {
            failed.set(true);
            errorMessages.add(error.getMessage());
            future.complete(null);
        }
    ));

    return future;
}
```

Warning: I didn't test this code though :)

xinyual: Since we use Actionlisener here, CountDownLatch won't block the thread and we don't need to await. And since we don't contact between each downstream thread, CountDownLatch is already enough here.

zane-neo (dismissed):

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +462,90 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+        for (T pipelineConfiguration : pipelineConfigurations) {
+            Map<String, Object> config = getConfigFunction.apply(pipelineConfiguration);
+            if (searchThroughConfig(config, candidateModelId)) {
+                dependentPipelineConfigurations.add(getIdFunction.apply(pipelineConfiguration));
+            }
+        }
+        return dependentPipelineConfigurations;
+    }
+
+    // This method is to go through the pipeline configs and he configuration is a map of string to objects.
```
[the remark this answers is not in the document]
xinyual: Done

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -333,6 +462,90 @@ private Boolean isModelNotDeployed(MLModelState mlModelState) {
...
+
+        while (!stack.isEmpty()) {
+            // Pop an item from the stack
+            Pair<String, Object> current = stack.pop();
+            String currentKey = current.getLeft();
+            Object currentCandidate = current.getRight();
+
+            if (currentCandidate instanceof String) {
+                // Check for a match
+                if (Objects.equals(currentKey, PIPELINE_TARGET_MODEL_KEY) && Objects.equals(candidateId, currentCandidate)) {
```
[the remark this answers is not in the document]
xinyual: Done

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/AgentModelsSearcher.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AgentModelsSearcher(Map<String, Tool.Factory> toolFactories) {
+        relatedModelIdSet = new HashSet<>();
+        for (Map.Entry<String, Tool.Factory> entry : toolFactories.entrySet()) {
+            Tool.Factory toolFactory = entry.getValue();
+            relatedModelIdSet.addAll(toolFactory.getAllModelKeys());
+        }
+    }
+
+    public SearchRequest constructQueryRequest(String candidateModelId) {
```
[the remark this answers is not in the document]
xinyual: Already add some.

plugin/src/test/java/org/opensearch/ml/plugin/DummyWrongTool.java
xinyual: This is for test.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Already changed.

spi/src/main/java/org/opensearch/ml/common/spi/tools/Tool.java
xinyual: I think it's like an attribute of Tool like version/name. Do you want to implement another class to default return the empty list in this function?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -32,7 +32,7 @@
  */
 @Log4j2
 @ToolAnnotation(AgentTool.TYPE)
-public class AgentTool implements Tool {
+public class AgentTool implements WithoutModelTool {
```
dhrubo-os, line 35: I would like to think the other way. Regardless of having model, any tool developer can extend the Tool interface. So let's not change the Tool interface. Instead we can extend the Tool interface to `ToolWithModel` which can have this extra method and the tools which aren't using models can implement Tool and the tools which are using models can implement `ToolWithModel` interface?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
dhrubo-os: I prefer `CompletableFuture` which is more cleaner, but upto you. 

In addition, can we fail fast, rather than waiting? If we see a model is being used in Agent, we don't need to wait to get result from Search or Index pipeline, right?

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
```diff
@@ -234,6 +262,106 @@ public void onFailure(Exception e) {
         });
     }
 
+    private void checkAgentBeforeDeleteModel(String modelId, ActionListener<Boolean> actionListener) {
```
dhrubo-os, line 265: The current method order doesn't align with the call hierarchy, making it harder to follow.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: It's a trade off. In current way, although we wait for all three downstream checking finishing, but in one call customer could collect messages from all downstream task, so he doesn't need to call it more times.

dhrubo-os: If there is a use case where customers want to know about all the resources where a model is being used, we can build a separate API for that purpose.

Please consider failing fast to reduce the latency of the delete API. The primary goal of the delete API is to delete the resource with minimal latency, not to inform customers about which resources are using the model (that is just a byproduct). Therefore, I don’t think there’s any question about trade-offs.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Sure. Already reorder the functions.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
xinyual: Sure, Please check it. Also with the logic in AgentModelSearcher to construct query.

plugin/src/main/java/org/opensearch/ml/action/models/DeleteModelTransportAction.java
xinyual: Got it. Now once one search process fails, it will return the result.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
xinyual: But I have concern. For this, we should make sure when a new tool with mlmodel is developed, it needs to implement  `ToolWithModel` but not `Tool`. It's a requirement for code review and it's more like a soft rule. So I'm afraid in the future it may skip this logic if we code reviewers miss this.

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/AgentModelsSearcher.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    /**
+     * Construct a should query to search all agent which containing candidate model Id
+    
+     @param candidateModelId the candidate model Id
+     @return a should search request towards agent index.
+     */
+    public SearchRequest constructQueryRequestToSearchModelIdInsideAgent(String candidateModelId) {
```
dhrubo-os, line 37: Can we filter out hidden agent here as well, as we are not doing anything with hidden model. So we should not search for hidden agent too.

xinyual: Sure. Let me change it.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -269,6 +269,8 @@ private MLCommonsSettings() {}
     public static final Setting<Boolean> ML_COMMONS_CONTROLLER_ENABLED = Setting
         .boolSetting("plugins.ml_commons.controller_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
 
+    public static final Setting<Boolean> ML_COMMONS_SAFE_DELETE_MODEL = Setting
```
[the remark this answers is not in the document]
xinyual: I already rename is to  ML_COMMONS_SAFE_DELETE_WITH_USAGE_CHECK

ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/AgentModelsSearcher.java
xinyual: I add must query when we create query in agentModelSearcher to make sure its is_hidden is false. Also remove  UT about hidden model, Please check it.

dhrubo-os: what if is_hidden field is null or empty, I meant the version when we didn't introduce hidden_model feature.

xinyual: Good point! Now I tune the DSL and if an agent doesn't have is_hidden field and use candidate model, we will also block it.

dhrubo-os (approved):

zane-neo (approved):

Issue #3238, opened 2024-11-26
[BUG] Sync up job not working as expected when upgrading cluster from prev versions to 2.17 


rbhavna:
**What is the bug?**
On 2.18, model auto redeploy feature has been removed for remote models in this [PR](https://github.com/opensearch-project/ml-commons/pull/2976) meaning after b/g or node/cluster restart, remote models wont be auto redeployed to new nodes automatically. 

When the same had been back-ported to AOS 2.17, a bug  was identified due to which ML Sync-up job would stop running automatically when a cluster is upgraded from 2.15 or any older versions to 2.17 on AOS. [This](https://github.com/opensearch-project/ml-commons/blame/main/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java#L282) condition returns directly if the model encountered by auto redeployer is a `remote` model and [sync-up job is never getting started](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java#L159) 

The sync-up is only starting again with manual intervention by changing the sync-up interval setting to a different value than the previous one(Default is 10 seconds)
```
PUT /_cluster/settings
{
  "persistent": {
    "plugins.ml_commons.sync_up_job_interval_in_seconds": 5
  }
}
````

**How can one reproduce the bug?**
1. On AOS, create a cluster on 2.15 version. Verify the sync_up job interval by getting the setting `plugins.ml_commons.sync_up_job_interval_in_seconds`
2. Check the logs to see the ML Sync-up logs at the intervals specified by above setting
3. Upgrade the cluster to 2.17 version
4. Check the logs again and you can see the sunc-up logs are missing

**What is the expected behavior?**
Sync-up job should run automatically when cluster is upgraded to 2.17 version

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #3241, opened 2024-11-27
fix for sync up job not working in 2.17 when upgraded from previous versions

rbhavna:
### Description
fix for sync up job not working in 2.17 when upgraded from previous versions

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3238

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
+++ b/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
@@ -186,6 +186,10 @@ public class MLModelAutoReDeployer {
                             modelAutoRedeployArrangements.add(modelAutoRedeployArrangement);
                     });
                 redeployAModel();
+            } else {
+                log.info("Could not find any models in the index, not performing auto reloading!");
+                startCronjobAndClearListener();
+                return;
             }
         }, e -> {
             if (e instanceof IndexNotFoundException) {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
```diff
@@ -261,6 +261,7 @@ private void queryRunningModels(ActionListener<SearchResponse> listener) {
     private void triggerModelRedeploy(ModelAutoRedeployArrangement modelAutoRedeployArrangement) {
         if (modelAutoRedeployArrangement == null) {
             log.info("No more models in arrangement, skipping the redeployment");
+            redeployAModel();
```
dhrubo-os, line 264: Why should we redeploy the model if `modelAutoRedeployArrangement` is null?

rbhavna: This method calls for redeploying the next model. Not the same one. Ideally if this is null, [this](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java#L157) condition should have been satisfied and triggerModelRedeploy shouldn't be called. Alternatively we can call startCronjobAndClearListener() here as well.

ylwu-amzn (dismissed):

plugin/src/main/java/org/opensearch/ml/autoredeploy/MLModelAutoReDeployer.java
zane-neo: Please change this to `startCronjobAndClearListener` method call.

zane-neo: Also, can you add an else to this [if](https://github.com/opensearch-project/ml-commons/pull/3241/files#diff-a81f529712d8f8f4b41ad5f5a2e7c7cbf8712211f26d04c91a4b208c1c1436b3L168) to ensure the cronjob started in edge cases, e.g. user deleted all the models in the index.

ylwu-amzn (approved):

Zhangxunmt (approved):

Issue #3182, opened 2024-10-29
[Enhancement] Change CatIndexTool implementation from _cat/index action to _list/index action


zane-neo:
**Is your feature request related to a problem?**

As `_cat/index` API is not performant in titan cluster and a new paginated `_list/index` API has been introduced recently: https://github.com/opensearch-project/OpenSearch/issues/14258. We need to review the implementation of [CatIndexTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java) as it uses the old [RestIndicesAction](https://github.com/opensearch-project/OpenSearch/blob/2.17/server/src/main/java/org/opensearch/rest/action/cat/RestIndicesAction.java) which could cause issue on titan clusters. In this [PR](https://github.com/opensearch-project/OpenSearch/pull/15986) a new cluster setting `cat.indices.response.limit.number_of_indices` is introduced to control the maximum indices number in `_cat/index` API, so CatIndexTool could throw exception in a titan cluster w/o modifying the implementation.

**What solution would you like?**
Replace the CatIndexTool implementation to `_list/index` approach.

**What alternatives have you considered?**
Still use `_cat/index` API and honer the cluster setting `cat.indices.response.limit.number_of_indices` to make sure no exception thrown during runtime, but the drawback is the tool may not able to fetch all the indices in a titan cluster since the amount of the indices might more than the setting's value.

**Do you have any additional context?**

dhrubo-os:
What do you mean by titan cluster?

ylwu-amzn:
Why not build another ListIndexTool? 
List index API introduced 2.18 https://opensearch.org/docs/latest/api-reference/list/list-indices/. Seems hard to backport fix/enhance of CatIndexTool to older versions by changing it to _list/indices ,

zane-neo:
> What do you mean by titan cluster?

Titan cluster means clusters that has a lot of number of nodes and on those clusters a `_cat/index` invocation might never return.

zane-neo:
> Why not build another ListIndexTool? List index API introduced in 2.18 https://opensearch.org/docs/latest/api-reference/list/list-indices/. Seems hard to backport fix/enhance of CatIndexTool to older versions by changing it to _list/indices ,

Agreed, I'll add a new tool instead of modifying the CatIndexTool.

PR #3243, opened 2024-11-29
Use _list/indices API instead of _cat/index API in CatIndexTool

zane-neo:
### Description
Use _list/indices API instead of _cat/index API in CatIndexTool

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3182

### Check List


plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Before
+    public void setUpCluster() throws Exception {
+        registerCatIndexFlowAgent();
+    }
+
+    private List<String> createIndices(int count) throws IOException {
+        List<String> indices = new ArrayList<>();
+        for (int i = 0; i < count; i++) {
+            String indexName = "test" + i;
```
dbwiddis, line 41: As mentioned, this "test" string plus an incrementing integer can be the prefix of other strings and be less useful for testing matching using substrings (`contains()` or `startswith()`).  If you're not using delimiters in the test it's important to make all of these unique substrings.  Is it possible to use a random string here?

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
```diff
@@ -0,0 +1,99 @@
...
+        String toolOutput = extractResult(responseStr);
+        String[] actualLines = toolOutput.split("\\n");
+        long testIndexCount = Arrays.stream(actualLines).filter(x -> x.contains("test")).count();
+        assert testIndexCount == indices.size();
+        for (String index : indices) {
+            assert Objects.requireNonNull(toolOutput).contains(index);
+        }
+    }
+
+    private String extractResult(String responseStr) {
```
dbwiddis, line 82: would be helpful for others maintaining this code to have an expected sample value of `responseStr` in comments here.  I can't really evaluate the code below without seeing the JSON I'm supposed to be parsing here.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
```diff
@@ -0,0 +1,99 @@
...
+public class CatIndexToolIT extends RestBaseAgentToolsIT {
+    private String agentId;
+    private final String question = "{\"parameters\":{\"question\":\"please help list all the index status in the current cluster?\"}}";
+
+    @Before
+    public void setUpCluster() throws Exception {
+        registerCatIndexFlowAgent();
+    }
+
+    private List<String> createIndices(int count) throws IOException {
```
dbwiddis, line 38: This is called multiple times in this test. How are we cleaning up the indices between tests?  Is that in the superclass?  Otherwise creating 10 will duplicate the first 10 of the 101 given the current setup.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
```diff
@@ -106,13 +119,13 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
         final String[] indices = indexList.toArray(Strings.EMPTY_ARRAY);
 
         final IndicesOptions indicesOptions = IndicesOptions.strictExpand();
-        final boolean local = parameters.containsKey("local") ? Boolean.parseBoolean("local") : false;
-        final TimeValue clusterManagerNodeTimeout = DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT;
+        final boolean local = parameters.containsKey("local") && Boolean.parseBoolean(parameters.get("local"));
```
dbwiddis, line 122: Thanks for fixing what was obviously a bug in the previous version. :)
```suggestion
        final boolean local = Boolean.parseBoolean(parameters.get("local"));
```
`parseBoolean(null)` returns false so you don't need to bother with checking the key existence.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
```diff
@@ -131,57 +144,150 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+        ClusterStateResponse clusterStateResponse,
+        GetSettingsResponse getSettingsResponse,
+        IndicesOptions subRequestIndicesOptions,
+        ActionListener<Table> originalListener
+    ) {
+        final ActionListener<PageToken> iterativeListener = ActionListener.wrap(r -> {
+            // when previous response returns, build next request with response and invoke again.
+            PageParams nextPageParams = new PageParams(r.getNextToken(), pageParams.getSort(), pageParams.getSize());
+            // when next page doesn't exist or reaches max supported page size, return.
+            if (r.getNextToken() == null || pageResults.size() >= MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE) {
```
dbwiddis, line 233: I may be missing something here but this comment is confusing.

I get that if the next page doesn't exist, we can return, but why return even if the next token is not null but we've hit the max?  This seems like we'd never hit the next page (since we would only have a next token in the case this page hit the max and had to paginate).

Also it seems the "max" could fit in one page.  Should this be `>` rather than `>=` here?  Or will we never get `>` and it could be `==` and we're just coding defensively?

This seems not right.  I don't see a unit or IT test at the edge case of `MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE` which is really necessary to properly test this.

dbwiddis (commented):
I don't see where we're properly testing the pagination here.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
```diff
@@ -0,0 +1,99 @@
...
+        String toolOutput = extractResult(responseStr);
+        String[] actualLines = toolOutput.split("\\n");
+        long testIndexCount = Arrays.stream(actualLines).filter(x -> x.contains("test")).count();
+        assert testIndexCount == indices.size();
+        for (String index : indices) {
+            assert Objects.requireNonNull(toolOutput).contains(index);
+        }
+    }
+
+    public void testCatIndexWithMoreThan100Indices() throws IOException {
```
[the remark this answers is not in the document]
zane-neo: The two test case indeed seems similar, for 10 indices one it's to test where pagination is not triggered and for 101 it's to test pagination is triggered. It's ok to change the numbers to `MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE ` and `MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE  + 1` but I think current code is enough to cover these cases and we can save a little bit time by not creating 100 indices.

Not sure why it's a good idea to combine tests together as we need to maintain the order between them and one test case change might impact another implicitly and developer might need to dig into this for more time to debug. 

For deleting indices, please check here: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/test/java/org/opensearch/ml/rest/RestBaseAgentToolsIT.java#L89

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
```diff
@@ -0,0 +1,99 @@
...
+    public void testCatIndexWithFewIndices() throws IOException {
+        List<String> indices = createIndices(10);
+        Response response = TestHelper.makeRequest(client(), "POST", "/_plugins/_ml/agents/" + agentId + "/_execute", null, question, null);
+        String responseStr = TestHelper.httpEntityToString(response.getEntity());
+        String toolOutput = extractResult(responseStr);
+        String[] actualLines = toolOutput.split("\\n");
+        long testIndexCount = Arrays.stream(actualLines).filter(x -> x.contains("test")).count();
+        assert testIndexCount == indices.size();
+        for (String index : indices) {
+            assert Objects.requireNonNull(toolOutput).contains(index);
```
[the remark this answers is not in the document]
zane-neo: I think this part is already covered by `assert testIndexCount == indices.size();` if result only contains `test10` without a `test1`, then the `testIndexCount` is not equals to indices.size().

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
zane-neo: Same above, I'm open to change to random string if you have a concrete example.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
zane-neo: please refer to: https://github.com/opensearch-project/ml-commons/pull/3243#discussion_r1868825815

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
zane-neo: This is from OS [core](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/action/list/RestIndicesListAction.java#L67-L69), it's limiting the max page size to the value of `MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE` means you can get at most this number of page results, so here I'm using `>=`, this can be changed to `==` but using `>=` for defensive.

ylwu-amzn:
Add a comment on the issue https://github.com/opensearch-project/ml-commons/issues/3182#issuecomment-2515304447
> Why not build another ListIndexTool?
List index API introduced in 2.18 https://opensearch.org/docs/latest/api-reference/list/list-indices/. Seems hard to backport fix/enhance of CatIndexTool to older versions by changing it to _list/indices ,

Any thoughts, @zane-neo , @dbwiddis ?

dbwiddis:
> Any thoughts, @zane-neo , @dbwiddis ?

Agreed a new tool would be cleaner.  `RestIndicesListAction` extends `RestIndicesAction` (that Cat Index uses) so it'd be better to logically extend the other tool with just the updated parameters.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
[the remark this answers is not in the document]
dbwiddis: I agree it's already covered; the point here is this line is a useless test (or at least not as useful as it could be) as it doesn't actually guarantee the thing being tested.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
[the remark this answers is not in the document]
dbwiddis: OK, got it, was looking at the wrong constant. Still I think you need to test the edge case of "this completely fills the first page of results but there is no second page".

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/CatIndexTool.java
dbwiddis: OK, now that I understand the purpose of this constant, this means you may have reached many pages but even if there are more past this max, you're going to stop.

plugin/src/test/java/org/opensearch/ml/tools/CatIndexToolIT.java
zane-neo: I was thinking an example could show that the current test can't cover, anyway it's not a big change so I can make this change.

dbwiddis: I already gave the example where you were testing for “test1” and your current test would have passed for “test10” or “teat100”. I’m not in favor of having tests that pass when they should fail. 

It’s just a suggestion.  Using an end of word delimiter would be equally effective.

zane-neo: changed to random string

zane-neo:
@dbwiddis @ylwu-amzn @dhrubo-os I've addressed your comments, can you review and approve? Thanks.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -0,0 +1,577 @@
...
+                return mlModelOutputs.get(0).getMlModelTensors().get(0).getDataAsMap().get("response");
+            }
+        };
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        // TODO: This logic exactly matches the OpenSearch _cat/indices REST action. If code at
+        // o.o.rest/action/cat/RestIndicesAction.java changes those changes need to be reflected here
+        // https://github.com/opensearch-project/ml-commons/pull/1582#issuecomment-1796962876
```
dbwiddis, line 114: This comment still references _cat/indices, but as it's now _list/indices as a subclass there should be a bit more clarity about the relationship.

dbwiddis (approved):
LGTM with a minor point on comment clarity!

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
zane-neo: Done

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -0,0 +1,576 @@
...
+        @SuppressWarnings("unchecked")
+        List<String> indexList = parameters.containsKey("indices")
+            ? gson.fromJson(parameters.get("indices"), List.class)
+            : Collections.emptyList();
+        final String[] indices = indexList.toArray(Strings.EMPTY_ARRAY);
+
+        final IndicesOptions indicesOptions = IndicesOptions.strictExpand();
+        final boolean local = parameters.containsKey("local") && Boolean.parseBoolean(parameters.get("local"));
+        final boolean includeUnloadedSegments = Boolean.parseBoolean(parameters.get("include_unloaded_segments"));
+        final PageParams pageParams = new PageParams(null, PageParams.PARAM_ASC_SORT_VALUE, DEFAULT_PAGE_SIZE);
```
dhrubo-os, line 123: can we take page size as a parameter input like:

```
final int pageSize = parameters.containsKey("page_size")
            ? Integer.parseInt(parameters.get("page_size"))
            : DEFAULT_PAGE_SIZE;

File: ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java [hunk 10]
```diff
@@ -0,0 +1,576 @@
...
+        ClusterStateResponse clusterStateResponse,
+        GetSettingsResponse getSettingsResponse,
+        IndicesOptions subRequestIndicesOptions,
+        ActionListener<Table> originalListener
+    ) {
+        final ActionListener<PageToken> iterativeListener = ActionListener.wrap(r -> {
+            // when previous response returns, build next request with response and invoke again.
+            PageParams nextPageParams = new PageParams(r.getNextToken(), pageParams.getSort(), pageParams.getSize());
+            // when next page doesn't exist or reaches max supported page size, return.
+            if (r.getNextToken() == null || pageResults.size() >= MAX_SUPPORTED_LIST_INDICES_PAGE_SIZE) {
```
dhrubo-os, line 232: Right now we are pulling pages one by one. Can we parallelize this one? Something like this maybe?


```
private void fetchPagesInParallel(
    String[] indices,
    boolean local,
    boolean includeUnloadedSegments,
    PageParams initialPageParams,
    ClusterStateResponse clusterStateResponse,
    GetSettingsResponse getSettingsResponse,
    ActionListener<Table> listener
) {
    // Thread-safe collection for storing page results
    Queue<Collection<ActionResponse>> pageResults = new ConcurrentLinkedQueue<>();

    // ExecutorService to manage parallel threads
    ExecutorService executor = Executors.newFixedThreadPool(5); // Adjust concurrency as needed

    // Start fetching pages
    fetchPageRecursively(indices, local, includeUnloadedSegments, initialPageParams, clusterStateResponse, getSettingsResponse, pageResults, executor)
        .whenComplete((ignored, ex) -> {
            executor.shutdown(); // Shutdown the thread pool
            if (ex != null) {
                log.error("Error during parallel page fetching", ex);
                listener.onFailure(ex);
            } else {
                Table table = buildTable(clusterStateResponse, getSettingsResponse, pageResults);
                listener.onResponse(table);
            }
        });
}

private CompletableFuture<Void> fetchPageRecursively(
    String[] indices,
    boolean local,
    boolean includeUnloadedSegments,
    PageParams currentPageParams,
    ClusterStateResponse clusterStateResponse,
    GetSettingsResponse getSettingsResponse,
    Queue<Collection<ActionResponse>> pageResults,
    ExecutorService executor
) {
    return CompletableFuture.runAsync(() -> {
        fetchSinglePage(
            indices,
            local,
            includeUnloadedSegments,
            currentPageParams,
            clusterStateResponse,
            getSettingsResponse,
            pageResults
        );
    }, executor).thenCompose(ignored -> {
        // Check if there's a next page to fetch
        PageToken nextToken = getNextPageToken(currentPageParams, pageResults);
        if (nextToken != null) {
            PageParams nextPageParams = new PageParams(nextToken, currentPageParams.getSort(), currentPageParams.getSize());
            return fetchPageRecursively(indices, local, includeUnloadedSegments, nextPageParams, clusterStateResponse, getSettingsResponse, pageResults, executor);
        } else {
            return CompletableFuture.completedFuture(null); // No more pages to fetch
        }
    });
}

private void fetchSinglePage(
    String[] indices,
    boolean local,
    boolean includeUnloadedSegments,
    PageParams pageParams,
    ClusterStateResponse clusterStateResponse,
    GetSettingsResponse getSettingsResponse,
    Queue<Collection<ActionResponse>> pageResults
) {
    try {
        GroupedActionListener<ActionResponse> groupedListener = createGroupedListener(
            pageResults,
            pageParams.getRequestedToken(),
            ActionListener.wrap(
                response -> log.info("Successfully fetched page {}", pageParams.getRequestedToken()),
                error -> log.error("Failed to fetch page {}", pageParams.getRequestedToken(), error)
            )
        );

        // Fetch index stats
        sendIndicesStatsRequest(indices, IndicesOptions.lenientExpandHidden(), includeUnloadedSegments, ActionListener.wrap(
            groupedListener::onResponse,
            groupedListener::onFailure
        ));

        // Fetch cluster health
        sendClusterHealthRequest(indices, IndicesOptions.lenientExpandHidden(), local, DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT, client, ActionListener.wrap(
            groupedListener::onResponse,
            groupedListener::onFailure
        ));
    } catch (Exception e) {
        log.error("Unexpected error while fetching a single page", e);
    }
}

private PageToken getNextPageToken(PageParams pageParams, Queue<Collection<ActionResponse>> pageResults) {
    // Determine the next token based on current page's results or null if no more pages
    // Replace this with actual logic to extract the next token
    return pageResults.isEmpty() ? null : new PageToken(); // Placeholder logic
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -0,0 +1,577 @@
...
+import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
```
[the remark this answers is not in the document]
zane-neo: Done

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
zane-neo: Done

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
zane-neo: No we can't as we need to get the page token from previous response and use it in next request.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -0,0 +1,580 @@
...
+import org.opensearch.ml.common.spi.tools.Tool;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@ToolAnnotation(ListIndexTool.TYPE)
+public class ListIndexTool implements Tool {
```
dhrubo-os, line 66: I don't see any corresponding Unit Test class for this class. Can we add one?

plugin/src/test/java/org/opensearch/ml/tools/ListIndexToolIT.java
```diff
@@ -0,0 +1,118 @@
...
+        String toolOutput = extractResult(responseStr);
+        String[] actualLines = toolOutput.split("\\n");
+        long testIndexCount = Arrays.stream(actualLines).filter(x -> x.contains("test")).count();
+        assert testIndexCount == indices.size();
+        for (String index : indices) {
+            assert Objects.requireNonNull(toolOutput).contains(index);
+        }
+    }
+
+    public void testCatIndexWithMoreThan100Indices() throws IOException {
```
dhrubo-os, line 71: why cat? Aren't we writing integ test for list here?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
dbwiddis: > Right now we are pulling pages one by one. Can we parallelize this one?

The whole purpose of using `_list` instead of `_cat` is to reduce CPU load on large clusters (hundreds of nodes) with a large number of indices (tens of thousands).   See 
 - https://github.com/opensearch-project/OpenSearch/issues/15954
 - https://github.com/opensearch-project/OpenSearch/pull/15986

Parallelizing would defeat the purpose of using `_list` -- just use `_cat` if you want the minimum latency but it comes at a cost of CPU.  The `_list` API intentionally trades away latency (for very large responses) in order to spread CPU out over a longer time period via pagination.

jngz-es:
Two high level comments.

1. Looks like finally we try to add a new `ListIndexTool`, so can change the title/description of issue according to the changes to avoid confusion when people reviewing issues?
2. I saw that this new `ListIndexTool` has the same default tool description with `CatIndexTool`. When users configure these two tools in an agent, I don't think LLM can distinguish between these two tools based on their descriptions. So from agent/LLM standard point, why do we need two tools having same functionality? If we cannot use them at the same time, why we need both of them? It's a little bit confusing for users about what the different use cases are for these two tools.

zane-neo:
> Two high level comments.
> 
> 1. Looks like finally we try to add a new `ListIndexTool`, so can change the title/description of issue according to the changes to avoid confusion when people reviewing issues?
> 2. I saw that this new `ListIndexTool` has the same default tool description with `CatIndexTool`. When users configure these two tools in an agent, I don't think LLM can distinguish between these two tools based on their descriptions. So from agent/LLM standard point, why do we need two tools having same functionality? If we cannot use them at the same time, why we need both of them? It's a little bit confusing for users about what the different use cases are for these two tools.

@ylwu-amzn @dbwiddis, I would like to discuss on the second point that @jngz-es pointed out, it's definitely correct that we shouldn't use same prompt for two different tools, there are two options to change this:
1. Change the CatIndexTool prompt to add more details about the limitation of this tool after OS 2.18 based on OS version.
cons: the default prompt is dynamic and user needs to read carefully the doc or the code to ensure the correctness of using this tool.
2. Change the CatIndexTool implementation to `_list/indices` API approach without adding the new ListIndexTool.
pros: user has consistent experience using this tool, we still make the tool's name and type CatIndexTool.

I believe you have concern on the second option that it can't be backport to older versions. But it seems backward compatible to change the CatIndexTool implementation to me, so please elaborate more on what scenarios we need to backport this. Also please put your thoughts here on which option you prefer, thanks.

zane-neo:
> As we are going to 3.x, however, I'm in favor of deprecating CatIndexTool and exclusively using ListIndexTool making this limitation more clear (and perhaps more user-configurable in the tool).

Yes, user can configure the page_size now in the tool and I think this is an option to deprecate CatIndexTool, but my thought is it's very easy to ensure BWC(not changing CatIndexTool's name and type only change the implementation to `_list/indices`), and if it's this easy, why not keep it? This is also the second option in my previous comment.

dbwiddis:
> not changing CatIndexTool's name 

Point being that there are two separate APIs with different behavior.
 - `_list/indices` is paginated by default and has REST-based params to control it, and won't fail. You can add a param to go up to the max limit.
 - `_cat/indices` will fail based on exceeding limits in cluster settings (see https://github.com/opensearch-project/OpenSearch/pull/15986) that the tool can't easily change, and are usually below the max limit.

To use a single tool:
 - we'd have to define a parameter (pagination size) that would go unused on `_cat` and be potentially confusing to users setting up the tool
 - for the case where we're below both the settings-based limit and pagination limit, they'd behave the same (other than not allowing any configuration)
 - for the case where we're above the settings-based limit, `_cat` would block, while `_list` might be successful depending on the pagination limit; otherwise would give partial results.  This is inconsistent behavior, particularly if the cluster setting is much lower than the desired pagination setting

So I still favor two tools:
 - CAT for customers who want a cluster-based hard limit to prevent AI tools from hammering their cluster, which blocks the call immediately and tells them to scope their request
 - LIST for customers who want to allow AI tools to use a higher limit than user-based CAT commands, favoring getting results (even partial results) rather than an error message.

jngz-es:
I agree with @zane-neo . From my point of view, `_list/indices` is more like an upgrade version of `_cat/indices`. We can use `_list/indices` instead of `_cat/indices` in all cases. I don't see much benefit or use cases for users to use both tools in a case. If not, we only need one tool to avoid confusing.

dbwiddis:
> in all cases. I don't see much benefit or use cases for users to use both tools in a case. If not, we only need one tool to avoid confusing.

Which is generally what I've been trying to say.

New List tool.
Deprecate Cat tool and eventually eliminate it (possibly with 3.0.0, honestly).

jngz-es:
Sounds good, thanks @dbwiddis . 

Hi @zane-neo , what do you think? Are you OK to deprecate catIndexTool?

zane-neo:
> Sounds good, thanks @dbwiddis .
> 
> Hi @zane-neo , what do you think? Are you OK to deprecate catIndexTool?

I'm open to both option, I'll make remove the CatIndexTool in this PR.

jngz-es (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -96,23 +110,25 @@ public Object parse(Object o) {
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        // TODO: This logic exactly matches the OpenSearch _cat/indices REST action. If code at
-        // o.o.rest/action/cat/RestIndicesAction.java changes those changes need to be reflected here
-        // https://github.com/opensearch-project/ml-commons/pull/1582#issuecomment-1796962876
+        // TODO: This logic exactly matches the OpenSearch _list/indices REST action. If code at
+        // o.o.rest/action/list/RestIndicesListAction.java changes those changes need to be reflected here
```
dhrubo-os, line 114: how are we planning to track changes of `o.o.rest/action/list/RestIndicesListAction.java` to reflect here?

dhrubo-os:
> Sounds good, thanks @dbwiddis .
> 
> Hi @zane-neo , what do you think? Are you OK to deprecate catIndexTool?

Sorry for the late response! I agree that we can deprecate catIndexTool, but I’d suggest following the standard deprecation process—announcing it first and then fully deprecating after 1–2 versions. Just want to make sure we give users enough time to adapt. What do you think?"

dbwiddis:
> I’d suggest following the standard deprecation process—announcing it first and then fully deprecating after 1–2 versions. Just want to make sure we give users enough time to adapt. What do you think?"

Given how long we are taking between major versions of OpenSearch, one version is probably enough!

We likely just should follow the same deprecation pattern as cat indices itself.   The key point here is that cat indices can be set to have a much lower response limit not in control of the LLM/tool, is more likely to fail without any response.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
zane-neo: It's not easy to keep tracking of that, maybe sometime we encounter error in the IT then we can realize that OS core has made changes that we should reflect, if the change is not critical then we may not find any issue with this tool, in this case I think it's fine not to track the changes. BTW, I have a plan to support dynamic tool in the future so that we don't need to duplicate the code of OpenSearch core or other plugins, you can check here: https://github.com/opensearch-project/skills/issues/459

zane-neo:
> ```
> 118 tests completed, 3 failed, 11 skipped
> Tests with failures:
>  - org.opensearch.ml.rest.RestBedRockInferenceIT.test_bedrock_embedding_v2_model_with_postProcessFunction
>  - org.opensearch.ml.rest.RestCohereInferenceIT.test_cohereInference_withDifferent_postProcessFunction
>  - org.opensearch.ml.rest.RestMLRemoteInferenceIT.testPredictWithAutoDeployAndTTL_RemoteModel
> ```

These tests are flaky and first two are introduced by me recently, I'm still trying to reproduce it and locate the root cause, let's merge this PR first, I'll keep tracking the first two failing ITs.

dhrubo-os:
> > ```
> > 118 tests completed, 3 failed, 11 skipped
> > Tests with failures:
> >  - org.opensearch.ml.rest.RestBedRockInferenceIT.test_bedrock_embedding_v2_model_with_postProcessFunction
> >  - org.opensearch.ml.rest.RestCohereInferenceIT.test_cohereInference_withDifferent_postProcessFunction
> >  - org.opensearch.ml.rest.RestMLRemoteInferenceIT.testPredictWithAutoDeployAndTTL_RemoteModel
> > ```
> 
> These tests are flaky and first two are introduced by me recently, I'm still trying to reproduce it and locate the root cause, let's merge this PR first, I'll keep tracking the first two failing ITs.

Sorry I'm not comfortable to merger a PR without passing any integration test suite. This is not a good practice IMO.

xinyual (approved):

jngz-es (approved):

.github/workflows/CI-workflow.yml
```diff
@@ -200,6 +195,24 @@ jobs:
           flags: ml-commons
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  Precommit-codecov:
```
dhrubo-os, line 198: Why did we add this?

This is failing all the time: https://github.com/opensearch-project/ml-commons/actions/runs/13911015452/job/38928296296

Please take a look @zane-neo

Issue #2993, opened 2024-09-26
[Enhancement] Enhance validation for create connector API


ylwu-amzn:
From https://github.com/opensearch-project/ml-commons/issues/2986, user found they can create connector with invalid input like a wrong `pre_process_function`. We should add more validation.

mingshl:
- check the mandatory fields in the preprocessing functions. 
- validate input and output

akolarkunnu:
please assign to me

dhrubo-os:
We can have more validations in few phases like:

1. Validate if connector payload has all the required fields. If not provided, throw the illegal argument exception
2. Then for pre-processing or post processing functions, if the LLM service uri is available url, we need to check if we are using the pre-processing function for the same llm service or not. As an example, for `"url": "https://${parameters.endpoint}/openai/deployments/${parameters.deploy-name}/embeddings?api-version=${parameters.api-version}"` pre-processing function needs to be: `"connector.pre_process.openai.embedding",` LLM service is openai here, so we need to use the pre-processing function of openai. Same for post processing function.

3. We can now perform more deeper level check. If we notice in the previous url: the model endpoint is `embeddings` so we are using the openai.embedding for pre-processing function. Same for post processing function.



We can raise PRs phase by phase accordingly.

akolarkunnu:
A comment from @mingshl  :
Just a following up thoughts, if we can check if the credentials is valid in the create API, that would be even better as a further enhancement coming next. Then we can help user ensure the credentials work before calling the model.

PR #3260, opened 2024-12-06
[Enhancement] Enhance validation for create connector API

### LINKED ISSUE #1382: [FEATURE]Improve test coverage for RemoteModel.java
We need to improve test coverage in our repo.

This issue will target only to improve test coverage for RemoteModel.java

From this url we can see where test coverage is missing for this class: https://app.codecov.io/gh/opensearch-project/ml-commons/blob/2.x/ml-algorithms%2Fsrc%2Fmain%2Fjava%2Forg%2Fopensearch%2Fml%2Fengine%2Falgorithms%2Fremote%2FRemoteModel.java

akolarkunnu:
### Description
This PR addresses the first part of this enhancement "Validate if connector payload has all the required fields. If not provided, throw the illegal argument exception". Validation of field credentials is added in this fix. All other validations already exist. Added new test cases corresponding to this validation, fixed all failing test cases because of this new validation and refactored some of the test case code.

### Related Issues
Partially Resolves #1382

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
~- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -55,13 +55,13 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         String postProcessFunction
     ) {
         if (actionType == null) {
-            throw new IllegalArgumentException("action type can't null");
+            throw new IllegalArgumentException("action type can't be null");
         }
         if (url == null) {
-            throw new IllegalArgumentException("url can't null");
+            throw new IllegalArgumentException("url can't be null");
         }
         if (method == null) {
-            throw new IllegalArgumentException("method can't null");
+            throw new IllegalArgumentException("method can't be null");
         }
         this.actionType = actionType;
         this.method = method;
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
@@ -93,6 +93,9 @@ public class MLCreateConnectorInput implements ToXContentObject, Writeable {
             if (protocol == null) {
                 throw new IllegalArgumentException("Connector protocol is null");
             }
+            if (credential == null || credential.isEmpty()) {
+                throw new IllegalArgumentException("Connector credential is null or empty list");
+            }
         }
         this.name = name;
         this.description = description;
--- a/common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
+++ b/common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.ml.common.connector;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 import static org.opensearch.ml.common.connector.ConnectorAction.ActionType.isValidActionInModelPrediction;
 
 import java.io.IOException;
@@ -12,10 +14,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.junit.Assert;
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.ExpectedException;
 import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.XContentType;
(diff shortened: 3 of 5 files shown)


common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
```diff
@@ -57,15 +59,28 @@ public void constructor_NullMethod() {
...
         String url = "https://test.com";
-        new ConnectorAction(actionType, method, url, null, null, null, null);
+        String requestBody = "{\"input\": \"${parameters.input}\"}";
+        new ConnectorAction(actionType, method, url, null, requestBody, null, null);
+    }
+
+    @Test
+    public void constructor_NullRequestBody() {
+        exceptionRule.expect(IllegalArgumentException.class);
+        exceptionRule.expectMessage("request body can't null");
```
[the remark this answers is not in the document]
akolarkunnu: ok, updated to assertThrows in all places of tests MLCreateConnectorInputTests and ConnectorActionTest(wherever it's applicable).

common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
```diff
@@ -103,12 +118,18 @@ public void toXContent_NullValue() throws IOException {
...
+        ConnectorAction action = new ConnectorAction(actionType, method, url, null, requestBody, null, null);
 
         XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
         action.toXContent(builder, ToXContent.EMPTY_PARAMS);
         String content = TestHelper.xContentBuilderToString(builder);
-        Assert.assertEquals("{\"action_type\":\"PREDICT\",\"method\":\"http\",\"url\":\"https://test.com\"}", content);
+        Assert
+            .assertEquals(
+                "{\"action_type\":\"PREDICT\",\"method\":\"http\",\"url\":\"https://test.com\","
+                    + "\"request_body\":\"{\\\"input\\\": \\\"${parameters.input}\\\"}\"}",
```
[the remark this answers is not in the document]
akolarkunnu: ok, updated all multi line strings to text blocks in the tests MLCreateConnectorInputTests and ConnectorActionTest.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -336,7 +336,7 @@ public static ConnectorAction createConnectorAction(Connector connector, Connect
 
         // Initialize the default method and requestBody
         String method = "POST";
-        String requestBody = null;
```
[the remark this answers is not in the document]
dhrubo-os: @rbhavna do you have any input here? In this PR, we are doing following in the create connector validation.

```
if (requestBody == null) {
            throw new IllegalArgumentException("request body can't null");
        }
        ```
        
     Could that be any issue for batch prediction?

File: common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java [hunk 4]
```diff
@@ -93,6 +93,15 @@ public MLCreateConnectorInput(
             if (protocol == null) {
                 throw new IllegalArgumentException("Connector protocol is null");
             }
+            if (description == null) {
```
dhrubo-os, line 96: I think description should not be a required field. May be we need to update the documentation from where you got this reference.

common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
```diff
@@ -27,49 +26,67 @@
...
     public void constructor_NullUrl() {
-        exceptionRule.expect(IllegalArgumentException.class);
-        exceptionRule.expectMessage("url can't null");
-        ConnectorAction.ActionType actionType = ConnectorAction.ActionType.PREDICT;
-        String method = "post";
-        String url = null;
-        new ConnectorAction(actionType, method, url, null, null, null, null);
+        Throwable exception = assertThrows(IllegalArgumentException.class, () -> {
+            ConnectorAction.ActionType actionType = ConnectorAction.ActionType.PREDICT;
+            String method = "post";
```
dhrubo-os, line 47:
```
// Shared test data for the class
    private static final ConnectorAction.ActionType TEST_ACTION_TYPE = ConnectorAction.ActionType.PREDICT;
    private static final String TEST_METHOD = "post";
    private static final String TEST_REQUEST_BODY = "{\"input\": \"${parameters.input}\"}";

```
And then we can reuse rather than defining in every lambda method?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -63,6 +63,9 @@ public ConnectorAction(
         if (method == null) {
             throw new IllegalArgumentException("method can't null");
```
dhrubo-os, line 64: `method can't be null`

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -63,6 +63,9 @@ public ConnectorAction(
...
         }
+        if (requestBody == null) {
+            throw new IllegalArgumentException("request body can't null");
```
dhrubo-os, line 67: `request body can't be null`

common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
```diff
@@ -27,49 +26,67 @@
...
+            String url = null;
+            String requestBody = "{\"input\": \"${parameters.input}\"}";
+            new ConnectorAction(actionType, method, url, null, requestBody, null, null);
+        });
+        assertEquals("url can't null", exception.getMessage());
```
dhrubo-os, line 52: let's make this error message to: `url can't be null`

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
akolarkunnu: Yes, I also had the same opinion while coding it. Since it is clearly marked is as required field in documentation, thought of discussing it on the review if someone raise the same concern. Similarly, I feel fields "parameters" and "credential" can also be null, if there is any parameters then user can configure otherwise can keep these fields as not configured. Also, let's wait for input from @rbhavna on "requestBody", that was intentionally kept as null and even explicit assertions for null were there in unit test cases. Same time documentation says, it's a required field.

I didn't add validation for backend_roles, access_mode, add_all_backend_roles though these were marked as required fields in documentation. These are inter related fields and corresponding validations exist in TransportCreateConnectorAction.validateRequest4AccessControl() and TransportCreateConnectorAction.validateSecurityDisabledOrConnectorAccessControlDisabled(). If none of these parameters configured we don't throw exception, but if one configure then others also has to configure properly(accordingly).

I referred the documentation - https://opensearch.org/docs/latest/ml-commons-plugin/remote-models/blueprints/#configuration-parameters

I wanted to discuss this before raising PR, but I thought it will be more easy to understand when reviewers see the code changes.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: Yes, I just followed existing error message text, will update it everywhere after we decide on above point.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
dhrubo-os: Yeah I agree parameters could be empty too. But not credentials. For connector, credentials are required. May be we can apply validation for Credentials but not for parameters?

dhrubo-os: Yeah, I think we add default role if role is not provided. So no need to have validation for roles as well. @rbhavna please correct me if I'm wrong.

rbhavna: @akolarkunnu Thanks for your contribution. And yes that is correct. We don't need any validation here for roles. We will validate them in transport layer after deciding if its security enabled or disabled cluster and accordingly validate the fields. So we can skip it here.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
rbhavna: @akolarkunnu @dhrubo-os Actually request body can be null sometimes. We don't pass in body for few requests like GET/CANCEL: https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/batch-predict/#prerequisites

Lets remove this null check here.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
rbhavna: As @dhrubo-os mentioned, `description` and `parameters` can be optional too. Looks like opensearch documentation needs to be updated here https://opensearch.org/docs/latest/ml-commons-plugin/remote-models/blueprints/#configuration-parameters

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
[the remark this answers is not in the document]
akolarkunnu: So removed requestBody null check according to @rbhavna ' s suggestion.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
akolarkunnu: ok, so new validation is added only for "credentials". All others removed.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: Removed the null check for "requestBody"

common/src/test/java/org/opensearch/ml/common/connector/ConnectorActionTest.java
akolarkunnu: ok, done

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: ok, done

mingshl:
LGTM, thanks for adding the null or empty checks for credentials!

Just a following up thoughts, if we can check if the credentials is valid in the create API, that would be even better as a further enhancement coming next. Then we can help user ensure the credentials work before calling the model.

mingshl (approved):

akolarkunnu:
> LGTM, thanks for adding the null or empty checks for credentials!
> 
> Just a following up thoughts, if we can check if the credentials is valid in the create API, that would be even better as a further enhancement coming next. Then we can help user ensure the credentials work before calling the model.

Ok, copied this to bug #2993 to include in the upcoming enhancement.

Issue #3276, opened 2024-12-13
[BUG] MLInferenceIngestProcessor has xContentRegistry as null 


brianf-aws:
**What is the bug?**
[Commit that introduces local models to MLInferenceIngestProcessor](https://github.com/opensearch-project/ml-commons/commit/7cd52915d04d8ac7ddb6e37a74a256603587ce69#diff-5fb7a6291568399efed0b4be65f70e91206634d3a1e5d1265243f1c204c0d81fR1011)

On creation of the Ingest processor the MLInferenceIngestProcessor has no valid [xContentRegistry](https://github.com/opensearch-project/ml-commons/blob/7cd52915d04d8ac7ddb6e37a74a256603587ce69/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L1011) which is used to create valid [MLInput](https://github.com/opensearch-project/ml-commons/blob/684627ac57f73c47e772d0cd9710932c3326bd39/common/src/main/java/org/opensearch/ml/common/input/MLInput.java#L233-L238) objects used for model prediction 

Contrast this with the [MLInferenceSearchResponseProcessor](https://github.com/opensearch-project/ml-commons/blob/c2a40c195402827fc5f113b865c12701f76fa60a/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L1035) That has a valid contentRegistry, this processor is able to correctly make objects and can produce valid model input.

If you want to test this locally I reccomend you [follow the steps](https://github.com/opensearch-project/ml-commons/pull/3258) to create a local embedding 

Then run the following 

```
PUT {{ _.domain }}/_ingest/pipeline/asymmetric_embedding_ingest_pipeline
{
	"description": "ingest passage text and generate a embedding using an asymmetric model",
	"processors": [
        {
	        "ml_inference": {
        
		        "model_input": "{\"text_docs\":[\"${input_map.text_docs}\"],\"target_response\":[\"sentence_embedding\"],\"parameters\":{\"content_type\":\"query\"}}",
		        "function_name": "text_embedding",
		        "model_id": "{{ _.model_id }}",
		        "input_map": [
			        {
				        "text_docs": "description"
			        }
		        ],
		        "output_map": [
			        {
				        "fact_embedding": "$.inference_results.*.output.*.data",
				        "embedding_size": "$.inference_results.*.output.*.shape[0]"
			        }
		        ]
	        }
        }
	]
}


[...]

PR #3281, opened 2024-12-16
Adds preset contentRegistry for IngestProcessors

brianf-aws:
### Background

Currently local models that use the parameters map within the payload to create a request can not create objects to be used for local model prediction, when using the MLInferenceIngestProcessor. 
e.g.
```json
## This is a cutom Asymmetric model prediction this passes
POST {{ _.domain }}/_plugins/_ml/_predict/text_embedding/{{ _.model_id }}
{
  "parameters": {
    "content_type": "query"
  },
  "text_docs": ["What day is it today?"],
}

## Ingest Pipeline configuration (Using the MLInferenceIngest Processor) that fails to create the MLInput needed for model prediciton

PUT {{ _.domain }}/_ingest/pipeline/asymmetric_embedding_ingest_pipeline
{
	"description": "ingest passage text and generate a embedding using an asymmetric model",
	"processors": [
		{
			"ml_inference": {

				"model_input": "{\"text_docs\":[\"${input_map.text_docs}\"],\"target_response\":[\"sentence_embedding\"],\"parameters\":{\"content_type\":\"query\"}}",
				"function_name": "text_embedding",
				"model_id": "{{ _.model_id }}",
				"input_map": [
					{
						"text_docs": "description"
					}
				],
				"output_map": [
					{
						"fact_embedding": "$.inference_results.*.output.*.data",
						"embedding_size": "$.inference_results.*.output.*.shape[0]"
					}
				]
			}
		}
	]
}
```

This requires a opensearch core change because it needs the contentRegistry,however given there is not much dependency on the registry (currently) we can give it the preset registry given in the MachineLearningPlugin class via the getNamedXContent() class to temporarily unblock this use case while a OpenSearch Core fix gives a proper change. 

### Context
Please see this issue https://github.com/opensearch-project/ml-commons/issues/3276. 

### Related Issues
Temporarily resolves https://github.com/opensearch-project/ml-commons/issues/3276 while waiting for a OpenSearch core fix.

### Check List
- [X] New functionality includes manual testing.

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
@@ -182,7 +182,10 @@ public class MLInferenceIngestProcessorTests extends OpenSearchTestCase {
 
             processor.execute(ingestDocument, handler);
             verify(handler)
-                .accept(isNull(), argThat(exception -> exception instanceof NullPointerException && exception.getMessage().equals(npeMessage)));
+                .accept(
+                    isNull(),
+                    argThat(exception -> exception instanceof NullPointerException && exception.getMessage().equals(npeMessage))
+                );
         } catch (Exception e) {
             assertEquals("this catch block should not get executed.", e.getMessage());
         }
--- a/plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
@@ -31,6 +31,7 @@ import org.opensearch.client.Client;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ingest.IngestDocument;
+import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.output.model.MLResultDataType;
@@ -136,6 +137,57 @@ public class MLInferenceIngestProcessorTests extends OpenSearchTestCase {
             assertEquals("this method should not get executed.", e.getMessage());
         }
 
+    }
+
+    /**
+     * Models that use the parameters field need to have a valid NamedXContentRegistry object to create valid MLInputs. For example
+     * <pre>
+     * PUT   /_plugins/_ml/_predict/text_embedding/model_id
+     *  {
+     *     "parameters": {
+     *         "content_type" : "query"
+     *     },
+     *     "text_docs" : ["what day is it today?"],
+     *     "target_response" : ["sentence_embedding"]
+     *   }
+     * </pre>
+     * These types of models like Local Asymmetric embedding models use the parameters field.
+     * And as such we need to test that having the contentRegistry throws an exception as it can not
+     * properly create a valid MLInput to perform prediction
+     *
+     * @implNote If you check the stack trace of the test you will see it tells you that it's a direct consequence of xContentRegistry being null
+     */
+    public void testExecute_xContentRegistryNullWithLocalModel_throwsException() throws Exception {
+        // Set the registry to null and reset after exiting the test
+        xContentRegistry = null;
+
+        String localModelInput =
+            "{ \"text_docs\": [\"What day is it today?\"],\"target_response\": [\"sentence_embedding\"], \"parameters\": { \"contentType\" : \"query\"} }";
+
+        MLInferenceIngestProcessor processor = createMLInferenceProcessor(
+            "local_model_id",
+            null,
+            null,
+            null,
+            false,
+            FunctionName.TEXT_EMBEDDING.toString(),
+            false,
+            false,
+            false,
+            localModelInput
+        );
+        try {
+            String npeMessage =
+                "Cannot invoke \"org.opensearch.ml.common.input.MLInput.setAlgorithm(org.opensearch.ml.common.FunctionName)\" because \"mlInput\" is null";
+
+            processor.execute(ingestDocument, handler);
+            verify(handler)
+                .accept(isNull(), argThat(exception -> exception instanceof NullPointerException && exception.getMessage().equals(npeMessage)));
+        } catch (Exception e) {
+            assertEquals("this catch block should not get executed.", e.getMessage());
+        }
+        // reset to mocked object
+        xContentRegistry = mock(NamedXContentRegistry.class);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


austintlee:
Can you create a test (maybe an IT test) that fails w/o this fix?

brianf-aws:
> Can you create a test (maybe an IT test) that fails w/o this fix?

Hey @austintlee, Im a bit confused do you mean create a test that shows this change works? If I create a test that fails. i.e. one that uses the `parameters: { ... }` it will pass since it now can create the MLInput properly. Also Im thinking if we should actually use the [asymmetric model ](https://huggingface.co/intfloat/multilingual-e5-small) as IT its 283 mb

mingshl:
@brianf-aws can you look up the usage of the parameters field and check what are the local model types that are impacted by this change? we know that asymmetric model type is impacted, wondering what other model types are impacted? 

then we can test the changes to all impacted local model scenarios.

mingshl:
I added the backport labels to make sure this bug fix can fix all the way back to when the local model is supported in ingest processors on June 11, https://github.com/opensearch-project/ml-commons/pull/2508 in 2.15, please make sure the bwc tests passed in the backport PRs. 

Also, would be nice if a second set of eyes can help double check the versions @brianf-aws https://github.com/opensearch-project/ml-commons/commits/2.15

brianf-aws:
> @brianf-aws can you look up the usage of the parameters field and check what are the local model types that are impacted by this change? we know that asymmetric model type is impacted, wondering what other model types are impacted?

Good call out Im seeing a pattern where the machineLearning plugin gets invoked by the Node class and gets `getNamedXContent()` e.g. [here](https://github.com/opensearch-project/OpenSearch/blob/35c366ddc794e0600184cf406c06ae65061e28ce/server/src/main/java/org/opensearch/node/Node.java#L785).
```java
        NamedXContentRegistry xContentRegistry = new NamedXContentRegistry(
            Stream.of(
                NetworkModule.getNamedXContents().stream(),
                IndicesModule.getNamedXContents().stream(),
                searchModule.getNamedXContents().stream(),
                pluginsService.filterPlugins(Plugin.class).stream().flatMap(p -> p.getNamedXContent().stream()),
                ClusterModule.getNamedXWriteables().stream()
            ).flatMap(Function.identity()).collect(toList())
        );

```
 So it will call the registries as specified here
https://github.com/opensearch-project/ml-commons/blob/67c562a76f8a010dfd9be7359022ab0070aba1dc/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L912-L930

If you click through every registry it will invoke a method that parses and collects fields from the parameter map.

Here is an example for [k-means](https://opensearch.org/docs/latest/ml-commons-plugin/algorithms/#k-means) [Albeit not what we think of when we think of a model], What this means is that the ML Processor has the ability to run any models.
https://github.com/opensearch-project/ml-commons/blob/684627ac57f73c47e772d0cd9710932c3326bd39/common/src/main/java/org/opensearch/ml/common/input/parameter/clustering/KMeansParams.java#L70-L88

dhrubo-os:
```
RestBedRockInferenceIT > test_bedrock_multimodal_model FAILED
    org.opensearch.client.ResponseException: method [POST], host [http://127.0.0.1:39501/], URI [/_plugins/_ml/models/_register], status line [HTTP/1.1 500 Internal Server Error]
    {"error":{"root_cause":[{"type":"null_pointer_exception","reason":"Cannot invoke \"org.opensearch.cluster.metadata.MappingMetadata.getSourceAsMap()\" because the return value of \"org.opensearch.cluster.metadata.IndexMetadata.mapping()\" is null"}],"type":"null_pointer_exception","reason":"Cannot invoke \"org.opensearch.cluster.metadata.MappingMetadata.getSourceAsMap()\" because the return value of \"org.opensearch.cluster.metadata.IndexMetadata.mapping()\" is null"},"status":500}
 ```

File: plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java [hunk 1]
```diff
@@ -138,6 +139,60 @@ public void testExecute_Exception() throws Exception {
...
+            null,
+            null,
+            false,
+            FunctionName.TEXT_EMBEDDING.toString(),
+            false,
+            false,
+            false,
+            localModelInput
+        );
+        try {
```
dhrubo-os, line 179:
```
// Act & Assert: Verify NullPointerException and its message
    NullPointerException exception = assertThrows(
        NullPointerException.class,
        () -> processor.execute(ingestDocument, handler),
        "Expected NullPointerException due to null xContentRegistry"
    );

    assertTrue(exception.getMessage().contains("Cannot invoke"),
               "Exception message should indicate a failure due to null mlInput");

```

What do you think about this?

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
```diff
@@ -138,6 +139,60 @@ public void testExecute_Exception() throws Exception {
...
+     *     "target_response" : ["sentence_embedding"]
+     *   }
+     * </pre>
+     * These types of models like Local Asymmetric embedding models use the parameters field.
+     * And as such we need to test that having the contentRegistry throws an exception as it can not
+     * properly create a valid MLInput to perform prediction
+     *
+     * @implNote If you check the stack trace of the test you will see it tells you that it's a direct consequence of xContentRegistry being null
+     */
+    public void testExecute_xContentRegistryNullWithLocalModel_throwsException() throws Exception {
```
dhrubo-os, line 160: Can't we add another test for success case with Asymmetric model?

brianf-aws: Yes its possible but that would involve hosting the model zip somewhere so we can test that it makes embeddings.

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
brianf-aws: I like this, but the problem is that the **exception is passed by the handler its not done by the method itself.** So this wouldn't be possible thats the reason why this class and more specifically this method has a catch to make sure that an exception is not possible. i.e. the handler passes an exception only.

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceIngestProcessorTests.java
dhrubo-os: Can we take some idea from here: https://github.com/opensearch-project/ml-commons/pull/2123/files#diff-cd8365f3263802b111604d62b751e5264a2e51df44b46b9fb650ca001466328a

brianf-aws: Oh! I didn't realize the community user did that, I can definitely use that thanks!

mingshl (approved):

PR #3289, opened 2024-12-18
Fix remote model with embedding input issue

### LINKED ISSUE #3261: [BUG] Bedrock multimodal model interface can't work with neural search
**What is the bug?**
Model interface for Bedrock multi-modal model can't work with neural search 

1. Create connector and model following this [blueprint](https://github.com/opensearch-project/ml-commons/blob/main/docs/remote_inference_blueprints/bedrock_connector_titan_multimodal_embedding_blueprint.md) 
2. Create ingest pipeline and try to ingest doc following this [doc](https://opensearch.org/docs/latest/search-plugins/multimodal-search/) , then can see such error when ingest doc

```
PUT /my-nlp-index/_doc/1
{
 "image_description": "Orange table"
}
```
error
```
{
  "error": {
    "root_cause": [
      {
        "type": "status_exception",
        "reason": "Error validating input schema: Validation failed: [$: required property 'parameters' not found] for instance: {\"algorithm\":\"REMOTE\",\"text_docs\":[\"Orange table\"],\"return_bytes\":false,\"return_number\":true,\"target_response\":[\"sentence_embedding\"]} with schema: {\n    \"type\": \"object\",\n    \"properties\": {\n        \"parameters\": {\n            \"type\": \"object\",\n            \"properties\": {\n                \"inputText\": {\n                    \"type\": \"string\"\n                },\n                \"inputImage\": {\n                    \"type\": \"string\"\n                }\n            }\n        }\n    },\n    \"required\": [\n        \"parameters\"\n    ]\n}"
      }
    ],
    "type": "status_exception",
    "reason": "Error validating input schema: Validation failed: [$: required property 'parameters' not found] for instance: {\"algorithm\":\"REMOTE\",\"text_docs\":[\"Orange table\"],\"return_bytes\":false,\"return_number\":true,\"target_response\":[\"sentence_embedding\"]} with schema: {\n    \"type\": \"object\",\n    \"properties\": {\n        \"parameters\": {\n            \"type\": \"object\",\n            \"properties\": {\n                \"inputText\": {\n                    \"type\": \"string\"\n                },\n                \"inputImage\": {\n                    \"type\": \"string\"\n                }\n            }\n        }\n    },\n    \"required\": [\n        \"parameters\"\n    ]\n}"
  },
  "status": 400
}
```

This is a break change. We should not break BWC by introducing model interface. 

 - OS: Linux
 - Version 2.17.1
 - Plugins

b4sjoo:
### Description
All remote embedding model will throw exception when use embedding input instead of remote input. However Neural Search uses a hardcoded value `FunctionName.TEXT_EMBEDDING` when instantiating the `MLInput`:
https://github.com/opensearch-project/neural-search/blob/7feacd67b3c7694ff4a1c1c2b430f2447a1ed4ab/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L293

Therefore, we removed the required string `parameter` in input interface to enable the dual usage of remote model. In the future, we plan to use different interface on different usage in same connector as a long term fix.

### Related Issues
Resolves #3261

### Check List

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/utils/MLNodeUtilsTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/utils/MLNodeUtilsTests.java
@@ -6,6 +6,9 @@
 package org.opensearch.ml.utils;
 
 import static java.util.Collections.emptyMap;
+import static org.opensearch.ml.common.utils.ModelInterfaceUtils.BEDROCK_COHERE_EMBED_ENGLISH_V3_MODEL_INTERFACE;
+import static org.opensearch.ml.common.utils.ModelInterfaceUtils.BEDROCK_TITAN_EMBED_MULTI_MODAL_V1_MODEL_INTERFACE;
+import static org.opensearch.ml.common.utils.ModelInterfaceUtils.BEDROCK_TITAN_EMBED_TEXT_V1_MODEL_INTERFACE;
 import static org.opensearch.ml.utils.TestHelper.ML_ROLE;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


dhrubo-os:
1. Can we have any unit test which can reflect this change?
2. Is there any documentation for model interface? Do we need to update the documentation too?

b4sjoo:
> 1. Can we have any unit test which can reflect this change?
> 2. Is there any documentation for model interface? Do we need to update the documentation too?

UT added. 
Interface doc: https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/register-model/#the-interface-parameter 
No need to added because we are trying to fix the interface back to the expected behavior.

jngz-es (approved):

plugin/src/test/java/org/opensearch/ml/utils/MLNodeUtilsTests.java
```diff
@@ -66,6 +66,31 @@ public void testValidateSchema() throws IOException {
         MLNodeUtils.validateSchema(schema, json);
     }
 
+    @Test
+    public void testValidateEmbeddingInputWithRemoteSchema() throws IOException {
+        String schema = "{\n"
```
ylwu-amzn, line 71: This unit test ensures this `schema` can work for text embedding. 
I think this is not enough, we should add unit test to cover the three model interfaces changed in this PR.

b4sjoo: Got it

plugin/src/test/java/org/opensearch/ml/utils/MLNodeUtilsTests.java
```diff
@@ -66,6 +66,149 @@ public void testValidateSchema() throws IOException {
...
+            + "        }\n"
+            + "    }\n"
+            + "}";
+        String json = "{\"parameters\": {\"texts\": [\"Hello\",\"world\"]}}";
+        MLNodeUtils.validateSchema(schema, json);
+    }
+
+    @Test
+    public void testValidateEmbeddingInputWithTitanTextRemoteSchema() throws IOException {
+        String schema = "{\n"
```
ylwu-amzn, line 121: Why not just use `TITAN_TEXT_EMBEDDING_MODEL_INTERFACE_INPUT`? 
So in future if someone changed `TITAN_TEXT_EMBEDDING_MODEL_INTERFACE_INPUT`, we can catch the error with this IT. 
If we define a new variable here and someone changed `TITAN_TEXT_EMBEDDING_MODEL_INTERFACE_INPUT` by mistake, how can we ensure this UT can catch that error ?

b4sjoo: Changed

ylwu-amzn (approved):

Zhangxunmt (approved):

Issue #2650, opened 2024-07-15
[META] Migrate feature/multi_tenancy branch to main


dbwiddis:
I will be starting the process of migrating the code developed on the [`feature/multi_tenancy`](https://github.com/opensearch-project/ml-commons/tree/feature/multi_tenancy) branch to `main`.  I'm creating this issue to establish context for reviewers and pre-answer some frequently asked questions. 

Important context to understand:

1. There are two main purposes of this branch/feature:
   - Abstract out the CRUD/S operations on system indices into a generic metadata storage option, that can use system indices, or a remote OpenSearch cluster, or DynamoDB, or (in the future) other storage options.
   - Add functionality to pass along a Tenant ID which will be used in the remote storage options to allow multiple tenants to use the same (stateless) plugin instance.

2. There will be new packages created in the `opensearch-ml-common` module:, `org.opensearch.sdk` containing a new metadata interface and associated request/response objects, and `org.opensearch.sdk.client` containing a default client implementation based on NodeClient, which implements existing behavior.
   - The location of this code in ML-Commons is temporary; it will eventually be migrated somewhere else (possibly a module on the core project, possibly a separate plugin, possibly the Java SDK project).  
   - The additional clients are presently in the `opensearch-ml-plugin` module `org.opensearch.ml.sdkclient` package. Like the above interfaces, this is not their final location.

3. Because of the way the `createComponents()` method on the `Plugin` interface works (binding those instances to their runtime class, not their interface), the client will be a concrete class which delegates method calls to an internal delegate instance.  This is not the ideal object-oriented structure, but seemed the best approach to simplify injection of the client into classes which are instantiated inside `createComponents()` and would otherwise not have access to an interface-bound instance created in a separate module.

4. The vast majority of changes across the codebase are copy-and-paste using the following general format (Get is shown, similar for Index, Update, Delete, and Search):

Old code (some variations on wrappers or anonymous subclasses, etc.):
```java
GetRequest getRequest = new GetRequest(<system index>).id(<doc id>).fetchSourceContext(fetchSourceContext);
client.get(getRequest, ActionListener.runBefore(ActionListener.wrap(r -> {
    // response code goes here
}, e -> {
    // exception code goes here
}), context::restore));
```

New code (some variations):
```java
GetDataObjectRequest getDataObjectRequest = GetDataObjectRequest
    .builder()
    .index(<system index>)
    .id(<doc id>)
    .tenantId(tenantId)
[...]

PR #3307, opened 2024-12-30
Primary setup for Multi-tenancy

dhrubo-os:
### Description

In this PR, I'm just adding the primary setup required for [Multi-tenancy feature](https://github.com/opensearch-project/ml-commons/tree/feature/multi_tenancy). After this PR, I'll start raising other PRs for adding multi-tenancy in connector, model, agents etc.

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/settings/SettingsChangeListener.java
+++ b/common/src/main/java/org/opensearch/ml/common/settings/SettingsChangeListener.java
@@ -1,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.settings;
+
+public interface SettingsChangeListener {
+    void onMultiTenancyEnabledChanged(boolean isEnabled);
+}
--- a/common/src/main/java/org/opensearch/ml/common/settings/SettingsChangeListener.java
+++ b/common/src/main/java/org/opensearch/ml/common/settings/SettingsChangeListener.java
@@ -5,6 +5,18 @@
 
 package org.opensearch.ml.common.settings;
 
+/**
+ * Interface for handling settings changes in the OpenSearch ML plugin.
+ */
 public interface SettingsChangeListener {
+    /**
+     * Callback method that gets triggered when the multi-tenancy setting changes.
+     *
+     * @param isEnabled A boolean value indicating the new state of the multi-tenancy setting:
+     *                  <ul>
+     *                    <li><code>true</code> if multi-tenancy is enabled</li>
+     *                    <li><code>false</code> if multi-tenancy is disabled</li>
+     *                  </ul>
+     */
     void onMultiTenancyEnabledChanged(boolean isEnabled);
 }
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -87,7 +87,9 @@ lombok {
 configurations.all {
     resolutionStrategy.force 'com.google.protobuf:protobuf-java:3.25.5'
     resolutionStrategy.force 'org.apache.commons:commons-compress:1.26.0'
+    resolutionStrategy.force 'software.amazon.awssdk:bom:2.29.12'
 }
+
 
 jacocoTestReport {
     reports {
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -57,7 +57,6 @@ dependencies {
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     // Multi-tenant SDK Client
     implementation "org.opensearch:opensearch-remote-metadata-sdk:${opensearch_version}"
-//    implementation "jakarta.json:jakarta.json-api:2.1.3"
 
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation("com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}")
@@ -359,7 +358,6 @@ configurations.all {
     resolutionStrategy.force 'org.slf4j:slf4j-api:1.7.36'
     resolutionStrategy.force 'org.codehaus.plexus:plexus-utils:3.3.0'
     resolutionStrategy.force 'org.eclipse.platform:org.eclipse.core.runtime:3.29.0'
-
     resolutionStrategy.force "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     resolutionStrategy.force "org.apache.httpcomponents.core5:httpcore5:5.3.1"
     resolutionStrategy.force "org.apache.httpcomponents.core5:httpcore5-h2:5.3.1"
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -55,6 +55,10 @@ dependencies {
 
     implementation group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
+    // Multi-tenant SDK Client
+    implementation "org.opensearch:opensearch-remote-metadata-sdk:${opensearch_version}"
+//    implementation "jakarta.json:jakarta.json-api:2.1.3"
+
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation("com.fasterxml.jackson.core:jackson-annotations:${versions.jackson}")
     implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}")
@@ -342,6 +346,7 @@ jacocoTestCoverageVerification {
 check.dependsOn jacocoTestCoverageVerification
 
 configurations.all {
+    exclude group: "org.jetbrains", module: "annotations"
     resolutionStrategy.force 'org.apache.commons:commons-lang3:3.10'
     resolutionStrategy.force 'commons-logging:commons-logging:1.2'
     resolutionStrategy.force 'org.objenesis:objenesis:3.2'
@@ -354,7 +359,16 @@ configurations.all {
     resolutionStrategy.force 'org.slf4j:slf4j-api:1.7.36'
     resolutionStrategy.force 'org.codehaus.plexus:plexus-utils:3.3.0'
     resolutionStrategy.force 'org.eclipse.platform:org.eclipse.core.runtime:3.29.0'
-    exclude group: "org.jetbrains", module: "annotations"
+
+    resolutionStrategy.force "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
+    resolutionStrategy.force "org.apache.httpcomponents.core5:httpcore5:5.3.1"
+    resolutionStrategy.force "org.apache.httpcomponents.core5:httpcore5-h2:5.3.1"
+    resolutionStrategy.force "org.apache.httpcomponents.client5:httpclient5:5.4.1"
+    resolutionStrategy.force "com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}"
+    resolutionStrategy.force "com.fasterxml.jackson.core:jackson-core:${versions.jackson_databind}"
+    resolutionStrategy.force "org.apache.logging.log4j:log4j-api:2.24.2"
+    resolutionStrategy.force "org.apache.logging.log4j:log4j-core:2.24.2"
+    resolutionStrategy.force "jakarta.json:jakarta.json-api:2.1.3"
 }
 
 apply plugin: 'com.netflix.nebula.ospackage'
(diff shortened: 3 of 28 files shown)


plugin/src/main/java/org/opensearch/ml/action/connector/GetConnectorTransportAction.java
```diff
@@ -77,6 +80,15 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLConn
                         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                         Connector mlConnector = Connector.createConnector(parser);
                         mlConnector.removeCredential();
+                        if (!Objects.equals(tenantId, mlConnector.getTenantId())) {
+                            actionListener
+                                .onFailure(
+                                    new OpenSearchStatusException(
+                                        "You don't have permission to access this connector",
```
mingshl, line 87: When permission denied, maybe also adding the tenant id and the connector id and name to exception message, so it helps the users to get proper permission.

mingshl:
@dhrubo-os do we have issues to explain this change?

dhrubo-os:
> @dhrubo-os do we have issues to explain this change?

RFC: https://github.com/opensearch-project/ml-commons/issues/2358

another issue: https://github.com/opensearch-project/ml-commons/issues/2650

plugin/src/main/java/org/opensearch/ml/action/connector/GetConnectorTransportAction.java
dhrubo-os: This is a Get connector method where customer provided the connector id. So we shouldn't expose any unnecessary information from where customer can take advantage.

plugin/build.gradle
```diff
@@ -55,6 +55,10 @@ dependencies {
 
     implementation group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
+    // Multi-tenant SDK Client
+    implementation "org.opensearch:opensearch-remote-metadata-sdk:${opensearch_version}"
+//    implementation "jakarta.json:jakarta.json-api:2.1.3"
```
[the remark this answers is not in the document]
dhrubo-os: done

plugin/build.gradle
```diff
@@ -354,7 +359,16 @@ configurations.all {
     resolutionStrategy.force 'org.slf4j:slf4j-api:1.7.36'
     resolutionStrategy.force 'org.codehaus.plexus:plexus-utils:3.3.0'
     resolutionStrategy.force 'org.eclipse.platform:org.eclipse.core.runtime:3.29.0'
-    exclude group: "org.jetbrains", module: "annotations"
+
```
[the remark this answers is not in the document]
dhrubo-os: done

common/src/main/java/org/opensearch/ml/common/settings/SettingsChangeListener.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.settings;
+
+public interface SettingsChangeListener {
```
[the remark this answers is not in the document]
dhrubo-os: done

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -480,6 +494,25 @@ public Collection<Object> createComponents(
         mlIndicesHandler = new MLIndicesHandler(clusterService, client);
         encryptor = new EncryptorImpl(clusterService, client, mlIndicesHandler);
 
+        SdkClient sdkClient = SdkClientFactory
+            .createSdkClient(
+                client,
+                xContentRegistry,
+                // Here we assume remote metadata client is only used with tenant awareness.
+                // This may change in the future allowing more options for this map
+                ML_COMMONS_MULTI_TENANCY_ENABLED.get(settings)
```
brianf-aws, line 503: This part is a little bit hard to follow, maybe we can extract it? I say this because you state you are going to add more in the future too.

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
```diff
@@ -0,0 +1,103 @@
...
+    public static boolean validateTenantResource(
+        MLFeatureEnabledSetting mlFeatureEnabledSetting,
+        String tenantIdFromRequest,
+        String tenantIdFromResource,
+        ActionListener<?> listener
+    ) {
+        if (mlFeatureEnabledSetting.isMultiTenancyEnabled() && !Objects.equals(tenantIdFromRequest, tenantIdFromResource)) {
+            listener.onFailure(new OpenSearchStatusException("You don't have permission to access this resource", RestStatus.FORBIDDEN));
+            return false;
+        } else
```
brianf-aws, line 61: same here

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
```diff
@@ -0,0 +1,103 @@
...
+            QueryBuilder queryBuilder = searchSourceBuilder.query();
+            if (queryBuilder instanceof TermQueryBuilder) {
+                TermQueryBuilder termQuery = (TermQueryBuilder) queryBuilder;
+                return TENANT_ID_FIELD.equals(termQuery.fieldName()); // Tenant filtering is enabled
+            }
+        }
+        return false; // Tenant filtering is not enabled
+    }
+
+    public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest restRequest) {
```
brianf-aws, line 77: There happens to be a lot of nesting maybe lets refactor to fail early ([gaurd clause](https://www.codementor.io/@clintwinter/use-guard-clauses-for-cleaner-code-1rrsczgwxp))? 

```java
 public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest restRequest) {
   if (!isMultiTenancyEnabled)
        return null;
    Map<String, List<String>> headers = restRequest.getHeaders();
  if (headers == null)
     throw new OpenSearchStatusException("Rest request header can't be null", RestStatus.FORBIDDEN);
  if (!headers.containsKey(Constants.TENANT_ID_HEADER))
      throw new OpenSearchStatusException("Tenant ID header is missing", RestStatus.FORBIDDEN);
  List<String> tenantIdList = headers.get(Constants.TENANT_ID_HEADER);
  
   if (tenantIdList == null || tenantIdList.isEmpty()) 
         throw new OpenSearchStatusException("Tenant ID header is missing", RestStatus.FORBIDDEN);
         
  String tenantId = tenantIdList.get(0);
  if (tenantId == null)
       throw new OpenSearchStatusException("Tenant ID can't be null", RestStatus.FORBIDDEN);
               
    return tenantId
}

```

My indentation is off but the point is to check your edge cases first and let it trickle down to the happy case

brianf-aws (changes requested):
I didnt get to reviewing the tests but left some comments hope they help :)

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
dhrubo-os: We are just building the SDKClient here. Similar like [this](https://github.com/opensearch-project/flow-framework/pull/980/files#diff-9d60d6086ce87e2240d7d6bcf8d2a8b79c73400fc6ec6865f74fc571e03bf1af)

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Validates the tenant ID based on the multi-tenancy feature setting.
+     *
+     * @param mlFeatureEnabledSetting The settings that indicate whether the multi-tenancy feature is enabled.
+     * @param tenantId The tenant ID to validate.
+     * @param listener The action listener to handle failure cases.
+     * @return true if the tenant ID is valid or if multi-tenancy is not enabled; false if the tenant ID is invalid and multi-tenancy is enabled.
+     */
+    public static boolean validateTenantId(MLFeatureEnabledSetting mlFeatureEnabledSetting, String tenantId, ActionListener<?> listener) {
+        if (mlFeatureEnabledSetting.isMultiTenancyEnabled() && tenantId == null) {
```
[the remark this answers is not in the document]
dhrubo-os: From the plugin level, we don't have visibility if the tenant id is valid or not. We perform this validation from the interceptor level. Here we just check if multi-tenancy is enabled then tenant id needs to be present.

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public static String getTenantID(Boolean isMultiTenancyEnabled, RestRequest restRequest) {
+        if (isMultiTenancyEnabled) {
+            Map<String, List<String>> headers = restRequest.getHeaders();
+            if (headers != null) {
+                if (headers.containsKey(Constants.TENANT_ID_HEADER)) {
+                    List<String> tenantIdList = headers.get(Constants.TENANT_ID_HEADER);
+                    if (tenantIdList != null && !tenantIdList.isEmpty()) {
+                        String tenantId = tenantIdList.get(0);
```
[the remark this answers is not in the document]
dhrubo-os: HTTP headers are represented as a Map<String, List<String>> because the HTTP protocol allows multiple values for a single header key (e.g., Set-Cookie). Even though most headers, like Tenant-ID, typically have only one value, the list structure ensures flexibility and compatibility. In practice, we usually retrieve the first value (list.get(0)) since only one value is needed in most cases.

plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java
```diff
@@ -263,4 +268,35 @@ private MLCommonsSettings() {}
...
+    // OPENSEARCH_JAVA_OPTS="-Dplugins.ml_commons.multi_tenancy_enabled=true" ./bin/opensearch
+    // 3. In the opensearch.yml configuration file:
+    // plugins.ml_commons.multi_tenancy_enabled: true
+    // After setting it, a full cluster restart is required for the changes to take effect.
+    public static final Setting<Boolean> ML_COMMONS_MULTI_TENANCY_ENABLED = Setting
+        .boolSetting("plugins.ml_commons.multi_tenancy_enabled", false, Setting.Property.NodeScope);
+
+    /** This setting sets the remote metadata type */
+    public static final Setting<String> REMOTE_METADATA_TYPE = Setting
+        .simpleString("plugins.ml_commons." + REMOTE_METADATA_TYPE_KEY, Setting.Property.NodeScope, Setting.Property.Final);
```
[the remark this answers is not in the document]
dhrubo-os: Nice!!

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
dhrubo-os: Yeah this is better.

dhrubo-os:
```
Tests with failures:
 - org.opensearch.ml.tools.VisualizationsToolIT.testVisualizationNotFound
 - org.opensearch.ml.tools.VisualizationsToolIT.testVisualizationFound
111 tests completed, 2 failed, 10 skipped
```


Same flaky errors

ml-algorithms/build.gradle
```diff
@@ -87,8 +87,10 @@ lombok {
 configurations.all {
     resolutionStrategy.force 'com.google.protobuf:protobuf-java:3.25.5'
     resolutionStrategy.force 'org.apache.commons:commons-compress:1.26.0'
+    resolutionStrategy.force 'software.amazon.awssdk:bom:2.29.12'
```
b4sjoo, line 90: Do you know why use a force here? If the version matter then you should be aware that it will cause version conflict when goes to the managed service.

dhrubo-os: refer to this: https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/25#issuecomment-2580952651

mingshl (approved):

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
dbwiddis: > if (headers == null)
>     throw new OpenSearchStatusException("Rest request header can't be null", RestStatus.FORBIDDEN);

FYI, this case is impossible.  I tried to write a unit test for the Flow Framework version of this util and could not (short of using reflection) make the headers null.  The only way to make it null is using a constructor, where it passes through a `headers.get()` to get the content type from the header; and if that didn't stop it, `Collections.unmodifiableMap()` [would throw an NPE on null headers](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/util/Collections.java#L1443-L1444) as well.

b4sjoo (approved):

plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java
dhrubo-os: Yeah, that's a good point! I agree but I prefer to still keep this as a defensive coding practice.

Issue #3285, opened 2024-12-17
[BUG] Model undeploying giving empty response


gaurav7830:
**What is the bug?**
When the model is in partially deployed state and undeployed is being called on that domain than the undeploy is giving emtpy response.

**How can one reproduce the bug?**
1. Create a domain.
2. Update the model state to partially deployed state.
3. Call model undeploy api.
4. Response will be empty.

**What is the expected behavior?**
Model should be in undeployed state after calling undeploy api.

**What is your host/environment?**
OS 2.17 version

**Do you have any screenshots?**
None

**Do you have any additional context?**
None

brianf-aws:
Hey Gaurav, Thank you for creating this issue I will look into it!

brianf-aws:
Hey @gaurav7830 Im trying to replicate locally. but I can not seem to have that empty response even when I purposely bring down a node during deployement it seems smart enough to give back correct results.

Im wondering if you have the `plugins.ml_commons.sync_up_job_interval_in_seconds: (number)` set on your cluster
https://opensearch.org/docs/latest/ml-commons-plugin/cluster-settings/#set-sync-job-intervals

Currently I tested with the sync up as "disabled". But please let me know what interval you are using so I can closely get as close.

brianf-aws:
You can get the scenario organically if do the following.

Suppose you have a OS cluster with nodes a,b,c,d

1. Deploy to all nodes a,b,c,d
2. Check the model status and also the profile API `GET {{ _.domain }}/_plugins/_ml/profile/models/{{ _.model_id }}` Notice that the model is deployed on all nodes worker_nodes : {a,b,c,d}
3. Bring down one node (Lets use node **a**). While this occurs make sure the sync up job is running for example 3 seconds should change state
4. After some time youll see that the model changes state to `PARTIALLY_DEPLOYED`
--  --
> At this point Ive theorized that the cache of node is volatile (i.e. It gets lost when the node gets shut down)

5. (Stop the sync up job) Bring back **a**, if you check the profile api youll notice that even though the the node is back the profile API states that only 3 nodes service the node.
-- --
> If you undeploy at this state you'll get back a valid response stating that it could undeploy from **{b,c,d}** and that node **a** had a not_found response
https://github.com/opensearch-project/ml-commons/blob/7209a10ae2021631234fc4bbe8101049145a6fcb/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java#L1872-L1884

[...]

PR #3380, opened 2025-01-11
Undeploy models with no WorkerNodes

brianf-aws:
This PR  aims to undeploy modelIds that have no nodes associated to them so as to keep the intention of undeploy truthful.

### Description
When performing undeploy if the model has no nodes associated to it then it will reset the index to UNDEPLOY status

Here is an example of why this code change is needed

This secnario is for the `PARTIALLY_DEPLOYED` issue.
1. Have nodes a,b,c,d in cluster associated with `modelID:@` i.e. peform deploy on it
2. Bring a,b down while having the syncup job running
3. By Now sync up will make this PARTIALLY_UNDEPLOYED
4. stop sync up
5. Bring other 2 c,d down and bring 2 nodes (these now have new ids 1,2)
6. bring the other two nodes back which have different ids so now the cluster has (1,2,3,4)
But the model index says `PARTIALLY_DEPLOYED` and no nodes are servicing

This code fix says. If no nodes are servicing this model then I need to set the index to UNDEPLOYED no matter if its already UNDEPLOYED or not.

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3285

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
@@ -228,22 +228,22 @@ public class TransportUndeployModelsAction extends HandledTransportAction<Action
         }
 
         bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-        log.info("No nodes service: {}", Arrays.toString(modelIds));
+        log.info("No nodes running these models: {}", Arrays.toString(modelIds));
 
         try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
             ActionListener<MLUndeployModelsResponse> listenerWithContextRestoration = ActionListener
                 .runBefore(listener, () -> threadContext.restore());
             ActionListener<BulkResponse> bulkResponseListener = ActionListener.wrap(br -> {
-                log.debug("Successfully set modelIds to UNDEPLOY in index");
+                log.debug("Successfully set the following modelId(s) to UNDEPLOY in index: {}", Arrays.toString(modelIds));
                 listenerWithContextRestoration.onResponse(new MLUndeployModelsResponse(response));
             }, e -> {
-                log.error("Failed to set modelIds to UNDEPLOY in index", e);
+                log.error("Failed to set the following modelId(s) to UNDEPLOY in index: {}", Arrays.toString(modelIds), e);
                 listenerWithContextRestoration.onFailure(e);
             });
 
             client.bulk(bulkUpdateRequest, bulkResponseListener);
         } catch (Exception e) {
-            log.error("Unexpected error while setting modelIds to UNDEPLOY status to index", e);
+            log.error("Unexpected error while setting the following modelId(s) to UNDEPLOY in index: {}", Arrays.toString(modelIds), e);
             listener.onFailure(e);
         }
 
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
@@ -7,6 +7,7 @@ package org.opensearch.ml.action.undeploy;
 
 import static org.opensearch.ml.common.CommonValue.ML_MODEL_INDEX;
 
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -14,10 +15,13 @@ import java.util.stream.Collectors;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
+import org.opensearch.action.bulk.BulkRequest;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.action.support.WriteRequest;
+import org.opensearch.action.update.UpdateRequest;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
@@ -33,6 +37,7 @@ import org.opensearch.index.query.TermQueryBuilder;
 import org.opensearch.index.query.TermsQueryBuilder;
 import org.opensearch.ml.cluster.DiscoveryNodeHelper;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.model.MLModelState;
 import org.opensearch.ml.common.transport.deploy.MLDeployModelRequest;
 import org.opensearch.ml.common.transport.undeploy.MLUndeployModelAction;
 import org.opensearch.ml.common.transport.undeploy.MLUndeployModelNodesRequest;
@@ -57,6 +62,7 @@ import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 
 import lombok.extern.log4j.Log4j2;
 
@@ -180,8 +186,34 @@ public class TransportUndeployModelsAction extends HandledTransportAction<Action
         mlUndeployModelNodesRequest.setTenantId(tenantId);
 
         client.execute(MLUndeployModelAction.INSTANCE, mlUndeployModelNodesRequest, ActionListener.wrap(r -> {
+            if (r.getNodes().isEmpty()) {
+                bulkSetModelIndexToUndeploy(modelIds);
+            }
             listener.onResponse(new MLUndeployModelsResponse(r));
         }, listener::onFailure));
+    }
+
+    private void bulkSetModelIndexToUndeploy(String[] modelIds) {
+        BulkRequest bulkUpdateRequest = new BulkRequest();
+        for (String modelId : modelIds) {
+            UpdateRequest updateRequest = new UpdateRequest();
+            Instant now = Instant.now();
+            ImmutableMap.Builder<String, Object> builder = ImmutableMap.builder();
+            builder.put(MLModel.MODEL_STATE_FIELD, MLModelState.UNDEPLOYED.name());
+
+            builder.put(MLModel.PLANNING_WORKER_NODES_FIELD, List.of());
+            builder.put(MLModel.PLANNING_WORKER_NODE_COUNT_FIELD, 0);
+
+            builder.put(MLModel.LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+            builder.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, 0);
+            updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(builder.build());
+            bulkUpdateRequest.add(updateRequest);
+        }
+        bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+        log.info("No models service: {}", modelIds.toString());
+        client.bulk(bulkUpdateRequest, ActionListener.wrap(br -> { log.debug("Successfully set modelIds to UNDEPLOY in index"); }, e -> {
+            log.error("Failed to set modelIds to UNDEPLOY in index", e);
+        }));
     }
 
     private void validateAccess(String modelId, String tenantId, ActionListener<Boolean> listener) {
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
@@ -217,7 +217,7 @@ public class TransportUndeployModelsAction extends HandledTransportAction<Action
         }
 
         bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
-        log.info("No nodes service: {}", modelIds.toString());
+        log.info("No nodes service: {}", Arrays.toString(modelIds));
 
         client.bulk(bulkUpdateRequest, ActionListener.wrap(br -> {
             log.debug("Successfully set modelIds to UNDEPLOY in index");
--- a/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
@@ -16,6 +16,7 @@ import org.opensearch.ExceptionsHelper;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.bulk.BulkRequest;
+import org.opensearch.action.bulk.BulkResponse;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.ActionFilters;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -157,10 +163,36 @@ private void undeployModels(String[] targetNodeIds, String[] modelIds, ActionLis
...
+
+            builder.put(MLModel.LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+            builder.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, 0);
+            updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(builder.build());
+            bulkUpdateRequest.add(updateRequest);
+        }
+        bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+        log.info("No models service: {}", modelIds.toString());
+        client.bulk(bulkUpdateRequest, ActionListener.wrap(br -> { log.debug("Successfully set modelIds to UNDEPLOY in index"); }, e -> {
+            log.error("Failed to set modelIds to UNDEPLOY in index", e);
```
ylwu-amzn, line 192: Should we send this exception back to client side ? If yes, we should pass listener to this method and add this line here
```
listener.onFailure(e);
```

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -157,10 +163,36 @@ private void undeployModels(String[] targetNodeIds, String[] modelIds, ActionLis
         MLUndeployModelNodesRequest mlUndeployModelNodesRequest = new MLUndeployModelNodesRequest(targetNodeIds, modelIds);
 
         client.execute(MLUndeployModelAction.INSTANCE, mlUndeployModelNodesRequest, ActionListener.wrap(r -> {
+            if (r.getNodes().isEmpty()) {
```
mingshl, line 166: it make sense that when execute undeploy model, the response return no worder nodes, then set the model to undeploy. 

but this doesn't fix the partially undeployed issue, right?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
dhrubo-os: +1

dhrubo-os:
Can we add unit test?

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
brianf-aws: `PARTIALLY_UNDEPLOYED` is a bit of a mixture of different scenarios one of them like so 
1. Have nodes a,b,c,d in cluster associated with `modelID:@` i.e. peform deploy on it
2. Bring a,b down while having the syncup job running
3. By Now sync up will make this PARTIALLY_UNDEPLOYED
4. stop sync up
5. Bring other 2 c,d down and bring 2 nodes (these now have new ids 1,2)
6. bring the other two nodes back which have different ids so now the cluster has (1,2,3,4)
But the model index says `PARTIALLY_DEPLOYED` and no nodes are servicing

This code fix says. If no nodes are servicing this model then I need to set the index to UNDEPLOYED no matter if its already UNDEPLOYED or not.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
brianf-aws: I'm not sure the user is concerned with the failure? The only way I see this as a problem if the model index does not exist and user does undeploy, this might cause a write issue.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -157,10 +163,36 @@ private void undeployModels(String[] targetNodeIds, String[] modelIds, ActionLis
...
+            builder.put(MLModel.PLANNING_WORKER_NODE_COUNT_FIELD, 0);
+
+            builder.put(MLModel.LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+            builder.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, 0);
+            updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(builder.build());
+            bulkUpdateRequest.add(updateRequest);
+        }
+        bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+        log.info("No models service: {}", modelIds.toString());
+        client.bulk(bulkUpdateRequest, ActionListener.wrap(br -> { log.debug("Successfully set modelIds to UNDEPLOY in index"); }, e -> {
```
Zhangxunmt, line 191: I think you can return MLUndeployModelsResponse including the original nodes rather than empty. You can find some examples in the tests how to create a  new MLUndeployModelsResponse.

ylwu-amzn: If we don't return in `br -> { log.debug("Successfully set modelIds to UNDEPLOY in index"); }`, it's possible that when client side receive the undeploy response, the model still on DEPLOYED state.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -157,10 +163,36 @@ private void undeployModels(String[] targetNodeIds, String[] modelIds, ActionLis
         MLUndeployModelNodesRequest mlUndeployModelNodesRequest = new MLUndeployModelNodesRequest(targetNodeIds, modelIds);
 
         client.execute(MLUndeployModelAction.INSTANCE, mlUndeployModelNodesRequest, ActionListener.wrap(r -> {
+            if (r.getNodes().isEmpty()) {
+                bulkSetModelIndexToUndeploy(modelIds);
```
ylwu-amzn, line 167:
```suggestion
                bulkSetModelIndexToUndeploy(modelIds, listener);
                return;
```

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
brianf-aws: Im not sure how I feel about creating a new one based on the failures,I think it will be misleading.  I will pass the original response r instead.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
brianf-aws: Thanks added the listener and response and returned early

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -156,11 +163,48 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUnde
...
+            builder.put(MLModel.LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+            builder.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, 0);
+            updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(builder.build());
+            bulkUpdateRequest.add(updateRequest);
+        }
+
+        bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+        log.info("No nodes service: {}", modelIds.toString());
+
+        client.bulk(bulkUpdateRequest, ActionListener.wrap(br -> {
```
ylwu-amzn, line 199: Wrap this with 
```
try (ThreadContext.StoredContext threadContext = client.threadPool().getThreadContext().stashContext()) {
```
Refer to https://github.com/opensearch-project/ml-commons/blob/main/memory/src/main/java/org/opensearch/ml/memory/index/ConversationMetaIndex.java#L161

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
Zhangxunmt: The failures should just return a message. But for the success case we should return something rather than {}

dhrubo-os: if success case shows:

```
{
"node id 1" : Not Found,
"node id 2" : Not Found
}
```

I don't think this will add much value from customer's POV.

brianf-aws:
> Can we add unit test?

Added 2 UTs for code fix
1. Check that the bulk write occurred when undeploy returned {}. This is a sign that the stale model index is `UNDEPLOYED`
2. Check that bulk write did not occur when some nodes have a response to the model, undeploy occured and changed index.

https://github.com/opensearch-project/ml-commons/blob/22b558d1f985d6de2c7b4713a6928fde73ea0030/plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelAction.java#L192-L195

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -156,11 +163,48 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUnde
     private void undeployModels(String[] targetNodeIds, String[] modelIds, ActionListener<MLUndeployModelsResponse> listener) {
         MLUndeployModelNodesRequest mlUndeployModelNodesRequest = new MLUndeployModelNodesRequest(targetNodeIds, modelIds);
 
-        client.execute(MLUndeployModelAction.INSTANCE, mlUndeployModelNodesRequest, ActionListener.wrap(r -> {
-            listener.onResponse(new MLUndeployModelsResponse(r));
+        client.execute(MLUndeployModelAction.INSTANCE, mlUndeployModelNodesRequest, ActionListener.wrap(response -> {
+            if (response.getNodes().isEmpty()) {
```
dhrubo-os, line 167: [nit]Let's add a detailed explanation here to ensure that future developers understand the reason behind this logic and aren't confused about why it's implemented this way.

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
rbhavna: This will still give empty response which is not accurate. Since we cannot send nodes as response in this case, lets send something to show model/models undeployed successfully. Something like

```
{
<model_id_1>: "UNDEPLOYED SUCCESSFULLY",
<model_id_2>: "UNDEPLOYED SUCCESSFULLY"
}
```

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
brianf-aws: Thanks for the insight, Just added this change

plugin/src/main/java/org/opensearch/ml/action/undeploy/TransportUndeployModelsAction.java
```diff
@@ -156,11 +164,68 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLUnde
...
+            builder.put(MLModel.PLANNING_WORKER_NODE_COUNT_FIELD, 0);
+
+            builder.put(MLModel.LAST_UPDATED_TIME_FIELD, Instant.now().toEpochMilli());
+            builder.put(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD, 0);
+            updateRequest.index(ML_MODEL_INDEX).id(modelId).doc(builder.build());
+            bulkUpdateRequest.add(updateRequest);
+        }
+
+        bulkUpdateRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+        log.info("No nodes service: {}", Arrays.toString(modelIds));
```
ylwu-amzn, line 208: How about clarifying more ?
```suggestion
        log.info("No nodes running these models: {}", Arrays.toString(modelIds));
```

brianf-aws: Valid feedback added to commit [77f6e5b](https://github.com/opensearch-project/ml-commons/pull/3380/commits/77f6e5bbb837191fe804ee26c5abd89a9942727e)

mingshl:
@brianf-aws you need to rebase first

mingshl (approved):

mingshl:
LGTM. I added 2.x backport label, did you figure out what other versions you need to backport to? I will add the labels for you

brianf-aws:
> LGTM. I added 2.x backport label, did you figure out what other versions you need to backport to? I will add the labels for you

Thank you! Yes I think we can do OS.2.11 - 2.18

jngz-es (approved):

brianf-aws:
Sorry can we make this to 2.x ? since this works of multi-tenancy and would break every other backport

brianf-aws:
Looks like an IT is causing an issue
```
RestBedRockInferenceIT > test_bedrock_multimodal_model_empty_imageInput_null_textInput STANDARD_OUT
    [2025-01-28T13:06:27,997][INFO ][o.o.m.r.RestBedRockInferenceIT] [test_bedrock_multimodal_model_empty_imageInput_null_textInput] before test
Error: 25-01-28T13:06:48,987][ERROR][o.o.m.r.MLCommonsRestTestCase] [test_bedrock_multimodal_model_empty_imageInput_null_textInput] method [POST], host [http://[::1]:38167], URI [/_plugins/_ml/_predict/TEXT_EMBEDDING/UFOGrpQB2p_oggmIDsWU], status line [HTTP/1.1 400 Bad Request]
    {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"No input text or image provided"}],"type":"illegal_argument_exception","reason":"No input text or image provided"},"status":400}
    org.opensearch.client.ResponseException: method [POST], host [http://[::1]:38167], URI [/_plugins/_ml/_predict/TEXT_EMBEDDING/UFOGrpQB2p_oggmIDsWU], status line [HTTP/1.1 400 Bad Request]
    {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"No input text or image provided"}],"type":"illegal_argument_exception","reason":"No input text or image provided"},"status":400}
    	at org.opensearch.client.RestClient.convertResponse(RestClient.java:501) ~[opensearch-rest-client-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
    	at org.opensearch.client.RestClient.performRequest(RestClient.java:384) ~[opensearch-rest-client-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
    	at org.opensearch.client.RestClient.performRequest(RestClient.java:359) ~[opensearch-rest-client-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
    	at org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:183) ~[test/:?]
    	at org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:156) ~[test/:?]
    	at org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:145) ~[test/:?]
    	at org.opensearch.ml.rest.MLCommonsRestTestCase.predictTextEmbeddingModel(MLCommonsRestTestCase.java:925) ~[test/:?]
    	at org.opensearch.ml.rest.RestBedRockInferenceIT.test_bedrock_multimodal_model_empty_imageInput_null_textInput(RestBedRockInferenceIT.java:233) ~[test/:?]
```

Is this a flaky known issue if not can we merge? cc: @jngz-es, @mingshl

PR #3423, opened 2025-01-23
run auto deploy remote model in partially deployed status

Zhangxunmt:
### Description
Currently the remote model auto-deploy only happens when the model is not deployed at all, by checking the running worker nodes == 0. But in some edge cases, we'd like to auto-deploy the model even it's in PARTIALLY_DEPLOYED status.  
The planning worker nodes are synced in the memory so when there is a partially deployed status, the auto deploy will apply.  

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -538,6 +538,21 @@ public class MLModelCacheHelper {
     }
 
     /**
+     * Get target worker nodes of model.
+     *
+     * @param modelId model id
+     * @return array of node id; return null if model not exists in cache
+     */
+    public String[] getTargetWorkerNodes(String modelId) {
+        MLModelCache modelCache = modelCaches.get(modelId);
+        if (modelCache == null) {
+            return null;
+        }
+        return modelCache.getTargetWorkerNodes();
+    }
+
+
+    /**
      * Add worker node of model.
      * 
      * @param modelId model id
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -632,7 +632,7 @@ public class MLModelCacheHelper {
         log.debug("sync model planning worker nodes");
         modelPlanningWorkerNodes.entrySet().forEach(entry -> {
             MLModelCache modelCache = getOrCreateModelCache(entry.getKey());
-            modelCache.syncPlanningWorkerNode(entry.getValue());
+            modelCache.syncPlanningWorkerNodes(entry.getValue());
         });
     }
 
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
@@ -551,7 +551,6 @@ public class MLModelCacheHelper {
         return modelCache.getTargetWorkerNodes();
     }
 
-
     /**
      * Add worker node of model.
      * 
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -537,6 +537,21 @@ public String[] getWorkerNodes(String modelId) {
...
+     *
+     * @param modelId model id
+     * @return array of node id; return null if model not exists in cache
+     */
+    public String[] getTargetWorkerNodes(String modelId) {
+        MLModelCache modelCache = modelCaches.get(modelId);
+        if (modelCache == null) {
+            return null;
+        }
+        return modelCache.getTargetWorkerNodes();
```
ylwu-amzn, line 551: We should consider deploy to all nodes case. 
If deploy to all nodes is try and target work nodes may be [node1, node2] on day1. 
Then day2, user add one more nodes to cluster, now the target work nodes should be [node1, node2, node3], so we can deploy to all nodes.
I think the work node from cache will still be [node1, node2], right ?

Zhangxunmt: From what I see the synup job only syncs up the worker nodes, but not the target worker nodes. This is another enhancement needed. Both the target worker nodes and running worker nodes needs to be up-to-date in the memory so it can cover all kinds of cases.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -2460,6 +2460,10 @@ public int getWorkerNodesSize(String modelId, FunctionName functionName) {
         return getWorkerNodes(modelId, functionName, false).length;
     }
 
+    public String[] getTargetWorkerNodes(String modelId) {
```
dhrubo-os, line 2463: Can we add java doc to the public method?

dhrubo-os:
Can we add tests?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -2460,6 +2460,10 @@ public int getWorkerNodesSize(String modelId, FunctionName functionName) {
...
+        return modelCacheHelper.getTargetWorkerNodes(modelId);
```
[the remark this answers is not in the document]
Zhangxunmt: The current GET functions in the modelCacheHelper returns null when there is a GET miss in the cache. So I think this is consistent in the code. Also I ensured that the only place calling this function has the null check.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
Zhangxunmt: planning worker nodes are synced in the new commit so I think this is not an issue anymore.

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -180,7 +180,9 @@ public void dispatchTask(
                 }
             }, listener::onFailure);
             String[] workerNodes = mlModelManager.getWorkerNodes(modelId, functionName, true);
-            if (workerNodes == null || workerNodes.length == 0) {
+            String[] targetWorkerNodes = mlModelManager.getTargetWorkerNodes(modelId);
+
+            if (requiresAutoDeployment(workerNodes, targetWorkerNodes)) {
```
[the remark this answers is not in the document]
Zhangxunmt: Good call out! But here are the things to consider 1) this is the call in the runtime for model prediction so less steps means less overheads. For most cases this syncUp shouldn't be needed so adding another syncUp layer for all model predictions may be expensive overall to justify it. 2) This syncUp is only for memory so it would cause inconsistency between memory and index before the next SyncUp job(which syncs for both).

In case the predict runs in between syncUpCronJobs and the nodes haven't been updated, the prediction will still proceed with the partially loaded nodes, and will sync up automatically to route to all nodes after the next SyncUp. So the impact is minimum to CX if not None. 

With that said, I think it's not justified to add another syncUp given all the overheads it brings to the system.

plugin/src/main/java/org/opensearch/ml/model/MLModelCacheHelper.java
```diff
@@ -609,6 +623,19 @@ public void syncWorkerNodes(Map<String, Set<String>> modelWorkerNodes) {
         });
     }
 
+    /**
+     * Sync planning worker nodes for all models.
+     *
+     * @param modelPlanningWorkerNodes planning worker nodes of all models
+     */
+    public void syncPlanningWorkerNodes(Map<String, Set<String>> modelPlanningWorkerNodes) {
+        log.debug("sync model planning worker nodes");
```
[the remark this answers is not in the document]
Zhangxunmt: This is a Map of <modelId: set of Nodes>, so showing this full matrix looks too much in the logs. Imagine what's looks like when there is hundreds of models and nodes in the domain. We can use the Profile API to check the memory to help debug.

b4sjoo (approved):

PR #3432, opened 2025-01-24
multi-tenancy + sdk client related changes in agents

dhrubo-os:
### Description
[multi-tenancy + sdk client related changes in agents]

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
@@ -5,30 +5,33 @@
 
 package org.opensearch.ml.engine.algorithms.agent;
 
+import static org.opensearch.common.xcontent.json.JsonXContent.jsonXContent;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_AGENT_INDEX;
 
-import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
+import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.ResourceNotFoundException;
-import org.opensearch.action.get.GetRequest;
+import org.opensearch.action.get.GetResponse;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
-import org.opensearch.common.xcontent.XContentHelper;
-import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.core.action.ActionListener;
-import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.common.Strings;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.IndexNotFoundException;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLAgentType;
 import org.opensearch.ml.common.agent.MLAgent;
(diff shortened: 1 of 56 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -70,26 +78,37 @@ public class MLAgentExecutor implements Executable {
...
         this.clusterService = clusterService;
         this.xContentRegistry = xContentRegistry;
         this.toolFactories = toolFactories;
         this.memoryFactoryMap = memoryFactoryMap;
+        this.isMultiTenancyEnabled = isMultiTenancyEnabled;
+    }
+
+    @Override
+    public void onMultiTenancyEnabledChanged(boolean isEnabled) {
+        this.isMultiTenancyEnabled = isEnabled;
```
jngz-es, line 111: I remember this setting is static as https://github.com/opensearch-project/ml-commons/blob/570edaf8ca48f08491706175755330bd24075143/plugin/src/main/java/org/opensearch/ml/settings/MLCommonsSettings.java#L305, are we expecting it can be changed dynamically?

jngz-es:
I have a high level question. It turns out we are to add a tenant id in MLToolSpec that means for every tool we will check the permission according to tenant id? From my point of view, we don't have any tools related APIs like create/delete, as we don't think of a tool as a resource. On the contrary, we consider the agent to be the resource for that we have create/delete APIs. So I am wondering if it is enough that we only have resource control on agents per tenant id.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
dhrubo-os: So I developed this piece of code when we are thinking to change tenancy dynamically. But for now, we are keeping it as a static settings. But still piece of code applied, it doesn't harm. If we in the long run want to turn it to dynamic settings  we won't need to do anything from our end.

dhrubo-os:
> I have a high level question. It turns out we are to add a tenant id in MLToolSpec that means for every tool we will check the permission according to tenant id? From my point of view, we don't have any tools related APIs like create/delete, as we don't think of a tool as a resource. On the contrary, we consider the agent to be the resource for that we have create/delete APIs. So I am wondering if it is enough that we only have resource control on agents per tenant id.

That a very good question. I agree with that. But the problem is, tool can make prediction to the model. But that model needs to be tenant specific. So which is why we are forwarding this tenant id everywhere. This way when tool wants to perform any operation we can check if this tool has enough permission to perform the operation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
dbwiddis: Pretty sure that as a static setting it will never trigger this method at all. And you have to actually register listeners to get these notifications and I don't see that done.  So it's half implemented. I agree it's harmless here but I'd rather see none or all.

dbwiddis (approved):
LGTM.   A lot more complicated than I thought!

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
dhrubo-os: That registration happened in this PR: https://github.com/opensearch-project/ml-commons/pull/3307/files

dbwiddis: Right but you need to add a settings update consumer in the class where you want it to update.  See https://github.com/search?q=repo%3Aopensearch-project%2Fml-commons%20addsettingsupdateconsumer&type=code 

But I never see us register multitenancy (we did at one time on the feature branch but I think I removed it): https://github.com/opensearch-project/ml-commons/blob/570edaf8ca48f08491706175755330bd24075143/plugin/src/main/java/org/opensearch/ml/settings/MLFeatureEnabledSetting.java#L46-L73

dhrubo-os: Yeah, we just need to add it back [here](https://github.com/opensearch-project/ml-commons/pull/2968/files#diff-c18df262e27ed8263e59dcbd7d1589294a7e2c8615b03248a19190d63193194c), that's all. So for now I'll keep the code as it is.

jngz-es (approved):

ylwu-amzn (approved):

PR #3443, opened 2025-01-28
[BACKPORT 2.x] applying multi-tenancy in search [model, model group, agent, connector] (#3433)

dhrubo-os:
[BACKPORT 2.x] applying multi-tenancy in search [model, model group, agent, connector] (#3433)

* applying multi-tenancy in search

* addressed comments

---------

### Description
[Describe what this change achieves]

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3459, opened 2025-01-29
changing MLSearchActionRequest to an instance subclass of SearchActionRequest

dhrubo-os:
In this PR, I'm making MLSearchActionRequest as a subclass of SearchActionRequest to follow the standard. Previously I wasn't quite extending the SearchActionRequest, as I had another SearchActionRequest object in the class and it was being read and written twice. 

### Description

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
@@ -25,7 +25,6 @@ import lombok.Getter;
  */
 @Getter
 public class MLSearchActionRequest extends SearchRequest {
-    SearchRequest searchRequest;
     String tenantId;
 
     /**
@@ -36,7 +35,7 @@ public class MLSearchActionRequest extends SearchRequest {
      */
     @Builder
     public MLSearchActionRequest(SearchRequest searchRequest, String tenantId) {
-        this.searchRequest = searchRequest;
+        super(searchRequest);
         this.tenantId = tenantId;
     }
 
@@ -49,10 +48,8 @@ public class MLSearchActionRequest extends SearchRequest {
     public MLSearchActionRequest(StreamInput input) throws IOException {
         super(input);
         Version streamInputVersion = input.getVersion();
-        if (input.readBoolean()) {
-            searchRequest = new SearchRequest(input);
-        }
         this.tenantId = streamInputVersion.onOrAfter(VERSION_2_19_0) ? input.readOptionalString() : null;
+
     }
 
     /**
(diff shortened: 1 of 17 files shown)


common/src/test/java/org/opensearch/ml/common/transport/search/MLSearchActionRequestTest.java
```diff
@@ -1,22 +1,15 @@
 package org.opensearch.ml.common.transport.search;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
+import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
dhrubo-os: addressed. That's in test file and we have a lot like this in ml-commons for test classes.

common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
```diff
@@ -36,7 +35,7 @@ public class MLSearchActionRequest extends SearchRequest {
      */
     @Builder
     public MLSearchActionRequest(SearchRequest searchRequest, String tenantId) {
-        this.searchRequest = searchRequest;
+        super(searchRequest);
```
arjunkumargiri, line 38: This constructs a new search request instead of reusing existing search request: https://github.com/opensearch-project/OpenSearch/blob/6f644e1c12de709b2136b999c8cb112a27c62100/server/src/main/java/org/opensearch/action/search/SearchRequest.java#L133

Would that be a memory concern?

arjunkumargiri (commented):
Add  details to github issues to provide description about why this PR is required, changes implemented as part of this PR and testing details.

common/src/main/java/org/opensearch/ml/common/transport/search/MLSearchActionRequest.java
dhrubo-os: Not sure how much memory concern will it be there. May be we can do something like:

```
// Use existing SearchRequest constructor to ensure minimal duplication
    super(searchRequest.indices(), searchRequest.source()); 
    this.routing(searchRequest.routing());
    this.preference(searchRequest.preference());
    this.scroll(searchRequest.scroll());
    this.allowPartialSearchResults(searchRequest.allowPartialSearchResults());
```


@msfroh do you have any input on this?

msfroh: Given that the `searchRequest.source()` (and pretty much everything else) is copied by reference, the added memory is pretty negligible.

Doing the math, I count 14 object references (each at most 8 bytes, depending on CompressedOOPS, if I recall correctly), 2 booleans (which I think are 1 byte each -- I don't think they get packed smaller than that), and 2 ints (4 bytes each). So, I think a rough estimate would be 122 bytes?

msfroh (approved):
This approach seems fine to me, since you've already made `MLSearchActionRequest` extend `SearchRequest`.

It's also valid to make `MLSearchActionRequest` extend `ActionRequest` directly and hold a reference to `SearchRequest`. Functionally, there's not much difference either way.

ylwu-amzn (approved):

Zhangxunmt (dismissed):

dhrubo-os:
No need add backport label, as this change has added to this PR in 2.x branch: https://github.com/opensearch-project/ml-commons/pull/3443/files

ylwu-amzn (approved):

jngz-es (approved):

PR #3474, opened 2025-01-30
addressing client changes due to adding tenantId in the apis

dhrubo-os:
### Description
[addressing client changes due to adding tenantId in the apis]

### Related Issues

### Check List

### DIFF
--- a/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
+++ b/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
@@ -237,7 +237,7 @@ public interface MachineLearningClient {
      */
     default ActionFuture<DeleteResponse> deleteTask(String taskId) {
         PlainActionFuture<DeleteResponse> actionFuture = PlainActionFuture.newFuture();
-        deleteModel(taskId, actionFuture);
+        deleteTask(taskId, actionFuture);
         return actionFuture;
     }
 
@@ -361,7 +361,7 @@ public interface MachineLearningClient {
      * Undeploy model
      * For additional info on deploy, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/undeploy-model/
      * @param modelIds the model ids
-     * @param modelIds the node ids. May be null for all nodes.
+     * @param nodeIds the node ids. May be null for all nodes.
      * @param listener a listener to be notified of the result
      */
     default void undeploy(String[] modelIds, String[] nodeIds, ActionListener<MLUndeployModelsResponse> listener) {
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -480,8 +480,7 @@ default ActionFuture<DeleteResponse> deleteAgent(String agentId) {
      * @param listener a listener to be notified of the result
      */
     default void deleteAgent(String agentId, ActionListener<DeleteResponse> listener) {
-        PlainActionFuture<DeleteResponse> actionFuture = PlainActionFuture.newFuture();
-        deleteAgent(agentId, null, actionFuture);
+        deleteAgent(agentId, null, listener);
```
Zhangxunmt, line 483: will it trigger BWC issue by using a listener?

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
```diff
@@ -361,7 +361,7 @@ default ActionFuture<MLUndeployModelsResponse> undeploy(String[] modelIds, @Null
      * Undeploy model
      * For additional info on deploy, refer: https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/undeploy-model/
      * @param modelIds the model ids
-     * @param modelIds the node ids. May be null for all nodes.
+     * @param nodeIds the node ids. May be null for all nodes.
```
Zhangxunmt, line 364: "May be null for all nodes" - this sounds not a description but a suggestion. Why not just put "Use null for all nodes" if you are fully sure.

Zhangxunmt (commented):
Are these change only applying to new client APIs? Can you double check any BWC issues due to this client change?

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
dhrubo-os: I didn't change anything in the description. It was like that before. I just fixed `nodeIds`, before it was `modelIds`

dhrubo-os:
> Are these change only applying to new client APIs? Can you double check any BWC issues due to this client change?

We already tested it before with other apis like for model (which is mostly being used in other dependencies like skills/FF). If the client get the updated client from their maven snapshot, then it won't cause any issue, but before that it might cause some integ failure. But maven snapshot updates daily, so we should be fine here.

client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java
dhrubo-os: nope. Please check other apis.

b4sjoo (approved):

Zhangxunmt (approved):

PR #3492, opened 2025-02-03
Run the Get Connector request in a stashed threadcontext

cwperks:
### Description

This PR fixes CI issues seen in the skills repo when testing with security. 

ref: https://github.com/opensearch-project/skills/actions/runs/13076372692/job/36489480535

This PR surrounds the call to the `.plugins-ml-connector` index in a stashed context to ensure the action runs in a trusted system context.

The security plugin has logic to scrub results when searching on system indices or getting individual documents for regular users. Plugins must surround calls to system indices to run in the trusted system context.

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -2077,40 +2077,43 @@ public class MLModelManager {
             .tenantId(tenantId)
             .build();
 
-        sdkClient.getDataObjectAsync(getDataObjectRequest).whenComplete((r, throwable) -> {
-            log.debug("Completed Get Connector Request, id:{}", connectorId);
-            if (throwable != null) {
-                Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
-                if (ExceptionsHelper.unwrap(cause, IndexNotFoundException.class) != null) {
-                    log.error("Failed to get connector index", cause);
-                    listener.onFailure(new OpenSearchStatusException("Failed to find connector", RestStatus.NOT_FOUND));
-                } else {
-                    log.error("Failed to get ML connector {}", connectorId, cause);
-                    listener.onFailure(cause);
-                }
-            } else {
-                try {
-                    GetResponse gr = r.parser() == null ? null : GetResponse.fromXContent(r.parser());
-                    if (gr != null && gr.isExists()) {
-                        try (
-                            XContentParser parser = MLNodeUtils
-                                .createXContentParserFromRegistry(NamedXContentRegistry.EMPTY, gr.getSourceAsBytesRef())
-                        ) {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -2077,40 +2077,43 @@ public void getConnector(String connectorId, String tenantId, ActionListener<Con
...
-        sdkClient.getDataObjectAsync(getDataObjectRequest).whenComplete((r, throwable) -> {
-            log.debug("Completed Get Connector Request, id:{}", connectorId);
-            if (throwable != null) {
-                Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
-                if (ExceptionsHelper.unwrap(cause, IndexNotFoundException.class) != null) {
-                    log.error("Failed to get connector index", cause);
-                    listener.onFailure(new OpenSearchStatusException("Failed to find connector", RestStatus.NOT_FOUND));
+        try (ThreadContext.StoredContext ctx = client.threadPool().getThreadContext().stashContext()) {
+            sdkClient.getDataObjectAsync(getDataObjectRequest).whenComplete((r, throwable) -> {
+                log.debug("Completed Get Connector Request, id:{}", connectorId);
```
dhrubo-os, line 2082: We can restore the context here? `ctx.restore();`?

cwperks: wdyt should I add that in? It doesn't look like the context is referenced inside this block at all but it doesn't hurt to restore the context once the action has finished either.

cwperks: FYI I will be driving a campaign shortly to remove and replace usages of stashContext: https://github.com/opensearch-project/opensearch-plugins/issues/238

Example of replacing the calls in geospatial plugin: https://github.com/opensearch-project/geospatial/pull/715

dhrubo-os: >wdyt should I add that in? It doesn't look like the context is referenced inside this block at all but it doesn't hurt to restore the context once the action has finished either.

The "action" is to get the object from the system index which should be done when we reach at line 2082. That's why I was suggesting to restore the context something like [this](https://github.com/opensearch-project/flow-framework/blob/6d6116d1bb688787f06a58fc0f6c2d9b09854007/src/main/java/org/opensearch/flowframework/transport/DeleteWorkflowTransportAction.java#L149)

cwperks: Added a line to restore the context once the action has completed.

dhrubo-os (approved):

b4sjoo (approved):

PR #3560, opened 2025-02-17
support embedding types

zane-neo:
### Description
BedRock [titan model V2](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html#model-parameters-titan-embed-text-request-response) and Cohere V2 model now supports multiple [embedding types](https://docs.cohere.com/reference/embed#request.body.embedding_types) in their API, which means you can specify different embedding types as a list in the request, and model generates different results based on embedding types in response, e.g.:
```
import cohere
import requests
import base64

co = cohere.ClientV2()

image = requests.get("https://cohere.com/favicon-32x32.png")
stringified_buffer = base64.b64encode(image.content).decode("utf-8")
content_type = image.headers["Content-Type"]
image_base64 = f"data:{content_type};base64,{stringified_buffer}"

response = co.embed(
    model="embed-english-v3.0",
    input_type="image",
    embedding_types=["float", "binary", "int8"],
    images=[image_base64],
)

print(response)
```
Response: 
```
{
  "id": "5807ee2e-0cda-445a-9ec8-864c60a06606",
  "embeddings": {
    "float": [
      [
        -0.007247925,
        -0.041229248,
        -0.023223877,
       ......
      ]
    ],
    "binary": [
        [
            12,
            -13,
            14
        ]
    ],
    "int8": [
        [
            19,
            20,
            21
        ]
    ]
  },
  "texts": [],
  "images": [
    {
      "width": 400,
      "height": 400,
      "format": "jpeg",
      "bit_depth": 24
    }
  ],
  "meta": {
    "api_version": {
      "version": "2"
    },
    "billed_units": {
      "images": 1
    }
  }
}
```

The current bedrock and cohere post process function has default response filter to `$.embedding` and `$.embeddings` respectively and the extracted results are both `List<List<Float>>` type, after they support embedding types, the response structure is different and the default response filter extracts `List<List<Float>>` and `Map<String, List<List<Number>>` type separately. This PR is to add support to user who wants to extract different types of results in model response, and this can be classified into several categories:
#### BedRock (The post process function applies for both v1 and v2 model)
1. The default case(response filter not set) is exactly same as before to maintain BWC.
2. When user specified the response filter to `$.embeddingsByType`, the response will be captured into `dataAsMap` in ModelTensor, and user can get all embedding types result.
3. When user specified the response filter to `$.embeddingsByType.{concrete_embedding_type}`, the response will be captured similar to default case to the `data` field in ModelTensor.

#### Cohere V2 (A new v2 post process function is added to v2 model)
1. The default case(response filter not set) is exactly same between v1 and v2. 
2. When user specified the response filter to `$.embeddings`, the response will be captured into `dataAsMap` in ModelTensor, and user can get all embedding types result.
3. When user specified the response filter to `$.embeddings.{concrete_embedding_type}`, the response will be captured similar to default case to the `data` field in ModelTensor.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3093

### Check List


Zhangxunmt (commented):
lgtm, but can you add ITs for the added embedding types in this PR?

zane-neo:
> lgtm, but can you add ITs for the added embedding types in this PR?

Added more ITs, also updated the blueprint under tutorial.

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/ConnectorPostProcessFunction.java
```diff
@@ -6,22 +6,23 @@
...
 
 import java.util.List;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 
+import org.opensearch.ml.common.output.model.MLResultDataType;
 import org.opensearch.ml.common.output.model.ModelTensor;
 
-public abstract class ConnectorPostProcessFunction<T> implements Function<Object, List<ModelTensor>> {
+public abstract class ConnectorPostProcessFunction<T> implements BiFunction<Object, MLResultDataType, List<ModelTensor>> {
```
jngz-es, line 14: This interface change looks like a breaking change, and not all PostProcessFunctions are using the ResultDataType like BedrockBatchJobArnPostProcessFunction. So what I am thinking is 1. to use extend instead of implements to support additional parameter MLResultDataType, or 2. to have MLResultDataType covering all cases and all PostProcessFunctions to use/set it, for example, not use null as dateType but have a default type something like FLOAT32.

zane-neo: I considered to change the Function to a self-defined interface/abstract class, but that needs more code changes and I want to limit the changes in this PR for easier reviewing. I'm open to make these changes in this PR if you insist or create a separate PR for better reviewing experience.

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockV2PostProcessFunctionInferenceIT.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+@Log4j2
+public class RestBedRockV2PostProcessFunctionInferenceIT extends MLCommonsRestTestCase {
```
Zhangxunmt, line 24: Why nor merge into the BedRock Inference IT to share the IT resources, reducing the overall IT run time length?
https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java

plugin/src/test/java/org/opensearch/ml/rest/RestCohereInferenceIT.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class RestCohereInferenceIT extends MLCommonsRestTestCase {
```
Zhangxunmt, line 21: This IT can be merged into the RemoteInferenceIT?

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockV2PostProcessFunctionInferenceIT.java
zane-neo: The template of RestBedRockInferenceIT and RestBedRockV2PostProcessFunctionInferenceIT are different, so if we merge them together, the code will seem mess up, separating them is for better code reading experience.

plugin/src/test/java/org/opensearch/ml/rest/RestCohereInferenceIT.java
zane-neo: The RemoteInferenceIT class file is already too big, I prefer we use separate file to test single feature for better maintenance.

Zhangxunmt: btw, the spotless is not applied yet.

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockV2PostProcessFunctionInferenceIT.java
zane-neo: Ok, I can take a look again to find a way to merge them and maintain the readability in best effort.

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/ConnectorPostProcessFunction.java
jngz-es: Yeah, for this pr, I am OK. We can have more discussion about if we need refactoring later.

jngz-es (approved):
LGTM, thanks for changes.

Zhangxunmt (approved):

plugin/src/test/java/org/opensearch/ml/rest/RestCohereInferenceIT.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Optional;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Before;
+import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
+import org.opensearch.ml.common.input.MLInput;
+
+public class RestCohereInferenceIT extends MLCommonsRestTestCase {
+    private final String COHERE_KEY = Optional.ofNullable(System.getenv("COHERE_KEY")).orElse("UzRF34a6gj0OKkvHOO6FZxLItv8CNpK5dFdCaUDW");
```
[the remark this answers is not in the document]
ylwu-amzn: @zane-neo, let's avoid adding credentials in code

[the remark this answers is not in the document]
dhrubo-os: [git-secrets](https://github.com/awslabs/git-secrets) could be helpful here.

Issue #3563, opened 2025-02-19
[BUG] IgnoreMissing in ml inference search request processors doesn't accept optional model input


mingshl:
### **Current Implementation:**
ML inference search request and response processors are the only search processors with an `ignoreMissing` flag. In the current implementation:

All fields in input_maps and out_maps are treated as required, similar to text_embedding processors.
When a field is missing:

- If ignoreMissing is false (default), an IllegalArgumentException is thrown.
- If ignoreMissing is true, processing is skipped, returning the original query or search hit.

### **Pros of Current Implementation:**

- Straightforward and predictable behavior
- Efficient for scenarios where inference is needed only for specific fields. **This is particular helpful, when the document has a lot of fields, for example 100 fields, but only 1 field `name` is for model inference. Users can bind this processor with the index, only call model inference when this field `name` is present in the query and match the json path, when querying other 99 fields, the processor is skipped, and no prediction task is wasted.**  
- Prevents unnecessary processing when critical fields are absent

###  **Cons of Current Implementation:**

- Lacks support for optional inputs, which is crucial for multi-model scenarios
- Inconsistent with ML inference ingest processors, potentially causing confusion

### Proposed Solutions:

**Proposal 1: Align Search Processors with Ingest Processors**

Allow absent fields in inputMaps when ignoreMissing is true
Proceed with model predictions even with missing inputs

**Pros:**
- Consistent with ingest processor behavior
- Supports optional inputs in predictions

**Cons:**
- May lead to unnecessary prediction attempts, potentially wasting resources

**Proposal 2: Introduce New Flags for Optional Fields**

Add optionalInputs and optionalOutputs flags
Maintain current ignoreMissing logic
Process predictions when at least one required input is present

**Pros:**

- Offers greater flexibility and control
- Maintains backward compatibility
- Efficient resource utilization
- Supports multi-model scenarios

**Cons:**

- Adds complexity to processor configuration
- Recommendation: Implement Proposal 2 to provide a more flexible and efficient solution while maintaining backward compatibility. This approach addresses the current limitations and accommodates future needs for complex input configurations in multi-model scenarios.

Additional Context:

Original issue: https://github.com/opensearch-project/ml-commons/issues/3211

martin-gaievski:
I do like the recommended option 2, this way we both can be backward compatible and allow more flexibility to the end user. I think that's fine if we have one extra parameter until that is well documented.

mingshl:
optional mapping is supported now,. closed this issue

PR #3587, opened 2025-02-26
fix optional mappings in ml inference search processors

mingshl:
### Description
fix optional mappings in ml inference search processors by adding new parameters `optional_input_map` and  `optional_output_map`. Please refer to the REST IT test to see the sample settings.

### Related Issues
[[Resolves #[Issue number to be closed when this PR is merged]
](https://github.com/opensearch-project/ml-commons/issues/3563)](https://github.com/opensearch-project/ml-commons/issues/3563)

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
+++ b/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
@@ -208,11 +208,11 @@ public class MLInferenceSearchRequestProcessor extends AbstractProcessor impleme
         }
 
         try {
-            if (!validateQueryFieldInQueryString(processInputMap, processOutputMap, queryString)) {
+            if (!validateQueryFieldInQueryString(processInputMap, processOutputMap, queryString, ignoreMissing)) {
                 requestListener.onResponse(request);
             }
         } catch (Exception e) {
-            if (ignoreMissing) {
+            if (ignoreFailure) {
                 requestListener.onResponse(request);
                 return;
             } else {
@@ -397,7 +397,11 @@ public class MLInferenceSearchRequestProcessor extends AbstractProcessor impleme
         }, Math.max(inputMapSize, 1));
     }
 
-    private boolean validateRequiredInputMappingFields(List<Map<String, String>> processInputMap, String queryString) {
+    private boolean validateRequiredInputMappingFields(
+        List<Map<String, String>> processInputMap,
+        String queryString,
+        boolean ignoreMissing
+    ) {
         // Suppress errors thrown by JsonPath and instead return null if a path does not exist in a JSON blob.
         Configuration suppressExceptionConfiguration = Configuration.defaultConfiguration().addOptions(Option.SUPPRESS_EXCEPTIONS);
         ReadContext jsonData = JsonPath.using(suppressExceptionConfiguration).parse(queryString);
@@ -409,14 +413,24 @@ public class MLInferenceSearchRequestProcessor extends AbstractProcessor impleme
                 String queryField = entry.getValue();
                 Object pathData = jsonData.read(queryField);
                 if (pathData == null) {
-                    throw new IllegalArgumentException("cannot find field: " + queryField + " in query string: " + jsonData.jsonString());
+                    if (!ignoreMissing) {
+                        throw new IllegalArgumentException(
+                            "cannot find field: " + queryField + " in query string: " + jsonData.jsonString()
+                        );
+                    } else {
+                        return false;
+                    }
                 }
             }
         }
         return true;
     }
 
-    private boolean validateRequiredOutputMappingFields(List<Map<String, String>> processOutputMap, String queryString) {
+    private boolean validateRequiredOutputMappingFields(
+        List<Map<String, String>> processOutputMap,
+        String queryString,
+        boolean ignoreMissing
+    ) {
         // Suppress errors thrown by JsonPath and instead return null if a path does not exist in a JSON blob.
         Configuration suppressExceptionConfiguration = Configuration.defaultConfiguration().addOptions(Option.SUPPRESS_EXCEPTIONS);
         ReadContext jsonData = JsonPath.using(suppressExceptionConfiguration).parse(queryString);
@@ -430,9 +444,13 @@ public class MLInferenceSearchRequestProcessor extends AbstractProcessor impleme
                     if (queryField.startsWith("query.") || queryField.startsWith("$.query.")) {
                         Object pathData = jsonData.read(queryField);
                         if (pathData == null) {
-                            throw new IllegalArgumentException(
-                                "cannot find field: " + queryField + " in query string: " + jsonData.jsonString()
-                            );
+                            if (!ignoreMissing) {
+                                throw new IllegalArgumentException(
+                                    "cannot find field: " + queryField + " in query string: " + jsonData.jsonString()
+                                );
+                            } else {
+                                return false;
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -376,23 +392,12 @@ public void onFailure(Exception e) {
...
-     * @param processOutputMap the list of output mappings
-     * @param queryString      the query string to be validated
-     * @return true if all query fields exist in the query string, false otherwise
-     */
-    private boolean validateQueryFieldInQueryString(
-        List<Map<String, String>> processInputMap,
-        List<Map<String, String>> processOutputMap,
-        String queryString
-    ) {
+    private boolean validateRequiredInputMappingFields(List<Map<String, String>> processInputMap, String queryString) {
```
jngz-es, line 395: This function `validateRequiredInputMappingFields` has same declare and very similar implementation with redundant code as the next function  `validateRequiredOutputMappingFields`. Could we abstract another function to implement the common logic?

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -419,17 +433,45 @@ private boolean validateQueryFieldInQueryString(
...
+    ) {
+        boolean isInputValid;
+        boolean isOutputValid;
+        if (processInputMap != null && processInputMap.size() > 0) {
+            isInputValid = validateRequiredInputMappingFields(processInputMap, queryString);
+        } else {
+            isInputValid = true;
+        }
+
+        if (processOutputMap != null && processOutputMap.size() > 0) {
```
jngz-es, line 459: We have a util class from opensearch `org.opensearch.core.common.util.CollectionUtils` which has a method able to check if a list is null or empty.
`CollectionUtils.isEmpty(processInputMap)`

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -403,6 +408,15 @@ private boolean validateQueryFieldInQueryString(
                 }
             }
         }
+        return true;
+    }
+
+    private boolean validateRequiredOutputMappingFields(List<Map<String, String>> processOutputMap, String queryString) {
```
jngz-es, line 414: Actually I didn't where to return false in these two input/output validation functions. It turns out you use the return value from these two functions at below code.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -419,17 +433,45 @@ private boolean validateQueryFieldInQueryString(
...
+        String queryString
+    ) {
+        boolean isInputValid;
+        boolean isOutputValid;
+        if (processInputMap != null && processInputMap.size() > 0) {
+            isInputValid = validateRequiredInputMappingFields(processInputMap, queryString);
+        } else {
+            isInputValid = true;
+        }
+
```
jngz-es, line 458: We should check if the `isInputValid` is false here to early stop when we want to return `isInputValid && isOutputValid` in the end.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -577,8 +621,26 @@ public MLInferenceSearchRequestProcessor create(
...
+            List<Map<String, String>> optionalOutputMaps = ConfigurationUtils
+                .readOptionalList(TYPE, processorTag, config, OPTIONAL_OUTPUT_MAP);
+
+            if ((inputMaps == null || inputMaps.isEmpty()) && (optionalInputMaps == null || optionalInputMaps.isEmpty())) {
+                throw new IllegalArgumentException(
+                    "Please provide at least one non-empty input_map or optional_input_map for ML Inference Search Request Processor"
+                );
+            }
+
+            if ((outputMaps == null || outputMaps.isEmpty()) && (optionalOutputMaps == null || optionalOutputMaps.isEmpty())) {
```
jngz-es, line 638: Using `CollectionUtils` could make this condition more readable.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
mingshl: we could but the logic is different for input map and output map, for example, if there is a query_template then we can skip checking output mapping field in query, it's easier to understand and faster return. If we abstract the codes, it makes it difficult to read and maintain

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
mingshl: it will throw exceptions when the path is null. the try catch logic will trigger if return response or return onFailure based on ignoreMissing true or false

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
mingshl: added 
```
        if(!isInputValid) {
            return false;
        }
```

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
mingshl: change to use CollectionUtils.isEmpty(), very pretty!

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
mingshl: added

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
jngz-es: Sounds good, thanks for the explanation.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
jngz-es: If we rely on exception throwing, we don't have to have a return value to check, for example, https://github.com/opensearch-project/ml-commons/blob/ce5d1ffcc1a2d257bb64c4b7972fe2482a4244ac/common/src/main/java/org/opensearch/ml/common/connector/Connector.java#L98.
It's minor, I am OK for this pr.

jngz-es (dismissed):

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
```diff
@@ -381,4 +381,37 @@ default void setRequestContextFromExt(SearchRequest request, PipelineProcessingC
...
+            Map<String, String> combinedMap = new HashMap<>();
+            if (i < mapList1.size()) {
+                combinedMap.putAll(mapList1.get(i));
+            }
+            if (i < mapList2.size()) {
+                combinedMap.putAll(mapList2.get(i));
+            }
+            combinedMaps.add(combinedMap);
+        }
+        return combinedMaps;
```
Zhangxunmt, line 415: How big are these lists? For large lists, using Streaming IntStream.range(0, maxSize) is more efficient.

Zhangxunmt (approved):

plugin/src/main/java/org/opensearch/ml/processor/ModelExecutor.java
mingshl: it's small list, the maxPredictionTask is 10, it won't go above 10. For performance wise, it won't make much difference.

jngz-es (approved):

Zhangxunmt (approved):

PR #3597, opened 2025-02-28
Use model type to check local or remote model

b4sjoo:
### Description
In stead of using user input algorithm we use model type extracting from model cache to check local or remote model, this will fix disable local model will cause remote neural search failure issue

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
@@ -126,10 +126,14 @@ public class RestMLPredictionAction extends BaseRestHandler {
     }
 
     /**
-     * Creates a MLPredictionTaskRequest from a RestRequest
+     * Creates a MLPredictionTaskRequest from a RestRequest. This method validates the request based on
+     * enabled features and model types, and parses the input data for prediction.
      *
-     * @param request RestRequest
-     * @return MLPredictionTaskRequest
+     * @param modelId The ID of the ML model to use for prediction
+     * @param modelType The type of the ML model, extracted from model cache to specify if its a remote model or a local model
+     * @param userAlgorithm The algorithm specified by the user for prediction, this is used todetermine the interface of the model
+     * @param request The REST request containing prediction input data
+     * @return MLPredictionTaskRequest configured with the model and input parameters
      */
     @VisibleForTesting
     MLPredictionTaskRequest getRequest(String modelId, String modelType, String userAlgorithm, RestRequest request) throws IOException {
--- a/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
@@ -19,6 +19,7 @@ import static org.opensearch.ml.utils.TenantAwareHelper.getTenantID;
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.opensearch.common.util.concurrent.ThreadContext;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -132,12 +132,12 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
      * @return MLPredictionTaskRequest
      */
     @VisibleForTesting
-    MLPredictionTaskRequest getRequest(String modelId, String algorithm, RestRequest request) throws IOException {
+    MLPredictionTaskRequest getRequest(String modelId, String modelType, String userAlgorithm, RestRequest request) throws IOException {
```
dhrubo-os, line 135: Let's add java doc for this method. I'm afraid this could be confusing later if we don't have enough documentation about this in the code.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -103,7 +103,7 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
                 client
                     .execute(
                         MLPredictionTaskAction.INSTANCE,
-                        getRequest(modelId, algoName, request),
+                        getRequest(modelId, functionName.get().name(), algoName, request),
```
dhrubo-os, line 106: This can break here. Because what if model isn't in cache. And it's just trying to pull model information from the system index?

dhrubo-os: If the model isn't deployed yet then we will find function name as null, and then algoName as "Remote" and this won't work.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
b4sjoo: Added

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -92,7 +92,7 @@ public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client
         }
 
         if (algorithm != null) {
-            MLPredictionTaskRequest mlPredictionTaskRequest = getRequest(modelId, algorithm, request);
+            MLPredictionTaskRequest mlPredictionTaskRequest = getRequest(modelId, functionName.get().name(), algorithm, request);
```
Zhangxunmt, line 95: the modelType is basically FunctionName, and algorithm = functionName.get().name(). So what is the difference between modelType and algorithm? If you want to define modelType to be "remote", "local", etc, you can add an internal function in the FunctionName to return the modelType FunctionName.modelType().

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionAction.java
```diff
@@ -83,27 +84,24 @@ public List<Route> routes() {
...
         String modelId = getParameterId(request, PARAMETER_MODEL_ID);
         Optional<FunctionName> functionName = modelManager.getOptionalModelFunctionName(modelId);
 
-        if (algorithm == null && functionName.isPresent()) {
-            algorithm = functionName.get().name();
-        }
-
-        if (algorithm != null) {
-            MLPredictionTaskRequest mlPredictionTaskRequest = getRequest(modelId, algorithm, request);
+        if (userAlgorithm != null && functionName.isPresent()) {
```
dhrubo-os, line 91: the edge case of this implementation is:

when customer didn't provide any algorithm in the request and the model is in cache, it will still go to the next block and make a query to the index, which is unnecessary.

ylwu-amzn (approved):
LGTM, make sure CI can pass

dhrubo-os:
@b4sjoo let's take updates from upstream.

b4sjoo:
Rebased

dhrubo-os (approved):

dhrubo-os:
@b4sjoo majority auto backports are failing. Are you planning to raise PR for those?

dhrubo-os:
> @b4sjoo majority auto backports are failing. Are you planning to raise PR for those?

@b4sjoo are you planning to do this?

PR #3632, opened 2025-03-08
Revert "Added Conversation API in MLClient"

owaiskazi19:
Reverts opensearch-project/ml-commons#3475 to unblock Flow Framework

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -16,7 +16,6 @@ plugins {
 dependencies {
     implementation project(path: ":${rootProject.name}-spi", configuration: 'shadow')
     implementation project(path: ":${rootProject.name}-common", configuration: 'shadow')
-    implementation project(path: ":${rootProject.name}-memory")
     compileOnly group: 'org.opensearch', name: 'opensearch', version: "${opensearch_version}"
     testImplementation group: 'junit', name: 'junit', version: '4.13.2'
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '5.15.2'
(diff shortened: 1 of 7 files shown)


owaiskazi19:
Looks like the issue is from the upstream core

```
| class: org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumerWrapper
| jar1: /Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch/opensearch/3.0.0-alpha1-SNAPSHOT/ca0d2b98c3e4748966e1e8bd784b5af016b187f7/opensearch-3.0.0-alpha1-SNAPSHOT.jar
| jar2: /Users/kazabdu/.gradle/caches/modules-2/files-2.1/org.opensearch/opensearch-ml-client/3.0.0.0-alpha1-SNAPSHOT/459c26e6ce20883267dcf225e8bffc434e6c9566/opensearch-ml-client-3.0.0.0-alpha1-SNAPSHOT.jar
|       at org.opensearch.common.bootstrap.JarHell.checkClass(JarHell.java:316)
|       at org.opensearch.common.bootstrap.JarHell.checkJarHell(JarHell.java:215)
|       at org.opensearch.common.bootstrap.JarHell.checkJarHell(JarHell.java:102)
|       at org.opensearch.common.bootstrap.JarHell.main(JarHell.java:86)

```

Let's not merge this until we decide on that

jngz-es:
Hi @owaiskazi19 , it turns out some API breaking changes which would break any use case using it. Usually we have deprecation plan at first, then we could remove it. I don't think we should remove APIs.
And also could you provide more details about how it is blocking Flow framework?

owaiskazi19:
> Hi @owaiskazi19 , it turns out some API breaking changes which would break any use case using it. Usually we have deprecation plan at first, then we could remove it. I don't think we should remove APIs directly. And also could you provide more details about how it is blocking Flow framework?

This is the error and the API was added today so no dependencies
https://github.com/opensearch-project/flow-framework/actions/runs/13729568900/job/38403720785?pr=1074

jngz-es:
Got it, it is a pretty new pr needed reverting. I am OK with that. BTW it's a jarhell issue, why you don't solve http-core dependency but remove some new added methods?

jngz-es (approved):

mingshl (approved):

mingshl:
let's merge this PR first to avoid the jar hell issue. When you test it fully, let's raise another new PR to bring back add conversation API @owaiskazi19

PR #3648, opened 2025-03-13
Use stream optional enum set from core in MLStatsInput

q-andy:
### Description
Refactors MLStatsInput to use `StreamInput::readOptionalEnumSet` and `StreamOutput::writeOptionalEnumSet` from OpenSearch core, added in this recent PR: https://github.com/opensearch-project/OpenSearch/pull/17556

The currently this is implemented in ml-commons using a private helper. I added this functionality to core using ml-commons as a reference so it could be reused in neural search, so the implementation is identical. While I made this change so other plugins could use the same function, I figure it might make sense to use it in ml-commons as well.

Compare
[ml-commons implmentation](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/stats/MLStatsInput.java#L142-L149)
```
private void writeEnumSet(StreamOutput out, EnumSet<?> set) throws IOException {
    if (set != null && set.size() > 0) {
        out.writeBoolean(true);
        out.writeEnumSet(set);
    } else {
        out.writeBoolean(false);
    }
}
```

[Core implementation](https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamOutput.java#L1262-L1269)
```
public <E extends Enum<E>> void writeOptionalEnumSet(@Nullable EnumSet<E> enumSet) throws IOException {
    if (enumSet != null && enumSet.size() > 0) {
        writeBoolean(true);
        writeEnumSet(enumSet);
    } else {
        writeBoolean(false);
    }
}
```
### Related Issues
n/a

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/stats/MLStatsInput.java
+++ b/plugin/src/main/java/org/opensearch/ml/stats/MLStatsInput.java
@@ -117,35 +117,26 @@ public class MLStatsInput implements ToXContentObject, Writeable {
     }
 
     public MLStatsInput(StreamInput input) throws IOException {
-        targetStatLevels = input.readBoolean() ? input.readEnumSet(MLStatLevel.class) : EnumSet.noneOf(MLStatLevel.class);
-        clusterLevelStats = input.readBoolean() ? input.readEnumSet(MLClusterLevelStat.class) : EnumSet.noneOf(MLClusterLevelStat.class);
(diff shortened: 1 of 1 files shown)


akolarkunnu:
Changes look good.

mingshl (approved):

mingshl:
thanks for bringing over this change to mlcommons, rerunning CI, please make sure the changes doesn't break any existing tests.

PR #3700, opened 2025-04-02
fix config index masterkey pull up for multi-tenancy

dhrubo-os:
### Description

In the config index the source can be:

```
{
  "_tenant_id": {
    "S": "058264223758:oasistestm8w8ikom"
  },
  "_id": {
    "S": "master_key_xJWBUnkbw8Gsj7pez-f-g0ccK2-JUSRHcD9uh64Mna0="
  },
  "_seq_no": {
    "N": "0"
  },
  "_source": {
    "M": {
      "create_time": {
        "N": "1743375153989"
      },
      "master_key": {
        "S": "tOVgn8xhoaYWjpgaAWUkvCUiQYcCEDfXv6aA6NJje8Q="
      },
      "tenant_id": {
        "S": "058264223758:oasistestm8w8ikom"
      }
    }
  }
}
```

When we store an ML encryption key, we save it in a config document with:
A unique _id like:  master_key_xJWBUnkbw8Gsj7pez-f-g0ccK2-JUSRHcD9uh64Mna0=
And the actual key value inside the _source field as: "master_key": "<actual AES key>"
The mistake was that when reading the key back, the code incorrectly tried to look it up using the full document ID (like master_key_xJWBUnkbw8...) as the field name — instead of just "master_key".
Since the _source does not contain a field with that name, it returned null. We then attempted to insert that null into the in-memory cache, which caused a NullPointerException when we tried to decode the key later.
What did we fix?
We changed the code to always look up the key from the correct field in _source, using "master_key" as the key — not the document ID.
We also added a check to make sure the retrieved value is a non-null String before using it.
Why this matters:
This fixes a bug that broke connector creation (and possibly other flows that use encryption). Without this fix, a newly created tenant or connector would crash when trying to initialize its master key.

This is only happening for multi-tenancy use case, but not for single tenancy. 

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
@@ -254,9 +254,15 @@ public class EncryptorImpl implements Encryptor {
         try {
             GetResponse getMasterKeyResponse = response.parser() == null ? null : GetResponse.fromXContent(response.parser());
             if (getMasterKeyResponse != null && getMasterKeyResponse.isExists()) {
-                this.tenantMasterKeys
-                    .put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) response.source().get(masterKeyId));
-                log.info("ML encryption master key already initialized, no action needed");
+
+                Object keyValue = response.source().get(MASTER_KEY);
+                if (keyValue instanceof String) {
+                    this.tenantMasterKeys.put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) keyValue);
+                    log.info("ML encryption master key already initialized, no action needed");
+                } else {
+                    log.error("Master key not found or not a string for tenantId: {}, masterKeyId: {}", tenantId, masterKeyId);
+                    exceptionRef.set(new ResourceNotFoundException(MASTER_KEY_NOT_READY_ERROR));
+                }
                 latch.countDown();
             } else {
                 initializeNewMasterKey(tenantId, masterKeyId, exceptionRef, latch, context);
@@ -416,9 +422,14 @@ public class EncryptorImpl implements Encryptor {
         } else {
             GetResponse getMasterKeyResponse = response1.parser() == null ? null : GetResponse.fromXContent(response1.parser());
             if (getMasterKeyResponse != null && getMasterKeyResponse.isExists()) {
-                this.tenantMasterKeys
-                    .put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) response1.source().get(masterKeyId));
-                log.info("ML encryption master key already initialized, no action needed");
+                Object keyValue = response1.source().get(MASTER_KEY);
+                if (keyValue instanceof String) {
+                    this.tenantMasterKeys.put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) keyValue);
+                    log.info("ML encryption master key already initialized, no action needed");
+                } else {
+                    log.error("Master key not found or not a string for tenantId: {}, masterKeyId: {}", tenantId, masterKeyId);
+                    exceptionRef.set(new ResourceNotFoundException(MASTER_KEY_NOT_READY_ERROR));
+                }
                 latch.countDown();
             } else {
                 exceptionRef.set(new ResourceNotFoundException(MASTER_KEY_NOT_READY_ERROR));
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
@@ -33,6 +33,7 @@ import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.Strings;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.index.engine.VersionConflictEngineException;
 import org.opensearch.ml.common.exception.MLException;
 import org.opensearch.ml.engine.indices.MLIndicesHandler;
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -416,9 +422,14 @@ private void handleVersionConflictResponse(
         } else {
             GetResponse getMasterKeyResponse = response1.parser() == null ? null : GetResponse.fromXContent(response1.parser());
             if (getMasterKeyResponse != null && getMasterKeyResponse.isExists()) {
-                this.tenantMasterKeys
-                    .put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) response1.source().get(masterKeyId));
-                log.info("ML encryption master key already initialized, no action needed");
+                Object keyValue = response1.source().get(MASTER_KEY);
+                if (keyValue instanceof String) {
```
dbwiddis, line 426: When (other than null) would it be anything other than a `String`?

Is this really just a null check?  If so the else block would probably work better in a `keyExists()` check.

ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
```diff
@@ -565,6 +564,49 @@ private GetResponse prepareMLConfigResponse(String tenantId) throws IOException
...
+
+        doAnswer(invocation -> {
+            ActionListener<Boolean> actionListener = (ActionListener) invocation.getArgument(0);
+            actionListener.onResponse(true); // init index success
+            return null;
+        }).when(mlIndicesHandler).initMLConfigIndex(any());
+
+        // Prepare a GetResponse where the _source has ONLY "master_key"
+        Map<String, Object> sourceMap = Map
+            .of(MASTER_KEY, "m+dWmfmnNRiNlOdej/QelEkvMTyH//frS2TBeS2BP4w=", CREATE_TIME_FIELD, Instant.now().toEpochMilli());
```
dbwiddis, line 562: this string is repeated a lot. Would be nice if it were a constant. Would be even nicer if it were randomly generated.

dbwiddis (approved):
LGTM. Some readability suggestions.

ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
dhrubo-os: I tried random, but then assertion will be hard...

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
dhrubo-os: it's a paranoid null check :D

ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
dhrubo-os: Addressed to put that in a constant.

dbwiddis: I'm fine with not random as long as the string isn't related to any other string being used!

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
dbwiddis: (Alternately a comment reminding me it's a null check with additional protection from wrong type/mapping/etc.)

dbwiddis (approved):
Re-approving with my suggestions addressed.

ylwu-amzn:
Should we backport to old versions like 2.19?

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -254,9 +254,15 @@ private void handleGetDataObjectSuccess(
         try {
             GetResponse getMasterKeyResponse = response.parser() == null ? null : GetResponse.fromXContent(response.parser());
             if (getMasterKeyResponse != null && getMasterKeyResponse.isExists()) {
-                this.tenantMasterKeys
-                    .put(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID), (String) response.source().get(masterKeyId));
-                log.info("ML encryption master key already initialized, no action needed");
+
+                Object keyValue = response.source().get(MASTER_KEY);
```
mingshl, line 258: is source() alway has the MASTER_KEY field?  if some response doesn't have the MASTER_KEY field, it will throw exception in the get() and the exception is not caught here

dhrubo-os: null pointer only can be thrown if `response.source()` is null, which I addressed in the next commit.

rbhavna (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
mingshl: looks better, approved

mingshl (approved):

dbwiddis (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -351,7 +363,8 @@ private void handlePutDataObjectFailure(
         CountDownLatch latch
     ) {
         Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable, OpenSearchStatusException.class);
-        if (cause instanceof VersionConflictEngineException) {
+        if (cause instanceof VersionConflictEngineException
```
dbwiddis, line 367: For DDB client this conditional works because there is no wrapped exception, but for Remote Client, using `OpenSearchStatusException` on line 365 will result in `cause` being an `OpenSearchException` if that is left as-is.

Two options to fix:
1. Just remove that `OpenSearchStatusException.class` vararg on line 365
2. Change this conditional to be `instanceof OpenSearchException` (since OpenSearchStatusException extends it, this works).

mingshl (approved):

PR #3719, opened 2025-04-09
 Add BERT Tokenizer as OpenSearch built-in analyzer

### LINKED ISSUE #3708: [RFC] Add BERT Tokenizer as OpenSearch built-in analyzer in ml-commons
### What problems are you trying to solve?
We're implementing analyzer-based neural sparse query support in the neural-search plugin, which requires BERT tokenizer functionality. ([RFC link](https://github.com/opensearch-project/neural-search/pull/1088)). 

The BERT tokenizer is widely used by OpenSearch pretrained text embedding models. Customers can use the analyzer to debug their input. Besides, the chunking processor also depends on analyzers to count token numbers.

However, we've encountered an issue that **native libraries (i.e. DJL) cannot be loaded by two different classloaders.** Since ml-commons already has dependencies on DJL and handles native library loading, we need to implement the BERT tokenizer functionality within ml-commons to avoid native library loading conflicts.

### What are you proposing?
Build `bert-base-uncased` tokenizer and `google-bert/bert-base-multilingual-uncased` as OpenSearch built in analyzers in ml-common. This will include:

The implementation will leverage DJL's HuggingFaceTokenizer, which is already a dependency in ml-commons. This will serve as the foundation for analyzer-based neural sparse queries in the neural-search plugin.

## Implementations

The main ambiguity is how we load the tokenizer config file ([example](https://huggingface.co/google-bert/bert-base-multilingual-uncased/blob/main/tokenizer.json)) and the idf file (saves the query token weight for neural sparse. [example](https://huggingface.co/opensearch-project/opensearch-neural-sparse-encoding-doc-v2-distill/blob/main/idf.json)). Here are several options:

### Option 1: Load from java resources dir

pros: no remote connect, most efficient
cons: increase the size for ml-commons jar file

### Option 2: Fetch from Hugging Face

pros: easy to implement, better extensibility for custom tokenizer
cons: security challenges if there is mallicious tokenizer.config at HF; can not work if remote website fails

### Option 3: Fetch from OpenSearch artifacts website

Fetch from https://artifacts.opensearch.org/ , the model file is identical with sparse tokenize files.

pros: don't increase jar size, the file is managed by opensearch
cons:  need to hard code the map of analyzer name and model artifact URL at some place.

Note for all options, we can build it in a lazy load fashion. That is we don't load one tokenizer multiple times, and only load the tokenizer when it's really invoked.

## Open discussion

We can also make the HF analyzer configurable by providing HF model id. But this need to be further reviewed by security teams. The custom analyzer is not in the scope now, but we do want to collect feedbacks about this.

zhichao-aws:
### Description
This PR add [bert-base-uncased](google-bert/bert-base-cased) tokenizer and [bert-base-multilingual-uncased](google-bert/bert-base-multilingual-uncased) tokenizer as OpenSearch built-in analyzer/tokenizer. Users can use them via analyze API without doing any special settings:
```
POST /_analyze
{
    "tokenizer" : "mbert-uncased",
    "text" : "hello world 你好"
}
```

We also have a follow up PR in neural-search to make neural-sparse search work with the analyzer.

### Related Issues
Resolves #3708 

### Check List


zhichao-aws:
We can merge this PR after 3.0.0-beta1 get released

mingshl:
would you consider option 3 from the issue? https://github.com/opensearch-project/ml-commons/issues/3708#issuecomment-2802261066

zhichao-aws:
> would you consider option 3 from the issue? [#3708 (comment)](https://github.com/opensearch-project/ml-commons/issues/3708#issuecomment-2802261066)

I've changed to option 3.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizer.java
```diff
@@ -0,0 +1,123 @@
...
+import ai.djl.huggingface.tokenizers.Encoding;
+import ai.djl.huggingface.tokenizers.HuggingFaceTokenizer;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * A Lucene Tokenizer implementation that uses Hugging Face tokenizer for tokenization.
+ * Supports token weighting and handles overflow scenarios.
+ */
+@Log4j2
+public class HFModelTokenizer extends Tokenizer {
```
mingshl, line 31: similarly, is the the tokenizer can be any customMLTokenizer ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,115 @@
...
+        Map<String, Float> token_weights;
+        String name;
+
+        BaseTokenizerHolder(String name, String url, String hashCode) {
+            this.name = name;
+
+            try {
+                Files.createDirectories(DJLUtils.getMlEngine().getAnalysisRootPath());
+                Path filePath = DJLUtils.getMlEngine().getAnalysisRootPath().resolve(name + ZIP_PREFIX);
+                downloadTokenizerFile(url, filePath, hashCode);
```
mingshl, line 55: question here, the file path is coded here, so it will download the default tokenizer for bert-uncased, can user change to use other tokenizer instead of this default tokenizer? it looks like it's not allowed at the moment.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,115 @@
...
+                ZipUtils.unzip(filePath.toFile(), DJLUtils.getMlEngine().getAnalysisRootPath().resolve(name));
+            } catch (IOException | PrivilegedActionException e) {
+                log.error("Failed to download analyzer zip file for {}. {}", name, e);
```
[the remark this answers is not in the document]
bzhangam: I think if we don't throw exception here it will throw null pointer exception later when we try to use it but it's not clear. So I think it's better to throw the exception here with clear error message. For retry I think we may want to have some auto retry mechanism in `downloadTokenizerFile`.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+        private void downloadTokenizerFile(String url, Path filePath, String hashCode) throws PrivilegedActionException {
+            AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                DownloadUtils.download(url, filePath.toString(), new ProgressBar());
+                String hash = calculateFileHash(filePath.toFile());
+                if (!hash.equals(hashCode)) {
+                    log.error("Analyzer zip file content hash can't match original hash value.");
+                    throw (new IllegalArgumentException("Analyzer " + name + " zip file content changed."));
```
bzhangam, line 77: Use `RuntimeException` since the url is hardcoded and it's not something that user can fix?

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/DJLUtils.java
```diff
@@ -0,0 +1,80 @@
...
+        return AccessController.doPrivileged((PrivilegedExceptionAction<T>) () -> {
+            ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+            try {
+                System.setProperty("java.library.path", mlEngine.getMlCachePath().toAbsolutePath().toString());
+                System.setProperty("DJL_CACHE_DIR", mlEngine.getMlCachePath().toAbsolutePath().toString());
+                Thread.currentThread().setContextClassLoader(ai.djl.Model.class.getClassLoader());
+
+                return action.call();
+            } finally {
+                Thread.currentThread().setContextClassLoader(contextClassLoader);
```
bzhangam, line 46: Do we also need to reset the system properties?

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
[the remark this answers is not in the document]
zhichao-aws: Break the OpenSearch node is not acceptable. So we can only log errors instead of throw exceptions during initialization.

I've added retry logics, it will be triggered when the analyzer is invoked, and it tries to download file when the tokenizer is not initialized

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/DJLUtils.java
zhichao-aws: The logics are consistent with other place using djl. We can keep them even finished.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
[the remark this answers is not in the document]
bzhangam: Will the null pointer exception be caught and not break the node? It's not clear to me why the null pointer exception is better than a clear exception explicitly threw by us.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+@Log4j2
+public class HFModelTokenizerFactory extends AbstractTokenizerFactory {
+    public static final String DEFAULT_TOKENIZER_NAME = "bert-uncased";
+
+    /**
+     * Atomically loads the HF tokenizer in a lazy fashion once the outer class accesses the static final set the first time.;
+     */
+    private static abstract class BaseTokenizerHolder {
+        private static final String ZIP_PREFIX = ".zip";
```
zane-neo, line 41: prefix or suffix?

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,129 @@
...
+        private static final String TOKENIZER_FILE_NAME = "tokenizer.json";
+        private static final String TOKEN_WEIGHTS_FILE_NAME = "idf.json";
+
+        HuggingFaceTokenizer tokenizer;
+        Map<String, Float> token_weights;
```
zane-neo, line 46: Use camel case.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        private void downloadTokenizerFile(String url, Path filePath, String hashCode) throws PrivilegedActionException {
+            AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                DownloadUtils.download(url, filePath.toString(), new ProgressBar());
+                String hash = calculateFileHash(filePath.toFile());
+                if (!hash.equals(hashCode)) {
+                    log.error("Analyzer zip file content hash can't match original hash value.");
+                    throw (new RuntimeException("Analyzer " + name + " zip file content changed."));
```
zane-neo, line 77: unnecessary parentheses.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
```diff
@@ -0,0 +1,129 @@
...
+                }
+                return null;
+            });
+        }
+    }
+
+    private static class DefaultTokenizerHolder extends BaseTokenizerHolder {
+        private static final String URL =
```
zane-neo, line 85: Is it possible to move this to a config file?

zane-neo:
LGTM, will approve once nit comments resolved.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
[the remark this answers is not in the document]
zhichao-aws: > Will the null pointer exception be caught and not break the node?
Yes.

The problem is we shouldn't throw exception at some place not catched by out layers.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
zhichao-aws: suffix. changed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
zhichao-aws: ack

ml-algorithms/src/main/java/org/opensearch/ml/engine/analysis/HFModelTokenizerFactory.java
zhichao-aws: As we switch back to option 1 now, we only need a resource file path now. I think it's okay to put it in code.

xinyual:
LGTM!

xinyual (approved):

mingshl:
@zane-neo @xinyual @zhichao-aws we are already code freeze for any new feature PRs. This PR passed the code freeze deadline, would you work with release team to get an exception? or please revert this PR

zhichao-aws:
Checked with @peterzhuamazon . We can still push feature code to main, as long as not backport to 3.0 branch

mingshl:
> Checked with @peterzhuamazon . We can still push feature code to main, as long as not backport to 3.0 branch

thanks for double checking.

PR #3724, opened 2025-04-11
Support phasing off SecurityManager usage in favor of Java Agent

kumargu:
### Description
With Java agent as a replacement to security manager, we start enforcing access to UnixDomainSockets. On windows since UnixDomainSockets are used for loopback address windows builds are now starting to fail for [skills](https://github.com/opensearch-project/skills/pull/551) and [flow-framework](https://github.com/opensearch-project/flow-framework/pull/1105). 

Additionally for this plugin's integ test to work, it will need a new jvm args to be passed. this change also includes the JVM args included in build.gradle. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,7 +83,16 @@ allprojects {
     }
 
     project.getExtensions().getExtraProperties().set("versions", VersionProperties.getVersions());
+}
 
+configurations {
+    agent
+}
+
+dependencies {
+    agent "org.opensearch:opensearch-agent-bootstrap:${opensearch_version}"
+    agent "org.opensearch:opensearch-agent:${opensearch_version}"
+    agent "net.bytebuddy:byte-buddy:1.17.5"
 }
 
 subprojects {
@@ -114,4 +123,14 @@ task updateVersion {
         // Include the required files that needs to be updated with new Version
         ant.replaceregexp(file:'build.gradle', match: '"opensearch.version", "\\d.*"', replace: '"opensearch.version", "' + newVersion.tokenize('-')[0] + '-SNAPSHOT"', flags:'g', byline:true)
     }
+}
+
+task prepareAgent(type: Copy) {
+    from(configurations.agent)
+    into "$buildDir/agent"
+}
+
+tasks.withType(Test) {
+    dependsOn prepareAgent
+    jvmArgs += ["-javaagent:" + project.layout.buildDirectory.file("agent/opensearch-agent-${opensearch_version}.jar").get()]
 }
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -84,6 +84,25 @@ allprojects {
...
     project.getExtensions().getExtraProperties().set("versions", VersionProperties.getVersions());
 
+    configurations {
+        agent
+    }
+
+    dependencies {
+        agent "org.opensearch:opensearch-agent-bootstrap:${opensearch_version}"
+        agent "org.opensearch:opensearch-agent:${opensearch_version}"
+        agent "net.bytebuddy:byte-buddy:1.17.5"
```
kumargu, line 94: once this PR is green. I can try using ```agent "net.bytebuddy:byte-buddy:${versions.bytebuddy}"```

build.gradle
```diff
@@ -84,6 +84,25 @@ allprojects {
...
+
+    tasks.withType(Test) {
+        dependsOn prepareAgent
+        jvmArgs += ["-javaagent:" + project.layout.buildDirectory.file("agent/opensearch-agent-${opensearch_version}.jar").get()]
+    }
+
+    task prepareAgent(type: Copy) {
+       from(configurations.agent)
+       into "$buildDir/agent"
+    }
```
andrross, line 105: I think this should not be nested within `allprojects`?

kumargu: Yeah, i also think so. figuring that out.

kumargu:
> I have tested with this PR [opensearch-project/OpenSearch#17900](https://github.com/opensearch-project/OpenSearch/pull/17900), just adding `apply plugin: 'opensearch.java-agent'` inside `:opensearch-ml-plugin:yamlRestTest` should solve the issue we dont have to add it across the gradle projects.

that's only when https://github.com/opensearch-project/OpenSearch/pull/17900 is merged?

prudhvigodithi:
> that's only when [opensearch-project/OpenSearch#17900](https://github.com/opensearch-project/OpenSearch/pull/17900) is merged?

Yes Once merged, its under review hope to get the PR merged soon.
Thanks
@getsaurabh02

build.gradle
```diff
@@ -83,7 +83,16 @@ allprojects {
     }
 
     project.getExtensions().getExtraProperties().set("versions", VersionProperties.getVersions());
+}
+
+configurations {
```
dhrubo-os, line 88: Just curious, if we are already doing this in the root gradle file, do we still need to do the same for all the modules?

peterzhuamazon (approved):

peterzhuamazon:
@prudhvigodithi opened:
* https://github.com/opensearch-project/ml-commons/pull/3727

peterzhuamazon:
Closing as #3727 is merged now. Thanks.

PR #3759, opened 2025-04-23
Replace null GetResponse with valid response and not exists

dbwiddis:
### Description

Several unit tests assume that it is possible for `client.get(request, actionListener)` to complete the listener's `onResponse()` method with `null`.  This is impossible.  See [here](https://github.com/opensearch-project/OpenSearch/blob/caf5d71ff02c5250a820f3ec1b346eff37b1d230/server/src/main/java/org/opensearch/action/get/TransportGetAction.java#L146-L165) for the ultimate return value of get requests.

What was done:
 - many tests that used `actionListener.onResponse(null)` were replaced with a non-null `GetResponse` object with a `false` value for the document existing.
   - for ease of review, these really are largely copy/paste with the only change being the index name
 - in the case of `MLModelManagerTests` there were two nearly identical tests, one for not exists (which was fixed as above) and one for null (which is impossible, and was removed, along with the mock helper producing the impossible result)

### Related Issues
Resolves test failures resulting from https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/145 which assumes that an `ActionListener.onResponse()` on a `client.get()` call will never be null.

### Check List

### DIFF
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
@@ -5,6 +5,8 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.opensearch.index.seqno.SequenceNumbers.UNASSIGNED_PRIMARY_TERM;
+import static org.opensearch.index.seqno.SequenceNumbers.UNASSIGNED_SEQ_NO;
 import static org.opensearch.ml.common.CommonValue.CREATE_TIME_FIELD;
 import static org.opensearch.ml.common.CommonValue.MASTER_KEY;
 import static org.opensearch.ml.common.CommonValue.ML_CONFIG_INDEX;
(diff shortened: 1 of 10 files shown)


dbwiddis:
Coverage check failed.  Committed https://github.com/opensearch-project/ml-commons/pull/3759/commits/88d61f2c7d4d5e3f9750b3835dbf76a2bda1b8a9 for the failing check, but there are a lot of this style of (unnecessary) null checks across the code that should probably be removed in a separate PR.  Avoiding that to minimize the scope of this PR to bugfix.

dbwiddis:
Added more test coverage. 🙄

dhrubo-os (approved):

dhrubo-os:
@dbwiddis we need to backport this PR to 3.0 branch, right?

ylwu-amzn:
> @dbwiddis we need to backport this PR to 3.0 branch, right?

Per my understanding, this is mainly to fix the UT, not a real bug , right ?

dhrubo-os:
> > @dbwiddis we need to backport this PR to 3.0 branch, right?
> 
> Per my understanding, this is mainly to fix the UT, not a real bug , right ?

We made a recent change to the [sdk client](https://github.com/opensearch-project/opensearch-remote-metadata-sdk/pull/141) about response from local client, to incorporate this we need to make few more changes in the ml-commons too.

dhrubo-os:
And the SDKClient change will fix the aggregate search bug we have now. So I think we should be able to merge the bug fixes to the 3.0 branch.

dbwiddis:
Yes, this fixes a side-effect of the aggregation bugfix and needs to be backported to 3.0 branch.

PR #3773, opened 2025-04-24
Clean up JSM from MCP

rithin-pullela-aws:
### Description
- Since Java Security Manager is deprecated form OpenSearch, we can use the MCP SDK as dependency instead of a custom JAR file.
- All the permissions previously added are no longer necessary. (Other than `FilePermission` and `SocketPermission ` other entries in `plugin-security.policy`  are noop)
- Thread group created for MCP(ExecutorService) before is no longer necessary since the default thread group created inside the MCP SDK don't need privileged access anymore
### Related Issues

### Check List


rithin-pullela-aws:
Hi team, this PR does not have code coverage because my previous PR did not have ant test coverage, I am working on another PR to add test coverage for all this code.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutor.java
```diff
@@ -85,71 +78,57 @@ public List<MLToolSpec> getMcpToolSpecs() {
...
-                    clientBuilder.executor(executor).connectTimeout(connectionTimeout);
-                }).customizeRequest(headerConfig).build();
-
-                // Create and initialize client
-                McpSyncClient client = McpClient
-                    .sync(transport)
-                    .requestTimeout(readTimeout)
-                    .capabilities(McpSchema.ClientCapabilities.builder().roots(false).build())
-                    .build();
+            Duration connectionTimeout = Duration.ofSeconds(super.getConnectorClientConfig().getConnectionTimeout());
```
ylwu-amzn, line 81: Is it possible this is null ? If yes, should we add default value ?

rithin-pullela-aws: No, if user does not provide a ConnectorClientConfig, we create an use a default one, so this will not be an issue.

ylwu-amzn (approved):

jngz-es (approved):

PR #3781, opened 2025-04-26
initial commit for MCP server in OpenSearch

zane-neo:
### Description
This PR is to support MCP server within OpenSearch, the feature is experimental and you'll need to enable transport-reactor-netty4 to test, for more details can refer this: https://docs.opensearch.org/docs/latest/install-and-configure/configuring-opensearch/network-settings/#selecting-the-transport

There are several experimental APIs are introduced:
```
POST /_plugins/_ml/mcp_tools/_register
POST /_plugins/_ml/mcp_tools/_remove
GET /_plugins/_ml/sse
POST /_plugins/_ml/sse/message
```
The later twos are standard MCP's sse and message APIs. An example of using these API is like below:
1. Create tools for MCP server:
```
POST /_plugins/_ml/mcp_tools/_register
{
    "tools": [
        {
            "name": "PPLTool",
            "description": "Use this tool to transfer natural language to generate PPL and execute PPL to query inside. Use this tool after you know the index name, otherwise, call IndexRoutingTool first. The input parameters are: {index:IndexName, question:UserQuestion}",
            "params": {
                "model_type": "FINETUNE",
                "model_id": "${your_model_id}"
            },
            "schema": {
                "type": "object",
                "properties": {
                    "parameters": {
                        "type": "object",
                        "question": {
                            "type": "string"
                        },
                        "index": {
                            "type": "string"
                        }
                    }
                },
                "required": [
                    "question",
                    "index"
                ]
            }
        },
        {
            "name": "ListIndexTool",
            "description": "This is my first list index tool",
            "params": {},
            "schema": {
                "type": "object",
                "properties": {
                    "indices": {
                        "type": "array",
                        "items": {
                            "type": "string"
                        },
                        "description": "OpenSearch index name list, separated by comma. for example: [\"index1\", \"index2\"], use empty array [] to list all indices in the cluster"
                    }
                },
                "additionalProperties": false
            }
        }
    ]
}
```
Sample response:
```
{
    "_ZNV5BrNTVm6ilcM7Jn1pw": {
        "created": true
    },
    "NZ9aiUCrSp2b5KBqdJGJKw": {
        "created": true
    }
}
```
2. Get a connection session for current client:
```
GET /_plugins/_ml/sse
```
Sample response:
```
event: endpoint
data: /sse/message?sessionId=9d6aeca5-89f6-4df3-82fc-625098362198
```
3. List tools in MCP server:
```
POST /_plugins/_ml/sse/message?sessionId=9d6aeca5-89f6-4df3-82fc-625098362198
{
  "jsonrpc": "2.0",
  "id": "1",
  "method": "tools/list"
}
```
Sample response:
```
event: message
data: {"jsonrpc":"2.0","id":"100","result":{"tools":[{"name":"ListIndexTool","description":"This is my first list index tool","inputSchema":{"type":"object","properties":{"indices":{"description":"OpenSearch index name list, separated by comma. for example: [\"index1\", \"index2\"], use empty array [] to list all indices in the cluster","type":"array","items":{"type":"string"}}},"additionalProperties":false}}]}}
```
4. Execute tool in MCP server:
```
POST /_plugins/_ml/sse/message?sessionId=9d6aeca5-89f6-4df3-82fc-62509836219
{
    "jsonrpc": "2.0",
    "id": "2",
    "method": "tools/call",
    "params": {
        "name": "ListIndexTool",
        "arguments": {
            "input": {
                "index": ["test"]
            }
        }
    }
}
```
Sample response:
```
event: message
data: {"jsonrpc":"2.0","id":"110","result":{"content":[{"type":"text","text":"row,health,status,index,uuid,pri(number of primary shards),rep(number of replica shards),docs.count(number of available documents),docs.deleted(number of deleted documents),store.size(store size of primary and replica shards),pri.store.size(store size of primary shards)\n1,green,open,.plugins-ml-config,2E1HgTkGS1-jwEYlyfd0og,1,0,1,0,4kb,4kb\n2,green,open,top_queries-2025.04.25-51488,I6BkcS5kSrOfh7Lgi4IcDg,1,0,28,12,138.9kb,138.9kb\n3,green,open,.opendistro_security,VSZdbnKYRKmvnmihS5TJ9Q,1,0,9,0,77.8kb,77.8kb\n"}],"isError":false}}
```
5. Remove tools in MCP server:
```
POST /_plugins/_ml/mcp_tools/_remove
[
    "ListIndexTool", "PPLTool"
]
```
Sample response:
```
{
    "_ZNV5BrNTVm6ilcM7Jn1pw": {
        "removed": true
    },
    "NZ9aiUCrSp2b5KBqdJGJKw": {
        "removed": true
    }
}
```

A example project that uses MCP client to connect to MCP server in OpenSearch can be found here: https://github.com/zane-neo/opensearch-mcpserver-test-example

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/17577

### Check List


common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTools.java
```diff
@@ -0,0 +1,105 @@
+package org.opensearch.ml.common.transport.mcpserver.requests.register;
```
ylwu-amzn, line 1: Add license header

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/MLMcpToolsRegisterNodeRequest.java
```diff
@@ -0,0 +1,64 @@
...
+        this.mcpTools = mcpTools;
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        mcpTools.writeTo(out);
+    }
+
+    public static MLMcpToolsRegisterNodeRequest fromActionRequest(TransportRequest actionRequest) {
```
rithin-pullela-aws, line 49: We are using `ActionRequest` in place of `TransportRequest` in other parts of the code, is there a reason behind using raw `TransportRequest`

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/remove/MLMcpToolsRemoveNodeRequest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.opensearch.transport.TransportRequest;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+public class MLMcpToolsRemoveNodeRequest extends TransportRequest {
+    private List<String> tools;
+
+    public MLMcpToolsRemoveNodeRequest(StreamInput in) throws IOException {
```
rithin-pullela-aws, line 22: Add `super(in)`

Similar to other parts of code, [example](https://github.com/opensearch-project/ml-commons/blob/920685d042c14a0a4613efb466d9d9c277f91e80/memory/src/main/java/org/opensearch/ml/memory/action/conversation/UpdateConversationRequest.java#L48)

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpServerTransportProvider.java
```diff
@@ -0,0 +1,208 @@
...
+            }).then();
+        }
+
+        @Override
+        public <T> T unmarshalFrom(Object data, TypeReference<T> typeRef) {
+            return objectMapper.convertValue(data, typeRef);
+        }
+
+        @Override
+        public Mono<Void> closeGracefully() {
```
rithin-pullela-aws, line 194: Should we also clean up the sessions map when the connection is closed gracefully?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,150 @@
...
+    @Override
+    protected MLMcpRegisterNodeResponse nodeOperation(MLMcpToolsRegisterNodeRequest request) {
+        return registerToolsOnNode(request.getMcpTools());
+    }
+
+    private MLMcpRegisterNodeResponse registerToolsOnNode(McpTools mcpTools) {
+        Flux.fromStream(mcpTools.getTools().stream()).flatMap(tool -> {
+            // check if user request contains tools that not in our system.
+            String toolName = tool.getName();
+            Tool.Factory factory = toolFactoryWrapper.getToolsFactories().get(toolName);
```
rithin-pullela-aws, line 114: Should we address the case where factory might be null?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,150 @@
...
+            return McpAsyncServerHolder.asyncServer
+                .removeTool(toolName)
+                .onErrorResume(e -> Mono.empty())
+                .then(McpAsyncServerHolder.asyncServer.addTool(toolSpecification));
+        })
+            .doOnComplete(() -> { log.debug("Successfully register tools on node: {}", clusterService.localNode().getId()); })
+            .doOnError(e -> {
+                log.error("Failed to register tools on node: {}", clusterService.localNode().getId(), e);
+            })
+            .subscribe();
```
rithin-pullela-aws, line 146: Here we are calling .subscribe() directly inside the method and immediately returning a MLMcpRegisterNodeResponse.
The .subscribe() method triggers an async operation.
This introduces a race condition — the response is returned before the tool registration actually completes. If a tool fails to register later, the caller will have already received a success response.


Changing the method Signature into this might resolve the issue:
```
private Mono<MLMcpRegisterNodeResponse> registerToolsOnNode(McpTools mcpTools) {
    return Flux.fromStream(mcpTools.getTools().stream())
        .flatMap(tool -> {
            // Same removeTool + addTool logic
        })
        .then(Mono.fromCallable(() -> new MLMcpRegisterNodeResponse(clusterService.localNode(), true)))
        .doOnError(e -> log.error("Failed to register tools on node", e))
        .doOnSuccess(r -> log.debug("Successfully registered tools on node"));
}

```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRemoveOnNodesAction.java
```diff
@@ -0,0 +1,116 @@
...
+    protected MLMcpRemoveNodeResponse newNodeResponse(StreamInput in) throws IOException {
+        return new MLMcpRemoveNodeResponse(in);
+    }
+
+    @Override
+    protected MLMcpRemoveNodeResponse nodeOperation(MLMcpToolsRemoveNodeRequest request) {
+        return removeMcpTools(request.getTools());
+    }
+
+    private MLMcpRemoveNodeResponse removeMcpTools(List<String> tools) {
```
rithin-pullela-aws, line 99: Same as above, would lead to race condition in .subscribe()

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpConnectionMessageStreamingAction.java
```diff
@@ -0,0 +1,164 @@
...
+                        })).subscribe()
+                    )
+                    .doOnError(e -> Mono.fromRunnable(() -> {
+                        try {
+                            channel.sendResponse(new BytesRestResponse(channel, new Exception(e)));
+                        } catch (IOException ex) {
+                            log.error("Failed to send exception response to client during messaging due to IOException");
+                        }
+                    }))
+                    .subscribe();
```
rithin-pullela-aws, line 145: In success path we would perform subsrcibe twice.

rithin-pullela-aws:
This is wonderful @zane-neo.

The only things I would like to know about is the clean up of clients when they close the connections gracefully and the race conditions I mentioned above.

Also, do we need to clean up stale clients who did not close the connection?

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
```diff
@@ -0,0 +1,136 @@
...
+ * description, parameters, and schema.
+ */
+@Log4j2
+@Data
+public class McpTool implements ToXContentObject, Writeable {
+    private static final String NAME_FIELD = "name";
+    private static final String DESCRIPTION_FIELD = "description";
+    private static final String PARAMS_FIELD = "params";
+    private static final String SCHEMA_FIELD = "schema";
+    private final String name;
```
ylwu-amzn, line 37: This is only for name or the tool type ? Current Agent Framework supports `name` and `type` when defining tool. That's for supporting multiple custom tools with same type. For example, user can define two MLModelTools with different name because they are for different purpose
```
   {
      "type": "MLModelTool",
     "name": "summarize"
    },
   {
      "type": "MLModelTool",
      "name": "translate"
    },
```

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
```diff
@@ -0,0 +1,136 @@
...
+    private final String description;
+    private Map<String, Object> params;
+    private Map<String, Object> schema;
```
ylwu-amzn, line 40: Suggest keep consistent with current tool definition, for example
```
    {
      "type": "IndexMappingTool",
      "attributes": {
        "input_schema": {
          "type": "object",
          "properties": {
            "index": {
              "type": "array",
              "description": "OpenSearch index name list, separated by comma. for example: index1, index2",
              "items": {
                "type": "string"
              }
            }
          },
          "required": ["index"],
          "additionalProperties": false
        },
        "strict": true
      }
    }
```

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
```diff
@@ -0,0 +1,136 @@
...
+                    break;
+                case SCHEMA_FIELD:
+                    schema = parser.map();
+                    break;
+                default:
+                    parser.skipChildren();
+                    break;
+            }
+        }
+        if (name == null) {
```
ylwu-amzn, line 95: As line 58 already checked null name, we can skip checking here

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -56,12 +56,22 @@ public class SearchIndexTool implements Tool {
...
+            {
+              "input": {
+                "index": {
+                  "type": "string",
+                  "description": "OpenSearch index name. for example: index1"
+                },
+                "query": {
+                  "type": "object",
+                  "description": "OpenSearch search index query. You need to get index mapping to write correct search query. It must be a valid OpenSearch query. Valid value:\\n{\\"query\\":{\\"match\\":{\\"population_description\\":\\"seattle 2023 population\\"}},\\"size\\":2,\\"_source\\":\\"population_description\\"}\\nInvalid value: \\n{\\"match\\":{\\"population_description\\":\\"seattle 2023 population\\"}}\\nThe value is invalid because the match not wrapped by \\"query\\".",
+                  "additionalProperties": false
```
ylwu-amzn, line 70: This part is not a direct string reformat. I see `\"required\":[\"index\",\"query\"]` was removed.

Can you explain why? To keep this PR scope clear, I would suggest put such reformatting/refactoring in a new PR

jngz-es:
I have some high level questions needed to clarify to help me understand this pr well.
1. Does this pr provide fundamental `ingest/query` opensearch functions as tools through MCP protocol? If yes, what's the definition of them look like? Could you have an example? From my standpoint, there should be way more use cases for `ingest/query`, but `PPLTool` for now.
2. From my understanding, we should only have single endpoint for opensearch cluster, not sure why we have multiple endpoints. 
a. A standard MCP client would discover what tools the opensearch server has by using the `endpoint` + `method: tools/list`. But in this pr, the endpoint is `/_plugins/_ml/sse/message?sessionId=9d6aeca5-89f6-4df3-82fc-625098362198`. How the client knows the `sessionId` before discovering? Why don't we put the `sessionId` in the tool input schema and use single endpoint?
b. Why do we have this rest api `POST /_plugins/_ml/mcp_tools/_register`? Are we expecting a cx/client to configure tools like ingest/query? What I am thinking is that we only need to provide a standard MCP server to allow a standard MCP client to interact.
c. Regarding the API `GET /_plugins/_ml/sse`, I can't relate this url path to MCP functionality, it is more like a common function for the entire ml-commons.
In summarization, I think we only need one new endpoint like `/_plugins/_ml/mcpserver` with supporting methods like `tools/list`, `tools/call`, `resources/list`, `resources/read` etc.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLRegisterMcpToolsAction.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        ActionRequestValidationException exception = new ActionRequestValidationException();
+        MLMcpToolsRegisterNodesRequest registerNodesRequest = getRequest(request);
+        if (CollectionUtils.isEmpty(registerNodesRequest.getMcpTools().getTools())) {
+            exception.addValidationError("tools list can not be null");
+            throw exception;
+        }
+        Set<String> buildInToolNames = toolFactories.values().stream().map(Tool.Factory::getDefaultType).collect(Collectors.toSet());
```
ylwu-amzn, line 70: Is it possible to register all of these built-in tools by default ? Why have to ask user register specific tools by themselves?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    protected MLMcpRegisterNodeResponse nodeOperation(MLMcpToolsRegisterNodeRequest request) {
+        return registerToolsOnNode(request.getMcpTools());
+    }
+
+    private MLMcpRegisterNodeResponse registerToolsOnNode(McpTools mcpTools) {
+        Flux.fromStream(mcpTools.getTools().stream()).flatMap(tool -> {
+            // check if user request contains tools that not in our system.
```
ylwu-amzn, line 112: What if the tool not in our system ? Should we handle this edge case ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpAsyncServerHolder.java
```diff
@@ -0,0 +1,29 @@
...
+public class McpAsyncServerHolder {
+
+    public static OpenSearchMcpServerTransportProvider mcpServerTransportProvider = new OpenSearchMcpServerTransportProvider(
+        new ObjectMapper()
+    );
+
+    public static final McpAsyncServer asyncServer = McpServer
+        .async(mcpServerTransportProvider)
+        .capabilities(McpSchema.ServerCapabilities.builder().tools(true).logging().build())
+        .serverInfo("OpenSearch-MCP-Server", "1.0.0")
```
ylwu-amzn, line 27: Version 1.0.0 traditionally signals that the software (or API, library, app, etc.) is stable, production-ready.

Let's use `0.x.y` versions (0.1.0, 0.9.0, etc.)  — meaning this is still experimental, APIs can change, breaking changes are allowed at any time

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,150 @@
...
+                    ActionListener<String> actionListener = ActionListener
+                        .wrap(r -> sink.success(new McpSchema.CallToolResult(List.of(new McpSchema.TextContent(r)), false)), e -> {
+                            log.error("Failed to execute tool, tool name: {}", toolName, e);
+                            sink.error(e);
+                        });
+                    actualTool.run(StringUtils.getParameterMap(arguments), actionListener);
+                })
+            );
+            return McpAsyncServerHolder.asyncServer
+                .removeTool(toolName)
```
ylwu-amzn, line 138: Re-register a tool with same name will remove the existing registered tool ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,150 @@
...
+                (exchange, arguments) -> Mono.create(sink -> {
+                    ActionListener<String> actionListener = ActionListener
+                        .wrap(r -> sink.success(new McpSchema.CallToolResult(List.of(new McpSchema.TextContent(r)), false)), e -> {
+                            log.error("Failed to execute tool, tool name: {}", toolName, e);
+                            sink.error(e);
+                        });
+                    actualTool.run(StringUtils.getParameterMap(arguments), actionListener);
+                })
+            );
+            return McpAsyncServerHolder.asyncServer
```
ylwu-amzn, line 137: Per my understanding, we will start a new MCP server on each node.
1. How do we route each session to correct node ? 
2. If a node restarted, will MCP server on that node be gone forever ?

rithin-pullela-aws:
Few High Level Questions about this:
1. Do we need experimental streaming feature in core for this to work (I could not make it work by cloning)
2. I think the server Endpoint should be `/_plugins/_ml/mcp` because only `/sse` seems generic.
3. Generally when we initialize an MCP client using MCP SDK, we need to provide the URL.
The SDK internally does these things:
 - `GET {url}/sse` to get the sessionID and the message endpoint like: `/messages/?session_id=c381902fafd3436295cd07df156498d2`
 - Next, the client performs the POST call to initialize the connection: `{url}/messages/?session_id=c381902fafd3436295cd07df156498d2`
 - 3rd initialization step and tool calls follow.

But in the code I see that we have these URLs:
- `GET /_plugins/_ml/sse`
- `POST /_plugins/_ml/sse/message?sessionId=9d6aeca5-89f6-4df3-82fc-625098362198`

Perhaps we do not need `/sse` in messages endpoint.

ylwu-amzn (approved):
Approved to unblock 3.0

model-collapse (approved):

model-collapse:
Merge after all CI passed.

mingshl:
If this is targeting 3.0 please add backport 3.0 label @zane-neo

jngz-es:
One more high level question.
Did you try do a load test on the execute tool api? I strongly suggest you to do a little bit.
As I mentioned above, your server logic is in rest handler `RestMcpConnectionMessageStreamingAction`, not through transport call. If my understanding is correct, mcp tools call would use netty thread which would block other incoming rest requests. Usually rest handler should be light-weight, I recommend to use another thread pool as what I did for model inference streaming, https://github.com/jngz-es/ml-commons/blob/7b7e0b16ceeeafe67eb42e752ca5b37a78e3bf90/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java#L191.

ylwu-amzn:
Add cluster setting to enable/disable this feature

zane-neo:
> One more high level question. Did you try do a load test on the execute tool api? I strongly suggest you to do a little bit. As I mentioned above, your server logic is in rest handler `RestMcpConnectionMessageStreamingAction`, not through transport call. If my understanding is correct, mcp tools call would use netty thread which would block other incoming rest requests. Usually rest handler should be light-weight, I recommend to use another thread pool as what I did for model inference streaming, https://github.com/jngz-es/ml-commons/blob/7b7e0b16ceeeafe67eb42e752ca5b37a78e3bf90/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java#L191.

@jngz-es, if there are any blocking operations either in rest handler or in transport handler, it'll block the `transport_worker` threads which are the NIO threads of netty which indeed can cause serious problems, but if you check the code here: https://github.com/opensearch-project/ml-commons/pull/3781/files#diff-5efd0e0dcc84b1e4e115cbb4f4dda4c812bfd5e1ae8317056a99c7fa3b92e861R127-R135 which is where the tools executed, they're been executed in a non-blocking fashion(using action listener), so it seems fine in the performance perspective. Anyway I'll do a performance test for this once I complete the following bug fixes, thanks!

ylwu-amzn:
Use `_plugins/_ml/mcp` as URL prefix for all MCP related APIs

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTools.java
zane-neo: done

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/MLMcpToolsRegisterNodeRequest.java
zane-neo: I think both are fine as they're all in OpenSearch core so the code won't have runtime issue, I believe I copied from other files this.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/remove/MLMcpToolsRemoveNodeRequest.java
zane-neo: done

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpServerTransportProvider.java
zane-neo: done

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
zane-neo: It doesn't seem able to fix the issue:
1. The suggestion code is returning a Mono<MLMcpRegisterNodeResponse> instead of MLMcpRegisterNodeResponse, and this is consistent to the framework TransportNodesAction's nodeOperation method. 
2. The framework forced a synchronize return, so the only way is to blocking wait for the response which is risky.
3. This is a framework drawback and I have a PR to fix this: https://github.com/opensearch-project/OpenSearch/pull/15166, but only one approve till today...

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRemoveOnNodesAction.java
zane-neo: Same above

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpConnectionMessageStreamingAction.java
zane-neo: I think they're not the same event, I tried to remove the inner one and feature breaks.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
zane-neo: Will change this in next PR.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ *  This class represents a tool that can be registered with OpenSearch. It contains information about the tool's name,
+ * description, parameters, and schema.
+ */
+@Log4j2
+@Data
+public class McpTool implements ToXContentObject, Writeable {
+    private static final String NAME_FIELD = "name";
+    private static final String DESCRIPTION_FIELD = "description";
+    private static final String PARAMS_FIELD = "params";
```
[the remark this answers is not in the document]
zane-neo: Done

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
zane-neo: Done

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String SCHEMA_FIELD = "schema";
+    private final String name;
+    private final String description;
+    private Map<String, Object> params;
+    private Map<String, Object> schema;
+    private static final String nameNotShownExceptionMessage = "name field required";
+
+    public McpTool(StreamInput streamInput) throws IOException {
+        name = streamInput.readString();
+        if (name == null) {
```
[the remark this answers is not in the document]
zane-neo: Removed.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/register/McpTool.java
zane-neo: Done

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/action/MLMcpToolsRegisterOnNodesAction.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+package org.opensearch.ml.common.transport.mcpserver.action;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.ml.common.transport.mcpserver.responses.register.MLMcpRegisterNodesResponse;
+
+public class MLMcpToolsRegisterOnNodesAction extends ActionType<MLMcpRegisterNodesResponse> {
+    public static MLMcpToolsRegisterOnNodesAction INSTANCE = new MLMcpToolsRegisterOnNodesAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/mcp_tools/register_on_nodes";
```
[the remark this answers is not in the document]
zane-neo: Yes, every data node will be registered with mcp tools.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
zane-neo: reverted.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLRegisterMcpToolsAction.java
zane-neo: Some tools needs init parameters like PPLTool needs model_id, we can't figure out this by default.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
zane-neo: This is already validated in the rest action: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLRegisterMcpToolsAction.java#L89

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpAsyncServerHolder.java
zane-neo: done

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
zane-neo: Yes, it's an override operation.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
zane-neo: This session management implemented in this PR: https://github.com/opensearch-project/ml-commons/pull/3803, for the tools definition persistent, we can add this feature in next release.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLRegisterMcpToolsAction.java
ylwu-amzn: Another option is let user pass in model_id for PPLTool. That way can avoid such register tool step, will make user experience much easier.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/action/MLMcpToolsRegisterOnNodesAction.java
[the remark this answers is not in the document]
ylwu-amzn: @zane-neo Thanks for confirming. I have some concerns about registering MCP tools on each data node:

- Consistency challenges: If multiple users try to register tools simultaneously, we could end up with inconsistent tool registrations across nodes (e.g., node 1 has user A's tool while node 2 has user B's tool). How do we maintain consistency?

- Race conditions: What happens if tool registration requests conflict across nodes? We need a mechanism to handle concurrent registrations.

- Synchronization overhead: Keeping tool registrations in sync across all data nodes adds complexity and network overhead.

- Failure handling: How do we handle cases where tool registration succeeds on some nodes but fails on others? We need a robust rollback mechanism.

What are your thoughts on handling these concerns?

I would highly suggest considering this option https://github.com/opensearch-project/ml-commons/pull/3781#discussion_r2064685985
> Is it possible to register all of these built-in tools by default ?

Issue #3527, opened 2025-02-11
[FEATURE] Add Recommended Space Type to OpenSearch-provided Pretrained Dense Models


bzhangam:
**Is your feature request related to a problem?**
When use KNN query with a dense model it's not easy to decide what [space type](https://opensearch.org/docs/1.0/search-plugins/knn/approximate-knn#spaces) we should use. If the recommended space type can be part of the model metadata for OpenSearch-provided Pretrained Dense Models it can help users who are not very familiar with ML models and KNN query.

We have an [issue](https://github.com/opensearch-project/k-NN/issues/2356) in KNN plugin about this. And we also have a [feature request](https://github.com/opensearch-project/neural-search/issues/803) in neural search plugin that expects us to be able to automatically decide the space type for the dense models.

**What solution would you like?**
For the OpenSearch-provided Pretrained Dense Models we should be able to know the recommended space type for KNN query from the model metadata.

For custom model we should allow users provide the recommended space type. Let's say a model owner want to recommend the space type there should be a way to add this info and then other people who want to use that model can easily access that info.

**What alternatives have you considered?**
We can document/hardcode the recommended space type.

**Do you have any additional context?**
N/A

bzhangam:
> [@bzhangam](https://github.com/bzhangam) are you looking into this? can we assign this to you?

No. I haven't looked into it yet. But if no one can pick this up I'll if I can work on it.

ylwu-amzn:
Need to add space type to pretrain model config file. Seems we don't need to change ml-commons code. 
For remote model, we can update blueprint document to add this space type parameter in model_config.

nathaliellenaa:
Can I get assigned to this?

PR #3786, opened 2025-04-28
Add space type mapping for pre-trained embedding models, add new additional_config field and BaseModelConfig class

nathaliellenaa:
### Description
- Add a new `additional_config` field to `model_config` to store new fields, such as `space_type`.
- Add a new `BaseModelConfig` class to handle model configuration for all models, except metrics correlation, question answering, text embedding, and remote models. 
- Add a new `RemoteModelConfig` class to handle model configuration for remote models
- Add space type mapping for pre-trained embedding models.

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3527
Resolves https://github.com/opensearch-project/ml-commons/issues/3791

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -32,7 +32,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.controller.MLRateLimiter;
-import org.opensearch.ml.common.model.DefaultModelConfig;
+import org.opensearch.ml.common.model.BaseModelConfig;
 import org.opensearch.ml.common.model.Guardrails;
 import org.opensearch.ml.common.model.MLDeploySetting;
 import org.opensearch.ml.common.model.MLModelConfig;
@@ -40,6 +40,7 @@ import org.opensearch.ml.common.model.MLModelFormat;
 import org.opensearch.ml.common.model.MLModelState;
 import org.opensearch.ml.common.model.MetricsCorrelationModelConfig;
 import org.opensearch.ml.common.model.QuestionAnsweringModelConfig;
+import org.opensearch.ml.common.model.RemoteModelConfig;
 import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
 
 import lombok.Builder;
@@ -281,8 +282,10 @@ public class MLModel implements ToXContentObject {
                     modelConfig = new QuestionAnsweringModelConfig(input);
                 } else if (algorithm.equals(FunctionName.TEXT_EMBEDDING)) {
                     modelConfig = new TextEmbeddingModelConfig(input);
+                } else if (algorithm.equals(FunctionName.REMOTE)) {
+                    modelConfig = new RemoteModelConfig(input);
                 } else {
-                    modelConfig = new DefaultModelConfig(input);
+                    modelConfig = new BaseModelConfig(input);
                 }
             }
             if (input.readBoolean()) {
@@ -628,8 +631,10 @@ public class MLModel implements ToXContentObject {
                         modelConfig = QuestionAnsweringModelConfig.parse(parser);
                     } else if (FunctionName.TEXT_EMBEDDING.name().equals(algorithmName)) {
                         modelConfig = TextEmbeddingModelConfig.parse(parser);
+                    } else if (FunctionName.REMOTE.name().equals(algorithmName)) {
+                        modelConfig = RemoteModelConfig.parse(parser);
                     } else {
-                        modelConfig = DefaultModelConfig.parse(parser);
+                        modelConfig = BaseModelConfig.parse(parser);
                     }
                     break;
                 case DEPLOY_SETTING_FIELD:
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -32,6 +32,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.controller.MLRateLimiter;
+import org.opensearch.ml.common.model.GeneralModelConfig;
 import org.opensearch.ml.common.model.Guardrails;
 import org.opensearch.ml.common.model.MLDeploySetting;
 import org.opensearch.ml.common.model.MLModelConfig;
@@ -278,8 +279,10 @@ public class MLModel implements ToXContentObject {
                     modelConfig = new MetricsCorrelationModelConfig(input);
                 } else if (algorithm.equals(FunctionName.QUESTION_ANSWERING)) {
                     modelConfig = new QuestionAnsweringModelConfig(input);
-                } else {
+                } else if (algorithm.equals(FunctionName.TEXT_EMBEDDING)) {
                     modelConfig = new TextEmbeddingModelConfig(input);
+                } else {
+                    modelConfig = new GeneralModelConfig(input);
                 }
             }
             if (input.readBoolean()) {
@@ -623,8 +626,10 @@ public class MLModel implements ToXContentObject {
                         modelConfig = MetricsCorrelationModelConfig.parse(parser);
                     } else if (FunctionName.QUESTION_ANSWERING.name().equals(algorithmName)) {
                         modelConfig = QuestionAnsweringModelConfig.parse(parser);
-                    } else {
+                    } else if (FunctionName.TEXT_EMBEDDING.name().equals(algorithmName)) {
                         modelConfig = TextEmbeddingModelConfig.parse(parser);
+                    } else {
+                        modelConfig = GeneralModelConfig.parse(parser);
                     }
                     break;
                 case DEPLOY_SETTING_FIELD:
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -32,7 +32,7 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.controller.MLRateLimiter;
-import org.opensearch.ml.common.model.GeneralModelConfig;
+import org.opensearch.ml.common.model.DefaultModelConfig;
 import org.opensearch.ml.common.model.Guardrails;
 import org.opensearch.ml.common.model.MLDeploySetting;
 import org.opensearch.ml.common.model.MLModelConfig;
@@ -282,7 +282,7 @@ public class MLModel implements ToXContentObject {
                 } else if (algorithm.equals(FunctionName.TEXT_EMBEDDING)) {
                     modelConfig = new TextEmbeddingModelConfig(input);
                 } else {
-                    modelConfig = new GeneralModelConfig(input);
+                    modelConfig = new DefaultModelConfig(input);
                 }
             }
             if (input.readBoolean()) {
@@ -629,7 +629,7 @@ public class MLModel implements ToXContentObject {
                     } else if (FunctionName.TEXT_EMBEDDING.name().equals(algorithmName)) {
                         modelConfig = TextEmbeddingModelConfig.parse(parser);
                     } else {
-                        modelConfig = GeneralModelConfig.parse(parser);
+                        modelConfig = DefaultModelConfig.parse(parser);
                     }
                     break;
                 case DEPLOY_SETTING_FIELD:
--- a/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
@@ -26,6 +26,10 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
 
+/**
+ * Base configuration class for ML models. This class handles 
+ * the basic configuration parameters that every ML model can support.
+ */
 @Setter
 @Getter
 public class BaseModelConfig extends MLModelConfig {
@@ -105,15 +109,11 @@ public class BaseModelConfig extends MLModelConfig {
             return;
         }
 
-        try {
-            Map<String, Object> allConfigMap = XContentHelper.convertToMap(XContentType.JSON.xContent(), allConfig, false);
-            Set<String> duplicateKeys = allConfigMap.keySet().stream().filter(additionalConfig::containsKey).collect(Collectors.toSet());
-            if (!duplicateKeys.isEmpty()) {
-                throw new Exception();
-            }
-        } catch (Exception e) {
+        Map<String, Object> allConfigMap = XContentHelper.convertToMap(XContentType.JSON.xContent(), allConfig, false);
+        Set<String> duplicateKeys = allConfigMap.keySet().stream().filter(additionalConfig::containsKey).collect(Collectors.toSet());
+        if (!duplicateKeys.isEmpty()) {
             throw new IllegalArgumentException(
-                "Duplicate keys found in both all_config and additional_config: " + String.join(", ", additionalConfig.keySet())
+                "Duplicate keys found in both all_config and additional_config: " + String.join(", ", duplicateKeys)
             );
         }
     }
--- a/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
@@ -28,7 +28,7 @@ import lombok.Getter;
 import lombok.Setter;
 
 /**
- * Base configuration class for ML models. This class handles 
+ * Base configuration class for ML local models. This class handles
  * the basic configuration parameters that every local model can support.
  */
 @Setter
--- a/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.model;
+
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.opensearch.common.xcontent.XContentHelper;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.ParseField;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.ToXContent;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public class BaseModelConfig extends MLModelConfig {
+    public static final String PARSE_FIELD_NAME = "base";
+    public static final NamedXContentRegistry.Entry XCONTENT_REGISTRY = new NamedXContentRegistry.Entry(
+        BaseModelConfig.class,
+        new ParseField(PARSE_FIELD_NAME),
+        it -> parse(it)
+    );
+
+    @Builder(toBuilder = true)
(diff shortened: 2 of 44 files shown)


common/src/test/java/org/opensearch/ml/common/transport/forward/MLForwardInputTest.java
```diff
@@ -26,9 +26,9 @@
 import org.opensearch.ml.common.dataset.MLInputDataType;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.input.parameter.clustering.KMeansParams;
+import org.opensearch.ml.common.model.GeneralModelConfig;
```
dhrubo-os, line 29: have we raised another PR for this?

dhrubo-os: I prefer `DefaultModelConfig`

nathaliellenaa: No, the `GeneralModelConfig` class is implemented in this PR. I'll create an issue about this and link to this PR as well.

common/src/main/java/org/opensearch/ml/common/model/DefaultModelConfig.java
```diff
@@ -0,0 +1,256 @@
...
+import org.opensearch.core.xcontent.XContentParser;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public class DefaultModelConfig extends MLModelConfig {
+    public static final String PARSE_FIELD_NAME = "default";
```
bzhangam, line 33: I feel default may not be a good name since it has `embedding_dimension` field. This doesn't sound like a default field for every model.

common/src/main/java/org/opensearch/ml/common/model/DefaultModelConfig.java
```diff
@@ -0,0 +1,256 @@
...
+        public static FrameworkType from(String value) {
+            try {
+                return FrameworkType.valueOf(value.toUpperCase(Locale.ROOT));
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Wrong framework type");
+            }
+        }
+    }
+
+    private void validateNoDuplicateKeys(String allConfig, Map<String, Object> additionalConfig) {
```
bzhangam, line 235: Can we add a validation that if it is a remote model and the type is explicitly defined as the text embedding then the embedding dimension and the space type should be provided? This should be backward compatible since for existing remote models user will not specify its type as text embedding. And in future if they want to specify that they should provide the related info.

common/src/main/java/org/opensearch/ml/common/model/DefaultModelConfig.java
nathaliellenaa: I've created `RemoteModelConfig` to handle remote models and `BaseModelConfig` to handle all other models.

common/src/main/java/org/opensearch/ml/common/model/DefaultModelConfig.java
nathaliellenaa: Added this validation in the `RemoteModelConfig` class.

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,120 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public class BaseModelConfig extends MLModelConfig {
```
dhrubo-os, line 31: let's add a javadoc for this class

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,120 @@
...
+    private void validateNoDuplicateKeys(String allConfig, Map<String, Object> additionalConfig) {
+        if (allConfig == null || additionalConfig == null || additionalConfig.isEmpty()) {
+            return;
+        }
+
+        try {
+            Map<String, Object> allConfigMap = XContentHelper.convertToMap(XContentType.JSON.xContent(), allConfig, false);
+            Set<String> duplicateKeys = allConfigMap.keySet().stream().filter(additionalConfig::containsKey).collect(Collectors.toSet());
+            if (!duplicateKeys.isEmpty()) {
+                throw new Exception();
```
dhrubo-os, line 112: What kind of exception are we throwing here?

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+        } catch (Exception e) {
+            throw new IllegalArgumentException(
```
dhrubo-os, line 115: Hmm I'm not following. If try fails to parse, then we are coming here, right? So doesn't that mean parsing fails somehow?

common/src/main/java/org/opensearch/ml/common/model/MLModelConfig.java
```diff
@@ -33,13 +36,28 @@ public MLModelConfig(String modelType, String allConfig) {
...
 
+    public MLModelConfig(String modelType, String allConfig, Map<String, Object> additionalConfig) {
+        this(modelType, allConfig);
+        this.additionalConfig = additionalConfig;
+    }
+
     public MLModelConfig(StreamInput in) throws IOException {
         this.modelType = in.readString();
         this.allConfig = in.readOptionalString();
+        this.additionalConfig = in.readMap();
```
dhrubo-os, line 47: I think we should add BWC check here?

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
```diff
@@ -106,6 +109,29 @@ public class MLRegisterModelInput implements ToXContentObject, Writeable {
     private Map<String, String> modelInterface;
     private String tenantId;
 
+    private static final Map<String, String> MODEL_SPACE_TYPE_MAPPING = Map
```
dhrubo-os, line 112: What is this for? This doesn't seem extensible? Can you please explain what's the idea behind this?

common/src/main/resources/index-mappings/ml_model.json
```diff
@@ -88,6 +88,9 @@
         },
         "all_config": {
           "type": "text"
+        },
+        "additional_config": {
```
dhrubo-os, line 92: we should increase the version as well.

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
nathaliellenaa: I'll modify it to `IllegalArgumentException` with `Duplicate keys found...` message.

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
nathaliellenaa: I'll remove the try and catch for parsing here since `XContentHelper.convertToMap()` already throws an exception when it fails to parse the JSON.

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
nathaliellenaa: This space type mapping helps users assign the correct space type for pre-trained embedding models. When users use knn query with embedding model, they often don't know which space type to specify. This automatic mapping will add the appropriate space type to the model metadata. Users don't need to specify the space type themselves, and they can check the assigned space type using the get model API. For example:
```
{
  "name": "huggingface/sentence-transformers/all-distilroberta-v1",
  "version": "1.0.1",
}

GET /_plugins/_ml/models/zRAI0ZYBsoymZW_H2XKt

// Response
{
    "name": "huggingface/sentence-transformers/all-distilroberta-v1",
    "model_group_id": "2cU50pYBb_8U7MgQh5zI",
    "algorithm": "TEXT_EMBEDDING",
    "model_version": "1",
    "model_state": "REGISTERED",
    "model_content_size_in_bytes": 330811571,
    "model_content_hash_value": "92bc10216c720b57a6bab1d7ca2cc2e559156997212a7f0d8bb70f2edfedc78b",
    "model_config": {
        "model_type": "roberta",
        "embedding_dimension": 768,
        "framework_type": "SENTENCE_TRANSFORMERS",
        "all_config": "{\"_name_or_path\":\"distilroberta-base\",\"architectures\":[\"RobertaForMaskedLM\"],\"attention_probs_dropout_prob\":0.1,\"bos_token_id\":0,\"eos_token_id\":2,\"gradient_checkpointing\":false,\"hidden_act\":\"gelu\",\"hidden_dropout_prob\":0.1,\"hidden_size\":768,\"initializer_range\":0.02,\"intermediate_size\":3072,\"layer_norm_eps\":0.00001,\"max_position_embeddings\":514,\"model_type\":\"roberta\",\"num_attention_heads\":12,\"num_hidden_layers\":6,\"pad_token_id\":1,\"position_embedding_type\":\"absolute\",\"transformers_version\":\"4.8.2\",\"type_vocab_size\":1,\"use_cache\":true,\"vocab_size\":50265}",
        "additional_config": {
            "space_type": "l2"
        }
    },
    "created_time": 1747283708708,
    "last_updated_time": 1747283749309,
    "last_registered_time": 1747283749307,
    "total_chunks": 34,
    "is_hidden": false
}
```

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,120 @@
...
+            builder.field(ALL_CONFIG_FIELD, allConfig);
+        }
+        if (additionalConfig != null) {
+            builder.field(ADDITIONAL_CONFIG_FIELD, additionalConfig);
+        }
+        builder.endObject();
+        return builder;
+    }
+
+    private void validateNoDuplicateKeys(String allConfig, Map<String, Object> additionalConfig) {
```
jngz-es, line 107: Why don't we allow duplicated keys? For better availability I think we could merge two configs, something like keeping allConfig if any conflicts. It's not friendly to throw an exception only as a duplicated key. Basically I don't think duplicated keys is an error.

nathaliellenaa: Right now we don't merge the two configs, so it will be confusing if users provide the same keys in both `all_config` and `additional_config`. Therefore, we want to make sure when registering the models, we don't allow the same key in both configs fields.

common/src/main/java/org/opensearch/ml/common/model/MLModelConfig.java
```diff
@@ -33,13 +38,33 @@ public MLModelConfig(String modelType, String allConfig) {
         this.allConfig = allConfig;
     }
 
+    public MLModelConfig(String modelType, String allConfig, Map<String, Object> additionalConfig) {
```
jngz-es, line 41: Hmm, you already have a new BaseModelConfig extending this abstract class to add a new additionalConfig. I am not sure what the difference is for these two classes. I think you don't have to add additionalConfig here that causes a version control for the existing interface change. We keep this interface unchanged, and have like TextEnbeddingModelConfig extending your new BaseModelConfig that would have less impact.

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -263,4 +290,17 @@ public static FrameworkType from(String value) {
         }
     }
 
+    private void validateNoDuplicateKeys(String allConfig, Map<String, Object> additionalConfig) {
```
jngz-es, line 293: You can use the same function in base class with changing `private` to `protected/public`. You don't have to hold a duplicated function for each child class.
Or putting this function in some util class is also OK, just don't have exactly same functions at multiple places.

nathaliellenaa:
Failing test is flaky
```
REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.tools.VisualizationsToolIT.testVisualizationNotFound" -Dtests.seed=814B407AF3E0187B -Dtests.security.manager=false -Dtests.locale=kde -Dtests.timezone=Australia/ACT -Druntime.java=23
VisualizationsToolIT > testVisualizationNotFound STANDARD_ERROR
    REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests "org.opensearch.ml.tools.VisualizationsToolIT.testVisualizationNotFound" -Dtests.seed=814B407AF3E0187B -Dtests.security.manager=false -Dtests.locale=kde -Dtests.timezone=Australia/ACT -Druntime.java=23

VisualizationsToolIT > testVisualizationNotFound FAILED
    java.lang.AssertionError: The response failed to meet condition after 5 attempts. Attempted to perform GET : /_plugins/_ml/models/oJ2I2pYBurGl-2fxAHRR
```

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
jngz-es: I think we shouldn't have hard-coded types here and new additional_config depending on it. This space_type is only for pretrained models which have config files for example, https://artifacts.opensearch.org/models/ml-models/huggingface/sentence-transformers/all-distilroberta-v1/1.0.1/torch_script/config.json. We should add additional_config in this config file, then ml-commons only needs to read and index it, don't have to know its content.

nathaliellenaa: Will create a PR in py-ml to update the config file with the space type mapping.

common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -278,8 +280,12 @@ public MLModel(StreamInput input) throws IOException {
                     modelConfig = new MetricsCorrelationModelConfig(input);
                 } else if (algorithm.equals(FunctionName.QUESTION_ANSWERING)) {
                     modelConfig = new QuestionAnsweringModelConfig(input);
-                } else {
+                } else if (algorithm.equals(FunctionName.TEXT_EMBEDDING)) {
```
dhrubo-os, line 283: What about Sparse encoding? Are you sure that won't break that model functionality?

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public String getWriteableName() {
+        return PARSE_FIELD_NAME;
+    }
+
+    public BaseModelConfig(StreamInput in) throws IOException {
+        super(in);
+        this.additionalConfig = in.readMap();
```
dhrubo-os, line 88: do we need to have BWC check here?

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeMap(additionalConfig);
```
dhrubo-os, line 94: same here

dhrubo-os:
Can you please add more details about how did you test this functionalities specially with Sparse models and also check with Neural plugin as well (like if your change breaking anything in neural end or not).

common/src/main/java/org/opensearch/ml/common/MLModel.java
nathaliellenaa: Old model config for sparse models might include fields like embedding dimension and framework type, which were previously required since we're using TextEmbeddingModelConfig. However, sparse encoding models don't actually need these fields to function. We have skipChildren [here](https://github.com/opensearch-project/ml-commons/blob/25919cb3486f0bba1525b16cef610eb69611e508/common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java#L74), which will skip parsing for the unknown field. So, this won't break the model functionality.

nathaliellenaa:
> Can you please add more details about how did you test this functionalities specially with Sparse models and also check with Neural plugin as well (like if your change breaking anything in neural end or not).

Confirmed with neural plugin that for now they are not using model config directly except for the new feature they're adding so it should not break anything in the neural end.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java
```diff
@@ -152,8 +151,7 @@ public void testRegisterAsymmetricModelRequest() throws Exception {
         assertEquals("test_model", registerModelInput.getModelName());
         assertEquals("1", registerModelInput.getVersion());
         assertEquals("TORCH_SCRIPT", registerModelInput.getModelFormat().toString());
-        assertEquals("query: ", ((TextEmbeddingModelConfig) registerModelInput.getModelConfig()).getQueryPrefix());
-        assertEquals("passage: ", ((TextEmbeddingModelConfig) registerModelInput.getModelConfig()).getPassagePrefix());
+        assertEquals(null, ((BaseModelConfig) registerModelInput.getModelConfig()).getAdditionalConfig());
```
dhrubo-os, line 154: What happened here? Could you please explain why we needed to change the assertEquals here?

common/src/main/java/org/opensearch/ml/common/model/TextEmbeddingModelConfig.java
```diff
@@ -54,32 +55,46 @@ public TextEmbeddingModelConfig(
...
+            poolingMode,
+            normalizeResult,
+            modelMaxLength,
+            null,
+            null
+        );
     }
 
-    @Builder(toBuilder = true)
+    @Builder(builderMethodName = "textEmbeddingConfigBuilder")
```
dhrubo-os, line 77: why do we need this? `textEmbeddingConfigBuilder`?

plugin/src/test/java/org/opensearch/ml/rest/RestMLInferenceIngestProcessorIT.java
```diff
@@ -519,7 +519,7 @@ public void testMLInferenceIngestProcessor_simulatesIngestPipelineSuccessfully_w
 
     private MLRegisterModelInput registerAsymmetricEmbeddingModelInput() {
         MLModelConfig modelConfig = TextEmbeddingModelConfig
-            .builder()
+            .textEmbeddingConfigBuilder()
```
dhrubo-os, line 522: why is this change necessary? Why can't we have simple `builder`?

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
```diff
@@ -0,0 +1,135 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Base configuration class for ML models. This class handles 
+ * the basic configuration parameters that every ML model can support.
```
dhrubo-os, line 32: >every ML model can support --> local or remote or both?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java
nathaliellenaa: The `getRestRequest` [here](https://github.com/opensearch-project/ml-commons/blob/76f0d63b40579664723bc9317841c02c6a31592f/plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java#L210) doesn't specify the function name as text embedding, so the model config is created as a `BaseModelConfig`.

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
```diff
@@ -0,0 +1,254 @@
...
+    public static final String MODEL_MAX_LENGTH_FIELD = "model_max_length";
+
+    private final Integer embeddingDimension;
+    private final FrameworkType frameworkType;
+    private final PoolingMode poolingMode;
+    private final boolean normalizeResult;
+    private final Integer modelMaxLength;
+
+    @Builder(builderMethodName = "remoteModelConfigBuilder")
+    public RemoteModelConfig(
```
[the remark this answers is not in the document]
nathaliellenaa: It's intended for any remote model, right now we only have a validation check for text embedding, but we can add more validations for other models in the future.

common/src/main/java/org/opensearch/ml/common/model/BaseModelConfig.java
nathaliellenaa: local model, I'll make the description more specific.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRegisterModelActionTests.java
dhrubo-os: I see, can you please check if asymmetric model have any issue with this change or not?

dhrubo-os (approved):

jngz-es (approved):

Issue #3639, opened 2025-03-12
[BUG] Enhance Input Validation for UpdateModel and UpdateModelGroup APIs


rithin-pullela-aws:
**What is the bug?**
The 2 APIs have potential vulnerability: UpdateModel and UpdateModelGroup

Currently, these APIs accept script-like variables in their request bodies without raising exceptions, which could potentially lead to security risks.

Impact: This vulnerability could potentially be exploited to inject malicious scripts, posing a risk to the system's integrity and security.

**How can one reproduce the bug?**
- Create a model or a model group
- Make this API Call:
```
PUT /_plugins/_ml/models/w2jYiJUBIo18BRbyKFkp
{
  "name": "<script>alert(1)</script>",
  "description": "<script>alert(1)</script>"
}
```
- Observe a successful update.

**What is the expected behavior?**
Apply stricter regex matching to validate inputs

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #3805, opened 2025-04-29
add validation for name and description for model model group and connector resources

dhrubo-os:
add validation for name and description for model model group and connector resources

### Description
resolved https://github.com/opensearch-project/ml-commons/issues/3639

Integ test will fail as code coverages is failing for previously merged PR.

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.common.transport.connector;
 
 import static org.opensearch.action.ValidateActions.addValidationError;
+import static org.opensearch.ml.common.utils.StringUtils.isSafeText;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -39,8 +40,25 @@ public class MLCreateConnectorRequest extends ActionRequest {
     @Override
     public ActionRequestValidationException validate() {
         ActionRequestValidationException exception = null;
+
         if (mlCreateConnectorInput == null) {
-            exception = addValidationError("ML Connector input can't be null", exception);
+            return addValidationError("ML Connector input can't be null", exception);
+        }
+
+        String modelName = mlCreateConnectorInput.getName();
+        String description = mlCreateConnectorInput.getDescription();
+
+        if (modelName != null && !isSafeText(modelName)) {
+            exception = addValidationError(
+                "Model connector name can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.)",
+                exception
+            );
+        }
+        if (description != null && !isSafeText(description)) {
+            exception = addValidationError(
+                "Model connector description can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.)",
+                exception
+            );
         }
 
         return exception;
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
@@ -6,12 +6,14 @@
 package org.opensearch.ml.common.transport.connector;
 
 import static org.opensearch.action.ValidateActions.addValidationError;
-import static org.opensearch.ml.common.utils.StringUtils.isSafeText;
+import static org.opensearch.ml.common.utils.StringUtils.validateFields;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
@@ -39,29 +41,14 @@ public class MLCreateConnectorRequest extends ActionRequest {
 
     @Override
     public ActionRequestValidationException validate() {
-        ActionRequestValidationException exception = null;
-
         if (mlCreateConnectorInput == null) {
-            return addValidationError("ML Connector input can't be null", exception);
+            return addValidationError("ML Connector input can't be null", null);
         }
+        Map<String, String> fieldsToValidate = new HashMap<>();
+        fieldsToValidate.put("Model connector name", mlCreateConnectorInput.getName());
+        fieldsToValidate.put("Model connector description", mlCreateConnectorInput.getDescription());
 
-        String modelName = mlCreateConnectorInput.getName();
-        String description = mlCreateConnectorInput.getDescription();
-
-        if (modelName != null && !isSafeText(modelName)) {
-            exception = addValidationError(
-                "Model connector name can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.)",
-                exception
-            );
-        }
-        if (description != null && !isSafeText(description)) {
-            exception = addValidationError(
-                "Model connector description can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.)",
-                exception
-            );
-        }
-
-        return exception;
+        return validateFields(fieldsToValidate);
     }
 
     @Override
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
@@ -21,6 +21,7 @@ import org.opensearch.core.common.io.stream.InputStreamStreamInput;
 import org.opensearch.core.common.io.stream.OutputStreamStreamOutput;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.ml.common.utils.FieldDescriptor;
 
 import lombok.Builder;
 import lombok.Getter;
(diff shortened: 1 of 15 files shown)


common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
```diff
@@ -39,8 +40,25 @@ public MLCreateConnectorRequest(StreamInput in) throws IOException {
...
-            exception = addValidationError("ML Connector input can't be null", exception);
+            return addValidationError("ML Connector input can't be null", exception);
+        }
+
+        String modelName = mlCreateConnectorInput.getName();
+        String description = mlCreateConnectorInput.getDescription();
+
+        if (modelName != null && !isSafeText(modelName)) {
+            exception = addValidationError(
+                "Model connector name can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.)",
```
ylwu-amzn, line 53: This error message is different with MLUpdateConnectorRequest.java one. Should we keep them consistent ?

```
"Model connector name can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.). Max length: 1000 characters.",
```

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
```diff
@@ -57,6 +58,22 @@ public ActionRequestValidationException validate() {
 
         if (updateContent == null) {
             exception = addValidationError("Update connector content can't be null", exception);
+        } else {
+            String modelName = updateContent.getName();
+            String description = updateContent.getDescription();
+
+            if (modelName != null && !isSafeText(modelName)) {
+                exception = addValidationError(
+                    "Model connector name can only contain letters, digits, spaces, underscores (_), hyphens (-), and dots (.). Max length: 1000 characters.",
```
ylwu-amzn, line 67: Don't see where validate `Max length: 1000 characters."` , Is that built-in logic in `isSafeText`?

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelRequest.java
```diff
@@ -45,8 +46,25 @@ public MLUpdateModelRequest(StreamInput in) throws IOException {
...
+
         if (updateModelInput == null) {
-            exception = addValidationError("Update Model Input can't be null", exception);
+            return addValidationError("Update Model Input can't be null", exception);
+        }
+
+        String modelName = updateModelInput.getName();
+        String description = updateModelInput.getDescription();
+
+        if (modelName != null && !isSafeText(modelName)) {
```
ylwu-amzn, line 57: Similar checking in multiple places ? Can we build some common util method ?

common/src/main/java/org/opensearch/ml/common/transport/connector/MLUpdateConnectorRequest.java
dhrubo-os: My bad, I removed that restriction from the regex. But then I forgot to remove that from the error message. 

Initially I was thinking to provide 1000 characters restriction. 

What do you think? Should we add such kind of restriction?

ylwu-amzn: Ignore the comment, find that internal logic in `isSafeText`

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java
dhrubo-os: Yes, in the next revision I kept it consistent.

common/src/main/java/org/opensearch/ml/common/transport/model/MLUpdateModelRequest.java
dhrubo-os: Yeah, I thought about that and felt too lazy to refactor 😄. But I fixed it in the second revision. Thanks for insisting on the highest standards!

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -497,4 +503,44 @@ public static String hashString(String input) {
...
+    public static ActionRequestValidationException validateFields(Map<String, String> fields) {
+        ActionRequestValidationException exception = null;
+
+        for (Map.Entry<String, String> entry : fields.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+
+            if (value != null && !isSafeText(value)) {
+                exception = addValidationError(
+                    key + " can only contain letters, digits, spaces, underscores (_), hyphens (-), dots (.), and colons (:)",
```
Zhangxunmt, line 537: To improve the performance, using StringBuilder msgBuilder = new StringBuilder() for error messages will reduce the memory usage.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -497,4 +503,44 @@ public static String hashString(String input) {
...
+     * Checks if the input is safe (non-null, non-blank, matches safe character set).
+     *
+     * @param value The input string to validate
+     * @return true if input is safe, false otherwise
+     */
+    public static boolean isSafeText(String value) {
+        if (value == null || value.isBlank()) {
+            return false;
+        }
+        return value.matches(SAFE_INPUT_REGEX);
```
Zhangxunmt, line 516: Pattern SAFE_TEXT_PATTERN = Pattern.compile(SAFE_INPUT_REGEX); Using Pattern matcher is more efficient as it only Compile pattern once rather than Compiles pattern every time in String.matches():

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -60,6 +63,9 @@ public class StringUtils {
         + "      return input;"
         + "\n    }\n";
 
+    // Regex allows letters, digits, spaces, hyphens, underscores, and dots.
+    private static final String SAFE_INPUT_REGEX = "^[a-zA-Z0-9 _\\-\\.:,'()]+$";
```
rithin-pullela-aws, line 67: Just curious how we determined the regex is this a standard regex generally used? I believe we want to prevent `<, >, ;, ", /, \, =`.
Since we are using this for description, maybe we need to allow `!, @` etc?

rithin-pullela-aws: Sounds good, allowing unicode makes a lot of sense. I think this regex is good. The same regex will apply for name field as well right?

dhrubo-os: Yes

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
dhrubo-os: The compiler (JAVA 9+) uses string concatenation factories (invokedynamic) and optimizes constants

For simple, non-loop cases, a + b + c is often as fast (or faster) than manual StringBuilder.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
dhrubo-os: Agree, addressed in the next commit.

jngz-es:
I have some high level comments.
1. If this is the case, should we also apply the same validation for the other rest APIs like `register agent`, `create memory` etc.?
2. Why do we only check `name` and `description` fields? We still have a lot of other fields with `String` type, should we also validate them like `function_name`, `version` etc. in `MLRegisterModelInput`?
3. I am curious do we have a concrete example of how name/description could inject and run malicious code?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -60,6 +63,12 @@ public class StringUtils {
         + "      return input;"
         + "\n    }\n";
 
+    // Regex allows letters, digits, spaces, hyphens, underscores, and dots.
+    private static final Pattern SAFE_INPUT_PATTERN = Pattern.compile("^[\\p{L}\\p{N}\\s.,!?():@\\-_'\"]*$");
```
jngz-es, line 67: We might not want to have the regex match empty string, right? So use `+` instead of `*`?

dhrubo-os: That's a good point. So for name I understand we don't want empty string, but description should be optional. Which is why I kept `+` what do you think?

dhrubo-os: Thanks for the comment. I refactored my PR in a way, so that it can be used for required or non required fields both. Please check.

jngz-es: Thanks for the addressing, LGTM.

jngz-es (approved):

rbhavna (approved):

chloewqg:
Hello! Do we also consider adding `/` in model name. Currently I am not able to create and register a local model: 

```
{
  "name": "huggingface/sentence-transformers/msmarco-distilbert-base-tas-b",
  "version": "1.0.2",
  "model_group_id": "Z1eQf4oB5Vm0Tdw8EIP2",
}

{
	"error": {
		"root_cause": [
			{
				"type": "action_request_validation_exception",
				"reason": "Validation Failed: 1: Model name can only contain letters, numbers, whitespace, and basic punctuation (.,!?():@-_'\");"
			}
		],
		"type": "action_request_validation_exception",
		"reason": "Validation Failed: 1: Model name can only contain letters, numbers, whitespace, and basic punctuation (.,!?():@-_'\");"
	},
	"status": 400
}
```

PR #3814, opened 2025-04-30
excluding circuit breaker for Agent

dhrubo-os:
### Description
[excluding circuit breaker for Agent]

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
+++ b/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
@@ -975,7 +975,9 @@ public class MLModelManager {
      * @param runningTaskLimit limit
      */
     public void checkAndAddRunningTask(MLTask mlTask, Integer runningTaskLimit) {
-        if (Objects.nonNull(mlTask) && mlTask.getFunctionName() != FunctionName.REMOTE) {
+
+        // for agent and remote model prediction we don't need to check circuit breaker
+        if (Objects.nonNull(mlTask) && mlTask.getFunctionName() != FunctionName.REMOTE && mlTask.getFunctionName() != FunctionName.AGENT) {
             checkOpenCircuitBreaker(mlCircuitBreakerService, mlStats);
         }
         mlTaskManager.checkLimitAndAddRunningTask(mlTask, runningTaskLimit);
--- a/plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
+++ b/plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
@@ -112,8 +112,7 @@ public abstract class MLTaskRunner<Request extends MLTaskRequest, Response exten
             if (clusterService.localNode().getId().equals(nodeId)) {
                 // Execute ML task locally
                 log.debug("Execute ML request {} locally on node {}", request.getRequestID(), nodeId);
-                checkOpenCircuitBreaker(mlCircuitBreakerService, mlStats);
-                executeTask(request, listener);
+                checkCBAndExecute(functionName, request, listener);
             } else {
                 // Execute ML task remotely
                 log.debug("Execute ML request {} remotely on node {}", request.getRequestID(), nodeId);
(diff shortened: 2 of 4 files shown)


plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
```diff
@@ -112,8 +112,7 @@ public void dispatchTask(
             if (clusterService.localNode().getId().equals(nodeId)) {
                 // Execute ML task locally
                 log.debug("Execute ML request {} locally on node {}", request.getRequestID(), nodeId);
-                checkOpenCircuitBreaker(mlCircuitBreakerService, mlStats);
-                executeTask(request, listener);
+                checkCBAndExecute(functionName, request, listener);
```
b4sjoo, line 115: Nit: Personally I don't like abbreviation. Try to use `CircuitBreaker` instead of CB?

rbhavna (approved):

b4sjoo (approved):

jngz-es (approved):

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
Zhangxunmt: are you using the latest main branch? I think this change is redundant in the main as checkCBAndExecute is already there?

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
```diff
@@ -975,7 +975,9 @@ private <T> ThreadedActionListener<T> threadedActionListener(String threadPoolNa
      * @param runningTaskLimit limit
      */
     public void checkAndAddRunningTask(MLTask mlTask, Integer runningTaskLimit) {
-        if (Objects.nonNull(mlTask) && mlTask.getFunctionName() != FunctionName.REMOTE) {
+
+        // for agent and remote model prediction we don't need to check circuit breaker
+        if (Objects.nonNull(mlTask) && mlTask.getFunctionName() != FunctionName.REMOTE && mlTask.getFunctionName() != FunctionName.AGENT) {
```
Zhangxunmt, line 980: For agent using pre-trained model, does this skip the CB?

dhrubo-os: That a good question. It will skip agent level execution. But under the agent execution when a model prediction is going to happen, then function name is not remote or agent, so that time it will check CB.

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
dhrubo-os:
```
public void run(FunctionName functionName, Request request, TransportService transportService, ActionListener<Response> listener) {
        if (!request.isDispatchTask()) {
            log.debug("Run ML request {} locally", request.getRequestID());
            checkCBAndExecute(functionName, request, listener);
            return;
        }
        dispatchTask(functionName, request, transportService, listener);
    }
```

This is where we applied.

dhrubo-os: I'm using latest main branch.

plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java
Zhangxunmt: Sure we can merge for now but need to remember that this would cause problems when agent is not using a remote model.

ylwu-amzn: For old versions, can user update setting to disable circuit breaker checking ?

dhrubo-os: From [code](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/breaker/MLCircuitBreakerService.java#L78-L81), looks like we only check [memory](https://docs.opensearch.org/docs/latest/ml-commons-plugin/cluster-settings/#set-jvm-heap-memory-threshold) and [disk](https://docs.opensearch.org/docs/latest/ml-commons-plugin/cluster-settings/#set-jvm-heap-memory-threshold). 

So if somebody sets up, `jvm_heap_memory_threshold` to 100 & `disk_free_space_threshold` to -1 then circuit breaker will be disabled. 

Does that answer your question @ylwu-amzn ?

ylwu-amzn: Sounds good, @dhrubo-os , can you check if we have this in our document?

dhrubo-os: which document are you referring to? I shared the memory and disk link from docs.opensearch.org.

Issue #3748, opened 2025-04-22
[FEATURE] expose agent update API


ruanyl:
**Is your feature request related to a problem?**
When developing and iterating on agents, we frequently need to tweak the prompt during the development cycle. Currently, there is no API available to update an existing agent’s configuration (such as the prompt). As a result, we are forced to create a new agent every time a change is needed, which generates a new agent ID each time. This creates unnecessary overhead and complicates agent management.

**What solution would you like?**
Expose an API that allows updating an existing agent's configuration, prompt, tools, etc.

**What alternatives have you considered?**
No

**Do you have any additional context?**
No

ylwu-amzn:
This is a known backlog. @jngz-es can you help on this ?

jngz-es:
Sure, let's have it on the roadmap.

brianf-aws:
Hey @rcho93 This feels like it can fit in with your feature (Prompt Management) you are developing. Lets keep this in mind

https://github.com/opensearch-project/ml-commons/issues/3711

jiapingzeng:
@jngz-es I'd like to work on this

PR #3820, opened 2025-05-01
Expose Update Agent API

jiapingzeng:
### Description
Exposes Agent Update API

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3748

### Check List


plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
```diff
@@ -0,0 +1,178 @@
...
+            if (throwable != null) {
+                Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
+                log.error("Failed to update ML Agent {}", agentId, cause);
+                wrappedListener.onFailure(cause);
+            } else {
+                try {
+                    UpdateResponse updateResponse = r.updateResponse();
+                    assert updateResponse != null : "Failed to update Agent";
+                    if (updateResponse.getResult() == DocWriteResponse.Result.UPDATED) {
+                        log.info("Successfully updated ");
```
nathaliellenaa, line 163: Agent ID can be added here

nathaliellenaa:
@jiapingzeng Are users allowed to update all fields within an agent, or are there any restrictions?

jiapingzeng:
> @jiapingzeng Are users allowed to update all fields within an agent, or are there any restrictions?

Users can update all fields within an agent. However the request body must be a valid `MLAgent` similar to the Create Agent API, so for example users cannot update a flow agent to a conversational agent without providing the additional `llm` field.

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
jiapingzeng: added, thanks

jngz-es:
> Users can update all fields within an agent. However the request body must be a valid `MLAgent` similar to the Create Agent API, so for example users cannot update from a flow agent to a conversational agent without providing the additional `llm` field.

Hmm, this is more like a new agent creation. Regarding `update agent`, I think some fields like `agent type` should keep unchanged, as agents look much different in terms of different `type` and we also don't allow updating `function_name` for `update model` API, as models look way different according to different `function name`.

dhrubo-os:
We have [hidden agent](https://docs.opensearch.org/docs/latest/ml-commons-plugin/agents-tools/agents/index/#hidden-agents), which admin/any other user can't update. So we need to check that condition. Please check update model api and you will get the idea

jiapingzeng:
> We have [hidden agent](https://docs.opensearch.org/docs/latest/ml-commons-plugin/agents-tools/agents/index/#hidden-agents), which admin/any other user can't update. So we need to check that condition. #

I am making a GET call first to check on the existing agent, if it is a hidden agent, a 403 will be returned for non-superadmins

https://github.com/opensearch-project/ml-commons/blob/5acd28c326f19d15ea4996ba5bb5c5454546b2d8/plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java#L118-L128

jiapingzeng:
> Hmm, this is more like a new agent creation. Regarding `update agent`, I think some fields like `agent type` should keep unchanged, as agents look much different in terms of different `type` and we also don't allow updating `function_name` for `update model` API, as models look way different according to different `function name`.

@jngz-es I decoupled the request input from `MLAgent` so that agent `type` cannot be changed anymore. Are there any other fields we would want to exclude from `update`?

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateRequest.java
```diff
@@ -0,0 +1,74 @@
...
+        this.mlAgentUpdateInput = mlAgentUpdateInput;
+    }
+
+    public MLAgentUpdateRequest(StreamInput in) throws IOException {
+        super(in);
+        this.mlAgentUpdateInput = new MLAgentUpdateInput(in);
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
```
nathaliellenaa, line 46: Can we also add a validation to check for empty/null agent ID here?

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
```diff
@@ -0,0 +1,183 @@
...
+                } else {
+                    try {
+                        GetResponse getResponse = r.getResponse();
+                        assert getResponse != null : "Failed to get Agent";
+                        XContentParser parser = JsonXContent.jsonXContent
+                            .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, getResponse.getSourceAsString());
+                        XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                        MLAgent retrievedAgent = MLAgent.parse(parser);
+
+                        if (TenantAwareHelper
```
Zhangxunmt, line 119: If TenantAwareHelper.validateTenantResource return false, it seems doesn't return back an error message.

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
```diff
@@ -0,0 +1,183 @@
...
+            sdkClient.getDataObjectAsync(getDataObjectRequest).whenComplete((r, throwable) -> {
+                log.debug("Completed Get Agent request for Agent ID {}", agentId);
+                if (throwable != null) {
+                    Exception cause = SdkClientUtils.unwrapAndConvertToException(throwable);
+                    log.error("Failed to get ML Agent {}", agentId, cause);
+                    wrappedListener.onFailure(cause);
+                } else {
+                    try {
+                        GetResponse getResponse = r.getResponse();
+                        assert getResponse != null : "Failed to get Agent";
```
Zhangxunmt: what exception are throw here if the response is null? Ideally we should clearly define the exception to be thrown. For example in this case should be IllegalStateException or a 400 exception indicating that the agent does not exist. 
also, AssertionError is an Error, not an Exception. Assertions are disabled by default in production.
The catch block catches Exception but not Error

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
```diff
@@ -0,0 +1,262 @@
...
+    private LLMSpec llm;
+    private List<MLToolSpec> tools;
+    private Map<String, String> parameters;
+    private MLMemorySpec memory;
+    private String appType;
+    private Instant lastUpdateTime;
+    private String tenantId;
+
+    @Builder(toBuilder = true)
+    public MLAgentUpdateInput(
```
Zhangxunmt, line 63: I think it worth to add some validations for some fields, like the memory should be an Id format, the name shouldn't exceed 128 character length, etc. So can you put some validation here based on your thinking?

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateRequest.java
jiapingzeng: I don't think agent ID can be null or empty here as it is a part of request URL `PUT /_plugins/_ml/agents/{{agent_id}}`, tried not specifying an agent ID and got the below error:
```
HTTP/1.1 400 Bad Request
{"error":"no handler found for uri [/_plugins/_ml/agents/] and method [PUT]"}
```
Request would get rejected before reaching the update agent route, which is defined [here](https://github.com/opensearch-project/ml-commons/pull/3820/files#diff-e80cb4e2c9efd81eb1eefbdd0cdba9af2018a3d13286b60a186885b8373dcbfaR48-R52)

If the provided agent ID is invalid/not found, it will fail during the GET request in UpdateAgentTransportAction.doExecute(). I added a 404 error in new commit when that happens.

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
jiapingzeng: validateTenantResource() should respond with a 403 if it fails:

https://github.com/opensearch-project/ml-commons/blob/2a9baeb2f70348cbf9fc3f1d17e854e5ed5262a6/plugin/src/main/java/org/opensearch/ml/utils/TenantAwareHelper.java#L52-L63

Added a log.error message in new commit when this happens.

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
jiapingzeng: Nice catch, addressed in new commit

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
jiapingzeng: Added validation in new commit for:
1. name non-blank and length<=128 check
2. if memory.type is specified, it has to be "conversation_index" (as that is the only one we support a the moment)
3. no duplicate tools allowed

We should probably backport these validation to `MLAgent` class as well for create API for consistency, I can do that in a future PR.

common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java
```diff
@@ -102,6 +104,11 @@ private void validate() {
         if (name == null) {
             throw new IllegalArgumentException("Agent name can't be null");
         }
+        if (name.isBlank() || name.length() > AGENT_NAME_MAX_LENGTH) {
```
dhrubo-os, line 107: Can we follow the same pattern here: https://github.com/opensearch-project/ml-commons/blob/a866cecf0555eed49e0da877ca09def7dfe10944/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java#L50-L54C16 ?

jiapingzeng: Seems that `StringUtils.validateFields` is for validating that fields only has safe text, it does not include blank or max length check. 

I can address validation enhancements in a future PR if that works?

dhrubo-os: Sure

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
```diff
@@ -0,0 +1,187 @@
...
+import org.opensearch.transport.client.Client;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class UpdateAgentTransportAction extends HandledTransportAction<ActionRequest, UpdateResponse> {
+
+    Client client;
```
akolarkunnu, line 53: Is there any specific reason these variables to be non-private ?

common/src/test/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInputTest.java
```diff
@@ -0,0 +1,437 @@
...
+    @Test
+    public void testToXContentIncomplete() throws Exception {
+        updateAgentInput = MLAgentUpdateInput.builder().agentId("test-agent-id").build();
+        String jsonStr = serializationWithToXContent(updateAgentInput);
+        assertTrue(jsonStr.contains("\"agent_id\":\"test-agent-id\""));
+        assertFalse(jsonStr.contains("\"name\""));
+    }
+
+    @Test
+    public void parseSuccess() throws Exception {
```
akolarkunnu: Usually we start all test method names with prefix "test"
Please update in all applicable places

plugin/src/main/java/org/opensearch/ml/action/agents/UpdateAgentTransportAction.java
jiapingzeng: This is consistent with other `TransportAction` classes, so that these variables can be accessed in unit testing.

```
@Test
public void testConstructor() {
    assertEquals(updateAgentTransportAction.client, client);
    assertEquals(updateAgentTransportAction.xContentRegistry, xContentRegistry);
}
```

common/src/test/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInputTest.java
jiapingzeng: Updated test names in new commit

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
```diff
@@ -0,0 +1,288 @@
...
+            );
+        }
+        if (memory != null && !memory.getType().equals("conversation_index")) {
+            throw new IllegalArgumentException(String.format("Invalid memory type: %s", memory.getType()));
+        }
+        if (tools != null) {
+            Set<String> toolNames = new HashSet<>();
+            for (MLToolSpec toolSpec : tools) {
+                String toolName = Optional.ofNullable(toolSpec.getName()).orElse(toolSpec.getType());
+                if (toolNames.contains(toolName)) {
```
dhrubo-os, line 280: Do we need to throw the exception for duplicate? Can't we just ignore? What's the requirement here?

dhrubo-os:
Can you please update the PR description with more details, like you did you test?

What was the request you made and what was the response? If you can cover the happy and sad cases, that would be awesome. Thanks.

common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
jiapingzeng: This is for consistency with `MLAgent` which is used for Create API, as it requires no duplicate tools as well: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/agent/MLAgent.java#L101-L120

PR #3838, opened 2025-05-07
Execlude trusted connector check for hidden model

dhrubo-os:
### Description
[Execlude trusted connector check for hidden model]

### Related Issues

### Check List


jngz-es:
This pr is only to skip url validation for inline connectors, correct?

dhrubo-os:
> This pr is only to skip url validation for inline connectors, correct?

Yeah. We support hidden model with inline connector. So I'm putting the validation only here for simplicity. Do you see any concern?

jngz-es:
No, it's OK to me. Thanks for confirming.

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -171,7 +172,8 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
                 "To upload custom model user needs to enable allow_registering_model_via_url settings. Otherwise please use OpenSearch pre-trained models."
             );
         }
-        registerModelInput.setIsHidden(RestActionUtils.isSuperAdminUser(clusterService, client));
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
+        registerModelInput.setIsHidden(isSuperAdmin);
```
ylwu-amzn, line 176: Should we check hidden field in model configuration? 
Is it possible super admin create non-hidden model?

dhrubo-os: In current design this is not possible for super admin to create non-hidden model.

rbhavna (approved):

plugin/src/main/java/org/opensearch/ml/action/register/TransportRegisterModelAction.java
```diff
@@ -171,7 +172,8 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLRegi
                 "To upload custom model user needs to enable allow_registering_model_via_url settings. Otherwise please use OpenSearch pre-trained models."
             );
         }
-        registerModelInput.setIsHidden(RestActionUtils.isSuperAdminUser(clusterService, client));
+        boolean isSuperAdmin = isSuperAdminUserWrapper(clusterService, client);
```
Zhangxunmt, line 175: This method for a single line adds complexity, additional method maintenance and reduced readability. this isn't typically considered best practice unless there's a specific testing requirement, but this static method is not covered in the UT at all.

dhrubo-os: If you notice, I have mocked the behavior in both of the tests:

```
        // Mock super admin check
        doReturn(false).when(transportRegisterModelAction).isSuperAdminUserWrapper(any(), any());
```

Issue #3843, opened 2025-05-09
Clean dependencies by following versions in core


jngz-es:
Currently we have many dependencies like Guava using different version with core and we force them to use local dependency to load the related classes. The better way is that we always use the same version with core if core has this dependency, then we CompileOnly it. We don't have to force any more. And also it will reduce jar hell issues if extending some plugins.

brianf-aws:
Hey @jngz-es can we close this issue?

krisfreedain:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

brianf-aws:
Hey @jngz-es can we close this issue?

Issue #3842, opened 2025-05-09
[BUG] gradlew run failed on main branch


jngz-es:
**What is the bug?**
```
=== Standard output of node `node{:opensearch-ml-plugin:integTest-0}` ===

»    ↓ errors and warnings from /local/home/jngz/github/jngz-es/ml-commons/plugin/build/testclusters/integTest-0/logs/opensearch.stdout.log ↓
» WARN ][stderr                   ] [integTest-0] May 09, 2025 9:03:27 AM org.opensearch.javaagent.bootstrap.AgentPolicy setPolicy
» WARN ][stderr                   ] [integTest-0] INFO: Policy attached successfully: org.opensearch.bootstrap.OpenSearchPolicy@54f4a7f0
» WARN ][stderr                   ] [integTest-0] May 09, 2025 9:03:27 AM org.apache.lucene.internal.vectorization.PanamaVectorizationProvider <init>
» WARN ][stderr                   ] [integTest-0] INFO: Java vector incubator API enabled; uses preferredBitSize=512; FMA enabled
» WARN ][stderr                   ] [integTest-0] WARNING: A restricted method in java.lang.foreign.Linker has been called
» WARN ][stderr                   ] [integTest-0] WARNING: java.lang.foreign.Linker::downcallHandle has been called by the unnamed module
» WARN ][stderr                   ] [integTest-0] WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for this module
» ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [integTest-0] fatal error in thread [main], exiting
»  java.lang.NoClassDefFoundError: com/google/common/collect/ImmutableSet
»       at org.opensearch.ml.task.MLTaskManager.<clinit>(MLTaskManager.java:84) ~[?:?]
»       at org.opensearch.ml.plugin.MachineLearningPlugin.createComponents(MachineLearningPlugin.java:589) ~[?:?]
»       at org.opensearch.node.Node.lambda$new$21(Node.java:1037) ~[opensearch-3.1.0-SNAPSHOT.jar:3.1.0-SNAPSHOT]
»       at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:273) ~[?:?]
»       at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708) ~[?:?]
[...]

PR #3844, opened 2025-05-09
quick fix for guava noclass issue

jngz-es:
### Description
A quick fix. In long term way, we should address the issue https://github.com/opensearch-project/ml-commons/issues/3843.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3842

### Check List

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -58,7 +58,6 @@ dependencies {
     implementation project(':opensearch-ml-algorithms')
     implementation project(':opensearch-ml-search-processors')
     implementation project(':opensearch-ml-memory')
-    compileOnly "com.google.guava:guava:32.1.3-jre"
 
     implementation group: 'software.amazon.awssdk', name: 'aws-core', version: "2.30.18"
     implementation group: 'software.amazon.awssdk', name: 's3', version: "2.30.18"


plugin/build.gradle
```diff
@@ -58,7 +58,6 @@ dependencies {
     implementation project(':opensearch-ml-algorithms')
     implementation project(':opensearch-ml-search-processors')
     implementation project(':opensearch-ml-memory')
-    compileOnly "com.google.guava:guava:32.1.3-jre"
```
dhrubo-os, line 61: Don't we need to do the same like this: https://github.com/opensearch-project/anomaly-detection/pull/1473/files ?

owaiskazi19: We need to import the dependency during runtime using `implementation` here

Zhangxunmt (approved):

b4sjoo (approved):

vibrantvarun:
I have a question. If we merge this PR then it might create jar hell issue for other plugins who are using guava and consuming ml-commons plugin.

plugin/build.gradle
jngz-es: Yeah, we have it in the following.

jngz-es:
> I have a question. If we merge this PR then it might create jar hell issue for other plugins who are using guava and consuming ml-commons plugin, because the same dependency would be found in 2 different locations.

1. We have `implementation` from long time ago, recently changed to `complieOnly` due to introducing job-scheduler.
2. Yeah, you are right. That's exactly the same case when we extended job-scheduler.
3. We definitely need some best practices for extensible plugins, as I mentioned in the issue https://github.com/opensearch-project/OpenSearch/issues/18113.

PR #3845, opened 2025-05-13
Add error handling for plan&execute agent

Hailong-am:
### Description
1. Remove line separator from tool's response
2. Index mapping tool support raw input for index parameter
3. Add json extract logic for LLM output is not json format 

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3848

### Check List


Hailong-am:
> Thanks for the fixes @Hailong-am ! Let me know when it is ready for review. Can you also please create an issue for the fixes involved in this PR? Thanks!

@pyek-bot PR description updated with issue number and the what's the actual fix, please help to take a look

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -95,7 +95,11 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
         try {
             List<String> indexList = new ArrayList<>();
             if (StringUtils.isNotBlank(parameters.get("index"))) {
-                indexList = gson.fromJson(parameters.get("index"), List.class);
+                try {
+                    indexList = gson.fromJson(parameters.get("index"), List.class);
+                } catch (Exception e) {
+                    indexList.add(parameters.get("index"));
```
mingshl, line 102: this will add the "index" even if it's not a list, is this change on purpose? can you add unit test to verify the change?

Hailong-am: Yes, it's on purpose.The expected input is a list, but LLM may not always follow the instructions and sometime the index is a single value of index name. 

For the unit test, let's me add one to cover this.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -528,6 +528,11 @@ String extractJsonFromMarkdown(String response) {
             if (response.contains("```")) {
                 response = response.substring(0, response.lastIndexOf("```"));
             }
+        } else {
```
dhrubo-os, line 558: 1. This method doesn't have any null check for `response`
2. Is there any chance that multiple code block can exist?
3. what about json array?
4. what happens if there's any string literals like ` {"data": "{nested}"}`
5. Possible IndexOutOfBoundsException if markers aren't found
6. why do we need second trim?
7. Using substring multiple times creates multiple intermediate string objects
Each operation like contains(), indexOf(), and lastIndexOf() scans through the string

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
dhrubo-os: Can we please add a comment in the code too?

dhrubo-os:
@Hailong-am fix spotless please, `./gradlew spotlessApply`

Hailong-am:
> @Hailong-am fix spotless please, `./gradlew spotlessApply`

@dhrubo-os @ylwu-amzn I have applied the spotless, can you review it again.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
Hailong-am: > This method doesn't have any null check for response

`response` passed to this method could not be null, otherwise it will throw NPE before call it. But it's a good practice to add null check here

> Is there any chance that multiple code block can exist?
> what about json array?
> what happens if there's any string literals like  {"data": "{nested}"}

Yes, all of them are possible. the logic here is try our best to extract the expected json from LLM response.

> Possible IndexOutOfBoundsException if markers aren't found

if markers are not found, it will not perform substring

> why do we need second trim?

the content extracted from ```json ``` block may have spaces, so trim is needed here before check it's a valid json string.

> Using substring multiple times creates multiple intermediate string objects
Each operation like contains(), indexOf(), and lastIndexOf() scans through the string

that's not a big concern. the response is not large, and gc will help to handle intermediate string objects

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
Hailong-am: comments added

Issue #3711, opened 2025-04-04
[FEATURE] Prompt Management in ML-Commons


brianf-aws:
**Is your feature request related to a problem?**
ML-commons provides users the ability to interact with ML Models used for Semantic Search, MultiModal Capabilities, RAG, ML Agents, and LLM Integrations. This access to models allow for users to use OpenSearch as their preferred platform for GenAI applications. At the forefront of GenAI applications there exists a strong need for **prompt management** which enables users to get the most out of their LLMs by **tracking, auditing, versioning, testing and evaluating their prompts.** This project proposes and expedites the need for ML-Commons to support prompt management, further establishing **OpenSearch** as the **desired platform** for users’ **GenAI solutions**. 

**What solution would you like?**
This solution requires ML-Commons to have a Prompt Management API to handle my prompts. In doing this I can track my prompts place tags, test my prompts against remote ML Models and integrate with existing ML-Commons APIs. Below are functional and non-functional requirements of the solution. Note not all have to be achieved at a certain time but can be implemented over time (Integration with other APIS/SDKs)

#### Functional Requirements

* Prompt Management Core Functions
    * Create new prompts
    * Read/retrieve existing prompts
    * Update prompts
    * Delete prompts
    * Store prompts in a logical structured format that can be semantically is shown across other prompt management systems
* Version Control Features
    * Track prompt version history
    * Revert to previous versions
    * View change history
    * Compare different versions
* Tagging functionality
    * Add tags to prompts
    * Remove tags from prompts
    * Filter prompts by tags
    * Search prompts using tags
* LLM Integration
    * Connect with ML Models
    * Execute prompts through LLMs
    * Ability to set/updated LLM HyperParameters to test
* Scoring System
    * Calculate prompt effectiveness scores
    * Track prompt performance metrics
    * Compare prompt performances

#### Non-Functional Requirements

* Performance
    * Low Latency with Many prompts
* Testability
    * 85% Code coverage 
    * Integration tests
* Maintainability
    * Has ability to extend to other features and existing ML-Commons APIs
    * Has documentation to accompany the feature
    * Implementes SOLID Principles
    * Has proper JavaDocs and naming conventions
* Usability
    * User interface responsiveness
    * Intuitive version control system
    * Can be used from a Front end developer perspective to integrate ML-Commons prompt management

**What alternatives have you considered?**
The most prominent Prompt management solution I know is Langfuse, but I would like integration within OpenSearch to reduce friction on my GenAI Application.

[...]

PR #3849, opened 2025-05-14
Prompt Management SEARCH API 

rcho93:
### Description
[Propose Search Prompt API]

### Related Issues
Resolves #[https://github.com/opensearch-project/ml-commons/issues/3711]

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/prompt/MLCreatePromptInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/prompt/MLCreatePromptInput.java
@@ -38,6 +38,7 @@ public class MLCreatePromptInput implements ToXContentObject, Writeable {
 
     public static final String PROMPT_FIELD_USER_PROMPT = "user";
     public static final String PROMPT_FIELD_SYSTEM_PROMPT = "system";
+    public static final String PROMPT_VERSION_INITIAL_VERSION = "1";
 
     private String name;
     private String description;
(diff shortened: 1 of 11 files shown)


common/src/test/java/org/opensearch/ml/common/prompt/MLPromptTest.java
jiapingzeng, line 1: nit: this test file should be one level higher to match code class MLPrompt.java file location

common/src/main/java/org/opensearch/ml/common/transport/prompt/MLCreatePromptInput.java
```diff
@@ -69,25 +70,19 @@ public MLCreatePromptInput(
...
             throw new IllegalArgumentException("MLPrompt prompt field requires " + PROMPT_FIELD_USER_PROMPT + " parameter");
         }
-        if (version == null) {
-            throw new IllegalArgumentException("MLPrompt version field is null");
-        }
 
         this.name = name;
         this.description = description;
-        this.version = version;
+        this.version = version == null ? PROMPT_VERSION_INITIAL_VERSION : version;
```
jiapingzeng, line 85: I like that version will now be auto initialized to 1 when creating a new prompt; version management should be invisible to the end user when possible.

What is the expected flow for creating a new version of prompt? Should the user call create with the same prompt name but specify version 2, or call update? If the latter, will update auto increment the version?

rcho93: Yes, each prompts will have unique name. So the user can either bump the version by providing the same name & updated metadatas on create or the version will auto-increment on every update.

Issue #3839, opened 2025-05-07
[BUG] Model prediction is working, but if we use the same model in the agent then agent execution is not working


srikanthpadakanti:
OpenSearch ML Commons throws two inconsistent and contradictory errors when using the predict and execute endpoints on remotely hosted models:

/_predict returns no handler found
/_execute on a deployed model returns a misleading private IP error even though the IP/domain is whitelisted in trusted_connector_endpoints_regex and private_ip_enabled is true. 

Private IP enabled and added the trusted endpoints 
{
  "persistent": {
    "plugins.ml_commons.connector.private_ip_enabled": true,
    "plugins.ml_commons.trusted_connector_endpoints_regex": [
      "https://.*\\.apple\\.com/.*"
    ]
  }
}

Below steps are followed to reproduce the issue 

`curl --location 'https://localhost:9200/_plugins/_ml/connectors/_create' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic YWRtaW46TXlQYXNzd29yZDEyMyE=' \
--data '{
  "name": "gemini-flash",
  "description": "Gemini 2.0 Flash connector via Apple interlinked API using predict",
  "version": "1",
  "protocol": "http",
  "path": "/",
  "parameters": {
    "model": "gemini-2.0-flash",
    "contents": [
      {
        "role": "user",
        "parts": [
          { "text": "${parameters.question}" }
        ]
      }
    ],
    "input": "Please say, Hakunamatata......",
    "temperature": 0.6
  },
  "credential": {
    "api_token": "<API-TOKEN>"
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "url": "https://interlinked.apple.com/api/v1/v1beta/models/${parameters.model}:generateContent",
      "headers": {
        "Authorization": "Bearer ${credential.api_token}",
        "Content-Type": "application/json"
      },
      "request_body": "{ \"generation_config\": { \"temperature\": ${parameters.temperature} }, \"contents\": ${parameters.contents} }"
    }
  ]
}'
`

`{
  "connector_id": "yhc4rJYBrzBYBKz7Uz7y"
}
`

`curl --location 'https://localhost:9200/_plugins/_ml/model_groups/_register' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic YWRtaW46TXlQYXNzd29yZDEyMyE=' \
--data '{
  "name": "Remote_model_group",
[...]

PR #2534, opened 2024-06-11
add setting to allow private IP

ylwu-amzn:
### Description
We have limited no private IP allowed in connector. Some user want to use private IP. Confirmed with security team, should be fine to have a setting to control allow private IP or not. By default it's disabled, user can enable it by update setting
```
PUT _cluster/settings
{
    "plugins.ml_commons.connector.private_ip_enabled": true
  }
}
```
 
### Issues Resolved
Resolve #2142
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3862, opened 2025-05-20
Fix connector private IP validation when executing agent without remote model

nathaliellenaa:
### Description
Fix private IP validation issue that occurs when executing agent with connector directly (without remote model). Initially, RemoteConnectorExecutor has an empty [setConnectorPrivateIpEnabled](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java#L159) setter which is normally set through [RemoteModel](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteModel.java#L118) during initialization. However, when agent uses connector directly, this setting remains uninitialized, causing IP [validation](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java#L142) to fail. This change will ensure proper IP validation when connector is used directly by agents or using it through remote model.

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3839

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
@@ -5,6 +5,11 @@
 
 package org.opensearch.ml.engine.algorithms.remote;
 
+import static org.opensearch.ml.common.settings.MLCommonsSettings.ML_COMMONS_CONNECTOR_PRIVATE_IP_ENABLED;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.ml.common.connector.Connector;
 import org.opensearch.ml.common.connector.ConnectorClientConfig;
 
@@ -15,6 +20,8 @@ import lombok.Setter;
 @Getter
 public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor {
     private ConnectorClientConfig connectorClientConfig;
+    protected ClusterService clusterService;
+    protected final AtomicBoolean connectorPrivateIpEnabled = new AtomicBoolean(false);
 
     public void initialize(Connector connector) {
         if (connector.getConnectorClientConfig() != null) {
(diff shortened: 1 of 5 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -23,4 +30,15 @@ public void initialize(Connector connector) {
             connectorClientConfig = new ConnectorClientConfig();
         }
     }
+
+    public boolean isConnectorPrivateIpEnabled() {
```
dhrubo-os: can we add this to the [MLFeatureEnabledSetting class](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/settings/MLFeatureEnabledSetting.java)

nathaliellenaa:
> After this change, the connectorExecutor is decided from the cluster setting so it becomes a cluster wide behavior rather than a model/connector level config. What if I want just some connector to enable the private IP?

I think the initial implementation decides whether private IP is enabled/disabled from the cluster settings rather than the model/connector level config ([reference](https://github.com/opensearch-project/ml-commons/pull/2534)). We can definitely consider this for future enhancement.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutorTest.java
```diff
@@ -90,6 +90,8 @@ public void invokeRemoteService_invalidIpAddress() {
             .actions(Arrays.asList(predictAction))
             .build();
         HttpJsonConnectorExecutor executor = new HttpJsonConnectorExecutor(connector);
+        AtomicBoolean privateIpEnabled = new AtomicBoolean(false);
```
dhrubo-os, line 93: Can we have a test where `privateIpEnabled` is enabled?

nathaliellenaa: We have this `invokeRemoteService_EnabledPrivateIpAddress` test where `privateIpEnabled` is set to `true`
https://github.com/opensearch-project/ml-commons/blob/d54f996eab13f98a7fdbf46a1346528633c589c2/ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutorTest.java#L111-L139

dhrubo-os (approved):

LeHeron-1:
Hi @nathaliellenaa , @dhrubo-os 
I'm currently facing the same issue with private IP validation when executing an agent without a remote model. I'm currently unable to upgrade the OpenSearch instance to the latest version. I'm using version 7.10.2

Is there a workaround available for this issue without upgrading ?
I have a few ideas

* At cluster level configuration 
 - i could try updating a specific cluster config to ensure the 'setConnectorPrivateIpEnabled' property is well initialized ?
* Review Agent-Connector
 - i could review the way i confugure my agent (or connector) to ensure that the property is set correctly ? ~ before the agent use the connector
 
 Would any of these ideas be feasible, or are there other workarounds you could suggest ?

Thanks you for your help !

Regards

Issue #3834, opened 2025-05-05
[BUG] MCP Server does not work with default input schema


rithin-pullela-aws:
**What is the bug?**
To register tools for MCP server, we post tools via this API: 
```
POST /_plugins/_ml/mcp/tools/_register
{
    "tools": [
        {
            "type": "ListIndexTool",
            "name": "My_ListIndexTool",
            "description": "Lists index of Opensearch CLuster A",
            "attributes": {
                "input_schema": {
                    "type": "object",
                    "properties": {
                        "input": {
                            "indices": {
                                "type": "array",
                                "items": {
                                    "type": "string"
                                },
                                "description": "OpenSearch index name list, separated by comma. for example: [\"index1\", \"index2\"], use empty array [] to list all indices in the cluster"
                            }
                        }
                    },
                    "additionalProperties": false
                }
            }
        }
}
```

All properties must be wrapped inside an "input" field within the input_schema for tools to work.

But for agents inside OpenSearch, we do not need to provide the "input" field. Ideally we should have the same interface for both the MCP Tool registration and normal Agent tool registration. 

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

zane-neo:
After a deeper look at the first category tools, only the SearchIndexTool has issue. Since [AbstractRetrieverTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java#L82-L85) and [RAGTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/RAGTool.java#L103-L105) considers `input` a string format which means `input` here is a concrete key instead of a wrapper. [McpSseTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java#L61) is a generic tool and we don't know the concrete keys so a wrapper key is mandatory.

PR #3883, opened 2025-05-30
change SearchIndexTool arguments parsing logic

zane-neo:
### Description
We want to align the schema of tools in agent and MCP, but now different tools has different ways to parse arguments from parameters map, we need to figure out how they done this and how to align based on the investigation.
**Example**
The registration request body of SearchIndexTool in ReAct agent could be: 
```
{
  "name": "SearchIndexTool",
  "type": "SearchIndexTool",
  "description": "Use this tool to search an index by providing two parameters: 'index' for the index name, and 'query' for the OpenSearch DSL formatted query. Only use this tool when both index name and DSL query is available.",
  "attributes": {
    "input_schema": {
      "type": "object",
      "properties": {
        "index": {
          "type": "string"
        },
        "query": {
          "type": "string"
        }
      },
      "additionalProperties": false
    }
  }
}
```
But in MCP tools, the tool's request body have to be like below, which has an extra layer of `input`:
```
{
  "name": "SearchIndexTool",
  "type": "SearchIndexTool",
  "description": "Use this tool to search an index by providing two parameters: 'index' for the index name, and 'query' for the OpenSearch DSL formatted query. Only use this tool when both index name and DSL query is available.",
  "attributes": {
    "input_schema": {
      "type": "object",
      "properties": {
        "input": {
          "type": "object",
          "properties": {
            "index": {
              "type": "string"
            },
            "query": {
              "type": "string"
            }
          },
          "additionalProperties": false
        }
      }
    }
  }
}

```
The reason is in ReAct agent, the LLM will parse the generated input and put actual arguments in the parameters map, but  the MCP server won't, which causing the difference, more details can found below.

#### How tools parsing arguments from `parameters` map
Based on the tools code, I can see there are three different categories that tools parsing their parameters:

**Parsing from key `input` from parameters map**
Under this category, the `input` key is treated as two different data structure:
1. [AbstractRetrieverTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java#L82-L85) which is super calss of NeuralSparseSearchTool and VectorDBTool. 
2. [RAGTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/RAGTool.java#L103-L105)
3. [McpSseTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java#L61)
4. [SearchIndexTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java#L121-L124)

**Parsing from concrete keys directly from parameters map**
1. [CreateAlertTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/CreateAlertTool.java#L138-L143)
2. [LogPatternTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/LogPatternTool.java#L111-L112)
3. [SearchAlertsTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/SearchAlertsTool.java#L74-L100)
4. [SearchAnomalyDetectorsTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java#L98-L113)
5. [SearchAnomalyResultsTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java#L88-L103)
6. [SearchMonitorsTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java#L88-L102)
7. [WebSearchTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/WebSearchTool.java#L105-L119)
8. [ConnectorTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java#L77)
9. [IndexMappingTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java#L97-L99)
10. [ListIndexTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java#L133-L147)
11. [MLModelTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/MLModelTool.java#L92)
12. [VisualizationsTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/VisualizationsTool.java#L80)

**Hybrid parsing from `input` or from concrete keys from parameters map**
1. [PPLTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/PPLTool.java#L593-L603)
2. [AgentTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java#L150-L162)
3. [CreateAnomalyDetectorTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java#L318-L332)

Some tools treats `input` as raw string like VisualizationsTool, so our target isn't to get rid of the `input` key itself, we're focusing on get rid of extra `input` if the tool's input is already a json.

For agent, the reason of unnecessary to add `input` key in the schema is because agent framework is not only extracting LLM returned tool input if it's a json, see [here](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java#L903-L906), but also add LLM returned tool input to parameters map with key `input`, see [here](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java#L920).

For MCP tools:
For cases `Parsing from concrete keys directly from parameters map` and case `Hybrid parsing from input or from concrete keys from parameters map`, they're working correctly without top level `input` key. But for case `Parsing from key input from parameters map`: it's not working well without top level `input` key as the tool parsing parameters from `input`, since it's difficult to parse the the MCP message and then transforming it to a different request body, so we can change these tools in this category to let them able to fetch both from `input` or directly from parameters map.

After a deeper look at the first category tools, only the SearchIndexTool has issue. Since [AbstractRetrieverTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java#L82-L85) and [RAGTool](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/RAGTool.java#L103-L105) considers input a string format which means input here is a concrete key instead of a wrapper. [McpSseTool](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java#L61) is a generic tool and we don't know the concrete keys so a wrapper key is mandatory.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3834

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
@@ -8,8 +8,6 @@ package org.opensearch.ml.engine.tools;
 import static org.opensearch.ml.common.CommonValue.*;
 
 import java.io.IOException;
-import java.security.AccessController;
-import java.security.PrivilegedExceptionAction;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -96,7 +94,12 @@ public class SearchIndexTool implements Tool {
 
     @Override
     public boolean validate(Map<String, String> parameters) {
-        return parameters != null && parameters.containsKey(INPUT_FIELD) && parameters.get(INPUT_FIELD) != null;
+        if (parameters == null || parameters.isEmpty()) {
+            return false;
+        }
+        boolean argumentsFromInput = parameters.containsKey(INPUT_FIELD) && parameters.get(INPUT_FIELD) != null;
+        boolean argumentsFromParameters = parameters.containsKey(INDEX_FIELD) && parameters.containsKey(QUERY_FIELD);
+        return argumentsFromInput || argumentsFromParameters;
     }
 
     private SearchRequest getSearchRequest(String index, String query) throws IOException {
@@ -120,8 +123,16 @@ public class SearchIndexTool implements Tool {
         try {
             String input = parameters.get(INPUT_FIELD);
             JsonObject jsonObject = GSON.fromJson(input, JsonObject.class);
-            String index = Optional.ofNullable(jsonObject).map(x -> x.get(INDEX_FIELD)).map(JsonElement::getAsString).orElse(null);
-            String query = Optional.ofNullable(jsonObject).map(x -> x.get(QUERY_FIELD)).map(JsonElement::toString).orElse(null);
+            String index = Optional
+                .ofNullable(jsonObject)
+                .map(x -> x.get(INDEX_FIELD))
+                .map(JsonElement::getAsString)
+                .orElse(parameters.getOrDefault(INDEX_FIELD, null));
+            String query = Optional
+                .ofNullable(jsonObject)
+                .map(x -> x.get(QUERY_FIELD))
+                .map(JsonElement::toString)
+                .orElse(parameters.getOrDefault(QUERY_FIELD, null));
             if (index == null || query == null) {
                 listener.onFailure(new IllegalArgumentException("SearchIndexTool's two parameter: index and query are required!"));
                 return;
@@ -134,10 +145,8 @@ public class SearchIndexTool implements Tool {
                 if (hits != null && hits.length > 0) {
                     StringBuilder contextBuilder = new StringBuilder();
                     for (SearchHit hit : hits) {
-                        String doc = AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> {
-                            Map<String, Object> docContent = processResponse(hit);
-                            return GSON.toJson(docContent);
-                        });
+                        Map<String, Object> docContent = processResponse(hit);
+                        String doc = GSON.toJson(docContent);
                         contextBuilder.append(doc).append("\n");
                     }
                     listener.onResponse((T) contextBuilder.toString());
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/SearchIndexToolTests.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/SearchIndexToolTests.java
@@ -95,8 +95,21 @@ public class SearchIndexToolTests {
 
     @Test
     @SneakyThrows
-    public void testValidate() {
+    public void testValidateWithInputKey() {
         Map<String, String> parameters = Map.of("input", "{}");
+        assertTrue(mockedSearchIndexTool.validate(parameters));
+    }
+
+    @Test
+    @SneakyThrows
+    public void testValidateWithActualKeys() {
+        Map<String, String> parameters = Map
+            .of(
+                SearchIndexTool.INDEX_FIELD,
+                "test-index",
+                SearchIndexTool.QUERY_FIELD,
+                "{\n" + "    \"query\": {\n" + "        \"match_all\": {}\n" + "    }\n" + "}"
+            );
         assertTrue(mockedSearchIndexTool.validate(parameters));
     }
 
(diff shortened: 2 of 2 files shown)


ylwu-amzn:
Suggest explain more details in the description with some example

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -96,7 +94,12 @@ public String getVersion() {
 
     @Override
     public boolean validate(Map<String, String> parameters) {
-        return parameters != null && parameters.containsKey(INPUT_FIELD) && parameters.get(INPUT_FIELD) != null;
+        if (parameters == null || parameters.isEmpty()) {
+            return false;
+        }
+        boolean argumentsFromInput = parameters.containsKey(INPUT_FIELD) && parameters.get(INPUT_FIELD) != null;
+        boolean argumentsFromParameters = parameters.containsKey(INDEX_FIELD) && parameters.containsKey(QUERY_FIELD);
```
Zhangxunmt, line 101: Better to check for bothel empty strings and null values
private boolean isValidValue(Map<String, String> params, String key) {
    String value = params.get(key);
    return value != null && !value.trim().isEmpty();
}
boolean argumentsFromInput = isValidValue(parameters, INPUT_FIELD);

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -96,7 +94,12 @@ public String getVersion() {
...
+        return argumentsFromInput || argumentsFromParameters;
```
Zhangxunmt, line 102: if (!argumentsFromInput && !argumentsFromParameters) {
        log.debug("Validation failed: missing required fields");
    }

add this logging before returning?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -120,8 +123,16 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
         try {
             String input = parameters.get(INPUT_FIELD);
```
Zhangxunmt, line 124: if (input != null && !input.isEmpty()) {
                try {
                    jsonObject = GSON.fromJson(input, JsonObject.class);
                } catch (JsonSyntaxException e) {
                    log.warn("Invalid JSON input: {}", input, e);
                }
            }
consider adding this null check and error handling?

Zhangxunmt (approved):

jiapingzeng (approved):
tested PER agent with SearchIndexTool, works from my side

Zhangxunmt (commented):
how about the other tools in the second and third category?

PR #3892, opened 2025-06-07
Prompt Execution via Predict

### LINKED ISSUE #3711: [FEATURE] Prompt Management in ML-Commons

rcho93:
### Description
Enables prompt execution through predict model. This PR also allow unique name across prompts

### Related Issues
Resolves #3711

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -190,12 +190,12 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
                                                     );
                                             } else {
                                                 validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
-                                                checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
+                                                checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener);
                                                 executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                             }
                                         } else {
                                             validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
-                                            checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
+                                            checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener);
                                             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                         }
                                     }
@@ -243,27 +243,23 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
         }
     }
 
-    public void checkIfPullPromptExists(
-        MLPredictionTaskRequest mlPredictionTaskRequest,
-        ActionListener<MLTaskResponse> wrappedListener,
-        String modelId
-    ) {
+    public void checkIfPullPromptExists(MLPredictionTaskRequest mlPredictionTaskRequest, ActionListener<MLTaskResponse> wrappedListener) {
         MLInputDataset inputDataset = mlPredictionTaskRequest.getMlInput().getInputDataset();
         Map<String, String> inputParameters = inputDataset instanceof RemoteInferenceInputDataSet
             ? ((RemoteInferenceInputDataSet) inputDataset).getParameters()
             : new HashMap<>();
         // prompt or messages
-        String promptOrMessages = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
+        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
             ? PARAMETERS_MESSAGES_FIELD
             : (inputParameters.containsKey(PARAMETERS_PROMPT_FIELD) ? PARAMETERS_PROMPT_FIELD : null);
-        if (!inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) || promptOrMessages == null) {
+        if (!inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) || promptType == null) {
             return;
         }
         AtomicReference<Map<String, String>> swappedParameter = new AtomicReference<>(inputParameters);
         StepListener<Map<String, String>> buildInputParameterListener = new StepListener<>();
 
         mlPromptManager
-            .buildInputParameters(promptOrMessages, inputParameters, mlPredictionTaskRequest.getTenantId(), buildInputParameterListener);
+            .buildInputParameters(promptType, inputParameters, mlPredictionTaskRequest.getTenantId(), buildInputParameterListener);
         buildInputParameterListener.whenComplete(swappedParameter::set, wrappedListener::onFailure);
 
         ((RemoteInferenceInputDataSet) inputDataset).setParameters(swappedParameter.get());
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -11,6 +11,8 @@ import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_FIELD;
 import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_PARAMETERS_FIELD;
 import static org.opensearch.ml.utils.MLExceptionUtils.LOCAL_MODEL_DISABLED_ERR_MSG;
 
+import java.util.Map;
+
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.support.ActionFilters;
@@ -52,9 +54,6 @@ import org.opensearch.transport.client.Client;
 import lombok.AccessLevel;
 import lombok.experimental.FieldDefaults;
 import lombok.extern.log4j.Log4j2;
-
-import java.util.HashMap;
-import java.util.Map;
 
 @Log4j2
 @FieldDefaults(level = AccessLevel.PRIVATE)
@@ -238,19 +237,20 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
         }
     }
 
-    private void checkIfPullPromptExists(
-            MLPredictionTaskRequest mlPredictionTaskRequest,
-            ActionListener<MLTaskResponse> wrappedListener,
-            String modelId
+    public void checkIfPullPromptExists(
+        MLPredictionTaskRequest mlPredictionTaskRequest,
+        ActionListener<MLTaskResponse> wrappedListener,
+        String modelId
     ) {
-        Map<String, String> inputParameters = new HashMap<>();
         RemoteInferenceInputDataSet inputDataSet = (RemoteInferenceInputDataSet) mlPredictionTaskRequest.getMlInput().getInputDataset();
-        inputParameters = inputDataSet.getParameters();
+        Map<String, String> inputParameters = inputDataSet.getParameters();
         // prompt or messages
-        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD) ? PARAMETERS_MESSAGES_FIELD :
-                (inputParameters.containsKey(PARAMETERS_PROMPT_FIELD) ? PARAMETERS_PROMPT_FIELD : null);
+        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
+            ? PARAMETERS_MESSAGES_FIELD
+            : (inputParameters.containsKey(PARAMETERS_PROMPT_FIELD) ? PARAMETERS_PROMPT_FIELD : null);
         if (inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) && promptType != null) {
-            mlPromptManager.buildInputParameters(
+            mlPromptManager
+                .buildInputParameters(
                     promptType,
                     inputParameters,
                     mlPredictionTaskRequest.getTenantId(),
@@ -258,7 +258,7 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
                         inputDataSet.setParameters(inputParametersAfter);
                         mlPredictionTaskRequest.getMlInput().setInputDataset(inputDataSet);
                     }, e -> wrappedListener.onFailure(e))
-            );
+                );
             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
         } else {
             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -11,6 +11,7 @@ import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_FIELD;
 import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_PARAMETERS_FIELD;
 import static org.opensearch.ml.utils.MLExceptionUtils.LOCAL_MODEL_DISABLED_ERR_MSG;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import org.opensearch.OpenSearchStatusException;
@@ -30,6 +31,7 @@ import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.exception.MLResourceNotFoundException;
 import org.opensearch.ml.common.input.MLInput;
@@ -242,8 +244,10 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
         ActionListener<MLTaskResponse> wrappedListener,
         String modelId
     ) {
-        RemoteInferenceInputDataSet inputDataSet = (RemoteInferenceInputDataSet) mlPredictionTaskRequest.getMlInput().getInputDataset();
-        Map<String, String> inputParameters = inputDataSet.getParameters();
+        MLInputDataset inputDataset = mlPredictionTaskRequest.getMlInput().getInputDataset();
+        Map<String, String> inputParameters = inputDataset instanceof RemoteInferenceInputDataSet
+            ? ((RemoteInferenceInputDataSet) inputDataset).getParameters()
+            : new HashMap<>();
         // prompt or messages
         String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
             ? PARAMETERS_MESSAGES_FIELD
@@ -255,8 +259,8 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
                     inputParameters,
                     mlPredictionTaskRequest.getTenantId(),
                     ActionListener.wrap(inputParametersAfter -> {
-                        inputDataSet.setParameters(inputParametersAfter);
-                        mlPredictionTaskRequest.getMlInput().setInputDataset(inputDataSet);
+                        ((RemoteInferenceInputDataSet) inputDataset).setParameters(inputParametersAfter);
+                        mlPredictionTaskRequest.getMlInput().setInputDataset(((RemoteInferenceInputDataSet) inputDataset));
                     }, e -> wrappedListener.onFailure(e))
                 );
             executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -13,9 +13,11 @@ import static org.opensearch.ml.utils.MLExceptionUtils.LOCAL_MODEL_DISABLED_ERR_
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
+import org.opensearch.action.StepListener;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.cluster.service.ClusterService;
@@ -189,10 +191,12 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
                                             } else {
                                                 validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
                                                 checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
+                                                executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                             }
                                         } else {
                                             validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
                                             checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
+                                            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
                                         }
                                     }
                                 }
@@ -249,24 +253,21 @@ public class TransportPredictionTaskAction extends HandledTransportAction<Action
             ? ((RemoteInferenceInputDataSet) inputDataset).getParameters()
             : new HashMap<>();
         // prompt or messages
-        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
+        String promptOrMessages = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
             ? PARAMETERS_MESSAGES_FIELD
             : (inputParameters.containsKey(PARAMETERS_PROMPT_FIELD) ? PARAMETERS_PROMPT_FIELD : null);
-        if (inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) && promptType != null) {
-            mlPromptManager
-                .buildInputParameters(
-                    promptType,
-                    inputParameters,
-                    mlPredictionTaskRequest.getTenantId(),
-                    ActionListener.wrap(inputParametersAfter -> {
-                        ((RemoteInferenceInputDataSet) inputDataset).setParameters(inputParametersAfter);
-                        mlPredictionTaskRequest.getMlInput().setInputDataset(((RemoteInferenceInputDataSet) inputDataset));
-                    }, e -> wrappedListener.onFailure(e))
-                );
-            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
-        } else {
-            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+        if (!inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) || promptOrMessages == null) {
+            return;
         }
+        AtomicReference<Map<String, String>> swappedParameter = new AtomicReference<>(inputParameters);
+        StepListener<Map<String, String>> buildInputParameterListener = new StepListener<>();
+
+        mlPromptManager
+            .buildInputParameters(promptOrMessages, inputParameters, mlPredictionTaskRequest.getTenantId(), buildInputParameterListener);
+        buildInputParameterListener.whenComplete(swappedParameter::set, wrappedListener::onFailure);
+
+        ((RemoteInferenceInputDataSet) inputDataset).setParameters(swappedParameter.get());
+        mlPredictionTaskRequest.getMlInput().setInputDataset(((RemoteInferenceInputDataSet) inputDataset));
     }
 
     private void executePredict(
--- a/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
@@ -6,6 +6,9 @@
 package org.opensearch.ml.action.prediction;
 
 import static org.opensearch.ml.common.settings.MLCommonsSettings.ML_COMMONS_MODEL_AUTO_DEPLOY_ENABLE;
+import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_MESSAGES_FIELD;
+import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_FIELD;
+import static org.opensearch.ml.prompt.MLPromptManager.PARAMETERS_PROMPT_PARAMETERS_FIELD;
 import static org.opensearch.ml.utils.MLExceptionUtils.LOCAL_MODEL_DISABLED_ERR_MSG;
 
 import org.opensearch.OpenSearchStatusException;
@@ -25,6 +28,7 @@ import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.exception.MLResourceNotFoundException;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +164,348 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
     public static boolean validateTags(List<String> tags) {
         return tags.size() <= MAX_NUMBER_OF_TAGS && tags.stream().allMatch(tag -> tag.length() <= MAX_LENGTH_OF_TAG);
     }
+
+    public void validateUniquePromptName(String name, String tenantId, ActionListener<SearchResponse> listener) {
```
akolarkunnu: Please add javadoc

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +164,348 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+            MLPrompt mlPrompt = getPrompt(getDataObjectRequest);
+            Map<String, String> promptField = mlPrompt.getPrompt();
+            // check if the specified key is defined in the prompt
+            if (!promptField.containsKey(key)) {
+                throw new IllegalArgumentException("Content for specified key is not defined in ML Prompt: " + promptId);
+            }
+            // retrieve the user-defined content during prompt creation
+            String content = promptField.get(key);
+            // populate the placeholder variable with user input, if needed
+            if (!promptParameters.isEmpty() && content.contains("${prompt_parameters.")) {
```
akolarkunnu: Better to define a constant for "${prompt_parameters." Used same text in multiple places.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +164,348 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+            // this checks if all the required input values are provided by users.
+            if (content.contains("${prompt_parameters.")) {
+                throw new IllegalArgumentException("Failed to replace all the placeholders for prompt: " + promptId);
+            }
+            return content;
+        } catch (Exception e) {
+            if (e instanceof IllegalArgumentException) {
+                throw e;
+            }
+            throw new OpenSearchStatusException("Failed to find a ML Prompt with prompt id: " + promptId, RestStatus.NOT_FOUND);
```
akolarkunnu: You are not including actual exception "e" details in this OpenSearchStatusException. Then end user will the actual root cause.

plugin/src/test/java/org/opensearch/ml/action/prediction/TransportPredictionTaskActionTests.java
```diff
@@ -20,6 +20,8 @@
 import java.util.HashSet;
 import java.util.Map;
 
+import javax.swing.*;
```
akolarkunnu: import with * is not preferred.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: okay. I will address this in my next commit

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+        try {
+            Map<String, String> parameters = new HashMap<>(inputParameters);
+            parameters.remove(PARAMETERS_PROMPT_PARAMETERS_FIELD);
+            String prompt = inputParameters.get(promptType);
+            String JsonStrPromptParameters = inputParameters.get(PARAMETERS_PROMPT_PARAMETERS_FIELD);
+            PromptParameters promptParam = PromptParameters.buildPromptParameters(JsonStrPromptParameters);
+            switch (promptType) {
+                case PARAMETERS_PROMPT_FIELD:
+                    String promptId = prompt.split("\\(")[1].split("\\)")[0];
+                    String key = prompt.split("\\.")[1];
```
Zhangxunmt: This parsing is very brittle and prone to ArrayIndexOutOfBoundsException. Consider using regex or a more robust parsing approach:
```
// More robust parsing with validation
Pattern pattern = Pattern.compile("pull_prompt\\(([^)]+)\\)\\.(.+)");
Matcher matcher = pattern.matcher(prompt);
if (!matcher.matches()) {
    throw new IllegalArgumentException("Invalid pull_prompt syntax. Expected: pull_prompt(prompt_id).key");
}
String promptId = matcher.group(1);
String key = matcher.group(2);
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+     * @param tenantId tenant id
+     * @param listener the listener to notified with new map containing modified request body
+     */
+    public void buildInputParameters(
+        String promptType,
+        Map<String, String> inputParameters,
+        String tenantId,
+        ActionListener<Map<String, String>> listener
+    ) {
+        try {
```
Zhangxunmt, line 238:
```
// Add null validation
if (promptType == null || inputParameters == null || tenantId == null || listener == null) {
    throw new IllegalArgumentException("Required parameters cannot be null");
}

String prompt = inputParameters.get(promptType);
if (prompt == null || prompt.trim().isEmpty()) {
    throw new IllegalArgumentException("Prompt cannot be null or empty");
}
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+                    parameters.put(PARAMETERS_MESSAGES_FIELD, Messages.toJsonString(messages));
+                    break;
+                default:
+                    log.error("Wrong prompt type is provided: {}, should provide either prompt or messages", promptType);
+                    throw new IllegalArgumentException(
+                        "Wrong prompt type is provided: " + promptType + ", should provide either prompt or messages"
+                    );
+            }
+            listener.onResponse(parameters);
+        } catch (Exception exception) {
```
Zhangxunmt, line 261: The exception conversion in the catch block might mask legitimate ArrayIndexOutOfBoundsException from other parts:
```
} catch (ArrayIndexOutOfBoundsException e) {
    // Be more specific about which operation failed
    throw new IllegalArgumentException(
        "Invalid pull_prompt syntax. Expected format: pull_prompt(prompt_id).key", e
    );
} catch (Exception exception) {
    log.error("Failed to build Input Parameters: ", exception);
    listener.onFailure(exception);
}
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+
+    /**
+     *  Builds a new map containing modified request body after pull_prompt is invoked
+     *
+     * @param promptType type of prompt, either prompt or messages
+     * @param inputParameters a map containing full request body received during predict
+     * @param tenantId tenant id
+     * @param listener the listener to notified with new map containing modified request body
+     */
+    public void buildInputParameters(
```
Zhangxunmt, line 232: Consider breaking down the method into smaller, testable units:
```
private String parsePromptId(String prompt) { /* ... */ }
private String parseKey(String prompt) { /* ... */ }
private void handlePromptField(Map<String, Object> parameters, String prompt, 
                              PromptParameters promptParam, String tenantId) { /* ... */ }
private void handleMessagesField(Map<String, Object> parameters, String prompt, 
                                PromptParameters promptParam, String tenantId) { /* ... */ }
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+                    "}"
+                );
+                content = substitutor.replace(content);
+            }
+            // this checks if all the required input values are provided by users.
+            if (content.contains(PROMPT_PARAMETER_PLACEHOLDER)) {
+                throw new IllegalArgumentException("Failed to replace all the placeholders");
+            }
+            return content;
+        } catch (Exception e) {
```
Zhangxunmt: better approach
```
} catch (IllegalArgumentException e) {
    throw new OpenSearchStatusException(
        "Failed to process prompt: " + promptRef + " - " + e.getMessage(),
        RestStatus.BAD_REQUEST
    );
} catch (ResourceNotFoundException e) {
    throw new OpenSearchStatusException(
        "Prompt not found: " + promptRef + " - " + e.getMessage(),
        RestStatus.NOT_FOUND
    );
} catch (IOException e) {
    throw new OpenSearchStatusException(
        "Failed to retrieve prompt: " + promptRef + " - " + e.getMessage(),
        RestStatus.INTERNAL_SERVER_ERROR
    );
} catch (Exception e) {
    log.error("Unexpected error processing prompt: {}", promptRef, e);
    throw new OpenSearchStatusException(
        "Unexpected error processing prompt: " + promptRef,
        RestStatus.INTERNAL_SERVER_ERROR
    );
}
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+     * <p>
+     *     1. fails to replace all the placeholder variables.
+     *     2. the specified key is not defined in the prompt template.
+     *     3. Failed to build a prompt or messages instance
+     *     4. Incorrect pull_prompt(prompt_id).key syntax is provided
+     *     5. Provided role does not match specified key during predict
+     * </p>
+     * @throws OpenSearchStatusException if the ML Prompt is not found
+     */
+    public String pullPrompt(String promptRef, String key, PromptParameters promptParameters, String tenantId) throws IOException {
```
Zhangxunmt: Instead of just returning a String, consider returning a more structured response
public class PromptResult {
    private final String content;
    private final String promptId;
    private final Map<String, String> appliedParameters;
    // constructors, getters
}

Zhangxunmt: The method tries to do too much in one place. Breaking it down into smaller, focused methods would improve readability, testability, and maintainability

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+    static class Messages {
+        private List<Message> messages;
+
+        @Builder(toBuilder = true)
+        public Messages(List<Message> messages) {
+            this.messages = messages;
+        }
+
+        // convert into json format string
+        public static String toJsonString(Messages messages) {
```
Zhangxunmt, line 475: Consider using a StringBuilder for large lists
Consider handling null messages parameter

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+            }
+            return StringUtils.toJson(jsonString);
+        }
+
+        public int size() {
+            return this.messages.size();
+        }
+
+        // initialize a Messages instance from json format input string
+        public static Messages buildMessages(String input) throws IOException {
```
Zhangxunmt: Consider creating custom exceptions instead of using generic ones
Add more specific error messages

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +168,361 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+        private String content;
+        private String promptId;
+        private String key;
+
+        public Message(String role, String content) {
+            this.role = role;
+            this.content = content;
+            this.promptId = this.content.split("\\(")[1].split("\\)")[0];
+            this.key = this.content.split("\\.")[1];
+        }
```
Zhangxunmt:
```
    // Should add null/empty checks
    if (role == null || content == null) {
        throw new IllegalArgumentException("Role and content cannot be null");
    }
    // Should handle cases where content doesn't match expected format
    try {
        this.role = role;
        this.content = content;
        this.promptId = this.content.split("\\(")[1].split("\\)")[0];
        this.key = this.content.split("\\.")[1];
    } catch (ArrayIndexOutOfBoundsException e) {
        throw new IllegalArgumentException("Invalid content format");
    }
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Null validation check on parameters get done before they get passed into the function. 
But, I will add some validation for prompt content as you mentioned.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: With regix and more robust parsing, this check will no long be needed. Thank you

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Got it, I will make this change in my next commit, thank you!

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -226,6 +239,36 @@ public void onFailure(Exception e) {
...
+                .buildInputParameters(
+                    promptType,
+                    inputParameters,
+                    mlPredictionTaskRequest.getTenantId(),
+                    ActionListener.wrap(inputParametersAfter -> {
+                        ((RemoteInferenceInputDataSet) inputDataset).setParameters(inputParametersAfter);
+                        mlPredictionTaskRequest.getMlInput().setInputDataset(((RemoteInferenceInputDataSet) inputDataset));
+                    }, e -> wrappedListener.onFailure(e))
+                );
+            executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
```
brianf-aws: seems predict happens regardless if condition is met?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -175,11 +188,11 @@ public void onResponse(MLModel mlModel) {
                                                     );
                                             } else {
                                                 validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
-                                                executePredict(mlPredictionTaskRequest, wrappedListener, modelId);
+                                                checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
```
brianf-aws: Can we separate this into two steps instead of trying to do 2 things in one method?

1. Parsing input if pull prompt present
2. execute predict

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -226,6 +239,36 @@ public void onFailure(Exception e) {
...
+    ) {
+        MLInputDataset inputDataset = mlPredictionTaskRequest.getMlInput().getInputDataset();
+        Map<String, String> inputParameters = inputDataset instanceof RemoteInferenceInputDataSet
+            ? ((RemoteInferenceInputDataSet) inputDataset).getParameters()
+            : new HashMap<>();
+        // prompt or messages
+        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
+            ? PARAMETERS_MESSAGES_FIELD
+            : (inputParameters.containsKey(PARAMETERS_PROMPT_FIELD) ? PARAMETERS_PROMPT_FIELD : null);
+        if (inputParameters.containsKey(PARAMETERS_PROMPT_PARAMETERS_FIELD) && promptType != null) {
```
brianf-aws: What is `PARAMETERS_PROMPT_PARAMETERS_FIELD ` ? I tried seaching but I cant seem to find it.

Also I'm seeing you added promptType I assume this Is ML or LangFuse. do you think this extra check is really needed? What occurs If I dont provide and my assumption is that I want to use regular ML commons PM?

plugin/src/main/java/org/opensearch/ml/action/prompt/TransportCreatePromptAction.java
```diff
@@ -98,6 +106,14 @@ protected void doExecute(Task task, MLCreatePromptRequest mlCreatePromptRequest,
         }
 
         try {
+            SearchResponse searchResponse = mlPromptManager
+                .validateUniquePromptName(mlCreatePromptInput.getName(), mlCreatePromptInput.getTenantId());
+            if (nameAlreadyExists(searchResponse)) {
+                SearchHit hit = searchResponse.getHits().getAt(0);
+                String id = hit.getId();
+                listener.onFailure(new IllegalArgumentException(UNIQUE_NAME_ERR_MESSAGE + id));
```
brianf-aws: user sees doc id they may get confused why "my_prompt" gets a error when I see a `abcId123` back as a error Maybe it should say that the name already exists?

plugin/src/main/java/org/opensearch/ml/action/prompt/TransportCreatePromptAction.java
```diff
@@ -98,6 +106,14 @@ protected void doExecute(Task task, MLCreatePromptRequest mlCreatePromptRequest,
         }
 
         try {
+            SearchResponse searchResponse = mlPromptManager
+                .validateUniquePromptName(mlCreatePromptInput.getName(), mlCreatePromptInput.getTenantId());
+            if (nameAlreadyExists(searchResponse)) {
```
brianf-aws: Please see my comment about `validateUniquePromptName ` in MLPromptManager class

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
     }
+
+    /**
+     * Check if a given name already exists in the prompt system index by using a term query on the name keyword field to
+     * enforce unique naming
+     *
+     * @param name name to search if it already exists
+     * @param tenantId tenant id
+     */
+    public SearchResponse validateUniquePromptName(String name, String tenantId) {
```
brianf-aws: Validate to me sounds more of a check condition is met return true or false. WHile the code tells me "Search for the index with the prompt_name with field `name`"

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
     public static boolean validateTags(List<String> tags) {
         return tags.size() <= MAX_NUMBER_OF_TAGS && tags.stream().allMatch(tag -> tag.length() <= MAX_LENGTH_OF_TAG);
     }
+
+    /**
+     * Check if a given name already exists in the prompt system index by using a term query on the name keyword field to
```
brianf-aws: Again this is more of a search operation lets be careful about the semantics, my expectation as your client is that your method will block me from creating a prompt with the same name (The javadoc and method name tell me that). But the story is something else :)

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+        }
+    }
+
+    /**
+     * Returns the boolean statement whether the name already exists or not based on retrieved Search Response
+     *
+     * @param searchResponse SearchResponse
+     * @return True if the name already exists in the prompt system index, False Otherwise.
+     */
+    public static boolean nameAlreadyExists(SearchResponse searchResponse) {
```
brianf-aws: Makes sense I'm surprised core doesnt have a default is empty method haha. Ran a search for `searchResponse.getHits().getTotalHits().value() != 0` and got no results in core

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+            if (e instanceof IndexNotFoundException || MLExceptionUtils.getRootCause(e) instanceof IndexNotFoundException) {
+                log.debug("ML Prompt index does not exist. ML Prompt name: {} will be used to init ML Prompt index", name);
+                return null;
+            }
+            throw new IllegalArgumentException("Failed to search ML Prompt index due to following cause: " + e);
+        }
+    }
+
+    /**
+     * Returns the boolean statement whether the name already exists or not based on retrieved Search Response
```
brianf-aws, line 213: This is more of a SearchResult is Empty

If we wish to accomplish the naming it would be 
```
    public static boolean MlPromptNameAlreadyExists(name, tenantId) {
               searchResponse =  searchForNameInPromptIndex(name, tenantId)
               return searchResponse != null
                          && searchResponse.getHits().getTotalHits() != null
                          && searchResponse.getHits().getTotalHits().value() != 0;.
```

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+        try {
+            Map<String, String> parameters = new HashMap<>(inputParameters);
+            parameters.remove(PARAMETERS_PROMPT_PARAMETERS_FIELD);
+            String inputContent = inputParameters.get(promptType);
+            if (inputContent == null || inputContent.trim().isEmpty()) {
+                throw new IllegalArgumentException("Missing required input: Either prompt or messages must be provided");
+            }
+            String JsonStrPromptParameters = inputParameters.get(PARAMETERS_PROMPT_PARAMETERS_FIELD);
+            PromptParameters promptParam = PromptParameters.buildPromptParameters(JsonStrPromptParameters);
+            switch (promptType) {
```
brianf-aws, line 247: I'm a bit lost on this part can you elaborate and show me an example?

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+
+            if (!matcher.matches()) {
+                throw new InvalidPullPromptSyntaxException(
+                    "Invalid pull_prompt syntax is provided: " + input + ". Expected: pull_prompt(prompt_id).key"
+                );
+            }
+            String promptId = matcher.group(1);
+            String key = matcher.group(2);
+
+            return List.of(promptId, key);
```
brianf-aws: Feels like you are looking for a tuple?

https://github.com/opensearch-project/OpenSearch/blob/0d54c1600b16081c15902a095111a9a530767099/libs/common/src/main/java/org/opensearch/common/collect/Tuple.java#L44

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
```diff
@@ -145,4 +172,500 @@ private void processResponse(GetDataObjectResponse getAsyncResponse, String prom
...
+    /**
+     * A class that represents prompt parameters.
+     */
+    static class PromptParameters {
+        /*
+            e.g. input before parsing:
+            {
+                "prompt_id":
+                {
+                    "name": "jeff"
```
brianf-aws, line 591: Nice!

brianf-aws (changes requested):
Left some comments hope it makes sense

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
rcho93: yes, if the statement is true, then the request body needs to be swapped before execute happens. Otherwise, it will just execute it.

brianf-aws: Right so do we really need to have executePredict on either branch? Is my understanding wrong?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
rcho93: It is located @ MLPromptManager:78
```
public static final String PARAMETERS_PROMPT_PARAMETERS_FIELD = "prompt_parameters";
```
The variable "promptType" refers to an unified input content for the model. This can be either prompt or messages, depending on the model's expected input format. 

If the name is confusing, I can maybe change to promptOrMessages or inputPayload?

plugin/src/main/java/org/opensearch/ml/action/prompt/TransportCreatePromptAction.java
rcho93: Yes you are right, but I also wanted to provide users the prompt Id to help them locate the conflicting prompt. But I will also include the name they provided which caused an error.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: I changed the function name to searchPromptByName to make the function more clear for the users. Let me know if  I still need better naming.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Got it! I will address this alongside revised javadoc in my next commit.

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Ahh, Make senses, I will address this in my next commit!

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Yes I agree, since we are only storing fixed number of variables that hold different meanings, I think it is better to use Tuple here. Thanks!

plugin/src/main/java/org/opensearch/ml/prompt/MLPromptManager.java
rcho93: Yes! As I mentioned in the earlier [comment ](https://github.com/opensearch-project/ml-commons/pull/3892#discussion_r2149132281), The variable promptType can either "prompt" or "messages", depending on the model's expected input format.

For example, 

openai chat model expected "messages" field in its payload whereas openai text model expects "prompt" field. 

Thus, the line of code above handles this different depending on this promptType. But I will fix this name for less confusion for readers in next commit.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
brianf-aws: Yeah this naming is a bit strange. So this `prompt_parameters` is whats actually being used as a parameters for `prompt_type` right?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -189,10 +191,12 @@ public void onResponse(MLModel mlModel) {
                                             } else {
                                                 validateInputSchema(modelId, mlPredictionTaskRequest.getMlInput());
                                                 checkIfPullPromptExists(mlPredictionTaskRequest, wrappedListener, modelId);
```
brianf-aws: Dont think model id is being used by prompt mgr right?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
```diff
@@ -249,24 +253,21 @@ public void checkIfPullPromptExists(
             ? ((RemoteInferenceInputDataSet) inputDataset).getParameters()
             : new HashMap<>();
         // prompt or messages
-        String promptType = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
+        String promptOrMessages = inputParameters.containsKey(PARAMETERS_MESSAGES_FIELD)
```
brianf-aws: Lets change back to promptType, using promptOrMessages locks us into only 2 values (semantically) that is we used 2or3 for a variable name in algebra.

brianf-aws (changes requested):
Left some slight enhancements. can you also verify your code coverage?

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
rcho93: Ahh, you are right. I will address this in my next commit. Thank!

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionTaskAction.java
rcho93: Oh prompt_parameters is used as a holder for all the  user inputs to placeholder variables. 

```
POST /_plugins/_ml/models/6wCqeZcB7X2-1Atlzb_P/_predict 
{
  "parameters": 
  {
    "prompt": "pull_prompt(weatherPrompt).user",
    "prompt_parameters": 
    {
      "weatherPrompt": 
      {
        "city": "Seattle"
      }
    }
  }
}
```

In this case the variable `promptType` would be "prompt", and 
the `prompt_parameters` would be Map.of("weatherPrompt", Map.of("city, "Seattle"))

PR #3934, opened 2025-06-20
refresh ml key

jngz-es:
### Description
Refresh ML key from the index to keep the consistence between memory and index which is under a flag. In some edge cases, we could get divergence between the key cached in memory and in the system index. Currently we only have to restart the node to re-fetch the key from the index to mitigate the issue. With the fix, we can only use a flag to control.

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
@@ -121,6 +121,18 @@ public class EncryptorImpl implements Encryptor {
         return Base64.getEncoder().encodeToString(keyBytes);
     }
 
+    // Refresh the key from the config index.
+    public void refreshMasterKey(ActionListener<Boolean> listener) {
+        try {
+            // Currently, we only handle no tenant case.
+            initMasterKeyFromIndex(null);
+            listener.onResponse(true);
+        } catch (Exception e) {
+            log.info("Refreshing ML key failed.");
+            listener.onFailure(e);
+        }
+    }
+
     private JceMasterKey createJceMasterKey(String tenantId) {
         byte[] bytes = Base64.getDecoder().decode(tenantMasterKeys.get(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID)));
         return JceMasterKey.getInstance(new SecretKeySpec(bytes, "AES"), "Custom", "", "AES/GCM/NOPADDING");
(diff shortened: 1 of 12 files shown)


jngz-es:
I am adding more IT for cover the fix.

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -121,6 +121,18 @@ public String generateMasterKey() {
         return Base64.getEncoder().encodeToString(keyBytes);
     }
 
+    // Refresh the key from the config index.
+    public void refreshMasterKey(ActionListener<Boolean> listener) {
+        try {
+            // Currently, we only handle no tenant case.
```
Zhangxunmt, line 127: Is that because the tenantId is not available when running the syncUp on each node? otherwise just fill in the tenantId in the initMasterKeyFromIndex(tenantId) it should refresh the masterKey.

Zhangxunmt (dismissed):

Zhangxunmt (dismissed):

dhrubo-os:
>In some edge cases, we could get divergence between the key cached in memory and in the system index.

Is this possible to explain this more? 

Does that mean node has a stale cached key? We create key only one time. How node's key information can be different? 

Also how important is it to refresh in a regular interval? Seems like one time refresh is required, why do we need to keep refreshing?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -100,9 +100,9 @@ public MLSyncUpCron(
     @Override
     public void run() {
         initMLConfig();
-        if (!clusterService.state().metadata().indices().containsKey(ML_MODEL_INDEX)) {
-            // no need to run sync up job if no model index
-            log.debug("Skipping sync up job - ML model index not found");
+        if (!clusterService.state().metadata().indices().containsKey(ML_MODEL_INDEX) && !mlFeatureEnabledSetting.isKeyRefreshEnabled()) {
+            // no need to run sync up job if no model index and ML key refresh disabled
+            log.info("Skipping sync up job - ML model index not found");
```
ylwu-amzn: suggest use debug level https://github.com/opensearch-project/ml-commons/issues/3947

jngz-es: Done.

PR #3946, opened 2025-06-26
MLAgentTracer Class

### LINKED ISSUE #3971: [RFC] ML-Commons Agent Tracing & Observability
### Summary
As AI agent systems become more prevalent in OpenSearch, there is an increasing need for robust tracing and observability capabilities to provide insights into agent behavior and performance. The current tracing implementation using memory in ML-Commons is limited in its ability to capture the complexity of agent workflows, especially for non-linear executions and multi-agent interactions. This RFC proposes a new approach leveraging OpenTelemetry to provide comprehensive tracing and observability for ML-Commons agents.

### Design Tenets

- Provide fine-grained tracing of agent operations, including planning, execution, and tool usage.
- Ensure that agent tasks start true root spans for accurate trace trees.
- Attach meaningful attributes (e.g., agent/task/tool/model details) to each span for better searchability and diagnostics.
- Integrate tracing with minimal performance impact and without disrupting existing agent logic.
- Allow future extension to other agent types, tools, and telemetry backends.

### Design
I propose introducing a new tracing layer for ML agents, centered around the MLAgentTracer and AbstractMLTracer classes. All agent runners (starting with MLPlanExecuteAndReflectAgentRunner) will use this tracer to create, manage, and end spans for key operations.

Image

Class Structure:
1. AbstractMLTracer
    * Base class for ML-specific tracers
    * Wraps OpenSearch's Tracer and provides utility methods for span creation and ending
2. MLAgentTracer
    * Singleton, initialized at plugin startup, extending AbstractMLTracer
    * Provides methods to start agent spans with explicit parent/child relationships
3. MLPlanExecuteAndReflectAgentRunner
    * Uses MLAgentTracer for all major operations (agent task, planning, execution steps, tool invocations, state transitions)
    * Adds rich attributes to each span (agent name, task, tool, model, etc.)
    * Maintains correct parent/child relationships between spans for trace hierarchy
4. AgentUtils
    * Provides utility methods for attribute construction and span creation

Integration with Agent Runners:
Image

Span Hierarchy:
Image

    
### Persistence
I will leverage OpenTelemetry's data collection capabilities and OpenSearch's existing infrastructure for trace persistence. All trace data will be collected via the OpenTelemetry SDK and exported through the OpenTelemetry Collector, which serves as the primary ingestion point for trace data. The collector will process and route traces to their designated storage destinations.

Traces will be stored in the otel-v1-apm-span-agent index within OpenSearch, allowing seamless integration with existing trace analytics features in OpenSearch Dashboards. This index will utilize the standard OpenTelemetry trace data model to maintain compatibility with existing tooling and visualization capabilities. However, it will also have static mappings for pre-defined span attributes. Data Prepper will handle the processing and transformation of raw OpenTelemetry data before indexing, enabling any necessary enrichment or formatting specific to ML agent traces.

Image

The persistence layer is designed with future extensibility in mind, particularly to accommodate potential integration with specialized backends for advanced trace analysis. The OpenTelemetry Collector's multi-exporter capability provides the flexibility to add support for additional storage backends as requirements evolve, without requiring changes to the core tracing implementation.

Image

### Visualization
Trace visualization will be implemented using OpenSearch Dashboards, leveraging and extending its existing trace analytics capabilities to provide intuitive, interactive views of ML agent executions. I will develop custom visualizations tailored to the unique needs of ML agent workflows, offering both high-level overviews and detailed drill-down capabilities.

The primary visualizations will be a timeline-based waterfall visualization and hierarchical tree view. This will offer a chronological perspective of the agent's operations, clearly showing the sequence and duration of each step as well as parent-child relationships. Users will be able to pan through the timeline and expand nested spans for more detailed inspection.

Complementing the list view, I will provide a hierarchical graph view. Nodes in this graph will correspond to different span types (e.g., agent tasks, planning steps, execution steps, tool calls), with edges representing the flow and dependencies between operations. This graph view will allow users to quickly grasp the overall structure of an agent's execution, including branching decisions, loops, and parallel operations.

Both visualizations will be interactive, allowing users to click on individual nodes or spans to inspect detailed attributes, such as input/output data, performance metrics, and error information.

### Implementation

1. Integrate OpenTelemetry into ML-Commons
2. Implement MLAgentTracer and AbstractMLTracer
3. Modify agent runners to use the new agent tracer 
4. Implement span creation for key agent operations
5. Add rich attribute collection for spans
6. Integrate with OpenSearch's existing OpenTelemetry exporter
7. Extend OpenSearch Dashboards for ML-specific trace visualization

- OpenTelemetry is already being utilized in OpenSearch core, query-insights, and performance analyzer, providing a solid foundation for this implementation.
- Similar solutions exist in tools like LangSmith, demonstrating proven value in this approach.
- The implementation will initially focus on the Plan-Execute-Reflect agent, with potential extension to other agent types in the future.

chriswlai:
### Description
Adds new MLAgentTracer class to capture traces within ML-commons and settings to disable or enable tracing.

Screenshot 2025-07-14 at 2 16 39 PM

### Related Issues
Resolves #3971 

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
@@ -45,6 +45,23 @@ public abstract class AbstractMLTracer {
     }
 
     /**
+     * Starts a new span for agent tracing with the specified name and attributes, and no parent span.
+     * <p>
+     * This is a convenience overload for starting a root span. It is equivalent to calling
+     * {@link #startSpan(String, Map, Span)} with {@code parentSpan} set to {@code null}.
+     * <p>
+     * The returned span should be passed to {@link #endSpan(Span)} when the operation completes.
+     *
+     * @param name The name of the span. Should follow the naming convention defined by
+     *             the span constants (e.g., AGENT_TASK_SPAN, AGENT_TOOL_CALL_SPAN).
+     * @param attributes A map of key-value pairs to associate with the span. These
+     *                  provide additional context about the operation being traced.
+     *                  May be null or empty if no attributes are needed.
+     * @return A Span object representing the started root span, or null if tracing is disabled.
+     */
+    public abstract Span startSpan(String name, Map<String, String> attributes);
+
+    /**
      * Starts a new span for tracing ML operations.
      * 
      * This method creates a new span with the specified name and attributes. The span
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
@@ -192,6 +192,25 @@ public class MLAgentTracer extends AbstractMLTracer {
     }
 
     /**
+     * Starts a new span for agent tracing with the specified name and attributes, and no parent span.
+     * <p>
+     * This is a convenience overload for starting a root span. It is equivalent to calling
+     * {@link #startSpan(String, Map, Span)} with {@code parentSpan} set to {@code null}.
+     * <p>
+     * The returned span should be passed to {@link #endSpan(Span)} when the operation completes.
+     *
+     * @param name The name of the span. Should follow the naming convention defined by
+     *             the span constants (e.g., AGENT_TASK_SPAN, AGENT_TOOL_CALL_SPAN).
+     * @param attributes A map of key-value pairs to associate with the span. These
+     *                  provide additional context about the operation being traced.
+     *                  May be null or empty if no attributes are needed.
+     * @return A Span object representing the started root span, or null if tracing is disabled.
+     */
+    public Span startSpan(String name, Map<String, String> attributes) {
+        return startSpan(name, attributes, null);
+    }
+
+    /**
      * Ends the given span.
      * 
      * This method marks the completion of a span and finalizes its timing information.


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,113 @@
...
+            return;
+        }
+        span.endSpan();
+    }
+
+    public Tracer getTracer() {
+        return tracer;
+    }
+
+    static void resetForTest() {
```
brianf-aws, line 251: Should we use a `@ VisibleForTesting` annotation also what else do we need to reset?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,113 @@
...
+            instance = null;
+            return;
+        }
+        Tracer tracerToUse = mlFeatureEnabledSetting.isAgentTracingEnabled() ? tracer : NoopTracer.INSTANCE;
+        instance = new MLAgentTracer(tracerToUse, mlFeatureEnabledSetting);
+    }
+
+    public static synchronized MLAgentTracer getInstance() {
+        if (instance == null) {
+            throw new IllegalStateException("MLAgentTracer is not initialized. Call initialize() first or check feature flag.");
```
brianf-aws: Lets be specific which feature flag needs to be flipped
`plugins.ml_commons.agent_tracing_feature_enabled` or
`plugins.ml_commons.agent_tracing_enabled`?

As a side note how can I tell what setting I need to switch on?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -109,6 +110,7 @@ public class MLAgentExecutor implements Executable, SettingsChangeListener {
     private volatile Boolean isMultiTenancyEnabled;
     private Encryptor encryptor;
     private static volatile boolean mcpConnectorIsEnabled;
+    private MLAgentTracer agentTracer;
```
mingshl: you can use getter and setter and save the method below

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
+import org.opensearch.telemetry.tracing.Span;
+import org.opensearch.telemetry.tracing.Tracer;
+
+public abstract class AbstractMLTracer {
+    protected final Tracer tracer;
+    protected final MLFeatureEnabledSetting mlFeatureEnabledSetting;
+
+    protected AbstractMLTracer(Tracer tracer, MLFeatureEnabledSetting mlFeatureEnabledSetting) {
```
mingshl, line 42: are you expecting other types of the MLTracer other than MLAgentTracer, if so, write the java doc to explain the intension of the the abstract class

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
```diff
@@ -0,0 +1,26 @@
...
+        this.tracer = tracer;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+    }
+
+    public abstract Span startSpan(String name, Map<String, String> attributes, Span parentSpan);
```
mingshl, line 82: does the the span has id? thinking if you have multiple start and end, would be helpful the start and end method would return some identifier, so that it would be helpful for future debugging. thinking if you would start multiple span, one the span failed, how do you know which span goes wrong?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+import org.opensearch.telemetry.tracing.SpanContext;
+import org.opensearch.telemetry.tracing.SpanCreationContext;
+import org.opensearch.telemetry.tracing.Tracer;
+import org.opensearch.telemetry.tracing.attributes.Attributes;
+import org.opensearch.telemetry.tracing.noop.NoopTracer;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class MLAgentTracer extends AbstractMLTracer {
```
mingshl, line 40: each class and method should have proper javadoc

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+    private static MLAgentTracer instance;
```
mingshl, line 53: why every tracer is called an instance?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+    private MLAgentTracer(Tracer tracer, MLFeatureEnabledSetting mlFeatureEnabledSetting) {
+        super(tracer, mlFeatureEnabledSetting);
+    }
+
+    public static synchronized void initialize(Tracer tracer, MLFeatureEnabledSetting mlFeatureEnabledSetting) {
+        if (mlFeatureEnabledSetting == null || !mlFeatureEnabledSetting.isTracingEnabled()) {
+            instance = null;
+            return;
+        }
+        Tracer tracerToUse = mlFeatureEnabledSetting.isAgentTracingEnabled() ? tracer : NoopTracer.INSTANCE;
```
mingshl: I would suggest that the OR condition exception should be separated in two different exceptions. so that it gives useful and easy info, when the flag is not set, throw message to set the flag. when not initialized call not initialized exception to suggest next step. so that exception is throw as early as possible and kill the rest of the process and save operations

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+                java.lang.reflect.Method createSpanMethod = tracingTelemetry
+                    .getClass()
+                    .getMethod("createSpan", SpanCreationContext.class, Span.class);
+                createSpanMethod.setAccessible(true);
+
+                newSpan = (Span) createSpanMethod.invoke(tracingTelemetry, context, null);
+
+                newSpan.addAttribute("thread.name", Thread.currentThread().getName());
+            } catch (Exception e) {
+                log.warn("Failed to create root span for agent.task, falling back to normal span creation", e);
```
mingshl: in which case that the root span failed? if you catch exception, what happens if you create manual span? is there differences between?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+                if (key != null && value != null) {
+                    attrBuilder.addAttribute(key, value);
+                }
+            }
+        }
+        SpanCreationContext context = SpanCreationContext.server().name(name).attributes(attrBuilder);
+
+        Span newSpan;
+        if ("agent.task".equals(name)) {
+            // Force agent.task spans to be root span
```
mingshl, line 161: if the root span is null, would you like to check as early as possible?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,117 @@
...
+        if (instance == null) {
+            throw new IllegalStateException(
+                "MLAgentTracer is not initialized. Call initialize() first or enable plugins.ml_commons.tracing_enabled setting."
+            );
+        }
+        return instance;
+    }
+
+    @Override
+    public Span startSpan(String name, Map<String, String> attributes, Span parentSpan) {
```
mingshl, line 147: one agent tracer would start one span only? does it have race condition here?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracerTests.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.engine.algorithms.agent.tracing;
+
+import static org.junit.Assert.*;
```
mingshl: avoid import *

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
chriswlai: Yes it does, however it is standard practice to use the Span object as the unique identifier

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: If creating a root span fails for whatever reason, the fallback will create an agent.task span that will show as having a parent span when no spans exist with that Span ID

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,321 @@
...
+     * 
+     * The method uses reflection to access the underlying tracing telemetry components,
+     * as the OpenSearch tracing API doesn't provide direct access to context propagation.
+     * If the reflection fails, the method logs a warning but doesn't throw an exception.
+     * 
+     * @param span The span whose context to inject. If null, this method is a no-op.
+     * @param carrier The map to inject context into. The span context will be added
+     *               as key-value pairs to this map. Must not be null.
+     */
+    public void injectSpanContext(Span span, Map<String, String> carrier) {
```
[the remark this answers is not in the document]
chriswlai: These are used in Plan-Execute-Reflect and Conversational agents to connect the conversational agent span with the p-e-r agent span

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
```diff
@@ -0,0 +1,321 @@
...
+                String key = entry.getKey();
+                String value = entry.getValue();
+                if (key != null && value != null) {
+                    attrBuilder.addAttribute(key, value);
+                }
+            }
+        }
+        SpanCreationContext context = SpanCreationContext.server().name(name).attributes(attrBuilder);
+        Span newSpan;
+        if (name != null && name.startsWith(AGENT_TASK_SPAN) && !(tracer instanceof NoopTracer)) {
```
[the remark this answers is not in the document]
chriswlai: We check for NoopTracer since although it gracefully handles public methods of Tracer, we use reflection to access internals (for custom span creation and injection/extraction) and NoopTracer doesn't have these so it will give a NoSuchFieldException. 

The agent execution will still run as normal, but I'm just checking to avoid the exception.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: We use instance to maintain the singleton pattern. This ensures only one instance of tracer exists which matches the standard for metrics/traces.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracerTests.java
chriswlai: fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/AbstractMLTracer.java
chriswlai: fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: good point! The new implementation avoids any null tracer or instance, so the new exception only asks to call initialize() which would then either choose NoopTracer or WrappedTracer

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: New implementation avoids null spans

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/tracing/MLAgentTracer.java
chriswlai: Tracer is designed to be thread-safe and can be used to start and end multiple spans concurrently.

mingshl (approved):

PR #3953, opened 2025-06-30
Fix exposed connector URL in error message

rithin-pullela-aws:
### Description
This pull request addresses a security concern where user input, in this case specifically the connector URL was being exposed in error messages. Exposing such information can inadvertently leak sensitive data and does not align with security best practices.

Fix:
Updated the error handling logic to avoid printing the connector URL in error messages, thereby preventing potential information disclosure.

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -182,7 +182,7 @@ public interface Connector extends ToXContentObject, Writeable {
                 }
             }
             if (!hasMatchedUrl) {
-                throw new IllegalArgumentException("Connector URL is not matching the trusted connector endpoint regex, URL is: " + url);
+                throw new IllegalArgumentException("Connector URL is not matching the trusted connector endpoint regex");
             }
         }
     }
--- a/common/src/main/java/org/opensearch/ml/common/connector/McpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/McpConnector.java
@@ -438,7 +438,7 @@ public class McpConnector implements Connector {
             }
         }
         if (!hasMatchedUrl) {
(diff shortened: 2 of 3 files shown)


mingshl:
this exception message is throwing when user try to create a model using the connectors, so we can hide the url to avoid exposing endpoint info to users, but please suggest either update the trusted endpoint in connectors or check the connector url. I think these suggestions would be helpful

rithinpullela:
> this exception message is throwing when user try to create a model using the connectors, so we can hide the url to avoid exposing endpoint info to users, but please suggest either update the trusted endpoint in connectors or check the connector url. I think these suggestions would be helpful

@mingshl, this error is thrown when we try to create the connector. So I believe the error message is good enough. `Connector URL is not matching the trusted connector endpoint regex`

rithinpullela:
I created an issue https://github.com/opensearch-project/ml-commons/issues/3954 to expand this solution to other places where we expose use input.

mingshl:
> > this exception message is throwing when user try to create a model using the connectors, so we can hide the url to avoid exposing endpoint info to users, but please suggest either update the trusted endpoint in connectors or check the connector url. I think these suggestions would be helpful
> 
> @mingshl, this error is thrown when we try to create the connector. So I believe the error message is good enough. `Connector URL is not matching the trusted connector endpoint regex`

then you can suggest to update the trusted endpoint in the cluster setting

dhrubo-os:
> > > this exception message is throwing when user try to create a model using the connectors, so we can hide the url to avoid exposing endpoint info to users, but please suggest either update the trusted endpoint in connectors or check the connector url. I think these suggestions would be helpful
> > 
> > 
> > @mingshl, this error is thrown when we try to create the connector. So I believe the error message is good enough. `Connector URL is not matching the trusted connector endpoint regex`
> 
> then you can suggest to update the trusted endpoint in the cluster setting

I think based on the input users can decide if they need to add trusted endpoint or not. What if they had some spelling mistake in their url?

PR #3579, opened 2025-02-21
[Enhancement] Enhance validation for create connector API

akolarkunnu:
### Description
This change will address the second part of validation "pre and post processing function validation".
Moved the method getRemoteServerFromURL() from ConnectorUtils.java to ConnectorAction.java, to avoid the cyclic dependency

### Related Issues
Partially resolves #2993

### Check List
- [x] New functionality includes testing.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #2986, opened 2024-09-25
[BUG] action pre_process_function does not appear to do anything at all


svdasein:
**What is the bug?**
It seems that regardless of what I say in my model definition's action section regarding pre_process_function, the effect is the same - no effect at all. In fact I have been unable to get anything to error out when I assign garbage to the property; it is as if it's simply not being used.

**How can one reproduce the bug?**
1) Set up an ollama based instance so you have an external model to talk to
2) pull the nomic-embed-text model
3) In dev tools do this:
```
POST /_plugins/_ml/models/_register
{
  "name": "nomic-embed-text",
  "function_name": "remote",
  "description": "Embedding model on myhost under Ollama",
  "connector": {
    "name": "myhostllama",
    "description": "Ollama on MyHost",
    "version": "1.0",
    "protocol": "http",
    "credential": {},
    "parameters": {

      "input_type": "text_docs",
      "model": "nomic-embed-text"
    },
    "actions": [
      {
        "action_type": "predict",
        "method": "POST",
        "headers": {
          "content-type": "application/json"
        },
        "url": "http://myhost:11434/api/embed",
        "request_body": """{"model": "${parameters.model}", "input": "${parameters.input}" }""",
        "pre_process_function": "connector.pre_process.idon'texist.embedding"
      }
    ]
  }
}

```
(You will note that it happily takes this input and registers the model despite the pre_process_function being meaningless)
![image](https://github.com/user-attachments/assets/4056c46a-c0c1-49cf-8561-b018a3f53bbd)

5) Now try some queries:
a)
```
POST /_plugins/_ml/models/<modelid>/_predict
{
  "parameters": {
    "input": "scary swedish cult"
  }
}

```
--- This works for me but only because nothing is trying to coerce the input string into an array
b)
```
POST /_plugins/_ml/models/<modelid>/_predict
{
  "parameters": {
    "input": 
      ["scary swedish cult"]
  }
}
```
--- This does not work for me - once it's in array format the framework complains that it has generated invalid json
c)
```
GET /rb-movies/_search
{
  "_source": {
    "excludes": [
      "overviewVector"
    ]
  },
  "query": {
    "bool": {
      "should": [
        {
          "neural": {
            "overviewVector": {
              "query_text": "scary swedish cult",
              "model_id": "<modelid>"",
              "k": 100
            }
          }
        }
      ]
    }
  }
}
```
--- This does not work for me - once it's in array format the framework complains that it has generated invalid json

[...]

PR #3967, opened 2025-07-08
[Enhancement] Enhance validation for create connector API - Part 3

### LINKED ISSUE #2993: [Enhancement] Enhance validation for create connector API

akolarkunnu:
### Description
This change will address the third part of validation "validation of pre and post embeddings processing functions".
If the text "embed" exists in the url, make sure pre and post processing functions are an embedding functions. If the check is false then shows a warning message in the log.
Pre and post processing functions are corresponding to llm service is already validated as part of https://github.com/opensearch-project/ml-commons/pull/3579.
Added the test case for negative scenario, positive scenarios already coverer in the existing tests of ConnectorActionTest.

### Related Issues
Resolves #2993

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -284,10 +284,8 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         String processFunction,
         String funcNameForWarnText
     ) {
-        if (endPointUrl.contains(EMBED)) {
-            if (!processFunction.contains(EMBEDDING)) {
-                logWarningForInvalidProcessFunc(remoteServer + " " + EMBEDDING, funcNameForWarnText);
-            }
+        if (endPointUrl.contains(EMBED) && !(processFunction.contains(EMBEDDING))) {
+            logWarningForInvalidProcessFunc(remoteServer + " " + EMBEDDING, funcNameForWarnText);
         }
     }
 


common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -232,7 +234,7 @@ public static String getRemoteServerFromURL(String url) {
         return SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES.stream().filter(url::contains).findFirst().orElse("");
     }
 
-    private void validateProcessFunctions(String remoteServer, String processFunction, String funcNameForWarnText) {
+    private void validateProcessFunctions(String endPointUrl, String remoteServer, String processFunction, String funcNameForWarnText) {
         if (isInBuiltProcessFunction(processFunction)) {
```
zane-neo, line 238: Based on the [issue](https://github.com/opensearch-project/ml-commons/issues/2986), I believe we should block the connector creation if the input process function doesn't exist. But based on your PRs I didn't find such validation, am I missing something?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -255,6 +257,7 @@ private void validateProcessFunctions(String remoteServer, String processFunctio
                         logWarningForInvalidProcessFunc(SAGEMAKER, funcNameForWarnText);
                     }
             }
+            validateEmbeddingProcessFunctions(endPointUrl, remoteServer, processFunction, funcNameForWarnText);
         }
     }
```
zane-neo, line 263: It's a weak validation of only checking the prefix, better to use `org.opensearch.ml.common.connector.MLPreProcessFunction#contains` to check if it's a valid precess function.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: Thanks @zane-neo for review.
Already two PRs done for fixing this bug and this is third part of it.
https://github.com/opensearch-project/ml-commons/pull/3260
https://github.com/opensearch-project/ml-commons/pull/3579.

So validation of required parameters for connector creation is done as part of https://github.com/opensearch-project/ml-commons/pull/3260 . Out of that, process functions are optional field according to https://docs.opensearch.org/docs/latest/ml-commons-plugin/remote-models/blueprints/#configuration-parameters and we didn't add any validation for it. All other parameters are also discussed in that PR.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: Yes, it's a week validation only. Otherwise, it will be hard to maintain this code, whenever we add a new pre or post functions we have to come and update this validation. We followed same strategy in https://github.com/opensearch-project/ml-commons/pull/3579 also.
Also we are just giving a warning message in the log if any validation fails, not throwing any exception. This also discussed in https://github.com/opensearch-project/ml-commons/pull/3579 .

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
zane-neo: I understand it's an optional field but the point is: if it shows up it's either a painless script or a build-in process function, not any other value. So I believe the validation should be try to compile the value to painless script, if exception occurred, then check if it's a valid value in Pre/PostProcessFunction. Without this validation, user still can pass any value to the process_function field and not error message letting them know it's not a correct value, during runtime they found the function not working at all not knowing the root cause.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
zane-neo: All the valid build-in functions are added to `MLPre/MLPostProcessFunction` internal map, and they have contains/containsKey method, invoking these methods and they'll complete the validation job, you don't need to change this every time new pre/post function added.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
akolarkunnu: @zane-neo I agree with that, that's the perfect validation here. But there is a technical limitation to implement that. As per my code walkthrough ScriptService instance is not available during the connector creation code flow. It will be available during the actual processing of  pre or post process functions(inside the class DefaultPreProcessFunction). Are there anyway I can get ScriptService instance in ConnectorActions class ?

akolarkunnu: @zane-neo Any suggestion here, please ?

Issue #3921, opened 2025-06-17
[BUG] Agent Run Failed when access key is encryted


mingshl:
**What is the bug?**
when creating an connector, register a model and then run the agent, it triggers  `java.lang.NullPointerException: host must not be null.`

**How can one reproduce the bug?**
try reproduce the issue in any agent, or follow this [doc](https://docs.opensearch.org/docs/latest/tutorials/gen-ai/agents/build-plan-execute-reflect-agent/) for plan and execute agent, it triggers the exception.

here is the stack trace

```
[2025-06-17T23:26:38,952][INFO ][o.o.m.m.i.ConversationMetaIndex] [data2] Successfully created the memory with id : a0Y3gJcB8ZcY5qRgU1Vi
[2025-06-17T23:26:38,975][INFO ][o.o.m.m.a.c.CreateInteractionTransportAction] [data2] Updating the memory a0Y3gJcB8ZcY5qRgU1Vi after the message bEY3gJcB8ZcY5qRgU1V6 is created
[2025-06-17T23:26:38,975][INFO ][o.o.m.m.i.InteractionsIndex] [data2] Successfully created the message with id : bEY3gJcB8ZcY5qRgU1V6
[2025-06-17T23:26:38,989][INFO ][o.o.m.e.a.a.MLAgentExecutor] [data2] Created parent interaction ID: bEY3gJcB8ZcY5qRgU1V6
[2025-06-17T23:26:38,991][INFO ][o.o.r.m.c.i.LocalClusterIndicesClient] [data2] Indexing data object in .plugins-ml-task
[2025-06-17T23:26:39,002][INFO ][o.o.r.m.c.i.LocalClusterIndicesClient] [data2] Creation status for id bUY3gJcB8ZcY5qRgU1WQ: CREATED
[2025-06-17T23:26:39,003][INFO ][o.o.m.e.a.a.MLAgentExecutor] [data2] Task creation result: CREATED, Task id: bUY3gJcB8ZcY5qRgU1WQ
[2025-06-17T23:26:39,018][ERROR][o.o.m.e.a.a.MLPlanExecuteAndReflectAgentRunner] [data2] Failed to run deep research agent
org.opensearch.transport.RemoteTransportException: [data1][172.31.24.244:9300][cluster:admin/opensearch/ml/predict]
Caused by: java.lang.NullPointerException: host must not be null.
	at software.amazon.awssdk.utils.Validate.paramNotNull(Validate.java:156) ~[?:?]
[...]

PR #3972, opened 2025-07-09
Add validation for creating uri in connectors

mingshl:
### Description
Add validation for creating uri in connectors  

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/3921

### Check List


dhrubo-os:
apply spotless

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -324,10 +324,20 @@ public static SdkHttpFullRequest buildSdkRequest(
             throw new IllegalArgumentException("Content length is 0. Aborting request to remote model");
         }
         String endpoint = connector.getActionEndpoint(action, parameters);
+        URI uri;
+        try {
+            uri = URI.create(endpoint);
+            if (uri.getHost() == null) {
+                throw new IllegalArgumentException("Invalid URI: " + endpoint + ". Please check if the endpoint is valid.");
+            }
+        } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
mingshl: yes, the URI.create() will throw exception saying the host cannot be null. When users saw it this native exception, it's hard to find out how to fix the problem. So I rethrow the exception with the message to indicate to check the endpoint from connector.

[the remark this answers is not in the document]
akolarkunnu: I feel you don't need to catch and re-throw IllegalArgumentException here. If host is null, you are already throwing the IllegalArgumentException here.
According to stack trace in #3921, exception happened from software.amazon.awssdk.utils.Validate.paramNotNull() , not from URI.create(). This is happening when we invoke builder.build().

 Whenever we get IllegalArgumentException from URI.create(endpoint), that is self explanatory. Eg: 
"Illegal character in path at index 4: inva lid-endpoint"

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtilsTest.java
```diff
@@ -71,19 +71,19 @@ public void processInput_TextDocsInputDataSet_NoPreprocessFunction() {
         MLInput mlInput = MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(dataSet).build();
 
         ConnectorAction predictAction = ConnectorAction
-            .builder()
-            .actionType(PREDICT)
-            .method("POST")
-            .url("http://test.com/mock")
-            .requestBody("{\"input\": \"${parameters.input}\"}")
-            .build();
+                .builder()
```
akolarkunnu: Added unnecessary tabs

mingshl: it was added by `./gradlew spotlessApply.`

mingshl: @ akolarkunnu as you can see from the stacktrace, the exception from software.amazon.awssdk.utils is not in our package, and when reaching to ConnectorUtil in ml-commons, the exception message is `host cannot be null`.  which is confusing to users. That's the intention to throw more meaningful message to customer, would you also try to mimic a invalid endpoint with a placeholder and see if the message make sense to you?

akolarkunnu: I agree with that. You are already doing that in line no: 331. 
Why you need to catch and re-throw it again line no: 334.

mingshl: was trying to specify when the uri became null. let me update the message to be clear

b4sjoo (approved):

jngz-es (approved):

Issue #4005, opened 2025-07-25
[RFC] Agentic Search in OpenSearch


mingshl:
**Is your feature request related to a problem?**
**1. Introduction**

This RFC proposes the introduction of a new query type in OpenSearch, termed "Agentic Search." This feature aims to enhance the search experience by allowing users to pose natural language questions that trigger an intelligent agent within OpenSearch to rewrite queries. This agent will leverage multiple tools, fetch documents to understand indices, perform smart query planning, generates smart queries, conduct queries, return most relevant documents related to the user's question, optionally configured to return document summary to the users questions. 

**2. Motivation**

Traditional search in OpenSearch primarily relies on keyword matching and predefined query structures. While powerful, this approach often requires users to formulate precise queries and manually sift through results to synthesize answers, especially for complex or nuanced information needs.

Agentic Search addresses these limitations by:

* Simplifying User Interaction: Users can ask questions in natural language, reducing the cognitive load of query construction.
* Enhancing Relevance: Agents can dynamically construct and refine queries, leveraging context and external tools to find more relevant information than static queries.
* Providing Summarized Answers: Beyond just document hits, users receive a direct, synthesized answer to their question, improving efficiency and user satisfaction.
* Enabling Advanced Capabilities: Integrates agentic capabilities directly into the search workflow, paving the way for more sophisticated information retrieval and knowledge synthesis.

**3. Proposed Solution: Agentic Search Query Type**

We propose a new query type, agentic_search, that will serve as the entry point for triggering an agent-driven search workflow.

![Image](https://github.com/user-attachments/assets/6f6bf2dc-1f91-4a20-a254-820d104dfa7e)

Preferred Interface

The preferred interface integrates 

agentic_search directly within the standard _search API, making it a first-class query type.

```
GET /<index_name>/_search
{
  "query": {
    "agentic_search": {
      "question": "Was Abraham Lincoln a good politician",
      "agent_id": "asn123456789".
      "fields": ["title", "description", "location"],
      "summarization": "true" //optional field
    }
  }
}
```

Explanation:

* question: The natural language question posed by the user. This is a mandatory field.
* agent_id: The identifier for a pre-configured agent within OpenSearch. This agent will be responsible for executing the search logic. This is a mandatory field.
* fields: The fields of the indices on which user wants to create DSL against
* summarization: This is an optional field which can be pass when user needs summarization of the results
    

**3. Requirements**

The Agentic Search feature should adhere to the following requirements:

[...]

PR #4006, opened 2025-07-28
Initiate query planning tool

mingshl:
### Description
This query planner tool is going to take a query text as a nature language question. Based on the question, the agent should intelligently identify relevant indices and construct appropriate queries. This includes:

    * generating queries using an LLM (Large Language Model).
    * Or Leveraging default search/query templates.
    * Or Use user defined search/query templates、

Interface：

```
{
  "name": "Test agent for embedding model",
  "type": "flow",
  "tools": [
    {
      "type": "QueryPlanningTool",
      "description": "A general query generation tool to answer any question",
      "parameters": {
        "type": "llmGenerated",
        "model_id": "h5AUWo0BkIylWTeYT4SU",
        "prompt": "\n\nHuman:You are a query generation agent for OpenSearch. You will always generate a query dsl string only.Please generate the query dsl for the question ${parameters.query_text}"
      }
    }
  ]
}
```

In this PR, we will first introduce `llmGenerated` type first, and we will introduce the rest two types when templateSearchTool is available. 

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4005

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -8,8 +8,8 @@ package org.opensearch.ml.engine.tools;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.text.StringSubstitutor;
 import org.opensearch.core.action.ActionListener;
-import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.ml.common.spi.tools.Parser;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.ml.common.spi.tools.WithModelTool;
@@ -47,8 +47,10 @@ public class QueryPlanningTool implements WithModelTool {
     @Getter
     @Setter
     private String description = DEFAULT_DESCRIPTION;
+    private String defaultQuery =
+        "{ \"query\": { \"multi_match\" : { \"query\":    \"${parameters.query_text}\",  \"fields\": ${parameters.query_fields:-[\"*\"]} } } }";
     private String defaultPrompt =
-        "You are an OpenSearch Query DSL generation assistant; using the provided index mapping ${parameters.ListIndexTool.output:-}, specified fields ${parameters.fields:-}, and the given sample queries as examples, generate an OpenSearch Query DSL to retrieve the most relevant documents for the user provided natural language question: ${parameters.query_text}\n";
+        "You are an OpenSearch Query DSL generation assistant; try using the optional provided index mapping ${parameters.index_mapping:-}, specified fields ${parameters.query_fields:-}, and the given sample queries as examples, generate an OpenSearch Query DSL to retrieve the most relevant documents for the user provided natural language question: ${parameters.query_text}, please return the query dsl only in a string format, no other texts.\n";
     @Getter
     private Client client;
     @Getter
@@ -73,13 +75,26 @@ public class QueryPlanningTool implements WithModelTool {
         if (!parameters.containsKey(PROMPT_FIELD)) {
             parameters.put(PROMPT_FIELD, defaultPrompt);
         }
-        ActionListener<List<ModelTensors>> modelListener = ActionListener.wrap(r -> {
+        if (!validate(parameters)) {
+            listener.onFailure(new IllegalArgumentException("Empty parameters for QueryPlanningTool: " + parameters));
+            return;
+        }
+        ActionListener<T> modelListener = ActionListener.wrap(r -> {
             try {
-                @SuppressWarnings("unchecked")
-                T result = (T) outputParser.parse(r);
-                listener.onResponse(result);
+                String queryString = (String) r;
+                if (queryString == null || queryString.isBlank() || queryString.isEmpty() || queryString.equals("null")) {
+                    StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
+                    String defaultQueryString = substitutor.replace(this.defaultQuery);
+                    listener.onResponse((T) defaultQueryString);
+                } else {
+                    listener.onResponse((T) queryString);
+                }
             } catch (Exception e) {
-                listener.onFailure(e);
+                IllegalArgumentException parsingException = new IllegalArgumentException(
+                    "Error processing query string: " + r + ". Try using response_filter in agent registration if needed.",
+                    e
+                );
+                listener.onFailure(parsingException);
             }
         }, listener::onFailure);
         queryGenerationTool.run(parameters, modelListener);
@@ -145,7 +160,7 @@ public class QueryPlanningTool implements WithModelTool {
                 type = LLM_GENERATED_TYPE_FIELD;
             }
 
-            // TODO to add in , SYSTEM_SEARCH_TEMPLATES_TYPE_FIELD, USER_SEARCH_TEMPLATES_TYPE_FIELD when searchTemplatesTool is
+            // TODO to add in SYSTEM_SEARCH_TEMPLATES_TYPE_FIELD, USER_SEARCH_TEMPLATES_TYPE_FIELD when searchTemplatesTool is
             // implemented.
             if (!LLM_GENERATED_TYPE_FIELD.equals(type)) {
                 throw new IllegalArgumentException("Invalid generation type: " + type + ". The current supported types are llmGenerated.");
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -10,7 +10,6 @@ import java.util.Map;
 
 import org.apache.commons.text.StringSubstitutor;
 import org.opensearch.core.action.ActionListener;
-import org.opensearch.ml.common.spi.tools.Parser;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.ml.common.spi.tools.WithModelTool;
 import org.opensearch.ml.repackage.com.google.common.annotations.VisibleForTesting;
@@ -18,7 +17,6 @@ import org.opensearch.transport.client.Client;
 
 import lombok.Getter;
 import lombok.Setter;
-import lombok.extern.log4j.Log4j2;
 
 /**
  * This tool supports different types of query planning,
@@ -26,7 +24,7 @@ import lombok.extern.log4j.Log4j2;
  * //TODO only support llmGenerated for now.
  * //TODO to add in systemSearchTemplates or userSearchTemplates when searchTemplatesTool is implemented.
  */
-@Log4j2
+
 @ToolAnnotation(QueryPlanningTool.TYPE)
 public class QueryPlanningTool implements WithModelTool {
     public static final String TYPE = "QueryPlanningTool";
@@ -35,6 +33,7 @@ public class QueryPlanningTool implements WithModelTool {
     public static final String PROMPT_FIELD = "prompt";
     private static final String GENERATION_TYPE_FIELD = "generation_type";
     private static final String LLM_GENERATED_TYPE_FIELD = "llmGenerated";
+    @Getter
     private final String generationType;
     @Setter
     @Getter
@@ -51,38 +50,26 @@ public class QueryPlanningTool implements WithModelTool {
         "{ \"query\": { \"multi_match\" : { \"query\":    \"${parameters.query_text}\",  \"fields\": ${parameters.query_fields:-[\"*\"]} } } }";
     private String defaultPrompt =
         "You are an OpenSearch Query DSL generation assistant; try using the optional provided index mapping ${parameters.index_mapping:-}, specified fields ${parameters.query_fields:-}, and the given sample queries as examples, generate an OpenSearch Query DSL to retrieve the most relevant documents for the user provided natural language question: ${parameters.query_text}, please return the query dsl only in a string format, no other texts.\n";
-    @Getter
-    private Client client;
-    @Getter
-    private String modelId;
-    @Setter
-    @Getter
-    @VisibleForTesting
-    private Parser outputParser;
-    @Setter
-    @Getter
-    private String responseField;
 
-    public QueryPlanningTool(Client client, String modelId, String generationType, MLModelTool queryGenerationTool) {
-        this.client = client;
-        this.modelId = modelId;
+    public QueryPlanningTool(String generationType, MLModelTool queryGenerationTool) {
         this.generationType = generationType;
         this.queryGenerationTool = queryGenerationTool;
     }
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        if (!parameters.containsKey(PROMPT_FIELD)) {
-            parameters.put(PROMPT_FIELD, defaultPrompt);
-        }
+
         if (!validate(parameters)) {
             listener.onFailure(new IllegalArgumentException("Empty parameters for QueryPlanningTool: " + parameters));
             return;
         }
+        if (!parameters.containsKey(PROMPT_FIELD)) {
+            parameters.put(PROMPT_FIELD, defaultPrompt);
+        }
         ActionListener<T> modelListener = ActionListener.wrap(r -> {
             try {
                 String queryString = (String) r;
-                if (queryString == null || queryString.isBlank() || queryString.isEmpty() || queryString.equals("null")) {
+                if (queryString == null || queryString.isBlank() || queryString.equals("null")) {
                     StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
                     String defaultQueryString = substitutor.replace(this.defaultQuery);
                     listener.onResponse((T) defaultQueryString);
@@ -111,16 +98,6 @@ public class QueryPlanningTool implements WithModelTool {
     }
 
     @Override
-    public String getName() {
-        return this.name;
-    }
-
-    @Override
-    public void setName(String s) {
-        this.name = s;
-    }
-
-    @Override
     public boolean validate(Map<String, String> parameters) {
         if (parameters == null || parameters.size() == 0) {
             return false;
@@ -130,8 +107,7 @@ public class QueryPlanningTool implements WithModelTool {
 
     public static class Factory implements WithModelTool.Factory<QueryPlanningTool> {
         private Client client;
-
-        private static Factory INSTANCE;
+        private static volatile Factory INSTANCE;
 
         public static Factory getInstance() {
             if (INSTANCE != null) {
@@ -165,7 +141,7 @@ public class QueryPlanningTool implements WithModelTool {
             if (!LLM_GENERATED_TYPE_FIELD.equals(type)) {
                 throw new IllegalArgumentException("Invalid generation type: " + type + ". The current supported types are llmGenerated.");
             }
-            return new QueryPlanningTool(client, (String) map.get(MODEL_ID_FIELD), type, queryGenerationTool);
+            return new QueryPlanningTool(type, queryGenerationTool);
         }
 
         @Override
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -42,7 +42,7 @@ public class QueryPlanningTool implements WithModelTool {
     @Setter
     private Map<String, Object> attributes;
     @VisibleForTesting
-    static String DEFAULT_DESCRIPTION = "Use this tool to generate query plans for a given query text.";
+    static String DEFAULT_DESCRIPTION = "Use this tool to generate opensearch query dsl for a given natural language question.";
     @Getter
     @Setter
     private String description = DEFAULT_DESCRIPTION;
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.engine.tools;
+
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+import org.opensearch.ml.common.spi.tools.WithModelTool;
+import org.opensearch.ml.repackage.com.google.common.annotations.VisibleForTesting;
+import org.opensearch.transport.client.Client;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This tool supports different types of query planning,
+ * llmGenerated, systemSearchTemplates or userSearchTemplates.
+ * //TODO only support llmGenerated for now.
+ * //TODO to add in systemSearchTemplates or userSearchTemplates when searchTemplatesTool is implemented.
+ */
+@Log4j2
+@ToolAnnotation(QueryPlanningTool.TYPE)
+public class QueryPlanningTool implements WithModelTool {
+    public static final String TYPE = "QueryPlanningTool";
+    public static final String MODEL_ID_FIELD = "model_id";
+    private final MLModelTool queryGenerationTool;
+    public static final String PROMPT_FIELD = "prompt";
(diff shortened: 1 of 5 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+    @Getter
+    @VisibleForTesting
+    private Parser outputParser;
+    @Setter
+    @Getter
+    private String responseField;
+
+    public QueryPlanningTool(Client client, String modelId, String generationType, MLModelTool queryGenerationTool) {
+        this.client = client;
+        this.modelId = modelId;
```
rithinpullela: nit:

We do not use the modelID field stored here anywhere. In the method `public QueryPlanningTool create(Map<String, Object> map) {`, we load the model ID in the ` MLModelTool queryGenerationTool`. 

And since the model ID is stored in agent Index, do we need it here?

plugin/build.gradle
```diff
@@ -225,7 +225,9 @@ integTest {
     // The --debug-jvm command-line option makes the cluster debuggable; this makes the tests debuggable
     if (System.getProperty("test.debug") != null) {
         jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'
+    jvmArgs '--add-opens', 'java.base/java.time=ALL-UNNAMED'
     }
+    jvmArgs '--add-opens', 'java.base/java.time=ALL-UNNAMED'
```
rithinpullela, line 229: Just curious, what does this line do?

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
```diff
@@ -0,0 +1,204 @@
...
+                new StringEntity(query),
+                List.of(new BasicHeader(HttpHeaders.CONTENT_TYPE, "application/json"))
+            );
+    }
+
+    private void deleteAgent(String agentId) throws IOException {
+        TestHelper.makeRequest(client(), "DELETE", "/_plugins/_ml/agents/" + agentId, null, "", List.of());
+    }
+
+    public String registerModel(String modelContent) throws IOException {
```
rithinpullela: Looks like we are not using this function, maybe remove this?

rithinpullela:
Overall looks good to me, just few nitpicks.

plugin/build.gradle
dhrubo-os: What is this for?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+        return this.name;
+    }
+
+    @Override
+    public void setName(String s) {
+        this.name = s;
+    }
+
+    @Override
+    public boolean validate(Map<String, String> parameters) {
```
dhrubo-os, line 101: is there any critical fields to validate? Also this will never trigger as in line 76, we are doing `parameters.put(PROMPT_FIELD, defaultPrompt);`. So parameters shouldn't be empty?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+            parameters.put(PROMPT_FIELD, defaultPrompt);
+        }
+        if (!validate(parameters)) {
+            listener.onFailure(new IllegalArgumentException("Empty parameters for QueryPlanningTool: " + parameters));
+            return;
+        }
+        ActionListener<T> modelListener = ActionListener.wrap(r -> {
+            try {
+                String queryString = (String) r;
+                if (queryString == null || queryString.isBlank() || queryString.isEmpty() || queryString.equals("null")) {
```
dhrubo-os: ` queryString.equals("null")` --> when this can happen?

If this is really necessary, then we can do : `if (StringUtils.isBlank(queryString) || queryString.equals("null")) {`

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+                }
+            } catch (Exception e) {
+                IllegalArgumentException parsingException = new IllegalArgumentException(
+                    "Error processing query string: " + r + ". Try using response_filter in agent registration if needed.",
+                    e
+                );
+                listener.onFailure(parsingException);
+            }
+        }, listener::onFailure);
+        queryGenerationTool.run(parameters, modelListener);
```
dhrubo-os, line 87: Should we make sure `queryGenerationTool` isn't null?

dhrubo-os: Also how long will this run? Do we need any timeout here?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+        if (parameters == null || parameters.size() == 0) {
+            return false;
+        }
+        return true;
+    }
+
+    public static class Factory implements WithModelTool.Factory<QueryPlanningTool> {
+        private Client client;
+
+        private static Factory INSTANCE;
```
dhrubo-os: add `volatile`

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+        return TYPE;
+    }
+
+    @Override
+    public String getVersion() {
+        return null;
+    }
+
+    @Override
+    public String getName() {
```
dhrubo-os: Seems like redundant if you are already using `@Setter` & `@Getter `

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This tool supports different types of query planning,
+ * llmGenerated, systemSearchTemplates or userSearchTemplates.
+ * //TODO only support llmGenerated for now.
+ * //TODO to add in systemSearchTemplates or userSearchTemplates when searchTemplatesTool is implemented.
+ */
+@Log4j2
```
dhrubo-os: Where are we using this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: If we make the LLM return a literal "null", this can be leveraged in the prompt.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: Makes sense, ideally the query planner tool will have a model ID, if there is no model ID it would eventually cause an error. Perhaps we can also validate it here?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: Just to understand, is there a reason why we need to add volatile here?

dhrubo-os: Without volatile, instruction reordering by the compiler or CPU can make this happen out of order:

A thread could see a non-null INSTANCE, but it might still point to a partially constructed object!

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: right, if queryGenerationTool is returning null, it would return default query. I do have the null check here 

https://github.com/opensearch-project/ml-commons/pull/4006/files#diff-44030799dec7f81009ffbbeb95491d8ee59583c274d997cdc25e2115a140637fR85

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
mingshl: sure

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: right I can remove here too

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
dhrubo-os: no, I'm saying what if `queryGenerationTool` object is null?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
dhrubo-os: I didn't clearly follow you. Can you give any example?

dhrubo-os:
let's add more test to cover branches. Check [here](https://app.codecov.io/gh/opensearch-project/ml-commons/pull/4006?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=opensearch-project)

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+@ToolAnnotation(QueryPlanningTool.TYPE)
+public class QueryPlanningTool implements WithModelTool {
+    public static final String TYPE = "QueryPlanningTool";
+    public static final String MODEL_ID_FIELD = "model_id";
+    private final MLModelTool queryGenerationTool;
+    public static final String PROMPT_FIELD = "prompt";
+    private static final String GENERATION_TYPE_FIELD = "generation_type";
+    private static final String LLM_GENERATED_TYPE_FIELD = "llmGenerated";
+    private final String generationType;
+    @Setter
```
owaiskazi19, line 38: You can add `@Setter`, `@Getter` at the top of the class and can remove from every vars

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+    @Getter
+    private String name = TYPE;
+    @Getter
+    @Setter
+    private Map<String, Object> attributes;
+    @VisibleForTesting
+    static String DEFAULT_DESCRIPTION = "Use this tool to generate query plans for a given query text.";
```
owaiskazi19: Why do need this here? Can move to testing

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+    @Getter
+    @Setter
+    private String description = DEFAULT_DESCRIPTION;
+    private String defaultQuery =
+        "{ \"query\": { \"multi_match\" : { \"query\":    \"${parameters.query_text}\",  \"fields\": ${parameters.query_fields:-[\"*\"]} } } }";
+    private String defaultPrompt =
+        "You are an OpenSearch Query DSL generation assistant; try using the optional provided index mapping ${parameters.index_mapping:-}, specified fields ${parameters.query_fields:-}, and the given sample queries as examples, generate an OpenSearch Query DSL to retrieve the most relevant documents for the user provided natural language question: ${parameters.query_text}, please return the query dsl only in a string format, no other texts.\n";
```
owaiskazi19, line 52: We shouldn't keep index mapping as optional, otherwise we won't provide any context to the model except the nlq
```suggestion
        "You are an OpenSearch Query DSL generation assistant; try using the provided index mapping ${parameters.index_mapping:-}, specified fields ${parameters.query_fields:-}, and the given sample queries as examples, generate an OpenSearch Query DSL to retrieve the most relevant documents for the user provided natural language question: ${parameters.query_text}, please return the query dsl only in a string format, no other texts.\n";
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,191 @@
...
+    public void setName(String s) {
+        this.name = s;
+    }
+
+    @Override
+    public boolean validate(Map<String, String> parameters) {
+        if (parameters == null || parameters.size() == 0) {
+            return false;
+        }
+        return true;
```
owaiskazi19, line 105:
```suggestion
       return parameters == null || parameters.size() == 0
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
owaiskazi19: Instead of returning "null" from LLM, we should return "N/A" or something unique

owaiskazi19: We have a method for validating params, we can move this check to a method to validate the query
if (validateQuery)...

owaiskazi19: Also, isBlank() already checks for Empty(), so we can remove that

rithinpullela: I think we decided to return the defaultQuery right?

owaiskazi19: Ah, yes.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: this is needed for other types of agents. when they will pick and run the tools by llm judgement

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: we will update this prompt later @rithin-pullela-aws is working on the default prompt. But if passing from agentic search query builder, it would have indexMapping. But this agent can execute alone without agentic search query then it might not have indexMapping.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: it's creating from the factory.create() method, it won't be null. every query planning tool will carry a queryGenerationTool

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: the model id is checked in MLModelTool.run, so we don't need to validate the model id here. I moved the validate before adding default prompt, I think that make more sense. We are expecting some parameters to run, however the field name of the parameters is not strictly named.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
ylwu-amzn: Seems this description no so accurate. What does "query plans" mean?
It converts natural language question into OpenSearch DSL query, right?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
ylwu-amzn: Seems not removed yet ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: changed to `"Use this tool to generate opensearch query dsl for a given natural language question.";`

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: removed Empty(),

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: added

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: removed in the latest commit https://github.com/opensearch-project/ml-commons/pull/4006/commits/0dd6907eda6c69382239abb0d450f69c90e96a4b

plugin/build.gradle
mingshl: we need this for converting time to string. removing that line would cripple the project's ability to handle any JSON data that involves modern Java date/time objects, which is a very common requirement.

dhrubo-os: How so? Can you please explain this a bit more? I'm still not sure why we need this? Are you trying to import the time class in any test class where it's throwing error?

rithin-pullela-aws (approved):
LGTM! Thank you!

rithinpullela:
The failing UTs are being fixed in this PR: https://github.com/opensearch-project/ml-commons/pull/4037

plugin/build.gradle
ylwu-amzn: From AI

### Why is this needed?

- JPMS restrictions: Starting with Java 9+, access to internal classes is restricted. Many libraries, frameworks, or test utilities (like Mockito, Jackson, etc.) may use reflection to access private or package-private APIs in java.time. Without --add-opens, such access can fail with IllegalAccessException.
- Testing scenarios: If your integration tests or dependencies use such reflective access, you may encounter errors unless you explicitly open the module/package.

### Is it really necessary?

- If your tests run fine without it: You may not need this.
- If you see errors about illegal reflective access or module access violations involving java.time: Then this argument is necessary.
- Often added for compatibility: Sometimes, this is added preemptively for compatibility with CI environments or when using certain mocking frameworks that are known to access JDK internals.

### How to decide?

- Try running your tests without it. If everything passes and there are no warnings/errors about illegal access to java.time, you can remove it.
- If you hit reflective access errors, add it back.

### Summary:
It’s only necessary if you encounter module access violations relating to java.time during your tests. Otherwise, it can be omitted. Adding it does not affect production code, only the test JVM arguments.

ylwu-amzn: @mingshl 
> Try running your tests without it. If everything passes and there are no warnings/errors about illegal access to java.time, you can remove it.

Do you see any error without this line?

owaiskazi19: Are we seeing reflection errors like the below if removing it? 
`java.lang.reflect.InaccessibleObjectException: module java.base does not "opens java.time" to unnamed module `

ylwu-amzn:
Have you tried Bedrock Claude 3.7 ? The response looks not clean. I think we need to filter query from the response.
```
{
    "inference_results": [
        {
            "output": [
                {
                    "name": "response",
                    "result": "{\"metrics\":{\"latencyMs\":1804.0},\"output\":{\"message\":{\"content\":[{\"text\":\"{\\n  \\\"size\\\": 0,\\n  \\\"query\\\": {\\n    \\\"match_all\\\": {}\\n  },\\n  \\\"aggs\\\": {\\n    \\\"document_count\\\": {\\n      \\\"value_count\\\": {\\n        \\\"field\\\": \\\"_index\\\"\\n      }\\n    }\\n  }\\n}\"}],\"role\":\"assistant\"}},\"stopReason\":\"end_turn\",\"usage\":{\"cacheReadInputTokenCount\":0.0,\"cacheReadInputTokens\":0.0,\"cacheWriteInputTokenCount\":0.0,\"cacheWriteInputTokens\":0.0,\"inputTokens\":56.0,\"outputTokens\":71.0,\"totalTokens\":127.0}}"
                }
            ]
        }
    ]
}
```

Update: need to add `"response_filter": "$.output.message.content[0].text"` to QueryPlanningTool parameters

mingshl:
> Have you tried Bedrock Claude 3.7 ? The response looks not clean. I think we need to filter query from the response.
> 
> ```
> 
> {
> 
>     "inference_results": [
> 
>         {
> 
>             "output": [
> 
>                 {
> 
>                     "name": "response",
> 
>                     "result": "{\"metrics\":{\"latencyMs\":1804.0},\"output\":{\"message\":{\"content\":[{\"text\":\"{\\n  \\\"size\\\": 0,\\n  \\\"query\\\": {\\n    \\\"match_all\\\": {}\\n  },\\n  \\\"aggs\\\": {\\n    \\\"document_count\\\": {\\n      \\\"value_count\\\": {\\n        \\\"field\\\": \\\"_index\\\"\\n      }\\n    }\\n  }\\n}\"}],\"role\":\"assistant\"}},\"stopReason\":\"end_turn\",\"usage\":{\"cacheReadInputTokenCount\":0.0,\"cacheReadInputTokens\":0.0,\"cacheWriteInputTokenCount\":0.0,\"cacheWriteInputTokens\":0.0,\"inputTokens\":56.0,\"outputTokens\":71.0,\"totalTokens\":127.0}}"
> 
>                 }
> 
>             ]
> 
>         }
> 
>     ]
> 
> }
> 
> ```
> 
> 
> 
> Update: need to add `"response_filter": "$.output.message.content[0].text"` to QueryPlanningTool parameters

It works on Claude 3.7. You can use the response_filter to filter out the response from model response 

```

POST {{host}}/_plugins/_ml/agents/_register
{
  "name": "Agentic Search with Claude 3.7",
  "type": "flow",
  "tools": [
    {
      "type": "QueryPlanningTool",
      "description": "A general tool to answer any question",
      "parameters": {
        "model_id": "O4rMUpgBQXKIv6OARn9w",
        "response_filter": "$.output.message.content[0].text"
        
      }
    }
  ]
}



```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -0,0 +1,167 @@
...
+ * //TODO only support llmGenerated for now.
+ * //TODO to add in systemSearchTemplates or userSearchTemplates when searchTemplatesTool is implemented.
+ */
+
+@ToolAnnotation(QueryPlanningTool.TYPE)
+public class QueryPlanningTool implements WithModelTool {
+    public static final String TYPE = "QueryPlanningTool";
+    public static final String MODEL_ID_FIELD = "model_id";
+    private final MLModelTool queryGenerationTool;
+    public static final String PROMPT_FIELD = "prompt";
```
ylwu-amzn, line 33: I think we also support `system_prompt`, so user can customize system prompt in QueryPlanningTool

rithinpullela: This PR https://github.com/opensearch-project/ml-commons/pull/4046 adds the default system prompt. Can you please take a look?

mingshl: right, we should merge this first, @rithin-pullela-aws 's PR to add in prompt and feature flag is depending on this one.

plugin/build.gradle
mingshl: yes, @ylwu-amzn removing it would cause the same problem as @owaiskazi19 saw

mingshl: as adding it doesn't impact the production codes, that should be fine.

mingshl: Let me try to explain in detail. 

when I am running the IT test, I am trying to cast the agent execute response into string so that I can verify the query string.

`Map<String, Object> responseMap = gson.fromJson(responseBody, Map.class);`

then when casting the time to string, java needs to specify the argument from how to cast the time, this argument is added only for running IT tests. So the time can cast to string properly. 

I can remove this IT assert, but I don't see any harm adding this argument when running IT tests, because we can cast other time to string when asserting string, this will come in handy @dhrubo-os

dhrubo-os: Approving for now to unblock @rithin-pullela-aws , but I'll look into this more.

PR #4015, opened 2025-07-30
implement create and get memory container API

b4sjoo:
### Description
CREATE & GET API for memory container API

### Related Issues

### Check List


common/src/main/resources/index-mappings/ml_memory_container.json
```diff
@@ -0,0 +1,54 @@
+{
+  "_meta": {
+    "schema_version": 1
+  },
+  "properties": {
+    "container_id": {
```
ylwu-amzn: Every document in index has its own id, any difference between the doc id and this `container_id` ?

https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/resources/index-mappings/ml_model.json, we don't have `model_id` in index.

common/src/main/resources/index-mappings/ml_memory_container.json
```diff
@@ -0,0 +1,54 @@
...
+      "type": "keyword"
+    },
+    "container_name": {
```
ylwu-amzn: Suggest follow current naming convention of model/agent, just use `name`
```suggestion
    "name": {
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryContainer.java
```diff
@@ -0,0 +1,213 @@
...
+    public MLMemoryContainer(
+        String containerId,
+        String containerName,
+        String description,
+        User owner,
+        String tenantId,
+        Instant createdTime,
+        Instant lastUpdatedTime,
+        String indexName,
+        SemanticStorageConfig semanticStorage
```
dhrubo-os: Don't we need a llm_model_id? This model will be used to extract the facts. The model id inside semantic storage is to generate embedding.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,66 @@
...
+    public static final String FACT_ENCODING_FIELD = "fact_encoding";
+    public static final String TAGS_FIELD = "tags";
+
+    // KNN index settings
+    public static final String KNN_ENGINE = "faiss";
+    public static final String KNN_SPACE_TYPE = "cosinesimil";
+    public static final String KNN_METHOD_NAME = "hnsw";
+    public static final int KNN_EF_SEARCH = 100;
+    public static final int KNN_EF_CONSTRUCTION = 128;
+    public static final int KNN_M = 24;
```
dhrubo-os: Are these default configuration for K-nn index?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // REST API paths
+    public static final String BASE_MEMORY_CONTAINER_PATH = "/_plugins/_ml/memory_container";
+    public static final String CREATE_MEMORY_CONTAINER_PATH = BASE_MEMORY_CONTAINER_PATH + "/_create";
+    public static final String PARAMETER_CONTAINER_ID = "container_id";
```
dhrubo-os: should we call this `mem_container_id` just to be specific?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,66 @@
...
+    // Field names for MemoryContainer
+    public static final String CONTAINER_ID_FIELD = "container_id";
+    public static final String CONTAINER_NAME_FIELD = "container_name";
+    public static final String DESCRIPTION_FIELD = "description";
+    public static final String OWNER_FIELD = "owner";
+    public static final String CREATED_TIME_FIELD = "created_time";
+    public static final String LAST_UPDATED_TIME_FIELD = "last_updated_time";
+    public static final String INDEX_NAME_FIELD = "index_name";
+    public static final String SEMANTIC_STORAGE_FIELD = "semantic_storage";
+    public static final String TENANT_ID_FIELD = "tenant_id";
```
dhrubo-os: let's reuse this: https://github.com/opensearch-project/ml-commons/blob/7040e8a0e6bb143e7b899daf4951d7dc0d5bc1b9/common/src/main/java/org/opensearch/ml/common/CommonValue.java#L25

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // Field names for SemanticStorageConfig
+    public static final String SEMANTIC_STORAGE_ENABLED_FIELD = "semantic_storage_enabled";
+    public static final String MODEL_TYPE_FIELD = "model_type";
+    public static final String MODEL_ID_FIELD = "model_id";
```
dhrubo-os: let's call this embedding_model_id so that we can have clear distinction between embedding model id vs llm model id.

common/src/main/java/org/opensearch/ml/common/memorycontainer/SemanticStorageConfig.java
```diff
@@ -0,0 +1,122 @@
...
+@EqualsAndHashCode
+public class SemanticStorageConfig implements ToXContentObject, Writeable {
+
+    private boolean semanticStorageEnabled;
+    private FunctionName modelType;
+    private String modelId;
+    private Integer dimension;
+
+    public SemanticStorageConfig(boolean semanticStorageEnabled, FunctionName modelType, String modelId, Integer dimension) {
+        this.semanticStorageEnabled = semanticStorageEnabled;
```
dhrubo-os: Trying to understand the significance of this field. Let's say customer had false in this field but provided model and dimension. What will we do?

I think better to have a field to identify if we should create a k-nn index or regular index?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerRequest.java
```diff
@@ -0,0 +1,108 @@
...
+            if (semanticStorage.getModelType() == null) {
+                exception = addValidationError(SEMANTIC_STORAGE_MODEL_TYPE_REQUIRED_ERROR, exception);
+            } else {
+                FunctionName modelType = semanticStorage.getModelType();
+                if (modelType != FunctionName.TEXT_EMBEDDING && modelType != FunctionName.SPARSE_ENCODING) {
+                    exception = addValidationError(INVALID_MODEL_TYPE_ERROR, exception);
+                }
+
+                // Model ID is required when semantic storage is enabled
+                if (semanticStorage.getModelId() == null) {
```
dhrubo-os: For simplicity I was wondering if we should use the LLM_model_id to generate the embedding if customer doesn't provide the embedding_model_id? What do you think?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerRequest.java
```diff
@@ -0,0 +1,108 @@
...
+                    exception = addValidationError(SEMANTIC_STORAGE_MODEL_ID_REQUIRED_ERROR, exception);
+                }
+
+                // Validate dimension requirements
+                if (modelType == FunctionName.TEXT_EMBEDDING) {
+                    // Dimension is required for TEXT_EMBEDDING
+                    if (semanticStorage.getDimension() == null) {
+                        exception = addValidationError(TEXT_EMBEDDING_DIMENSION_REQUIRED_ERROR, exception);
+                    }
+                } else if (modelType == FunctionName.SPARSE_ENCODING) {
```
dhrubo-os: Do we need to throw this error? Can't we just ignore?

common/src/main/resources/index-mappings/ml_memory_container.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "created_time": {
+      "type": "date",
+      "format": "strict_date_time||epoch_millis"
+    },
+    "last_updated_time": {
+      "type": "date",
+      "format": "strict_date_time||epoch_millis"
+    },
+    "index_name": {
```
dhrubo-os: should we say `memory_index_name`?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -0,0 +1,292 @@
...
+        String tenantId = input.getTenantId();
+
+        // Check if memory container index exists, create if not
+        ActionListener<Boolean> indexCheckListener = ActionListener.wrap(created -> {
+            try {
+                // Generate container ID
+                String containerId = UUID.randomUUID().toString();
+
+                // Create memory container document
+                MLMemoryContainer memoryContainer = createMemoryContainer(input, containerId, user, tenantId);
```
dhrubo-os: can we rename this method? This is just to build the object?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -0,0 +1,292 @@
...
+            .indexName(input.getIndexName())
+            .semanticStorage(input.getSemanticStorage())
+            .build();
+    }
+
+    private void createMemoryDataIndices(MLMemoryContainer container, User user, ActionListener<Boolean> listener) {
+        String userId = user != null ? user.getName() : "default";
+        String baseIndexName = container.getIndexName();
+
+        if (baseIndexName == null) {
```
dhrubo-os, line 187: I don't think we should update the naming. What if customer setup roles for index `abc` but now created another index `abc+containerid-dhrubo`. 

I think we should create the index with the same name user provided.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -0,0 +1,292 @@
...
+            } else if (semanticStorage.getModelType() == FunctionName.SPARSE_ENCODING) {
+                baseIndexName = SPARSE_MEMORY_INDEX_PREFIX + container.getContainerId() + "-" + userId;
+            }
+        }
+
+        // Create the memory data index with appropriate mapping
+        createMemoryDataIndex(baseIndexName, container.getSemanticStorage(), listener);
+    }
+
+    private void createMemoryDataIndex(String indexName, SemanticStorageConfig semanticStorage, ActionListener<Boolean> listener) {
```
dhrubo-os: one way could be we can check first if the index exists or not? If not exist then we can try to create the index?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -0,0 +1,292 @@
...
+        try {
+            XContentBuilder builder = XContentType.JSON.contentBuilder();
+            container.toXContent(builder, ToXContent.EMPTY_PARAMS);
+
+            IndexRequest indexRequest = new IndexRequest(ML_MEMORY_CONTAINER_INDEX)
+                .id(container.getContainerId())
+                .source(builder)
+                .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+            client.index(indexRequest, ActionListener.wrap(response -> {
```
dhrubo-os: can we please follow the sdkclient approach?

common/src/main/resources/index-mappings/ml_memory_container.json
ylwu-amzn: Can we move this inside the `semantic_storage` ?

common/src/main/resources/index-mappings/ml_memory_container.json
```diff
@@ -0,0 +1,54 @@
...
+      "type": "keyword"
+    },
+    "semantic_storage": {
```
ylwu-amzn: How about renaming this to be more generic ? For example, some user may want to use the vanilla index rather than K-NN.  Maybe just `storage` or other generic name?

dhrubo-os: `memory_storage`?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Not default, recommended by KNN team

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryContainer.java
b4sjoo: Added

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Doesn't make sense, user can specify raw memory index name.

b4sjoo: > should we say `memory_index_name`?

Renamed.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: This is just for the case user does not specify any index name. Tested in the cluster.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: Creating an index which is already existed exception is handled by OpenSearch core.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: Please guide

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Renamed

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerRequest.java
b4sjoo: Better not, chat model's response is not stable enough, we never know if llm's answer contains other words. For instance:
```
{"message": "Sure! Let me generate the embedding for you: [0.52315,...,1.313143]"}
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerRequest.java
b4sjoo: No, this is the number we used to generate KNN index

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: Renamed to `buildMemoryContainer`

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Refactored to `memory_storage_config`. We used a flag `semantic_storage_enabled` and `embedding_model_type` to control the index type.

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: > Can we move this inside the `semantic_storage` ?

Moved

common/src/main/java/org/opensearch/ml/common/memorycontainer/SemanticStorageConfig.java
b4sjoo: This flag field is important if we want to enable default embedding model. At the time when user don't specify embedding_model_type and embedding_model_id but has `semanticStorageEnabled` to `true` we can set up a sparse model for user in serverless, just like 1-click semantic search.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Deleted and use the commonValue

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Renamed

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportGetMemoryContainerAction.java
```diff
@@ -0,0 +1,226 @@
...
+                            RestStatus.NOT_FOUND
+                        )
+                    );
+            }
+        } catch (Exception e) {
+            wrappedListener.onFailure(e);
+        }
+    }
+
+    private void validateMemoryContainerAccess(
```
nathaliellenaa, line 184: Should we move this function to an AccessControlHelper class, so it can be reused by other operations?

b4sjoo: You can refactor those to simplify your work.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: Oh I found how to write it. Changed.

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Check [L36](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/resources/index-mappings/ml_model.json#L36), we have.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Changed to memory_container_id

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: And we caught this exception as proof of index exists.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryContainer.java
```diff
@@ -49,6 +51,7 @@ public class MLMemoryContainer implements ToXContentObject, Writeable {
     private Instant createdTime;
     private Instant lastUpdatedTime;
     private MemoryStorageConfig memoryStorageConfig;
+    private List<String> modelIdsMonitoring;
```
dhrubo-os: What is this for?

b4sjoo: Discussed offline, will remove

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Removed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -0,0 +1,284 @@
...
+@Setter
+@Builder
+public class MLMemory implements ToXContentObject, Writeable {
+
+    // Core fields
+    private String memoryId;
+    private String sessionId;
+    private String memory;
+    private MemoryType memoryType;
+    private MemoryCharacteristic memoryCharacteristic;
```
dhrubo-os: I don't think we need this. Let's stick to the plan we discussed.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,97 @@
...
+
+    // Response fields
+    public static final String STATUS_FIELD = "status";
+
+    // Error messages
+    public static final String SEMANTIC_STORAGE_EMBEDDING_MODEL_TYPE_REQUIRED_ERROR =
+        "Embedding model type is required when semantic storage is enabled";
+    public static final String SEMANTIC_STORAGE_EMBEDDING_MODEL_ID_REQUIRED_ERROR =
+        "Embedding model ID is required when semantic storage is enabled";
+    public static final String SEMANTIC_STORAGE_LLM_MODEL_ID_REQUIRED_ERROR = "LLM model ID is required when semantic storage is enabled";
```
dhrubo-os, line 80: we can just extract fact and store as a message without embedding, right? Does this need to have a strong dependency?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStorageConfig.java
```diff
@@ -0,0 +1,272 @@
...
+     * Validates the memory storage configuration based on semantic storage settings
+     * and enforces field restrictions and limits.
+     */
+    public void validate() {
+        if (!semanticStorageEnabled) {
+            // When semantic storage is disabled, only allow specific fields
+            // Clear fields that aren't allowed
+            this.embeddingModelType = null;
+            this.embeddingModelId = null;
+            this.llmModelId = null;
```
dhrubo-os: My understanding is:

If customer want to have semantic search, then they will enable this field. And they will provide embedding model to embed the memory so that we can perform k-nn search.

The basic functionality of extracting the fact from user message and storing them as a long term memory is still in place. But embedding will not be there if `semanticStorageEnabled` is false. And during this time we will just perform basic search. 

Is this not the case? Are you thinking differently?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package org.opensearch.ml.common.memorycontainer;
+
+/**
+ * Enum representing the type of memory entry
+ */
+public enum MemoryType {
+    RAW_MESSAGE("RAW_MESSAGE"),
+    FACT("FACT");
```
dhrubo-os, line 13: https://github.com/mem0ai/mem0/blob/7d1d0ca8063d03e58265dacd7246229696b1260e/mem0/configs/enums.py#L4

What do you think about following this?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -0,0 +1,97 @@
...
+    public static final String TEXT_EMBEDDING_DIMENSION_REQUIRED_ERROR = "Dimension is required for TEXT_EMBEDDING";
+    public static final String SPARSE_ENCODING_DIMENSION_NOT_ALLOWED_ERROR = "Dimension is not allowed for SPARSE_ENCODING";
+    public static final String INVALID_EMBEDDING_MODEL_TYPE_ERROR = "Embedding model type must be either TEXT_EMBEDDING or SPARSE_ENCODING";
```
dhrubo-os, line 83: What if they want to use `REMOTE` like bedrock Titan embedding model or a embedding model is hosted in sagemaker?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * Represents a memory entry in a memory container
+ */
+@Getter
+@Setter
+@Builder
+public class MLMemory implements ToXContentObject, Writeable {
+
+    // Core fields
+    private String memoryId;
```
rithin-pullela-aws, line 36: Do we need memory ID, this will be doc ID in the index right?

rithinpullela: Do we need memory ID, this will be doc ID in the index right?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoryInput.java
```diff
@@ -0,0 +1,211 @@
...
+@Builder
+public class MLAddMemoryInput implements ToXContentObject, Writeable {
+
+    // Required fields
+    private String memoryContainerId;
+    private String memory; // Internally stored as 'memory', but exposed as 'message' in API
+
+    // Optional fields
+    private MemoryType memoryType;
+    private String memoryId;
```
rithinpullela: Looks like memory_id does not correspond to the doc ID, is that expected? I believe it should be the doc ID after indexing.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
```diff
@@ -0,0 +1,659 @@
...
+                return;
+            }
+
+            // Validate and determine infer value based on semantic storage
+            MemoryStorageConfig storageConfig = container.getMemoryStorageConfig();
+            boolean isSemanticEnabled = storageConfig != null && storageConfig.isSemanticStorageEnabled();
+            Boolean infer = input.getInfer();
+
+            if (!isSemanticEnabled) {
+                // Semantic storage disabled: infer must be false or null
```
rithinpullela: Do we need to have a tight coupling between infer and semantic storage?
Please correct me if I am wrong, but my assumption is, if infer is true then we extract facts via the LLM call and if infer is false we skip the fact extraction and directly store the raw messages. So I do not see a tight coupling between infer and the semantic storage. It makes sense to have infer coupled with the presence of an LLM Model ID to perform fact extraction.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
```diff
@@ -0,0 +1,659 @@
...
+            }
+
+            // Auto-determine memory type and characteristic based on infer
+            MemoryType memoryType = input.getMemoryType();
+            MemoryCharacteristic memoryCharacteristic;
+
+            if (!infer) {
+                // When infer is false: only RAW_MESSAGE allowed, characteristic is LONG_TERM
+                if (memoryType != null && memoryType != MemoryType.RAW_MESSAGE) {
+                    actionListener.onFailure(new IllegalArgumentException("When infer is false, only RAW_MESSAGE memory type is allowed"));
```
rithinpullela: Does that mean the user input for memory type can be FACT as well? I think we need to determine if the memory being stored is fact or raw message.

Eventually when we allow storing a different memory type  like an SOP or something, then we need to use the memoryType value. 

I think for now any value other than null and raw_message should throw an exception irrespective of the infer value. Looks like we don't throw an error if infer is true.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
```diff
@@ -0,0 +1,659 @@
...
+                memoryCharacteristic = MemoryCharacteristic.SHORT_TERM;
+            }
+
+            // Generate IDs - memory_id now serves as the unique message identifier
+            String memoryId = input.getMemoryId();
+            String sessionId = input.getSessionId();
+
+            if (memoryId == null) {
+                memoryId = UUID.randomUUID().toString();
+            }
```
rithinpullela: So the memoryID is not the doc ID? I believe the design suggests that the memoryId is the doc ID

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
```diff
@@ -0,0 +1,659 @@
...
+                    .onFailure(
+                        new OpenSearchStatusException("Memory container does not have a valid index name", RestStatus.INTERNAL_SERVER_ERROR)
+                    );
+                return;
+            }
+
+            // If sessionId not provided but memoryId exists, look up its sessionId
+            if (sessionId == null && input.getMemoryId() != null) {
+                searchSessionId(indexName, finalMemoryId, ActionListener.wrap(existingSessionId -> {
+                    String finalSessionId = existingSessionId != null ? existingSessionId : "sess_" + UUID.randomUUID().toString();
```
rithinpullela: Assuming memory ID is another field and not the doc ID, looks like we are trying to associate a session ID with the memoryID. For a given memory ID we are trying to find the corresponding sessionID. i.e we are tightly coupling the session ID with the memoryID. Is there a reason for doing this? I think this deviates from the design.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -0,0 +1,435 @@
...
+            properties.put(ROLE_FIELD, Map.of("type", "text")); // Text field for role (human/llm)
+            properties.put(CREATED_TIME_FIELD, Map.of("type", "date", "format", "strict_date_time||epoch_millis"));
+            properties.put(LAST_UPDATED_TIME_FIELD, Map.of("type", "date", "format", "strict_date_time||epoch_millis"));
+
+            if (memoryStorageConfig != null && memoryStorageConfig.isSemanticStorageEnabled()) {
+
+                if (memoryStorageConfig.getEmbeddingModelType() == FunctionName.TEXT_EMBEDDING) {
+                    // KNN index configuration
+                    indexSettings.put("index.knn", true);
+                    indexSettings.put("index.knn.algo_param.ef_search", KNN_EF_SEARCH);
```
rithinpullela, line 232: I was wondering, if we are taking a embeddingModelId as an input during create container operation, can we not use semantic type field and associate the model directly, instead of creating the whole index mapping

https://docs.opensearch.org/latest/field-types/supported-field-types/semantic/

rithin-pullela-aws (commented):
Looks like we deviated a little from our design from before, posted some high level questions to understand the proposed design

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
b4sjoo: Correct will remove

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoryInput.java
b4sjoo: Correct will change

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
b4sjoo: But LLM model is tightly stick to semantic storage flag? So I still insist the logic here

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
b4sjoo: Fact / raw_message should be controlled by infer flag. The only type user use can make some impact on the memory generation process is procedure. I am pretty sure Mem0 use this flow

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
b4sjoo: Will change

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
b4sjoo: Semantic does not support cosine similarity now. I checked with their team

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoryAction.java
b4sjoo: Right this is wrong, I will change it to the most straightforward flow.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: It has a strong dependency, we need semantic search

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: We allowed REMOTE and TEXT_EMBEDDING/SPARSE

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
b4sjoo: Will remove

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoryInput.java
```diff
@@ -0,0 +1,211 @@
...
+@Setter
+@Builder
+public class MLAddMemoryInput implements ToXContentObject, Writeable {
+
+    // Required fields
+    private String memoryContainerId;
+    private String memory; // Internally stored as 'memory', but exposed as 'message' in API
+
+    // Optional fields
+    private MemoryType memoryType;
```
rithinpullela: Remove this as we do not accept this as input

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java
b4sjoo: Discussed offline, stick to current right now

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Resolved offline

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStorageConfig.java
b4sjoo: Discussed offline, agree to decouple

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Discussed offline, agree to decouple

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoryInput.java
b4sjoo: Discussed offline, instead of removing we will throw exception

b4sjoo:
Use #4050 unless you have questions on add Memory API

PR #4032, opened 2025-08-01
[DO NOT MERGE] Get memory API

rithin-pullela-aws:
### Description
Adds Get Memory

This PR defines the MLMemory Object allowing storing: raw message, fact, or summary

The schema of the memory index propose here is different from the index created in `TransportCreateMemoryContainer.java`. 

Need to evaluate if we want to use this mapping.

### Related Issues

### Check List


common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -0,0 +1,245 @@
...
+@Getter
+@Setter
+@Builder
+@EqualsAndHashCode
+public class MLMemory implements ToXContentObject, Writeable {
+
+    private String rawMessage;
+    private String fact;
+    // type: raw message, fact, or summary
+    private String type;
```
dhrubo-os: I agree we should have a memory type. If we have a memory type do we need to add two fields like rawMessages or fact?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -0,0 +1,245 @@
...
+    private String summary;
+    private List<String> tags;
+    private String encoding;
```
dhrubo-os: you meant embedding?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -0,0 +1,245 @@
...
+    private String userId;
+    private String agentId;
+    private String sessionId;
+    private String tenantId;
```
dhrubo-os, line 58: I don't think we need tenantId for the memory. TenantId is required for the metadata. But at this point we are at the customer's index now.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportGetMemoryAction.java
```diff
@@ -0,0 +1,183 @@
...
+        ActionListener<MLMemoryGetResponse> actionListener
+    ) {
+        String memoryIndexName = mlMemoryContainer.getMemoryStorageConfig().getMemoryIndexName();
+
+        FetchSourceContext fetchSourceContext = new FetchSourceContext(true, Strings.EMPTY_ARRAY, Strings.EMPTY_ARRAY);
+        GetDataObjectRequest getMemoryRequest = GetDataObjectRequest
+            .builder()
+            .index(memoryIndexName)
+            .id(memoryId)
+            .tenantId(tenantId)
```
dhrubo-os: We need tenantId to get memory container, but we don't need tenantId to get memory from customer's index, as we are already inside tenant's index.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
dhrubo-os: We don't need tenantId here

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
rithinpullela: Will clean it up. It does not make sense here

dhrubo-os:
That's a great progress! Let's add unit tests too.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
rithinpullela: Agreed, having just "message" would work. 

I mainly wanted to point out that we need to edit the index creation part in create Memory Container.
I will refine this further.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportGetMemoryAction.java
rithinpullela: I see, will clean it up

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
rithinpullela: Will remove this

PR #4040, opened 2025-08-02
[FIX] allow partial updates to llm and memory fields in MLAgentUpdateInput

jiapingzeng:
### Description
Currently, agent update API allows partial updates to its fields. For example, 

```
PUT _plugins/_ml/agents/{{agent_id}}
{
  "name": "UPDATED agent"
}
```
This will only update the agent name field and leave other fields as-is. However, when updating sub-fields of `llm` and `memory`, `llm.model_id` and `memory.type` are required. 

This PR will make those fields also optional during an update operation (the result agent will keep values from before the update), so that it is consistent with other fields.

-----

Verified the below cases which are also included in UT:

1. Update `llm.parameters.max_iteration` only:

```
PUT {{endpoint}}/_plugins/_ml/agents/{{agent_id}}
{
  "llm": {
    "parameters": {
      "max_iteration": 50
    }
  }
}
```

2. Update `memory.session_id` only:
```
PUT {{endpoint}}/_plugins/_ml/agents/{{agent_id}}
{
  "memory": {
    "session_id": "asdf"
  }
}
```

3. Update other fields:
```
PUT {{endpoint}}/_plugins/_ml/agents/{{agent_id}}
{
  "name": "{{updated_agent_name}}",
  "type": "{{updated_agent_type}}",
  "description": "{{updated_agent_description}}",
  "llm": {
    "parameters": {
      "max_iteration": 10
    }
  },
  "tools": [
    {
      "type": "ListIndexTool",
      "include_output_in_agent_response": false
    },
    {
      "type": "SearchIndexTool",
      "include_output_in_agent_response": false
    }
  ],
  "parameters": {
    "_llm_interface": "bedrock/converse/claude"
  },
  "memory": {
    "type": "conversation_index"
  }
}
```

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
@@ -12,6 +12,7 @@ import static org.opensearch.ml.common.CommonValue.VERSION_2_19_0;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -41,9 +42,14 @@ public class MLAgentUpdateInput implements ToXContentObject, Writeable {
     public static final String AGENT_NAME_FIELD = "name";
     public static final String DESCRIPTION_FIELD = "description";
     public static final String LLM_FIELD = "llm";
+    public static final String LLM_MODEL_ID_FIELD = "model_id";
+    public static final String LLM_PARAMETERS_FIELD = "parameters";
     public static final String TOOLS_FIELD = "tools";
     public static final String PARAMETERS_FIELD = "parameters";
     public static final String MEMORY_FIELD = "memory";
+    public static final String MEMORY_TYPE_FIELD = "type";
+    public static final String MEMORY_SESSION_ID_FIELD = "session_id";
+    public static final String MEMORY_WINDOW_SIZE_FIELD = "window_size";
     public static final String APP_TYPE_FIELD = "app_type";
     public static final String LAST_UPDATED_TIME_FIELD = "last_updated_time";
 
@@ -51,10 +57,13 @@ public class MLAgentUpdateInput implements ToXContentObject, Writeable {
     private String agentId;
     private String name;
     private String description;
-    private LLMSpec llm;
+    private String llmModelId;
+    private Map<String, String> llmParameters;
     private List<MLToolSpec> tools;
     private Map<String, String> parameters;
-    private MLMemorySpec memory;
+    private String memoryType;
+    private String memorySessionId;
+    private Integer memoryWindowSize;
     private String appType;
     private Instant lastUpdateTime;
     private String tenantId;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/opensearch/ml/common/transport/agent/MLAgentUpdateInput.java
```diff
@@ -41,20 +42,28 @@ public class MLAgentUpdateInput implements ToXContentObject, Writeable {
     public static final String AGENT_NAME_FIELD = "name";
     public static final String DESCRIPTION_FIELD = "description";
     public static final String LLM_FIELD = "llm";
+    public static final String LLM_MODEL_ID_FIELD = "model_id";
+    public static final String LLM_PARAMETERS_FIELD = "parameters";
     public static final String TOOLS_FIELD = "tools";
     public static final String PARAMETERS_FIELD = "parameters";
     public static final String MEMORY_FIELD = "memory";
+    public static final String MEMORY_TYPE_FIELD = "type";
+    public static final String MEMORY_SESSION_ID_FIELD = "session_id";
```
ylwu-amzn, line 51: We don't store `session_id` in agent index. Do we need this in update agent API?

ylwu-amzn: I see why,  `MLMemorySpec` class has this field. It's ok to keep this. We can refactor together later.

ylwu-amzn:
Rebase this PR with the latest main branch ? The latest main branch has some fix can resolve the CI failures

dhrubo-os:
@jiapingzeng can you please push an empty PR. I take updates from branch, but I don't see the button to run the workflow.

jiapingzeng:
Hi @dhrubo-os, does the button show up now?

jiapingzeng:
looks like all tests passed other than codecov/project, I added more unit tests to improve coverage

mingshl:
Hi @jiapingzeng thanks for making the fix, but maybe it's a bit intensive to modify  all the related methods like parse(), toXContent(), writeTo(), etc.
To achieve the same partial update functionality while minimizing changes, we could:

Keep the original fields (LLMSpec and MLMemorySpec)

Only modify the streamInput method to make the fields optional:
```
public MLAgentUpdateInput(StreamInput in) throws IOException {
    agentId = in.readString();
    name = in.readOptionalString();
    description = in.readOptionalString();
    
    // Make LLM fields optional
    if (in.readBoolean()) {
        String modelId = in.readOptionalString();
        Map<String, String> params = in.readMap(StreamInput::readString, StreamInput::readOptionalString);
        llm = new LLMSpec(modelId, params);
    }
    
    // Make memory fields optional 
    if (in.readBoolean()) {
        String type = in.readOptionalString();
        String sessionId = in.readOptionalString(); 
        Integer windowSize = in.readOptionalInt();
        memory = new MLMemorySpec(type, sessionId, windowSize);
    }
    
    // Rest of the fields...
}```

This approach would be less invasive and  still achieve the goal of allowing partial updates

**Comment by jiapingzeng** [comment]
Hi @mingshl, the issue is that `LLMSpec` and `MLMemorySpec` require `llm.model` and `memory.type` respectively in their constructors. For example with this agent update request:

```
PUT {{endpoint}}/_plugins/_ml/agents/{{agent_id}}
{
  "llm": {
    "parameters": {
      "max_iteration": 50
    }
  }
}
```

There is no `llm.model` provided; in this case I want it to keep the model ID from the existing agent. However, methods other than `toMLAgent()` don't have access to the original agent, so there's no way to for them to instantiate an `LLMSpec` just from the update input.

I believe in your example `llm = new LLMSpec(modelId, params)` would throw `IllegalArgumentException` when `modelId` is null.

ylwu-amzn (approved):

PR #4096, opened 2025-08-12
Fix Default prompt of Query Planner Tool

rithin-pullela-aws:
### Description
The default prompt is updated to give more reliable responses.
With this change the instructions are sent in the user prompt instead of system prompt.

Steps to use the queryPlannerTool:

Step 1: Enable Agentic Search
```
PUT /_cluster/settings/
{
    "persistent": {
        "plugins.ml_commons.agentic_search_enabled": "true",
    }
}
```

Step 2: Register OpenAI Model
```
{
    "name": "My OpenAI model: gpt-4o",
    "function_name": "remote",
    "description": "test model",
    "connector": {
        "name": "My openai connector: gpt-40",
        "description": "The connector to openai chat model",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "model": "gpt-4o"
        },
        "credential": {
            "openAI_key": "<YOUR_OPENAI_KEY>"
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
                "url": "https://api.openai.com/v1/chat/completions",
                "headers": {
                },
                "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": [{\"role\":\"system\",\"content\":\"${parameters.system_prompt}\"},{\"role\":\"user\",\"content\":\"${parameters.user_prompt}\"}]}"
            }
        ]
    }
}
```

Step 3: Register a flow agent with queryPlanner tool 
```
{
  "name": "Agentic Search with Claude 3.7",
  "type": "flow",
  "tools": [
    {
      "type": "QueryPlanningTool",
      "description": "A general tool to answer any question",
      "parameters": {
        "model_id": "N_Wmn5gBKFS3OzyjJJ6u", // -> Generates query or chooses template
        "sys_p":"You are an expert in translating NLQ to OpenSearch DSL"
      }
    }
  ]
}
```

Step 4: Execute the Agent
```
{
    "parameters": {
        "query_text": "Recommend the best nike shoes",
        "query_fields": "brand, rating, category, description",
        "index_mapping": {
            "mappings": {
                "properties": {
                    "product_name": {
                        "type": "text"
                    },
                    "description": {
                        "type": "text"
                    },
                    "final_price": {
                        "type": "float"
                    },
                    "currency": {
                        "type": "keyword"
                    },
                    "in_stock": {
                        "type": "boolean"
                    },
                    "color": {
                        "type": "keyword"
                    },
                    "size": {
                        "type": "keyword"
                    },
                    "reviews_count": {
                        "type": "integer"
                    },
                    "category_tree": {
                        "type": "nested",
                        "properties": {
                            "name": {
                                "type": "keyword"
                            },
                            "url": {
                                "type": "keyword"
                            }
                        }
                    },
                    "rating": {
                        "type": "float"
                    },
                    "root_category": {
                        "type": "keyword"
                    },
                    "category": {
                        "type": "keyword"
                    },
                    "brand": {
                        "type": "keyword"
                    },
                    "all_available_sizes": {
                        "type": "keyword"
                    }
                }
            }
        },
        "response_filter": "$.choices[0].message.content"
    }
}
```

Expected Response:
```
{
    "inference_results": [
        {
            "output": [
                {
                    "name": "response",
                    "result": "{\"size\":10.0,\"query\":{\"bool\":{\"filter\":[{\"term\":{\"brand\":\"nike\"}},{\"term\":{\"category\":\"shoes\"}}]}},\"sort\":[{\"rating\":{\"order\":\"desc\"}},{\"reviews_count\":{\"order\":\"desc\"}}]}"
                }
            ]
        }
    ]
}
```

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
@@ -4,110 +4,138 @@ public class QueryPlanningPromptTemplate {
 
     public static final String DEFAULT_QUERY = "{\"size\":10,\"query\":{\"match_all\":{}}}";
 
-    public static final String QUERY_TYPE_RULES = "\nChoose query types based on user intent and fields: \n"
-        + "match: single-token full‑text searches on analyzed text fields, \n"
-        + "match_phrase: multi-token phrases on analyzed text fields (search string contains a space, hyphen, comma, etc.), \n"
-        + "term / terms:exact match on keyword, numeric, boolean, \n"
-        + "range:numeric/date comparisons (gt, lt, gte, lte), \n"
-        + "bool with must, should, must_not, filter: AND/OR/NOT logic, \n"
-        + "wildcard / prefix on keyword:\"starts with\", \"contains\", \n"
-        + "exists:field presence/absence, \n"
-        + "nested query / nested agg:Never wrap a field in nested unless the mapping for that exact path (or one of its parents) explicitly says \"type\": \"nested\". \n"
-        + "Otherwise use a normal query on the flattened field. \n";
+    // ==== RULES ====
+    public static final String QUERY_TYPE_RULES = "Use only fields present in the provided mapping; never invent names.\n"
+        + "Choose query types based on user intent and field types:\n"
+        + "- match: single-token full-text on analyzed text fields.\n"
+        + "- match_phrase: multi-token phrases on analyzed text fields (search string contains spaces, hyphens, commas, etc.).\n"
+        + "- multi_match: when multiple analyzed text fields are equally relevant.\n"
+        + "- term / terms: exact match on keyword, numeric, boolean.\n"
+        + "- range: numeric/date comparisons (gt, lt, gte, lte).\n"
+        + "- bool with must, should, must_not, filter: AND/OR/NOT logic.\n"
+        + "- wildcard / prefix on keyword: \"starts with\" / pattern matching.\n"
+        + "- exists: field presence/absence.\n"
+        + "- nested query / nested agg: ONLY if the mapping for that exact path (or a parent) has \"type\":\"nested\".\n"
+        + "\n"
+        + "Mechanics:\n"
+        + "- Put exact constraints (term, terms, range, exists, prefix, wildcard) in bool.filter (non-scoring). Put full-text relevance (match, match_phrase, multi_match) in bool.must.\n"
+        + "- Top N items/products/documents: return top hits (set \"size\": N as an integer) and sort by the relevant metric(s). Do not use aggregations for item lists.\n"
+        + "- Spelling tolerance: match_phrase does NOT support fuzziness; use match or multi_match with \"fuzziness\": \"AUTO\" when tolerant matching is needed.\n"
+        + "- Numeric note: use integers for sizes (e.g., \"size\": 5), not floats.\n";
 
-    public static final String AGGREGATION_RULES = "Aggregations (when asked for counts, averages, \"top N\", distributions): \n"
-        + "terms on field.keyword or numeric for grouping / top N, \n"
-        + "Metric aggs (avg, min, max, sum, stats, cardinality) on numeric fields, \n"
-        + "date_histogram, histogram, range for distributions, \n"
-        + "Always set \"size\": 0 when only aggregations are needed, \n"
-        + "Use sub‑aggregations + order for \"top N by metric\", \n"
-        + "If grouping by a text field, use its .keyword sub‑field.\n";
+    public static final String AGGREGATION_RULES = "Aggregations (counts, averages, grouped summaries, distributions):\n"
+        + "- Use aggregations when the user asks for grouped summaries (e.g., counts by category, averages by brand, or top N categories/brands).\n"
+        + "- terms on field.keyword or numeric for grouping / top N groups (not items).\n"
+        + "- Metric aggs (avg, min, max, sum, stats, cardinality) on numeric fields.\n"
+        + "- date_histogram, histogram, range for distributions.\n"
+        + "- Always set \"size\": 0 when only aggregations are needed.\n"
+        + "- Use sub-aggregations + order for \"top N groups by metric\".\n"
+        + "- If grouping/filtering exactly on a text field, use its .keyword sub-field when present.\n";
 
-    public static final String PROMPT_PREFIX =
-        "You are an OpenSearch DSL expert. Your job is to convert natural‑language questions into strict JSON OpenSearch search query bodies. \n"
-            + "Follow every rule: Use only the provided index mapping to decide which fields exist and their types, pay close attention to index mapping. \n"
-            + "Do not use fields that not present in mapping. \n"
-            + QUERY_TYPE_RULES
-            + AGGREGATION_RULES;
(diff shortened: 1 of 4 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -118,19 +146,22 @@ public class QueryPlanningPromptTemplate {
...
+        + "Query Fields: ${parameters.query_fields:-}\n\n"
+        + "==== OUTPUT ====\n"
+        + "GIVE THE OUTPUT PART ONLY IN YOUR RESPONSE (a single JSON object)\n"
         + "Output:";
 
-    public static final String DEFAULT_SYSTEM_PROMPT = PROMPT_PREFIX
-        + " \n "
+    public static final String DEFAULT_USER_PROMPT = PROMPT_PREFIX
+        + "\n\n"
         + OUTPUT_FORMAT_INSTRUCTIONS
```
owaiskazi19, line 162: Trying to understand, will it better to have `OUTPUT_FORMAT_INSTRUCTIONS` after `PROMPT_SUFFIX` or will it cause no change in the response?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -72,6 +73,9 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
         if (!parameters.containsKey(SYSTEM_PROMPT_FIELD)) {
             parameters.put(SYSTEM_PROMPT_FIELD, DEFAULT_SYSTEM_PROMPT);
         }
+        if (!parameters.containsKey(USER_PROMPT_FIELD)) {
+            parameters.put(USER_PROMPT_FIELD, DEFAULT_USER_PROMPT);
```
owaiskazi19, line 77: This default user prompt would be replaced if user provided a prompt, right?

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
```diff
@@ -85,14 +72,14 @@ public void deleteIndices() throws IOException {
...
-        if (AWS_ACCESS_KEY_ID == null) {
+        if (OPENAI_KEY == null) {
             return;
         }
         String agentName = "Test_QueryPlanningAgent_DefaultPrompt";
         String agentId = registerAgentWithQueryPlanningTool(agentName, queryPlanningModelId);
         assertNotNull(agentId);
 
-        String query = "{\"parameters\": {\"query_text\": \"How many iris flowers of type setosa are there?\"}}";
+        String query = "{\"parameters\": {\"query_text\": \"List 5 iris flowers of type setosa\"}}";
```
owaiskazi19, line 82: Why this change? List looks very specific

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
```diff
@@ -139,8 +126,8 @@ private String registerAgentWithQueryPlanningTool(String agentName, String model
     }
 
     private String registerQueryPlanningModel() throws IOException, InterruptedException {
-        String bedrockClaudeModelName = "bedrock claude model " + randomAlphaOfLength(5);
-        return registerRemoteModel(bedrockClaudeModelConnectorEntity, bedrockClaudeModelName, true);
+        String openaiModelName = "openai gpt-4o model " + randomAlphaOfLength(5);
```
owaiskazi19, line 129: Did gpt 4o perform better than claude in our testing? Also, did we try with gpt5?

owaiskazi19 (commented):
This is nice @rithin-pullela-aws. Few comments

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
rithinpullela: Prompt suffix has the user's question, index mapping and the query fields. So I think having it at the end makes more sense. Also the LLM can look at the output instructions and the examples right below.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: Yes, if the user's want to use their own prompt they can replace it via the `user_prompt` param.

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
rithinpullela: The IT was showing a flaky behavior because of expired AWS credentials in the CI. So replaced the model with open AI(uses API key) to unblock failing CI.

mingshl:
Overall, it's a good improvement to separate long prompt into: 

sytem prompt, 
instruction prompt,
and facts/examples.....

Two things to consider,
- query planning tool now only takes one user_prompt or system_prompt. If users want to update the prompt it's going to be a very long string, if we finalize the prompt structures and also get more industry standard on the prompt structures, we can open up more smaller parameters. For example, the user can just update the query_example, and leave the rest of the prompts, including system_prompt, instructions the same. That might be easier to update the prompt to better fit the users' indices, because they might have examples that fits better with their data. 

- how long is the entire prompt? how long can the questions and index_mappings can hold now? we need some more testings to see if the long prompt are limiting the abilities for the length of the questions, and the size of the index_mapping. If so, we should state in the documentation that currently, we suggest the question and index mapping can be less that a size limit.

plugin/src/test/java/org/opensearch/ml/rest/RestQueryPlanningToolIT.java
rithinpullela: With the new prompt the LLM understands to create a aggregation query so the term "query" is not present in the output. So I am asking it to list the products instead.

rithinpullela:
> Overall, it's a good improvement to separate long prompt into:
> 
> sytem prompt, instruction prompt, and facts/examples.....
> 
> Two things to consider,
> 
> * query planning tool now only takes one user_prompt or system_prompt. If users want to update the prompt it's going to be a very long string, if we finalize the prompt structures and also get more industry standard on the prompt structures, we can open up more smaller parameters. For example, the user can just update the query_example, and leave the rest of the prompts, including system_prompt, instructions the same. That might be easier to update the prompt to better fit the users' indices, because they might have examples that fits better with their data.
> * how long is the entire prompt? how long can the questions and index_mappings can hold now? we need some more testings to see if the long prompt are limiting the abilities for the length of the questions, and the size of the index_mapping. If so, we should state in the documentation that currently, we suggest the question and index mapping can be less that a size limit.



@mingshl to answer your questions,
1. I agree, ideally we should let the users update the small parts of the prompt. I created this issue #https://github.com/opensearch-project/ml-commons/issues/4098 to track it.
2. The prompt length should not be an issue for modern LLMs. But the limits are generally dependent on the kind of model being used.

rithinpullela:
The failing tests are not related to the changes made in this PR or to the query Planner Tool

jiapingzeng (approved):

PR #4111, opened 2025-08-13
Add ml-commons passthrough post process function

q-andy:
### Description
Adds a predefined post-process function to unwrap model output when making a remote call to second ml-commons cluster. When calling a remote second ml-commons predict API with a remote connector, the remote connector output will "double wrap" the second ml-commons. Tested compatibility with neural-search `SparseEncodingProcessor` and `TextEmbeddingProcessor`.

Predict call without post process function (double wrapped):
```
{
	"inference_results": [
		{
			"output": [
				{
					"name": "output",
					"dataAsMap": {
						"inference_results": [
							{
								"output": [
									{
										"name": "output",
										"dataAsMap": {
											"response": [
												{
													"this": 0.6228184700012207,
													"harrison": 0.5467907786369324,
													...
												},
												{
													"?": 0.1258760392665863,
													"day": 0.46702781319618225,
													...
												}
											]
										}
									}
								],
								"status_code": 200.0
							}
						]
					}
				}
			],
			"status_code": 200
		}
	]
}
```

After using post process function:
```
{
	"inference_results": [
		{
			"output": [
				{
					"name": "output",
					"dataAsMap": {
						"response": [
							{
								"increasingly": 0.028670792,
								"achievements": 0.4906937,
								...
							}
						]
					}
				},
				{
					"name": "output",
					"dataAsMap": {
						"response": [
							{
								"hesse": 0.07568395,
								"greeting": 0.296827,
								...
							}
						]
					}
				}
			],
			"status_code": 200
		}
	]
}
```

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
@@ -36,6 +36,7 @@ public class MLPostProcessFunction {
     public static final String BEDROCK_RERANK = "connector.post_process.bedrock.rerank";
     public static final String DEFAULT_EMBEDDING = "connector.post_process.default.embedding";
     public static final String DEFAULT_RERANK = "connector.post_process.default.rerank";
+    // ML commons passthrough unwraps a remote ml-commons response and reconstructs model tensors directly based on remote inference
     public static final String ML_COMMONS_PASSTHROUGH = "connector.post_process.mlcommons.passthrough";
 
     private static final Map<String, String> JSON_PATH_EXPRESSION = new HashMap<>();
--- a/common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
@@ -7,7 +7,9 @@ package org.opensearch.ml.common.connector.functions.postprocess;
 
 import static org.opensearch.ml.common.output.model.ModelTensors.OUTPUT_FIELD;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -26,11 +28,75 @@ public class RemoteMlCommonsPassthroughPostProcessFunction extends ConnectorPost
         }
     }
 
+    /**
+     * Example unwrapped response:
+     * {
+     * 	"inference_results": [
+    * 		        {
+     * 			"output": [
+     *                {
+     * 					"name": "output",
+     * 					"dataAsMap": {
+     * 						"inference_results": [
+     *                            {
+     * 								"output": [
+     *                                    {
+     * 										"name": "output",
+     * 										"dataAsMap": {
+     * 											"response": [
+     *                                                {
+     *                     								"increasingly": 0.028670792,
+     *                     								"achievements": 0.4906937,
+     *                     								...
+     *                                                }
+     * 											]
+     *                                        }
+     *                                    }
+     * 								],
+     * 								"status_code": 200.0
+     *                            }
+     * 						]
+     *                    }
+     *                }
+     * 			],
+     * 			"status_code": 200
+     *        }
+     * 	]
+     * }
+     *
+     * Example unwrapped response:
+     *
+     * {
+     * 	"inference_results": [
+     * 		        {
+     * 			"output": [
+     *                {
+     * 					"name": "output",
+     * 					"dataAsMap": {
+     * 						"response": [
+     *                            {
+     * 								"increasingly": 0.028670792,
+     * 								"achievements": 0.4906937,
+     * 								...
+     *                            }
+     * 						]
+     *                    }
+     *                },
+     * 			],
+     * 			"status_code": 200
+     *        }
+     * 	]
+     * }
+     *
+     * @param mlCommonsResponse raw remote ml commons response
+     * @param dataType the datatype of the result, not used since datatype is set based on the response body
+     * @return a list of model tensors representing the inner model tensors
+     */
     @Override
-    public List<ModelTensor> process(Map<String, Object> input, MLResultDataType dataType) {
+    public List<ModelTensor> process(Map<String, Object> mlCommonsResponse, MLResultDataType dataType) {
         // Check if this is an ML-Commons response with inference_results
-        if (input.containsKey("inference_results") && input.get("inference_results") instanceof List) {
-            List<Map<String, Object>> inferenceResults = (List<Map<String, Object>>) input.get("inference_results");
+        if (mlCommonsResponse.containsKey("inference_results") && mlCommonsResponse.get("inference_results") instanceof List) {
+            List<Map<String, Object>> inferenceResults = (List<Map<String, Object>>) mlCommonsResponse.get("inference_results");
 
             List<ModelTensor> modelTensors = new ArrayList<>();
             for (Map<String, Object> result : inferenceResults) {
@@ -51,7 +117,7 @@ public class RemoteMlCommonsPassthroughPostProcessFunction extends ConnectorPost
         }
 
         // Fallback for non-ML-Commons responses
-        ModelTensor tensor = ModelTensor.builder().name("response").dataAsMap(input).build();
+        ModelTensor tensor = ModelTensor.builder().name("response").dataAsMap(mlCommonsResponse).build();
 
         return List.of(tensor);
     }
@@ -69,9 +135,12 @@ public class RemoteMlCommonsPassthroughPostProcessFunction extends ConnectorPost
         Object uncastedName = map.get(ModelTensor.NAME_FIELD);
         String name = uncastedName instanceof String castedName ? castedName : OUTPUT_FIELD;
         String result = (String) map.get(ModelTensor.RESULT_FIELD);
+
+        // Handle data as map
         Map<String, Object> dataAsMap = (Map<String, Object>) map.get(ModelTensor.DATA_AS_MAP_FIELD);
 
-        // Handle data type
+        // Handle data type. For certain models like neural sparse and non-dense remote models, this field
+        // is not populated and left as null instead, which is still valid
         MLResultDataType dataType = null;
         if (map.containsKey(ModelTensor.DATA_TYPE_FIELD)) {
             Object dataTypeObj = map.get(ModelTensor.DATA_TYPE_FIELD);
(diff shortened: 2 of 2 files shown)


common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return List.of(tensor);
+    }
+
+    /**
+     * Creates a ModelTensor from a Map<String, Object> representation based on the API format
+     * of the /_predict API
+     */
+    @SuppressWarnings("unchecked")
+    private ModelTensor createModelTensorFromMap(Map<String, Object> map) {
```
Zhangxunmt, line 129: should add validation for the map to be not null and some fields exist.

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,117 @@
...
+        String name = (String) map.getOrDefault(ModelTensor.NAME_FIELD, OUTPUT_FIELD);
+        String result = (String) map.get(ModelTensor.RESULT_FIELD);
+        Map<String, Object> dataAsMap = (Map<String, Object>) map.get(ModelTensor.DATA_AS_MAP_FIELD);
+
+        // Handle data type
+        MLResultDataType dataType = null;
+        if (map.containsKey(ModelTensor.DATA_TYPE_FIELD)) {
+            Object dataTypeObj = map.get(ModelTensor.DATA_TYPE_FIELD);
+            if (dataTypeObj instanceof String) {
```
Zhangxunmt, line 147: nitpick: There are multiple unchecked type casting operations in this method. I think it worth adding type safety to avoid too much "unchecked" warnings.

q-andy: Removed the @ SuppressUnchecked warnings, and fixed most of the casts. The remaining casts are like `(List<Map<?, ?>) List<?>` which are difficult to avoid due to how Java does type erasure for generics.

For this case I think its ok, we are casting because we know the input will be in a specific format (we have knowledge that the Java language doesn't). And if the input isn't in the correct format, then we want it to throw an exception.

dhrubo-os:
@q-andy fix spotless :)

Zhangxunmt (approved):

Zhangxunmt (approved):

q-andy:
Most CI passed, just 2 flaky integ test failures are unrelated to this change. Could you take another look @dhrubo-os @Zhangxunmt

Zhangxunmt (approved):

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+public class RemoteMlCommonsPassthroughPostProcessFunction extends ConnectorPostProcessFunction<Map<String, Object>> {
+    @Override
+    public void validate(Object input) {
+        if (!(input instanceof Map) && !(input instanceof List)) {
+            throw new IllegalArgumentException("Post process function input must be a Map or List");
+        }
+    }
+
+    @Override
+    public List<ModelTensor> process(Map<String, Object> input, MLResultDataType dataType) {
```
dhrubo-os: instead of `Map<String, Object> input`, Since this parameter represents ML Commons inference results or response data, some better name suggestions would be:

    mlCommonsResponse
    inferenceResponse
    responseData
    modelResponse

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.ml.common.output.model.MLResultDataType;
+import org.opensearch.ml.common.output.model.ModelTensor;
+
+/**
+ * A post-processing function for calling a remote ml commons instance that preserves the original neural sparse response structure
+ * to avoid double-wrapping when receiving responses from another ML-Commons instance.
+ */
```
dhrubo-os, line 22: Can we add example in the doc for better understanding?

common/src/main/java/org/opensearch/ml/common/connector/MLPostProcessFunction.java
```diff
@@ -35,6 +36,7 @@ public class MLPostProcessFunction {
     public static final String BEDROCK_RERANK = "connector.post_process.bedrock.rerank";
     public static final String DEFAULT_EMBEDDING = "connector.post_process.default.embedding";
     public static final String DEFAULT_RERANK = "connector.post_process.default.rerank";
+    public static final String ML_COMMONS_PASSTHROUGH = "connector.post_process.mlcommons.passthrough";
```
dhrubo-os, line 40: as this is a special kind, can we please add a comment here?

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+        Map<String, Object> dataAsMap = (Map<String, Object>) map.get(ModelTensor.DATA_AS_MAP_FIELD);
+
+        // Handle data type
+        MLResultDataType dataType = null;
+        if (map.containsKey(ModelTensor.DATA_TYPE_FIELD)) {
+            Object dataTypeObj = map.get(ModelTensor.DATA_TYPE_FIELD);
+            if (dataTypeObj instanceof String) {
+                try {
+                    dataType = MLResultDataType.valueOf((String) dataTypeObj);
+                } catch (IllegalArgumentException e) {
```
dhrubo-os, line 150: `IllegalArgumentException` is a 500 level error. Should we treat this as a 5xx error?

dhrubo-os: Also what is the reason for leaving as null?

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+                    // Invalid data type, leave as null
+                }
+            }
+        }
+
+        // Handle shape
+        long[] shape = null;
+        if (map.containsKey(ModelTensor.SHAPE_FIELD)) {
+            Object shapeObj = map.get(ModelTensor.SHAPE_FIELD);
+            if (shapeObj instanceof List<?> shapeList) {
```
dhrubo-os: In the else we are sending `null` for shape, is that expected?

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+        }
+
+        // Handle data array
+        Number[] data = null;
+        if (map.containsKey(ModelTensor.DATA_FIELD)) {
+            Object dataObj = map.get(ModelTensor.DATA_FIELD);
+            if (dataObj instanceof List<?> dataList) {
```
dhrubo-os: what happens if we send data as null to the model Tensor?

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
```diff
@@ -0,0 +1,121 @@
...
+                    if (item instanceof Number) {
+                        shape[i] = ((Number) item).longValue();
+                    }
+                }
+            }
+        }
+
+        // Handle data array
+        Number[] data = null;
+        if (map.containsKey(ModelTensor.DATA_FIELD)) {
```
dhrubo-os, line 169: Seems like both of the underlying logic could be put in a common method like:

```
private <T> T[] processNumericArray(Object obj, Class<T> type) {
    if (obj instanceof List<?> list) {
        T[] result = (T[]) Array.newInstance(type, list.size());
        // ... process the list
        return result;
    }
    return null;
}
```

mingshl:
after using the post processing function, 

the name seems changed, the name used to be "response" but after the post processing become "output". Is this change intended? I am sure if the name is used somewhere but better not change the name if not intentionally. 

```
{
	"inference_results": [
		{
			"output": [
				{
					"name": "output",
					"dataAsMap": {
						"response": [
```

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
q-andy: Sure, didn't realize you could rename the parameter variables of an overridden function. Changed to `mlCommonsResponse`

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
q-andy: I left as null instead of treating like 5xx is because even if the model data type is invalid, there may be usecases where we can still parse and use the model data or dataAsMap. E.g. some model types like neural sparse or NER don't include data type as part of the model response, so null is still valid.

Right now since we're focused on neural sparse and dense, my thought process is its better to leave it flexible to be able to possible handle different model response formats. For example, in the future we me add a new datatype for dense models and we might inference that from an older version of ml-commons: perhaps we can still use the data by casting the data at the processor level. Updated the comment to explain this.

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
q-andy: Yes, e.g. neural sparse and dense models populate different fields, sparse leaves data_type, data, and shape as null.

Dense
```
{
	"name": "sentence_embedding",
	"data_type": "FLOAT32",
	"shape": [
		768
	],
	"data": [...]
}
```

Sparse
```
{
	"name": "output",
	"dataAsMap": {
		"response": [
			{ ... }
		]
	}
}
```

common/src/main/java/org/opensearch/ml/common/connector/functions/postprocess/RemoteMlCommonsPassthroughPostProcessFunction.java
q-andy: See above comment, this is valid for neural sparse and other remote model format. Looks like data field is primarily used for vector/numerical info, and if a model output doesn't include that, then it uses dataAsMap instead and data field being null is valid.

q-andy:
>the name seems changed, the name used to be "response" but after the post processing become "output". Is this change intended? I am sure if the name is used somewhere but better not change the name if not intentionally.

Fixed, the name is relevant for different model types. I changed it so the name will be passthroughed as well, this is just a typo in the PR description.

dhrubo-os (approved):

Zhangxunmt (approved):

q-andy:
Could we backport to 2.19 and 3.1? For 2.19 back port there will be some conflicts, I will resolve

cc: @dhrubo-os @Zhangxunmt

PR #4121, opened 2025-08-15
Parameter Passing for Predict via Remote Connector

### LINKED ISSUE #4105: [FEATURE] Parameter Passing for Predict via Remote Connector
**Background**

In ml-commons, remote models are accessed via a connector, which defines how to construct and send inference requests. This includes specifying endpoints, authentication, and the structure of the request payload—known as the request_body template. 

Currently, when invoking the predict API with text_embedding, the system does not fully support passing user-defined parameters from the predict call through the connector to the remote endpoint. For example, sending a request such as:

```
POST {{base_url}}/_plugins/_ml/_predict/text_embedding/{{model_id}}
{
    "text_docs": ["hello world"],
    "parameters": {
        "sparse_embedding_format": "TOKEN_ID"
    }
}
```

produces the same output as omitting the parameters field, meaning the value of sparse_embedding_format does not affect the result. This limitation stems from the current invocation method and parameter handling associated with the text_embedding function, which does not fully propagate user-supplied parameters into the final HTTP payload sent for inference.

This design aims to enhance the parameter passing mechanism, ensuring that user-provided parameters during predict calls are correctly integrated with the connector’s request template and properly formatted in the outgoing request to the remote endpoint.

**Implementation**

The connector’s request_body is designed as a JSON string template with placeholders. The connector’s request_body template includes some parameters hardcoded as fixed values (Static Parameters) and others as placeholders to be replaced by predict call parameters (Dynamic Parameters). When creating the connector, users can optionally define default values for dynamic parameters in the "parameters"field. During a predict call, if a dynamic parameter is not provided, the system will first check whether a default value exists. If a default value is found, it will be used; if no default is defined, the parameter will be removed from the final request payload. Static parameters’ values remain fixed and cannot be overridden.

**Example format:**

```
{
  "input": "${parameters.input}",
  "parameters": {
    "sparseEmbeddingFormat": "${parameters.sparseEmbeddingFormat}",
    "embeddingContentType": null
  }
}
```

**Example Usage Overview: SageMaker Model Deployment, Connector Creation, and API Invocation**

In the following section, we will outline how to use this feature within the existing workflow of deploying a model on SageMaker, configuring a connector with a specified request payload format for remote inference, and invoking the predict API. This overview focuses specifically on the changes and additions related to the parameter passing feature in the deployment and connector setup code, rather than providing a full end-to-end tutorial. Detailed code design and configuration snippets will be provided to clearly demonstrate how parameters flow and are managed between the client, connector, and remote endpoint in the context of this feature.

**Deploy sparse model to SageMaker**

Prepare the model file for SageMaker.
```
%%writefile handler/neural_sparse_handler.py

import os
import re
import itertools
import json
import torch

from ts.torch_handler.base_handler import BaseHandler
from sentence_transformers.sparse_encoder import SparseEncoder

model_id = os.environ.get(
    "MODEL_ID", "opensearch-project/opensearch-neural-sparse-encoding-doc-v3-gte"
)
max_bs = int(os.environ.get("MAX_BS", 32))
trust_remote_code = model_id.endswith("gte")

class SparseEncodingModelHandler(BaseHandler):
    def __init__(self):
        super().__init__()
        self.initialized = False

    def initialize(self, context):
        self.manifest = context.manifest
        properties = context.system_properties

        # Print initialization parameters
        print(f"Initializing SparseEncodingModelHandler with model_id: {model_id}")

        # load model and tokenizer
        self.device = torch.device(
            "cuda:" + str(properties.get("gpu_id"))
            if torch.cuda.is_available()
            else "cpu"
        )
        print(f"Using device: {self.device}")
        self.model = SparseEncoder(model_id, device=self.device, trust_remote_code=trust_remote_code)
        self.initialized = True

    def preprocess(self, requests):
        inputSentence = []
        batch_idx = []
        parameters = {}
    
        for request in requests:
            request_body = request.get("body")
            if isinstance(request_body, bytearray):
                request_body = request_body.decode("utf-8")
                request_json = json.loads(request_body)
            
            if isinstance(request_json, dict) and "input" in request_json:
                inputs = request_json["input"]
                parameters = request_json.get("parameters", {})
    
                if isinstance(inputs, list):
                    inputSentence += inputs
                    batch_idx.append(len(inputs))
                else:
                    inputSentence.append(inputs)
                    batch_idx.append(1)
            else:
                if isinstance(request_json, list):
                    inputSentence += request_json
                    batch_idx.append(len(request_json))
                else:
                    inputSentence.append(request_json)
                    batch_idx.append(1)
    
        return inputSentence, batch_idx, parameters


    def handle(self, data, context):
        inputSentence, batch_idx, parameters = self.preprocess(data)
        model_output = self.model.encode_document(inputSentence, batch_size=max_bs)
        sparse_embedding = list(map(dict,self.model.decode(model_output, parameters)))

        outputs = [sparse_embedding[s:e]
           for s, e in zip([0]+list(itertools.accumulate(batch_idx))[:-1],
                           itertools.accumulate(batch_idx))]
        return outputs
```

**Create connector**

```
{
  "name": "test",
  "description": "Test connector for Sagemaker model",
  "version": 1,
  "protocol": "aws_sigv4",
  "credential": {
    "access_key": "your access key",
    "secret_key": "your secret key"
  },
  "parameters": {
    "region": "{region}",
    "service_name": "sagemaker",
    "input_docs_processed_step_size": 2,
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "headers": {
      },
      "url": "https://runtime.sagemaker.{region}.amazonaws.com/endpoints/{predictor.endpoint_name}/invocations",
      "request_body": "{ \"input\": ${parameters.input}, \"parameters\": {\"sparse_embedding_format\": \"${parameters.sparseEmbeddingFormat}\" }}"
  }],
  "client_config":{
      "max_retry_times": -1,
      "max_connection": 60,
      "retry_backoff_millis": 10
  }
}
```

**Register model**

```
POST /_plugins/_ml/models/_register?deploy=true
{
"name": "test",
  "function_name": "remote",
  "version": "1.0.0",
  "connector_id": "{connector id}",
  "description": "Test connector for Sagemaker model"
}
```
**Invoking Text Embedding Prediction API**
```
POST {{base_url}}/_plugins/_ml/_predict/text_embedding/{{model_id}} 
{
    "text_docs":["hello world"],
    "parameters":{
        "sparse_embedding_format": "TOKEN_ID"
    }
}
```

Sq-eng:
### Description
This PR adds support for parameter propagation in the predict API when using `text_embedding` function by ensuring user-supplied parameters are correctly merged into the connector's request_body and included in the outgoing request.

### Related Issues
Resolves #4105

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -359,7 +359,9 @@ public class HttpConnector extends AbstractConnector {
     }
 
     public String removeMissingParameterFields(String payload, Map<String, String> params) {
-        // Match: "xxx": "${parameters.yyy}" or "xxx": {parameters.yyy}
+        if (params == null) {
+            return payload;
+        }
         Pattern pattern = Pattern.compile(
                 "\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?"
         );
@@ -368,7 +370,7 @@ public class HttpConnector extends AbstractConnector {
 
         while (matcher.find()) {
             String paramName = matcher.group(2); // yyy
-            if (!params.containsKey(paramName)) {
+            if (!params.containsKey(paramName) && !"input".equals(paramName)) {
                 matcher.appendReplacement(sb, "");
             } else {
                 matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(0)));
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -358,13 +358,15 @@ public class HttpConnector extends AbstractConnector {
         return (T) parameters.get("http_body");
     }
 
+    /**
+     * Removes fields from the given JSON payload string that correspond to parameters
+     * not present in the provided parameter map.
+     */
     public String removeMissingParameterFields(String payload, Map<String, String> params) {
         if (params == null) {
             return payload;
         }
-        Pattern pattern = Pattern.compile(
-                "\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?"
-        );
+        Pattern pattern = Pattern.compile("\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?");
         Matcher matcher = pattern.matcher(payload);
         StringBuffer sb = new StringBuffer();
 
@@ -379,7 +381,6 @@ public class HttpConnector extends AbstractConnector {
         matcher.appendTail(sb);
         return sb.toString().replaceAll(",\\s*}", "}").replaceAll(",\\s*]", "]");
     }
-
 
     protected String fillNullParameters(Map<String, String> parameters, String payload) {
         List<String> bodyParams = findStringParametersWithNullDefaultValue(payload);
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -346,6 +346,7 @@ public class HttpConnector extends AbstractConnector {
             String payload = connectorAction.get().getRequestBody();
             payload = fillNullParameters(parameters, payload);
             parseParameters(parameters);
+            payload = removeMissingParameterFields(payload, parameters);
             StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
             payload = substitutor.replace(payload);
 
@@ -356,6 +357,27 @@ public class HttpConnector extends AbstractConnector {
         }
         return (T) parameters.get("http_body");
     }
+
+    private String removeMissingParameterFields(String payload, Map<String, String> params) {
+        // Match: "xxx": "${parameters.yyy}" or "xxx": {parameters.yyy}
+        Pattern pattern = Pattern.compile(
+                "\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?"
+        );
+        Matcher matcher = pattern.matcher(payload);
+        StringBuffer sb = new StringBuffer();
+
+        while (matcher.find()) {
+            String paramName = matcher.group(2); // yyy
+            if (!params.containsKey(paramName)) {
+                matcher.appendReplacement(sb, "");
+            } else {
+                matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(0)));
+            }
+        }
+        matcher.appendTail(sb);
+        return sb.toString().replaceAll(",\\s*}", "}").replaceAll(",\\s*]", "]");
+    }
+
 
     protected String fillNullParameters(Map<String, String> parameters, String payload) {
         List<String> bodyParams = findStringParametersWithNullDefaultValue(payload);
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -346,7 +346,6 @@ public class HttpConnector extends AbstractConnector {
             String payload = connectorAction.get().getRequestBody();
             payload = fillNullParameters(parameters, payload);
             parseParameters(parameters);
-            payload = removeMissingParameterFields(payload, parameters);
             StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
             payload = substitutor.replace(payload);
 
@@ -356,30 +355,6 @@ public class HttpConnector extends AbstractConnector {
             return (T) payload;
         }
         return (T) parameters.get("http_body");
-    }
-
-    /**
-     * Removes fields from the given JSON payload string that correspond to parameters
-     * not present in the provided parameter map.
-     */
-    public String removeMissingParameterFields(String payload, Map<String, String> params) {
-        if (params == null) {
-            return payload;
-        }
-        Pattern pattern = Pattern.compile("\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?");
-        Matcher matcher = pattern.matcher(payload);
-        StringBuffer sb = new StringBuffer();
-
-        while (matcher.find()) {
-            String paramName = matcher.group(2); // yyy
-            if (!params.containsKey(paramName) && !"input".equals(paramName)) {
-                matcher.appendReplacement(sb, "");
-            } else {
-                matcher.appendReplacement(sb, Matcher.quoteReplacement(matcher.group(0)));
-            }
-        }
-        matcher.appendTail(sb);
-        return sb.toString().replaceAll(",\\s*}", "}").replaceAll(",\\s*]", "]");
     }
 
     protected String fillNullParameters(Map<String, String> parameters, String payload) {
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -358,7 +358,7 @@ public class HttpConnector extends AbstractConnector {
         return (T) parameters.get("http_body");
     }
 
-    private String removeMissingParameterFields(String payload, Map<String, String> params) {
+    public String removeMissingParameterFields(String payload, Map<String, String> params) {
         // Match: "xxx": "${parameters.yyy}" or "xxx": {parameters.yyy}
         Pattern pattern = Pattern.compile(
                 "\\s*\"[^\"]+\"\\s*:\\s*(\"?\\$?\\{parameters\\.([^}]+)\\}\"?)\\s*,?"
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
@@ -5,23 +5,20 @@
 
 package org.opensearch.ml.engine.algorithms.remote;
 
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+//import static org.opensearch.ml.common.dataset.SearchQueryInputDataset.xContentRegistry;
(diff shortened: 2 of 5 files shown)


common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -357,6 +358,26 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
         return (T) parameters.get("http_body");
     }
 
+    public String removeMissingParameterFields(String payload, Map<String, String> params) {
```
xinyual: Can you add description for this function?

Sq-eng: Done.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -9,13 +9,8 @@
...
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
+import java.io.IOException;
+import java.util.*;
```
zhichao-aws: we should aviod using import *, and restore to the previous import

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -28,11 +23,12 @@
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.TokenBucket;
+import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.commons.ConfigConstants;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
-import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.*;
```
zhichao-aws: same as above

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
zhichao-aws: I think we don't need remove the unused field. We should keep consistent behavior with previous code. I.e. if user set a parameter in template and don't provide values, we just keep it. Otherwise it may change the behavior at user side

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
Sq-eng: Done.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
Sq-eng: Done.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -227,6 +246,23 @@ && getUserRateLimiterMap().get(user.getName()) != null
         }
     }
 
+    default Map<String, String> getParams(MLInput mlInput) throws IOException {
```
zhichao-aws: Can we make it a static method? it doesn't have dependency on other fields or method

Sq-eng: The method has been updated to `static`.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
Sq-eng: I understand. I’ve updated the code to follow this logic – now it keeps the unused fields consistent with the previous behavior.

xinyual:
LGTM.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -227,6 +246,23 @@ && getUserRateLimiterMap().get(user.getName()) != null
         }
     }
 
+    static Map<String, String> getParams(MLInput mlInput) throws IOException {
```
zane-neo, line 250: you can reuse this method: `org.opensearch.ml.common.utils.StringUtils#getParameterMap`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -187,6 +195,17 @@ default void preparePayloadAndInvoke(
...
         }
         parameters.putAll(inputParameters);
+
+        MLAlgoParams algoParams = mlInput.getParameters();
+        if (algoParams != null) {
+            try {
+                Map<String, String> parametersMap = getParams(mlInput);
+                parameters.putAll(parametersMap);
+            } catch (IOException e) {
+                actionListener.onFailure(e);
```
zane-neo, line 205: Add return here.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
Sq-eng: Thanks for the suggestion. However, most of the logic here is unrelated to `StringUtils#getParameterMap`—it’s mainly handling the extraction and preprocessing of parameters from `mlInput`. Only the last two lines are somewhat similar to `getParameterMap`, but the implementation details differ slightly. I believe reusing `StringUtils#getParameterMap` here isn’t necessary, and keeping the current logic is clearer.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
Sq-eng: Done.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -227,6 +247,23 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
     }
 
+    static Map<String, String> getParams(MLInput mlInput) throws IOException {
+        Map<String, String> result = new HashMap<>();
+        XContentBuilder builder = XContentFactory.jsonBuilder();
+        mlInput.getParameters().toXContent(builder, ToXContent.EMPTY_PARAMS);
+        builder.flush();
+        String json = builder.toString();
+
+        ObjectMapper mapper = new ObjectMapper();
```
zane-neo: Let's make a performance optimization here: 
1. Change the `MAPPER` here: `org.opensearch.ml.common.utils.StringUtils#MAPPER` to public.
2. Use the public MAPPER here for deserialization.

Sq-eng: Done.

zane-neo (dismissed):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -227,6 +246,21 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+        XContentBuilder builder = XContentFactory.jsonBuilder();
+        mlInput.getParameters().toXContent(builder, ToXContent.EMPTY_PARAMS);
+        builder.flush();
+        String json = builder.toString();
+
+        Map<String, Object> tempMap = StringUtils.MAPPER.readValue(json, Map.class);
+
+        HashMap<String, String> paramMap = new HashMap<>();
+        for (Map.Entry<String, Object> entry : tempMap.entrySet()) {
+            paramMap.put(entry.getKey(), entry.getValue() != null ? entry.getValue().toString() : null);
```
zane-neo: If the Object is a map/list type, the toString by default returns their address in memory instead of a json, it seems not correct here, do you have test cases to cover this case?

zane-neo:
> Patch coverage is 60.00000% with 8 lines in your changes missing coverage.

The code coverage in this PR is little bit low, can you add more UTs to increase the coverage?

Sq-eng:
> > Patch coverage is 60.00000% with 8 lines in your changes missing coverage.
> 
> The code coverage in this PR is little bit low, can you add more UTs to increase the coverage?

I've already added UT for it. But the Codecov Report hasn't been updated yet. Could you please allow CI/CD so it can rerun the test?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
Sq-eng: I’ve addressed this issue by updating the code so that all types except basic types and String are serialized to JSON strings. Also added a unit test.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -227,6 +248,33 @@ && getUserRateLimiterMap().get(user.getName()) != null
...
+                paramMap.put(entry.getKey(), null);
+            } else if (value instanceof String) {
+                paramMap.put(entry.getKey(), value.toString());
+            } else {
+                try {
+                    String jsonValue = StringUtils.MAPPER.writeValueAsString(value);
+                    paramMap.put(entry.getKey(), jsonValue);
+                } catch (JsonProcessingException e) {
+                    paramMap.put(entry.getKey(), null);
+                }
```
zane-neo: This part can be replaced with the existing util method: `getParameterMap` in `org.opensearch.ml.common.utils.StringUtils`, right? A good practice in coding is reusing and avoid duplication.

Sq-eng: Yes, you are right. I’ve updated the code to reuse `getParameterMap `from `StringUtils` to avoid duplication. Thanks for the suggestion!

zane-neo (approved):

xinyual (approved):

PR #4131, opened 2025-08-22
Add provided dependency section to ml-client POM

### LINKED ISSUE #3860: [BUG]
**What is the bug?**

In https://github.com/opensearch-project/ml-commons/pull/3805, the `validate()` method of the `MLCreateConnectorRequest` was used to add field validation using `StringUtils`.  

Unfortunately, [`validate()` is called from the `TransportActionNodeProxy` class](https://github.com/opensearch-project/OpenSearch/blob/54af44541698069242e865fd2ef3420aa0868c19/server/src/main/java/org/opensearch/action/TransportActionNodeProxy.java#L59-L72) with an `org.opensearch.action.Request` class, and the `StringUtils` class is not available to the classloader used for this call.

1. Attempt to create a connector using the ML Client's `createConnector()` method from another plugin (for example, any Flow Framework plugin integ test that creates a connector).
2. Observe the request failing silently with an exception (probably class loading?) swallowed silently somewhere.

A connector is created or field validation fails with a 400 error.

Easy to reproduce locally on macOS running Flow Framework integ tests on current `main` branch beginning with the commit including https://github.com/opensearch-project/ml-commons/pull/3805.

Also has caused integ test failures on all platforms on GitHub Action integ tests as well as Jenkins test environment, see https://github.com/opensearch-project/flow-framework/issues/1143

I've worked around the problem locally by separating the field validation code from the `validate()` method.

```java
    @Override
    public ActionRequestValidationException validate() {
        // This is called with a Request class and a different classloader so must be simple
        if (mlCreateConnectorInput == null) {
            return addValidationError("ML Connector input can't be null", null);
        }
        return null;
    }

    // This must be called from a thread with access to this class's packages
    public static ActionRequestValidationException validateFields(MLCreateConnectorInput mlCreateConnectorInput) {
        Map<String, FieldDescriptor> fieldsToValidate = new HashMap<>();
        fieldsToValidate.put("Model connector name", new FieldDescriptor(mlCreateConnectorInput.getName(), true));
        fieldsToValidate.put("Model connector description", new FieldDescriptor(mlCreateConnectorInput.getDescription(), false));

        return StringUtils.validateFields(fieldsToValidate);
    }
```

And then, from `TransportCreateConnectorAction`:
```java
    @Override
    protected void doExecute(Task task, ActionRequest request, ActionListener<MLCreateConnectorResponse> listener) {
        MLCreateConnectorRequest mlCreateConnectorRequest = MLCreateConnectorRequest.fromActionRequest(request);
        MLCreateConnectorInput mlCreateConnectorInput = mlCreateConnectorRequest.getMlCreateConnectorInput();
        // existing protocol, tenant, and dryrun conditionals
        String connectorName = mlCreateConnectorInput.getName();
        try {
            // add this line
            MLCreateConnectorRequest.validateFields(mlCreateConnectorInput);
            // remainder of the class
```

Updating tests is left as an exercise for the reader.

dbwiddis:
### Description

The ML Client requires a few runtime dependencies that are not included in the shadow JAR, primarily associated with validation of ML Connector fields.  The requirement to have these dependencies by downstream consumers of the client is not documented nor obvious.

In the Maven dependency ecosystem, the [**`provided`** dependency scope](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope) is a clear indicator of a required runtime dependency that the project is not providing itself:
> **provided**
> This is much like compile, but indicates you expect the JDK or a container to provide the dependency at runtime. For example, when building a web application for the Java Enterprise Edition, you would set the dependency on the Servlet API and related Java EE APIs to scope provided because the web container provides those classes. A dependency with this scope is added to the classpath used for compilation and test, but not the runtime classpath. It is not transitive.

This PR appends a `<dependency>` section to the [published opensearch-ml-client POM file on the Central Repository](https://central.sonatype.com/artifact/org.opensearch/opensearch-ml-client). The additional section added will be:
```xml
<dependencies>
  <dependency>
    <groupId>org.json</groupId>
    <artifactId>json</artifactId>
    <version>20231013</version>
    <scope>provided</scope>
  </dependency>
  <dependency>
    <groupId>com.jayway.jsonpath</groupId>
    <artifactId>json-path</artifactId>
    <version>2.9.0</version>
    <scope>provided</scope>
  </dependency>
  <dependency>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-text</artifactId>
    <version>1.10.0</version>
    <scope>provided</scope>
  </dependency>
</dependencies>
```

Note that there is an additional dependency in `StringUtils` on `com.networknt:json-schema-validator:1.4.0`.  Currently this is not required by the ML Client, but if any future changes to the client include schema validation, that should be added to this list. 

_**This has no impact on actual included dependencies; it's simply a documentation signal to clearly indicate to consumers of this client that they must provide these dependencies themselves.**_

### Related Issues
 - Resolves #3860
 - Supercedes #3861 which can be closed (thanks @jngz-es for trying to fix!)

### Check List

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to specify external library dependencies that must be provided by the runtime environment rather than bundled with the application.

### DIFF
--- a/client/build.gradle
+++ b/client/build.gradle
@@ -108,6 +108,24 @@ publishing {
                         url = "https://github.com/opensearch-project/ml-commons"
                     }
                 }
+                // Add a dependencies section to signal that consumers of the client need to provide these
+                withXml {
+                    def dependencies = asNode().appendNode('dependencies')
+                    
+                    def addProvidedDependency = { group, artifact, version ->
+                        def dependency = dependencies.appendNode('dependency')
+                        dependency.appendNode('groupId', group)
+                        dependency.appendNode('artifactId', artifact)
+                        dependency.appendNode('version', version)
+                        dependency.appendNode('scope', 'provided')
+                    }
+
+                    addProvidedDependency('org.json', 'json', '20231013')
+                    addProvidedDependency('com.jayway.jsonpath', 'json-path', '2.9.0')
+                    addProvidedDependency('org.apache.commons', 'commons-text', '1.14.0')
+                    addProvidedDependency('com.fasterxml.jackson.core', 'jackson-annotations', "${versions.jackson_annotations}")
+                    addProvidedDependency('tools.jackson.core', 'jackson-databind', "${versions.jackson3_databind}")
+                }
             }
         }
 


client/build.gradle
```diff
@@ -117,6 +117,22 @@ publishing {
                         url = "https://github.com/opensearch-project/ml-commons"
                     }
                 }
+                // Add a dependencies section to signal that consumers of the client need to provide these
+                withXml {
+                    def dependencies = asNode().appendNode('dependencies')
+                    
+                    def addProvidedDependency = { group, artifact, version ->
```
ylwu-amzn, line 115: Thanks for fixing this issue. Have you verified this? Any impact to other plugins using ml-commons clients like neural-search?

dbwiddis: This really only changes the POM which wouuld in theory add transitive dependencies, ecxept:
 - we don't do transitive dependencies in OpenSearch (the source of the problem!)
 - it's in scope `<provided>` which means it wouldn't be imported transitively if it did.  This is purely a user documentation change.

dbwiddis: I did test it with flow framework by deleting those imports (and getting runtime errors) even with the new snapshot containing the new values in the pom in my maven local repository, so it confirms these imports aren't included.

PR #4154, opened 2025-09-03
Search Template Support for QueryPlanningTool

joshpalis:
### Description
WIP - Adds support for QueryPlanningTool `user_templates` generation type. 

### Check List


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -70,6 +92,59 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
             listener.onFailure(new IllegalArgumentException("Empty parameters for QueryPlanningTool: " + parameters));
             return;
         }
+
+        if (generationType != USER_SEARCH_TEMPLATES_TYPE_FIELD) {
+            // Use default search template, skip template selection
+            if (!parameters.containsKey(TEMPLATE_FIELD)) {
+                parameters.put(TEMPLATE_FIELD, DEFAULT_SEARCH_TEMPLATE);
+            }
```
rithinpullela: To make it less confusing, should we expect a field called "default_template" from the user instead of "template"

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -70,6 +92,59 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
...
+            return;
+        }
+
+        // Template Selection
+        if (!parameters.containsKey(SYSTEM_PROMPT_FIELD)) {
+            parameters.put(SYSTEM_PROMPT_FIELD, TEMPLATE_SELECTION_SYSTEM_PROMPT);
+        }
+        if (!parameters.containsKey(USER_PROMPT_FIELD)) {
+            parameters.put(USER_PROMPT_FIELD, TEMPLATE_SELECTION_USER_PROMPT);
+        }
```
rithinpullela: If the user provides the parameters  "system_prompt" and "user_prompt", how do we determine if we want to replace the system and default prompts of Template Selection LLM call or the Query Generation LLM call?

Maybe we should fall back to using default prompt for template selection? We can raise another PR to have separate fields for prompts of Template Selection LLM call and Query Generation LLM call in the future?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -70,6 +92,59 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
...
+        if (!parameters.containsKey(SEARCH_TEMPLATES_FIELD)) {
+            parameters.put(SEARCH_TEMPLATES_FIELD, gson.toJson(parameters.get(SEARCH_TEMPLATES_FIELD)));
+        }
```
rithinpullela: Do we need to check the params for SEARCH_TEMPLATES_FIELD? Can we skip the check and directly add the parameter?

We seem to be extracting the search templates in the factory class: ` public QueryPlanningTool create(Map<String, Object> map) {` Lines 245-265

This will also break/invalidate the validation done in Lines 245-265

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -70,6 +92,59 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
...
+
+        ActionListener<T> templateSelectionListener = ActionListener.wrap(r -> {
+            try {
+                String templateId = (String) r;
+                if (templateId == null || templateId.isBlank() || templateId.equals("null")) {
+                    // Default search template if LLM does not choose
+                    if (!parameters.containsKey(TEMPLATE_FIELD)) {
+                        parameters.put(TEMPLATE_FIELD, DEFAULT_SEARCH_TEMPLATE);
+                    }
```
rithinpullela: Same as one of my comments above should we replace this by "default_template"?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -70,6 +92,59 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
...
+                    executeQueryPlanning(parameters, listener);
+                } else {
+                    // Retrieve search template by ID
+                    GetStoredScriptRequest getStoredScriptRequest = new GetStoredScriptRequest(templateId);
+                    client.admin().cluster().getStoredScript(getStoredScriptRequest, ActionListener.wrap(getStoredScriptResponse -> {
+                        if (!parameters.containsKey(TEMPLATE_FIELD)) {
+                            parameters.put(TEMPLATE_FIELD, getStoredScriptResponse.getSource().getSource());
+                        }
```
rithinpullela: We are replacing the template selected by the LLM with the parameter provided by the user "template". I think this is not the ideal behavior, could be confusing to the end users.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -164,4 +164,67 @@ public class QueryPlanningPromptTemplate {
...
+        + TEMPLATE_SELECTION_CRITERIA
+        + "\n"
+        + "==== VALIDATION ====\n"
+        + TEMPLATE_SELECTION_VALIDATION
+        + "\n"
+        + "==== INPUTS ====\n"
+        + TEMPLATE_SELECTION_INPUTS
+        + "\n"
+        + "==== EXAMPLES ====\n"
+        + TEMPLATE_SELECTION_EXAMPLES;
```
rithinpullela: nit: Should we add the TEMPLATE_SELECTION_INPUTS at the end after the examples?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -154,16 +229,43 @@ public QueryPlanningTool create(Map<String, Object> map) {
...
+                if (searchTemplatesObj != null) {
+                    if (searchTemplatesObj instanceof List) {
+                        searchTemplates = (List<Map<String, Object>>) searchTemplatesObj;
+                        // Validate search templates structure
+                        for (Map<String, Object> template : searchTemplates) {
+                            if (!template.containsKey("template_id") || !template.containsKey("template_description")) {
+                                throw new IllegalArgumentException(
+                                    "Each search template must contain 'template_id' and 'template_description' fields"
+                                );
+                            }
```
rithinpullela: nit: Should we ignore fields other than `template_id ` and `template_description `?
If they are added to the `searchTemplates` variable, it might confuse the LLM

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -164,4 +164,67 @@ public class QueryPlanningPromptTemplate {
...
+        + "   - Prefer general-purpose search over analytics if the question asks to “find/search/browse” items; prefer analytics if it asks for “most sold/revenue/total/average”.";
+
+    public static final String TEMPLATE_SELECTION_VALIDATION =
+        "- Your output MUST be exactly one of the provided template ids (regex: ^[A-Za-z0-9_-]+$).\n"
+            + "- If no perfect match exists, pick the closest by the criteria above. Never output “none” or invent an id.";
+
+    public static final String TEMPLATE_SELECTION_INPUTS = "question: ${parameters.query_text}\n"
+        + "templates: ${parameters.search_templates}";
+
+    public static final String TEMPLATE_SELECTION_EXAMPLES = "Example A: \n"
```
rithinpullela, line 206: nit: Adding an example response could help LLM

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: This is a good point, I can create a separate parameters map strictly for the first LLM call for template selection. The original parameters we can keep for use in the query planning call to the LLM

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: We actually dont expect the user to directly provide the `template` field. Its chosen by the llm or its defaulted if a choice is omitted. Currently the default is just a blank string placeholder until we determine which default search template we want to use

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: I had the wrong assumption on tool configuration parsing, seems that non-string objects are just converted to a json string. This string we can pass directly to the LLM.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: The user will never see the `template` field, it is something that we pass to the Query Planning prompt during the second LLM invocation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: I removed the check for `template` within the parameters, as this value will never be provided by the user

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -164,4 +164,67 @@ public class QueryPlanningPromptTemplate {
...
+
+    public static final String TEMPLATE_SELECTION_CRITERIA = "(apply in order)\n"
+        + "1) INTENT MATCH: Identify the user’s primary intent (e.g., product search/browse, analytical reporting, trend/sales analysis, inventory, support lookup). Prefer templates whose descriptions explicitly support that intent.\n"
+        + "2) SIGNAL ALIGNMENT: Count strong lexical/semantic matches between the question and each template’s description/placeholders.\n"
+        + "   - Attribute filters (brand, category, size, color, price, rating, etc.) → favor product/item search templates.\n"
+        + "   - Metrics (sales value, revenue, units sold, conversion, time windows) → favor analytics/aggregation templates.\n"
+        + "   - Temporal phrases (“last week”, “by month”, “trending”, “top sellers”) → favor templates with date/time and aggregations.\n"
+        + "   - Opinion/quality words (“highly rated”, “best”, “top reviewed”) → favor templates with rating/review placeholders.\n"
+        + "3) SPECIFICITY: If multiple templates match, prefer the one whose description/placeholders are the most specific to the question’s entities and constraints.\n"
```
owaiskazi19, line 194: Why are we selecting only one template here? We can multiple templates send back to LLM for context if they are right

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -164,4 +164,67 @@ public class QueryPlanningPromptTemplate {
...
+        + "4) TIE-BREAK:\n"
+        + "   - Prefer templates intended for the user’s domain (e.g., “products” vs “sales analytics”).\n"
```
owaiskazi19, line 196: How would context know about user's domain?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
owaiskazi19: Yes, we should add an example

owaiskazi19 (commented):
Looks good overall. Need tests and can you share few examples of LLm returned templates and then the generated queries?
Also, what's the overall latency?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
joshpalis: We need the LLM to choose one template that we retrieve from the cluster state. Each request to retrieve a template only returns one and submits a cluster state task which is an expensive operation. Retrieving multiple templates would increase this overhead

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
joshpalis: done

rithin-pullela-aws (approved):
Thanks @joshpalis LGTM! Please make sure to add ITs in another PR

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -155,6 +155,7 @@ public class QueryPlanningPromptTemplate {
         + "Query Fields: ${parameters.query_fields:-}\n\n"
         + "==== OUTPUT ====\n"
         + "GIVE THE OUTPUT PART ONLY IN YOUR RESPONSE (a single JSON object)\n"
+        + "Use this template provided by the user as reference to generate the query: ${parameters.template}\n\n"
```
mingshl, line 158: should we different prompt prefix for different generation type? for LLM generated type, we don't have search template, the template would be null. then this prompt might be miss leading through.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
```diff
@@ -164,4 +165,105 @@ public class QueryPlanningPromptTemplate {
...
+        + "          \"type\": \"{{#lex_type}}{{lex_type}}{{/lex_type}}{{^lex_type}}best_fields{{/lex_type}}\","
+        + "          \"operator\": \"{{#lex_operator}}{{lex_operator}}{{/lex_operator}}{{^lex_operator}}or{{/lex_operator}}\","
+        + "          \"boost\": {{#lex_boost}}{{lex_boost}}{{/lex_boost}}{{^lex_boost}}1.0{{/lex_boost}}"
+        + "        }"
+        + "      }{{#sem_enabled}},"
+        + "      {"
+        + "        \"neural\": {"
+        + "          \"{{sem_field}}\": {"
+        + "            \"query_text\": \"{{sem_query_text}}\","
+        + "            \"model_id\": \"{{sem_model_id}}\","
```
mingshl, line 253: for the default search template, we need a model id, should we open up the parameters should that we would let users provide model id?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -154,16 +213,32 @@ public QueryPlanningTool create(Map<String, Object> map) {
...
-            return new QueryPlanningTool(type, queryGenerationTool);
+
+            // Parse search templates if generation type is user_templates
+            String searchTemplates = null;
+            if (USER_SEARCH_TEMPLATES_TYPE_FIELD.equals(type)) {
+                if (!map.containsKey(SEARCH_TEMPLATES_FIELD)) {
+                    throw new IllegalArgumentException("search_templates field is required when generation_type is 'user_templates'");
+                } else {
+                    // array is parsed as a json string
+                    searchTemplates = gson.toJson((String) map.get(SEARCH_TEMPLATES_FIELD));
```
mingshl, line 236: may be use StringUtils.toJosn is better, it has different types conversion in StringUtils package

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
rithinpullela: We will have a default search template in case of llmGeneratedType

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
rithinpullela: Yes, we plan to open up more parameters to enable semantic. But even in this state the users can provide the model ID in their query if they want to write a semantic search query.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java
mingshl: did you test the default search template with the llmGeneratedType? Did it help?

joshpalis: With the LLM generated type, we will provide a default search template, so we will always have a search template

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
joshpalis: I see that the StringUtils.toJson just casts the given object back as a string, in our case we need to use gson.toJson to format the provided string with correctly escaped quotes in order to pass validation. Given this requirement, I will leave this as is

mingshl: I am mainly concerned about the question string has a special character that might break the json. have you test it with the question having special character> for example, "how many order I placed on 07/10?"

joshpalis: just tested this now and gson.toJson still correctly formats the special character. Internally, StringsUtils.toJson still uses gson.toJson so its effectively the same

mingshl: I would rather to keep the repo consistent when using the same package in different places, but if this works for now, I can let this go

joshpalis:
Unrelated tests are failing, @ylwu-amzn @mingshl can you help merge this PR in : 
```
 25-09-11T16:25:27,517][ERROR][o.o.m.r.MLCommonsRestTestCase] [test_bedrock_multimodal_model_empty_imageInput_null_textInput]
```

PR #4162, opened 2025-09-10
[MCP Server] Support Streamable HTTP and deprecate SSE in MCP server

### LINKED ISSUE #3813: [BUG] MCP Server not compatible with Python MCP Client
**What is the bug?**
### Summary

The current OpenSearch MCP server implementation is compatible with the Java SDK but fails with the official Python SDK. The two blocking problems are:

1. Endpoint URL resolution differences between Python & Java SDKs. (easy workaround + small code change)
2. Message‑endpoint response semantics (streaming vs. REST) that violate the MCP/SSE contract expected by the Python SDK. (need to change code)

Fixing these will allow both SDKs to interoperate with the same server.

### 1. Endpoint URL resolution differences:

The current implementation exposes these endpoints:

1. GET /_plugins/_ml/sse
2. POST /_plugins/_ml/sse/message?sessionId

When we do GET /_plugins/_ml/sse it returns 

```
event: endpoint
data: /sse/message?sessionId=06abf5d2-29c1-4cb3-b30f-89bb62c5ab03
```
Observe the end point returned has a leading /

### Python SDK

The python SDK [here](https://github.com/modelcontextprotocol/python-sdk/blob/017135434eab20a289eba78c50c17079f73eb607/src/mcp/client/sse.py#L64C40-L64C78) uses the urllib.parse.urljoin function to create the message endpoint

If the endpoint event retuned begins with /, urljoin discards the _plugins/_ml prefix and produces /sse/message …, which the server rejects.

The constructed URL is now: 

* `http://localhost:9200/sse/message?sessionId= `

The Correct URL should be: `http://localhost:9200/_plugins/_ml/sse/message?sessionId= `

When we try to resolve this by changing the returned endpoint data to have no leading /, like below the JAVA client fails
```
event: endpoint
data: sse/message?sessionId=06abf5d2-29c1-4cb3-b30f-89bb62c5ab03
```
### JAVA SDK

The JAVA SDK [here](https://github.com/modelcontextprotocol/java-sdk/blob/e610d853f922e36ba474b2240f5c6546166e4840/mcp/src/main/java/io/modelcontextprotocol/util/Utils.java#L71-L79) uses java’s URI.resolve() method, which uses different rule.

The current code works because,
```
Base: http://localhost:9200/_plugins/_ml
Endpoint: /sse/message
```
This makes call to URL: `http://localhost:9200/_plugins/_ml/sse/message`

When we respond with:
```
Base: http://localhost:9200/_plugins/_ml
Endpoint: sse/message
```
This makes call to URL: `http://localhost:9200/_plugins/_mlsse/message`
Note the missing / between _ml and sse

### Work Around for JAVA with `sse/message?sessionId=...`
```
McpClientTransport transport = HttpClientSseClientTransport.builder("http://localhost:9200/_plugins/_ml/")
                .sseEndpoint("sse")
                .customizeClient(clientBuilder -> {
            clientBuilder.connectTimeout(Duration.ofSeconds(1));
        }).build();
```
With this:

* The URL ends with / like this: `http://localhost:9200/_plugins/_ml/`
* SSE end point is “sse” different from the default /sse
* So when the GET call is made, the url will be `http://localhost:9200/_plugins/_ml/`
    
* We return sse/message?sessionId=... and this is appended to the base URL
* So the message end point becomes: `http://localhost:9200/_plugins/_ml/sse/message?sessionId=...`

To track the REST calls made using the JAVA client use this code: https://github.com/rithin-pullela-aws/opensearch-mcpserver-test-example

Although we have resolved both JAVA and Python cases for URL, there is another issue that makes it not work with python SDK:

### 2. Message‑endpoint response semantics (streaming vs. REST)

* Python Client: Awaits POST responses - Expects them to complete ([code](https://github.com/modelcontextprotocol/python-sdk/blob/017135434eab20a289eba78c50c17079f73eb607/src/mcp/client/sse.py#L116))
* Java Client: Handles POST as async - Does not wait for completion ([code](https://github.com/modelcontextprotocol/java-sdk/blob/e610d853f922e36ba474b2240f5c6546166e4840/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java#L423))

The current server implementation starts an SSE connection when receiving POST requests to the messages endpoint. This works with Java's non-blocking approach but causes the Python client to hang indefinitely waiting for the response to complete. Eventually it errors out because of read timeout. 

To resolve this issue, we need to make sure our server implementation does not open a streaming endpoint for messages, having a REST endpoint is the right approach. The same can be seen in the server implementations of JAVA and Python MCP SDKs.

If we make the message endpoint which receives the POST call to REST

### TODO:

1. Change the message endpoint returned to sse/message?sessionId=...
2. Update documentation to inform the customers to use custom sse endpoint as “sse” (default is “/sse”)
3. Make messages endpoint REST (no sse)

Try connecting with python client:

 - Plugins

rithin-pullela-aws:
### Description
This PR deprecates SSE server and now provides the Streamable HTTP server.

The new endpoint: `http://localhost:9200/_plugins/_ml/mcp/stream`

- Updated the mcp java sdk to `0.12.1` version

NOTE: The MCP Client (connectors) with the new protocol is not implemented yet. Will do it in a different PR to not clutter the current code.

### Testing:

**Enable MCP Server:**
```
PUT /_cluster/settings/
{
    "persistent": {
        "plugins.ml_commons.mcp_server_enabled":"true"
    }
}
```

**Register a Tool:**
```
POST /_plugins/_ml/mcp/tools/_register
{
    "tools": [
        {
            "type": "ListIndexTool",
            "name": "ListIndexTool"
        }
    ]
}
```

**Mock Client behavior via Rest Calls:**

**1. Initialize Connection:**
```
POST /_plugins/_ml/mcp

{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
        "protocolVersion": "2025-03-26",
        "capabilities": {
            "roots": {
                "listChanged": true
            },
            "sampling": {}
        },
        "clientInfo": {
            "name": "test-client",
            "version": "1.0.0"
        }
    }
}
```
Expected response:
```
200 OK
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "protocolVersion": "2025-03-26",
        "capabilities": {
            "logging": {},
            "prompts": {
                "listChanged": false
            },
            "resources": {
                "subscribe": false,
                "listChanged": false
            },
            "tools": {
                "listChanged": true
            }
        },
        "serverInfo": {
            "name": "OpenSearch-MCP-Stateless-Server",
            "version": "0.1.0"
        },
        "instructions": "OpenSearch MCP Stateless Server - provides access to ML tools without sessions"
    }
}
```

**2. Initialization Complete Notification**
```
POST /_plugins/_ml/mcp
{
    "jsonrpc": "2.0",
    "method": "notifications/initialized",
    "params": {}
}
```
Expected response:
```
202 OK
```

**3. List Available Tools**
```
POST /_plugins/_ml/mcp
{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "tools/list",
    "params": {}
}
```
Expected response:
```
200 OK
{
    "jsonrpc": "2.0",
    "id": 2,
    "result": {
        "tools": [
            {
                "name": "ListIndexTool",
                "description": "This tool returns information about indices in the OpenSearch cluster along with the index `health`, `status`, `index`, `uuid`, `pri`, `rep`, `docs.count`, `docs.deleted`, `store.size`, `pri.store. size `, `pri.store.size`, `pri.store`. Optional arguments: 1. `indices`, a comma-delimited list of one or more indices to get information from (default is an empty list meaning all indices). Use only valid index names. 2. `local`, whether to return information from the local node only instead of the cluster manager node (Default is false)",
                "inputSchema": {
                    "type": "object",
                    "properties": {
                        "indices": {
                            "type": "array",
                            "items": {
                                "type": "string"
                            },
                            "description": "OpenSearch index name list, separated by comma. for example: [\"index1\", \"index2\"], use empty array [] to list all indices in the cluster"
                        }
                    },
                    "additionalProperties": false
                }
            },
            {
                "name": "ListIndexTool",
                "description": "This tool returns information about indices in the OpenSearch cluster along with the index `health`, `status`, `index`, `uuid`, `pri`, `rep`, `docs.count`, `docs.deleted`, `store.size`, `pri.store. size `, `pri.store.size`, `pri.store`. Optional arguments: 1. `indices`, a comma-delimited list of one or more indices to get information from (default is an empty list meaning all indices). Use only valid index names. 2. `local`, whether to return information from the local node only instead of the cluster manager node (Default is false)",
                "inputSchema": {
                    "type": "object",
                    "properties": {
                        "indices": {
                            "type": "array",
                            "items": {
                                "type": "string"
                            },
                            "description": "OpenSearch index name list, separated by comma. for example: [\"index1\", \"index2\"], use empty array [] to list all indices in the cluster"
                        }
                    },
                    "additionalProperties": false
                }
            }
        ]
    }
}
```

**4. Call a Tool**
```
POST /_plugins/_ml/mcp
{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
        "name": "ListIndexTool",
        "arguments": {
            "indices": []
        }
    }
}
```
Expected Response:
```
{
    "jsonrpc": "2.0",
    "id": 3,
    "result": {
        "content": [
            {
                "type": "text",
                "text": "row,health,status,index,uuid,pri(number of primary shards),rep(number of replica shards),docs.count(number of available documents),docs.deleted(number of deleted documents),store.size(store size of primary and replica shards),pri.store.size(store size of primary shards)\n1,green,open,.plugins-ml-config,nKyzDAupTGCwuybs9S_iBA,1,0,1,0,3.9kb,3.9kb\n2,green,open,.plugins-ml-mcp-tools,k1QwQKmXSeqRexmB2JDJiw,1,0,1,0,5kb,5kb\n"
            }
        ],
        "isError": false
    }
}
```
### Other ways of Testing:
You can also test with various MCP clients

**1. Fast MCP client:**
```
import asyncio, logging
from fastmcp import Client

async def main():
    async with Client("http://localhost:9200/_plugins/_ml/mcp") as client:
        for t in await client.list_tools():
            print(t.name)
        r = await client.call_tool("ListIndexTool", {})
        print("result ->", r)

asyncio.run(main())
```
**2. PostMan MCP client:**
In the latest versions of postman we can directly test an MCP server by creating an MCP kind of request
image

### Related Issues
Resolves #3813 since we don't need the extra variable in streamable_http protocol.

### Check List


plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -357,12 +345,7 @@
 import org.opensearch.ml.rest.RestMemorySearchInteractionsAction;
 import org.opensearch.ml.rest.RestMemoryUpdateConversationAction;
 import org.opensearch.ml.rest.RestMemoryUpdateInteractionAction;
-import org.opensearch.ml.rest.mcpserver.RestMLMcpToolsListAction;
-import org.opensearch.ml.rest.mcpserver.RestMLMcpToolsRegisterAction;
-import org.opensearch.ml.rest.mcpserver.RestMLMcpToolsRemoveAction;
-import org.opensearch.ml.rest.mcpserver.RestMLMcpToolsUpdateAction;
-import org.opensearch.ml.rest.mcpserver.RestMcpConnectionMessageStreamingAction;
-import org.opensearch.ml.rest.mcpserver.ToolFactoryWrapper;
+import org.opensearch.ml.rest.mcpserver.*;
```
rithinpullela: Will change this, waiting for CI to pass.

rithinpullela:
The failing ITs are because of usage of deprecated Bedrock models.

```
{"status":404,"error":{"type":"OpenSearchStatusException","reason":"System Error","details":"Error from remote service: {\"message\":\"This model version has reached the end of its life. Please refer to the AWS documentation for more details.\"}"}}
```

Will raise another PR to fix

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
```diff
@@ -0,0 +1,154 @@
...
+import org.opensearch.rest.RestRequest;
+import org.opensearch.transport.client.node.NodeClient;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.modelcontextprotocol.spec.McpSchema;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class RestMcpStatelessStreamingAction extends BaseRestHandler {
```
ylwu-amzn: What does stateless mean? Do we also have or plan to build a stateful action?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+        if (!mlFeatureEnabledSetting.isMcpServerEnabled()) {
+            throw new OpenSearchException(ML_COMMONS_MCP_SERVER_DISABLED_MESSAGE);
+        }
+
+        return channel -> {
+            try {
+                if (request.content() == null) {
+                    sendErrorResponse(channel, null, -32700, "Parse error: empty body");
```
ylwu-amzn: What does this magic number mean: `-32700`? Define these json RPC error code as constants ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,109 @@
...
+                .prompts(false)           // We don't support prompts
+                .build();
+            // Build the server using the transport provider WITHOUT pre-loaded tools
+            // Tools will be loaded dynamically via the sync job
+            log.info("Building MCP server without pre-loaded tools (dynamic loading)...");
+            McpStatelessAsyncServer server = McpServer
+                .async(mcpStatelessServerTransportProvider)
+                .serverInfo("OpenSearch-MCP-Stateless-Server", "0.1.0")
+                .capabilities(serverCapabilities)
+                .instructions("OpenSearch MCP Stateless Server - provides access to ML tools without sessions")
```
ylwu-amzn: `OpenSearch MCP Stateless Server` this name looks confusing, do we have another type of stateful server ?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
```diff
@@ -0,0 +1,154 @@
...
+                final String requestBody = request.content().utf8ToString();
+                if (requestBody == null || requestBody.isBlank()) {
+                    sendErrorResponse(channel, null, -32700, "Parse error: empty body");
+                    return;
+                }
+
+                OpenSearchMcpStatelessServerTransportProvider transportProvider = McpStatelessServerHolder
+                    .getMcpStatelessServerTransportProvider();
+                if (transportProvider == null || !transportProvider.isHandlerReady()) {
+                    log.error("MCP transport provider not ready - server may not be properly initialized");
```
ylwu-amzn: Is this an error ? Change to warn level?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: MCP protocol supports streamable HTTP in a stateful manner as well, but that would require streaming to be enabled.
While we do not have any plans yet, by naming this endpoint stateless we are keeping the door open for potential future use cases. Also it makes it clear to readers what underlying methodology is being used keeping it consistent with MCP JAVA SDK

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: These are defined by the JSON RPC spec which is used by MCP Protocol. These numbers indicate different errors to the client. 

Reference: https://www.jsonrpc.org/specification#:~:text=NOT%20be%20included.-,5.1%20Error%20object,-When%20a%20rpc

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: The description here does not affect the functionality. But I believe it would not hurt to have an explicit mention of the server being stateless.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Yes, this is an error and we are throwing an exception in the next lines of the code.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
ylwu-amzn: For future extension , should we create a new API or extend current one ? If need to extend current one, the sateless name seems not reasonable.

From https://modelcontextprotocol.io/specification/2025-03-26/basic/transports
> The server MUST provide a single HTTP endpoint path (hereafter referred to as the MCP endpoint) that supports both POST and GET methods. For example, this could be a URL like https://example.com/mcp.

Does that mean we should use one HTTP endpoint for both stateless and stateful?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
```diff
@@ -0,0 +1,154 @@
...
+                if (msg instanceof McpSchema.JSONRPCNotification) {
+                    // Notifications: acknowledge and do not process
+                    channel.sendResponse(new BytesRestResponse(RestStatus.ACCEPTED, "application/json", BytesArray.EMPTY));
+                    return;
+                }
+
+                // Requests: capture id for any downstream error mapping
+                final Object id = (msg instanceof McpSchema.JSONRPCRequest) ? ((McpSchema.JSONRPCRequest) msg).id() : null;
+
+                transportProvider.handleRequest(requestBody).subscribe(response -> {
```
ylwu-amzn: Line 86 already called `deserializeJsonRpcMessage` , why not pass `msg` to `handleRequest` method?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    private final ObjectMapper objectMapper;
+    private McpStatelessServerHandler mcpHandler;
+
+    public OpenSearchMcpStatelessServerTransportProvider(ObjectMapper objectMapper) {
+        Assert.notNull(objectMapper, "ObjectMapper must not be null");
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public void setMcpHandler(McpStatelessServerHandler mcpHandler) {
```
ylwu-amzn, line 33: Where is this method being used ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
ylwu-amzn: Can the client side see/get the instructions and serverInfo ?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
ylwu-amzn: I see McpStatelessServerHolder.java line 72 using warn level log.  Let's keep consistent?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpStatelessServerTransportProvider.java
rithinpullela: Used internally by Java SDK framework. Need to override since it is required by the interface:
Reference: https://github.com/modelcontextprotocol/java-sdk/blob/3f70afd21f8064a8b9549c4624c46316fc1ab2c0/mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java#L127

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,109 @@
...
+            return mcpStatelessServerTransportProvider;
+        }
+        synchronized (McpStatelessServerHolder.class) {
+            if (mcpStatelessServerTransportProvider != null) {
+                return mcpStatelessServerTransportProvider;
+            }
+            mcpStatelessServerTransportProvider = new OpenSearchMcpStatelessServerTransportProvider(new ObjectMapper());
+            // initialize the server
+            if (mcpStatelessAsyncServer == null) {
+                mcpStatelessAsyncServer = createMcpStatelessServer(mcpStatelessServerTransportProvider);
```
ylwu-amzn: This method initializes `mcpStatelessAsyncServer` seems indirect. Someone reading the code may not immediately see that `mcpStatelessAsyncServer` is initialized inside the transport provider getter. Consider explicitly initializing the server inside `getMcpStatelessAsyncServerInstance()` for clarity? Or create a new init method.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,109 @@
...
+                .serverInfo("OpenSearch-MCP-Stateless-Server", "0.1.0")
+                .capabilities(serverCapabilities)
+                .instructions("OpenSearch MCP Stateless Server - provides access to ML tools without sessions")
+                .build();
+
+            log.info("Stateless MCP server created successfully");
+
+            // Load tools immediately
+            statelessToolsHelper
+                .autoLoadAllMcpTools(
```
ylwu-amzn: This `McpStatelessServerHolder` invokes `statelessToolsHelper.autoLoadAllMcpTools` method. 
The `statelessToolsHelper.autoLoadAllMcpTools` invokes `McpStatelessServerHolder.getMcpStatelessAsyncServerInstance` method.

This is a bidirectional dependency between the holder and the helper.

Suggest pass the server instance to the helper `autoLoadAllMcpTools` method. 
```
statelessToolsHelper.autoLoadAllMcpTools(server, ...);
```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Yes

ylwu-amzn: If they can see these information, let's design carefully.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Good point, I will refactor the method to accept the McpSchema.JSONRPCMessage instead of the whole requestBody

ylwu-amzn: We need a transport action to control permission on API level.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Will remove the stateless

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpStatelessServerTransportProvider.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+        } catch (Exception e) {
+            log.error("Failed to handle MCP request: {}", e.getMessage(), e);
+            return Mono.error(e);
+        }
+    }
+
+    public boolean isHandlerReady() {
+        boolean ready = mcpHandler != null;
```
ylwu-amzn: Is this enough ? should we also check requestHandlers ready or not , since requestHandlers created in async way

ylwu-amzn:
Could you add some sample HTTP requests to description part?

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Renamed the Rest and Transport Actions to remove "stateless"

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Used constants for JSON-RPC error codes

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Updated the server desc to remove "stateless"

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Removed the warning logging since we do not need to check isHandlerReady anymore

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMcpStatelessStreamingAction.java
rithinpullela: Added a transport Action

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Refactored to have an initialize method

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Refactored such that Tools helper does not depend on holder

plugin/src/main/java/org/opensearch/ml/action/mcpserver/OpenSearchMcpStatelessServerTransportProvider.java
rithinpullela: We do not need to check this since it happens synchronously

rithinpullela:
@ylwu-amzn, updated PR description with test HTTP calls

jiapingzeng (approved):
A note that when we release this in 3.3, we should update blog post https://opensearch.org/blog/introducing-mcp-in-opensearch/ to reflect the changes. This blog post is the top search result for "OpenSearch MCP" so it should be kept up-to-date.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Start the sync job for auto-reloading MCP tools
+     */
+    public void startSyncMcpToolsJob() {
+        ActionListener<Boolean> listener = ActionListener
+            .wrap(r -> { log.debug("Auto reload mcp tools schedule job run successfully!"); }, e -> {
+                log.error(e.getMessage(), e);
+            });
+        threadPool
+            .schedule(() -> autoLoadAllMcpTools(listener), TimeValue.timeValueSeconds(SYNC_MCP_TOOLS_JOB_INTERVAL), GENERAL_THREAD_POOL);
```
ylwu-amzn: Can we use a dedicated thread pool to run this ?

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/server/MLMcpServerRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import lombok.Getter;
+
+public class MLMcpServerRequest extends ActionRequest {
+    @Getter
+    private String requestBody;
+
+    public MLMcpServerRequest(StreamInput in) throws IOException {
+        super(in);
+        this.requestBody = in.readString();
```
ylwu-amzn, line 28: Could requestBody be null?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,162 @@
...
+public class McpStatelessServerHolder {
+
+    private final McpToolsHelper mcpToolsHelper;
+    private final Client client;
+    private final ThreadPool threadPool;
+    private static final int SYNC_MCP_TOOLS_JOB_INTERVAL = 10;
+    public static Map<String, Long> IN_MEMORY_MCP_TOOLS = new ConcurrentHashMap<>();
+    private static volatile McpStatelessAsyncServer mcpStatelessAsyncServer;
+    private static volatile OpenSearchMcpStatelessServerTransportProvider mcpStatelessServerTransportProvider;
+    private static Boolean initialized = false;
```
ylwu-amzn: add `volatile` ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpServerAction.java
```diff
@@ -0,0 +1,128 @@
...
+                } catch (Exception e) {
+                    log.error("Failed to serialize/send response", e);
+                    Map<String, Object> errorMessage = new HashMap<>();
+                    errorMessage.put(MESSAGE_FIELD, "Response serialization failed: " + e.getMessage());
+                    errorMessage.put(ID_FIELD, id);
+                    errorMessage.put(ERROR_CODE_FIELD, JSON_RPC_INTERNAL_ERROR);
+                    listener.onResponse(new MLMcpServerResponse(false, null, errorMessage));
+                }
+            }, error -> {
+                log.error("Failed to handle MCP request", error);
```
ylwu-amzn: Seems duplicate code for the error handling , extract into a method ?

ylwu-amzn:
> 3. List Available Tools

Thanks, why `3. List Available Tools` will return two `"name": "ListIndexTool",` ?

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/requests/server/MLMcpServerRequest.java
rithinpullela: No, it cannot be null. The rest layer will handle null request body.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: Done

rithinpullela:
> > 3. List Available Tools
> 
> Thanks, why `3. List Available Tools` will return two `"name": "ListIndexTool",` ?

@ylwu-amzn it is a typing mistake, we will only see 1 tool

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,169 @@
...
+
+                McpSchema.ServerCapabilities serverCapabilities = McpSchema.ServerCapabilities
+                    .builder()
+                    .tools(true)
+                    .logging()
+                    .resources(false, false)  // We don't support resources
+                    .prompts(false)           // We don't support prompts
+                    .build();
+
+                log.info("Building MCP server without pre-loaded tools (dynamic loading)...");
```
ylwu-amzn: What does "pre-loaded tools" mean?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,169 @@
...
+                log.info("Stateless MCP server created successfully");
+
+                autoLoadAllMcpTools(
+                    ActionListener
+                        .wrap(
+                            success -> log.info("Initial tool loading completed successfully"),
+                            error -> log.error("Initial tool loading failed", error)
+                        )
+                );
+                initialized = true;
```
ylwu-amzn, line 83: I think we should move this line inside `autoLoadAllMcpTools` success branch ? If load MCP tools failed, we should not set `initialized` as `true` right?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Start the sync job for auto-reloading MCP tools
+     */
+    public void startSyncMcpToolsJob() {
+        ActionListener<Boolean> listener = ActionListener
+            .wrap(r -> { log.debug("Auto reload mcp tools schedule job run successfully!"); }, e -> {
+                log.error(e.getMessage(), e);
+            });
+        threadPool
```
ylwu-amzn, line 161: Use dedicated thread pool ? 
Refer to MLExecuteTaskRunner 
```
threadPool.executor(EXECUTE_THREAD_POOL)
```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
```diff
@@ -231,23 +222,16 @@ private SearchRequest buildSearchRequest() {
...
-        });
+    private SearchRequest buildSearchRequest(List<String> toolNames) {
+        SearchRequest searchRequest = new SearchRequest();
+        searchRequest.indices(MLIndex.MCP_TOOLS.getIndexName());
+
+        BoolQueryBuilder queryBuilder = QueryBuilders.boolQuery();
+        toolNames.forEach(toolName -> queryBuilder.should(QueryBuilders.matchQuery("name", toolName)));
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.query(queryBuilder);
+        searchRequest.source(searchSourceBuilder);
```
ylwu-amzn, line 287: Add size ?
```
searchRequest.source().size(MAX_TOOL_NUMBER);
```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
```diff
@@ -184,24 +141,59 @@ public void searchAllToolsWithVersion(ActionListener<Map<String, Tuple<McpToolRe
...
+
+    private ActionListener<SearchResponse> createSearchResponseListener(ActionListener<List<McpToolRegisterInput>> listener) {
+        return ActionListener.wrap(r -> {
+            List<McpToolRegisterInput> mcpTools = new ArrayList<>();
+            Arrays.stream(Objects.requireNonNull(r.getHits().getHits())).forEach(x -> {
+                try {
+                    McpToolRegisterInput mcpTool = parseMcpTool(x.getSourceAsString());
+                    mcpTools.add(mcpTool);
+                } catch (IOException e) {
+                    listener.onFailure(e);
```
ylwu-amzn: If have multiple MCP tools parsing failed, will call this `listener.onFailure(e);` multiple times. Is that ok ? I remember that will trigger some exception. Have you done some test?

Should we use List to track all exceptions, then send back all exceptions with `listener.onFailure(...)` ?

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
```diff
@@ -184,24 +141,59 @@ public void searchAllToolsWithVersion(ActionListener<Map<String, Tuple<McpToolRe
...
+        }, e -> {
+            String errMsg = String.format(Locale.ROOT, "Failed to search mcp tools index with error: %s", e.getMessage());
+            log.error(errMsg, e);
+            listener.onFailure(new OpenSearchException(errMsg));
+        });
+    }
+
+    public void searchToolsWithPrimaryTermAndSeqNo(List<String> toolNames, ActionListener<SearchResponse> listener) {
+        SearchRequest searchRequest = buildSearchRequest(toolNames);
+        searchRequest.source().seqNoAndPrimaryTerm(true);
```
ylwu-amzn, line 215: Add size ?
```
searchRequest.source().size(MAX_TOOL_NUMBER);
```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
```diff
@@ -111,12 +114,17 @@ protected MLMcpToolsRegisterNodeResponse nodeOperation(MLMcpToolsRegisterNodeReq
     private MLMcpToolsRegisterNodeResponse registerToolsOnNode(List<McpToolRegisterInput> mcpTools) {
         AtomicReference<Throwable> exception = new AtomicReference<>();
         Flux.fromStream(mcpTools.stream()).flatMap(tool -> {
-            if (!McpAsyncServerHolder.IN_MEMORY_MCP_TOOLS.containsKey(tool.getName())) {
-                return McpAsyncServerHolder
-                    .getMcpAsyncServerInstance()
+            // Use putIfAbsent to make check-and-act atomic
+            Long previousVersion = McpStatelessServerHolder.IN_MEMORY_MCP_TOOLS.putIfAbsent(tool.getName(), tool.getVersion());
+            if (previousVersion == null) {
```
ylwu-amzn, line 119: What if `previousVersion != null` ? I think we should delete the old version and add the new version, right?

ylwu-amzn:
The current design will only start one MCP server inside OpenSearch cluster with one custom set of tools. That sounds not flexible if multiple users want to have different MCP servers with different sets of tools. No need to change the design in this PR. But some draft idea for supporting multiple MCP servers:

## Register MCP server
```
POST /_plugins/_ml/mcp_servers/_register
{
    "name": "search_index_server",
    "description": "This is a MCP server for searching OpenSearch index.",
    "tools": [
        {
            "type": "ListIndexTool"
        },
        {
            "type": "SearchIndexTool"
        }
    ]
}
```
Sample response
```
{
"mcp_server_id": "abc123"
}
```

## Initialize connection
```
POST /_plugins/_ml/mcp_servers/abc123/mcp

{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
        "protocolVersion": "2025-03-26",
        "capabilities": {
            "roots": {
                "listChanged": true
            },
            "sampling": {}
        },
        "clientInfo": {
            "name": "test-client",
            "version": "1.0.0"
        }
    }
}
```

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: The initialized indicates that the MCP server object has been built successfully

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: We are not creating the server with tools. We are adding tools later. I wanted to convey that with the message, but that seems like an unnecessary detail which cannot lead to any conclusive action from the cx. Will reword this part

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpStatelessServerHolder.java
rithinpullela: I created a thread pool called MCP_TOOLS_SYNC_THREAD_POOL. I am using that already

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
rithinpullela: Makes sense, will modify to return a list of exceptions

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpToolsRegisterOnNodesAction.java
rithinpullela: No, the idea is to prevent race condition with the background job.

So first we index the new tools into the system index and then send out a request to add to MCP server across all the nodes. This part of code that you see is a registration request on each node.

This path is triggered only when a tool is being added to the MCP server for the first time. Before my change, we were trying to update the MCP server directly with this new tool. But the older code did not consider a race condition where the request itself and the background thread pool directly try to update the MCP server.

So when the background job updates the MCP server(by looking at the system index) before we register the tools in this path, we need to backoff and not do any operations. So the check here: `Long previousVersion = McpStatelessServerHolder.IN_MEMORY_MCP_TOOLS.putIfAbsent(tool.getName(), tool.getVersion());` is an atomic operation to add the tool into the MCP server if it is not present in the map. The background job as well does the same. It will first check if the key is present in the map. 

If the key is present then the previous version is not null, that means the key has already been added by the other job. So we need to back off and not add it again.

rithinpullela:
@ylwu-amzn, thanks for the proposal. The main idea of this PR was to change the protocol. The multiple MCP servers part does sound interesting, I will surely look into this alongside other possible designs for the future.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -1169,6 +1166,14 @@ public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
...
             false
         );
+        FixedExecutorBuilder mcpThreadPool = new FixedExecutorBuilder(
+            settings,
+            MCP_TOOLS_SYNC_THREAD_POOL,
+            Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) - 1),
+            10,
+            ML_THREAD_POOL_PREFIX + MCP_TOOLS_SYNC_THREAD_POOL,
+            false
+        );
```
nathaliellenaa, line 1195: Just curious, how did you come up with the number here?

ylwu-amzn (dismissed):

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
rithinpullela: The idea was to provide a minimum number of threads for tool reloading and not put a lot of pressure on the thread pool.

PR #4180, opened 2025-09-18
Add Get Agent to ML Client

owaiskazi19:
### Description
Add Get Agent to ML Client

### Related Issues

### Check List


ylwu-amzn:
nit: this PR has 23 commits , 21 are for merging main branch. Suggest use rebase which can keep the commit history clean

mingshl:
I am thinking whether using the agent API is the only way to decide which type of agent that you are executing cc @rithin-pullela-aws

owaiskazi19:
> I am thinking whether using the agent API is the only way to decide which type of agent that you are executing cc @rithin-pullela-aws

This is for reading the agent type and also the parameters of the agent to see if it already has system prompt available

mingshl:
> I am thinking whether using the agent API is the only way to decide which type of agent that you are executing cc @rithin-pullela-aws

Instead of use logic based on the agent type, should we change the parsing logic in the agent response?

I think we can save one api call just to parse the agent response properly, 

if we enhance the parsing logic here, is it better to save this get api call? 

https://github.com/opensearch-project/neural-search/blob/8ede32df642c7a3acb624a7acd586ff05e3e3ac1/src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java#L130



the other point is, I still don't think that's reasonable that a flow agent and conversational agent has different output format, can we standardize them in 3.3? this is a good time @dhrubo-os @ylwu-amzn @rithin-pullela-aws @pyek-bot

owaiskazi19:
>  Instead of use logic based on the agent type, should we change the parsing logic in the agent response?

Having the same response for all the agent types would be the ideal way, would avoid writing a separate parser based on agent type on the search side.

Anyway, that decision shouldn't block this PR from merging since it's just addition of get agent API to ML Client. We would still need to verify if the user has already provided a system prompt if not add it from search side.

mingshl:
since we already have get model, get tool, it is reasonable to have get agent api, in the ml client.

But I still would like to have a standardize agent response, and have the parsing agent response logic not depending on the agent type

mingshl (approved):

joshpalis (approved):

Issue #4170, opened 2025-09-15
[BUG] MLSdkAsyncHttpResponseHandler reports incorrect generic MLException


brianf-aws:
**What is the bug?**
Currently Ml Commons reports a generic MLException which hides the error underneath

```
            actionListener.onFailure(new MLException("Fail to execute " + action + " in aws connector", e));
```

Our Application uses ML-Commons and extracts the response code from ML Commons in order to declare whether an error is a 5xx or 4xx usually this is done with `OpenSearchStatusException.status().getStatus()` And if it can't be parsed from said exception it reports as a 5xx. This step is incorrect and should instead capture the exception underneath so downstream clients can correctly decide what to do/report.

**How can one reproduce the bug?**
On a high level. Create a connector with a post processing function that will fail on the ml output
1. Perform prediction
2. Observe that the post processing fails and a MLException is reported.

**What is the expected behavior?**
We want to be able to extract what occurred and be able to intelillegently parse what the error was as placing another branch in our application for `if e instanceOf MLException  then parseUntilGetErrorCode(e)` would not be ideal 
**What is your host/environment?**
 - OS: Mac
 - Version OS.2.17
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

PR #4182, opened 2025-09-18
make MLSdkAsyncHttpResponseHandler return IllegalArgumentException

brianf-aws:
### Description
This is a mitigation for https://github.com/opensearch-project/ml-commons/issues/4170

But would need a more thorough code change so that clients can handle properly all situations.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4170 can be closed but discussion can be done to consider different types of failures

### Testing
 ./gradlew spotlessApply
./gradlew test

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
@@ -206,6 +206,8 @@ public class MLSdkAsyncHttpResponseHandler implements SdkAsyncHttpResponseHandle
             ModelTensors tensors = processOutput(action, body, connector, scriptService, parameters, mlGuard);
             tensors.setStatusCode(statusCode);
             actionListener.onResponse(new Tuple<>(executionContext.getSequence(), tensors));
+        } catch (IllegalArgumentException e) {
+            actionListener.onFailure(e);
         } catch (Exception e) {
             log.error("Failed to process response body: {}", body, e);
             actionListener.onFailure(new MLException("Fail to execute " + action + " in aws connector", e));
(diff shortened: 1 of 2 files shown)


Zhangxunmt (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java
```diff
@@ -206,6 +206,8 @@ private void response() {
             ModelTensors tensors = processOutput(action, body, connector, scriptService, parameters, mlGuard);
             tensors.setStatusCode(statusCode);
             actionListener.onResponse(new Tuple<>(executionContext.getSequence(), tensors));
+        } catch (IllegalArgumentException e) {
```
dhrubo-os, line 209: Can we also try to catch the OpensearchStatus Exception too in another catch block?

brianf-aws: Ah good catch would need to look what scenarios cause a OpensearchStatus to occur

brianf-aws: After checking again the try block does not do any other operations that would produce OpensearchStatusException. All the edge cases would only result in a illegalArgumentException

https://github.com/opensearch-project/ml-commons/blob/2a15074ea21a797f94966dc4598e85b8e6a4cb1d/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java#L206

If we happen to find such error we will be able to see if through the log statement on the generic exception which would be very unlikely. The only other method that can throw a exception would be the

connector.parseResponse method which throws a IOException

https://github.com/opensearch-project/ml-commons/blob/5cabf63fb8b8e730e98df42fbf91f440aff58d59/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java#L257


TLDR: adding a catch block OpensearchStatusException would be dead code, and making a test for it wouldnt be useful as we would just be mocking a scenario that doesn't exist

Issue #3630, opened 2025-03-07
[RFC] Remote Model Inference Streaming


jngz-es:
# [RFC] Remote Model Inference Streaming

## Overview

* Popular LLMs provide the ability to stream responses back to a client for better customer experience, like [OpenAI streaming](https://platform.openai.com/docs/api-reference/streaming) etc.
* By supporting this feature, OpenSearch ML provides the ability to integrate in a streaming system as a component.

## Scope

* Support remote model only at the first step.
* Provide a new REST streaming model prediction API.
* Introduce the feature as experimental. The required dependent features like arrow flight, rest streaming channel, reactor-netty4 etc., either the development is still ongoing or they are experimental too. 

## Out of Scope

* Support agent/tool streaming execution.

## Challenges

* REST streaming API. The default external HTTP-based communication module `transport-netty4` not support streaming. We have to change the network settings to use another module `transport-reactor-netty4` (experimental) which supports a new rest channel, [StreamingRestChannel](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/StreamingRestChannel.java). Accordingly we need to implement the REST action in reactor way.
* Internal TCP-based communication between nodes. Currently OpenSearch doesn’t have a network module supporting it in streaming way. Fortunately we will have a new feature being released in 3.0 which is about Arrow Flight. More details from [Search streams using Apache Arrow and Flight](https://github.com/opensearch-project/OpenSearch/issues/16679) and [Arrow Flight Server bootstrap logic](https://github.com/opensearch-project/OpenSearch/pull/16962). With this feature, we are able to stream LLM responses between coordinate node and ml node.
* Support [SSE](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events) between OpenSearch and LLM. Popular LLMs like OpenAI, Claude etc. follow SSE standard. For different LLMs, they recommend to use their clients as it is the easiest way to interact with the streaming API. But from our end, it’s not the case. 1. Usually they don’t have java clients. 2. We need to support many LLMs and don’t want to introduce a different client for every LLM.

## System Architecture

![Image](https://github.com/user-attachments/assets/eda8a5b1-cde7-41ea-b239-49c3f278173c)

## Apache Arrow

OpenSearch is introducing [Apache Arrow](https://github.com/apache/arrow) in 3.0 release. There are some key takeaways from arrow for our use cases.

[...]

PR #4187, opened 2025-09-23
[FEATURE] Predict Stream

nathaliellenaa:
### Description
Add a new predict stream API as an experimental feature to support streaming predictions. The implementation currently utilized `transport-reactor-netty4` for REST streaming. This is a temporary solution as we need to do more research to determine a long-term approach because `transport-reactor-netty4` from the OpenSearch core is still experimental.

This predict stream API currently supports 2 models:
- [OpenAI Chat Completion model](https://platform.openai.com/docs/api-reference/completions)
- [Bedrock Converse Stream model](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html)

> Note: This PR only covers predict streaming, will open another PR to cover agent streaming

**API Endpoint:**
```
POST /_plugins/_ml/models/{model_id}/_predict/stream
```

**Sample workflow:**
```
// Enable feature flag
PUT /_cluster/settings
{
    "persistent": {
        "plugins.ml_commons.stream_enabled": true
    }
}

// Register OpenAI chat completion model
{
    "name": "openai gpt 3.5 turbo",
    "function_name": "remote",
    "description": "openai model",
    "connector": {
        "name": "OpenAI Chat Connector",
        "description": "The connector to public OpenAI model service for GPT 3.5",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "endpoint": "api.openai.com",
            "model": "gpt-3.5-turbo"
        },
        "credential": {
            "openAI_key": "<your_api_key>"
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
                "headers": {
                },
                "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": ${parameters.messages} }",
                "response_filter": "$.choices[0].delta.content"
            }
        ]
    }
}

// Run predict stream API
POST /_plugins/_ml/models/mm92cpkBiefJiMfDtq8_/_predict/stream
{
  "parameters": {
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "Can you summarize Prince Hamlet of William Shakespeare in around 100 words?"
      }
    ],
    "_llm_interface": "openai/v1/chat/completions"
  }
}

// Sample response
data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":"Sure","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":"!","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":" \"","is_last":false}}]}]}

....
data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":" psyche","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":".","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":"","is_last":true}}]}]}
```

**Error handling:**
1. Invalid model ID
```
{
    "error": {
        "root_cause": [
            {
                "type": "status_exception",
                "reason": "Failed to find model"
            }
        ],
        "type": "status_exception",
        "reason": "Failed to find model"
    },
    "status": 404
}
```
2. Error from remote service (e.g., invalid payload, invalid API key, etc)
Error message will be sent as one chunk and the connection will be closed
```
data: {"error": "Error from remote service: {\n  \"error\": {\n    \"message\": \"Missing required parameter: 'messages[1].role'. You provided 'roles', did you mean to provide 'role'?\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"messages[1].role\",\n    \"code\": \"missing_required_parameter\"\n  }\n}"}
```
3. Error in the middle of streaming
Error message will be sent as one chunk and the connection will be closed
```
data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":"tering his father's ghost, who","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"response","dataAsMap":{"content":" claims Claudius murdered him, Hamlet struggles","is_last":false}}]}]}

data: {"error": "Error from remote service: Too many requests, please wait before trying again. You have sent too many requests.  Wait before trying again. (Service: bedrock, Status Code: 429, Request ID: xxx)"}
```

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3630

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -367,6 +367,10 @@ public class HttpConnector extends AbstractConnector {
     }
 
     private boolean neededStreamParameterInPayload(Map<String, String> parameters) {
+        if (parameters == null) {
+            return false;
+        }
+
         boolean isStream = parameters.containsKey("stream");
         if (!isStream) {
             return false;
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -19,12 +19,14 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.text.StringEscapeUtils;
 import org.apache.commons.text.StringSubstitutor;
 import org.opensearch.Version;
 import org.opensearch.common.io.stream.BytesStreamOutput;
@@ -35,6 +37,9 @@ import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.AccessMode;
 import org.opensearch.ml.common.transport.connector.MLCreateConnectorInput;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
@@ -351,10 +356,35 @@ public class HttpConnector extends AbstractConnector {
 
             if (!isJson(payload)) {
                 throw new IllegalArgumentException("Invalid payload: " + payload);
+            } else if (neededStreamParameterInPayload(parameters)) {
+                JsonObject jsonObject = JsonParser.parseString(payload).getAsJsonObject();
+                jsonObject.addProperty("stream", true);
+                payload = jsonObject.toString();
             }
             return (T) payload;
         }
         return (T) parameters.get("http_body");
+    }
+
+    private boolean neededStreamParameterInPayload(Map<String, String> parameters) {
+        boolean isStream = parameters.containsKey("stream");
+        if (!isStream) {
+            return false;
+        }
+
+        String llmInterface = parameters.get("_llm_interface");
+        if (llmInterface.isBlank()) {
+            return false;
+        }
+
+        llmInterface = llmInterface.trim().toLowerCase(Locale.ROOT);
+        llmInterface = StringEscapeUtils.unescapeJava(llmInterface);
+        switch (llmInterface) {
+            case "openai/v1/chat/completions":
+                return true;
+            default:
+                return false;
+        }
     }
 
     protected String fillNullParameters(Map<String, String> parameters, String payload) {
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
@@ -1,3 +1,4 @@
+
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
@@ -6,16 +7,22 @@
 package org.opensearch.ml.engine.algorithms.remote;
 
 import static org.opensearch.ml.common.connector.ConnectorProtocols.AWS_SIGV4;
+import static org.opensearch.ml.engine.algorithms.agent.AgentUtils.LLM_INTERFACE_BEDROCK_CONVERSE_CLAUDE;
+import static org.opensearch.ml.engine.algorithms.agent.MLChatAgentRunner.LLM_INTERFACE;
 import static software.amazon.awssdk.http.SdkHttpMethod.GET;
 import static software.amazon.awssdk.http.SdkHttpMethod.POST;
 
 import java.security.AccessController;
+import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.time.Duration;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.commons.text.StringEscapeUtils;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.common.util.TokenBucket;
@@ -26,18 +33,31 @@ import org.opensearch.ml.common.exception.MLException;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.model.MLGuard;
 import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.transport.MLTaskResponse;
 import org.opensearch.ml.engine.annotation.ConnectorExecutor;
 import org.opensearch.ml.engine.httpclient.MLHttpClientFactory;
 import org.opensearch.script.ScriptService;
+import org.opensearch.transport.StreamTransportService;
 import org.opensearch.transport.client.Client;
 
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
+import okhttp3.OkHttpClient;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.core.internal.http.async.SimpleHttpContentPublisher;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.http.async.AsyncExecuteRequest;
 import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeAsyncClient;
+import software.amazon.awssdk.services.bedrockruntime.model.ContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ContentBlockDeltaEvent;
+import software.amazon.awssdk.services.bedrockruntime.model.ConverseStreamRequest;
+import software.amazon.awssdk.services.bedrockruntime.model.ConverseStreamResponseHandler;
+import software.amazon.awssdk.services.bedrockruntime.model.Message;
 
 @Log4j2
 @ConnectorExecutor(AWS_SIGV4)
@@ -63,6 +83,16 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
 
     private SdkAsyncHttpClient httpClient;
 
+    private OkHttpClient okHttpClient;
+
+    private BedrockRuntimeAsyncClient bedrockRuntimeAsyncClient;
+
+    @Setter
+    @Getter
+    private StreamTransportService streamTransportService;
+
+    private AtomicBoolean isStreamClosed = new AtomicBoolean(false);
+
     public AwsConnectorExecutor(Connector connector) {
         super.initialize(connector);
         this.connector = (AwsConnector) connector;
@@ -70,6 +100,19 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
         Duration readTimeout = Duration.ofSeconds(super.getConnectorClientConfig().getReadTimeout());
         Integer maxConnection = super.getConnectorClientConfig().getMaxConnections();
         this.httpClient = MLHttpClientFactory.getAsyncHttpClient(connectionTimeout, readTimeout, maxConnection);
+        this.bedrockRuntimeAsyncClient = buildBedrockRuntimeAsyncClient(httpClient);
+        try {
+            AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                this.okHttpClient = new OkHttpClient.Builder()
+                    .connectTimeout(10, TimeUnit.SECONDS)
+                    .readTimeout(1, TimeUnit.MINUTES)
+                    .retryOnConnectionFailure(true)
+                    .build();
+                return null;
+            });
+        } catch (PrivilegedActionException e) {
+            throw new RuntimeException("Failed to build OkHttpClient.", e);
+        }
     }
 
     @Override
@@ -126,6 +169,70 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
         }
     }
 
+    @Override
+    public void invokeRemoteServiceStream(
+        String action,
+        MLInput mlInput,
+        Map<String, String> parameters,
+        String payload,
+        ExecutionContext executionContext,
+        StreamPredictActionListener<MLTaskResponse, ?> actionListener
+    ) {
+        try {
+            String llmInterface = parameters.get(LLM_INTERFACE);
+            llmInterface = llmInterface.trim().toLowerCase(Locale.ROOT);
+            llmInterface = StringEscapeUtils.unescapeJava(llmInterface);
+            validateLLMInterface(llmInterface);
+
+            ConverseStreamRequest request = ConverseStreamRequest
+                .builder()
+                .modelId(parameters.get("model"))
+                .messages(Message.builder().role("user").content(ContentBlock.builder().text(parameters.get("inputs")).build()).build())
+                .build();
+
+            ConverseStreamResponseHandler handler = ConverseStreamResponseHandler.builder().onResponse(response -> {
+                // Handle initial response
+                log.debug("Initial converse stream response: {}", response);
+            }).onError(error -> {
+                // Handle errors
+                log.error("Converse stream error: {}", error.getMessage());
+            }).onComplete(() -> {
+                // Handle completion
+                log.debug("Converse stream complete");
+                sendCompletionResponse(isStreamClosed, actionListener);
+            }).subscriber(event -> {
+                log.debug("Converse stream event: {}", event);
+                switch (event.sdkEventType()) {
+                    case CONTENT_BLOCK_DELTA:
+                        ContentBlockDeltaEvent contentEvent = (ContentBlockDeltaEvent) event;
+                        String chunk = contentEvent.delta().text();
+                        sendContentResponse(chunk, false, actionListener);
+                        break;
+                    default:
+                        // Ignore the other event types for now.
+                        break;
+                }
+            }).build();
+            bedrockRuntimeAsyncClient.converseStream(request, handler);
+        } catch (Exception e) {
+            log.error("Failed to execute streaming", e);
+            actionListener.onFailure(new MLException("Fail to execute streaming", e));
+        }
+    }
+
+    private BedrockRuntimeAsyncClient buildBedrockRuntimeAsyncClient(SdkAsyncHttpClient sdkAsyncHttpClient) {
+        AwsSessionCredentials credentials = AwsSessionCredentials
+            .create(connector.getAccessKey(), connector.getSecretKey(), connector.getSessionToken());
+        AwsCredentialsProvider awsCredentialsProvider = StaticCredentialsProvider.create(credentials);
+
+        return BedrockRuntimeAsyncClient
+            .builder()
+            .region(Region.of(connector.getRegion()))
+            .credentialsProvider(awsCredentialsProvider)
+            .httpClient(sdkAsyncHttpClient)
+            .build();
+    }
+
     private SdkHttpFullRequest signRequest(SdkHttpFullRequest request) {
         String accessKey = connector.getAccessKey();
         String secretKey = connector.getSecretKey();
@@ -134,5 +241,14 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
         String region = connector.getRegion();
 
         return ConnectorUtils.signRequest(request, accessKey, secretKey, sessionToken, signingName, region);
+    }
+
+    private void validateLLMInterface(String llmInterface) {
+        switch (llmInterface) {
+            case LLM_INTERFACE_BEDROCK_CONVERSE_CLAUDE:
+                break;
+            default:
+                throw new MLException(String.format("Unsupported llm interface: %s", llmInterface));
+        }
     }
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
@@ -13,13 +13,11 @@ import static software.amazon.awssdk.http.SdkHttpMethod.GET;
 import static software.amazon.awssdk.http.SdkHttpMethod.POST;
 
 import java.security.AccessController;
-import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.time.Duration;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.text.StringEscapeUtils;
@@ -43,7 +41,6 @@ import org.opensearch.transport.client.Client;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
-import okhttp3.OkHttpClient;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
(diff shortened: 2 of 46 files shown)


nathaliellenaa:
Will push more commit to add UTs

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -63,13 +83,36 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
 
     private SdkAsyncHttpClient httpClient;
 
+    private OkHttpClient okHttpClient;
```
ylwu-amzn: I see we have `bedrockRuntimeAsyncClient`, do we still need this `okHttpClient` ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/StreamPredictActionListener.java
```diff
@@ -0,0 +1,55 @@
+package org.opensearch.ml.engine.algorithms.remote;
```
ylwu-amzn, line 6: Add license header

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
nathaliellenaa: Right, we don't need `okHttpClient`. Will remove it

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -141,4 +206,101 @@ private void validateHttpClientParameters(String action, Map<String, String> par
...
+         * @param t The error object
+         * @param response The response
+         */
+        @Override
+        public void onFailure(EventSource eventSource, Throwable t, Response response) {
+            log.error("SSE failure.");
+            if (t != null) {
+                log.error("Error: " + t.getMessage(), t);
+                if (t instanceof StreamResetException && t.getMessage().contains("NO_ERROR")) {
+                    // TODO: reconnect
```
dhrubo-os, line 282: are we planning to keep like this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/StreamPredictActionListener.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    @Override
+    public final void onResponse(Response response) {
+        onStreamResponse(response, true);
+    }
+
+    @Override
+    public void onFailure(Exception e) {
+        try {
+            channel.sendResponse(e);
```
dhrubo-os: Don't we need to `completeStream` here? What is the actual expectation during Failure?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -63,13 +83,36 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
...
+
+    private BedrockRuntimeAsyncClient bedrockRuntimeAsyncClient;
+
+    @Setter
+    @Getter
+    private StreamTransportService streamTransportService;
+
+    private AtomicBoolean isStreamClosed = new AtomicBoolean(false);
```
dhrubo-os: This will be shared across multiple streaming requests (I assume), which can cause race conditions between concurrent streams?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
nathaliellenaa: Yes, we will keep it like this for current implementation. We can implement this as enhancement in the future

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/StreamPredictActionListener.java
nathaliellenaa: The connection should be closed during failure and the error message should be displayed in regular HTTP response, not streaming. Working on enhancing this part.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
nathaliellenaa: Right, I moved the shared instance variable `isStreamClosed` to a local variable inside `invokeRemoteServiceStream` method. This way the flag will be local per request. I tested it with concurrent requests and confirmed this change eliminates race condition. https://github.com/opensearch-project/ml-commons/blob/dbd3f0aa75774609a760f04ff994afca38ba0d4d/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java#L154-L163

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/StreamPredictActionListener.java
nathaliellenaa: The model validation check occurs before streaming start, so it will be send as a regular HTTP response. However, remote service error happens after the streaming connection is established, so it cannot be sent as a regular HTTP response. Current implementation is to send the remote service error message as one data chunk and close the connection. I put more details about the error handling in the PR description.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionStreamAction.java
```diff
@@ -0,0 +1,407 @@
...
+            @Override
+            public void handleStreamResponse(StreamTransportResponse<MLTaskResponse> streamResponse) {
+                try {
+                    MLTaskResponse response = streamResponse.nextResponse();
+                    if (response != null) {
+                        HttpChunk responseChunk = convertToHttpChunk(response);
+                        channel.sendChunk(responseChunk);
+
+                        // Recursively handle the next response
+                        client.threadPool().executor(STREAM_PREDICT_THREAD_POOL).execute(() -> handleStreamResponse(streamResponse));
```
rishabhmaurya, line 262: Why are we handling it recursively and why each response is handled by a separate thread? Isn't is sequential processing?

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionStreamAction.java
```diff
@@ -0,0 +1,407 @@
...
+                    } else {
+                        log.info("No more responses, closing stream");
```
rishabhmaurya, line 264: close the `streamResponse` here and at any other place where you are using `handleStreamResponse()`

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionStreamAction.java
```diff
@@ -0,0 +1,407 @@
...
+            }
+
+            @Override
+            public MLTaskResponse read(StreamInput in) throws IOException {
+                return new MLTaskResponse(in);
+            }
+        };
+
+        // Send the streaming request using transport service
+        TransportPredictionStreamTaskAction.streamTransportService
```
rishabhmaurya: I didn't like this static declaration of stream transport service. I have created issue on opensearch to address this shortcoming - https://github.com/opensearch-project/OpenSearch/issues/19439

For now, let's define it as a singleton instance in `TransportPredictionStreamTaskAction` instead of static variable.

plugin/src/main/java/org/opensearch/ml/action/prediction/TransportPredictionStreamTaskAction.java
```diff
@@ -0,0 +1,277 @@
...
+            mlPredictionTaskRequest.setUser(user);
+        }
+        final User userInfo = user;
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            ActionListener<MLTaskResponse> wrappedListener = ActionListener.runBefore(listener, context::restore);
+            MLModel cachedMlModel = modelCacheHelper.getModelInfo(modelId);
+            ActionListener<MLModel> modelActionListener = new ActionListener<>() {
+                @Override
+                public void onResponse(MLModel mlModel) {
```
rishabhmaurya, line 161: can we improve the readability here

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionStreamAction.java
nathaliellenaa: The recursion ensures that responses within each individual stream are processed sequentially in the correct order, the threading allows multiple concurrent streams to be handled simultaneously without blocking each other

rishabhmaurya:
@nathaliellenaa  I have looked from stream transport perspective and changes looks good other than the few comments I have added. I would suggest to run a load test and check for both performance and resource leaks.

plugin/src/main/java/org/opensearch/ml/rest/RestMLPredictionStreamAction.java
nathaliellenaa: Discussed offline with @rishabhmaurya, the current recursive thread pool implementation for handling streaming responses has some potential issues. Each streaming responses create a new thread from the pool, which could lead to thread pool exhaustion when handling multiple streaming requests. 

The alternative implementations to call `channel.sendChunk` in the thread or remove the thread pool usage entirely will introduce additional latency to the streaming response processing. For now, we will maintain the current implementation, I'll open an issue to track and fix this in the future.

plugin/build.gradle
```diff
@@ -429,6 +429,10 @@ configurations.all {
     resolutionStrategy.force "org.opensearch:opensearch:${opensearch_version}"
     resolutionStrategy.force "org.bouncycastle:bcprov-jdk18on:1.78.1"
     resolutionStrategy.force 'io.projectreactor:reactor-core:3.7.0'
+    resolutionStrategy.force "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10"
```
ylwu-amzn, line 436: Curious , why we need kotlin dependency?

nathaliellenaa: We use okhttp 4.12.0 which depends on kotlin

mingshl:
I took a skim through the PR, I have a high level question, the implementation for openai and bedrock models are totally different. 

-Bedrock: Uses AWS SDK client (BedrockRuntimeAsyncClient) with native streaming
-OpenAI: Uses OkHttp with SSE parsing and manual event handling

This creates two completely different code paths that are hard to maintain and extend. Have you ever considering making a generic StreamingHandler interface and use factory pattern for the handler, so that we can a unified streaming client interface that can be implemented by different model providers.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -351,12 +356,41 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+            return false;
+        }
+
+        boolean isStream = parameters.containsKey("stream");
+        if (!isStream) {
+            return false;
+        }
+
+        String llmInterface = parameters.get("_llm_interface");
+        if (llmInterface.isBlank()) {
```
dhrubo-os, line 380: `StringUtils.isBlank()` check null, otherwise this one may throw null exception.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -141,4 +207,112 @@ private void validateHttpClientParameters(String action, Map<String, String> par
...
+                break;
+            default:
+                throw new MLException(String.format("Unsupported llm interface: %s", llmInterface));
+        }
+    }
+
+    public final class HTTPEventSourceListener extends EventSourceListener {
+        private StreamPredictActionListener<MLTaskResponse, ?> streamActionListener;
+        private final String llmInterface;
+        private volatile AtomicBoolean isStreamClosed;
```
dhrubo-os, line 223: AtomicBoolean already provides thread-safe operations; volatile is unnecessary.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -133,6 +169,36 @@ public void invokeRemoteService(
...
+        MLInput mlInput,
+        Map<String, String> parameters,
+        String payload,
+        ExecutionContext executionContext,
+        StreamPredictActionListener<MLTaskResponse, ?> actionListener
+    ) {
+        try {
+            String llmInterface = parameters.get(LLM_INTERFACE);
+            llmInterface = llmInterface.trim().toLowerCase(Locale.ROOT);
+            llmInterface = StringEscapeUtils.unescapeJava(llmInterface);
```
dhrubo-os, line 184:
```
String llmInterface = StringEscapeUtils.unescapeJava(
    parameters.get(LLM_INTERFACE).trim().toLowerCase(Locale.ROOT)
);
```

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -351,12 +356,41 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+            return false;
+        }
+
+        llmInterface = llmInterface.trim().toLowerCase(Locale.ROOT);
+        llmInterface = StringEscapeUtils.unescapeJava(llmInterface);
+        switch (llmInterface) {
+            case "openai/v1/chat/completions":
```
dhrubo-os, line 387: [nit] can we put this in a static variable? 

Also why only for openai?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -63,13 +80,20 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
 
     private SdkAsyncHttpClient httpClient;
 
+    private BedrockRuntimeAsyncClient bedrockRuntimeAsyncClient;
```
dhrubo-os, line 83: this creates a hard dependency on Bedrock SDK for ALL AWS connectors, even if they're not using Bedrock.

nathaliellenaa:
> This creates two completely different code paths that are hard to maintain and extend. Have you ever considering making a generic StreamingHandler interface and use factory pattern for the handler, so that we can a unified streaming client interface that can be implemented by different model providers.

@mingshl Right, I agree with this. We can make a generic stream handler so it will be easier to extend this for other model providers. Created streaming handler factory in agent stream PR https://github.com/opensearch-project/ml-commons/pull/4212

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -126,6 +150,83 @@ public void invokeRemoteService(
...
+            }).subscriber(event -> {
+                log.debug("Converse stream event: {}", event);
+                switch (event.sdkEventType()) {
+                    case CONTENT_BLOCK_DELTA:
+                        ContentBlockDeltaEvent contentEvent = (ContentBlockDeltaEvent) event;
+                        String chunk = contentEvent.delta().text();
+                        sendContentResponse(chunk, false, actionListener);
+                        break;
+                    default:
+                        // Ignore the other event types for now.
```
dhrubo-os, line 195: Just curious what is the experience here? Can we throw an client side exception like bad request or not supported?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -135,4 +236,13 @@ private SdkHttpFullRequest signRequest(SdkHttpFullRequest request) {
 
         return ConnectorUtils.signRequest(request, accessKey, secretKey, sessionToken, signingName, region);
     }
+
+    private void validateLLMInterface(String llmInterface) {
+        switch (llmInterface) {
+            case LLM_INTERFACE_BEDROCK_CONVERSE_CLAUDE:
+                break;
+            default:
+                throw new MLException(String.format("Unsupported llm interface: %s", llmInterface));
```
dhrubo-os, line 245: `MLException` is a 5XX error. Should this be 4XX instead?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -358,6 +360,32 @@ public static SdkHttpFullRequest buildSdkRequest(
         return builder.build();
     }
 
+    public static Request buildOKHttpRequestPOST(String action, Connector connector, Map<String, String> parameters, String payload) {
+        okhttp3.RequestBody requestBody;
+        if (payload != null) {
+            requestBody = okhttp3.RequestBody.create(payload, MediaType.parse("application/json; charset=utf-8"));
+        } else {
+            throw new IllegalArgumentException("Content length is 0. Aborting request to remote model");
```
dhrubo-os, line 368: What happens after that? What does customer see finally?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/HttpJsonConnectorExecutor.java
```diff
@@ -68,13 +87,30 @@ public class HttpJsonConnectorExecutor extends AbstractConnectorExecutor {
...
         Duration readTimeout = Duration.ofSeconds(super.getConnectorClientConfig().getReadTimeout());
         Integer maxConnection = super.getConnectorClientConfig().getMaxConnections();
         this.httpClient = MLHttpClientFactory.getAsyncHttpClient(connectionTimeout, readTimeout, maxConnection);
+        try {
+            AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
+                this.okHttpClient = new OkHttpClient.Builder()
+                    .connectTimeout(10, TimeUnit.SECONDS)
+                    .readTimeout(1, TimeUnit.MINUTES)
+                    .retryOnConnectionFailure(true)
+                    .build();
```
dhrubo-os, line 108: shouldn't these come from the values customers are setting in their connector configuration?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -358,6 +360,32 @@ public static SdkHttpFullRequest buildSdkRequest(
...
+        Request.Builder requestBuilder = new Request.Builder();
+        Map<String, String> headers = connector.getDecryptedHeaders();
+        if (headers != null) {
+            for (String key : headers.keySet()) {
+                requestBuilder.addHeader(key, headers.get(key));
+            }
+        }
+        requestBuilder.addHeader("Accept-Encoding", "");
+        requestBuilder.addHeader("Accept", "text/event-stream");
+        requestBuilder.addHeader("Cache-Control", "no-cache");
```
dhrubo-os, line 381: These headers are for Server sent events, right?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -358,6 +360,32 @@ public static SdkHttpFullRequest buildSdkRequest(
         return builder.build();
     }
 
+    public static Request buildOKHttpRequestPOST(String action, Connector connector, Map<String, String> parameters, String payload) {
```
dhrubo-os, line 363: it's specifically for streaming requests, not general POST requests. Can we rename the method accordingly?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -358,6 +360,32 @@ public static SdkHttpFullRequest buildSdkRequest(
...
+        okhttp3.RequestBody requestBody;
+        if (payload != null) {
+            requestBody = okhttp3.RequestBody.create(payload, MediaType.parse("application/json; charset=utf-8"));
+        } else {
+            throw new IllegalArgumentException("Content length is 0. Aborting request to remote model");
+        }
+
+        String endpoint = connector.getActionEndpoint(action, parameters);
```
dhrubo-os, line 371: Do we need any endpoint validation?

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
nathaliellenaa: Currently we support 2 models for streaming, openai and bedrock. For bedrock, we don't need to put stream parameter in the payload.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Yes, they're SSE headers

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Will rename it to `buildOKHttpStreamingRequest`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
nathaliellenaa: Will change it to `IllegalArgumentException`

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: The connection will be closed and user will see this error message
```
{
    "error": "Content length is 0. Aborting request to remote model"
}
```

ylwu-amzn (approved):
Approved to unblock agent stream PR. 
Please add more unit test to increase coverage in new PR.

dhrubo-os (approved):

dhrubo-os:
In the follow up PR let's address the comments and also improve the code coverage. Thanks

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
dhrubo-os: This is not a helpful message and also this is a 4xx error. At this point as customer I won't have any idea what to do to fix this issue.

nathaliellenaa:
Thanks, will address all comments and improve coverage in the agent stream PR

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
nathaliellenaa: Moved all Bedrock SDK related imports and implementation to a separate [BedrockStreamingHandler](https://github.com/opensearch-project/ml-commons/blob/cfe0cb43a2c4a44023021074a9c6159886dca5a1/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java) class and used reflection to access the class.

https://github.com/opensearch-project/ml-commons/blob/cfe0cb43a2c4a44023021074a9c6159886dca5a1/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/StreamingHandlerFactory.java#L39-L45

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
nathaliellenaa: https://github.com/opensearch-project/ml-commons/blob/cfe0cb43a2c4a44023021074a9c6159886dca5a1/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java#L211-L214

PR #4188, opened 2025-09-23
Working memory related APIs

b4sjoo:
### Description

### Related Issues

### Check List


ylwu-amzn:
Suggest publish PR to `feature/agentic_memory_new` branch first, so can review the changes more easily.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -42,20 +44,17 @@ public class MemoryContainerConstants {
     public static final String LONG_TERM_MEMORY_INDEX = "long_term_memory_index";
     public static final String LONG_TERM_MEMORY_HISTORY_INDEX = "long_term_memory_history_index";
 
-
     // Memory data index field names
     public static final String USER_ID_FIELD = "user_id";
     public static final String AGENT_ID_FIELD = "agent_id";
     public static final String SESSION_ID_FIELD = "session_id";
-    public static final String SHORT_TERM_MEMORY_ID_FIELD = "short_term_memory_id";
+    public static final String EVENT_ID_FIELD = "event_id";
```
ylwu-amzn: We are not event based architecture, using `event_id` seems confusing?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
 
 import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
-import org.opensearch.ml.common.memorycontainer.ShortTermMemoryType;
-import org.opensearch.ml.common.utils.StringUtils;
 
 /**
- * Input data for adding memory to a memory container
+ * Input payload for creating short-term events in a memory container
```
ylwu-amzn: This is not just trigger short-term memory creation. It will also trigger long-term is the long term strategies configured in container. 
Based on this understanding, I think we can just keep the original name `MLAddMemoriesInput`

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
+        // Auto-detect memory type and set infer defaults
+        if (messages != null && !messages.isEmpty()) {
+            this.memoryType = ShortTermMemoryType.CONVERSATIONAL;
+            this.infer = infer; // User can override
+        } else if (data != null) {
+            this.memoryType = ShortTermMemoryType.DATA;
+            this.infer = false; // Always false for data
+        } else {
+            this.memoryType = memoryType; // Use provided type if neither messages nor data
+        }
```
ylwu-amzn: We should always set memory type as user provided value, no matter the messages/data value is null or not.

```suggestion
        if (memoryType != null) {
            this.memoryType = memoryType;
        } else if (messages != null && !messages.isEmpty()) {
            this.memoryType = ShortTermMemoryType.CONVERSATIONAL;
            this.infer = infer; // User can override
        } else if (data != null) {
            this.memoryType = ShortTermMemoryType.DATA;
            this.infer = false; // Always false for data
        }
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -87,7 +86,7 @@ public MemoryConfiguration(
         validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
 
         // Assign values after validation
-        this.indexPrefix = indexPrefix == null ? "memory-" + UUID.randomUUID() : indexPrefix;
+        this.indexPrefix = indexPrefix == null ? ".plugins-ml" : indexPrefix;
```
ylwu-amzn, line 89: How about use specific prefix 
```suggestion
        this.indexPrefix = indexPrefix == null ? ".plugins-ml-memory" : indexPrefix;
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -270,18 +269,19 @@ public static MemoryConfiguration parse(XContentParser parser) throws IOExceptio
     }
 
     public String getSessionIndexName() {
-        return indexPrefix + "-session";
+        return indexPrefix + "-agentic-memory-sessions";
     }
 
     public String getShortTermMemoryIndexName() {
-        return indexPrefix + "-short-term-memory";
+        return indexPrefix + "-agentic-memory-events";
```
ylwu-amzn: Events looks confusing, we are not storing events, we are storing short term memory. Create a new concept `event` here looks not necessary.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
 @Getter
 @Setter
 @Builder
-public class MLAddMemoriesInput implements ToXContentObject, Writeable {
+public class MLCreateEventInput implements ToXContentObject, Writeable {
 
     // Required fields
     private String memoryContainerId;
-    private ShortTermMemoryType memoryType;
+    private ShortTermMemoryType memoryType;  // Auto-detected from fields
```
ylwu-amzn:
```suggestion
    private ShortTermMemoryType memoryType;  // Auto-detected from fields if not provided
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
     private List<MessageInput> messages;
-    private String binaryData;
```
ylwu-amzn, line 55: I think we need binary data to support storing multi-modal data like image/video/pdf etc.
Storing such binary data in `data` is ok, but `data` will be stored as `flat_object` in index. That's not optimal comparing store it as binary type.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
         this.metadata = metadata;
-        this.tags = tags;
+
+        // Auto-detect memory type and set infer defaults
+        if (messages != null && !messages.isEmpty()) {
+            this.memoryType = ShortTermMemoryType.CONVERSATIONAL;
+            this.infer = infer; // User can override
+        } else if (data != null) {
+            this.memoryType = ShortTermMemoryType.DATA;
+            this.infer = false; // Always false for data
```
ylwu-amzn: What if both `messages` and `data` not null, and `memoryType` not provided by user? Should we throw exception to let user provide `memoryType` ?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
+        } else {
+            this.memoryType = memoryType; // Use provided type if neither messages nor data
+        }
+
         validate();
     }
 
     public void validate() {
+        // Check mutual exclusion
+        if (messages != null && !messages.isEmpty() && data != null) {
```
ylwu-amzn: User may want to store both the messages and state data as a checkpoint. Let's remove this limitation?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
+            throw new IllegalArgumentException("Cannot specify both 'messages' and 'data' fields in the same request");
+        }
+
+        // Check at least one is provided
+        if ((messages == null || messages.isEmpty()) && data == null) {
+            throw new IllegalArgumentException("Must specify either 'messages' or 'data' field");
+        }
+
+        // Validate infer field for data type
+        if (data != null && infer) {
```
ylwu-amzn, line 100:
```suggestion
        if (messages == null && infer) {
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
+            throw new IllegalArgumentException("infer=true is not supported for data memory type");
+        }
```
ylwu-amzn:
```suggestion
        if ((messages == null || messages.isEmpty()) && infer) {
            throw new IllegalArgumentException("Can not infer when messages are null.");
        }
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -33,76 +28,96 @@
...
+        // Validate infer with messages
         if (messages == null || messages.isEmpty()) {
             if (infer) {
                 throw new IllegalArgumentException("No messages provided when inferring memory");
             }
         }
 
-        if (infer && memoryType != ShortTermMemoryType.CONVERSATION) {
-            throw new IllegalArgumentException("Infer is only supported for conversation memory");
+        if (infer && memoryType != ShortTermMemoryType.CONVERSATIONAL) {
```
ylwu-amzn: Seems too strict. We could do infer if the messages not null or empty. So if user want to preload some memory, they could use DATA type and set `messages` data.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -261,51 +220,55 @@ public static MLAddMemoriesInput parse(XContentParser parser, String memoryConta
...
+        } else if (data != null) {
+            // Data type: always false, error if user tries to set to true
+            if (inferSpecified != null && inferSpecified) {
+                throw new IllegalArgumentException("infer=true is not supported for data memory type");
+            }
+            infer = false;
+        } else {
+            // Use provided value or false as default
+            infer = inferSpecified != null ? inferSpecified : false;
+        }
```
ylwu-amzn: Seems duplicate. We already do validation in constructor method, let's remove this validation ?

ylwu-amzn:
Have some high idea for the API naming

## Add memories
For add memories API, the payload could be messages which is short-term memories, but the API itself actually not just add it to short-term memory. It also extract long term memory.  So the existing API `POST _plugins/_ml/memory_containers/{{mem_container_id}}/memories` should be good enough to show this is to add some new memories (short-term and long term).

## Different types of memory APIs
### short-term/working memory
From AI:

> Many authors use “short-term memory” as a general term to describe what an agent holds in immediate context (recent messages, recent tool outputs, current session). It is simpler, more intuitive, and covers what people first think of when they talk about memory in chat agents. 

>But increasingly, people use “working memory” (or “short-term/working memory”) when they want to emphasize that the agent is not just storing context, but is manipulating, reasoning with it, holding intermediate results, sub-goals, internal states etc. For example, in agent memory taxonomies or architectures, “working memory” is used to refer to what the agent’s active “workspace” is.

> Working Memory = Short-Term Memory + the operations on it / structure for reasoning

Seems working memory is a more accurate word here as we are planning to store not only dialogue snippets but also agent state, tool invocation results etc.

We could use this API patten 
```
GET _plugins/_ml/memory_containers/{{mem_container_id}}/memories/{{memory_type}}/_search
```
For working memory , it will be
```
GET _plugins/_ml/memory_containers/{{mem_container_id}}/memories/working/_search
```
### Long term meory
```
GET _plugins/_ml/memory_containers/{{mem_container_id}}/memories/longterm/_search
```

dhrubo-os:
>Working Memory = Short-Term Memory + the operations on it / structure for reasoning

Another llm explanation:

```
Yes, you're right! The Short-term vs Long-term memory distinction is actually more intuitive and easier to understand than Working Memory, especially in terms of agentic memory. Here's why:

Short-term vs Long-term Memory:

Clear Time-based Distinction
Short-term: immediate, recent memories
Long-term: stored memories from past
Easy to understand the time difference
Everyday Examples
Short-term: remembering a phone number you just heard
Long-term: remembering your childhood home
People can easily relate to these examples
Simple Concept
Based on how long we remember things
Clear categories
Matches common experience
Natural Understanding
Matches how people think about memory
Reflects daily experiences
Makes intuitive sense
Working Memory:

More Complex
Involves multiple components
Includes processing and storage
More technical concept
Less Intuitive
Requires understanding of cognitive processes
More abstract
Harder to relate to daily life
More Academic
Used more in scientific contexts
Less common in everyday language
More specialized concept
You're correct that Short-term vs Long-term memory makes more sense in understanding agentic memory because it's:

More intuitive
Easier to understand
Based on common experience
Matches how people naturally think about memory
```


Bedrock also uses short term memory: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-memory-operations.html

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportCreateEventAction.java
```diff
@@ -188,41 +217,64 @@ private void processAndIndexMemory(
...
         IndexRequest indexRequest = new IndexRequest(shortTermMemoryIndex);
 
         try {
             XContentBuilder builder = XContentFactory.jsonBuilder();
-            // Add memory content from input object
-            mlAddMemoriesInput.toXContentWithTimeStamp(builder, ToXContent.EMPTY_PARAMS);
+            builder.startObject();
+
+            // Serialize event content (this duplicates MLCreateEventInput.toXContent() logic
+            // but we need to add timestamps, so we can't use toXContent directly)
```
ylwu-amzn: We could use the method `toXContentWithTimeStamp`, right?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
```diff
@@ -198,59 +199,17 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
...
             builder.field(METADATA_FIELD, metadata);
         }
-        if (tags != null && !tags.isEmpty()) {
-            builder.field(TAGS_FIELD, tags);
-        }
         builder.endObject();
         return builder;
     }
 
-    public XContentBuilder toXContentWithTimeStamp(XContentBuilder builder, ToXContent.Params params) throws IOException {
```
ylwu-amzn, line 208: This method was being used by `TransportAddMemoriesAction`. I think you removed this one to avoid duplicate code with method `toXContent`.

How about change this method to `public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params params, boolean withTimeStamp) throws IOException {`

Then change the `toXContent` method to 
```
    @Override
    public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params params) throws IOException {
        return toXContent(builder, params, false);
    }
```

We could use this in `TransportAddMemoriesAction`
```
mlAddMemoriesInput.toXContent(builder, ToXContent.EMPTY_PARAMS, true);
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportCreateEventAction.java
```diff
@@ -273,76 +326,84 @@ private Map<String, String> getStrategyNameSpace(MemoryStrategy strategy, Map<St
...
+                                decisions,
+                                memoryConfig,
+                                strategyNameSpace,
+                                user,
+                                input,
+                                memoryConfig,
+                                ActionListener.wrap(operationResults -> {
+                                    MLCreateEventResponse response = MLCreateEventResponse
+                                        .builder()
+                                        .eventId(null)
```
ylwu-amzn: Why have to set it as null explicitly ?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportDeleteEventAction.java
```diff
@@ -0,0 +1,113 @@
...
+            deleteEventFromIndex(eventId, container.getConfiguration(), actionListener);
+        }, actionListener::onFailure));
+    }
+
+    private void deleteEventFromIndex(String eventId, MemoryConfiguration configuration, ActionListener<DeleteResponse> actionListener) {
+        try {
+            String eventIndexName = configuration.getShortTermMemoryIndexName();
+            DeleteRequest deleteRequest = new DeleteRequest(eventIndexName, eventId);
+
+            client.delete(deleteRequest, ActionListener.wrap(deleteResponse -> {
```
ylwu-amzn: Since we store memory to system index by default if user not provide index prefix. We should check if the index is system index or not. If it's system index, we should check user's permission and stash context.

The memory should be user level. Memory container should support multiple users' memories. So only checking if user have memory container access is not enough. We should also record user id in memory index and add filter when search memory.

Add comment here, but the change should be applied to all memory APIs

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -149,20 +149,16 @@ private MLMemoryContainer buildMemoryContainer(MLCreateMemoryContainerInput inpu
...
         String userId = user != null ? user.getName() : "default";
         MemoryConfiguration configuration = container.getConfiguration();
         String indexPrefix = configuration != null ? configuration.getIndexPrefix() : null;
 
         // Convert to lowercase as OpenSearch doesn't support uppercase in index names
-        final String sessionIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-session";
-        final String shortTermMemoryIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-short-term-memory";
-        final String longTermMemoryIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-long-term-memory";
-        final String longTermMemoryHistoryIndexName = configuration.getLongMemoryHistoryIndexName();
+        final String sessionIndexName = configuration.getSessionIndexName().toLowerCase(Locale.ROOT);
```
ylwu-amzn: Let's move `toLowerCase` to `configuration.getSessionIndexName` to avoid duplicate this in multiple places?

A better way is converting `indexPrefix` to lowercase in `MemoryConfiguration` class constructor method. So just need to convert to lower case one time, not need to convert lower case for every getXXXIndexName request

b4sjoo:
Will choose working memory and long term memory as our names.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
b4sjoo: Will revert

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Will revert

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
b4sjoo: Will revert

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: I am fine with separate those

ylwu-amzn:
> > Working Memory = Short-Term Memory + the operations on it / structure for reasoning
> 
> Another llm explanation:
> 
> ```
> Yes, you're right! The Short-term vs Long-term memory distinction is actually more intuitive and easier to understand than Working Memory, especially in terms of agentic memory. Here's why:
> 
> Short-term vs Long-term Memory:
> 
> Clear Time-based Distinction
> Short-term: immediate, recent memories
> Long-term: stored memories from past
> Easy to understand the time difference
> Everyday Examples
> Short-term: remembering a phone number you just heard
> Long-term: remembering your childhood home
> People can easily relate to these examples
> Simple Concept
> Based on how long we remember things
> Clear categories
> Matches common experience
> Natural Understanding
> Matches how people think about memory
> Reflects daily experiences
> Makes intuitive sense
> Working Memory:
> 
> More Complex
> Involves multiple components
> Includes processing and storage
> More technical concept
> Less Intuitive
> Requires understanding of cognitive processes
> More abstract
> Harder to relate to daily life
> More Academic
> Used more in scientific contexts
> Less common in everyday language
> More specialized concept
> You're correct that Short-term vs Long-term memory makes more sense in understanding agentic memory because it's:
> 
> More intuitive
> Easier to understand
> Based on common experience
> Matches how people naturally think about memory
> ```
> 
> Bedrock also uses short term memory: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-memory-operations.html


Use different prompt, the AI will give different answer. Re-asked ChatGPT with this

```
Please read https://github.com/opensearch-project/ml-commons/pull/4188#issuecomment-3331355058 and https://github.com/opensearch-project/ml-commons/pull/4188#issuecomment-3332042950 And check which name is better in this PR : short-term or working Read this This tutorial doc which built earlier for a PoC purpose https://github.com/opensearch-project/ml-commons/blob/feature/agentic_memory_new/docs/tutorials/agentic_memory/agentic_memory_tutorial.md The key point is the short-term/working memory not only store the conversation messages, it also store agent state as checkpoint, other intemedeitate result etc
```
The answer is long, copy the final parts

>My recommendation

>I lean toward “working memory” (or at least a hybrid name like “short-term/working memory”) as the better name in this PR, because:

>Accuracy / expressiveness: You are storing more than just transient chat history — internal state, partial computation, checkpoints. “Working memory” captures that this is an active workspace.

>Avoid semantic mismatch: If you name it “short-term memory” but treat it as a workspace (with manipulation, state, etc.), readers might misunderstand or think it's just a buffer.

>You can still document equivalence: You can clarify in documentation that “working memory is your short-term memory + workspace for reasoning & state.”

>However, to preserve intuitiveness and avoid confusion, you could:

>Use “working memory” in internal API names / implementation, but in external doc/UI mention “working memory (short-term memory)” to bridge terminologies.

>Or choose a combined name like “short-term working memory” (or “short-term / working memory”) so it carries both senses.

>If forced to pick a single name, I’d choose “working memory” as primary and document that it corresponds to your short-term memory concept in the agentic memory taxonomy.

>So choose “working memory” as the canonical name (paired with “long-term memory”) for this PR, with clear documentation bridging to “short-term memory.” If later feedback from users shows confusion, you can always add context or aliasing.

I don't have strong opinions between these two names: shortterm vs working. We could explain the details in doc.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportDeleteEventAction.java
b4sjoo: Will change when implementing other under developing APIs

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportCreateEventAction.java
b4sjoo: Reverted

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Reverted first to unblock feature branch then will find a better solution when submitting PR to main

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportCreateEventAction.java
b4sjoo: Reverted first to unblock feature branch then will find a better solution when submitting PR to main

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Reverted

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: I agreed that user might have requirements in pre-load messages as data. But I still insist we don't need infer and type together.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Removed

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Reverted

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLCreateEventInput.java
b4sjoo: Reverted

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteWorkingMemoryAction.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package org.opensearch.ml.common.transport.memorycontainer.memory;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.action.delete.DeleteResponse;
+
+public class MLDeleteWorkingMemoryAction extends ActionType<DeleteResponse> {
+    public static final MLDeleteWorkingMemoryAction INSTANCE = new MLDeleteWorkingMemoryAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/memory_containers/memory/working/delete";
```
ylwu-amzn, line 13: Not urgent, we could discuss and fix in next PR. we may want to keep it consistent with the REST API 
```suggestion
    public static final String NAME = "cluster:admin/opensearch/ml/memory_containers/memories/working/delete";
```

b4sjoo: Check with connector or model, this place is singular.

ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteWorkingMemoryAction.java
b4sjoo: Ack

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
```diff
@@ -152,7 +152,7 @@ private void createNewSessionIfAbsent(
 
             if (!userProvidedSessionId && input.getMemoryType() == ShortTermMemoryType.CONVERSATION && !configuration.isDisableSession()) {
```
dhrubo-os, line 153: if we are treating short term memory as working memory, let's change `ShortTermMemoryType` to `WorkingMemoryType` to keep this consistent.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -85,11 +86,15 @@ public class MemoryContainerConstants {
     public static final String CREATE_MEMORY_CONTAINER_PATH = BASE_MEMORY_CONTAINERS_PATH + "/_create";
     public static final String PARAMETER_MEMORY_CONTAINER_ID = "memory_container_id";
     public static final String PARAMETER_MEMORY_ID = "memory_id";
+    public static final String PARAMETER_WORKING_MEMORY_ID = "working_memory_id";
```
dhrubo-os, line 89: [nit] we already have `public static final String WORKING_MEMORY_ID_FIELD = "working_memory_id";`

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -31,23 +30,25 @@ public class MemoryContainerConstants {
     public static final String LLM_ID_FIELD = "llm_id";
     public static final String MAX_INFER_SIZE_FIELD = "max_infer_size";
     public static final String STRATEGIES_FIELD = "strategies";
+    public static final String STRATEGY_FIELD = "strategy";
     public static final String INDEX_SETTINGS_FIELD = "index_settings";
+    public static final String ID_FIELD = "id";
+    public static final String ENABLED_FIELD = "enabled";
```
dhrubo-os, line 36: What is this field for? What are we enabling?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLGetWorkingMemoryRequest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(memoryContainerId);
+        out.writeString(workingMemoryId);
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
```
dhrubo-os, line 43: Can we follow this for the validation: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorRequest.java#L48-L54

We had a security concern before to allow any type of characters.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
```diff
@@ -60,18 +65,18 @@ public void writeTo(StreamOutput out) throws IOException {
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
         builder.startObject();
 
-        builder.field("id", id);
-        builder.field("enabled", enabled);
-        builder.field("type", type);
-        builder.field("namespace", namespace);
+        builder.field(ID_FIELD, id);
+        builder.field(ENABLED_FIELD, enabled);
+        builder.field(STRATEGY_FIELD, type);
```
ylwu-amzn, line 70: Ask ChatGPT
```
Which one do you prefer
"strategies": [
      {
        "enabled": true,
        "strategy": "SEMANTIC",
        "namespace": [
          "user_id"
        ],
        "configuration": {
          "test": "value"
        }
      }
    ],

"strategies": [
    {
        "enabled": true,
        "type": "SEMANTIC",
        "namespace": ["user_id"],
        "configuration": {
            "test": "value"
        }
    }
],
```

Response
```
I’d go with "type" rather than "strategy".

Reasoning:

1. strategies is already the array name, so using strategy again inside each object is a bit redundant.

2. type is a more natural way to describe the kind of strategy (e.g., SEMANTIC, LEXICAL, HYBRID).

3. It improves readability: strategies[].type immediately tells you it’s describing the classification of that strategy.
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -149,49 +149,67 @@ private MLMemoryContainer buildMemoryContainer(MLCreateMemoryContainerInput inpu
...
-    ) {
+    private void createMemoryDataIndices(MLMemoryContainer container, User user, ActionListener<String> listener) {
         String userId = user != null ? user.getName() : "default";
         MemoryConfiguration configuration = container.getConfiguration();
         String indexPrefix = configuration != null ? configuration.getIndexPrefix() : null;
 
         // Convert to lowercase as OpenSearch doesn't support uppercase in index names
         final String sessionIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-session";
-        final String shortTermMemoryIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-short-term-memory";
+        final String workingMemoryIndexName = indexPrefix.toLowerCase(Locale.ROOT) + "-working-memory";
```
ylwu-amzn, line 159: Change to `final String sessionIndexName = configuration.getSessionIndexName();` to keep consistent

Same comments for line 158, 160

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
b4sjoo: I would rather go to `strategy_type`

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
b4sjoo: Agree

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
ylwu-amzn: It's already in the `strategies` object, we dont' need to `strategy_` prefix. Keep it simple as `type` should be enough. Just like other fields , we don't name as `strategy_enabled` , just `enabled`.

Issue #4186, opened 2025-09-23
[BUG]NPE when run flow agent with MLAgentTool


Hailong-am:
**What is the bug?**

```
[2025-09-23T16:27:11,652][ERROR][o.o.m.e.a.a.MLAgentExecutor] [node-2] Failed to run flow agent
org.opensearch.transport.SendRequestTransportException: [node-1][127.0.0.1:9301][cluster:admin/opensearch/ml/predict]
	at org.opensearch.transport.TransportService.sendRequestInternal(TransportService.java:1083) ~[opensearch-3.2.0.jar:3.2.0]
	at org.opensearch.transport.TransportService.sendRequestAsync(TransportService.java:1847) ~[opensearch-3.2.0.jar:3.2.0]
	at org.opensearch.transport.TransportService.sendRequest(TransportService.java:978) ~[opensearch-3.2.0.jar:3.2.0]
	at org.opensearch.transport.TransportService.sendRequest(TransportService.java:937) ~[opensearch-3.2.0.jar:3.2.0]
	at org.opensearch.ml.task.MLPredictTaskRunner.lambda$dispatchTask$1(MLPredictTaskRunner.java:179) ~[?:?]
	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) ~[opensearch-core-3.2.0.jar:3.2.0]
	at org.opensearch.ml.task.MLTaskDispatcher.lambda$dispatchPredictTask$2(MLTaskDispatcher.java:84) ~[?:?]
	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) ~[opensearch-core-3.2.0.jar:3.2.0]
	at org.opensearch.ml.task.MLTaskDispatcher.dispatchTaskWithRoundRobin(MLTaskDispatcher.java:99) ~[?:?]
	at org.opensearch.ml.task.MLTaskDispatcher.dispatchPredictTask(MLTaskDispatcher.java:82) ~[?:?]
	at org.opensearch.ml.task.MLPredictTaskRunner.dispatchTask(MLPredictTaskRunner.java:243) ~[?:?]
	at org.opensearch.ml.task.MLPredictTaskRunner.dispatchTask(MLPredictTaskRunner.java:90) ~[?:?]
	at org.opensearch.ml.task.MLTaskRunner.run(MLTaskRunner.java:93) ~[?:?]
[...]

PR #4053, opened 2025-08-04
Support output filter, unify tool parameter handling and improve Sear…

ylwu-amzn:
…chIndexTool output parsing

### Description

Fix issues when build [Agentic RAG with automatic index selection tutorial](https://github.com/ylwu-amzn/ml-commons/blob/agent-rag-tutorial/docs/tutorials/agent_framework/Agentic_RAG.md#approach-2-agentic-rag-with-automatic-index-selection)

1. In the tutorial, the Chat Agent uses AgentTool to call another Flow Agent. The flow agent generates OpenSearch query and search index. The Flow Agent returns extra info, but we only need the query response part. So we need tool output filter.

3. The Flow Agent will pass the generated OpenSearch query to SearchIndexTool to search data. But the escape is wrong. The query is already valid json. If we run `escapeJson(outputResponse)` again, the query json string become invalid. 

4. In this tutorial, we use one LLM which has multiple placeholders like `_chat_history`, `tool_configs`. The root Chat Agent is Chat Agent, which has chat history and tool configs in parameters. When Chat Agent run the Flow Agent through a tool call, it will also pass these parameters to Flow Agent. Since Flow Agent use same LLM, these parameter will cause wrong result in Flow Agent. So we need to support required parameter on tool level to filter out only the required parameters.

5. SearchIndexTool always returns the search hits part. But for an aggregation query, the search hits part is empty. Add a new parameter to return the full response. 
For example
```
GET opensearch_dashboards_sample_data_flights/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "OriginCountry": "CA"
          }
        },
        {
          "term": {
            "DestCountry": "US"
          }
        }
      ]
    }
  },
  "size": 0,
  "aggs": {
    "flight_count": {
      "value_count": {
        "field": "FlightNum"
      }
    }
  }
}
```
Response
```
{
  "took": 14,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 94,
      "relation": "eq"
    },
    "max_score": null,
    "hits": []
  },
  "aggregations": {
    "flight_count": {
      "value": 94
    }
  }
}
```

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
[...]

PR #4189, opened 2025-09-23
Fix NPE when execute flow agent with mutli tenancy is off

Hailong-am:
### Description

Fix the NPE when execute flow agent, the null value comes from `tenant_id` that will cause NPE when do serialization.

https://github.com/opensearch-project/ml-commons/blob/247a88c54a7e07ee777b455f4974205ddb9ea9ee/common/src/main/java/org/opensearch/ml/common/dataset/remote/RemoteInferenceInputDataSet.java#L79-L81

We can fix it with another way for class `RemoteInferenceInputDataSet`, but this is may have BWC issue when two node have different version.

```diff
- streamOutput.writeMap(parameters, StreamOutput::writeString, StreamOutput::writeString);
+ streamOutput.writeMap(parameters, StreamOutput::writeString, StreamOutput::writeOptionalString);

- parameters = streamInput.readMap(StreamInput::readString, StreamInput::readString);
+ parameters = streamInput.readMap(StreamInput::readString, StreamInput::readOptionalString);
```

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4186

https://github.com/opensearch-project/ml-commons/pull/4053

### Check List


common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -195,6 +195,8 @@ public static Map<String, String> filteredParameterMap(Map<String, ?> parameterO
     @SuppressWarnings("removal")
     public static Map<String, String> getParameterMap(Map<String, ?> parameterObjs) {
         Map<String, String> parameters = new HashMap<>();
+        if (parameterObjs == null)
+            return parameters;
```
akolarkunnu, line 199: Instead of keeping two return statements in single method, I would suggest:
if (parameterObjs != null) {
   for loop and remaining logic
}
return parameters;

It will handle this use case.

Hailong-am:
> @Hailong-am this may be error prone as the original issue of parsing still exists. If another code flow passes null, it will fail. I recommend fixing using the alternate solution mentioned in the description as that solves the root cause. We can use version checks for BWC.
> 
> What do you think?

@pyek-bot Yes, i was thinking BWC maybe tricky for existing field as i misunderstand the version of StreamInput and StreamOutput. The version default is the node version, but for remote transport communication, its value comes from remote node. For example two nodes with version 3.2 and 3.3, the version should be 3.2. so this will not break existing logic. That make sense we can have version check for this. 

Another thought i may have is remove the null value entries before perform serialization?

Hailong-am:
Change it to way to use `writeOptionalString` and `readOptionalString` with version check. @pyek-bot please help to take a look again.

Hailong-am:
> Changes look good, thanks @Hailong-am ! Please make sure you have tested the BWC change by upgrading

Sure, I can setup a cluster with two nodes one is 3.2 another is 3.3 to test it out.

Hailong-am:
> > Changes look good, thanks @Hailong-am ! Please make sure you have tested the BWC change by upgrading
> 
> Sure, I can setup a cluster with two nodes one is 3.2 another is 3.3 to test it out.

@pyek-bot @dhrubo-os I have tested a cluster with version 3.2 and 3.3 mixed nodes, agent execute api will always go through 3.2 logic branch. So it will not break existing logic. But the NPE still exists for mixed version cluster, that's expected.

Hailong-am:
> @Hailong-am can you look at the code coverage? should be a simple test

Unit test added @pyek-bot 

jar hell issue will handled by https://github.com/opensearch-project/ml-commons/pull/4214

dhrubo-os (approved):

PR #4212, opened 2025-09-26
[FEATURE] Agent Execute Stream

nathaliellenaa:
### Description
Add a new agent execute stream API as an experimental feature to support agent streaming.

Supported agent - model for this agent execute stream API:
- [ ] Conversational agent - OpenAI Chat Completion model
- [ ] Conversational agent - Bedrock Converse Stream model

> Note: This PR depends on the predict stream implementation https://github.com/opensearch-project/ml-commons/pull/4187

**API Endpoint:**
```
POST /_plugins/_ml/agents/{agent_id}/_execute/stream
```

**Sample workflow:**
```
// Enable feature flag
PUT /_cluster/settings
{
    "persistent": {
        "plugins.ml_commons.stream_enabled": true
    }
}

// Register OpenAI chat completion model
{
    "name": "openai gpt 3.5 turbo",
    "function_name": "remote",
    "description": "openai model",
    "connector": {
        "name": "OpenAI Chat Connector",
        "description": "The connector to public OpenAI model service for GPT 3.5",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "endpoint": "api.openai.com",
            "model": "gpt-3.5-turbo"
        },
        "credential": {
            "openAI_key": "<your_api_key>"
        },
        "actions": [{
            "action_type": "predict",
            "method": "POST",
            "headers": {
            },
            "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": [{\"role\":\"developer\",\"content\":\"${parameters.system_prompt}\"},${parameters._chat_history:-}{\"role\":\"user\",\"content\":\"${parameters.prompt}\"}${parameters._interactions:-}]${parameters.tool_configs:-} }"
        }]
    }
}

// Register conversational agent using OpenAI model
{
    "name": "Chat Agent with RAG",
    "type": "conversational",
    "llm": {
        "model_id": "<model id created in previous step>",
        "parameters": {
            "max_iteration": 5,
            "system_prompt": "You are a helpful assistant. You are able to assist with a wide range of tasks, from answering simple questions to providing in-depth explanations and discussions on a wide range of topics.\nIf the question is complex, you will split it into several smaller questions, and solve them one by one. For example, the original question is:\nhow many orders in last three month? Which month has highest?\nYou will spit into several smaller questions:\n1.Calculate total orders of last three month.\n2.Calculate monthly total order of last three month and calculate which months order is highest. You MUST use the available tools everytime to answer the question",
            "prompt": "${parameters.question}"
        }
    },
    "memory": {
        "type": "conversation_index"
    },
    "parameters": {
        "_llm_interface": "openai/v1/chat/completions"
    },
    "tools": [
        {
            "type": "IndexMappingTool",
            "name": "DemoIndexMappingTool",
            "parameters": {
                "index": "${parameters.index}",
                "input": "${parameters.question}"
            }
        },
        {
            "type": "ListIndexTool",
            "name": "RetrieveIndexMetaTool",
            "description": "Use this tool to get OpenSearch index information: (health, status, index, uuid, primary count, replica count, docs.count, docs.deleted, store.size, primary.store.size)."
        }
    ],
    "app_type": "chat_with_rag"
}

// Run agent execute stream API
POST /_plugins/_ml/agents/evCIh5kB66VN-aC_0aNf/_execute/stream
{
    "parameters": {
        "question": "How many indices are in my cluster?"
    }
}

// Sample response
data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"[{\"index\":0.0,\"id\":\"call_HjpbrbdQFHK0omPYa6m2DCot\",\"type\":\"function\",\"function\":{\"name\":\"RetrieveIndexMetaTool\",\"arguments\":\"\"}}]","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"[{\"index\":0.0,\"function\":{\"arguments\":\"{}\"}}]","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"{\"choices\":[{\"message\":{\"tool_calls\":[{\"type\":\"function\",\"function\":{\"name\":\"RetrieveIndexMetaTool\",\"arguments\":\"{}\"},\"id\":\"call_HjpbrbdQFHK0omPYa6m2DCot\"}]},\"finish_reason\":\"tool_calls\"}]}","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"row,health,status,index,uuid,pri(number of primary shards),rep(number of replica shards),docs.count(number of available documents),docs.deleted(number of deleted documents),store.size(store size of primary and replica shards),pri.store.size(store size of primary shards)\n1,green,open,.plugins-ml-model-group,Msb1Y4W5QeiLs5yUQi-VRg,1,1,2,0,17.1kb,5.9kb\n2,green,open,.plugins-ml-memory-message,1IWd1HPeSWmM29qE6rcj_A,1,1,658,0,636.4kb,313.5kb\n3,green,open,.plugins-ml-memory-meta,OETb21fqQJa3Y2hGQbknCQ,1,1,267,7,188kb,93.9kb\n4,green,open,.plugins-ml-config,0mnOWX5gSX2s-yP27zPFNw,1,1,1,0,8.1kb,4kb\n5,green,open,.plugins-ml-model,evYOOKN4QPqtmUjxsDwJYA,1,1,5,5,421.5kb,210.7kb\n6,green,open,.plugins-ml-agent,I0SpBovjT3C6NABCBzGiiQ,1,1,6,0,205.5kb,111.3kb\n7,green,open,.plugins-ml-task,_Urzn9gdSuCRqUaYAFaD_Q,1,1,100,4,136.1kb,45.3kb\n8,green,open,top_queries-2025.09.26-00444,jb7Q1FiLSl-wTxjdSUKs_w,1,1,1736,126,1.8mb,988kb\n9,green,open,.plugins-ml-connector,YaJORo4jT0Ksp24L5cW1uA,1,1,2,0,97.8kb,48.9kb\n","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"There","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" are","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" ","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"9","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" indices","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" in","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" your","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":" cluster","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":".","is_last":false}}]}]}

data: {"inference_results":[{"output":[{"name":"memory_id","result":"LvU1iJkBCzHrriq5hXbN"},{"name":"parent_interaction_id","result":"L_U1iJkBCzHrriq5hXbs"},{"name":"response","dataAsMap":{"content":"","is_last":true}}]}]}
```

**Error handling:**
[In progress]

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/3630

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -26,6 +26,7 @@ import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.commons.text.StringSubstitutor;
 import org.opensearch.Version;
@@ -56,6 +57,7 @@ public class HttpConnector extends AbstractConnector {
     public static final String PARAMETERS_FIELD = "parameters";
     public static final String SERVICE_NAME_FIELD = "service_name";
     public static final String REGION_FIELD = "region";
+    public static final String LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS = "openai/v1/chat/completions";
 
     // TODO: add RequestConfig like request time out,
 
@@ -377,14 +379,14 @@ public class HttpConnector extends AbstractConnector {
         }
 
         String llmInterface = parameters.get("_llm_interface");
-        if (llmInterface.isBlank()) {
+        if (StringUtils.isBlank(llmInterface)) {
             return false;
         }
 
         llmInterface = llmInterface.trim().toLowerCase(Locale.ROOT);
         llmInterface = StringEscapeUtils.unescapeJava(llmInterface);
         switch (llmInterface) {
-            case "openai/v1/chat/completions":
+            case LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS:
                 return true;
             default:
                 return false;
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -57,6 +57,7 @@ public class HttpConnector extends AbstractConnector {
     public static final String PARAMETERS_FIELD = "parameters";
     public static final String SERVICE_NAME_FIELD = "service_name";
     public static final String REGION_FIELD = "region";
+    // TODO: move the AgentUtils class from algorithm module to common module
     public static final String LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS = "openai/v1/chat/completions";
 
     // TODO: add RequestConfig like request time out,
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
@@ -14,10 +14,19 @@ import org.opensearch.transport.TransportChannel;
 public interface Executable {
 
     /**
-     * Execute algorithm with given input data.
+     * Execute algorithm with given input data (non-streaming).
+     * @param input input data
+     * @param listener action listener
+     */
+    default void execute(Input input, ActionListener<Output> listener) throws ExecuteException {
+        execute(input, listener, null);
+    }
+
+    /**
+     * Execute algorithm with given input data (streaming).
      * @param input input data
      * @param listener action listener
      * @param channel transport channel
      */
-    void execute(Input input, ActionListener<Output> listener, TransportChannel channel) throws ExecuteException;
+    default void execute(Input input, ActionListener<Output> listener, TransportChannel channel) {}
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
@@ -9,12 +9,15 @@ import org.opensearch.core.action.ActionListener;
 import org.opensearch.ml.common.exception.ExecuteException;
 import org.opensearch.ml.common.input.Input;
 import org.opensearch.ml.common.output.Output;
+import org.opensearch.transport.TransportChannel;
 
 public interface Executable {
 
     /**
      * Execute algorithm with given input data.
      * @param input input data
+     * @param listener action listener
+     * @param channel transport channel
      */
-    void execute(Input input, ActionListener<Output> listener) throws ExecuteException;
+    void execute(Input input, ActionListener<Output> listener, TransportChannel channel) throws ExecuteException;
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentRunner.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentRunner.java
@@ -17,7 +17,17 @@ import org.opensearch.transport.TransportChannel;
 public interface MLAgentRunner {
 
     /**
-     * Function interface to execute agent.
+     * Function interface to execute agent (non-streaming)
+     * @param mlAgent
+     * @param params
+     * @param listener
+     */
+    default void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Object> listener) {
+        run(mlAgent, params, listener, null);
+    }
+
+    /**
+     * Function interface to execute agent (streaming)
      * @param mlAgent
      * @param params
      * @param listener
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentRunner.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentRunner.java
@@ -9,6 +9,7 @@ import java.util.Map;
 
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.ml.common.agent.MLAgent;
+import org.opensearch.transport.TransportChannel;
 
 /**
  * Agent executor interface definition. Agent executor will be used by {@link MLAgentExecutor} to invoke agents.
@@ -20,6 +21,7 @@ public interface MLAgentRunner {
      * @param mlAgent
      * @param params
      * @param listener
+     * @param channel
      */
-    void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Object> listener);
+    void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Object> listener, TransportChannel channel);
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
@@ -38,6 +38,7 @@ import static org.opensearch.ml.engine.tools.ReadFromScratchPadTool.SCRATCHPAD_N
 
 import java.security.PrivilegedActionException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -71,6 +72,7 @@ import org.opensearch.ml.common.spi.memory.Memory;
 import org.opensearch.ml.common.spi.memory.Message;
 import org.opensearch.ml.common.spi.tools.Tool;
 import org.opensearch.ml.common.transport.MLTaskResponse;
+import org.opensearch.ml.common.transport.prediction.MLPredictionStreamTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskRequest;
 import org.opensearch.ml.common.utils.StringUtils;
@@ -84,6 +86,7 @@ import org.opensearch.ml.engine.tools.MLModelTool;
 import org.opensearch.ml.repackage.com.google.common.collect.ImmutableMap;
 import org.opensearch.ml.repackage.com.google.common.collect.Lists;
 import org.opensearch.remote.metadata.client.SdkClient;
+import org.opensearch.transport.TransportChannel;
 import org.opensearch.transport.client.Client;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -160,7 +163,7 @@ public class MLChatAgentRunner implements MLAgentRunner {
     }
 
     @Override
-    public void run(MLAgent mlAgent, Map<String, String> inputParams, ActionListener<Object> listener) {
+    public void run(MLAgent mlAgent, Map<String, String> inputParams, ActionListener<Object> listener, TransportChannel channel) {
         Map<String, String> params = new HashMap<>();
         if (mlAgent.getParameters() != null) {
             params.putAll(mlAgent.getParameters());
@@ -246,7 +249,7 @@ public class MLChatAgentRunner implements MLAgentRunner {
                     }
                 }
 
-                runAgent(mlAgent, params, listener, memory, memory.getConversationId(), functionCalling);
+                runAgent(mlAgent, params, listener, memory, memory.getConversationId(), functionCalling, channel);
             }, e -> {
                 log.error("Failed to get chat history", e);
                 listener.onFailure(e);
@@ -260,7 +263,8 @@ public class MLChatAgentRunner implements MLAgentRunner {
         ActionListener<Object> listener,
         Memory memory,
         String sessionId,
-        FunctionCalling functionCalling
+        FunctionCalling functionCalling,
+        TransportChannel channel
     ) {
         List<MLToolSpec> toolSpecs = getMlToolSpecs(mlAgent, params);
 
@@ -269,7 +273,23 @@ public class MLChatAgentRunner implements MLAgentRunner {
             Map<String, Tool> tools = new HashMap<>();
             Map<String, MLToolSpec> toolSpecMap = new HashMap<>();
             createTools(toolFactories, params, allToolSpecs, tools, toolSpecMap, mlAgent);
-            runReAct(mlAgent.getLlm(), tools, toolSpecMap, params, memory, sessionId, mlAgent.getTenantId(), listener, functionCalling);
+            if (channel != null) {
+                runReActStream(
+                    mlAgent.getLlm(),
+                    mlAgent,
+                    tools,
+                    toolSpecMap,
+                    params,
+                    memory,
+                    sessionId,
+                    mlAgent.getTenantId(),
+                    listener,
+                    functionCalling,
+                    channel
+                );
+            } else {
+                runReAct(mlAgent.getLlm(), tools, toolSpecMap, params, memory, sessionId, mlAgent.getTenantId(), listener, functionCalling);
+            }
         };
 
         // Fetch MCP tools and handle both success and failure cases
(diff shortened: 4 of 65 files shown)


dhrubo-os:
resolve the conflict. Rebase from main?

nathaliellenaa:
Will push another commit to address comments from predict stream PR. Just cleaned up this PR to only include commit related to agent stream

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -71,17 +71,7 @@ default void executeAction(String action, MLInput mlInput, ActionListener<MLTask
...
-                public void onResponse(Tuple<Integer, ModelTensors> response) {
-                    actionListener.onResponse(new MLTaskResponse(new ModelTensorOutput(Arrays.asList(response.v2()))));
-                }
-
-                @Override
-                public void onFailure(Exception e) {
-                    actionListener.onFailure(e);
-                }
-            }, channel);
+            preparePayloadAndInvoke(action, mlInput, new ExecutionContext(0), null, actionListener, channel);
```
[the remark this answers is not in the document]
nathaliellenaa: Yes, it's intentional. We need to pass `ActionListener<MLTaskResponse> actionListener` for stream

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/StreamPredictActionListener.java
```diff
@@ -72,6 +99,37 @@ public void onFailure(Exception e) {
...
+
+                updatedTensors.add(ModelTensor.builder().name("memory_id").result(memoryId).build());
+                updatedTensors.add(ModelTensor.builder().name("parent_interaction_id").result(parentInteractionId).build());
+
+                updatedTensors.addAll(tensors.getMlModelTensors());
+                updatedOutputs.add(ModelTensors.builder().mlModelTensors(updatedTensors).build());
+            }
+
+            ModelTensorOutput updatedOutput = ModelTensorOutput.builder().mlModelOutputs(updatedOutputs).build();
+            return (Response) new MLTaskResponse(updatedOutput);
```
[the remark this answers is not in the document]
nathaliellenaa: Yes, right now streaming will only be supported for conversational agent. Will need to make a lot of modification to handle other agents.

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -364,17 +365,9 @@ protected void executeTask(MLPredictionTaskRequest request, ActionListener<MLTas
...
-        try {
-            if (request.getMlInput() != null && request.getMlInput().getInputDataset() instanceof RemoteInferenceInputDataSet) {
-                RemoteInferenceInputDataSet inputDataSet = (RemoteInferenceInputDataSet) request.getMlInput().getInputDataset();
-                Map<String, String> parameters = inputDataSet.getParameters();
-                return parameters != null && "true".equals(parameters.get("stream"));
-            }
-        } catch (Exception e) {
-            log.debug("Failed to check streaming parameter, defaulting to non-streaming", e);
-        }
-        return false;
```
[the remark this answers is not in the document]
nathaliellenaa: This comes from the predict stream PR https://github.com/opensearch-project/ml-commons/pull/4187, it added the new stream parameter

plugin/src/main/java/org/opensearch/ml/rest/RestMLExecuteStreamAction.java
```diff
@@ -0,0 +1,342 @@
...
+                        }
+                        if (dataMap.containsKey("is_last")) {
+                            isLast = Boolean.TRUE.equals(dataMap.get("is_last"));
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            log.error("Failed to extract values from response", e);
+        }
```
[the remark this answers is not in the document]
nathaliellenaa: Yes, need to make more modifications to make the implementation more general and can be used by other agents

nathaliellenaa:
Addressed all comments from https://github.com/opensearch-project/ml-commons/pull/4187

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/StreamPredictActionListener.java
```diff
@@ -72,6 +99,37 @@ public void onFailure(Exception e) {
...
+    private Response addMetadataToResponse(Response response) {
+        if (!(response instanceof MLTaskResponse)) {
+            return response;
+        }
+
+        // Only add metadata for agent streaming
+        if (memoryId == null && parentInteractionId == null) {
+            return response;
+        }
+
```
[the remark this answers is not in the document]
nathaliellenaa: Right, will change it to 
```
if (agentListener == null) {
            return response;
}
```

plugin/src/main/java/org/opensearch/ml/action/execute/TransportExecuteStreamTaskAction.java
```diff
@@ -0,0 +1,84 @@
...
+        doExecute(task, request, streamListener, channel);
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLExecuteTaskResponse> listener) {
+        // This should never be called for streaming action
+        listener.onFailure(new UnsupportedOperationException("Use doExecute with TransportChannel for streaming requests"));
+    }
+
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLExecuteTaskResponse> listener, TransportChannel channel) {
```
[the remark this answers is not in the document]
nathaliellenaa: No, we don't need one. It's a new method, there is no parent method to override

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -552,6 +572,362 @@ private void runReAct(
...
+        Map<String, MLToolSpec> toolSpecMap,
+        Map<String, String> parameters,
+        Memory memory,
+        String sessionId,
+        String tenantId,
+        ActionListener<Object> listener,
+        FunctionCalling functionCalling,
+        TransportChannel channel
+    ) {
+        Map<String, String> tmpParameters = constructLLMParams(llm, parameters);
```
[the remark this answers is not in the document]
nathaliellenaa: That makes sense, main difference between runReAct and runReActStream:

- Final answer handling:
Regular react: uses sendFinalAnswer() to send the complete response at once
Streaming react: uses saveToMemory() to create and send final response as chunk
- Tool response and LLM request processing:
Regular react: uses MLPredictionTaskAction for LLM requests
Streaming react: sends tool responses as individual chunks and uses MLPredictionStreamTaskAction for LLM request. 

I think this should be doable, will try to combine this into runReAct method and handle streaming by checking for channel

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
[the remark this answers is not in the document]
nathaliellenaa: Sure, will address in future PR

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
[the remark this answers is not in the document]
nathaliellenaa: Yes, I refactored it in the latest commit

plugin/src/main/java/org/opensearch/ml/action/execute/TransportExecuteStreamTaskAction.java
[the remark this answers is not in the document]
nathaliellenaa: The parent method should be defined in TransportAction class in the OpenSearch core. I don't think we need to define a parent method since it's a method overloading?

plugin/src/main/java/org/opensearch/ml/rest/RestMLExecuteStreamAction.java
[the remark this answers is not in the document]
nathaliellenaa: Added more todos

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -56,6 +57,7 @@ public class HttpConnector extends AbstractConnector {
     public static final String PARAMETERS_FIELD = "parameters";
     public static final String SERVICE_NAME_FIELD = "service_name";
     public static final String REGION_FIELD = "region";
+    public static final String LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS = "openai/v1/chat/completions";
```
mingshl, line 61: nit, there is predefined variables. then try to import [ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java](https://github.com/opensearch-project/ml-commons/blob/5883f540ccec99208520b4a1f3914591a9a1b165/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java#L118) 

So in case we changed any variables, we don't have any left overs

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
```diff
@@ -117,6 +118,19 @@ public void dispatchTask(
...
+                            node,
+                            getTransportStreamActionName(),
+                            request,
+                            TransportRequestOptions.builder().withType(TransportRequestOptions.Type.STREAM).build(),
+                            getResponseStreamHandler(request)
+                        );
+                } else {
+                    transportService.sendRequest(node, getTransportActionName(), request, getResponseHandler(listener));
+                }
                 transportService.sendRequest(node, getTransportActionName(), request, getResponseHandler(listener));
```
mingshl: is this sendRequest duplicated in this logic? it looks like jumping to else condition , it will call two times

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -357,23 +389,37 @@ private void runReAct(
...
-                            parentInteractionId,
-                            verbose,
-                            traceDisabled,
-                            traceTensors,
-                            conversationIndexMemory,
-                            traceNumber,
-                            additionalInfo,
-                            finalAnswer
-                        );
+                        if (isStreaming) {
```
mingshl: there's significant duplication between streaming and non-streaming logic. The runReAct method has become very complex with conditional 
streaming logic throughout. This creates maintenance burden and increases the risk 
of bugs.

Instead of having conditional logic throughout runReAct, consider:
• Extract common functionality into helper methods
• Use strategy pattern for streaming vs non-streaming behavior
• Create a streaming wrapper that handles the differences

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/sample/LocalSampleCalculator.java
```diff
@@ -37,7 +38,7 @@ public LocalSampleCalculator(Client client, Settings settings) {
     }
 
     @Override
-    public void execute(Input input, ActionListener<Output> listener) {
+    public void execute(Input input, ActionListener<Output> listener, TransportChannel channel) {
```
mingshl: these are public methods, can you make the channel optional?

Consider making the channel parameter optional or using method overloading:
java
default void execute(Input input, ActionListener<Output> listener) {
    execute(input, listener, null);
}
void execute(Input input, ActionListener<Output> listener, TransportChannel channel);

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/tool/MLToolExecutor.java
```diff
@@ -63,7 +64,7 @@ public MLToolExecutor(
     }
 
     @Override
-    public void execute(Input input, ActionListener<Output> listener) {
+    public void execute(Input input, ActionListener<Output> listener, TransportChannel channel) {
         if (!(input instanceof ToolMLInput)) {
```
mingshl, line 67: same comments for all execute(), for the public methods, 
Consider making the channel parameter optional or using method overloading:
java
default void execute(Input input, ActionListener<Output> listener) {
    execute(input, listener, null);
}
void execute(Input input, ActionListener<Output> listener, TransportChannel channel);

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -160,73 +149,19 @@ public void invokeRemoteServiceStream(
...
-                    default:
-                        // Ignore the other event types for now.
-                        break;
-                }
-            }).build();
-            if (bedrockRuntimeAsyncClient == null) {
-                bedrockRuntimeAsyncClient = buildBedrockRuntimeAsyncClient(httpClient);
-            }
-            bedrockRuntimeAsyncClient.converseStream(request, handler);
+            StreamingHandler handler = StreamingHandlerFactory.createHandler(llmInterface, connector, httpClient, executionContext, null);
```
mingshl: this is a good pattern

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
nathaliellenaa: I don't think we can import from AgentUtils, since they're in different module. Got an error when trying to import from AgentUtils `Cannot resolve symbol 'AgentUtils'`

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
nathaliellenaa: Nice catch, will delete one

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
dhrubo-os: Yeah, Algorithm module has dependency on common module. Not the other way. One solution could be, we can move the AgentUtils class from algorithm module to common module. But we can refactor that later. Not urgent.

ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
```diff
@@ -9,12 +9,15 @@
...
 public interface Executable {
 
     /**
      * Execute algorithm with given input data.
      * @param input input data
+     * @param listener action listener
+     * @param channel transport channel
      */
-    void execute(Input input, ActionListener<Output> listener) throws ExecuteException;
+    void execute(Input input, ActionListener<Output> listener, TransportChannel channel) throws ExecuteException;
```
dhrubo-os: do we need this transport channel for every regular execution or only for streaming?

dhrubo-os: can't we use method overloading?

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
```diff
@@ -128,6 +142,14 @@ public void dispatchTask(
 
     protected abstract TransportResponseHandler<Response> getResponseHandler(ActionListener<Response> listener);
 
+    protected TransportResponseHandler<Response> getResponseStreamHandler(Request request) {
+        throw new UnsupportedOperationException("Streaming is not supported.");
+    }
+
+    protected boolean isStreamingRequest(Request request) {
```
dhrubo-os, line 148: Is that expected?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
dhrubo-os: Let's do this. Otherwise it won't be maintainable in future.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+            );
+
+        ModelTensor tensor = ModelTensor.builder().name("response").dataAsMap(wrappedResponse).build();
+        ModelTensors tensors = ModelTensors.builder().mlModelTensors(List.of(tensor)).build();
+        ModelTensorOutput output = ModelTensorOutput.builder().mlModelOutputs(List.of(tensors)).build();
+        return new MLTaskResponse(output);
+    }
+
+    private String[] extractToolInfoFromContentBlockStart(ConverseStreamOutput event) {
+        String eventString = event.toString();
```
dhrubo-os: this seems too brittle. What if they change their pattern and this string manipulation will break. I prefer:

```
case CONTENT_BLOCK_START:
    if (event.contentBlockStart() != null) {
        ContentBlockStartEvent startEvent = event.contentBlockStart();
        // Now we can access proper SDK methods on startEvent
        // instead of string parsing
    }
    break;
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+        StreamPredictActionListener<MLTaskResponse, ?> listener,
+        AtomicBoolean isStreamClosed,
+        AtomicBoolean functionCallInProgress,
+        AtomicBoolean agentExecutionInProgress,
+        AtomicReference<String> toolName,
+        AtomicReference<Map<String, Object>> toolInput,
+        AtomicReference<String> toolUseId,
+        StringBuilder toolInputAccumulator
+    ) {
+        switch (event.sdkEventType()) {
```
dhrubo-os: Can we do something like this:

```
private volatile StreamState currentState = StreamState.STREAMING_CONTENT;

private void handleStateTransition(StreamEvent event) {
    switch (currentState) {
        case STREAMING_CONTENT:
            if (isToolUseDetected(event)) {
                currentState = StreamState.TOOL_CALL_DETECTED;
            } else if (isStreamComplete(event)) {
                currentState = StreamState.COMPLETED;
            }
            break;
            
        case TOOL_CALL_DETECTED:
            if (toolResponseReady()) {
                sendToolResponse();
                currentState = StreamState.WAITING_FOR_TOOL_RESULT;
            }
            break;
            
        case WAITING_FOR_TOOL_RESULT:
            // Don't close stream - wait for tool execution
            break;
            
        case COMPLETED:
            closeStream();
            break;
    }
}
```

Just an example of state management approach.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+        log.debug("Extracted tool - name: {}, id: {}", toolName, toolUseId);
+        return new String[] { toolName, toolUseId };
+    }
+
+    private String extractWithPattern(String text, Pattern pattern) {
+        Matcher matcher = pattern.matcher(text);
+        return matcher.find() ? matcher.group(1).trim() : null;
+    }
+
+    private void accumulateToolInput(
```
dhrubo-os, line 307:
```
Example of Failure:
json
Fragment 1: {"tool": "search", "params": {"query": "hello"
Fragment 2: , "limit": 10}}

After fragment 1: starts with {, doesn't end with } → won't parse (good)
After fragment 2: starts with {, ends with } → tries to parse → success (good)

But what about:
json
Fragment 1: {"tool": "search", "params": {"nested": "}"
Fragment 2: }}

After fragment 1: starts with {, ends with " → won't parse (good)
After fragment 2: starts with {, ends with } → tries to parse → could succeed with malformed JSON
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+            .build();
+    }
+
+    private Document buildDocumentFromJsonNode(JsonNode node) {
+        if (node.isObject()) {
+            Map<String, Document> map = new HashMap<>();
+            node.fields().forEachRemaining(entry -> map.put(entry.getKey(), buildDocumentFromJsonNode(entry.getValue())));
+            return Document.fromMap(map);
+        }
+        if (node.isArray()) {
```
dhrubo-os, line 495:
```
if (node.isArray()) {
    List<Document> list = new ArrayList<>();
    for (JsonNode item : node) {  // ← Proper array iteration
        list.add(buildDocumentFromJsonNode(item));
    }
    return Document.fromList(list);
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+            // Build Bedrock client
+            BedrockRuntimeAsyncClient bedrockClient = buildBedrockRuntimeAsyncClient();
+
+            // Parse payload to build ConverseStreamRequest
+            ConverseStreamRequest request = buildConverseStreamRequest(payload, parameters);
+
+            // Build response handler with comprehensive raw logging
+            ConverseStreamResponseHandler handler = ConverseStreamResponseHandler.builder().onResponse(response -> {}).onError(error -> {
+                log.error("Converse stream error: {}", error.getMessage());
+                listener.onFailure(new MLException("Error from remote service: " + error.getMessage(), error));
```
dhrubo-os: Can you please check [here](https://github.com/opensearch-project/ml-commons/blob/72312d6a2ca9a61840da49cb9f7c2c2c91a7e4a7/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java#L90).

We should carefully treat this error. If this is a throttling error from the remote model service, then to show 5XX error is confusing to customer. Because it seems like the issue is in ml-commons but actually problem is in remote service what customer is using. So in that cases error should be 4XX level so that customer can take action on that.

dhrubo-os:
Is there any chance Stream could hang indefinitely? Do we have any timeouts for streaming?

ml-algorithms/src/main/java/org/opensearch/ml/engine/Executable.java
nathaliellenaa: Yes, will change it to use method overloading

plugin/src/main/java/org/opensearch/ml/task/MLTaskRunner.java
nathaliellenaa: Yes this is expected, default is false. We only override this method in MLPredictTaskRunner and MLExecuteTaskRunner

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+            // Add messages if present
+            if (payloadJson.has("messages")) {
+                requestBuilder.messages(parseMessages(payloadJson.get("messages")));
+            }
+
+            // Add tool configuration if present
+            if (payloadJson.has("toolConfig")) {
+                requestBuilder.toolConfig(parseToolConfig(payloadJson.get("toolConfig")));
+            }
+            return requestBuilder.build();
```
Zhangxunmt: Why not use the ConverseStreamRequest.builder() to simplify this block? 
return ConverseStreamRequest.builder()
            .modelId(parameters.get("model"))
            .system(getOptionalNode(payloadJson, "system").map(this::parseSystemMessages).orElse(null))
            .messages(getOptionalNode(payloadJson, "messages").map(this::parseMessages).orElse(null))
            .toolConfig(getOptionalNode(payloadJson, "toolConfig").map(this::parseToolConfig).orElse(null))
            .build();

private Optional<JsonNode> getOptionalNode(JsonNode json, String field) {
    return Optional.ofNullable(json.get(field));
}

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
```diff
@@ -0,0 +1,435 @@
...
+    @Override
+    public void handleError(Throwable error, StreamPredictActionListener<MLTaskResponse, ?> listener) {
+        log.error("HTTP streaming error", error);
+        listener.onFailure(new MLException("Fail to execute streaming", error));
+    }
+
+    private MLTaskResponse createToolUseResponse(
+        AtomicReference<String> toolName,
+        AtomicReference<Map<String, Object>> toolInput,
+        AtomicReference<String> toolUseId
```
Zhangxunmt, line 272: // Validate inputs
    if (toolName == null || toolInput == null || toolUseId == null) {
        throw new IllegalArgumentException("Tool references cannot be null");
    }

This null safety check should be needed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
nathaliellenaa: thanks for the suggestion, I created a streaming wrapper to handle the difference

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
nathaliellenaa: Fixed the implementation

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/BedrockStreamingHandler.java
nathaliellenaa: Thanks for the suggestion, applied

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentRunner.java
```diff
@@ -20,6 +21,7 @@ public interface MLAgentRunner {
      * @param mlAgent
      * @param params
      * @param listener
+     * @param channel
      */
-    void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Object> listener);
+    void run(MLAgent mlAgent, Map<String, String> params, ActionListener<Object> listener, TransportChannel channel);
```
dhrubo-os, line 36: Can we do the same here like `Executable.java`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -540,17 +552,108 @@ private void runReAct(
...
-                .build(),
-            null,
-            tenantId
-        );
-        client.execute(MLPredictionTaskAction.INSTANCE, request, firstListener);
+        ActionRequest request = wrapper.createPredictionRequest(llm, tmpParameters, tenantId);
+        wrapper.executeRequest(request, firstListener);
+    }
+
+    private class StreamingWrapper {
```
dhrubo-os: This class is already large. Can we take this out in a separate class?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -204,11 +193,21 @@ default void setUserRateLimiterMap(Map<String, TokenBucket> userRateLimiterMap)
 
     default void setMlGuard(MLGuard mlGuard) {}
 
+    default void preparePayloadAndInvoke(
```
dhrubo-os, line 196: how both can be default? at line 205 I'm seeing `default void preparePayloadAndInvoke(`

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
nathaliellenaa: Added, I'll also go through every comments during refactoring to make sure I didn't miss anything

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
nathaliellenaa: Created a separate `StreamingWrapper` class

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
nathaliellenaa: I think we can have both as default right? If we set line 205 to `void preparePayloadAndInvoke` we need to move all the implementation inside it since interface abstract method cannot have body.

dhrubo-os: My bad, method signature is different, which is why you can't have both as default. Both methods CAN be default - The first method is a convenience method that calls the second method with null values for the additional parameters. This is a common pattern for method overloading with default values.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -16,6 +16,7 @@
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.ml.common.transport.MLTaskResponse;
+import org.opensearch.ml.engine.algorithms.remote.streaming.StreamPredictActionListener;
```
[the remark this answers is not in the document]
nathaliellenaa: Right, will remove `sendContentResponse` and `sendCompletionResponse` method and remove this import

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/StreamingWrapper.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public ActionRequest createPredictionRequest(LLMSpec llm, Map<String, String> parameters, String tenantId) {
+        return new MLPredictionTaskRequest(
+            llm.getModelId(),
+            RemoteInferenceMLInput
+                .builder()
+                .algorithm(FunctionName.REMOTE)
+                .inputDataset(RemoteInferenceInputDataSet.builder().parameters(parameters).build())
+                .build(),
+            !isStreaming, // set dispatchTask to false for streaming
```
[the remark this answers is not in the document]
nathaliellenaa: No, it should be `!isStreaming`. Currently agent streaming is only working when it executes the entire process locally on the receiving node, that's why we need to specify dispatchTask=false for agent streaming. 

If we set dispatchTask=true, it will cause a flaky behavior. I debugged this issue and found that when running across multiple nodes (e.g., from node 2 to node 1's REST layer), there's a type mismatch in the output (expects MODEL_TENSOR but receives TRAINING type instead). Will add todo here and need to debug this issue further and try to fix this in future PR.

nathaliellenaa:
> @nathaliellenaa should agent streaming have its own feature flag? i know we have for predict and are reusing for agents

I think single feature flag for both predict and agent streaming is ok. Users don't need to enable 2 feature flags when they want to try to predict and agent streaming.

nathaliellenaa:
Failing CI seems unrelated, run both IT locally and they're passing
```
RestCohereInferenceIT > test_cohereInference_withDifferent_postProcessFunction FAILED
    java.lang.AssertionError: failed to run test with test name: connector.post_process.cohere_v2.embedding.float_test
        at __randomizedtesting.SeedInfo.seed([EA9F7EF19FCC2013:1911E683B8CE553B]:0)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.opensearch.ml.rest.RestCohereInferenceIT.test_cohereInference_withDifferent_postProcessFunction(RestCohereInferenceIT.java:77)


RestMLRAGSearchProcessorIT > testBM25WithBedrockConverse FAILED
    org.opensearch.client.ResponseException: method [POST], host [http://[::1]:33109], URI [/test/_search?size=5&search_pipeline=pipeline_test], status line [HTTP/1.1 429 Too Many Requests]
    {"error":{"root_cause":[{"type":"remote_connector_throttling_exception","reason":"Error from remote service: The request was denied due to remote server throttling. To change the retry policy and behavior, please update the connector client_config."}],"type":"remote_connector_throttling_exception","reason":"Error from remote service: The request was denied due to remote server throttling. To change the retry policy and behavior, please update the connector client_config."},"status":429}
```

ylwu-amzn (approved):

dhrubo-os:
approving to merge the PR by 2:00 PM.

dhrubo-os (approved):

PR #4159, opened 2025-09-06
[Agentic Memory] Allow other LLMs for Fact Extraction

rithin-pullela-aws:
### Description
With this PR we expect the users to pass `response_filter` in parameters of the LLM connectors and thus allowing more flexible LLM options for fact extraction in Agentic Memory. This also means we can have a more flexible connector request body without tying to only claude models.

To use a Claude model:
```
{
  "name": "Bedrock infer model",
  "function_name": "remote",
  "description": "LLM model for memory processing",
  "connector": {
    "name": "Amazon Bedrock Connector: LLM",
    "description": "The connector to bedrock Claude 3.7 sonnet model",
    "version": 1,
    "protocol": "<AWS_REGION>",
    "parameters": {
      "region": "us-east-1",
      "service_name": "bedrock",
      "max_tokens": 8000,
      "temperature": 1,
      "anthropic_version": "bedrock-2023-05-31",
      "model": "us.anthropic.claude-3-5-sonnet-20240620-v1:0",
      "response_filter": "$.content[0].text"
    },
    "credential": {
        "access_key": <AWS_ACCESS_KEY>,
        "secret_key": <AWS_SECRET_KEY>,
        "session_token": <AWS_SESSION_TOKEN>
    },
    "actions": [{
      "action_type": "predict",
      "method": "POST",
      "headers": {"content-type": "application/json"},
      "url": "https://bedrock-runtime.${parameters.region}.amazonaws.com/model/${parameters.model}/invoke",
      "request_body": "{ \"system\": \"${parameters.system_prompt}\", \"anthropic_version\": \"${parameters.anthropic_version}\", \"max_tokens\": ${parameters.max_tokens}, \"temperature\": ${parameters.temperature}, \"messages\": [{\"role\":\"user\",\"content\":[{ \"type\": \"text\", \"text\":\"${parameters.messages}\"}]}]}"
    }]
  }
}

```

To use Open AI models:
```
{
    "name": "My OpenAI model: gpt-4o",
    "function_name": "remote",
    "description": "test model",
    "connector": {
        "name": "My openai connector: gpt-4o mini",
        "description": "The connector to openai chat model",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "model": "gpt-4o-mini",
            "response_filter": "$.choices[0].message.content"
        },
        "credential": {
            "openAI_key": <OPENAI_API_KEY>
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
[...]

PR #4218, opened 2025-09-29
Refactor Agentic Memory

ylwu-amzn:
### Description

Refactor agentic memory. Read these docs to learn more details

https://github.com/opensearch-project/ml-commons/blob/feature/agentic_memory_new/docs/tutorials/agentic_memory/agentic_memory_apis.md
https://github.com/opensearch-project/ml-commons/blob/feature/agentic_memory_new/docs/tutorials/agentic_memory/agentic_memory_tutorial.md

### Related Issues

### Check List


plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -216,40 +273,35 @@ private List<String> parseFactsFromLLMResponse(MLOutput mlOutput) {
...
-                                            facts.add(fact);
-                                        }
-                                    } else {
-                                        parser.skipChildren();
-                                    }
-                                }
+            String llmResultPath = Optional
+                .ofNullable(strategy.getStrategyConfig().get("llm_result_path"))
+                .map(Object::toString)
+                .orElse(DEFAULT_LLM_RESULT_PATH);
```
rithinpullela, line 285: Do we expect users to provide the llm_result_path in the strategy config?

Can we use  the response filter during the registration of the model directly? We use this pattern in other places as well so things can be consistent for the users.

The users can provide: `"response_filter": "$.output.message.content[0].text"` for Claude models

And `"response_filter": "$.choices[0].message.content"` for GPT models.

rithinpullela: I tried to do that in this PR: https://github.com/opensearch-project/ml-commons/pull/4159

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -109,9 +143,15 @@ public class MemoryContainerConstants {
...
...ound and experiences</item>\n<item>Contact information (if shared)</item>\n<item>Availability and schedule preferences</item>\n</include_list>\n</instruction>\n<instruction>Organize each piece of information as a separate fact.</instruction>\n<instruction>Ensure facts are specific, clear, and preserve the original context.</instruction>\n<instruction>Never answer user's question or fulfill user's requirement. You are a personal information manager, not a helpful assistant.</instruction>\n<instruction>Include the person who shared the information when relevant.</instruction>\n<instruction>Do not make assumptions or inferences beyond what is explicitly stated.</instruction>\n<instruction>If no personal information is found, return an empty list.</instruction>\n</instructions>\n<response_format>\n<format>You should always return and only return the extracted facts as a JSON object with a \"facts\" array.</format>\n<example>\n{\n  \"facts\": [\n    \"User's name is John Smith\",\n    \"John works as a software engineer at TechCorp\",\n    \"John enjoys hiking on weekends\",\n    \"John is looking to improve his Python skills\"\n  ]\n}\n</example>\n</response_format>\n</system_prompt>";
```
Zhangxunmt, line 146: This is almost identical to the PERSONAL_INFORMATION_ORGANIZER_PROMPT. I think we should use a different prompt for semantic extraction.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
ylwu-amzn: Use response filter is not so flexible considering support multiple strategies and multi-modal. We allow user configure multiple strategies in container. Define path filter in strategy is more flexible, so each strategy can configure their own filter path.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -50,36 +60,83 @@ public MemoryProcessingService(Client client, NamedXContentRegistry xContentRegi
...
+        List<MessageInput> messages,
+        MemoryConfiguration memoryConfig,
+        ActionListener<List<String>> listener
+    ) {
+        if ("semantic".equalsIgnoreCase(strategy.getType())
+            || "user_preference".equalsIgnoreCase(strategy.getType())
+            || "summary".equalsIgnoreCase(strategy.getType())) {
+            extractFactsFromConversation(messages, strategy, memoryConfig, listener);
+        } else {
+            listener.onFailure(new IllegalArgumentException("Unsupported memory strategy type: " + strategy.getType()));
```
Zhangxunmt, line 76: better to have add a log error here.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -317,4 +365,42 @@ private List<MemoryDecision> parseMemoryDecisions(MLTaskResponse response) {
             throw new RuntimeException("Failed to parse memory decisions", e);
         }
     }
+
+    public void summarizeMessages(List<MessageInput> messages, ActionListener<String> listener) {
+        if (messages == null || messages.isEmpty()) {
+            listener.onResponse("");
```
Zhangxunmt, line 377: Is it expected to response an empty String or should an exception to be thrown?
throw new IllegalStateException("Messages list cannot be empty");

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
```diff
@@ -41,46 +51,78 @@ public class MLAddMemoriesInput implements ToXContentObject, Writeable {
...
+        this.namespace = namespace;
+        this.infer = infer; // default infer is false
+        this.metadata = metadata;
         this.tags = tags;
+        this.ownerId = ownerId;
+        validate();
+    }
+
+    public void validate() {
+        if (messages == null || messages.isEmpty()) {
```
mingshl, line 102: we should add more validate in the future pr. Now it allows the messages to be empty when infer is false.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java
```diff
@@ -9,8 +9,9 @@
  * Enum representing the type of memory entry
  */
 public enum MemoryType {
-    RAW_MESSAGE("RAW_MESSAGE"),
-    FACT("FACT");
+    SEMANTIC("SEMANTIC"),
```
mingshl, line 12: we released fact type in 3.2, then this is not backward compatible. we also need documentation explanation of the new memory types

common/src/main/resources/index-mappings/ml_memory_sessions.json
```diff
@@ -0,0 +1,34 @@
...
+    "additional_info": {
+      "type": "flat_object"
+    },
+    "namespace": {
+      "type": "flat_object"
+    },
+    "namespace_size": {
+      "type": "integer"
+    },
+    "owner": USER_MAPPING_PLACEHOLDER,
```
mingshl, line 29: so once the user_mapping is defined, it cannot be changed anymore, is it the intension? will there be use cases when the user_mapping is changing? what is the example of this user_mapping_placeholder? 

the index mapping is critical, once we create the index, if we need to change mapping, that requires re-index, and it's an expensive operation.

mingshl (commented):
Overall it looks good, but the index strategy is getting more complicated now.

 I understand there is trade off when all memory are storing in one index, then the search performance would be degraded, but adding more indices would be operational complex. 

But those 3 different type of indices are very similar in mappings and usages, we could have started with one system index then, when it grows larger we could have index partitioning strategy. If you look at the .kibana index, it has .kibana_1, .kibana_2. Just my two cents

common/src/main/resources/index-mappings/ml_memory_working.json
```diff
@@ -0,0 +1,57 @@
...
+      "type": "keyword"
+    },
+    "memory_container_id": {
+      "type": "keyword"
+    },
+    "memory_type": {
+      "type": "keyword"
+    },
+    "messages": {
+      "type": "nested",
```
mingshl, line 16: Nested + flat_object can be expensive for large message arrays

common/src/main/resources/index-mappings/ml_memory_container.json
```diff
@@ -23,36 +23,70 @@
...
+          "type": "boolean"
+        },
+        "use_system_index": {
+          "type": "boolean"
         },
         "max_infer_size": {
           "type": "integer"
+        },
+        "strategies": {
+          "type": "object",
```
mingshl, line 57: it's an array of strategies, then should this be "nested"?

ylwu-amzn: make sense.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
```diff
@@ -148,31 +259,39 @@ public static MLAddMemoriesInput parse(XContentParser parser) throws IOException
                 case MEMORY_CONTAINER_ID_FIELD:
                     memoryContainerId = parser.text();
```
dhrubo-os, line 260: Is that correct? Currently we are expecting `memoryContainerId` as parameter. If we do `memoryContainerId = parser.text();`, why do we need to send `memoryContainerId` as parameter?

common/src/main/java/org/opensearch/ml/common/CommonValue.java
```diff
@@ -68,6 +72,9 @@ public class CommonValue {
     public static final String ML_MEMORY_META_INDEX_MAPPING_PATH = "index-mappings/ml_memory_meta.json";
     public static final String ML_MEMORY_MESSAGE_INDEX_MAPPING_PATH = "index-mappings/ml_memory_message.json";
     public static final String ML_MEMORY_CONTAINER_INDEX_MAPPING_PATH = "index-mappings/ml_memory_container.json";
+    public static final String ML_MEMORY_SESSION_INDEX_MAPPING_PATH = "index-mappings/ml_memory_sessions.json";
```
dhrubo-os, line 73: where is the long term memory index mapping?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -0,0 +1,305 @@
...
+        Map<String, Object> structuredData,
+        Map<String, String> namespace,
+        boolean infer,
+        Map<String, String> metadata,
+        Map<String, String> tags,
+        Instant createdTime,
+        Instant lastUpdateTime,
+        String ownerId
+    ) {
+        // MAX_MESSAGES_PER_REQUEST limit removed for performance testing
```
dhrubo-os, line 81: are we putting this in place?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject();
+        if (memoryContainerId != null) {
+            builder.field(MEMORY_CONTAINER_ID_FIELD, memoryContainerId);
+        }
+        builder.field(MEMORY_TYPE_FIELD, memoryType);
+        if (messages != null && messages.size() > 0) {
```
dhrubo-os, line 178: [nit] `!messages.isEmpty()`

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
```diff
@@ -0,0 +1,251 @@
...
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.ml.common.transport.memorycontainer.memory.MemoryEvent;
+import org.opensearch.ml.common.utils.StringUtils;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Input data for adding memory to a memory container
```
dhrubo-os, line 39: can we update this comment?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerInput.java
```diff
@@ -9,63 +9,89 @@
...
 
 @Data
+@Builder
 public class MLCreateMemoryContainerInput implements ToXContentObject, Writeable {
 
     public static final String NAME_FIELD = "name";
     public static final String DESCRIPTION_FIELD = "description";
-    public static final String MEMORY_STORAGE_CONFIG_FIELD = "memory_storage_config";
+    public static final String MEMORY_CONFIG_FIELD = "configuration";
+    public static final String BACKEND_ROLES_FIELD = "backend_roles";
```
dhrubo-os, line 34: can we use this [common field](https://github.com/opensearch-project/ml-commons/blob/d1dfa0301e1a5342115c732f013a8813d81f52ab/common/src/main/java/org/opensearch/ml/common/CommonValue.java#L101)

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
dhrubo-os: +1

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java
dhrubo-os:
```
public enum MemoryType {
    // New values
    SEMANTIC("SEMANTIC"),
    USER_PREFERENCE("USER_PREFERENCE"), 
    SUMMARY("SUMMARY"),
    
    // Deprecated but supported for migration
    @Deprecated
    RAW_MESSAGE("RAW_MESSAGE"),
    @Deprecated  
    FACT("FACT");

    // Enhanced fromString method
    public static MemoryType fromString(String value) {
        if (value == null) return null;
        
        // Handle legacy values with mapping
        switch (value.toUpperCase()) {
            case "RAW_MESSAGE": return USER_PREFERENCE;  // I know it doesn't make enough sense. But not sure what can be replaced with RAW_MESSAGE
            case "FACT": return SEMANTIC; // Map old to new
            default:
                // Normal enum lookup
                for (MemoryType type : MemoryType.values()) {
                    if (type.value.equalsIgnoreCase(value)) {
                        return type;
                    }
                }
        }
        throw new IllegalArgumentException("Invalid memory type: " + value);
    }
}
```

we can do something like this for backward compatibility.

plugin/src/test/resources/org/opensearch/ml/bwc/2.4.0.0/opensearch-ml-2.4.0.0.zip
dhrubo-os, line 1: Why is this here?

b4sjoo: This one  need to be delete

plugin/src/main/java/org/opensearch/ml/rest/RestMLSearchMemoryContainerAction.java
```diff
@@ -0,0 +1,38 @@
...
+import org.opensearch.ml.common.memorycontainer.MLMemoryContainer;
+import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
+import org.opensearch.ml.common.transport.memorycontainer.MLMemoryContainerSearchAction;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * This class consists of the REST handler to search ML Models.
+ */
+public class RestMLSearchMemoryContainerAction extends AbstractMLSearchAction<MLMemoryContainer> {
```
dhrubo-os, line 20: can we check feature enablement settings first before performing search operation?

plugin/src/main/java/org/opensearch/ml/rest/RestMLGetMemoryAction.java
```diff
@@ -46,7 +47,7 @@ public String getName() {
     }
 
     @Override
-    public List<Route> routes() {
+    public List<Route> routes() {// TODO
```
dhrubo-os, line 50: What is this `TODO` for?

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoriesByQueryAction.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+
+        // If no query provided, default to match_all
+        if (query == null) {
+            query = QueryBuilders.matchAllQuery();
+        }
+
+        // Create the delete by query request
+        MLDeleteMemoriesByQueryRequest deleteRequest = new MLDeleteMemoriesByQueryRequest(memoryContainerId, memoryType, query);
```
dhrubo-os, line 83: Can't we perform all these in the transport layer?

b4sjoo: Can you clarify your question?

dhrubo-os: Sure, let me know if that helps!

## **ml-commons Layer Separation Pattern**

In ml-commons, the typical pattern is:

1. REST Layer: Minimal - just parameter extraction and basic validation
2. Transport Layer: Heavy lifting - business logic, security, data processing

## **Current Anti-Pattern in RestMLDeleteMemoriesByQueryAction.java**

Looking at the new delete-by-query implementation, if it follows the same pattern as other REST handlers, it likely has:

• Query parsing logic
• Memory container validation  
• Index resolution
• Security checks
• Query transformation

## **Why This Should Move to Transport Layer**

1. Consistency with Existing Code
java
// Other ml-commons REST handlers are thin:
RestMLCreateMemoryContainerAction -> MLCreateMemoryContainerTransportAction
RestMLSearchMemoriesAction -> MLSearchMemoriesTransportAction


2. Reusability
• Transport actions can be called internally by other components
• REST handlers are only accessible via HTTP API

3. Security Model
• Transport layer handles user context, permissions, tenant isolation
• REST layer shouldn't duplicate this logic

4. Testing
• Transport actions are easier to unit test
• Business logic testing doesn't require HTTP mocking

## **What Should Be in Each Layer**
Sure, here's 


REST Layer (minimal):
java
// Just parameter extraction
String memoryContainerId = request.param("memory_container_id");
String memoryType = request.param("memory_type");
QueryBuilder query = parseQuery(request);


Transport Layer (business logic):
java
// All the heavy lifting:
- Validate memory container exists
- Check user permissions  
- Resolve target indices
- Execute delete by query
- Handle errors and responses

Zhangxunmt (approved):

dhrubo-os:
Approving to unblock the merging before code freeze cutoff time.

dhrubo-os (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
```diff
@@ -122,9 +151,127 @@ public void initMLAgentIndex(ActionListener<Boolean> listener) {
...
+        initIndexIfAbsent(indexName, StringUtils.toJson(indexMappings), indexSettings, 1, listener);
+    }
+
+    public void createLongTermMemoryHistoryIndex(String indexName, MemoryConfiguration configuration, ActionListener<Boolean> listener) {
+        String indexMappings = getMapping(ML_LONG_MEMORY_HISTORY_INDEX_MAPPING_PATH);
+        Map<String, Object> indexSettings = configuration.getMemoryIndexMapping(LONG_TERM_MEMORY_HISTORY_INDEX);
+        initIndexIfAbsent(indexName, StringUtils.toJson(indexMappings), indexSettings, 1, listener);
+    }
+
+    public void createLongTermMemoryIndex(
```
b4sjoo, line 196: I think we should store as .json instead of create a mapping here

dhrubo-os: Can we refactor the change then?

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoriesByQueryAction.java
b4sjoo: Looks reasonable, I will check if we can change it

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -13,47 +13,74 @@ public class MemoryContainerConstants {
...
+    public static final String NAMESPACE_FIELD = "namespace";
+    public static final String STRATEGY_CONFIG_FIELD = "configuration";
+    public static final String BINARY_DATA_FIELD = "binary_data";
+    public static final String STRUCTURED_DATA_FIELD = "structured_data";
+    public static final String NAMESPACE_SIZE_FIELD = "namespace_size";
     public static final String MEMORY_FIELD = "memory";
     public static final String MEMORY_EMBEDDING_FIELD = "memory_embedding";
+    public static final String METADATA_FIELD = "metadata";
     public static final String TAGS_FIELD = "tags";
+    public static final String STRATEGY_ID_FIELD = "strategy_id";
```
dhrubo-os, line 66: What is this for?

b4sjoo: Think about a case the user specified two SEMANTIC strategy object with one of them has customized prompt. Then we use this `strategy_id` to do the separation to let user audit the fact extraction quality on different prompt.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java
b4sjoo: Confirmed by @ylwu-amzn:

> We haven't GA this feature so no BWC need to be considered.

common/src/main/resources/index-mappings/ml_memory_working.json
b4sjoo: Discussed offline, we will keep this for now. But will do some testing and review the performance to see if this needs to be enhanced in the future.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
b4sjoo: This is expected. We do allow messages to be empty when infer is false.

common/src/main/java/org/opensearch/ml/common/CommonValue.java
b4sjoo: Need to be added as *.json.

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoriesByQueryAction.java
b4sjoo: Refactored.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
b4sjoo: Update for `MLMemoryHistory`, `MLMemorySession`, `MLWorkingMemory`

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
b4sjoo: Changed.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLCreateMemoryContainerInput.java
b4sjoo: Removed.

common/src/main/resources/index-mappings/ml_memory_sessions.json
b4sjoo: Check `ml_task.json` or `ml_model.json` or even `ml_connector.json` They all have this field. It's ML system index dev pattern.

b4sjoo: But yeah I do agree this is one way door which we need to carefully design and review.

plugin/src/main/java/org/opensearch/ml/rest/RestMLGetMemoryAction.java
b4sjoo: Removed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/indices/MLIndicesHandler.java
b4sjoo: Changed.

b4sjoo: Quick Notes: Issue here is we don't know what type of the index this system index will be until the user have explicitly give the configuration. It might be either a knn index, a sparse index, or static index

common/src/main/resources/index-mappings/ml_memory_container.json
b4sjoo: Fixed.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
b4sjoo: Agree

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
b4sjoo: FYI, `List.isEmpty()` does not have null check.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
b4sjoo: You are right. Fixed.

plugin/src/main/java/org/opensearch/ml/rest/RestMLSearchMemoryContainerAction.java
b4sjoo: Added

PR #4234, opened 2025-10-01
fix: refactor memory delete by query API to avoid anti pattern

b4sjoo:
### Description
Migrate API business logic from rest layer to transport layer to follow our ML plugin API develop pattern. Also try to add more UT coverage to fix CI

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.memorycontainer.memory;
+
+import java.io.IOException;
+
+import org.opensearch.core.action.ActionResponse;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.index.reindex.BulkByScrollResponse;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Response wrapper for delete memories by query action.
+ * Wraps BulkByScrollResponse and implements ToXContentObject for REST response serialization.
+ */
+@Getter
+@AllArgsConstructor
+public class MLDeleteMemoriesByQueryResponse extends ActionResponse implements ToXContentObject {
+
+    private final BulkByScrollResponse bulkResponse;
+
+    /**
+     * Constructor for deserialization
+     * @param in StreamInput to read from
+     * @throws IOException if deserialization fails
+     */
+    public MLDeleteMemoriesByQueryResponse(StreamInput in) throws IOException {
+        super(in);
+        this.bulkResponse = new BulkByScrollResponse(in);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        bulkResponse.writeTo(out);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject();
+
+        // Basic statistics
+        builder.field("took", bulkResponse.getTook().millis());
+        builder.field("timed_out", bulkResponse.isTimedOut());
+        builder.field("deleted", bulkResponse.getDeleted());
+        builder.field("batches", bulkResponse.getBatches());
+        builder.field("version_conflicts", bulkResponse.getVersionConflicts());
+        builder.field("noops", bulkResponse.getNoops());
+
+        // Retries information
+        builder.startObject("retries");
+        builder.field("bulk", bulkResponse.getBulkRetries());
+        builder.field("search", bulkResponse.getSearchRetries());
+        builder.endObject();
+
+        // Throttling information
+        builder.field("throttled_millis", bulkResponse.getStatus().getThrottled().millis());
+        builder.field("requests_per_second", bulkResponse.getStatus().getRequestsPerSecond());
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
```diff
@@ -0,0 +1,94 @@
...
+        builder.field("throttled_until_millis", bulkResponse.getStatus().getThrottledUntil().millis());
+
+        // Add bulk failures if any
+        if (bulkResponse.getBulkFailures() != null && !bulkResponse.getBulkFailures().isEmpty()) {
+            builder.startArray("bulk_failures");
+            for (var failure : bulkResponse.getBulkFailures()) {
+                builder.startObject();
+                builder.field("index", failure.getIndex());
+                builder.field("id", failure.getId());
+                builder.field("cause", failure.getCause().getMessage());
```
dhrubo-os, line 75: [nit] `builder.field("cause", failure.getCause() != null ? failure.getCause().getMessage() : "Unknown error");`

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
```diff
@@ -0,0 +1,94 @@
...
+     * @throws IOException if deserialization fails
+     */
+    public MLDeleteMemoriesByQueryResponse(StreamInput in) throws IOException {
+        super(in);
+        this.bulkResponse = new BulkByScrollResponse(in);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        bulkResponse.writeTo(out);
```
dhrubo-os, line 42: don't we need `super.writeTo(out)` here?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
```
ylwu-amzn, line 46: For search memory API, we just return the search response. Is it ok to just return the delete by query response directly?

b4sjoo: Let me check

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportDeleteMemoriesByQueryAction.java
ylwu-amzn, line 230: How about extracting this into method `deleteDataByQuery` and add to `MemoryContainerHelper`

ylwu-amzn, line 170: Seems we could tune `addOwnerIdFilter` method in `MemroyContainerHelper` to avoid duplicate logic.

dhrubo-os:
Should we merge this first while @b4sjoo can address comments in a follow up PR? 

The PRs workflow is not passing and blocked due to this: https://github.com/opensearch-project/ml-commons/actions/runs/18170353481/job/51723290024?pr=3715

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
b4sjoo: Used `BulkByScrollResponse.toXcontent`

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
b4sjoo: simplified

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportDeleteMemoriesByQueryAction.java
b4sjoo: Simplified

ylwu-amzn:
Merge this PR to unblock other PRs, @b4sjoo you can cut new PR to address comments

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLDeleteMemoriesByQueryResponse.java
b4sjoo: No, you don't need super.writeTo(out) in MLDeleteMemoriesByQueryResponse.

  Here's why:

  1. Pattern in ML Commons: Looking at all the Response classes in ML Commons that extend ActionResponse, none of them call super.writeTo(out) in their writeTo() methods. Examples:
    - MLTaskResponse: Only writes output.writeTo(out)
    - MLAgentGetResponse: Only writes mlAgent.writeTo(out)
    - MLCreateConnectorResponse: Only writes out.writeString(connectorId)
  2. ActionResponse base class: The ActionResponse class in OpenSearch is essentially an empty marker class. Its writeTo() method in the base class doesn't write any data that needs to be preserved. It's designed to be completely overridden by subclasses.
  3. Distinction from Request classes: You'll notice that Request classes (like MLTaskRequest, MLAgentGetRequest, etc.) DO call super.writeTo(out) because ActionRequest base class has important fields that need to be serialized. But Response classes don't have this requirement.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportDeleteMemoriesByQueryAction.java
b4sjoo: This required to update the `SearchSourceBuilder` to `QueryBuilder` class in `addOwnerIdFilter`, I will test it to see if this makes sense

PR #4246, opened 2025-10-02
add create session API; add message id to working memory; fix update api

ylwu-amzn:
### Description
add create session API; add message id to working memory; fix update api

#### 1. If provide session_id, will use the provided id to create session.
Example 
```
POST _plugins/_ml/memory_containers/{{mem_container_id}}/memories/sessions
{
  "session_id": "abc123",
  "metadata": {
    "key1": "value1"
  }
}
```

Response
```
{
  "session_id": "abc123",
  "status": "created"
}
```

#### 2. If not provide session_id, will create session with a random id.

```
POST _plugins/_ml/memory_containers/{{mem_container_id}}/memories/sessions
{
  "summary": "This is a test session",
  "metadata": {
    "key1": "value1"
  },
  "namespace": {
    "user_id": "bob"
  }
}
```
Response
```
{
  "session_id": "jTYm35kBt8CyICnjxJl9",
  "status": "created"
}
```

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
@@ -107,6 +107,10 @@ public class MemoryContainerConstants {
     public static final String UPDATE_MEMORY_PATH = MEMORIES_PATH + "/{" + PARAMETER_MEMORY_TYPE + "}" + "/{" + PARAMETER_MEMORY_ID + "}";
     public static final String GET_MEMORY_PATH = MEMORIES_PATH + "/{" + PARAMETER_MEMORY_TYPE + "}" + "/{" + PARAMETER_MEMORY_ID + "}";
     public static final String DELETE_MEMORIES_BY_QUERY_PATH = MEMORIES_PATH + "/{" + PARAMETER_MEMORY_TYPE + "}" + "/_delete_by_query";
+    public static final String MEM_CONTAINER_MEMORY_TYPE_SESSIONS = "sessions";
+    public static final String MEM_CONTAINER_MEMORY_TYPE_WORKING = "working";
+    public static final String MEM_CONTAINER_MEMORY_TYPE_LONG_TERM = "long-term";
+    public static final String MEM_CONTAINER_MEMORY_TYPE_HISTORY = "history";
 
     // Memory types are defined in MemoryType enum
 
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
@@ -124,8 +124,8 @@ public class TransportUpdateMemoryAction extends HandledTransportAction<ActionRe
                 }
 
                 // Prepare the update
-                Map<String, Object> updateFields = constructUpdateFields(updateRequest.getMlUpdateMemoryInput(), memoryType, originalDoc);
-                IndexRequest indexRequest = new IndexRequest(memoryIndexName).id(memoryId).source(updateFields);
+                Map<String, Object> newDoc = constructNewDoc(updateRequest.getMlUpdateMemoryInput(), memoryType, originalDoc);
+                IndexRequest indexRequest = new IndexRequest(memoryIndexName).id(memoryId).source(newDoc);
                 memoryContainerHelper.indexData(container.getConfiguration(), indexRequest, actionListener);
 
             }, actionListener::onFailure);
@@ -134,18 +134,18 @@ public class TransportUpdateMemoryAction extends HandledTransportAction<ActionRe
         }, actionListener::onFailure));
     }
 
-    public Map<String, Object> constructUpdateFields(MLUpdateMemoryInput input, String memoryType, Map<String, Object> originalDoc) {
+    public Map<String, Object> constructNewDoc(MLUpdateMemoryInput input, String memoryType, Map<String, Object> originalDoc) {
         Map<String, Object> updateFields = new HashMap<>();
         updateFields.putAll(originalDoc);
         Map<String, Object> updateContent = input.getUpdateContent();
         switch (memoryType) {
-            case "session":
+            case MEM_CONTAINER_MEMORY_TYPE_SESSIONS:
                 constructSessionMemUpdateFields(updateFields, updateContent);
                 break;
-            case "working":
+            case MEM_CONTAINER_MEMORY_TYPE_WORKING:
                 constructWorkingMemUpdateFields(updateFields, updateContent);
                 break;
-            case "long-term":
+            case MEM_CONTAINER_MEMORY_TYPE_LONG_TERM:
                 constructLongTermMemUpdateFields(updateFields, updateContent);
                 break;
             default:
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
@@ -21,6 +21,10 @@ import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MAX_INFER_SIZE_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MAX_INFER_SIZE_LIMIT_ERROR;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEMORY_INDEX_PREFIX_FIELD;
+import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEM_CONTAINER_MEMORY_TYPE_HISTORY;
+import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEM_CONTAINER_MEMORY_TYPE_LONG_TERM;
+import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEM_CONTAINER_MEMORY_TYPE_SESSIONS;
+import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEM_CONTAINER_MEMORY_TYPE_WORKING;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.PARAMETERS_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.SEMANTIC_STORAGE_EMBEDDING_MODEL_ID_REQUIRED_ERROR;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.SEMANTIC_STORAGE_EMBEDDING_MODEL_TYPE_REQUIRED_ERROR;
@@ -59,7 +63,13 @@ import lombok.Setter;
 @EqualsAndHashCode
 public class MemoryConfiguration implements ToXContentObject, Writeable {
 
-    public static final Set<String> VALID_MEMORY_TYPES = Set.of("session", "working", "long-term", "history");
+    public static final Set<String> VALID_MEMORY_TYPES = Set
+        .of(
+            MEM_CONTAINER_MEMORY_TYPE_SESSIONS,
+            MEM_CONTAINER_MEMORY_TYPE_WORKING,
+            MEM_CONTAINER_MEMORY_TYPE_LONG_TERM,
+            MEM_CONTAINER_MEMORY_TYPE_HISTORY
+        );
     private String indexPrefix;
     private FunctionName embeddingModelType;
     private String embeddingModelId;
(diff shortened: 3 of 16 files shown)


common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemorySession.java
```diff
@@ -59,6 +65,8 @@ public MLMemorySession(
         this.summary = summary;
         this.createdTime = createdTime;
         this.lastUpdateTime = lastUpdateTime;
+        this.metadata = metadata;
+        this.agents = agents;
```
b4sjoo, line 69: What is this agent field for? Shouldn't it be agent_id_field?

b4sjoo:
Could you please explain why we need this API?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -83,9 +88,11 @@ public MLWorkingMemory(
         this.memoryContainerId = memoryContainerId;
         this.memoryType = memoryType == null ? WorkingMemoryType.CONVERSATIONAL : memoryType;
         this.messages = messages;
+        this.messageId = messageId;
         this.binaryData = binaryData;
         this.structuredData = structuredData;
         this.namespace = namespace;
+        this.namespaceSize = namespace == null ? null : namespace.size();
```
rithinpullela, line 95: This variable can be derived from the namespace variable, do we really need this?

common/src/main/java/org/opensearch/ml/common/transport/session/MLCreateSessionAction.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.session;
+
+import org.opensearch.action.ActionType;
+
+public class MLCreateSessionAction extends ActionType<MLCreateSessionResponse> {
+    public static final MLCreateSessionAction INSTANCE = new MLCreateSessionAction();
+    public static final String NAME = "cluster:admin/opensearch/ml/sessions/create";
```
rithinpullela: Since this is related to memory should we call it something like`cluster:admin/opensearch/ml/memory_containers/sessions/create` to indicate these sessions are associated with agentic memory?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryAction.java
```diff
@@ -6,13 +6,14 @@
 package org.opensearch.ml.common.transport.memorycontainer.memory;
 
 import org.opensearch.action.ActionType;
+import org.opensearch.action.index.IndexResponse;
 import org.opensearch.action.update.UpdateResponse;
```
rithinpullela: Seems like an unused import

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -45,29 +47,32 @@
...
 @Getter
 @Setter
-@Builder
 public class MLWorkingMemory implements ToXContentObject, Writeable {
 
     // Required fields
     private String memoryContainerId;
     private WorkingMemoryType memoryType;
     private List<MessageInput> messages;
+    private Integer messageId;
```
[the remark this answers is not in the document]
ylwu-amzn: Some agent framework like Strands using messageId to track the message order, it's a int number. Start from 0.
It's an optional field

ylwu-amzn:
> Could you please explain why we need this API?

Some agent framework needs to create session explicitly.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemorySession.java
ylwu-amzn: Strands will save agents data to session. That shows which agent is being used for current sessoin

common/src/main/java/org/opensearch/ml/common/transport/session/MLCreateSessionResponse.java
```diff
@@ -0,0 +1,54 @@
...
+
+@Getter
+@Builder
+public class MLCreateSessionResponse extends ActionResponse implements ToXContentObject {
+
+    private static final String SESSION_ID_FIELD = "session_id";
+    private static final String STATUS_FIELD = "status";
+
+    private String sessionId;
+    private String status;
```
rithinpullela, line 27: Looks like we are setting only one value for status in the whole flow. Do we plan to add any more statuses in the future?

b4sjoo:
Another high level concern: right now you found gap for strands which need some new api and new fields. This is more like a customization. Then what if we found other agent frameworks requires other new APIs and new fields in the future? What if they are conflicting to each other? Is there any way that we can extend this in a more convenient way?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemorySession.java
rithinpullela: What kind of data is passed by strands? Can we call it agentMetadata? Because when I see `agents` I am thinking it is a list of agents, but looks like it is some sort of metadata added by strands about the agent?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryAction.java
```diff
@@ -6,13 +6,14 @@
...
 
-public class MLUpdateMemoryAction extends ActionType<UpdateResponse> {
+public class MLUpdateMemoryAction extends ActionType<IndexResponse> {
     public static final MLUpdateMemoryAction INSTANCE = new MLUpdateMemoryAction();
     public static final String NAME = "cluster:admin/opensearch/ml/memory_containers/memory/update";
 
     private MLUpdateMemoryAction() {
-        super(NAME, UpdateResponse::new);
+        super(NAME, IndexResponse::new);
```
b4sjoo, line 16: Why change to indexresponse

ylwu-amzn:
> Another high level concern: right now you found gap for strands which need some new api and new fields. This is more like a customization. Then what if we found other agent frameworks requires other new APIs and new fields in the future? What if they are conflicting to each other? Is there any way that we can extend this in a more convenient way?

Not so customized, it also generic.

It's challenging to say we 100% don't need any code change in future to support other frameworks. The goal is to gradually support more popular frameworks. We are not targeting write code one time and never change it in future.

common/src/main/java/org/opensearch/ml/common/transport/session/MLCreateSessionResponse.java
ylwu-amzn: Yes, It's more for future extension. We may add more status like if user create a session with a specific session_id, but that id exists, we may prefer to show a different status like "existing".

common/src/main/java/org/opensearch/ml/common/transport/session/MLCreateSessionAction.java
ylwu-amzn: make sense

b4sjoo:
So why strands required explicit call on session creation api? Can't we create session passively?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemorySession.java
ylwu-amzn: It's not just agent meta data , check example

```
    "agents": {
      "conversation_manager_state": {
        "removed_message_count": 0,
        "__name__": "SlidingWindowConversationManager"
      },
      "agent_id": "default",
      "updated_at": "2025-10-02T17:50:22.417762+00:00",
      "created_at": "2025-10-02T17:47:11.438447+00:00",
      "state": {}
    }
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryAction.java
ylwu-amzn: UpdateResponse not being used. 

Prevoiusly we use update request to update doc partially. that will cause flat object merging not replacing old values. 

Example
```
# old value of 
"metadata": {
 "abc": 123
  "def": 456
}

# update with 
"metadata": {
 "abc": 333
  "key1": value1
}

# new value will be 
"metadata": {
 "abc": 333
  "def": 456,
  "key1": value1
}

# With this PR, the new value will be 
"metadata": {
 "abc": 333
  "key1": value1
}

```

ylwu-amzn:
> So why strands required explicit call on session creation api? Can't we create session passively?

No,  check this strands API https://github.com/strands-agents/sdk-python/blob/main/src/strands/session/session_repository.py#L13. We have to implement create session explicitly 
```
class SessionRepository(ABC):
    """Abstract repository for creating, reading, and updating Sessions, AgentSessions, and AgentMessages."""

    @abstractmethod
    def create_session(self, session: Session, **kwargs: Any) -> Session:
        """Create a new Session."""
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
ylwu-amzn: Yes, we need, otherwise the get response will not show this field.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryAction.java
ylwu-amzn: Check reply here https://github.com/opensearch-project/ml-commons/pull/4246#discussion_r2400474176

b4sjoo:
coverage failed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
b4sjoo: Why we need get response show this field then?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
```diff
@@ -122,19 +124,19 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
                 }
 
                 // Prepare the update
-                Map<String, Object> updateFields = constructUpdateFields(updateRequest.getMlUpdateMemoryInput(), memoryType);
-                UpdateRequest updateMemoryRequest = new UpdateRequest(memoryIndexName, memoryId).doc(updateFields);
-
-                memoryContainerHelper.updateData(container.getConfiguration(), updateMemoryRequest, actionListener);
+                Map<String, Object> updateFields = constructUpdateFields(updateRequest.getMlUpdateMemoryInput(), memoryType, originalDoc);
```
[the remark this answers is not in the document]
b4sjoo: +1 Since this not follows OpenSearch native update logic, you need explicitly doc that.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -83,9 +88,11 @@ public MLWorkingMemory(
         this.memoryContainerId = memoryContainerId;
         this.memoryType = memoryType == null ? WorkingMemoryType.CONVERSATIONAL : memoryType;
         this.messages = messages;
+        this.messageId = messageId;
```
b4sjoo, line 91: Why we need a message id here? Why working memory ID cannot be used in your case?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -98,6 +100,7 @@ public class MemoryContainerConstants {
     public static final String PARAMETER_MEMORY_ID = "memory_id";
     public static final String PARAMETER_WORKING_MEMORY_ID = "working_memory_id";
     public static final String MEMORIES_PATH = BASE_MEMORY_CONTAINERS_PATH + "/{" + PARAMETER_MEMORY_CONTAINER_ID + "}/memories";
+    public static final String SESSIONS_PATH = MEMORIES_PATH + "/sessions";
```
b4sjoo, line 103: We use plural here but when it's MEMORY_TYPE we use singular, should we use plural to all to avoid confusion?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
```diff
@@ -122,19 +124,19 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
...
             }, actionListener::onFailure);
             memoryContainerHelper.getData(container.getConfiguration(), getRequest, getResponseActionListener);
 
         }, actionListener::onFailure));
     }
 
-    public Map<String, Object> constructUpdateFields(MLUpdateMemoryInput input, String memoryType) {
+    public Map<String, Object> constructUpdateFields(MLUpdateMemoryInput input, String memoryType, Map<String, Object> originalDoc) {
         Map<String, Object> updateFields = new HashMap<>();
+        updateFields.putAll(originalDoc);
```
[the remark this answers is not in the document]
ylwu-amzn: It's correct, index request is a full document replacement: OpenSearch's IndexRequest with an ID will replace the entire document, not do a partial update

plugin/src/main/java/org/opensearch/ml/action/session/TransportCreateSessionAction.java
```diff
@@ -0,0 +1,139 @@
...
+            actionListener.onFailure(new OpenSearchStatusException(ML_COMMONS_AGENTIC_MEMORY_DISABLED_MESSAGE, RestStatus.FORBIDDEN));
+            return;
+        }
+
+        MLCreateSessionInput input = request.getMlCreateSessionInput();
+
+        // Validate tenant ID
+        if (!TenantAwareHelper.validateTenantId(mlFeatureEnabledSetting, input.getTenantId(), actionListener)) {
+            return;
+        }
```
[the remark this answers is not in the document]
ylwu-amzn: I think it's ok, just reject request. `validateTenantId` is a unified method to handle the error.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
ylwu-amzn: Read this comment https://github.com/opensearch-project/ml-commons/pull/4246#discussion_r2399997302

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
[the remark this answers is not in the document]
ylwu-amzn: Make sense, will rename this to `constructNewDoc`
```suggestion
                Map<String, Object> newDoc = constructNewDoc(updateRequest.getMlUpdateMemoryInput(), memoryType, originalDoc);
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
ylwu-amzn: Good point, will change to plural

ylwu-amzn: AI suggestoins. If no concern, I will follow option1

My recommendation: You have two options:

- Option 1: Make session consistent with itself (minimal change)

Change session to use plural everywhere: `/sessions`, `/sessions/{id}`, `/sessions/_search`
Keep other types as-is: `/working/{id}`, `/long-term/{id}`, `/history/{id}`

- Option 2: Make all types plural (full consistency)

Sessions: `/sessions`, `/sessions/{id}`, `/sessions/_search`
Working: `/workings/{id}`, `/workings/_search` (or `/working-memories/...`)
Long-term: `/long-terms/{id}`, `/long-terms/_search` (or `/long-term-memories/...`)
History: `/histories/{id}`, `/histories/_search`

My preference: Option 1 - Just fix session to be consistent with itself. The other types don't have dedicated create endpoints, so the inconsistency is less confusing. Plus "workings" and "long-terms" sound awkward.

b4sjoo: Agree, go with option 1 then

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
ylwu-amzn: Make the code easier for client side

b4sjoo (approved):

PR #4255, opened 2025-10-03
fix wrong field name in get working memory API

ylwu-amzn:
### Description

Rename the memory type field to make it more clear

1. For add memories API, use `payload_type`
2. For working memory, use `payload_type` , which is same value with payload type value from add memories API
3. For long term memory, use `strategy_type` which matches the strategy type defined in container
4. For generic get/delete/update memory data API, use `memory_type`. We have four types of memory types in memory container: sessions, working, long-term, history

### Related Issues

### Check List


common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLSearchMemoriesInput.java
```diff
@@ -63,7 +63,7 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
             builder.field(MEMORY_CONTAINER_ID_FIELD, memoryContainerId);
         }
         if (memoryType != null) {
-            builder.field(MEMORY_TYPE_FIELD, memoryType);
+            builder.field(PARAMETER_MEMORY_ID, memoryType);
```
b4sjoo: Why use parameter_memory_id here

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
```diff
@@ -170,7 +170,7 @@ public static MLMemory parse(XContentParser parser) throws IOException {
                 case MEMORY_FIELD:
                     memory = parser.text();
                     break;
-                case MEMORY_TYPE_FIELD:
+                case LONG_TERM_MEMORY_TYPE_FIELD:
                     memoryType = MemoryType.fromString(parser.text());
```
dhrubo-os: should we rename [this memory type](https://github.com/opensearch-project/ml-commons/blob/1a3dbe9dd5dfd8cca9fb288ba4d63f296b0ed47a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryType.java#L18) to Long term memory type too? 

I think over the time it would be more confusing!

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLSearchMemoriesInput.java
ylwu-amzn: We have three memory types parameters in agentic memory 
1. long term memory type :  SEMANTIC, USER_REFERENCE , SUMMARY
2. working memory type:  conversational, data
3. memory type used in the API : sessions, working , long-term, history

PARAMETER_MEMORY_ID is for the third one. We should use that one, as this class is used in search memory API

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
b4sjoo: +1

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -68,7 +68,7 @@ public class MemoryContainerConstants {
     public static final String MEMORY_ACTION_FIELD = "action";
     public static final String MEMORY_BEFORE_FIELD = "before";
     public static final String MEMORY_AFTER_FIELD = "after";
-    public static final String MEMORY_TYPE_FIELD = "memory_type";
+    public static final String LONG_TERM_MEMORY_TYPE_FIELD = "memory_type";
```
dhrubo-os: `memory_type` --> should we mark this one to `long_term_memory_type`?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLSearchMemoriesInput.java
b4sjoo: Then I would say this name sounds confusing. Maybe something like memory_index_type?

b4sjoo:
BTW CI failed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
ylwu-amzn: TBH, I feel we don't need to add the `long_` and `workgin_` prefix, because the data already stored in long term meory and working memory index, adding such prefix won't help too much . Just like we don't use `connector_name`, `connector_description` in connector index.

ylwu-amzn: Use a generic `memory_type` name should be good enough

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLSearchMemoriesInput.java
ylwu-amzn: Sounds not good to expose implementation details by including index in the name

ylwu-amzn: This is internal logic, we can always tune in future, Not one-way door. We can change it to a better name in future.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemory.java
ylwu-amzn: TBH, I prefer to revert the working memory type and change the name from `working_memory_type` to `memory_type`, don't think we need to add `working_` and `long_term_` prefix.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLSearchMemoriesInput.java
b4sjoo: Eh but at least not `PARAMETER_MEMORY_ID`? I think even `PARAMETER_MEMORY_TYPE` is better

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -52,7 +52,7 @@ public class MLWorkingMemory implements ToXContentObject, Writeable {
 
     // Required fields
     private String memoryContainerId;
-    private WorkingMemoryType memoryType;
+    private PayloadType memoryType;
```
dhrubo-os: let's change `memoryType` to `payloadType`?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
```diff
@@ -253,7 +253,7 @@ public static MLWorkingMemory parse(XContentParser parser) throws IOException {
                 case MEMORY_CONTAINER_ID_FIELD:
                     memoryContainerId = parser.text();
                     break;
-                case MEMORY_TYPE_FIELD:
+                case PAYLOAD_TYPE_FIELD:
                     memoryType = parser.text();
```
dhrubo-os: [nit] `payloadType`

dhrubo-os (dismissed):

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLWorkingMemory.java
b4sjoo: Fix this one?

dhrubo-os (approved):

b4sjoo (approved):

PR #4258, opened 2025-10-04
fix: allow only container owner to delete memory container

b4sjoo:
### Description
1. Allow only container owner to delete memory container, in the meantime, add a delete_memories array field to control which memory index (indices) to delete

```
DELETE /_plugins/_ml/memory_containers/{container_id}?delete_all_memories=true
DELETE /_plugins/_ml/memory_containers/{container_id}?delete_memories=sessions,working

# Alternative: Use request body
DELETE /_plugins/_ml/memory_containers/{container_id}
{
    "delete_memories": ["sessions", "long-term", "history"]
}
```

2. Enabled update in strategy field
```
# GET
{
...
  "configuration": {
...
    "strategies": [
      {
        "id": "semantic_f904644b",
        "enabled": false,
        "type": "SEMANTIC",
        "namespace": [
          "user_id"
        ]
      }
    ]
  }
}

PUT _plugins/_ml/memory_containers/kqFHr5kBqfpW6aKup72J
{
  "strategies": [
      {
        "id": "semantic_f904644b",
        "enabled": true
      },
      {
        "type": "SUMMARY",
        "namespace": ["agent_id"],
      }
    ]
}

# GET
{
...
  "configuration": {
...
    "strategies": [
      {
        "id": "semantic_f904644b",
        "enabled": true,
        "type": "SEMANTIC",
        "namespace": [
          "user_id"
        ]
      },
      {
        "id": "summary_7hq08345",
        "enabled": true,
        "type": "SUMMARY",
        "namespace": [
          "agent_id"
        ]
      }
    ]
  }
}
```

### Related Issues

### Check List


common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
```diff
@@ -62,6 +79,42 @@ public ActionRequestValidationException validate() {
...
+        if (this.deleteAllMemories && this.deleteMemories != null && !this.deleteMemories.isEmpty()) {
+            exception = addValidationError(
+                "Cannot specify both delete_all_memories and delete_memories. Use either delete_all_memories=true OR delete_memories with specific types",
+                exception
+            );
+        }
+
+        // Validate memory types
+        if (this.deleteMemories != null && !this.deleteMemories.isEmpty()) {
+            List<String> validMemoryTypes = Arrays
```
ylwu-amzn: We have a todo to change this to enum. That could make the validation easier

common/src/test/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryContainerInputTests.java
```diff
@@ -0,0 +1,342 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.transport.memorycontainer.memory;
+
+import static org.junit.Assert.*;
```
ylwu-amzn: Avoid import *

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
```diff
@@ -165,20 +186,98 @@ private void handleDeleteResponse(
...
+                            case MEM_CONTAINER_MEMORY_TYPE_WORKING:
+                                indicesToDelete.add(configuration.getWorkingMemoryIndexName());
+                                break;
+                            case MEM_CONTAINER_MEMORY_TYPE_LONG_TERM:
+                                indicesToDelete.add(configuration.getLongMemoryIndexName());
+                                break;
+                            case MEM_CONTAINER_MEMORY_TYPE_HISTORY:
+                                indicesToDelete.add(configuration.getLongMemoryHistoryIndexName());
+                                break;
+                            default:
```
dhrubo-os: Shouldn't we throw bad reqeust/illegal argument exception in the first place before even coming here?

plugin/src/main/java/org/opensearch/ml/helper/StrategyMergeHelper.java
```diff
@@ -0,0 +1,112 @@
...
+                // Validate type not changed
+                if (update.getType() != null && !update.getType().equals(existingStrategy.getType())) {
+                    log
+                        .error(
+                            "Attempt to change strategy type from {} to {} for strategy {}",
+                            existingStrategy.getType(),
+                            update.getType(),
+                            update.getId()
+                        );
+                    throw new IllegalArgumentException(
```
dhrubo-os, line 74: I don't have much concern, but curious why do we need to have this restriction?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
```diff
@@ -29,20 +35,30 @@ public class MLMemoryContainerDeleteRequest extends ActionRequest {
...
         this.deleteAllMemories = deleteAllMemories;
+        this.deleteMemories = deleteMemories;
         this.tenantId = tenantId;
     }
 
     public MLMemoryContainerDeleteRequest(StreamInput input) throws IOException {
         super(input);
         this.memoryContainerId = input.readString();
         this.deleteAllMemories = input.readBoolean();
+        this.deleteMemories = input.readOptionalStringList();
```
dhrubo-os: can we put these in a set to avoid duplicates?

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoryContainerAction.java
```diff
@@ -62,11 +68,48 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+        boolean deleteAllMemories = request.paramAsBoolean(PARAMETER_DELETE_ALL_MEMORIES, false);
+        List<String> deleteMemories = null;
+        String[] memoryArray = request.paramAsStringArray(PARAMETER_DELETE_MEMORIES, Strings.EMPTY_ARRAY);
+        if (memoryArray != null && memoryArray.length > 0) {
+            deleteMemories = Arrays.asList(memoryArray);
+        }
+
+        // Parse request body if present (URL params take precedence)
+        if (request.hasContent()) {
+            try (XContentParser parser = request.contentParser()) {
```
dhrubo-os, line 86: no catch block for parsing exceptions?

plugin/src/main/java/org/opensearch/ml/helper/StrategyMergeHelper.java
b4sjoo: Because the strategy id itself is type+UUID, if we allow update strategy type then it will be a mismatch, which breaks our naming logic.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
b4sjoo: See the [validation](https://github.com/opensearch-project/ml-commons/pull/4258/files#diff-aea1a0490d3f3222a41887566b09b73cb564349258e4a40c327eebcfa15b1b8dR100) in REST layer, but there will still be issue if developer called api in transport layer directly. So here is still needed.

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoryContainerAction.java
b4sjoo: Why No Catch Is Needed

  1. Try-with-Resources Statement
  - This is a try-with-resources statement (introduced in Java 7)
  - The XContentParser implements AutoCloseable
  - The parser is automatically closed when exiting the try block
  - Pattern: try (Resource r = ...) { }

  2. Method Declares IOException
  - The method signature: prepareRequest(...) throws IOException
  - Any IOException from parser.contentParser() or parser.map() propagates up
  - The caller is responsible for handling exceptions

  3. No Recovery Logic Needed
  - If parsing fails, there's nothing to recover
  - The REST framework expects exceptions to propagate
  - OpenSearch's REST layer catches these and returns appropriate HTTP errors

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
b4sjoo: Nice catch, updated

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
```diff
@@ -94,7 +95,7 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
 
     public static MemoryStrategy parse(XContentParser parser) throws IOException {
         String id = null;
-        boolean enabled = true;  // Default to true
+        Boolean enabled = null;  // Null to detect if not provided
```
ylwu-amzn: Do we really need to track this ?

plugin/src/main/java/org/opensearch/ml/helper/StrategyHelper.java
```diff
@@ -0,0 +1,147 @@
...
+import org.opensearch.ml.common.memorycontainer.MemoryStrategy;
+
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * Helper class for memory strategy operations.
+ * Provides validation and merging logic for memory strategies.
+ */
+@Log4j2
+public class StrategyHelper {
```
ylwu-amzn: Refactoring Suggestions from AI:

1. StrategyHelper has too much responsibility - It's handling validation, merging logic, and field updates. Add update() method to MemoryStrategy

Move the field merging logic into MemoryStrategy itself:
```
/**
 * Updates this strategy with non-null values from another strategy.
 * Type cannot be changed.
 *
 * @param other The strategy containing updates
 * @throws IllegalArgumentException if attempting to change type
 */
public void update(MemoryStrategy other) {
    if (other.getType() != null && !other.getType().equals(this.type)) {
        throw new IllegalArgumentException(
            "Cannot change strategy type from " + this.type + " to " + other.getType() + " for strategy " + this.id
        );
    }
    
    if (other.getEnabled() != null) {
        this.enabled = other.getEnabled();
    }
    if (other.getNamespace() != null) {
        this.namespace = other.getNamespace();
    }
    if (other.getStrategyConfig() != null) {
        this.strategyConfig = other.getStrategyConfig();
    }
}
```

2. Move validation into MemoryStrategy

```
/**
 * Validates this strategy for required fields and valid values.
 *
 * @throws IllegalArgumentException if validation fails
 */
public void validate() {
    validateType(this.type);
    validateNamespace(this.namespace);
}

private static void validateType(String type) {
    if (type != null && 
        !("semantic".equalsIgnoreCase(type) || 
          "user_preference".equalsIgnoreCase(type) || 
          "summary".equalsIgnoreCase(type))) {
        throw new IllegalArgumentException(String.format(INVALID_STRATEGY_TYPE_ERROR, type));
    }
}

private static void validateNamespace(List<String> namespace) {
    if (namespace == null || namespace.isEmpty()) {
        throw new IllegalArgumentException("Strategy namespace is required. Please provide a non-empty namespace array.");
    }
}
```

b4sjoo: Will move validation into MemoryStrategy and rename `StrategyHelper` to `StrategyMergeHelper`

common/src/test/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryContainerInputTests.java
b4sjoo: Will change

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
b4sjoo: Think about a case the strategy has already been set to `false`, when update it the user does not declare the field. Then it will be set back to `true` and update to `true` in the end, which is not our expected result.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
b4sjoo: Changed to enum

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
ylwu-amzn: `type` is `String`, can we change to `MemoryStrategyType`?


```suggestion
                    MemoryStrategyType.fromString(parser.text());
```

With this , no need to add this check at line 161
```
        if (type != null
            && !("semantic".equalsIgnoreCase(type) || "user_preference".equalsIgnoreCase(type) || "summary".equalsIgnoreCase(type))) {
            throw new IllegalArgumentException(String.format(MemoryContainerConstants.INVALID_STRATEGY_TYPE_ERROR, type));
        }
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLGetMemoryResponse.java
```diff
@@ -102,23 +99,28 @@ public void writeTo(StreamOutput out) throws IOException {
         }
     }
 
-    public static MLGetMemoryResponse fromGetResponse(GetResponse getResponse, String memoryType) {
+    public static MLGetMemoryResponse fromGetResponse(GetResponse getResponse, String memoryTypeStr) {
+        MemoryType memoryType = MemoryType.fromString(memoryTypeStr);
+        if (memoryType == null) {
+            throw new IllegalArgumentException("Invalid memory type: " + memoryTypeStr);
+        }
```
ylwu-amzn: Can we move this check to `MemoryType.fromString` ?

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoryContainerAction.java
```diff
@@ -62,11 +69,48 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
         }
 
         String memoryContainerId = request.param(PARAMETER_MEMORY_CONTAINER_ID);
-        boolean deleteAllMemories = request.paramAsBoolean(PARAMETER_DELETE_ALL_MEMORIES, false);
         String tenantId = getTenantID(mlFeatureEnabledSetting.isMultiTenancyEnabled(), request);
+
+        // Parse URL parameters
+        boolean deleteAllMemories = request.paramAsBoolean(PARAMETER_DELETE_ALL_MEMORIES, false);
+        Set<String> deleteMemories = null;
```
ylwu-amzn: Use MemoryType enum ? 
```suggestion
        Set<MemoryType> deleteMemories = null;
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryRequest.java
ylwu-amzn:
```suggestion
    private MemoryType memoryType;
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
```diff
@@ -29,20 +33,31 @@ public class MLMemoryContainerDeleteRequest extends ActionRequest {
     @Getter
     boolean deleteAllMemories;
 
+    @Getter
+    Set<String> deleteMemories;
```
ylwu-amzn:
```suggestion
    Set<MemoryType> deleteMemories;
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
```diff
@@ -130,13 +131,44 @@ public static MemoryStrategy parse(XContentParser parser) throws IOException {
...
+        return enabled == null || enabled;
+    }
+
+    /**
+     * Validates a memory strategy for required fields and valid values.
+     *
+     * @param strategy The memory strategy to validate
+     * @throws IllegalArgumentException if validation fails with specific error message
+     */
+    public static void validate(MemoryStrategy strategy) {
```
ylwu-amzn, line 184: This validation isn't automatically performed in constructor method, meaning invalid objects could be created if the calling code doesn't explicitly call the validate() method.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
```diff
@@ -95,25 +100,71 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Delete
...
+                actionListener.onFailure(new OpenSearchStatusException("Only container owner can delete container", RestStatus.FORBIDDEN));
                 return;
             }
 
-            // Delete memory container
-            deleteMemoryContainer(memoryContainerId, container, tenantId, deleteAllMemories, actionListener);
-        }, actionListener::onFailure));
+            // CRITICAL: Check for shared index prefix BEFORE deleting container
+            boolean willDeleteIndices = deleteRequest.isDeleteAllMemories()
+                || (deleteRequest.getDeleteMemories() != null && !deleteRequest.getDeleteMemories().isEmpty());
```
ylwu-amzn: minor

```suggestion
                || !CollectionUtils.isEmpty(deleteRequest.getDeleteMemories()));
```

plugin/src/main/java/org/opensearch/ml/helper/StrategyMergeHelper.java
```diff
@@ -0,0 +1,121 @@
...
+        if (updates == null || updates.isEmpty()) {
+            return existing;
+        }
+
+        // Create map of existing strategies by ID for efficient lookup
+        Map<String, MemoryStrategy> strategyMap = existing.stream().collect(Collectors.toMap(MemoryStrategy::getId, Function.identity()));
+
+        // Process each update
+        for (MemoryStrategy update : updates) {
+            if (update.getId() != null && !update.getId().trim().isEmpty()) {
```
ylwu-amzn:
```suggestion
            if (update.getId() != null && !update.getId().isBlank()) {
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
b4sjoo: Refactored and removed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryStrategy.java
b4sjoo: Use strategytype enum to do the string validation. Here just focus on field validation.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLGetMemoryResponse.java
b4sjoo: Moved, with other class checked.

plugin/src/main/java/org/opensearch/ml/rest/RestMLDeleteMemoryContainerAction.java
b4sjoo: Refactored

plugin/src/main/java/org/opensearch/ml/helper/StrategyMergeHelper.java
b4sjoo: Done

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLUpdateMemoryRequest.java
b4sjoo: Refactored

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/MLMemoryContainerDeleteRequest.java
b4sjoo: Refactored

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
b4sjoo: Done

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLGetMemoryResponse.java
b4sjoo: Moved

ylwu-amzn (approved):

PR #4262, opened 2025-10-06
[Agentic Search]Use same model for Agent and QPT

rithinpullela:
### Description
This change allows using the same model ID for the agent and the Query Planner Tool.
Added a function to remove the unnecessary interactions and tool config parameters to keep the QPT response unaffected 

**Register Model:**
This can be used by both QPT and Agent
```
{
    "name": "My OpenAI model: gpt-5",
    "function_name": "remote",
    "description": "test model",
    "connector": {
        "name": "My openai connector: gpt-5",
        "description": "The connector to openai chat model",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "model": "gpt-5"
        },
        "credential": {
            "openAI_key": "YOUR API KEY"    
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
                "url": "https://api.openai.com/v1/chat/completions",
                "headers": {
                },
                "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": [{\"role\":\"developer\",\"content\":\"${parameters.system_prompt}\"},${parameters._chat_history:-}{\"role\":\"user\",\"content\":\"${parameters.user_prompt}\"}${parameters._interactions:-}], \"reasoning_effort\":\"low\"${parameters.tool_configs:-}}"
            }
        ]
    }
}
```

For claude Models:
```
{
    "name": "My Claude 3.7 model",
    "function_name": "remote",
    "description": "test model",
    "connector": {
        "name": "Amazon Bedrock Claude 3.7-sonnet connector",
        "description": "Connector to Amazon Bedrock service for the Claude model",
        "version": 1,
        "protocol": "aws_sigv4",
        "parameters": {
            "region": "us-east-1",
            "service_name": "bedrock",
            "model": "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
        },
        "credential": {
            "access_key": "access key",
            "secret_key": "secret key",
            "session_token": <Session token>
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
                "url": "https://bedrock-runtime.${parameters.region}.amazonaws.com/model/${parameters.model}/converse",
                "headers": {
                },
                "request_body": "{ \"system\": [{\"text\": \"${parameters.system_prompt}\"}], \"messages\": [${parameters._chat_history:-}{\"role\":\"user\",\"content\":[{\"text\":\"${parameters.user_prompt}\"}]}${parameters._interactions:-}]${parameters.tool_configs:-} }"
            }
        ]
    }
}
```

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
@@ -92,7 +92,7 @@ public class TransportRegisterAgentAction extends HandledTransportAction<ActionR
         }
 
         // Update QueryPlanningTool to include model_id if missing
-        List<MLToolSpec> updatedTools = processQueryPlanenrTools(agent);
+        List<MLToolSpec> updatedTools = processQueryPlannerTools(agent);
 
         String llmInterface = (agent.getParameters() != null) ? agent.getParameters().get(LLM_INTERFACE) : null;
         if (llmInterface != null) {
@@ -131,7 +131,7 @@ public class TransportRegisterAgentAction extends HandledTransportAction<ActionR
         }
     }
 
-    private List<MLToolSpec> processQueryPlanenrTools(MLAgent agent) {
+    private List<MLToolSpec> processQueryPlannerTools(MLAgent agent) {
         List<MLToolSpec> tools = agent.getTools();
         List<MLToolSpec> updatedTools = tools;
         if (tools != null) {


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -149,8 +168,19 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
 
             // Template Selection, replace user and system prompts
             Map<String, String> templateSelectionParameters = new HashMap<>(parameters);
-            templateSelectionParameters.put(SYSTEM_PROMPT_FIELD, TEMPLATE_SELECTION_SYSTEM_PROMPT);
-            templateSelectionParameters.put(USER_PROMPT_FIELD, TEMPLATE_SELECTION_USER_PROMPT);
+            if (templateSelectionParameters.containsKey(TEMPLATE_SELECTION_SYSTEM_PROMPT_FIELD)) {
+                templateSelectionParameters
+                    .put(SYSTEM_PROMPT_FIELD, templateSelectionParameters.get(TEMPLATE_SELECTION_SYSTEM_PROMPT_FIELD));
```
joshpalis: Why should we allow the template selection prompt to be configurable by the user?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -377,7 +412,7 @@ public static Factory getInstance() {
 
         public void init(Client client, MLFeatureEnabledSetting mlFeatureEnabledSetting) {
             this.client = client;
-            this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+            Factory.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
```
owaiskazi19: Why this change? We can remove this completely now as we have flipped the flag

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -121,10 +127,23 @@ public QueryPlanningTool(String generationType, MLModelTool queryGenerationTool,
         this.attributes = new HashMap<>(DEFAULT_ATTRIBUTES);
     }
 
+    private Map<String, String> stripAgentContextParameters(Map<String, String> originalParameters) {
+        // Drop agent-specific metadata that can bias or slow query planning; keep all other non-null params.
+        // This enables using the same LLM for both the agent and the Query Planning Tool.
+        // Excluded keys: _chat_history, _tools, _interactions, tool_configs
+        Set<String> blacklistedParams = Set.of("_chat_history", "_tools", "_interactions", "tool_configs");
```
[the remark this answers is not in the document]
rithinpullela: When we use the same model for Agent and Query Planner, these params will be filled with agent's context. I am making a copy of the parameters after removing these agent context specific params.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: There was a small linting error which suggested to use the class name when accessing the static variable.

I would prefer to have this setting because people can opt out of Agentic Search by flipping the flag to false.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: This can be configured by the users similar to what we are doinf with the conversational agent prompt and Query Planner prompt.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
owaiskazi19: The setting for agentic search has been removed from neural search. If users don't register a QPT tool then they won't have the setting. Will not block PR for this but we can remove it in the future PR

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -89,19 +90,33 @@ private void registerAgent(MLAgent agent, ActionListener<MLRegisterAgentResponse
...
+        List<MLToolSpec> updatedTools = tools;
         if (tools != null) {
-            for (MLToolSpec tool : tools) {
-                if (tool.getType().equals(QueryPlanningTool.TYPE) && !mlFeatureEnabledSetting.isAgenticSearchEnabled()) {
-                    listener.onFailure(new OpenSearchException(ML_COMMONS_AGENTIC_SEARCH_DISABLED_MESSAGE));
-                    return;
-                }
+            // Check if agentic search is enabled when QueryPlanningTool is present
+            boolean hasQueryPlanningTool = tools.stream().anyMatch(tool -> tool.getType().equals(QueryPlanningTool.TYPE));
+            if (hasQueryPlanningTool && !mlFeatureEnabledSetting.isAgenticSearchEnabled()) {
```
mingshl: we already removed the feature flag in 3.3 right?

mingshl: 1. if we are GA in 3.3, please remove the feature flag

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -89,19 +90,33 @@ private void registerAgent(MLAgent agent, ActionListener<MLRegisterAgentResponse
...
+        // Update QueryPlanningTool to include model_id if missing
         List<MLToolSpec> tools = agent.getTools();
+        List<MLToolSpec> updatedTools = tools;
         if (tools != null) {
-            for (MLToolSpec tool : tools) {
-                if (tool.getType().equals(QueryPlanningTool.TYPE) && !mlFeatureEnabledSetting.isAgenticSearchEnabled()) {
-                    listener.onFailure(new OpenSearchException(ML_COMMONS_AGENTIC_SEARCH_DISABLED_MESSAGE));
-                    return;
-                }
+            // Check if agentic search is enabled when QueryPlanningTool is present
```
mingshl: I am thinking these additional checks are not necessary,
we could do first load the agent.getLlm().getModelId(), 
and allow toolSpec to override the model_id, then it can be used in all tools? 

I am worried later I have other tools, then I am adding more and more types check of tool to load the model id.

mingshl: @pyek-bot is it possible: any sweet spot that I can load the model id first then it call tool_configs, I also don't want to iterate through the toolconfigs and modify the tool config?

rithinpullela: +1, the tool parameters can be overridden by execute params. Not sure if that is the ideal thing to do.

Perhaps we could have some fixed params during registration which cannot be overridden later. Could take this as an improvement.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
```diff
@@ -121,6 +114,25 @@ private void registerAgent(MLAgent agent, ActionListener<MLRegisterAgentResponse
         }
     }
 
+    private List<MLToolSpec> processQueryPlanenrTools(MLAgent agent) {
```
owaiskazi19: You can check if the tool type is not `QueryPlanningTool` and return early with tools

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
rithinpullela: Created an issue for this: https://github.com/opensearch-project/ml-commons/issues/4267

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
rithinpullela: ML Commons allows multiple tools of the same type to be registered. So we need to check each tool

owaiskazi19: Would we ever have 2 QPT tool? Since the job of the tool is to convert NLQ to DSL

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -311,12 +311,6 @@ private MLCommonsSettings() {}
     public static final Setting<Boolean> ML_COMMONS_MEMORY_FEATURE_ENABLED = Setting
         .boolSetting(ML_PLUGIN_SETTING_PREFIX + "memory_feature_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
 
-    public static final Setting<Boolean> ML_COMMONS_AGENTIC_SEARCH_ENABLED = Setting
```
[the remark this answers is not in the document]
owaiskazi19: It was experimental in 3.2, so we can remove the settings when going GA. This is expected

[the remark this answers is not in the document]
dhrubo-os: But what if somebody doesn't want agentic search? how can they turn it off?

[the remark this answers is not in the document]
mingshl: For MLCommons, it's just using query planning tool. If users doesn't want to use a tool, just don't configure it. It's fine

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
mingshl: I am fine if we can refactor this in a better structure later

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
dhrubo-os: In my understanding, in experimental feature, we keep the settings off (by default) so that if customer wants they can enable the feature. When we go on GA, we enable that feature by default. But we still keep the settings so that if customer wants they can turn it off.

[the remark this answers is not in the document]
mingshl: If that's the case to always keep feature flag after GA, how many feature flags the customer will have?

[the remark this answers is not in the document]
dhrubo-os: a flag per feature: https://docs.opensearch.org/latest/ml-commons-plugin/cluster-settings/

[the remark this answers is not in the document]
owaiskazi19: But this setting was to use a tool and not complete feature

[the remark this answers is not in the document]
rithinpullela: While I am not sure what is the right way of handling the flag, at the end of the day, what we are exposing from ML Commons side is the QueryGeneration Tool. The feature Agentic Search uses this tool, but the main logic behind the feature is controlled from neural search's end. 

If users want to they can use the Query Planner Tool with a normal agent to help in better query generation.

Hence I think it is okay to remove the flag from ML Commons.

plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java
rithinpullela: 1. Since it is a list, and not a heavy operation, the resources and time consumed will be minimal. 
2. While it does not make sense to have 2 tools of same kind, since it is possible for users to register them, we need to cover that use case as well. Perhaps they might have different templates attached to it and they might call them `index1_query_planner_tool` and `index2_query_planner_tool`

ylwu-amzn (approved):

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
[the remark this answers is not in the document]
mingshl: IMO, when a feature goes GA, it means `general available`, which literally means everyone can access this feature, it doesn't make sense to keep the feature flag after general available and keep option to turn it off.  I am cool with this change in this PR

PR #4278, opened 2025-10-07
fix json parsing error by extracing json first; add for each processor; support input processor

ylwu-amzn:
### Description
1. Fix json parsing error when extract long term memory. The LLM output could be invalid json like `"This is the answer {\"facts\": [...]}"`. Use extract json processor to extract facts json
2. Add for each processor to better support processing each item in array
3. Support input processor for model

Tried to test Strands + Agentic Memory for long term memory, find several issues
1. LLM result not valid json. We can solve it by using extract json processor
2. Strands message not valid , the message has no `"type":"text"` by default, so have to use a for each processor to add this. Otherwise Bedrock Claude invoke API will throw exception "messages.0.content.0.type: Field required" . Need to configure for each processor as model's input processor.
### Related Issues

### Check List


mingshl:
do you mean this is the strands agent response?  LLM response shouldn't be a string.


> The LLM output could be invalid json like "This is the answer {\"facts\": [...]}". Use extract json processor to extract facts json

mingshl:
If you testing using strands agent, I am assuming you are trying store the memory from agent response.

However, I think it might make more sense for users to store memory, no matter short term or long term, using hooks, and if I was the person trying to implement strands agent and store memory, I would use this hook, 
`MessageAddedEvent` , this is triggered when a message is added to the agent's conversation history. And in here, you should be to extract the message from json from strands agent side, I don't think we should do this parsing on agentic memory side. 

This is my personal take

ylwu-amzn:
> do you mean this is the strands agent response? LLM response shouldn't a string.
> 
> > The LLM output could be invalid json like "This is the answer {"facts": [...]}". Use extract json processor to extract facts json

When we ingest message to agentic memory, we will extract long term memory with LLM. The LLM output here means that extracted long term memory. Ideally it should be `{"facts": [...]}`, but LLM may add some description words before or after it which makes the model output not valid json

ylwu-amzn:
> If you testing using strands agent, I am assuming you are trying store the memory from agent response.
> 
> However, I think it might make more sense for users to store memory, no matter short term or long term, using hooks, and if I was the person trying to implement strands agent and store memory, I would use this hook, `MessageAddedEvent` , this is triggered when a message is added to the agent's conversation history. And in here, you should be to extract the message from json from strands agent side, I don't think we should do this parsing on agentic memory side.
> 
> This is my personal take

People could have different options. This is definitely one option. I'm using Agentic Memory to store short term memory. Then the Agentic Memory will extract long term memory automatically.  We can discuss this separately.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -242,6 +244,13 @@ default void preparePayloadAndInvoke(
         // override again to always prioritize the input parameter
         parameters.putAll(inputParameters);
         String payload = connector.createPayload(action, parameters);
+
+        List<Map<String, Object>> processorConfigs = ProcessorChain.extractProcessorConfigs(parameters, INPUT_PROCESSORS);
+        if (!processorConfigs.isEmpty()) {
+            ProcessorChain processorChain = new ProcessorChain(processorConfigs);
+            payload = StringUtils.toJson(processorChain.process(payload));
+        }
+
```
Zhangxunmt, line 253: This seems creating new ML processor instances and chains for each request since this is in the model invoke path. I think we can use caching processor chains to reduce overheads.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -307,6 +309,10 @@ private List<String> parseFactsFromLLMResponse(MemoryStrategy strategy, MLOutput
             return facts;
         }
 
+        List<Map<String, Object>> processorConfigs = new ArrayList<>();
+        processorConfigs.add(Map.of("type", MLProcessorType.EXTRACT_JSON.getValue(), "extract_type", "object"));
+        ProcessorChain processorChain = new ProcessorChain(processorConfigs);
```
Zhangxunmt: If this is the default/required processorChain, why not initialize it as the class member and use it directly. I think this processChain doesn't contain any memory data so there's no need to create a new ProcessorChain for each memory request.

ylwu-amzn:
> Does it require the customer to provide the process config in the add/get memory request? I haven't found in the PR where the processor config is defined to create the processorChain to add the missing "type":"text".

User need to change the model configuration by adding processors. We should not hard code the change in code.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
ylwu-amzn: Good point, will change this

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
ylwu-amzn: Hard to cache it, as this is dynamically created for each model request. The processors will be defined in parameters . Example
```
{
  "parameters": {
    "system_prompt": "you are a helpful assistant",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "text": "Can you find all error log indices in my cluster",
            "type": "text"
          }
        ]
      }
    ],
    "input_processors": [
      {
        "type":"remove_jsonpath",
        "paths": ["$..content[?(@.reasoning)]"]
      }
    ]
  }
}
```

Zhangxunmt (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/processor/MLForEachProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+                } catch (Exception e) {
+                    log.warn("Failed to process element at index {} in path '{}': {}", i, path, e.getMessage());
+                    // Keep original element on error
+                    processedArray.add(element);
+                }
+            }
+
+            // Set the processed array back
+            // Remove [*] wildcards from path for setting
+            String setPath = path.replaceAll("\\[\\*\\]$", "");
```
mingshl, line 208: will this work if the path has `[*]` in the middle? for example, `text_chunks[*].text_chunk`. it seems this regex pattern might only takes the trailing `[*]`. 

do you want to remove all `[*]` in the path? maybe we can do  
`String setPath = path.replaceAll("\\[\\*\\]", "");`

mingshl: oh I saw your test cases, testExtractFactsFromConversation_WithProcessorChainExtractingJson which is parsing `$.content[0].text`, should be fine. approved

PR #4283, opened 2025-10-08
fix llm result path; convert message to user prompt string

ylwu-amzn:
### Description

### Related Issues

### Check List


plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -138,32 +138,9 @@ public void extractFactsFromConversation(
...
-                message.toXContent(messagesBuilder, ToXContent.EMPTY_PARAMS);
-            }
-            messagesBuilder.endArray();
-            String messagesJson = messagesBuilder.toString();
-            stringParameters.put("messages", messagesJson);
-
-            log.debug("LLM request - processing {} messages", messages.size());
+            String conversationJson = serializeMessagesToJson(messages);
+            String userPrompt = "Analyze the following conversation and extract information:\n```json\n" + conversationJson + "\n```";
+            stringParameters.put("user_prompt", userPrompt);
```
dhrubo-os, line 153: So we aren't expecting prompts for strategies from user?

ylwu-amzn: User could provide their custom `system_prompt`, not `user_prompt`.

`user_prompt` is generated with the raw messages

ylwu-amzn: That could be another options. Choose to serialize the messages to json directly is the easiest way. If we parse it into custom format, need to consider different LLM's message format and build specific parsing logic.

b4sjoo:
Hmm I have some concerns on this change. 1. If we want to support a real multi-modal later this will be an blocker cause under this way all the thing is wrapped as a text 2. If you really want to wrapping all the thing in one-round conversation, then why don't we wrapped it behind the backend code? In such way the user interface won't change.

dhrubo-os:
@ylwu-amzn could you please resolve the conflict?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
dhrubo-os: Currently I was using this [JSON_ENFORCEMENT_MESSAGE](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java#L175-L178) to ensure the output format and it is working. Tested with claude and gpt mini model

If we are adding user prompt, may be we can add that JSON_ENFORCEMENT_MESSAGE in the user prompt instead?

ylwu-amzn: Line 148 already added that 
```
messages.add(enforcementMessage);
```

dhrubo-os: Sorry for the confusion. I was wondering if we can use `JSON_ENFORCEMENT_MESSAGE` as `user_prompt`?

But we can try to optimize it later by doing more investigation.

PR #4313, opened 2025-10-14
fix: prevent NPE and status inconsistency in index insight

shuangli-z:
### Description
Minor fixes for index insight feature:
1. Save lastUpdatedTime when task is failed to prevent NullPointerException during IndexInsight serialization
2. Ensure insight status resets to FAILED instead of remaining GENERATING on error
3. Refine error messages and remove redundant logging

### Related Issues

### Check List

### DIFF
--- a/common/src/test/java/org/opensearch/ml/common/indexInsight/FieldDescriptionTaskTests.java
+++ b/common/src/test/java/org/opensearch/ml/common/indexInsight/FieldDescriptionTaskTests.java
@@ -142,6 +142,7 @@ public class FieldDescriptionTaskTests {
         String statisticalContent = "{\"mapping\": {\"field1\": {\"type\": \"text\"}}}";
 
         mockGetSuccess(sdkClient, statisticalContent);
+        mockUpdateSuccess(sdkClient);
         mockMLConfigFailure(client, "Config not found");
 
         task.runTask("tenant-id", listener);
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/opensearch/ml/common/indexInsight/AbstractIndexInsightTask.java
```diff
@@ -203,9 +203,11 @@ protected void beginGeneration(String tenantId, ActionListener<IndexInsight> lis
             .lastUpdatedTime(Instant.now())
             .build();
 
-        writeIndexInsight(indexInsight, tenantId, ActionListener.wrap(r -> { runWithPrerequisites(tenantId, listener); }, e -> {
-            saveFailedStatus(tenantId, e, listener);
-        }));
+        writeIndexInsight(
+            indexInsight,
+            tenantId,
+            ActionListener.wrap(r -> { runWithPrerequisites(tenantId, listener); }, listener::onFailure)
```
xinyual, line 209: Why in here we don't save the failure status?

shuangli-z: Because if writing the initial GENERATING status fails, the insight will not be stored, so there is no need to update to FAILED status. Also, attempting to save a FAILED status would likely fail for the same reason.

xinyual:
Close as already fixed in the PR https://github.com/opensearch-project/ml-commons/pull/4449

PR #4332, opened 2025-10-20
Add name validation to index prefix

b4sjoo:
### Description
Add name validation to index prefix and remove sensitive logs

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
@@ -353,7 +353,7 @@ public class MemoryProcessingService {
                     "LLM predict result cannot be extracted with current llm_result_path with reason: "
                         + reason
                         + ". Please check either your llm configuration or your llm_result_path setting in memory container configuration",
-                    RestStatus.NOT_FOUND
+                    RestStatus.BAD_REQUEST
                 );
             }
         }
@@ -417,7 +417,7 @@ public class MemoryProcessingService {
                     "LLM predict result cannot be extracted with current llm_result_path with reason: "
                         + reason
                         + ". Please check either your llm configuration or your llm_result_path setting in memory container configuration",
-                    RestStatus.NOT_FOUND
+                    RestStatus.BAD_REQUEST
                 );
             }
         } catch (Exception e) {
@@ -507,7 +507,7 @@ public class MemoryProcessingService {
                 "LLM predict result cannot be extracted with current llm_result_path with reason: "
                     + reason
                     + ". Please check either your llm configuration or your llm_result_path setting in memory container configuration",
-                RestStatus.NOT_FOUND
+                RestStatus.BAD_REQUEST
             );
         }
     }
--- a/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingServiceTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingServiceTests.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.ml.action.memorycontainer.memory;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
@@ -30,6 +31,7 @@ import java.util.Map;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 import org.opensearch.OpenSearchStatusException;
@@ -1109,7 +1111,7 @@ public class MemoryProcessingServiceTests {
         MemoryConfiguration storageConfig = mock(MemoryConfiguration.class);
         when(storageConfig.getLlmId()).thenReturn("llm-model-123");
 
-        // Mock LLM response with path that will trigger NOT_FOUND (404)
+        // Mock LLM response with path that will trigger BAD_REQUEST (400)
         MLTaskResponse mockResponse = mock(MLTaskResponse.class);
         ModelTensorOutput mockOutput = mock(ModelTensorOutput.class);
         ModelTensors mockTensors = mock(ModelTensors.class);
@@ -1132,12 +1134,12 @@ public class MemoryProcessingServiceTests {
 
         memoryProcessingService.extractFactsFromConversation(messages, memoryStrategy, storageConfig, factsListener);
 
-        // Verify that 4XX error (NOT_FOUND) is preserved with its detailed message
+        // Verify that 4XX error (BAD_REQUEST) is preserved with its detailed message
         verify(factsListener)
             .onFailure(
                 argThat(
                     exception -> exception instanceof OpenSearchStatusException
-                        && ((OpenSearchStatusException) exception).status() == RestStatus.NOT_FOUND
+                        && ((OpenSearchStatusException) exception).status() == RestStatus.BAD_REQUEST
                         && exception.getMessage().contains("LLM predict result cannot be extracted")
                         && exception.getMessage().contains("llm_result_path")
                 )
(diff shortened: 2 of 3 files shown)


ylwu-amzn (dismissed):

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -410,11 +410,11 @@ public void summarizeMessages(MemoryConfiguration configuration, List<MessageInp
                         String summary = parseSessionSummary((ModelTensorOutput) response.getOutput(), llmResultPath);
                         listener.onResponse(summary);
                     } catch (Exception e) {
-                        log.error("Failed to parse memory decisions from LLM response", e);
+                        log.error("Failed to parse memory decisions from LLM response");
```
dhrubo-os: Are you sure security was complaining about logging error message? I thought their only concern was to log the fact data?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -133,6 +136,14 @@ private String buildIndexPrefix(String indexPrefix, boolean useSystemIndex) {
                 ? DEFAULT_MEMORY_INDEX_PREFIX
                 : UUID.randomUUID().toString().replace("-", "").substring(0, 8).toLowerCase();
         }
+        if (indexPrefix.indexOf('\r') >= 0 || indexPrefix.indexOf('\n') >= 0 || indexPrefix.chars().anyMatch(ch -> ch < 32)) {
```
dhrubo-os, line 139: Can we follow these restrictions: https://docs.opensearch.org/latest/api-reference/index-apis/create-index/?utm_source=chatgpt.com#index-naming-restrictions

b4sjoo: I have already added this validation by importing the metadata method. However we still need this line because that one does not address the CRLF issue.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
b4sjoo: Yes there's one concern on that.

Zhangxunmt (dismissed):

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -133,6 +136,14 @@ private String buildIndexPrefix(String indexPrefix, boolean useSystemIndex) {
...
+            throw new OpenSearchParseException(INDEX_PREFIX_INVALID_CHARACTERS_ERROR);
```
dhrubo-os, line 140: Can we throw OpensearchStatusException with 400 status code? As this should be treated as 4xx level error?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -133,6 +136,14 @@ private String buildIndexPrefix(String indexPrefix, boolean useSystemIndex) {
...
+        }
+        MetadataCreateIndexService
+            .validateIndexOrAliasName(
+                indexPrefix,
+                (s1, s2) -> new OpenSearchParseException("missing or invalid index prefix [" + s1 + "] " + s2)
```
dhrubo-os, line 145: Same here.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -425,12 +452,22 @@ public void summarizeMessages(MemoryConfiguration configuration, List<MessageInp
...
-            sessionSummary = StringUtils.toJson(filterdResult);
+        try {
+            Object filterdResult = JsonPath.read(dataAsMap, llmResultPath);
+            String sessionSummary = null;
+            if (filterdResult != null) {
+                sessionSummary = StringUtils.toJson(filterdResult);
+            }
+            return sessionSummary;
+        } catch (PathNotFoundException e) {
+            String reason = extractFirstSentence(e.getMessage());
```
dhrubo-os, line 504: Can we also log the reason as well?

b4sjoo: Makes sense, added

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
b4sjoo: OpenSearchParseException is 400 Bad request, I verified

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
b4sjoo: Same

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
b4sjoo: Reverted some of those which seems won't include sensitive info

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
dhrubo-os: Oh yeah, that's true. Thanks for checking.

dhrubo-os (dismissed):

ylwu-amzn (dismissed):

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -176,12 +178,21 @@ public void extractFactsFromConversation(
                 log.debug("Extracted {} facts from LLM response", facts.size());
                 listener.onResponse(facts);
             } catch (Exception e) {
+                // Preserve client errors (4XX) with their detailed messages
+                if (e instanceof OpenSearchStatusException) {
+                    OpenSearchStatusException osException = (OpenSearchStatusException) e;
+                    if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
```
ylwu-amzn, line 184: Why need to add this checking? If the exception is OpenSearchStatusException, should we just return this to client side ? 

For other errors which not in [400, 500] range, why need to wrap it to another `OpenSearchStatusException` ? The exception itself is already `OpenSearchStatusException`, so the wrapping just change the error code to 500?
```
listener.onFailure(new OpenSearchStatusException("Internal server error", RestStatus.INTERNAL_SERVER_ERROR));
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -297,31 +317,44 @@ private List<String> parseFactsFromLLMResponse(MemoryStrategy strategy, MemoryCo
...
-                    throw new IllegalArgumentException("Failed to extract content from LLM response", e);
                 }
+            } catch (PathNotFoundException e) {
+                String reason = extractFirstSentence(e.getMessage());
+                log.error("Failed to extract LLM result using path {}: {}", llmResultPath, reason);
+                throw new OpenSearchStatusException(
+                    "LLM predict result cannot be extracted with current llm_result_path with reason: "
+                        + reason
+                        + ". Please check either your llm configuration or your llm_result_path setting in memory container configuration",
+                    RestStatus.NOT_FOUND
```
ylwu-amzn: AI suggests 400 BAD_REQUEST

b4sjoo: Sure can change to BAD_REQUEST

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
ylwu-amzn: Got the point, this is to hide detail error message, just return `Internal server error`

ylwu-amzn (approved):

PR #4336, opened 2025-10-22
Fixing regex bypass issue

zane-neo:
### Description
Fix connector regex bypass issue

### Related Issues

### Check List

### DIFF
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -5,8 +5,8 @@ on:
   push:
     branches:
       - main
-      - 1.*
-      - 2.*
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.x'
 
 jobs:
   build-and-publish-snapshots:
--- a/build.gradle
+++ b/build.gradle
@@ -96,6 +96,15 @@ subprojects {
         // Force spotless depending on newer version of guava due to CVE-2023-2976. Remove after spotless upgrades.
         resolutionStrategy.force "com.google.guava:guava:32.1.3-jre"
         resolutionStrategy.force 'org.apache.commons:commons-compress:1.26.0'
+        resolutionStrategy.force "io.netty:netty-buffer:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-codec:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-codec-http:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-codec-http2:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-common:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-handler:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-resolver:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-transport:${versions.netty}"
+        resolutionStrategy.force "io.netty:netty-transport-native-unix-common:${versions.netty}"
     }
 
     apply plugin: 'com.diffplug.spotless'


common/src/main/java/org/opensearch/ml/common/httpclient/MLHttpClientFactory.java
```diff
@@ -5,96 +5,39 @@
...
-        // 127.0.0.1, 10.x.x.x, 172.16-31.x.x, 192.168.x.x, 169.254.x.x
-        return (first == 10)
-            || (first == 172 && second >= 16 && second <= 31)
-            || (first == 192 && second == 168)
-            || (first == 169 && second == 254);
+    public static SdkAsyncHttpClient getAsyncHttpClient(
+        Duration connectionTimeout,
+        Duration readTimeout,
+        int maxConnections,
+        AtomicBoolean connectorPrivateIpEnabled
```
Hailong-am: can we use bool here? what's the reason to use atomicBoolean

common/src/main/java/org/opensearch/ml/common/httpclient/ValidatingHttpClient.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Locale;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import lombok.extern.log4j.Log4j2;
+import software.amazon.awssdk.http.async.AsyncExecuteRequest;
+import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+
+@Log4j2
+public class ValidatingHttpClient implements SdkAsyncHttpClient {
```
ylwu-amzn: Suggest use a better class name.

rithin-pullela-aws (approved):
Other than the comment from @ylwu-amzn, looks good to me!

common/src/main/java/org/opensearch/ml/common/httpclient/MLHttpClientFactory.java
zane-neo: Agree that boolean is better, I tried to change all the occurrence of this `connectorPrivateIpEnabled` to boolean all the way back to where it's defined, but there's more UTs need to be changed. But changing to boolean is correct direction, let me try again this change.

ylwu-amzn:
let's backport to 3.3

common/src/main/java/org/opensearch/ml/common/httpclient/MLValidatableAsyncHttpClient.java
```diff
@@ -0,0 +1,110 @@
...
+     * @throws UnknownHostException Allow to use private IP or not.
+     */
+    public void validate(String protocol, String host, int port, boolean connectorPrivateIpEnabled) throws UnknownHostException {
+        if (protocol != null && !"http".equalsIgnoreCase(protocol) && !"https".equalsIgnoreCase(protocol)) {
+            log.error("Remote inference protocol is not http or https: {}", protocol);
+            throw new IllegalArgumentException("Protocol is not http or https: " + protocol);
+        }
+        // When port is not specified, the default port is -1, and we need to set it to 80 or 443 based on protocol.
+        if (port == -1) {
+            if (protocol == null || "http".equals(protocol.toLowerCase(Locale.getDefault()))) {
```
xinyual, line 63: Is there any possibility we are not in English region so Locale.getDefault() could be wrong? Is it better to use Locale.Root?

zane-neo: It's not a big issue, only when a local language has different encoding format for `http` four characters, then this could be an issue, but it doesn't seem a language like this. Anyway I can change to Locale.ROOT in next PR.

PR #4356, opened 2025-10-28
[Agentic Search]Add extract JSON processor in Query Planning Tool

rithinpullela:
### Description
Adds extract JSON processor so when a model produces extra text around the generated DSL, it can be ignored.

Example:
if model produces:
```
"Here is your query: {\"query\":{\"match\":{\"title\":\"test\"}}}"
```

The actual JSON will be extracted:
```
{\"query\":{\"match\":{\"title\":\"test\"}}}
```

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -253,6 +253,7 @@ public class QueryPlanningTool implements WithModelTool {
                         try {
                             String queryString = (String) r;
                             if (queryString == null || queryString.isBlank() || queryString.equals("null")) {
+                                log.debug("Model failed to generate the DSL query, returning the Default match all query");
                                 StringSubstitutor substitutor = new StringSubstitutor(parameters, "${parameters.", "}");
                                 String defaultQueryString = substitutor.replace(DEFAULT_QUERY);
                                 listener.onResponse((T) defaultQueryString);
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/processor/MLExtractJsonProcessor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/processor/MLExtractJsonProcessor.java
@@ -117,14 +117,22 @@ public class MLExtractJsonProcessor extends AbstractMLProcessor {
 
         String text = (String) input;
         if (text.trim().isEmpty()) {
-            return defaultValue != null ? defaultValue : input;
+            if (defaultValue != null) {
+                log.warn("Input text is empty, returning default value");
+                return defaultValue;
+            }
+            return input;
         }
 
         try {
             int start = findJsonStart(text);
             if (start < 0) {
+                if (defaultValue != null) {
+                    log.warn("No JSON found in input text, returning default value");
+                    return defaultValue;
+                }
                 log.debug("No JSON found in text");
-                return defaultValue != null ? defaultValue : input;
+                return input;
             }
 
             JsonNode jsonNode = mapper.readTree(text.substring(start));
@@ -133,16 +141,24 @@ public class MLExtractJsonProcessor extends AbstractMLProcessor {
                 if (jsonNode.isObject()) {
                     return mapper.convertValue(jsonNode, Map.class);
                 }
+                if (defaultValue != null) {
+                    log.warn("Expected JSON object but found {}, returning default value", jsonNode.getNodeType());
+                    return defaultValue;
+                }
                 log.debug("Expected JSON object but found {}", jsonNode.getNodeType());
-                return defaultValue != null ? defaultValue : input;
+                return input;
             }
 
             if (EXTRACT_TYPE_ARRAY.equalsIgnoreCase(extractType)) {
                 if (jsonNode.isArray()) {
                     return mapper.convertValue(jsonNode, List.class);
                 }
+                if (defaultValue != null) {
+                    log.warn("Expected JSON array but found {}, returning default value", jsonNode.getNodeType());
+                    return defaultValue;
+                }
                 log.debug("Expected JSON array but found {}", jsonNode.getNodeType());
-                return defaultValue != null ? defaultValue : input;
+                return input;
             }
 
             // auto detect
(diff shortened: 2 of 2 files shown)


rithinpullela:
> Will this work if user wants to add another output parser on top of this default parser? Ideally you would register output parser during tool registration. I get the idea that we need to parse it here to return raw tool response. Just make sure that if other parsers are used it can work with this.

@pyek-bot yes, users can add another parser on top of the default one.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -431,17 +439,48 @@ public QueryPlanningTool create(Map<String, Object> map) {
...
+         * @return Parser with extract_json as first processor, followed by any custom processors
+         */
+        private Parser createParserWithDefaultExtractJson(Map<String, Object> params) {
+            // Extract any existing custom processors from params
+            List<Map<String, Object>> customProcessorConfigs = ProcessorChain.extractProcessorConfigs(params);
+
+            // Create the default extract_json processor config
+            Map<String, Object> extractJsonConfig = new HashMap<>();
+            extractJsonConfig.put("type", "extract_json");
+            extractJsonConfig.put("extract_type", "object"); // Extract JSON objects only
```
mingshl, line 475: this will silently fall into default search template if json not found. how do you plan to notify this change? is this the expected behaviour? does it make it better controlling by a flag like `ignoreParsingFailure`? 
at least adding debug logging?

mingshl: another thing, if you want to further validate if this is a valid query, use opensearch query parsers. If it's json and not valid query, you can still fall back to default search template. IMO, this might be better validation.

rithinpullela: Thanks Mingshi, these are valid points. We are planning to work on this in the next versions of Agentic Search: https://github.com/opensearch-project/ml-commons/issues/4359

The current behavior causes a random error when the query is run against OpenSearch Index without any clear explanation to the end user. And it has been established in 3.3 release ([code ref](https://github.com/opensearch-project/ml-commons/blob/19eeb7bba69bcd2fa7d1193eb4a132cfcffa9898/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java#L254-L258)) that when the LLM does not have enough context or fails to produce an output we return the default match all query. Continuing the same behavior (while not ideal) we are returning the default match all query when the JSON is malformed.

We will need to discuss further in the next release how to change this behavior. I would like to keep the behavior same across 3.3 minor versions. And we can introduce a change in 3.4

rithinpullela:
The failing tests are not related to the change I made: 
These are probably flaky since they did not fail on other runs
```
MLModelGroupRestIT > test_search_MatchAllQuery_For_ModelGroups FAILED
    java.lang.AssertionError: expected null, but was:<org.opensearch.client.ResponseException: method [GET], host [https://localhost:9200], URI [/_plugins/_ml/model_groups/_search], status line [HTTP/2.0 500 Internal Server Error]
    {"error":{"root_cause":[{"type":"status_exception","reason":"Fail to search"}],"type":"status_exception","reason":"Fail to search"},"status":500}>
        at __randomizedtesting.SeedInfo.seed([DF82DC920B47ACA4:A2154E7F85406A33]:0)
        at org.junit.Assert.failNotNull(Assert.java:756)
        at org.junit.Assert.assertNull(Assert.java:738)
        at org.junit.Assert.assertNull(Assert.java:748)
        at org.opensearch.ml.rest.MLModelGroupRestIT.test_search_MatchAllQuery_For_ModelGroups(MLModelGroupRestIT.java:1186)
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: please create issue to track the enhancement, many of these offline discussions are going away while we just merging quick fix

rithinpullela: I attached the created issue in the above message

dhrubo-os (dismissed):

rithinpullela:
Raised a documentation PR describing the default behavior: https://github.com/opensearch-project/documentation-website/pull/11446

ylwu-amzn (approved):

rithinpullela:
Failing because of unrelated changes:
```
MLModelAutoReDeployerIT > testModelAutoRedeploy FAILED
REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests 'org.opensearch.ml.autoredeploy.MLModelAutoReDeployerIT.testModelAutoRedeploy' -Dtests.seed=B374BA3F198CB8FE -Dtests.security.manager=false -Dtests.locale=pcm-Latn-NG -Dtests.timezone=America/Asuncion -Druntime.java=21
    java.lang.AssertionError: method [POST], host [http://127.0.0.1:42527/], URI [/_plugins/_ml/models/null/_deploy], status line [HTTP/1.1 404 Not Found]


Suite: Test class org.opensearch.ml.autoredeploy.MLModelAutoReDeployerIT
    {"error":{"root_cause":[{"type":"status_exception","reason":"Failed to find model"}],"type":"status_exception","reason":"Failed to find model"},"status":404}
  2> Ọkt 28, 2025 9:34:19 FỌ ÍVNIN org.opensearch.client.RestClient logResponse
        at __randomizedtesting.SeedInfo.seed([B374BA3F198CB8FE:ADF94E4B10A2EF02]:0)
        at org.opensearch.ml.autoredeploy.MLModelAutoReDeployerIT.lambda$prepareModel$0(MLModelAutoReDeployerIT.java:52)
        at org.opensearch.ml.rest.MLCommonsRestTestCase.verifyResponse(MLCommonsRestTestCase.java:686)
        at org.opensearch.ml.rest.MLCommonsRestTestCase.getTask(MLCommonsRestTestCase.java:650)
```

PR #4361, opened 2025-10-28
Add tool parameter type validation

nathaliellenaa:
### Description
Add tool parameter type validation.
- Add new parameter `max_question_length`/`max_input_length` to control max size for `question`/`input` field. Default is 10000. Users can change the value by passing this parameter when register an agent or executing tool
```
// Register agent
{
  "name": "Test agent tool",
  "type": "flow",
  "tools": [
    {
      "type": "AgentTool",
      "description": "A general agent to answer any question",
      "parameters": {
        "agent_id": "9X7xWI0Bpc3sThaJdY9i",
        "max_question_length": 10
      }
    }
  ]
}

// Execute agent
POST /_plugins/_ml/tools/_execute/AgentTool
{
    "parameters": {
        "agent_id": "g-CELJoBkoz7WY_qrUkJ",
        "question": "how many indices in my cluster?"
        "max_question_length": 20000
      }
}

// Sample error response
{
    "status": 400,
    "error": {
        "type": "IllegalArgumentException",
        "reason": "Invalid Request",
        "details": "question length cannot exceed 10 characters"
    }
}
```
- Add validations for malformed input

Examples:
```
POST /_plugins/_ml/tools/_execute/ListIndexTool
{
  "parameters": {
    "question": ["invalid"]
  }
}

// Error response
{
    "status": 400,
    "error": {
        "type": "OpenSearchParseException",
        "reason": "Invalid Request",
        "details": "[question] property isn't a string, but of type [java.util.ArrayList]"
    }
}
```
More examples in this doc: 
[Tool Validation Test.pdf](https://github.com/user-attachments/files/23248118/Tool.Validation.Test.pdf)

### Related Issues

### Check List

## Summary by CodeRabbit

* **New Features**
  * Configurable max question length for agent queries.
  * Tools now expose parameter type definitions for runtime validation.
  * Inputs preserve the original raw parameter map for richer value handling.

* **Refactor**
  * Improved parameter parsing with JSON-aware type inference.
  * Centralized parameter validation integrated into tool and agent execution flows.

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -1012,8 +1012,8 @@ public class AgentUtils {
         }
         Tool tool = toolFactories.get(toolSpec.getType()).create(toolParams);
 
-        // Map<String, Class<?>> toolParamDef = tool.getToolParamsDefinition();
-        // validateToolParameters(toolParams, toolParamDef);
+        Map<String, Class<?>> toolParamDef = tool.getToolParamsDefinition();
+        validateToolParameters(toolParams, toolParamDef);
 
         String toolName = getToolName(toolSpec);
         tool.setName(toolName);
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/WriteToScratchPadTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/WriteToScratchPadTool.java
@@ -13,7 +13,6 @@ import java.util.List;
 import java.util.Map;
 
 import org.opensearch.core.action.ActionListener;
-import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.ml.common.spi.tools.Tool;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.ml.common.utils.StringUtils;
@@ -152,8 +151,6 @@ public class WriteToScratchPadTool implements Tool {
 
         @Override
         public WriteToScratchPadTool create(Map<String, Object> params) {
-            ConfigurationUtils.readStringProperty(TYPE, null, params, NOTES_KEY);
-            ConfigurationUtils.readBooleanProperty(TYPE, null, params, RETURN_HISTORY_KEY, false);
             return new WriteToScratchPadTool();
         }
 


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -117,6 +117,28 @@ public void setName(String s) {
 
     @Override
     public boolean validate(Map<String, String> parameters) {
+        if (parameters == null || parameters.isEmpty()) {
+            return false;
+        }
+
+        // Validate question length
+        String question = parameters.get("question");
+        if (question != null && question.length() > 10000) {
```
mingshl: can you make the max_length as a static variable? and can the variable to be changed by users?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java
```diff
@@ -100,6 +100,18 @@ public boolean validate(Map<String, String> parameters) {
         return parameters != null && !parameters.isEmpty();
     }
 
+    @Override
+    public boolean validateParameterTypes(Map<String, Object> parameters) {
+        // Validate response_filter must be String
+        Object responseFilterObj = parameters.get("response_filter");
```
mingshl: this validation method only check one field at a time, checkout this parser in create method, you can try validate the type of the each parameters when creating tools, this will be more efficient https://github.com/opensearch-project/ml-commons/blob/5964268b0522deecc5a2333fabce69bd4f3118aa/plugin/src/main/java/org/opensearch/ml/processor/MLInferenceIngestProcessor.java#L481

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
nathaliellenaa: Will add a default value for the `max_question_length` and allow users to change this value by specifying `max_question_length` parameters when registering a tool in an agent or directly executing the tool

Register agent
```
{
  "name": "Test agent tool",
  "type": "flow",
  "tools": [
    {
      "type": "AgentTool",
      "description": "A general agent to answer any question",
      "parameters": {
        "agent_id": "9X7xWI0Bpc3sThaJdY9i",
        "max_question_length": 100
      }
    }
  ]
}
```

Execute tool
```
POST /_plugins/_ml/tools/_execute/AgentTool
{
    "parameters": {
        "agent_id": "g-CELJoBkoz7WY_qrUkJ",
        "question": "how many indices in my cluster?"
        "max_question_length": 20000
      }
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java
nathaliellenaa: Good suggestion, I removed `validateParameterType` and used methods in `ConfigurationUtils` class to validate the parameter types

nathaliellenaa:
Working on adding tests

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
```diff
@@ -175,7 +179,17 @@ public String getType() {
...
-        return parameters != null && !parameters.isEmpty() && parameters.containsKey("index");
+        if (parameters == null || parameters.isEmpty() || !parameters.containsKey("index")) {
+            return false;
+        }
+
+        // Validate question length
+        String question = parameters.get("question");
+        if (question != null && question.length() > maxQuestionLength) {
+            throw new IllegalArgumentException("question length cannot exceed " + maxQuestionLength + " characters");
+        }
```
rithinpullela: Why would the index mapping tool, need a "question" attribute?
Looks like we are not using it anywhere

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
```diff
@@ -86,6 +87,9 @@ public class ListIndexTool implements Tool {
         + "for example: [\\\"index1\\\", \\\"index2\\\"], use empty array [] to list all indices in the cluster\"}},"
         + "\"additionalProperties\":false}";
     public static final Map<String, Object> DEFAULT_ATTRIBUTES = Map.of(TOOL_INPUT_SCHEMA_FIELD, DEFAULT_INPUT_SCHEMA, STRICT_FIELD, false);
+    public static final int DEFAULT_MAX_QUESTION_LENGTH = 10000;
+    public static final String MAX_QUESTION_LENGTH_FIELD = "max_question_length";
+    private int maxQuestionLength = DEFAULT_MAX_QUESTION_LENGTH;
```
rithinpullela: Same as above, we don't need the question parameter. The tool just ignores the question param

rithin-pullela-aws (commented):
Thanks for this PR Nathalie, these validations have been long time pending. Added 2 small comments, other than that LGTM!

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
nathaliellenaa: We use it during tool execution. Also, in the [doc](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/index-mapping-tool/#execute-parameters), it listed `question` as a required parameter. 
```
POST /_plugins/_ml/agents/9X7xWI0Bpc3sThaJdY9i/_execute
{
  "parameters": {
    "index": [ "sample-ecommerce" ],
    "question": "What fields are in the sample-ecommerce index?"
  }
}
```

I tested and we can run this tool without the question parameter. I'll update the doc to make it optional. 
We can keep the question length validation here since users can still pass the question parameter, but I'll modify this to optional

https://github.com/opensearch-project/ml-commons/blob/8932e94f97238c55ecaee547050aeab0151a6dba/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java#L229

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java
nathaliellenaa: Same thing here, we use it during tool execution. [Doc](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/list-index-tool/#execute-parameters) marked it as a required field. But looking at the code, we can actually run this tool without question parameter. I'll update the doc to make this parameter optional.

We can keep the question length validation here since users can still pass the question parameter, but I'll modify this to optional
https://github.com/opensearch-project/ml-commons/blob/8932e94f97238c55ecaee547050aeab0151a6dba/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java#L471

nathaliellenaa:
Some outdated tool information I found during testing, will open doc PR to update these
- [ListIndexTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/list-index-tool/#execute-parameters)
  - `indices` should be array type
  - `question` parameter should be optional
- [IndexMappingTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/index-mapping-tool/#register-parameters)
  - `input` and `question` parameter should be optional
 - [SearchIndexTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/search-index-tool/#execute-parameters)
   - Mark `input` parameter as required
- [ReadFromScratchPadTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/scratchpad-tools/#readfromscratchpadtool)
  - `persistent_notes` parameter is optional when executing tool directly

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
zane-neo: Although the question is shown in the parameter map passed to tools, but `question` is an agent level parameter, let's say an agent runs 10 tools, we only need to validate the `question` once when agent runs, instead of validating it 10 times when each tool runs.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -1014,4 +1017,32 @@ public static Tool createTool(Map<String, Tool.Factory> toolFactories, Map<Strin
 
         return tool;
     }
+
+    private static Object parseValue(String value) {
```
zane-neo, line 1031: I don't see the benefit of parsing these values, if a tool requires different types like map or double/float, then they're still being passed to tool in String format.

zane-neo:
High level comment:
1. Can we add an interface method in `org.opensearch.ml.common.spi.tools.Tool` like `getCreateToolParamsDefinition` so that every tool returns either a json string like `attributes` or a Map<String, Class> to indicate what type the parameter should be. And validate the parameters when creating the agent?
2. `question` is an agent level parameter and should be validated in agent instead of tools.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java
nathaliellenaa: That makes sense, I added `validateAgentParameters` method in the MLAgentExecutor class to validate the `question` param once when agent was being executed.

PR #4386, opened 2025-11-03
Enforce size limit for SearchIndexTool

yyfamazon:
### Description
Enforce size limit for SearchIndexTool. 
PER agent rely on LLM to generate DSL then call SearchIndexTool to execute the DSL. There is no size limit enforced by SearchIndexTool, if LLM generate a DSL that returns 1000 or even more data, SearchIndexTool will do so and when the result sending back to agent as context, it will fail due to Input is too long for requested model error. We need to add a size limit for SearchIndexTool to return at most 100? records no matter what's the size value is set by LLM.

### Related Issues

### Check List

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -711,16 +711,6 @@ configurations.all {
     resolutionStrategy.force "jakarta.json:jakarta.json-api:2.1.3"
     resolutionStrategy.force "org.opensearch:opensearch:${opensearch_version}"
     resolutionStrategy.force "org.bouncycastle:bcprov-jdk18on:1.78.1"
-    // Force consistent Netty versions to resolve conflicts
-    resolutionStrategy.force 'io.netty:netty-codec-http:4.1.124.Final'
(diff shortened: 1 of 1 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -202,6 +211,14 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
                 query = parameters.get(QUERY_FIELD);
             }
 
+            if (StringUtils.isNotEmpty(parameters.get(SIZE_FIELD))) {
+                try {
+                    size = Math.min(size == null ? 100 : size, Integer.parseInt(parameters.get(SIZE_FIELD)));
```
xinyual: The size here is also generated by LLM, right? Only when LLM doesn't generate anything we will use the default size 100 in PER agent.

xinyual:
Please apply spotless.

PR #4390, opened 2025-11-03
[MCP Server] Fix tool name bug and support ObjectType returning tools

rithinpullela:
### Changes
-  Modified type conversion in MCP server to use .toString() instead of direct string casting. Updated line 86 McpToolsHelper.java to handle non-string objects returned by output processors
- When Tool Name is not provided we automatically take the tool type as name

### Background
With the addition of output processors in 3.3, tools can now return non-string objects (e.g., Extract JSON returns HashMap). The existing MCP server code assumed all tool responses were strings wrapped in generic types, causing ClassCastException when trying to cast HashMap to String.

### Solution
Changed the conversion logic to accept any Object type and convert to string using .toString(), maintaining compatibility with both legacy string responses and new output processor results.

This ensures MCP server works with all tool types while preserving existing
functionality.

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
@@ -83,8 +83,8 @@ public class McpToolsHelper {
         return new McpStatelessServerFeatures.AsyncToolSpecification(
             new McpSchema.Tool(toolName, String.valueOf(description), schema),
             (ctx, request) -> Mono.create(sink -> {
-                ActionListener<String> actionListener = ActionListener
-                    .wrap(r -> sink.success(new McpSchema.CallToolResult(List.of(new McpSchema.TextContent(r)), false)), e -> {
+                ActionListener<Object> actionListener = ActionListener
+                    .wrap(r -> sink.success(new McpSchema.CallToolResult(List.of(new McpSchema.TextContent(r.toString())), false)), e -> {
                         log.error("Failed to execute tool, tool name: {}", toolName, e);
                         sink.error(e);
                     });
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
```diff
@@ -83,8 +83,8 @@ public McpStatelessServerFeatures.AsyncToolSpecification createToolSpecification
         return new McpStatelessServerFeatures.AsyncToolSpecification(
             new McpSchema.Tool(toolName, String.valueOf(description), schema),
             (ctx, request) -> Mono.create(sink -> {
-                ActionListener<String> actionListener = ActionListener
-                    .wrap(r -> sink.success(new McpSchema.CallToolResult(List.of(new McpSchema.TextContent(r)), false)), e -> {
+                ActionListener<Object> actionListener = ActionListener
```
brianf-aws, line 86: Just curious what is the reason for this change? Because it reads that adding the .toString() would still keep the same type

rithinpullela: Our tools return a generic Object type when executed (see run function in tools, [example](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ListIndexTool.java#L120)). Most of the tools have a string response and the string is wrapped in a generic type <T> before returning. 

Hence when the MCP execute tool expects a String, the generic type returned back from the Tool execution is converted back to String in line 86 without an issue.

But with the latest Output processors that are added in 3.3, the final return type might not be a string wrapped in a generic type anymore, for example Extract JSON converts the string into a map object when processed. The direct tool execution is not impacted because the HashMap type is wrapped in a generic <T>

But when we execute a tool with the output processors via the MCP server, the server tries to convert the <T> type to a string. which will fail now since the object inside is not a string anymore, it is a hashmap which cannot be casted into a string.

Hence I made this change which accepts any Object type and the `.toString()` converts it back to a string data type which is required by the MCP SDK dependency.

I hope this clarifies your question, please let me know if I need to add more details

PR #4412, opened 2025-11-11
upgrade dependencies to make it compatible with AD

jackiehanyang:
### Description
In AD Detection Insights feature, we're calling ml-commons metrics correlation algorithm from AD which adds ml-commons as a runtime dependency. So making some dependency updates to make AD and ml-commons run in the same JVM runtime inside that container and communicate via transport actions.

### Related Issues

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ subprojects {
 
     configurations.all {
         // Force spotless depending on newer version of guava due to CVE-2023-2976. Remove after spotless upgrades.
-        resolutionStrategy.force "com.google.guava:guava:32.1.3-jre"
+        resolutionStrategy.force "com.google.guava:guava:33.4.5-jre"
         resolutionStrategy.force 'org.apache.commons:commons-compress:1.26.0'
         resolutionStrategy.force "io.netty:netty-buffer:${versions.netty}"
         resolutionStrategy.force "io.netty:netty-codec:${versions.netty}"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -94,7 +94,7 @@ subprojects {
 
     configurations.all {
         // Force spotless depending on newer version of guava due to CVE-2023-2976. Remove after spotless upgrades.
-        resolutionStrategy.force "com.google.guava:guava:32.1.3-jre"
+        resolutionStrategy.force "com.google.guava:guava:33.4.5-jre"
```
dhrubo-os, line 97: May be both plugins can take version from core: https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml#L23 ?

jackiehanyang: No, AD needs it for a resource permission related change from security plugin side - https://github.com/opensearch-project/anomaly-detection/commit/ee8e38dda12cb5a6a97039ae97e7f4fc87ec5826 They are planning to make this change for all plugins just AD is the first one onboarded

rithinpullela, line 97: NIT: 
We are forcing the newer version in the root module, but looks like other modules still declare older versions 32.1.3
Should we change it in all the packages as well? This would be more for overall consistency since the resolutionStrategy.force in the root module would ensure all the modules use the latest version declared here.

Issue #4424, opened 2025-11-14
[BUG] Updating conversational agents causes agentic search to fail


ohltyler:
If updating conversational agents and re-running agentic search, it fails with the following error:
```
Error searching index: [illegal_argument_exception] Agentic search failed - Agent execution error - Agent ID: [7G6Bg5oBpAtrCWabdaOI], Error: [Model ID can't be null or empty]
```

From logs:
```
[2025-11-14T11:55:00,444][ERROR][o.o.m.e.a.a.MLAgentExecutor] [integTest-0] Failed to run conversational agent
java.lang.IllegalArgumentException: Model ID can't be null or empty
        at org.opensearch.ml.engine.tools.MLModelTool.<init>(MLModelTool.java:73) ~[?:?]
```

Using any underlying working model for agentic search, follow these steps to reproduce:

1. Register agent
```
POST /_plugins/_ml/agents/_register
{
  "name": "test",
  "type": "conversational",
  "description": "",
  "tools": [
    {
      "type": "QueryPlanningTool"
    }
  ],
  "llm": {
    "model_id": "<model-id>"
  },
  "memory": {
    "type": "conversation_index"
  },
  "parameters": {
    "_llm_interface": "<interface>"
  }
}
```

2. Run agentic search
```
GET /<index>/_search
{
  "query": {
    "agentic": {
      "query_text": "match all",
      "query_fields": []
    }
  },
  "search_pipeline": {
    "request_processors": [
      {
        "agentic_query_translator": {
            "agent_id": "<agent-id>"
        }
      }
    ]
  }
}
```

3. Update conversational agent (in this case, adding SearchIndexTool)
```
PUT /_plugins/_ml/agents/<agent-id>
{
  "name": "test",
  "type": "conversational",
  "description": "",
  "tools": [
    {
      "type": "QueryPlanningTool"
    },
    {
      "type": "SearchIndexTool"
    }
  ],
  "llm": {
    "model_id": "<model-id>"
  },
  "memory": {
    "type": "conversation_index"
  },
  "parameters": {
    "_llm_interface": "<interface>"
  }
}
```

4. Verify it's updated
```
GET /_plugins/_ml/agents/<agent-id>
```

5. Re-run agentic search
```
GET /<index>/_search
{
  "query": {
    "agentic": {
      "query_text": "match all",
      "query_fields": []
    }
  },
  "search_pipeline": {
    "request_processors": [
      {
        "agentic_query_translator": {
[...]

PR #4458, opened 2025-11-20
[Agentic Search] Fix model id parsing for QueryPlanningTool

owaiskazi19:
### Description
During tool creation, only mlAgent.getParameters() was being passed to tools but QPT has a special case since it uses the same model as the agent llm. Fixed this by parsing agent llm in the params and checked on QPT side if there's no model_id in the params then use the agent llm as model id.

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/4424

### Check List

## Summary by CodeRabbit

* **New Features**
  * Agent language model ID is now available to and utilized by tools
  * Query planning tool automatically applies agent's model ID as default when not explicitly specified
  * Model ID precedence ensures explicit tool settings override agent defaults

* **Tests**
  * Added comprehensive test coverage for model ID propagation and fallback behavior
  * Validates error handling when no model ID is available

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestMLRAGSearchProcessorIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestMLRAGSearchProcessorIT.java
@@ -716,8 +716,7 @@ public class RestMLRAGSearchProcessorIT extends MLCommonsRestTestCase {
         requestParameters.timeout = 60;
         requestParameters.imageFormat = "jpeg";
         requestParameters.imageType = "url";
-        requestParameters.imageData =
-            "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"; // imageContent;
+        requestParameters.imageData = "https://images.unsplash.com/photo-1506905925346-21bda4d32df4?w=800";
         Response response3 = performSearch(INDEX_NAME, "pipeline_test", 5, requestParameters);
         assertEquals(200, response2.getStatusLine().getStatusCode());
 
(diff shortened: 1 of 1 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -420,6 +421,10 @@ public void init(Client client) {
 
         @Override
         public QueryPlanningTool create(Map<String, Object> params) {
+            // Use agent's Agent model_id if tool doesn't have its own model_id
+            if (!params.containsKey(MODEL_ID_FIELD) && params.containsKey(AGENT_LLM_MODEL_ID)) {
+                params.put(MODEL_ID_FIELD, params.get(AGENT_LLM_MODEL_ID));
+            }
```
rithinpullela, line 427: When creating the Query Planner tool, during the registration of the agent, we are adding the agent's LLM Model ID as the model ID for QPT. But I believe this approach is a much cleaner way of doing things. 

Can we also clean up this code in [register Agent](https://github.com/opensearch-project/ml-commons/blob/87e742e41bd3a49eb226dff2cb99c2e8c3b5ceb0/plugin/src/main/java/org/opensearch/ml/action/agents/TransportRegisterAgentAction.java#L134-L152)?

owaiskazi19:
> @owaiskazi19 are these changes for 3.4?

Yes, any version aware settings I need to look into?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
mingshl: @rithin-pullela-aws earlier, you had a PR that check if it's query planning tool then you passed over the model id to query planning tool and that's specific for queryPlanningTool.I wanted something more generic to tools and I like this PR's approach to handle model id for all tool execution. 

Can you try find out that PR and then we can revert that commit and take this one.

owaiskazi19: [This](https://github.com/opensearch-project/ml-commons/pull/4262) is the PR but it has some other changes as well. I will leave that to @rithin-pullela-aws

mingshl: Approve to unblock. @rithin-pullela-aws please raise PR to revert the previous commit.

PR #4508, opened 2025-12-12
Optimize connector client configuration

zane-neo:
### Description

Current [connector client configuration](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java#L39-L40) are not optimal, the value are too big for a httpclient, if the remote endpoint performance is not good and holding the request for long time without response, then the `connections` will be exhausted in the connection pool. We need to set a better value for them. There isn't a silver bullet configuration for all scenarios, but since we're using pooling connections, we don't want the connection waiting too long(connection usually completed in milliseconds if a server is up), so we set the default connection timeout to 1s, we can't guarantee all LLM can respond quickly so we set the read timeout to 10s.

Update: changed the default connection timeout to 10s and default read timeout to 30s, see details in comments.

### Related Issues

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Timeout handling standardized to explicit millisecond units to avoid misinterpretation.

* **New Features**
  * Executors adopt a closeable lifecycle for standardized resource cleanup.
  * Connector configuration exposes millisecond-based timeout fields and uses millis throughout.

* **Breaking Changes**
  * Public timeout APIs renamed to indicate millis (e.g., getConnectionTimeoutMillis).

* **Chores**
  * Added "**/out" to ignore patterns and editor settings.

* **Tests**
  * Tests and fixtures updated for renamed timeout APIs and added read_timeout in configs.

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteModel.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteModel.java
@@ -9,8 +9,6 @@ import static org.opensearch.ml.common.connector.ConnectorAction.ActionType.PRED
 import static org.opensearch.ml.common.settings.MLCommonsSettings.REMOTE_METADATA_GLOBAL_TENANT_ID;
 
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
@@ -113,31 +111,53 @@ public class RemoteModel implements Predictable {
     }
 
     @Override
-    public CompletionStage<Boolean> initModelAsync(MLModel model, Map<String, Object> params, Encryptor encryptor) {
+    public void initModelAsync(MLModel model, Map<String, Object> params, Encryptor encryptor, ActionListener<Predictable> listener) {
         SdkClient sdkClient = (SdkClient) params.get(SDK_CLIENT);
-        return sdkClient.isGlobalResource(MLIndex.MODEL.getIndexName(), model.getModelId()).thenCompose(isGlobalResource -> {
+        sdkClient.isGlobalResource(MLIndex.MODEL.getIndexName(), model.getModelId()).whenComplete((isGlobalResource, throwable) -> {
+            if (throwable != null) {
+                log.error("Failed to init remote model.", throwable);
+                listener.onFailure(new MLException(throwable));
+            } else {
+                innerInitModelAsync(model, params, encryptor, isGlobalResource, listener);
+            }
+        });
+    }
+
+    private void innerInitModelAsync(
+        MLModel model,
+        Map<String, Object> params,
+        Encryptor encryptor,
+        Boolean isGlobalResource,
+        ActionListener<Predictable> listener
+    ) {
+        try {
             String decryptTenantId = Boolean.TRUE.equals(isGlobalResource)
                 ? REMOTE_METADATA_GLOBAL_TENANT_ID.get((Settings) params.get(SETTINGS))
                 : model.getTenantId();
             Connector connector = model.getConnector().cloneConnector();
-            connector.decrypt(PREDICT.name(), (credential, tenantId) -> encryptor.decrypt(credential, decryptTenantId), decryptTenantId);
-            // This situation can only happen for inline connector where we don't provide tenant id.
-            if (connector.getTenantId() == null && model.getTenantId() != null) {
-                connector.setTenantId(model.getTenantId());
-            }
-            this.connectorExecutor = MLEngineClassLoader.initInstance(connector.getProtocol(), connector, Connector.class);
-            this.connectorExecutor.setScriptService((ScriptService) params.get(SCRIPT_SERVICE));
-            this.connectorExecutor.setClusterService((ClusterService) params.get(CLUSTER_SERVICE));
-            this.connectorExecutor.setClient((Client) params.get(CLIENT));
-            this.connectorExecutor.setXContentRegistry((NamedXContentRegistry) params.get(XCONTENT_REGISTRY));
-            this.connectorExecutor.setRateLimiter((TokenBucket) params.get(RATE_LIMITER));
-            this.connectorExecutor.setUserRateLimiterMap((Map<String, TokenBucket>) params.get(USER_RATE_LIMITER_MAP));
-            this.connectorExecutor.setMlGuard((MLGuard) params.get(GUARDRAILS));
-            this.connectorExecutor.setConnectorPrivateIpEnabled((boolean) params.getOrDefault(CONNECTOR_PRIVATE_IP_ENABLED, false));
-            return CompletableFuture.completedStage(true);
-        }).exceptionally(e -> {
-            log.error("Failed to init remote model.", e);
-            throw new MLException(e);
-        });
+            ActionListener<Boolean> decryptSuccessfulListener = ActionListener.wrap(r -> {
+                // This situation can only happen for inline connector where we don't provide tenant id.
+                if (connector.getTenantId() == null && model.getTenantId() != null) {
+                    connector.setTenantId(model.getTenantId());
+                }
+                this.connectorExecutor = MLEngineClassLoader.initInstance(connector.getProtocol(), connector, Connector.class);
+                this.connectorExecutor.setScriptService((ScriptService) params.get(SCRIPT_SERVICE));
+                this.connectorExecutor.setClusterService((ClusterService) params.get(CLUSTER_SERVICE));
+                this.connectorExecutor.setClient((Client) params.get(CLIENT));
+                this.connectorExecutor.setXContentRegistry((NamedXContentRegistry) params.get(XCONTENT_REGISTRY));
+                this.connectorExecutor.setRateLimiter((TokenBucket) params.get(RATE_LIMITER));
+                this.connectorExecutor.setUserRateLimiterMap((Map<String, TokenBucket>) params.get(USER_RATE_LIMITER_MAP));
+                this.connectorExecutor.setMlGuard((MLGuard) params.get(GUARDRAILS));
+                this.connectorExecutor.setConnectorPrivateIpEnabled((boolean) params.getOrDefault(CONNECTOR_PRIVATE_IP_ENABLED, false));
+                listener.onResponse(this);
+            }, e -> {
+                log.error("Failed to init remote model.", e);
+                listener.onFailure(new MLException(e));
+            });
+            connector.decrypt(PREDICT.name(), encryptor::decrypt, decryptTenantId, decryptSuccessfulListener);
+        } catch (Throwable e) {
+            log.error("Failed to init remote model", e);
+            listener.onFailure(new MLException(e));
+        }
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
@@ -10,14 +10,17 @@ import static org.opensearch.ml.common.utils.StringUtils.isJson;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import org.apache.commons.text.StringSubstitutor;
+import org.opensearch.common.TriConsumer;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.authuser.User;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.AccessMode;
@@ -26,7 +29,9 @@ import org.opensearch.ml.common.utils.StringUtils;
 
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
 
+@Log4j2
 @Getter
 public abstract class AbstractConnector implements Connector {
     public static final String ACCESS_KEY_FIELD = "access_key";
@@ -156,5 +161,69 @@ public abstract class AbstractConnector implements Connector {
         }
         return predictEndpoint;
     }
+
+    @Override
+    public void encrypt(
+        TriConsumer<List<String>, String, ActionListener<List<String>>> function,
+        String tenantId,
+        ActionListener<Boolean> listener
+    ) {
+        if (credential == null || credential.isEmpty()) {
+            listener.onResponse(true);
+            return;
+        }
+        List<String> orderedEncryptKeys = new ArrayList<>();
+        List<String> orderedToEncrypt = new ArrayList<>();
+        for (String key : credential.keySet()) {
+            orderedEncryptKeys.add(key);
+            orderedToEncrypt.add(credential.get(key));
+        }
+        ActionListener<List<String>> updateEncryptedCredentialsListener = ActionListener.wrap(r -> {
+            Map<String, String> encryptedCredentials = new HashMap<>();
+            for (int i = 0; i < r.size(); i++) {
+                encryptedCredentials.put(orderedEncryptKeys.get(i), r.get(i));
+            }
+            credential = encryptedCredentials;
+            listener.onResponse(true);
+        }, e -> {
+            log.error("Failed to encrypt credentials in connector", e);
+            listener.onFailure(e);
+        });
+        function.apply(orderedToEncrypt, tenantId, updateEncryptedCredentialsListener);
+    }
+
+    @Override
+    public void decrypt(
+        String action,
+        TriConsumer<List<String>, String, ActionListener<List<String>>> function,
+        String tenantId,
+        ActionListener<Boolean> listener
+    ) {
+        if (credential == null || credential.isEmpty()) {
+            this.decryptedHeaders = createDecryptedHeaders(getAllHeaders(action));
+            listener.onResponse(true);
+            return;
+        }
+        List<String> orderedDecryptKeys = new ArrayList<>();
+        List<String> orderedToDecrypt = new ArrayList<>();
+        for (Map.Entry<String, String> entry : credential.entrySet()) {
+            orderedDecryptKeys.add(entry.getKey());
+            orderedToDecrypt.add(entry.getValue());
+        }
+        ActionListener<List<String>> updateDecryptedCredentialsListener = ActionListener.wrap(r -> {
+            decryptedCredential = new HashMap<>();
+            for (int i = 0; i < r.size(); i++) {
+                decryptedCredential.put(orderedDecryptKeys.get(i), r.get(i));
+            }
+            this.decryptedHeaders = createDecryptedHeaders(getAllHeaders(action));
+            listener.onResponse(true);
+        }, e -> {
+            log.error("Failed to decrypt credentials in connector", e);
+            listener.onFailure(e);
+        });
+        function.apply(orderedToDecrypt, tenantId, updateDecryptedCredentialsListener);
+    }
+
+    protected abstract Map<String, String> getAllHeaders(String action);
 
 }
--- a/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/Connector.java
@@ -16,14 +16,15 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.commons.text.StringSubstitutor;
+import org.opensearch.common.TriConsumer;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.authuser.User;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
@@ -79,9 +80,18 @@ public interface Connector extends ToXContentObject, Writeable {
 
     <T> T createPayload(String action, Map<String, String> parameters);
 
-    void decrypt(String action, BiFunction<String, String, String> function, String tenantId);
+    void decrypt(
+        String action,
+        TriConsumer<List<String>, String, ActionListener<List<String>>> function,
+        String tenantId,
+        ActionListener<Boolean> listener
+    );
 
-    void encrypt(BiFunction<String, String, String> function, String tenantId);
+    void encrypt(
+        TriConsumer<List<String>, String, ActionListener<List<String>>> function,
+        String tenantId,
+        ActionListener<Boolean> listener
+    );
 
     Connector cloneConnector();
 
@@ -91,7 +101,7 @@ public interface Connector extends ToXContentObject, Writeable {
 
     void writeTo(StreamOutput out) throws IOException;
 
-    void update(MLCreateConnectorInput updateContent, BiFunction<String, String, String> function);
+    void update(MLCreateConnectorInput updateContent);
 
     <T> void parseResponse(T orElse, List<ModelTensor> modelTensors, boolean b) throws IOException;
 
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -24,7 +24,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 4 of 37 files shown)


common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -35,17 +35,17 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
...
     public static final Integer RETRY_TIMEOUT_SECONDS_DEFAULT_VALUE = 30;
     public static final Integer MAX_RETRY_TIMES_DEFAULT_VALUE = 0;
     public static final RetryBackoffPolicy RETRY_BACKOFF_POLICY_DEFAULT_VALUE = RetryBackoffPolicy.CONSTANT;
     public static final Version MINIMAL_SUPPORTED_VERSION_FOR_RETRY = Version.V_2_15_0;
     private Integer maxConnections;
-    private Integer connectionTimeout;
-    private Integer readTimeout;
+    private Integer connectionTimeoutMillis;
+    private Integer readTimeoutSeconds;
     private Integer retryBackoffMillis;
```
Hailong-am, line 53: could we use milliseconds for both?

ylwu-amzn: +1, sounds very confusing to use different way to configure these timeouts

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -35,17 +35,17 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
     public static final String MAX_RETRY_TIMES_FIELD = "max_retry_times";
     public static final String RETRY_BACKOFF_POLICY_FIELD = "retry_backoff_policy";
 
-    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
-    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
-    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = 30;
+    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = 1000;
+    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = 10;
```
dhrubo-os: Originally we had 30 seconds for both of these cases. Can we get back to the original state? Thanks

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -5,22 +5,52 @@
...
+            Duration readTimeout = Duration.ofSeconds(connectorClientConfig.getReadTimeoutSeconds());
+            Integer maxConnection = connectorClientConfig.getMaxConnections();
+            this.httpClientRef
+                .compareAndSet(
+                    null,
+                    MLHttpClientFactory.getAsyncHttpClient(connectionTimeout, readTimeout, maxConnection, connectorPrivateIpEnabled)
+                );
+        }
+        return httpClientRef.get();
+    }
```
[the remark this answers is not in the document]
dhrubo-os: what do you think about making this change?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteModel.java
```diff
@@ -97,6 +97,7 @@ public void asyncPredict(MLInput mlInput, ActionListener<MLTaskResponse> actionL
 
     @Override
     public void close() {
+        this.connectorExecutor.close();
         this.connectorExecutor = null;
     }
```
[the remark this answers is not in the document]
dhrubo-os: +1

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
dhrubo-os: +1

zane-neo: In documentation, we specified that connection timeout is in millisecond and read timeout in second: https://docs.opensearch.org/latest/ml-commons-plugin/remote-models/blueprints/#configuration-parameters, changing this could break existing customer configurations.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zane-neo: Can you provide some evidence that connection/read timeout both setting to 30s is a good option?

dhrubo-os: I don't have evidence handy. But my point is this was 30s before until the bug we introduced. And to fix the bug if we change the value drastically that seems a breaking change to me.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
dhrubo-os: https://github.com/opensearch-project/documentation-website/issues/9460

based on this issue, the documentation was changed. Let's make the values consistent. I don't have preference about seconds vs miliseconds. But keep this consistent please. Thanks.

zane-neo: Ok, I can change this to milliseconds but this could be a breaking change for existing customer connector configurations, e.g. user configuring to 3s now become 3 milliseconds which basically means all remote model calls fails, @ylwu-amzn , please also confirm on this.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zane-neo: This change is not targeting for reverting code back, its purpose is to address some issues.

Screenshot 2025-12-23 at 10 53 50
This is a general timeout setting given by Claude, basically we shouldn't set connection timeout 30s that big in our case.

A good user experience of the E2E latency better less than 10s based on this article: https://www.nngroup.com/articles/response-times-3-important-limits. Note that this is only an internal http call instead of E2E latency.

From retry perspective, if retry mechanism is on and if a connection can't established and we wait for 30s, let's say we have 3 retries and that's 90s, I don't think user would like wait that long.

From serverless case, SGW's read timeout is 15s so if this number is bigger than 15s then SGW timed out first.

So all above I prefer to keep the current settings, if you still have concern please give some proof that we have to do that way, thanks!

dhrubo-os: Sure, if you want to introduce breaking change with optimal values may be we can look into this: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/timeouts.html ?

zane-neo: This documentation is talking about how to setup API level timeouts(including retries etc) and http level timeouts(pure http perspective), in our case we are using `NettyNioAsyncHttpClient` and we only enabled http timeouts for customer. I'm not following you on which part I should look into?

Hailong-am: My concerns about changing the default value may break existing connectors/use cases rely on the original value which is 30s, and connectors could connect to any http urls not limited to LLM.

zane-neo: The thing is most connectors are for LLM and LLM has basically highest latency compared to other APIs, also I insist we should consider user experience, if an API need 30s to return response it should be made async. Few users have such patient to wait for a response taking so long.

Another case I'm highlighting is SGW has 15s read timeout in serverless.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
ylwu-amzn: We should preserve backward compatibility. So no way to keep them consistent except breaking backward compatibility?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "kiroAgent.configureMCP": "Enabled"
```
ylwu-amzn: Remove this?

ylwu-amzn:
Resolve conflicts ?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -37,17 +37,17 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
...
 
-    public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
-    public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
-    public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
-    public static final Integer RETRY_BACKOFF_MILLIS_DEFAULT_VALUE = 200;
-    public static final Integer RETRY_TIMEOUT_SECONDS_DEFAULT_VALUE = 30;
-    public static final Integer MAX_RETRY_TIMES_DEFAULT_VALUE = 0;
+    public static final int MAX_CONNECTION_DEFAULT_VALUE = 30;
+    public static final int CONNECTION_TIMEOUT_DEFAULT_VALUE = 1000;
+    public static final int READ_TIMEOUT_DEFAULT_VALUE = 10000;
```
Hailong-am: can make default as 30000 for read timeout? for LLM like claude 4.5 opus with long input it will take more than 10s to return.

For connection timeout, 1s is too low. we should leave room for bad network case, maybe we can set it as 10s

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -15,23 +18,63 @@
...
+                    Optional
+                        .ofNullable(connectorClientConfig.getReadTimeoutMillis())
+                        .orElse(ConnectorClientConfig.READ_TIMEOUT_DEFAULT_VALUE)
+                );
+            int maxConnection = Optional
+                .ofNullable(connectorClientConfig.getMaxConnections())
+                .orElse(ConnectorClientConfig.MAX_CONNECTION_DEFAULT_VALUE);
+            SdkAsyncHttpClient newClient = MLHttpClientFactory
+                .getAsyncHttpClient(connectionTimeout, readTimeout, maxConnection, connectorPrivateIpEnabled);
+            httpClientRef.set(newClient);
```
Hailong-am, line 88: seems that logging been removed?
```
log
                .info(
                    "AwsConnectorExecutor creating HTTP client for connector: {} - maxConnections: {}, connectionTimeout: {}s, readTimeout: {}s",
                    connector.getName(),
                    maxConnection,
                    super.getConnectorClientConfig().getConnectionTimeout(),
                    super.getConnectorClientConfig().getReadTimeout()
                );
                ```

File: ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java [hunk 8]
```diff
@@ -61,8 +61,14 @@
...
 import lombok.Builder;
+import software.amazon.awssdk.utils.SdkAutoCloseable;
 
-public interface RemoteConnectorExecutor {
+/**
+ * This class is responsible for executing the remote connector actions like prediction, it internally encapsulates a SdkAsyncHttpclient
+ * which is used to send the request to the remote model service, when the executor is being closed, we need to close the internal
+ * HttpClient as well.
+ */
+public interface RemoteConnectorExecutor extends SdkAutoCloseable {
```
Hailong-am: nit: we can use java.io.Closeable as well, that should remove the dependency of awssdk

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zane-neo: I don't think 30s as default is a good idea as not all customers are using ml-commons for running agent with long context, they may run LLM with small context or run embedding generation. Also the 10s connection timeout is  too long, I think most users in production using their httpclient don't have this long connection timeout and read timeout configuration, which means their httpclient is timed out but in OpenSearch the request is still running, and the result won't be available to user.  But anyway, since we have several round discussion on this and you insist I'll make this change.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
zane-neo: Will add this back

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
zane-neo: Changed to java.io.Closeable.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
```diff
@@ -72,8 +72,9 @@ public HttpStreamingHandler(
         this.llmInterface = llmInterface;
         this.parameters = parameters;
 
-        Duration connectionTimeout = Duration.ofSeconds(connectorClientConfig.getConnectionTimeout());
-        Duration readTimeout = Duration.ofSeconds(connectorClientConfig.getReadTimeout());
+        // Get connector client configuration
+        Duration connectionTimeout = Duration.ofMillis(connectorClientConfig.getConnectionTimeoutMillis());
+        Duration readTimeout = Duration.ofMillis(connectorClientConfig.getReadTimeoutMillis());
```
ylwu-amzn: Any impact for the existing connector ?  
If connector has timeout as 10 seconds, now it becomes 10 milliseconds?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zane-neo: No, we don't have any option to manipulate existing connectors.

ylwu-amzn: Then we should prioritize the backward compatibility.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
zane-neo: Yes, which is a breaking change, let's discuss in this thread: https://github.com/opensearch-project/ml-commons/pull/4508#discussion_r2630169238

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -15,23 +18,71 @@
...
 @Getter
 public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor {
-    private ConnectorClientConfig connectorClientConfig;
+
+    @Setter
+    private volatile boolean connectorPrivateIpEnabled;
+
+    private final AtomicReference<SdkAsyncHttpClient> httpClientRef = new AtomicReference<>();
+
+    private ConnectorClientConfig connectorClientConfig = new ConnectorClientConfig();
```
ylwu-amzn: `connectorClientConfig` is not volatile. If initialize() is called from one thread while getHttpClient() reads it from another, you may see stale or partially constructed data.
Add `volatile` ?

plugin/src/test/java/org/opensearch/ml/rest/RestMLRAGSearchProcessorIT.java
```diff
@@ -89,6 +92,9 @@ public class RestMLRAGSearchProcessorIT extends MLCommonsRestTestCase {
         + OPENAI_KEY
         + "\"\n"
         + "    },\n"
+        + "    \"client_config\": {\n"
+        + "        \"read_timeout\": 60000\n"
```
mingshl: is it the IT has to raise up the setting for `read_timeout` then the test can pass? would it be a concern for users as well?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
zane-neo: Took a look and this doesn't seem to happen in current predict flow, but adding volatile won't impact, will add this.

plugin/src/test/java/org/opensearch/ml/rest/RestMLRAGSearchProcessorIT.java
zane-neo: I see it's flaky, and 60 is a number that is big enough to get the response, and this is our test account doesn't mean this has same behavior with users, so I think it's fine.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
zane-neo: Changed back to use seconds.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -5,22 +5,73 @@
...
+            return newClient;
+        }
+    }
+
+    public void close() {
+        SdkAsyncHttpClient httpClient = httpClientRef.getAndSet(null);
+        if (httpClient != null) {
+            httpClient.close();
         }
     }
```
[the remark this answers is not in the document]
zane-neo: Not in current scope.

mingshl:
I took another look at the PR, I am aligned that at least we are not changing the configuration name for the parameters, so it's backward compatible. But the actual unit counting in the same parameter is changing.

we should think about the customer experience:
- existing persisted connectors with "read_timeout": 30000 (old millis default) will be interpreted as 30000 seconds after this change — needs migration or backwards-compatible parsing. 
- Existing connectors with "connection_timeout" in seconds (matching old Duration.ofSeconds() usage) will now be treated as milliseconds — breaking change for users who set custom values. we should document this change.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
dhrubo-os: >Ok, I can change this to milliseconds but this could be a breaking change for existing customer connector configurations, e.g. user configuring to 3s now become 3 milliseconds which basically means all remote model calls fails, @ylwu-amzn , please also confirm on this.

why? Then you will change the value to : 

```
public static final int READ_TIMEOUT_DEFAULT_VALUE = 30000;
```

then we won't have any backward compatibility issue, right?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
```diff
@@ -15,23 +18,75 @@
...
         if (connector.getConnectorClientConfig() != null) {
             connectorClientConfig = connector.getConnectorClientConfig();
-        } else {
-            connectorClientConfig = new ConnectorClientConfig();
+        }
+    }
+
+    protected SdkAsyncHttpClient getHttpClient() {
+        // This block for high performance retrieval after http client is created.
+        SdkAsyncHttpClient existingClient = httpClientRef.get();
```
dhrubo-os, line 49: This lazy HTTP client creation introduces a resource leak for short-lived executor instances. RemoteModel.close() handles cleanup for deployed models, but several call sites create executors via 
MLEngineClassLoader.initInstance() for one-off operations and never call close():

- GetTaskTransportAction (line ~472)
- CancelBatchJobTransportAction (line ~262)
- ExecuteConnectorTransportAction (lines ~84-85)
- AgentUtils (lines ~763, 770)

Before this PR these executors didn't own an HTTP client, so there was nothing to leak. Now each invocation leaks an SdkAsyncHttpClient (Netty event loops, connection pools).


can we add try-with-resources at those call sites (since RemoteConnectorExecutor extends SdkAutoCloseable) ?

zane-neo:
> existing persisted connectors with "read_timeout": 30000 (old millis default) will be interpreted as 30000 seconds after this change — needs migration or backwards-compatible parsing.
Existing connectors with "connection_timeout" in seconds (matching old Duration.ofSeconds() usage) will now be treated as milliseconds — breaking change for users who set custom values. we should document this change.

This is the documentation:
```
connection_timeout: The maximum amount of time (in milliseconds) that the client will wait while trying to establish a connection to the server. A timeout prevents the client from waiting indefinitely and allows the client to recover when it encounters unreachable network endpoints.

read_timeout: The maximum amount of time (in seconds) that the client will wait for a response from the server after sending a request. This is useful when the server is slow to respond or encounters an issue while processing a request.
```
So we're changing the code to align with documentation, so it's backward compatible.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
zane-neo: We can refer to this discussion: https://github.com/opensearch-project/ml-commons/pull/4716#issuecomment-4045577687, we have a separate PR to optimize this.

zane-neo:
> why? Then you will change the value to :
public static final int READ_TIMEOUT_DEFAULT_VALUE = 30000;
then we won't have any backward compatibility issue, right?

This is our documentation:
```
connection_timeout: The maximum amount of time (in milliseconds) that the client will wait while trying to establish a connection to the server. A timeout prevents the client from waiting indefinitely and allows the client to recover when it encounters unreachable network endpoints.

read_timeout: The maximum amount of time (in seconds) that the client will wait for a response from the server after sending a request. This is useful when the server is slow to respond or encounters an issue while processing a request.
```
Changing read time to millisecond is definitely a breaking change to user, user configured: `10` will be interpreted as 10000 if we make that change.

zane-neo:
> IMO code should be fixed to represent the value documented in the documentation. Keep read_timeout as seconds and reduce the default value.
It was a bug that the value was so large and reducing this value should be acceptable. This change in behavior can be documented.

Agree, so this PR's change is following the documentation to ensure BWC, once this PR is merged I will create documentation PR to doc this.

dhrubo-os:
> IMO code should be fixed to represent the value documented in the documentation. Keep `read_timeout` as seconds and reduce the default value.

Originally code and doc both were in miliseconds and that was 30000. But then we changed it Duration.Seconds and introduced a `bug` and then we changed it to lower value. I want both of the values (read timeout, connection time out) to be in the same unit. Either miliseconds or Seconds. And then update the doc accordingly.

zane-neo:
@dhrubo-os , we had a lot of discussion in this PR, I checked the whole story history as below:

1. You created the ml-commons [PR](https://github.com/opensearch-project/ml-commons/commit/25f21223367fb334d1276b2543adfa9357d1322e) that introduced the connection_timeout and read_timeout and it's resolved to millisecond in code.
2. You created the documentation [PR](https://github.com/opensearch-project/documentation-website/commit/c5df18a165676999219fc536592893e6b26fa8b3#diff-96003f4b00074982bd947b82690ba02671e4802020c4865ee017847afc52197eR91-R92) that introduced the connection_timeout and read_timeout in documentation but the unit is not consistent to code implementation.
3. I changed the unit in code in this [PR](https://github.com/opensearch-project/ml-commons/pull/1958/) to second which is consistent to documentation but the default values not updated.
4. @pyek-bot created the [issue](https://github.com/opensearch-project/documentation-website/issues/9460) to update the incorrection in the documentation, 
and this is the [PR](https://github.com/opensearch-project/documentation-website/pull/11463) to update the connection_timeout to millisecond instead of second.

So this is buggy in the beggining and now this PR is trying to make code implementation consistent to documentation, this isn't a big change as no customer complained this. And seems we have differnt thoughts and conerns, I don't think we can get a consensus on this in short time, I'll close this PR for now, and since you're the author of these two configurations, can you help create PR to fix this following your thoughts? Thanks.

dhrubo-os:
>You created the documentation [PR](https://github.com/opensearch-project/documentation-website/commit/c5df18a165676999219fc536592893e6b26fa8b3#diff-96003f4b00074982bd947b82690ba02671e4802020c4865ee017847afc52197eR91-R92) that introduced the connection_timeout and read_timeout in documentation but the unit is not consistent to code implementation.

My bad on my side, as I thought I marked as milliseconds in doc as well, but both units are same. That's what I prefer. You don't have to close the PR on this... but anyway whenever I get time, I'll try to work on this if you are not willing to re-open this PR. Thanks.

PR #4516, opened 2025-12-26
Fix tool name matching to prefer exact match over prefix collision

rithinpullela:
### Description
**Issue:**
When an agent has multiple MCP tools with similar names (e.g., `get_order_history` and `get_order_history_simple`), the `getMatchedTool()` function could incorrectly match the shorter tool name due to substring matching.
For example, when the LLM calls `get_order_history_simple`, the function would check if `"get_order_history_simple".contains("get_order_history")` — which is true — and return `get_order_history` instead of the correct tool, causing validation errors on the MCP server.

**Fix:**
Updated getMatchedTool() to:
- Prioritize exact match (case-insensitive) — returns immediately if found
- Fall back to substring matching with longest-match preference to

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool matching logic to prioritize exact matches over partial matches with case-insensitive comparison for more accurate results.

* **Tests**
  * Added test coverage to verify exact match precedence in tool selection.

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -502,12 +502,21 @@ public class AgentUtils {
     }
 
     public static String getMatchedTool(Collection<String> tools, String action) {
+        String bestMatch = null;
+        String actionLower = action.toLowerCase(Locale.ROOT);
         for (String tool : tools) {
-            if (action.toLowerCase(Locale.ROOT).contains(tool.toLowerCase(Locale.ROOT))) {
+            // Exact match takes priority
+            if (action.equalsIgnoreCase(tool)) {
                 return tool;
             }
+            // Track longest substring match to avoid prefix collisions (e.g., "get_order" vs "get_order_simple")
+            if (actionLower.contains(tool.toLowerCase(Locale.ROOT))) {
+                if (bestMatch == null || tool.length() > bestMatch.length()) {
+                    bestMatch = tool;
+                }
+            }
         }
-        return null;
+        return bestMatch;
     }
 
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -502,12 +502,21 @@ public static String substitute(String template, Map<String, String> params, Str
     }
 
     public static String getMatchedTool(Collection<String> tools, String action) {
+        String bestMatch = null;
+        String actionLower = action.toLowerCase(Locale.ROOT);
         for (String tool : tools) {
-            if (action.toLowerCase(Locale.ROOT).contains(tool.toLowerCase(Locale.ROOT))) {
+            // Exact match takes priority
+            if (action.equalsIgnoreCase(tool)) {
```
mingshl, line 509: if tool doesn't match, what is the default behavior>

dhrubo-os: @rithin-pullela-aws What happens if `get_order_history` is after `get_order_history_simple` in order?

rithinpullela: If there is an exact name match, we will directly take that tool (line 509, 510)

If there is no exact match, we will keep iterating till we get the best match. Ideally I would prefer the call to fail and go back to LLM saying tool name you provided is wrong. 

But to maintain backward compatibility I added another logic, if the tool name does not match exactly, we will keep the best match determined by logic below

Consider 2 tools
get_order
get_order_history

If the LLM asks for "get_order_history_recent" tool, there is no perfect match, so we will give the best match as "get_order_history"

The older code would have returned either of the tools based on the order.

While I disagree with such incomplete matching, I wanted to have the same behavior as before if there is no exact match.

Issue #2994, opened 2024-09-26
[FEATURE] JsonPath configuration


mingshl:
**Is your feature request related to a problem?**
Checking in ml-commons repo, there are [some classes](https://github.com/search?q=repo%3Aopensearch-project%2Fml-commons%20JSONPATh&type=code),  using JsonPath packages, and in JsonPath, there are some options to config json path differently, referring [here](https://opensearch.org/docs/latest/field-types/supported-field-types/flat-object/).

Options:

1. DEFAULT_PATH_LEAF_TO_NULL
This option makes JsonPath return null for missing leafs.
2.  ALWAYS_RETURN_LIST
This option configures JsonPath to return a list even when the path is definite.
3. SUPPRESS_EXCEPTIONS
This option makes sure no exceptions are propagated from path evaluation.
4. REQUIRE_PROPERTIES
This option configures JsonPath to require properties defined in path when an indefinite path is evaluated.

**What solution would you like?**
Now that is using default configuration, but if required for different use case, we should add open parameters for users to config using the JsonPath differently.

brianf-aws:
Hey, can you assign this to me for now?

PR #4520, opened 2026-01-05
Add JSON validation in StringUtils fromJson/toJson

jiapingzeng:
### Description
Address the below security finding by:
- switch from Gson to Jackson in StringUtils.fromJson and toJson for validation features
- add max size check, duplicate key check
- fix test failures due to small differences between Gson and Jackson json handling

```
What's wrong?
Location: Multiple files, including:

ml-commons/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
Unvalidated Input: The code deserializes JSON from response.toString() without validating the structure or content first. If the response comes from an untrusted source (like an external LLM API), malicious JSON could be injected.

Type Safety: The method returns Map<String, Object> where Object can be anything - nested maps, lists, primitives, etc. Without type checking, malicious payloads could exploit downstream code that assumes certain types.

No Schema Validation: There's no validation that the deserialized data conforms to expected schemas before it's used in operations like:

Vulnerable Patterns in This Code
Looking at the usage patterns:

// Line 328 - No validation before use
Map<String, Object> llmResponse = StringUtils.fromJson(response.toString(), RESPONSE_FIELD);

// Line 344 - Directly reading from unvalidated map
String llmFinishReason = JsonPath.read(llmResponse, llmFinishReasonPath.substring(...));

// Line 418 - Used in final answer without sanitization
modelOutput.put(FINAL_ANSWER, StringUtils.toJson(postFilterFinalAnswer(parameters, llmResponse)));

Vulnerable Patterns in This Code
Looking at the usage patterns:

Recommendations
The code should add:

Input validation before deserialization
Type checking after deserialization
Schema validation for expected structure
Sanitization of values used in tool execution or final outputs
This is especially important since this is agent/tool execution code that processes LLM outputs and executes tools based on that data.
```

### Related Issues

### Check List

## Release Notes

* **New Features**
  * Introduced configurable maximum JSON size limit (default: 100 MB, -1 for unlimited) for controlling JSON payload sizes in system operations
  * Enhanced JSON parsing with stricter duplicate key detection to improve data integrity and prevent processing errors

* **Improvements**
  * Strengthened JSON serialization and validation mechanisms for more robust data handling

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
+++ b/common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
@@ -38,6 +38,7 @@ import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
@@ -81,15 +82,14 @@ public class StringUtils {
 
     public static final String SAFE_INPUT_DESCRIPTION = "can only contain letters, numbers, spaces, and basic punctuation (.,!?():@-_'/\")";
 
-    // Maximum allowed JSON string size (configurable via plugins.ml_commons.max_json_size)
-    private static volatile int maxJsonSize;
+    private static volatile MLFeatureEnabledSetting mlFeatureEnabledSetting;
 
     /**
-     * Sets the maximum allowed JSON string size for parsing.
-     * @param size the maximum size, or -1 for unlimited
+     * Sets the MLFeatureEnabledSetting instance for accessing dynamic settings.
+     * @param setting the MLFeatureEnabledSetting instance
      */
-    public static void setMaxJsonSize(int size) {
-        maxJsonSize = size;
+    public static void setMLFeatureEnabledSetting(MLFeatureEnabledSetting setting) {
+        mlFeatureEnabledSetting = setting;
     }
 
     public static final Gson PLAIN_NUMBER_GSON = new GsonBuilder()
@@ -186,8 +186,11 @@ public class StringUtils {
             throw new IllegalArgumentException("JSON string cannot be null");
         }
         // Check size limit if configured (maxJsonSize = -1 means unlimited)
-        if (maxJsonSize > 0 && jsonStr.length() > maxJsonSize) {
-            throw new IllegalArgumentException(String.format("JSON string size exceeds maximum allowed size (%d characters)", maxJsonSize));
+        if (mlFeatureEnabledSetting != null) {
+            int maxJsonSize = mlFeatureEnabledSetting.getMaxJsonSize();
+            if (maxJsonSize > 0 && jsonStr.getBytes(StandardCharsets.UTF_8).length > maxJsonSize) {
+                throw new IllegalArgumentException(String.format("JSON string size exceeds maximum allowed size (%d bytes)", maxJsonSize));
+            }
         }
         try {
             JsonNode jsonNode = MAPPER.readTree(jsonStr);
--- a/common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
+++ b/common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
@@ -13,6 +13,8 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 import static org.opensearch.ml.common.utils.StringUtils.*;
 
 import java.io.IOException;
@@ -35,6 +37,7 @@ import org.opensearch.ml.common.output.model.MLResultDataType;
 import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
 
 import com.google.gson.JsonElement;
 import com.google.gson.TypeAdapter;
@@ -1276,13 +1279,16 @@ public class StringUtilsTest {
 
     @Test
     public void testFromJson_ExceedsMaxSize() {
-        // Set a small limit for testing
-        StringUtils.setMaxJsonSize(1000);
+        // Create a mock MLFeatureEnabledSetting with a small limit for testing
+        MLFeatureEnabledSetting mockSetting = mock(MLFeatureEnabledSetting.class);
+        when(mockSetting.getMaxJsonSize()).thenReturn(100);
+
+        StringUtils.setMLFeatureEnabledSetting(mockSetting);
 
         try {
-            // Create a JSON string larger than the limit
+            // Create a JSON string larger than the limit (100 bytes)
             StringBuilder largeJson = new StringBuilder("{\"data\":\"");
-            for (int i = 0; i < 1000; i++) {
+            for (int i = 0; i < 200; i++) {
                 largeJson.append("a");
             }
             largeJson.append("\"}");
@@ -1293,7 +1299,7 @@ public class StringUtilsTest {
             );
             assertTrue(exception.getMessage().contains("JSON string size exceeds maximum allowed size"));
         } finally {
-            StringUtils.setMaxJsonSize(100_000_000);
+            StringUtils.setMLFeatureEnabledSetting(null);
         }
     }
 
--- a/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
+++ b/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
@@ -647,14 +647,10 @@ public class MachineLearningPlugin extends Plugin
         Settings settings = environment.settings();
         Path dataPath = environment.dataFiles()[0];
 
-        // Initialize max JSON size from settings and set up listener for dynamic updates
-        StringUtils.setMaxJsonSize(MLCommonsSettings.ML_COMMONS_MAX_JSON_SIZE.get(settings));
-        clusterService
-            .getClusterSettings()
-            .addSettingsUpdateConsumer(MLCommonsSettings.ML_COMMONS_MAX_JSON_SIZE, StringUtils::setMaxJsonSize);
-
         mlFeatureEnabledSetting = new MLFeatureEnabledSetting(clusterService, settings);
         mlFeatureEnabledSetting.addListener(mlTaskManager);
+
+        StringUtils.setMLFeatureEnabledSetting(mlFeatureEnabledSetting);
 
         mlIndicesHandler = new MLIndicesHandler(clusterService, client, mlFeatureEnabledSetting);
 
--- a/common/src/main/java/org/opensearch/ml/common/settings/MLFeatureEnabledSetting.java
+++ b/common/src/main/java/org/opensearch/ml/common/settings/MLFeatureEnabledSetting.java
@@ -63,6 +63,8 @@ public class MLFeatureEnabledSetting {
 
     private volatile Boolean isStreamEnabled;
 
+    private volatile Integer maxJsonSize;
+
     private final List<SettingsChangeListener> listeners = new ArrayList<>();
 
     public MLFeatureEnabledSetting(ClusterService clusterService, Settings settings) {
@@ -83,6 +85,7 @@ public class MLFeatureEnabledSetting {
         isAgenticMemoryEnabled = ML_COMMONS_AGENTIC_MEMORY_ENABLED.get(settings);
         isIndexInsightEnabled = ML_COMMONS_INDEX_INSIGHT_FEATURE_ENABLED.get(settings);
         isStreamEnabled = ML_COMMONS_STREAM_ENABLED.get(settings);
+        maxJsonSize = MLCommonsSettings.ML_COMMONS_MAX_JSON_SIZE.get(settings);
 
         clusterService
             .getClusterSettings()
@@ -118,6 +121,7 @@ public class MLFeatureEnabledSetting {
                 listener.onStaticMetricCollectionEnabledChanged(it);
             }
         });
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(MLCommonsSettings.ML_COMMONS_MAX_JSON_SIZE, it -> maxJsonSize = it);
     }
 
     /**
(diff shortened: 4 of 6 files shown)


ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/McpSseToolTests.java
```diff
@@ -56,7 +56,7 @@ public void testRunSuccess() {
         tool.run(validParams, listener);
 
         // Assert
-        verify(listener).onResponse("[{\"annotations\":null,\"text\":\"{\\\"foo\\\":\\\"bar\\\"}\",\"meta\":null}]");
+        verify(listener).onResponse("[{\"text\":\"{\\\"foo\\\":\\\"bar\\\"}\"}]");
```
mingshl, line 59: can you explain why the test has different results on response, were the annotations and meta fields used elsewhere?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -168,18 +174,27 @@ public static String toUTF8(String rawString) {
...
-        JsonElement jsonElement = JsonParser.parseString(jsonStr);
-        if (jsonElement.isJsonObject()) {
-            result = gson.fromJson(jsonElement, Map.class);
-        } else if (jsonElement.isJsonArray()) {
-            List<Object> list = gson.fromJson(jsonElement, List.class);
-            result = new HashMap<>();
-            result.put(defaultKey, list);
-        } else {
-            throw new IllegalArgumentException("Unsupported response type");
+        if (jsonStr == null) {
```
mingshl, line 185: I am a bit worry, this `fromJson` method is used in multiple features. is 10MB big enough? please share how you come up with this number? it's always great to add limitation on the length control. but trying to doublecheck if 10MB is suitable for most of the prediction results. is there any existing configuration that we can reuse here?

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
```diff
@@ -101,7 +101,7 @@ public void fromJson_NestedMap() {
         assertEquals("nested_value", nestedMap.get("nested_key"));
         List list = (List) nestedMap.get("nested_array");
         assertEquals(2, list.size());
-        assertEquals(1.0, list.get(0));
```
brianf-aws, line 104: are decimal numbers no longer supported? This will affect ML Inference Query rewrite flows (neural sparse and embedding ones)

ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/McpSseToolTests.java
jiapingzeng: MCP Java SDK uses Jackson's `@JsonInclude(JsonInclude.Include.NON_ABSENT)` annotation which only includes non-null fields. Since we were using Gson before, `Include.NON_ABSENT` did not apply and null fields are included. The annotations and meta fields would still show up if they do not have null values.

https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java#L1532

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -168,18 +174,27 @@ public static String toUTF8(String rawString) {
...
+            throw new IllegalArgumentException("JSON string cannot be null");
+        }
+        if (jsonStr.length() > MAX_JSON_SIZE) {
+            throw new IllegalArgumentException(String.format("JSON string size exceeds maximum allowed size (%d bytes)", MAX_JSON_SIZE));
```
brianf-aws: Can we say characters instead? Since some chars have different byte size other than 1 e.g.  ¬∆¥ 

Used this as reference. https://mothereff.in/byte-counter

If not should we use a more direct way of getting bytes like this 

```
// Check encoded sizes
final byte[] utf8Bytes = string.getBytes("UTF-8");
System.out.println(utf8Bytes.length); // prints "11"
```
Source https://stackoverflow.com/questions/4385623/bytes-of-a-string-in-java

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -100,7 +104,9 @@ public class StringUtils {
     }
     public static final String TO_STRING_FUNCTION_NAME = ".toString()";
 
-    public static final ObjectMapper MAPPER = new ObjectMapper();
+    public static final ObjectMapper MAPPER = new ObjectMapper()
+        .configure(com.fasterxml.jackson.core.JsonParser.Feature.STRICT_DUPLICATE_DETECTION, true)
+        .configure(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY, true);
```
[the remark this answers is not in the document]
jiapingzeng: ObjectMapper provides JSON validation for invalid JSON, then these settings add validation for duplicate keys. 

E.g.: `{ "user": "test", "user": "admin" }` would give value user:admin without the setting and error with the setting.

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
jiapingzeng: Gson parses integers as doubles: https://stackoverflow.com/questions/45734769/why-does-gson-parse-an-integer-as-a-double; the input JSON string from this test is integer 1:

```
    @Test
    public void fromJson_NestedMap() {
        Map<String, Object> response = StringUtils
            .fromJson("{\"key\": {\"nested_key\": \"nested_value\", \"nested_array\": [1, \"a\"]}}", "response");
        assertEquals(1, response.size());
        assertTrue(response.get("key") instanceof Map);
        Map nestedMap = (Map) response.get("key");
        assertEquals("nested_value", nestedMap.get("nested_key"));
        List list = (List) nestedMap.get("nested_array");
        assertEquals(2, list.size());
        assertEquals(1, list.get(0));
        assertEquals("a", list.get(1));
    }
```

so 1 (instead of 1.0) is actually the correct result here. Jackson parses integers as integers.

jiapingzeng (commented):
-

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: 10MB is an arbitrary number but should be enough for all current generation of LLM models.

For example, Claude Sonnet 4.5 supports [max 64000 output tokens](https://www.anthropic.com/claude/sonnet). If we estimate 4 characters per token per [this guide](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them), that's 256000 characters or approx. 0.25MB. GPT-5.2 pro supports [max 128000 output tokens](https://platform.openai.com/docs/models/gpt-5.2-pro) which is approx. 0.5MB.

I am open to either increasing this value or making it configurable.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: Oh interesting! Updated exception message to say "characters" in new commit.

ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/McpSseToolTests.java
jiapingzeng: `NON_ABSENT` drops the null fields which is why this test needs to be updated.

common/src/test/java/org/opensearch/ml/common/utils/StringUtilsTest.java
brianf-aws: Thanks for taking the time to explain!

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
brianf-aws: Cool will approve once revised :)

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
mingshl: normal llm response, 10MB is enough, but considering batch prediction or streaming, the size might be bigger. Instead of hard limiting the length size. 

It would be great to be configurable. and have 10 mb as default.

we can either reuse the openSearch core setting, `http.max_content_length (default: 100MB)`, but that's a bit different than this purpose, if's a prediction result, this configuration is suitable. but when it's for search hit, which may contain multiple prediction response, this configuration might be too big. 

or we make a new configuration in ml-common for json, `ml.common.max_json_size` (default: 10MB). I remembered @ohltyler brought up open up the json setting in ml-commons once. https://github.com/opensearch-project/ml-commons/issues/2994

jiapingzeng: added setting `plugins.ml_commons.max_json_size` with default 100MB, 
or can specify -1 for unlimited json size

brianf-aws (approved):

brianf-aws:
Oh, lets apply spotless

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
dhrubo-os: What is the fall back mechanism? I know this is not gonna happen usually, but what if the output exceeds 100MB, what happens then? Do we show error or we just show the truncated output?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -168,18 +182,28 @@ public static String toUTF8(String rawString) {
...
-            List<Object> list = gson.fromJson(jsonElement, List.class);
-            result = new HashMap<>();
-            result.put(defaultKey, list);
-        } else {
-            throw new IllegalArgumentException("Unsupported response type");
+        if (jsonStr == null) {
+            throw new IllegalArgumentException("JSON string cannot be null");
+        }
+        // Check size limit if configured (maxJsonSize = -1 means unlimited)
+        if (maxJsonSize > 0 && jsonStr.length() > maxJsonSize) {
```
dhrubo-os: how this can work? This checks character count, not byte size. The setting is documented as "Maximum allowed JSON string size in bytes" but String.length() returns characters.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -168,18 +182,28 @@ public static String toUTF8(String rawString) {
...
+        try {
+            JsonNode jsonNode = MAPPER.readTree(jsonStr);
+            if (jsonNode.isObject()) {
+                return MAPPER.convertValue(jsonNode, Map.class);
+            } else if (jsonNode.isArray()) {
+                List<Object> list = MAPPER.convertValue(jsonNode, List.class);
+                Map<String, Object> result = new HashMap<>();
+                result.put(defaultKey, list);
+                return result;
+            } else {
```
dhrubo-os, line 204: What other type can we expect here?

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -646,6 +647,12 @@ public Collection<Object> createComponents(
         Settings settings = environment.settings();
         Path dataPath = environment.dataFiles()[0];
 
+        // Initialize max JSON size from settings and set up listener for dynamic updates
+        StringUtils.setMaxJsonSize(MLCommonsSettings.ML_COMMONS_MAX_JSON_SIZE.get(settings));
```
dhrubo-os: Can we leverage this class: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/settings/MLFeatureEnabledSetting.java ?

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: JsonNode has a few more types such as `isTextual`, `isNumber`, etc. However, since here `jsonStr` is always a full json document instead of a nested node, we're processing the outermost node which can only be `{}` object or `[]` array.

The processing logic is unchanged in this PR but replaced `gson.fromJson` with `MAPPER.convertValue`.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: updated setting comment to say string length and default 100M characters

dhrubo-os: I'm not sure if this setting is reasonable in that case. How do customer know how many characters in a response.

jiapingzeng: This is to prevent malicious payload as LLMs will not return more than 100M chars in its response. But if customers somehow have a valid use case where they expect a very large size response, they can increase or remove the limit via the setting.

Are you suggesting that we should use a size limit instead of chars limit? Or that we should not provide this as a configurable setting?

dhrubo-os: I was just wondering if we should be using this settings? https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/network-settings/ : `http.max_content_length`, or similar like this. 

I am suggesting size limit instead of chars limit.

jiapingzeng: `http.max_content_length` already has a 100MB size limit by default so large requests would be rejected there before reaching ml-commons. `max_json_size` is a JSON parsing setting which I think can exist separately to control ML-specific payloads.

Will update logic to calculate byte size instead.

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: It will error. Since the purpose of this setting is to prevent malicious input, I think it makes sense to error instead of trying to process it and truncate. If customer has a valid use case for 100MB+ JSON inputs in ML workloads, they can increase or disable `plugins.ml_commons.max_json_size`.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
jiapingzeng: added in new commit

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
jiapingzeng: switched to byte size in new commit

common/src/main/java/org/opensearch/ml/common/utils/StringUtils.java
```diff
@@ -168,18 +182,31 @@ public static String toUTF8(String rawString) {
...
+                throw new IllegalArgumentException("Unsupported response type");
+            }
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException("Invalid JSON format: " + e.getMessage(), e);
         }
-        return result;
     }
```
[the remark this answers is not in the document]
jiapingzeng: we were already throwing `IllegalArgumentException` in this method before this PR; I think we can address this suggestion as a future enhancement.

.gitignore
```diff
@@ -11,3 +11,6 @@ plugin/build/
 .DS_Store
 */bin/
 **/*.factorypath
+
+test-requests
```
mingshl: can you remove this? this looks like left out

jiapingzeng: oops good catch, did not mean to include this

mingshl (approved):

dhrubo-os (approved):

PR #4530, opened 2026-01-10
Fix Issue 4351: SearchIndexTool frequently failed

### LINKED ISSUE #4351: SearchIndexTool frequently failed
when testing failed in conversational agents, observed in logging, 
indexMappingTool frequently failed, which wasted llm call and token usage. 

Exception message: 
```
"{""role"":""assistant"",""content"":null,""tool_calls"":[{""id"":""call_j9ALOKavukeMN1M0odPNyUDd"",""type"":""function"",""function"":{""name"":""SearchIndexTool"",""arguments"":""{\""index\"":\""opensearch-release\"",\""query\"":\""{\\\""query\\\"":{\\\""match_all\\\"":{}}}\""}""}}],""refusal"":null,""annotations"":[]}"
"{""role"":""tool"",""tool_call_id"":""call_j9ALOKavukeMN1M0odPNyUDd"",""content"":""Tool SearchIndexTool failed: Expected [START_OBJECT] but found [VALUE_STRING]""}"```
```
Expected correct function calling 

```
[{""id"":""call_2u4yS00o05eQ5Ay6UDDWaD4x"",""type"":""function"",""function"":{""name"":""SearchIndexTool"",""arguments"":""{\""index\"":\""opensearch-release\"",\""query\"":{\""query\"":{\""match_all\"":
```

then it will return search result properly

```
"{""role"":""tool"",""content"":""{\""_index\"":\""opensearch-release\"",\""_source\"":{\""release_note\"":\""{\\\""next_page\\\"":\\\""[https://duckduckgo.com/html?q\\\\u003dOpenSearch+releases+after+version+2.0\\\\u0026s\\\\u003d10\\\\u0026nextParams\\\\u003d\\\\u0026v\\\\u003dl\\\\u0026o\\\\u003djson\\\\u0026dc\\\\u003d11\\\\u0026api\\\\u003dd.js\\\\u0026vqd\\\\u003d4-302390017373502469848272537024929795231\\\\u0026kl\\\\u003dwt-wt\\\\\\](https://duckduckgo.com/html?q%5C%5C%5C%5Cu003dOpenSearch+releases+after+version+2.0%5C%5C%5C%5Cu0026s%5C%5C%5C%5Cu003d10%5C%5C%5C%5Cu0026nextParams%5C%5C%5C%5Cu003d%5C%5C%5C%5Cu0026v%5C%5C%5C%5Cu003dl%5C%5C%5C%5Cu0026o%5C%5C%5C%5Cu003djson%5C%5C%5C%5Cu0026dc%5C%5C%5C%5Cu003d11%5C%5C%5C%5Cu0026api%5C%5C%5C%5Cu003dd.js%5C%5C%5C%5Cu0026vqd%5C%5C%5C%5Cu003d4-302390017373502469848272537024929795231%5C%5C%5C%5Cu0026kl%5C%5C%5C%5Cu003dwt-wt%5C%5C%5C%5C%5C%5C)"",\\\""items\\\"":[{\\\""url\\\"":\\\""[https://opensearch.org/releases/\\\\\\](https://opensearch.org/releases/%5C%5C%5C%5C%5C%5C)"",\\\""title\\\"":\\\""\\\"",\\\""content\\\"":\\\""Skip to main content Search Close Search search Menu About Foundation Governing Board Charter Meeting Calendar Platform Platform OpenSearch CoreOpenSearch is a powerful search and analytics engine built on Apache Lucene. OpenSearch DashboardsOur data visualization toolset is a flexible, fully integrated solution for visually exploring and querying your data. OpenSearch Data PrepperA server-side data collector designed to enrich, transform, and aggregate data for downstream analytics with Opensearch. OpenSearch Vector EnginePower machine learning and generative AI applications with a high-performance database designed for the way AI understands data. Capabilities Machine learning and AI Vector Search Anomaly Detection Search E-Commerce Document Search Search Relevance Observability Performance …..
```

The exception is due to the SearchIndexTool is expected a proper escape query json string, most LLM tends to not able to write proper escape at the first time, and it might try second or third time to write the proper the query string in json format. 

There are two ways to fix this, 
1. in the "query " parameters on the searchIndexTool, check on the format and try to parse it properly when it's malformed by llm to prevent the exception from happening, this save one llm call
2. change the "query" parameters format to be an easy understanding format, for example, json format instead of json string. 

feel free to propose more solutions.

ishag4:
### Description
Fix SearchIndexTool JSON parsing to reduce wasted LLM calls and token usage

Problem: LLMs frequently generate malformed JSON queries causing SearchIndexTool to fail with 'Expected [START_OBJECT] but found [VALUE_STRING]' errors. This forces LLMs to retry multiple times, wasting tokens and degrading user experience.

Root Cause: LLM-generated queries like: "query":"{\"query\":{\"match_all\":{}}}" contain improperly escaped JSON that fails to parse.

Impact: Eliminates need for LLM retries, reduces token consumption, and enables seamless conversational agent integration with SearchIndexTool.

Before: LLM fails -> retries 2-3 times -> wastes tokens -> poor UX After: LLM query auto-normalized -> succeeds first time -> optimal token usage

Solution:

Add a framework-level ToolArgumentValidator that validates and normalizes LLM tool arguments against the tool's declared input schema before the tool executes.
Normalize stringified JSON values in place (normalizeJsonObject()), so a value like "query":"{\\"query\\":{\\"match_all\\":{}}}" becomes a real JSON object.
Unwrap outer-quoted JSON via unwrapStringifiedJson(). The unwrap is only accepted when the unwrapped content parses as valid JSON; otherwise the original input is kept.
Add a supportsStrictSchema() capability to FunctionCalling (true for OpenaiV1ChatCompletionsFunctionCalling, false by default) so MLChatAgentRunner applies framework validation only for providers without native strict-schema enforcement.
Update SearchIndexTool: DEFAULT_INPUT_SCHEMA now documents query as an object, STRICT_FIELD defaults to true, string queries still work for backward compatibility (normalizeQueryParameter()), and parse failures surface a clear ParsingException / IllegalArgumentException instead of a raw GSON error.

### Related Issues
Resolves #4351

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust JSON query handling: automatic normalization and conservative repairs for malformed, extra‑braced, or LLM-generated stringified queries; clearer parsing errors and improved resilience when building searches.

* **Documentation**
  * Input guidance updated to prefer object-based OpenSearch Query DSL while retaining backward-compatible string handling.

* **New Features**
  * Exposed default tool attributes to simplify schema and strict-mode defaults.

* **Tests**
  * Extensive new tests covering normalization, malformed/edge cases, and real-world LLM query patterns.


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -322,4 +475,4 @@ public Map<String, Object> getDefaultAttributes() {
             return DEFAULT_ATTRIBUTES;
         }
     }
-}
+}
```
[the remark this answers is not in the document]
ishag4: Fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,6 +134,142 @@ public boolean validate(Map<String, String> parameters) {
...
+            return PLAIN_NUMBER_GSON.toJson(parsed);
+        } catch (JsonSyntaxException e) {
+            log.debug("Initial query parsing failed, attempting to fix common LLM formatting issues: {}", e.getMessage());
+        }
+
+        String fixedQuery = queryString;
+
+        fixedQuery = fixedQuery.replaceAll("\\\\\\\\\"", "\\\\\"");
+
+        fixedQuery = fixedQuery.replaceAll("\\\\\"", "\"");
```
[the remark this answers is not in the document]
dhrubo-os: Let's address this comment. Otherwise it can break the valid json string:

```
json
{"name": "foo\"bar"}  // Valid JSON

Would become:
json
{"name": "foo"bar"}   // Invalid JSON
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,6 +134,142 @@ public boolean validate(Map<String, String> parameters) {
...
+            try {
+                Object parsed = PLAIN_NUMBER_GSON.fromJson(unwrapped, Object.class);
+                log.info("Successfully fixed stringified JSON query by removing outer quotes");
+                return PLAIN_NUMBER_GSON.toJson(parsed);
+            } catch (JsonSyntaxException ignored) {
+                // Keep original if unwrapping doesn't work
+            }
+        }
+
+        if (fixedQuery.startsWith("{") && fixedQuery.contains("\\\"") && fixedQuery.endsWith("}")) {
```
dhrubo-os: Is this piece of code reachable? Can we add test for this?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -212,7 +355,17 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
                 return;
             }
 
-            SearchRequest searchRequest = getSearchRequest(index, query);
+            SearchRequest searchRequest;
+            try {
+                searchRequest = getSearchRequest(index, query);
+            } catch (Exception e) {
+                log.error("Failed to parse search query. Index: {}, Query: {}", index, query, e);
```
dhrubo-os: Printing query could be security issue. 

log.error("Failed to parse search query. Index: {}", index, e);

ylwu-amzn:
@yuye-aws , Can you also help review this PR ?

yuye-aws:
> @yuye-aws , Can you also help review this PR ?

OK

xluo-aws:
@Hailong-am , @xinyual , please help review

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +131,197 @@ public boolean validate(Map<String, String> parameters) {
...
+
+        while (fixed.endsWith("}")) {
+            int braceBalance = countBraces(fixed);
+            if (braceBalance == 0) {
+                break;
+            } else if (braceBalance < 0) {
+                fixed = fixed.substring(0, fixed.length() - 1);
+                log.debug("Removed extra closing brace from malformed JSON");
+            } else {
+                break;
```
xinyual: Here means we have too many `{`, do we need to raise error here? Or can we manually add `{`?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +131,197 @@ public boolean validate(Map<String, String> parameters) {
...
+            log.warn("Could not auto-fix malformed query (length: {}), using original", queryString.length());
+            return queryString;
+        }
+    }
+
+    /**
+     * Fixes escape sequences while preserving legitimate internal quotes in JSON strings.
+     * Only applies aggressive fixes outside of quoted string values.
+     */
+    private String fixEscapeSequencesPreservingInternalQuotes(String input) {
```
xinyual: Not quit understand the logic. But can we use StringEscapeUtils.unescapeJson instead?

ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/SearchIndexToolTests.java
```diff
@@ -520,4 +521,183 @@ public void testRun_withRangeQuery_triggersPlainDoubleGson() {
...
+        // Test the specific LLM-generated malformed JSON pattern that was failing
+        String malformedInput = "{\"index\":\"test-index\",\"query\":\"{\\\"query\\\":{\\\"match_all\\\":{}}}}}\"}";
+        Map<String, String> parameters = Map.of("input", malformedInput);
+
+        ActionListener<String> listener = mock(ActionListener.class);
+        mockedSearchIndexTool.run(parameters, listener);
+
+        // Verify that the search was executed successfully (no failure callback)
+        verify(listener, never()).onFailure(any());
+        verify(client, times(1)).search(any(), any());
```
Hailong-am: can we verify the normalized dsl query match our expectation in all your unit test?  Or we can you have separate test for normalized part.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +131,197 @@ public boolean validate(Map<String, String> parameters) {
...
+    }
+
+    /**
+     * Basic heuristic to detect if the string might contain legitimate internal quotes.
+     * Looks for patterns like "field": "value with \"quotes\""
+     */
+    private boolean containsLegitimateInternalQuotes(String input) {
+        // Look for pattern: "key": "value with \"internal\" quotes"
+        return input.matches(".*\"[^\"]*\"\\s*:\\s*\"[^\"]*\\\\\"[^\"]*\".*");
+    }
```
yuye-aws: There could be missing edge cases. The containsLegitimateInternalQuotes function is checking whether the whole string contains legitimate internal quotes. However, if a string contains, perhaps some substring may need the aggressive fix. Can you check with this edge case:

```
{"field_1": "value with \"quotes\"", \\"field_2\\", "value"}
```

In this example, we need aggressive fix but containsLegitimateInternalQuotes returns true

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +131,197 @@ public boolean validate(Map<String, String> parameters) {
...
+        if (fixedQuery.startsWith("\"") && fixedQuery.endsWith("\"") && fixedQuery.length() > 1) {
+            String unwrapped = fixedQuery.substring(1, fixedQuery.length() - 1);
+            try {
+                Object parsed = PLAIN_NUMBER_GSON.fromJson(unwrapped, Object.class);
+                log.info("Successfully fixed stringified JSON query by removing outer quotes");
+                return PLAIN_NUMBER_GSON.toJson(parsed);
+            } catch (JsonSyntaxException ignored) {
+                // Continue with other fixes
+            }
+        }
```
yuye-aws: Could be multiple outer quotes?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +131,197 @@ public boolean validate(Map<String, String> parameters) {
...
+        String fixed = input.trim();
+
+        while (fixed.endsWith("}")) {
+            int braceBalance = countBraces(fixed);
+            if (braceBalance == 0) {
+                break;
+            } else if (braceBalance < 0) {
+                fixed = fixed.substring(0, fixed.length() - 1);
+                log.debug("Removed extra closing brace from malformed JSON");
+            } else {
```
yuye-aws: The while loop could possibly executes multiple times. If braceBalance is -5, it unnecessarily counts brackets 5 times.

ishag4:
Hi @dhrubo-os @xinyual @yuye-aws @Hailong-am, Could you please re-review?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +136,238 @@ public boolean validate(Map<String, String> parameters) {
...
+            String unwrapped = result.substring(1, result.length() - 1);
+            try {
+                // Test if the unwrapped content is valid JSON
+                PLAIN_NUMBER_GSON.fromJson(unwrapped, Object.class);
+                result = unwrapped;
+                layersRemoved++;
+            } catch (JsonSyntaxException e) {
+                // If unwrapping breaks JSON, stop here
+                break;
+            }
```
yuye-aws: Can you check the time complexity of the **substring**? If this is O(n), the overall time complexity could be O(n * m), where m stands for the number of outer quotes.

If I understand this correctly, you can simply check the value of m and finish with a single **substring** operation. Is there any use case to require multiple checks?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +136,238 @@ public boolean validate(Map<String, String> parameters) {
...
+        if (StringUtils.isEmpty(input)) {
+            return input;
+        }
+
+        String fixed = input.trim();
+        int braceBalance = countBraces(fixed);
+
+        if (braceBalance < 0) {
+            // Remove all extra closing braces from the end in a single pass
+            int extraClosing = Math.abs(braceBalance);
```
yuye-aws: [int] int extraClosing = -braceBalance;

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +136,238 @@ public boolean validate(Map<String, String> parameters) {
...
+            int endIndex = fixed.length();
+            while (extraClosing > 0 && endIndex > 0 && fixed.charAt(endIndex - 1) == '}') {
+                endIndex--;
+                extraClosing--;
+            }
```
yuye-aws: Why not simply check the number of "}" at the end of string and then minus endIndex?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -131,10 +136,238 @@ public boolean validate(Map<String, String> parameters) {
...
+            fixed = fixed.substring(0, endIndex);
+            log.debug("Removed {} extra closing brace(s) from malformed JSON", Math.abs(braceBalance) - extraClosing);
+
+        } else if (braceBalance > 0) {
+            // Unbalanced: more opening braces
+            log.debug("Unbalanced JSON: missing {} closing brace(s), cannot safely auto-fix without changing semantics", braceBalance);
+            return input;
```
yuye-aws: Check the beginning of string?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: The loop intentionally removes only the number of extra closing braces required to rebalance (`braceBalance < 0`), rather than all trailing `}`. This avoids over-removal when the string is partially valid and keeps the fix minimal.

For example, with `{"a":1}}}}` (braceBalance = -3), we remove exactly 3 closing braces, not all trailing ones. This preserves any valid structure and handles edge cases where `}` might appear in unexpected positions.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: We intentionally do not attempt to auto-add missing closing braces. Adding structure requires guessing intent and can silently change query semantics, which is especially risky for OpenSearch DSL queries.

The conservative approach of logging and returning the original input ensures we fail fast rather than potentially executing a different query than intended. This aligns with the defensive design principle throughout the JSON normalization logic.

Please let me know your thoughts on this.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: Good point on substring complexity. In practice, the number of outer quote layers from LLM output is very small (typically ≤2), and each iteration is guarded by a JSON parse which already dominates runtime.

Although I optimized this by counting consecutive quotes from both prefix and suffix, taking the minimum, and then doing a single substring operation.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: Agreed. Updated in the latest commit.

ishag4:
Hi @yuye-aws 
Could you please re-review?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
yuye-aws: Well, I mean you can simply saves this O(n) loop.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
yuye-aws: Make sense

yuye-aws:
@ishag4 Just replied to one comment. Other part looks good to me. @dhrubo-os Do you have any unresolved comment?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: Thanks for clarifying.

The key constraint here is that we should remove only the required number of extra closing braces (based on braceBalance), not all trailing }. That’s why I used the loop to stop once we’ve removed exactly -braceBalance.

That said, you’re right we can avoid the iterative loop and still preserve the same behavior by first counting trailing } and then trimming min(trailing, -braceBalance). (if I understood correctly)

Refactored it!

ishag4:
Hi @yuye-aws, I’ve updated the comments and made changes based on my understanding. Could you please take another look?

Hi @dhrubo-os, if everything looks good, we can go ahead and merge this PR.

yuye-aws:
Hi @ishag4 . No major comments from my side. @dhrubo-os Do you have any comments?

yuye-aws (approved):

ishag4:
Hi @yuye-aws Are we good to merge this PR?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -177,20 +442,21 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
             String index = null;
             String query = null;
             boolean returnFullResponse = Boolean.parseBoolean(parameters.getOrDefault(RETURN_RAW_RESPONSE, "false"));
+
             if (!StringUtils.isEmpty(input)) {
                 try {
-                    JsonObject jsonObject = GSON.fromJson(input, JsonObject.class);
+                    String fixedInput = fixMalformedJson(input);
```
mingshl:
``` 
String fixedInput = fixMalformedJson(input);
JsonObject jsonObject = GSON.fromJson(fixedInput, JsonObject.class);
  ```
  

This applies brace-balancing to the entire input JSON (which includes index + query), not
  just the query string. If the outer JSON has extra braces due to an LLM error, removing
  trailing } could corrupt the index field or other parameters. The fix should only be applied
  to the query portion after extraction, or at minimum validate that the result still parses to
  a JsonObject with the expected fields

Comment (ishag4) [reply ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java]: Good catch, thanks. I updated the implementation to avoid applying `fixMalformedJson()` on the entire input payload.

File: ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java [hunk 16]
```diff
@@ -71,11 +72,15 @@ public class SearchIndexTool implements Tool {
...
 
     public static final String DEFAULT_INPUT_SCHEMA = "{\"type\":\"object\","
-        + "\"properties\":{\"index\":{\"type\":\"string\",\"description\":\"OpenSearch index name. for example: index1\"},"
-        + "\"query\":{\"type\":\"object\",\"description\":\"OpenSearch search index query. You need to get index mapping to write correct search query. It must be a valid OpenSearch query."
-        + " Valid value:\\n{\\\"query\\\":{\\\"match\\\":{\\\"population_description\\\":\\\"seattle 2023 population\\\"}},\\\"size\\\":2,\\\"_source\\\":\\\"population_description\\\"}\\n"
-        + "Invalid value: \\n{\\\"match\\\":{\\\"population_description\\\":\\\"seattle 2023 population\\\"}}\\nThe value is invalid because the match not wrapped by \\\"query\\\".\","
-        + "\"additionalProperties\":false}},\"required\":[\"index\",\"query\"],\"additionalProperties\":false}";
+        + "\"properties\":{"
+        + "\"index\":{\"type\":\"string\",\"description\":\"OpenSearch index name. Example: index1\"},"
+        + "\"query\":{\"type\":\"object\",\"description\":\"OpenSearch Query DSL as a JSON object. "
```
rithinpullela: Looks like we removed "  `You need to get index mapping to write correct search query.`

Can we add this, because in practice we found the LLM tries to use this tool without getting the index mapping and falls into a retry loop which can be avoided if we can get the index mapping first.

rithinpullela (approved):
Thanks for the PR, this would surely solve one of the biggest problems of using this tool! 

Added a small nit, but approving the PR nonetheless since my comment is not blocking

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
ishag4: Added @rithinpullela , thank you for the review!

Hailong-am:
## Question: Is this complexity still needed as LLMs improve?

Modern LLMs (Claude, GPT-4o, Gemini) now support structured output / function calling that guarantees schema-valid JSON. Given that the schema already declares `query` as `"type":"object"`, a capable model will produce a proper object — not an escaped string.

### `strict` field is already wired for OpenAI

I checked the codebase — `strict` is not just metadata. In `OpenaiV1ChatCompletionsFunctionCalling.java`:

```java
"\"strict\": ${tool.attributes.strict:-false}"
```

This maps directly to OpenAI's [Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs) — the API **guarantees** the model output conforms to the schema. With `strict: true` and `query` declared as `"type":"object"`, the malformed-string-instead-of-object problem **cannot happen** on the OpenAI path.

However, this only applies to OpenAI-compatible providers:
- **Bedrock Converse** — no `strict` in template, just passes `inputSchema`
- **Gemini** — no `strict` in template either

### Concerns with the normalization approach

1. **Risks altering query semantics** — `StringEscapeUtils.unescapeJson()` can change the meaning of queries that legitimately contain escaped characters. Silently "fixing" input is harder to debug than a clear error + retry.

2. **Maintenance cost vs. shrinking problem** — This adds ~250 lines of heuristic JSON-fixing code + 800 lines of tests for an edge case that diminishes with every model generation. The retry-on-error contract already exists in the agent framework for exactly this purpose.

3. **Individual tools shouldn't compensate for framework gaps** — If Bedrock/Gemini don't enforce schemas, that should be fixed at the framework level (e.g., adding structured output support for those providers), not patched per-tool.

### Suggested alternative

- For OpenAI path: flip `STRICT_FIELD` to `true` (one-line fix, guaranteed enforcement)
- Keep the improved schema description from this PR (genuinely helpful for all providers)
- For Bedrock/Gemini: consider adding schema enforcement at the framework level

### Question

Which LLM provider / connector was producing the malformed queries in issue #4351? If it was OpenAI-compatible, `strict: true` alone would fully resolve this.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/SearchIndexTool.java
```diff
@@ -71,11 +72,16 @@ public class SearchIndexTool implements Tool {
...
+        + "You need to get index mapping to write correct search query. "
+        + "The object MUST follow OpenSearch Query DSL and MUST include a top-level 'query' field. "
+        + "Preferred format for reliable parsing. "
+        + "Example: {\\\"query\\\":{\\\"match\\\":{\\\"field\\\":\\\"value\\\"}},\\\"size\\\":10}. "
+        + "String format is also supported for backward compatibility, but object format is strongly recommended.\"}},"
+        + "\"required\":[\"index\",\"query\"],"
+        + "\"additionalProperties\":false}";
```
Hailong-am: nit:  we can use text blocks for better reading we're using jdk 21 now.

ishag4: added

ishag4:
Hi @Hailong-am
Thanks for the input. I agree with the concern that tool-specific repair logic would become unnecessary (and potentially risky) as provider-side structured outputs mature.

I've updated the implementation accordingly:

* Enabled `STRICT_FIELD=true` for `SearchIndexTool`, so providers that support strict schema enforcement (OpenAI) rely on native guarantees rather than local repair logic.
* Added `supportsStrictSchema()` as a provider capability, so enforcement is handled at the framework level instead of inside individual tools.
* OpenAI now bypasses additional normalization and relies on schema guarantees.
* Bedrock/Gemini continue using framework-level validation because they currently don't provide equivalent strict enforcement.
* Reduced the normalization behavior to conservative, intent-preserving handling only (mainly stringified JSON cases), avoiding broad transformations such as aggressive unescaping that could alter query semantics.
* Kept the schema improvements because they help guide all providers regardless of enforcement capability.

The goal of the change is to move responsibility away from tool-specific workarounds and toward a capability-driven framework approach while maintaining compatibility for providers that don't yet guarantee schema adherence.

Hailong-am (approved):

Hailong-am (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/BedrockConverseFunctionCalling.java
```diff
@@ -209,6 +209,11 @@ public TokenUsage extractTokenUsage(Map<String, ?> llmResponseDataAsMap) {
         }
     }
 
+    @Override
+    public boolean supportsStrictSchema() {
```
akolarkunnu: Default implementation returns false only. Then overriding and returning false won't make any differnece

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolArgumentValidator.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        String trimmed = value.trim();
+        return (trimmed.startsWith("{") && trimmed.endsWith("}")) || (trimmed.startsWith("[") && trimmed.endsWith("]"));
+    }
+
+    /**
+     * Validation result containing the outcome and any error details.
+     */
+    public static class ValidationResult {
```
akolarkunnu: Where is this class getting used?

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolArgumentValidator.java
```diff
@@ -0,0 +1,191 @@
...
+
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * Validates and normalizes tool arguments against JSON schemas.
+ * Provides conservative normalization that preserves semantic meaning
+ * while fixing common LLM output formatting issues.
+ */
+@Log4j2
+public class ToolArgumentValidator {
```
akolarkunnu: I don't see any specific state for this class, this is a kind of utility class with one public API "validateAndNormalize". If that's the case you can change this class as a static class.

ml-algorithms/src/test/java/org/opensearch/ml/engine/function_calling/FunctionCallingCapabilityTests.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    public void testFunctionCalling_defaultSupportsStrictSchema() {
+        // Test the default implementation
+        FunctionCalling defaultImpl = new FunctionCalling() {
+            @Override
+            public void configure(java.util.Map<String, String> params) {}
+
+            @Override
+            public java.util.List<java.util.Map<String, String>> handle(
```
akolarkunnu: use imports

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ToolArgumentValidator.java
ishag4: removed

ishag4:
Hi @mingshl 
Could you please review this PR? 
This is open for months now

mingshl (approved):

mingshl:
please fix the test

```
* What went wrong:
Execution failed for task ':opensearch-ml-algorithms:compileTestJava'.
> Compilation failed; see the compiler output below.
  Note: Recompile with -Xlint:deprecation for details.
  /__w/ml-commons/ml-commons/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/ToolArgumentValidatorTests.java:248: error: cannot find symbol
          assertTrue("Status should be an array", terms.get("status") instanceof List);
                                                                                 ^
    symbol:   class List
    location: class ToolArgumentValidatorTests
  /__w/ml-commons/ml-commons/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/SearchIndexToolTests.java:905: error: cannot find symbol
          Field instanceField = SearchIndexTool.Factory.class.getDeclaredField("INSTANCE");
          ^
    symbol:   class Field
    location: class SearchIndexToolTests

```

ishag4:
Hi @mingshl Could you please trigger the pipelines again?

mingshl:
this IT is also flaky 

```
»  	at java.base/java.lang.Thread.run(Thread.java:1583)
Error: -07-08T23:44:48,684][ERROR][o.o.m.e.a.a.MLAgentExecutor] [integTest-0] Agent execution failed. agentType=flow, agentId=P10eRJ8BFxm6wV4urE8g, tenantId=null, latencyMs=0, statusCode=unknown, error=SearchIndexTool's two parameters: index and query are required and should be in valid format
»  java.lang.IllegalArgumentException: SearchIndexTool's two parameters: index and query are required and should be in valid format

```

ishag4:
Hi @mingshl Please re-run

AirFunnyB:
Hey @ishag4, the description lists fixMalformedJson()/countBraces(), but those got removed in the e2cbbde27 refactor when it was consolidated into unwrapStringifiedJson can you update the description.

ishag4:
Hi @AirFunnyB Sure

PR #4537, opened 2026-01-14
fix tool result parsing in AGUI agent

jiapingzeng:
AGUI agents include both frontend and backend tools. When LLM makes a backend tool call followed by a frontend tool call, the agent would fail as it did not process the first tool result. This change ensures that
- all tool results are processed
- empty inputs are properly converted to empty JSON objects

---

Bedrock only expects tool results for the tools used in the most recent turn, so this logic filters out those tool results from the rest. Previously we only retrieved the latest one tool use, but a turn can have multiple tool use. This showed up when LLM made parallel tool use which we fixed previously by skipping tool use after the first in the same turn: https://github.com/opensearch-project/ml-commons/pull/4526/commits/4e46b6b21b8981588438cc767b7d8f4ecc5164a5. The previous fix addressed most of the Bedrock format errors.

A corner case that was not addressed in the previous fix was that when the LLM does not add any thought message between the tool uses, and the tool use has a mix of backend and frontend tools, the frontend will send a request like this after it completes the frontend tool execution:

```
        {
            "id": "msg_8244372463205",
            "role": "assistant",
            "content": "I see there's an issue with the timestamp field. Let me first check the mapping of the OTEL logs index to understand the correct field structure:",
            "toolCalls": [
                {
                    "id": "tooluse_ituGO_T7TCiUoiCFbTgG4Q",
                    "type": "function",
                    "function": {
                        "name": "ListIndexTool",
                        "arguments": "..."
                    }
                },
                {
                    "id": "tooluse_bNOEhvgaR7KV2vvtaQdg0Q",
                    "type": "function",
                    "function": {
                        "name": "execute_ppl_query",
                        "arguments": "..."
                    }
                }
            ]
        },
        {
            "id": "tool-result-tooluse_ituGO_T7TCiUoiCFbTgG4Q",
            "role": "tool",
            "content": "...",
            "toolCallId": "tooluse_ituGO_T7TCiUoiCFbTgG4Q"
        },
        {
            "id": "msg-1767996275334-nph503d6s",
            "role": "tool",
            "content": "...",
            "toolCallId": "tooluse_bNOEhvgaR7KV2vvtaQdg0Q"
        }
    ]
```

Which has 2 tool results in one turn like parallel tool use. This PR will process all tool results after the lastest `assistant` role message which ensures that each tool call in the assistant message has a matching tool result.

---

### Description
When LLM makes a backend tool call followed by a frontend tool call, the agent would fail as it did not process the first tool result. This change will ensure that all tool results are processed.

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAGUIAgentRunner.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAGUIAgentRunner.java
@@ -280,10 +280,6 @@ public class MLAGUIAgentRunner implements MLAgentRunner {
                     }
                 }
 
-                if (recentToolResults.isEmpty()) {
-                    recentToolResults = new ArrayList<>(toolResults);
-                }
-
                 String toolResultsJson = gson.toJson(recentToolResults);
                 params.put(AGUI_PARAM_TOOL_CALL_RESULTS, toolResultsJson);
 


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAGUIAgentRunner.java
```diff
@@ -266,9 +266,23 @@ private void processAGUIMessages(Map<String, String> params, String llmInterface
...
+                    for (int i = 0; i < toolResultMessageIndices.size(); i++) {
+                        int toolResultIndex = toolResultMessageIndices.get(i);
+                        if (toolResultIndex > lastToolCallIndex) {
+                            recentToolResults.add(toolResults.get(i));
+                        }
+                    }
+                }
+
+                if (recentToolResults.isEmpty()) {
+                    recentToolResults = new ArrayList<>(toolResults);
```
mingshl: if the last tool result is not getting back, the recentToolResults can be empty? can you explain a bit how this logic can ensure every tool input is followed by a tool response in such format?

jiapingzeng: Bedrock only expects tool results for the tools used in the most recent turn, so this logic filters out those tool results from the rest. Previously we only retrieved the latest one tool use, but a turn can have multiple tool use. This showed up when LLM made parallel tool use which we fixed previously by skipping tool use after the first in the same turn: https://github.com/opensearch-project/ml-commons/commit/4e46b6b21b8981588438cc767b7d8f4ecc5164a5. The previous fix addressed most of the Bedrock format errors.

A corner case that was not addressed in the previous fix was that when the LLM does not add any thought message between the tool uses, and the tool use has a mix of backend and frontend tools, the frontend will send a request like this after it completes the frontend tool execution:
```
        {
            "id": "msg_8244372463205",
            "role": "assistant",
            "content": "I see there's an issue with the timestamp field. Let me first check the mapping of the OTEL logs index to understand the correct field structure:",
            "toolCalls": [
                {
                    "id": "tooluse_ituGO_T7TCiUoiCFbTgG4Q",
                    "type": "function",
                    "function": {
                        "name": "ListIndexTool",
                        "arguments": "..."
                    }
                },
                {
                    "id": "tooluse_bNOEhvgaR7KV2vvtaQdg0Q",
                    "type": "function",
                    "function": {
                        "name": "execute_ppl_query",
                        "arguments": "..."
                    }
                }
            ]
        },
        {
            "id": "tool-result-tooluse_ituGO_T7TCiUoiCFbTgG4Q",
            "role": "tool",
            "content": "...",
            "toolCallId": "tooluse_ituGO_T7TCiUoiCFbTgG4Q"
        },
        {
            "id": "msg-1767996275334-nph503d6s",
            "role": "tool",
            "content": "...",
            "toolCallId": "tooluse_bNOEhvgaR7KV2vvtaQdg0Q"
        }
    ]
```
Which has 2 tool results in one turn like parallel tool use. This PR will process all tool results after the lastest assistant role message which ensures that each tool call in the assistant message has a matching tool result. 

We don't expect `recentToolResults` to be empty here unless the request is malformed (when there is no tool result after a tool use assistant message but has one before it), so this particular if statement was fallback logic to try to grab previous tool result and pass it in latest turn. Removed in new commit as it is not necessary.

ylwu-amzn:
Thanks for fixing the bug. Have you tested by deploying to some test env?

Update: confirmed with @jiapingzeng this already tested. The failure case passed now.

PR #3919, opened 2025-06-17
[FEATURE] Improve EncryptorImpl with Asynchronous Handling for Scalability

akolarkunnu:
### Description
Removed CountDownLatch completely and now it is purely based on action listeners.
Along with this, fixed the issue of duplicate master key generation. If key generation for a tenant is in progress and another request come for encryption/decryption with same tenant id, it will again try to generate another master key , because old request is in the process of creating key and not yet completed. So there is a chance of creating duplicate keys for single tenant.

Fix : Removed CountDownLatch completely and now it is purely based on action listeners.
Storing all tenants who are waiting for the key generation in the map tenantWaitingListenerMap. Whenever the key generation completes or error happened for a tenant, notify all requestors waiting for that tenant key generation.
Now both encrypt and decrypt APIs accepts list of Strings as parameter to encrypt/decrypt. Previously it was single String object.

Testing:
Added more test cases with multi threaded use cases.
Also manually tested single and multi tenant use cases :
  - Register the model and invoke predict API - This cover both encryption and decryption cycle.
  - Tested error scenarios by mocking errors. It's behaving as previously
  - Tested With single and multiple credentials
  - Tested Agents with MCP Connector

Commands used for testing(eg: Multi tenant):

Register the model:
curl -XPOST "http://localhost:9200/_plugins/_ml/models/_register" \
  -H 'Content-Type: application/json' \
  -H 'x-tenant-id: 1234567' \
  -d '{
    "name": "My OpenAI model: gpt-5",
    "function_name": "remote",
    "description": "test model",
    "connector": {
        "name": "My openai connector: gpt-5",
        "description": "The connector to openai chat model",
        "version": 1,
        "protocol": "http",
        "parameters": {
            "model": "gpt-5"
        },
        "credential": {
            "openAI_key": "..."
        },
        "actions": [
            {
                "action_type": "predict",
                "method": "POST",
                "url": "url",
                "headers": {
                    "Authorization": "Bearer ${credential.openAI_key}"
                },
                "request_body": "{ \"model\": \"${parameters.model}\", \"messages\": [{\"role\":\"developer\",\"content\":\"${parameters.system_prompt}\"},${parameters._chat_history:-}{\"role\":\"user\",\"content\":\"${parameters.user_prompt}\"}${parameters._interactions:-}], \"user\": \"abdulmun\", \"reasoning_effort\":\"minimal\"${parameters.tool_configs:-}}"
            }
        ]
    }
}'

Invoking PREDICT API:
[...]

PR #4543, opened 2026-01-14
[BUG FIX] Add expiry time to master key cache to prevent stale encryption key usage

### LINKED ISSUE #4542: [BUG] Add expiry time to master key cache to prevent stale encryption key usage
In multi-tenant environment, We store master keys for tenants in an in-memory HashMap ([code ref](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/
encryptor/EncryptorImpl.java#L65)). Currently, entries never expire, which can cause decryption failures in the following edge case:

Steps to Reproduce:

1. T1: Create a model → encryption key is cached in the map
2. T2: Delete the collection/domain
3. T3: Create a new collection/domain with the same name
4. T4: Create a model → uses stale encryption key from T1 (still in cache) to encrypt credentials
5. T5: Deployment occurs → HashMap is reset
6. T6: Predict on the model created at T4 → fails because:
   - Key not in cache (reset at T5)
   - Key not in index (new domain from T3)
   - New key is generated, but credentials were encrypted with the old key from T1

Root Cause:

The cache doesn't expire entries, so stale encryption keys persist across collection/domain recreation cycles.

Proposed Solution:

Add a TTL (e.g., 5 minutes) to cache entries. When an entry expires, fetch the encryption key from the index/DDB instead of using the cached value. This ensures we always use the correct encryption key for the current domain.

arjunkumargiri:
### Description

Fixes #4542 

This PR adds TTL-based expiry to the master key cache in `EncryptorImpl` to prevent stale encryption key usage in multi-tenant environments.

#### 1. Cache with TTL
- Replaced `ConcurrentHashMap<String, String>` with Guava's `Cache<String, String>`
- Added 5-minute TTL using `CacheBuilder.expireAfterWrite(5, TimeUnit.MINUTES)`
- When entries expire, keys are automatically re-fetched from the index/DDB
- This ensures we always use the correct encryption key for the current domain

#### 2. Race Condition Fix
Fixed a potential race condition where the key could expire between the cache check and usage:
- Introduced `getOrInitMasterKey()` method that atomically retrieves and returns the key
- Updated `encrypt()` and `decrypt()` to store the key in a local variable
- Prevents `NullPointerException` if cache expires mid-operation

#### 3. Testability
- Added package-private constructors accepting custom TTL for testing
- Allows tests to use 1-second TTL instead of waiting 5 minutes

### Changes

**EncryptorImpl.java:**
- Replaced `Map<String, String>` with `Cache<String, String>` with 5-minute expiry
- Added `MASTER_KEY_CACHE_TTL_MINUTES` constant (5 minutes)
- Added package-private constructors with configurable TTL for testing
- Introduced `getOrInitMasterKey()` to prevent race conditions
- Updated `createJceMasterKey()` to accept the key directly instead of tenant ID
- Updated all cache access to use `getIfPresent()` instead of `get()`

**EncryptorImplTest.java:**
- Added 5 comprehensive tests covering cache behavior:
  - `testMasterKeyRefetchAfterCacheExpiry()` - Verifies cache reuse and no redundant DDB calls
  - `testCacheExpiryConfiguration()` - Verifies DDB is called only once when key is cached
  - `testStaleMasterKeyScenarioWithCacheExpiry()` - Tests actual cache expiry with 1-second TTL
  - `testMultipleTenantsCacheIndependently()` - Ensures tenant isolation
  - `testCacheReturnsNullForExpiredKey()` - Validates null returns for missing keys
- Added `verify()` assertions to confirm DDB call counts
- Removed redundant tests

### Testing

All existing tests pass, plus 5 new tests specifically for cache expiry:

```bash
./gradlew :opensearch-ml-algorithms:test --tests "EncryptorImplTest"

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
@@ -52,6 +52,7 @@ import com.amazonaws.encryptionsdk.AwsCrypto;
 import com.amazonaws.encryptionsdk.CommitmentPolicy;
 import com.amazonaws.encryptionsdk.CryptoResult;
 import com.amazonaws.encryptionsdk.jce.JceMasterKey;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.RemovalListener;
@@ -92,21 +93,10 @@ public class EncryptorImpl implements Encryptor {
         this.client = client;
         this.sdkClient = sdkClient;
         this.mlIndicesHandler = mlIndicesHandler;
-
-        // Register listener for dynamic setting updates
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(ML_COMMONS_MASTER_KEY_CACHE_TTL_MINUTES, this::updateCacheTtl);
-    }
-
-    /**
-     * Updates the cache TTL when the cluster setting changes.
-     * Note: This only affects new cache entries. Existing entries will expire based on their original TTL.
-     */
-    private void updateCacheTtl(Integer newTtlMinutes) {
-        this.masterKeyCacheTtlMinutes = newTtlMinutes;
-        log.info("Master key cache TTL updated to {} minutes. Note: Existing cache entries retain their original TTL.", newTtlMinutes);
     }
 
     // Package-private constructor for testing with custom TTL
+    @VisibleForTesting
     EncryptorImpl(
         ClusterService clusterService,
         Client client,
--- a/common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
+++ b/common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
@@ -444,25 +444,10 @@ public final class MLCommonsSettings {
      * Default: 5 minutes
      * Range: 1 to 1440 minutes (1 day)
      * 
-     * This is a dynamic setting that can be updated without restarting the cluster:
-     * <pre>
-     * PUT _cluster/settings
-     * {
-     *   "persistent": {
-     *     "plugins.ml_commons.master_key_cache_ttl_minutes": 10
-     *   }
-     * }
-     * </pre>
+     * This is a static setting that must be configured in opensearch.yml and requires a cluster restart to take effect.
      */
     public static final Setting<Integer> ML_COMMONS_MASTER_KEY_CACHE_TTL_MINUTES = Setting
-        .intSetting(
-            ML_PLUGIN_SETTING_PREFIX + "master_key_cache_ttl_minutes",
-            5,
-            1,
-            1440,
-            Setting.Property.NodeScope,
-            Setting.Property.Dynamic
-        );
+        .intSetting(ML_PLUGIN_SETTING_PREFIX + "master_key_cache_ttl_minutes", 5, 1, 1440, Setting.Property.NodeScope);
 
     /** This setting sets the remote metadata type */
     public static final Setting<String> REMOTE_METADATA_TYPE = Setting

### CODE REVIEW

**Comment by ylwu-amzn** [comment]
Spotless check failed. Run './gradlew spotlessApply' to fix.

File: ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java [hunk 1]
```diff
@@ -62,23 +65,41 @@ public class EncryptorImpl implements Encryptor {
...
-    private final Map<String, String> tenantMasterKeys;
+    private final Cache<String, String> tenantMasterKeys;
     private MLIndicesHandler mlIndicesHandler;
 
     // concurrent map can't have null as a key. This is to support single tenancy
     // assigning some random string so that it can't be duplicate
     public static final String DEFAULT_TENANT_ID = "03000200-0400-0500-0006-000700080009";
+    
+    // TTL for master key cache entries (5 minutes)
+    private static final long MASTER_KEY_CACHE_TTL_MINUTES = 5;
```
ylwu-amzn: Should we make this configurable ?
The key will be removed from cache every 5 minutes. Then will retrieve it from index and put into cache. That may add some latency overhead.

brianf-aws: Agree with @ylwu-amzn  domain upgrade or restart of OS will be a bit unlikely so eventually it should refresh from datastore.

Also having it configurable has an operator an option to step in and refresh users key if they happen to report a problem.

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -62,23 +65,41 @@ public class EncryptorImpl implements Encryptor {
...
 
     public EncryptorImpl(String tenantId, String masterKey) {
-        this.tenantMasterKeys = new ConcurrentHashMap<>();
+        this(tenantId, masterKey, MASTER_KEY_CACHE_TTL_MINUTES, TimeUnit.MINUTES);
+    }
+
+    // Package-private constructor for testing with custom TTL
+    EncryptorImpl(String tenantId, String masterKey, long cacheTtl, TimeUnit timeUnit) {
+        this.tenantMasterKeys = CacheBuilder.newBuilder()
+            .expireAfterWrite(cacheTtl, timeUnit)
```
brianf-aws, line 132: If we make TTL very long, can we add logging so operators have visibility on evictions? While I do think your PR will solve tag mismatch issues  but we wouldn't be able to get a hint about the key in memory/cache

```
EncryptorImpl(String tenantId, String masterKey, long cacheTtl, TimeUnit timeUnit) {
    this.tenantMasterKeys = CacheBuilder.newBuilder()
        .expireAfterWrite(cacheTtl, timeUnit)
        .removalListener((RemovalNotification<String, MasterKey> notification) -> 
            logger.info("Cache entry removed - Key: {}, Cause: {}", 
                notification.getKey(), 
                notification.getCause()))
        .build();
}
```

ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
brianf-aws, line 1: thanks for making the tests very thorough!

brianf-aws (approved):
Very good changes, curious to know your thoughts on making the TTL configurable

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -121,13 +142,26 @@ public String generateMasterKey() {
...
         return JceMasterKey.getInstance(new SecretKeySpec(bytes, "AES"), "Custom", "", "AES/GCM/NOPADDING");
     }
 
+    private String getOrInitMasterKey(String tenantId) {
+        String masterKey = tenantMasterKeys.getIfPresent(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID));
+        if (masterKey != null) {
+            return masterKey;
+        }
+        initMasterKey(tenantId);
+        masterKey = tenantMasterKeys.getIfPresent(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID));
```
Zhangxunmt: Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID) is same called twice but can use a variable to remove the duplication. 
```
String effectiveTenantId = Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID); 
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -121,13 +142,26 @@ public String generateMasterKey() {
...
+        byte[] bytes = Base64.getDecoder().decode(masterKey);
         return JceMasterKey.getInstance(new SecretKeySpec(bytes, "AES"), "Custom", "", "AES/GCM/NOPADDING");
     }
 
+    private String getOrInitMasterKey(String tenantId) {
+        String masterKey = tenantMasterKeys.getIfPresent(Objects.requireNonNullElse(tenantId, DEFAULT_TENANT_ID));
+        if (masterKey != null) {
+            return masterKey;
+        }
+        initMasterKey(tenantId);
```
Zhangxunmt: nit: When the concurrency TPS is high, I think using the thread safe check here could avoid duplicate index writing in the initMasterKey(tenantId) to have better performance, since the MLConfigIndex writing/update is turned into sync operation in initMasterKey(tenantId). 

    synchronized (lock) {
        masterKey = tenantMasterKeys.getIfPresent(effectiveTenantId);
        if (masterKey == null) {
            initMasterKey(tenantId);
            masterKey = tenantMasterKeys.getIfPresent(effectiveTenantId);
        }
    }

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -62,23 +67,86 @@ public class EncryptorImpl implements Encryptor {
...
+    private final Object lock = new Object();
+    private volatile long masterKeyCacheTtlMinutes;
 
     // concurrent map can't have null as a key. This is to support single tenancy
     // assigning some random string so that it can't be duplicate
     public static final String DEFAULT_TENANT_ID = "03000200-0400-0500-0006-000700080009";
 
     public EncryptorImpl(ClusterService clusterService, Client client, SdkClient sdkClient, MLIndicesHandler mlIndicesHandler) {
-        this.tenantMasterKeys = new ConcurrentHashMap<>();
+        this.masterKeyCacheTtlMinutes = ML_COMMONS_MASTER_KEY_CACHE_TTL_MINUTES.get(clusterService.getSettings());
```
dhrubo-os, line 81: What do you think about adding this here: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/settings/MLFeatureEnabledSetting.java ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -62,23 +67,86 @@ public class EncryptorImpl implements Encryptor {
...
+     * Updates the cache TTL when the cluster setting changes.
+     * Note: This only affects new cache entries. Existing entries will expire based on their original TTL.
+     */
+    private void updateCacheTtl(Integer newTtlMinutes) {
+        this.masterKeyCacheTtlMinutes = newTtlMinutes;
+        log.info("Master key cache TTL updated to {} minutes. Note: Existing cache entries retain their original TTL.", newTtlMinutes);
+    }
+
+    // Package-private constructor for testing with custom TTL
+    EncryptorImpl(
```
dhrubo-os, line 100: [nit] can we add `@VisibleForTesting` ?

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
arjunkumargiri: Dynamic setting for ML master key cache will not be required. It needs to be a bootstrap configuration. 

Let me know if you have concerns.

mingshl: IMO, I don't want a B/G just because of increasing expiration time

ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java
```diff
@@ -62,23 +68,75 @@ public class EncryptorImpl implements Encryptor {
...
+
+    // Package-private constructor for testing with custom TTL
+    EncryptorImpl(String tenantId, String masterKey, long cacheTtl, TimeUnit timeUnit) {
+        this.tenantMasterKeys = CacheBuilder
+            .newBuilder()
+            .expireAfterWrite(cacheTtl, timeUnit)
+            .removalListener(new RemovalListener<String, String>() {
+                @Override
+                public void onRemoval(RemovalNotification<String, String> notification) {
+                    log.info("Master key cache entry removed - Tenant: {}, Cause: {}", notification.getKey(), notification.getCause());
```
mingshl, line 136: is it safe to log the tenant here?

dhrubo-os: yeah tenant id is a metadata. So this should be fine from security POV

mingshl: all good from me then, approve.

ml-algorithms/src/test/java/org/opensearch/ml/engine/encryptor/EncryptorImplTest.java
```diff
@@ -779,6 +793,188 @@ public void encrypt_GetSourceAsMapIsNull_ShouldThrowResourceNotFound() throws Ex
...
+        // Second encryption should use cached key (no additional DDB call)
+        String encrypted2 = encryptor.encrypt("test", TENANT_ID);
+        Assert.assertNotNull(encrypted2);
+
+        // Verify DDB (client.get) was called only once, not twice
+        verify(client, times(1)).get(any(), any());
+    }
+
+    @Test
+    public void testCacheExpiryConfiguration() throws Exception {
```
akolarkunnu, line 831: @arjunkumargiri What is the difference between tests testMasterKeyRefetchAfterCacheExpiry() and testCacheExpiryConfiguration(). I don't see any clear difference. Both are testing same functioanlity.

arjunkumargiri: It should be fine to remove it, I started testing the timeout config in testCacheExpiryConfiguration but ran into few challenges.

akolarkunnu: Ok, I will remove it in my PR https://github.com/opensearch-project/ml-commons/pull/3919

PR #4557, opened 2026-01-24
[Agentic Search] Validate model type as REMOTE for the Query Planner Tool

### LINKED ISSUE #4540: [FEATURE] enhance queryplanning tool to validate provided model type
in queryplanning tool, it's expected to use a llm model, 

add the validation in query planning tool.
if model type is not remote, try to throw exception to advise user to choose a llm model id. 

rithinpullela:
### Description
This PR validates the model type used in Query Planner Tool as REMOTE (ideally should validate as LLM but there is no way to determine it for now)

### Related Issues
Resolves #4540

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.engine.tools;
 
 import static org.opensearch.action.support.clustermanager.ClusterManagerNodeRequest.DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
 import static org.opensearch.ml.common.CommonValue.TOOL_INPUT_SCHEMA_FIELD;
 import static org.opensearch.ml.common.utils.StringUtils.gson;
 import static org.opensearch.ml.engine.algorithms.agent.AgentUtils.AGENT_LLM_MODEL_ID;
@@ -37,9 +38,13 @@ import org.opensearch.cluster.metadata.MappingMetadata;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.index.IndexNotFoundException;
 import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.spi.tools.Parser;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.ml.common.spi.tools.WithModelTool;
+import org.opensearch.ml.common.transport.model.MLModelGetAction;
+import org.opensearch.ml.common.transport.model.MLModelGetRequest;
 import org.opensearch.ml.common.utils.ToolUtils;
 import org.opensearch.ml.engine.processor.ProcessorChain;
 import org.opensearch.ml.engine.tools.parser.ToolParser;
@@ -94,6 +99,10 @@ public class QueryPlanningTool implements WithModelTool {
         .of(CHAT_HISTORY_FIELD, TOOLS_FIELD, INTERACTIONS_FIELD, TOOL_CONFIGS_FIELD);
     public static final String FALLBACK_QUERY_FIELD = "fallback_query";
 
+    // Model validation error messages
+    private static final String LLM_MODEL_NOT_FOUND_ERROR = "LLM model not found. Please provide a valid model ID.";
+    private static final String LLM_MODEL_NOT_REMOTE_ERROR = "Query Planning Tool requires a remote LLM model.";
+
     @Getter
     private final String generationType;
     @Getter
@@ -130,6 +139,8 @@ public class QueryPlanningTool implements WithModelTool {
     @Setter
     private String description = DEFAULT_DESCRIPTION;
     private final Client client;
+    private final String modelId;
+    private volatile boolean validated = false;
 
     @Setter
     @Getter
@@ -140,13 +151,15 @@ public class QueryPlanningTool implements WithModelTool {
         MLModelTool queryGenerationTool,
         Client client,
         String searchTemplates,
-        String fallbackQuery
+        String fallbackQuery,
+        String modelId
     ) {
         this.generationType = generationType;
         this.queryGenerationTool = queryGenerationTool;
         this.client = client;
         this.searchTemplates = searchTemplates;
         this.fallbackQuery = fallbackQuery;
+        this.modelId = modelId;
         this.attributes = new HashMap<>(DEFAULT_ATTRIBUTES);
     }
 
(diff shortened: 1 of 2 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -127,16 +136,25 @@ public class QueryPlanningTool implements WithModelTool {
...
     private String description = DEFAULT_DESCRIPTION;
     private final Client client;
+    private final String modelId;
+    private volatile boolean validated = false;
 
     @Setter
     @Getter
     private Parser outputParser;
 
-    public QueryPlanningTool(String generationType, MLModelTool queryGenerationTool, Client client, String searchTemplates) {
```
mingshl: Earlier the model id is not a required input as if the conversational agent has a model id, queryPlanningTool will reuse it. After you changed this interface, have you verified when model id is taken from the conversational agent?

Maybe instead of changing the interface you could have two interfaces. One is using default model id and the other is allow passing over model id. Up to you.

rithinpullela: > After you changed this interface, have you verified when model id is taken from the conversational agent?

Yes I did, and it works as expected. In fact my change does not impact the existing logic for pulling the Model ID from. Agent

The Model ID read from the Agent will be loaded in the MLModelTool in the Tool's Create function, so this  constructor above did not need a model ID since it was anyways loaded in the MLModelTool Object. With this change I am also passing the model ID, just to add additional validation in the run method. I did not do validation in the Create method of the Tool because the create method's signature is synchronous and the validation would add a blocking call which is not ideal, hence I added it in the run method.

mingshl: the model id was added in the tool.run parameters?  if that is the case, if taking the model id from conversational agent, will the queryPlanningTool constructor model_id is null? if you can confirm this is not happening, I am all good, 
 

> The Model ID read from the Agent will be loaded in the MLModelTool in the Tool's Create function

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
```diff
@@ -156,6 +174,25 @@ private Map<String, String> stripAgentContextParameters(Map<String, String> orig
     public <T> void run(Map<String, String> originalParameters, ActionListener<T> listener) {
         try {
             Map<String, String> parameters = stripAgentContextParameters(ToolUtils.extractInputParameters(originalParameters, attributes));
+            String tenantId = originalParameters.get(TENANT_ID_FIELD);
+
+            // Validate that the model is of REMOTE type (only on first run)
+            if (!validated) {
+                validateModelIsRemote(modelId, tenantId, ActionListener.wrap(valid -> {
```
owaiskazi19, line 186: Still this can be a concern for validation since a remote model could be anything. It could be a remote text embedding model or a LLM. Better approach would be to have an interface for the type of model registered with the agent. That would provide us an option to validate the type for the tool which needs to validate them like QPT.

rithinpullela: Agreed, I raised the same concerns in the issue https://github.com/opensearch-project/ml-commons/issues/4540

But currently we do not have a mechanism to verify if a remote model is LLM or Text Embedding or other kind. This PR just prevents from providing local models.

The same is an issue for agents as well, agents can also be registered with any model. Raised an issue here: https://github.com/opensearch-project/ml-commons/issues/4558

mingshl: I also made the same comment on the issue https://github.com/opensearch-project/ml-commons/issues/4558, to keep track of this progress :

 @pyek-bot has a plan to introduce simplified model registration with model interface, then we can have more granular model meta data to identify with. 

But the higher level of validation is to make sure QueryPlanningTool is using a remote model, then the next granular level is to make sure it's a LLM model. We can roll out as phases.

rithinpullela: Agreed we can improve this in next phases

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
rithinpullela: The model ID will not be null if QPT is taking it from the Conversational Agent. Please refer to lines 495-502.

If the Model ID is not provided we load model Id from Agent Model ID, so it will not be null.

mingshl: great, so the model id is only taking during tools creation and it won't got overwritten then this is fine.

mingshl (commented):
I think this validation is necessary. 

I am cross checking one last time, there is a Question and Answering type of local model. I am thinking if that is possible to use this Q&A model in queryplanningtool to generate a query string? @rithin-pullela-aws can you take a look at this sample model? https://huggingface.co/mrm8488/electra-small-finetuned-squadv2

mingshl:
@owaiskazi19 I know there is a text to PPL model, what is the plan for registering this text to ppl model, will it be remote model or will it be local model? 

if we block the local model for query planning tool, we should clear out the possibility using local model for QPT

ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/QueryPlanningToolTests.java
```diff
@@ -1535,4 +1588,81 @@ public void testFactoryCreate_NoModelIdProvided() {
...
+        actionListenerCaptor.getValue().onResponse(getIndexResponse);
+
+        // Should succeed - no exception thrown
+        String result = future.get();
+        assertNotNull(result);
+    }
+
+    @SneakyThrows
+    @Test
+    public void testRun_NonRemoteModel_ValidationFails() throws ExecutionException, InterruptedException {
```
owaiskazi19, line 2063: Can we also add a test for when model id is not provided validation added above?

ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/QueryPlanningToolTests.java
```diff
@@ -1535,4 +1588,81 @@ public void testFactoryCreate_NoModelIdProvided() {
...
+        Map<String, String> params = new HashMap<>();
+        params.put(QUESTION_FIELD, "test question");
+        params.put(INDEX_NAME_FIELD, "testIndex");
+
+        tool.run(params, listener);
+
+        // Should fail with validation error
+        ExecutionException exception = assertThrows(ExecutionException.class, () -> future.get());
+        assertTrue(exception.getCause() instanceof IllegalArgumentException);
+        assertTrue(exception.getCause().getMessage().contains("remote LLM model"));
```
owaiskazi19: we can make substring be a little clear with the test
```suggestion
        assertTrue(exception.getCause().getMessage().contains("requires a remote LLM model"));
```

PR #4596, opened 2026-02-04
Support tool messages in agent revamp

jiapingzeng:
### Description
Changes included:
- support tool messages in agent revamp for both Bedrock and OpenAI model provider
- move tool message processing logic from AG-UI agent to agent revamp
- support image in streaming

Request:
```
POST {{endpoint}}/_plugins/_ml/agents/{{agent_id}}/_execute
{
    "input": [
        {
            "role": "user",
            "content": [{"type": "text", "text": "I like red"}]
        },
        {
            "role": "assistant",
            "content": [{"type": "text", "text": "Thanks for telling me that! I'll remember it."}]
        },
        {
            "role": "user",
            "content": [{"type": "text", "text": "What's the weather like in Seattle?"}]
        },
        {
            "role": "assistant",
            "content": [{"type": "text", "text": "I'll help you get the weather information for Seattle."}],
            "toolCalls": [{
                "id": "weather_check_123",
                "type": "function",
                "function": {
                    "name": "get_weather",
                    "arguments": "{\"location\": \"Seattle, WA\"}"
                }
            }]
        },
        {
            "role": "tool",
            "toolCallId": "weather_check_123",
            "content": [{"type": "text", "text": "{\"temperature\": \"65°F\", \"condition\": \"Partly cloudy\", \"humidity\": \"72%\", \"wind\": \"8 mph NW\"}"}]
        },
        {
            "role": "assistant",
            "content": [{"type": "text", "text": "The weather in Seattle is currently 65°F and partly cloudy, with 72% humidity and winds at 8 mph from the northwest. It's a pleasant day!"}]
        },
        {
            "role": "user",
            "content": [{"type": "text", "text": "Given the current weather, is it good for outdoor activities?"}]
        },
        {
            "role": "assistant",
            "content": [{"type": "text", "text": "Yes, the current weather in Seattle is quite good for outdoor activities! Here's why:\n\n- **Temperature**: 65°F is comfortable - not too hot or cold. Perfect for most outdoor activities without needing heavy layers.\n- **Conditions**: Partly cloudy means you'll get some sun but also some cloud cover for shade, which is ideal.\n- **Wind**: 8 mph is a gentle breeze - enough to keep things comfortable but not so strong that it would interfere with most activities.\n- **Humidity**: 72% is moderate. It's not uncomfortably humid, though you might feel a bit of moisture in the air.\n\nThis weather would be great for:\n- Walking or hiking\n- Biking\n- Picnics in the park\n- Outdoor sports (tennis, frisbee, etc.)\n- Sightseeing around the city\n\nJust bring a light jacket in case it cools down or the clouds thicken. Overall, it's a lovely day to be outside!"}]
        },
        {
            "role": "user",
            "content": [{"type": "text", "text": "What tools have you used and what was the result?"}]
        }
    ]
}
```

Response:
```
{
  "inference_results": [
    {
      "output": [
        {
          "name": "memory_id",
          "result": "OR2VLJwBKTKVugvBFxdh"
        },
        {
          "name": "parent_interaction_id",
          "result": "PR2VLJwBKTKVugvBFxe9"
        },
        {
          "name": "response",
          "dataAsMap": {
            "response": "Great question! Let me review what tools I've used in our conversation:\n\n## Tools Used:\n\n1. **get_weather tool**\n   - **When**: When you asked \"What's the weather like in Seattle?\"\n   - **Parameters**: I provided the location as \"Seattle, WA\"\n   - **Result**: The tool returned:\n     - Temperature: 65°F\n     - Condition: Partly cloudy\n     - Humidity: 72%\n     - Wind: 8 mph NW\n\nThat's the only tool I've actually called during our conversation. \n\nFor your other messages:\n- \"I like red\" - No tool needed, just a conversational response\n- \"Given the current weather, is it good for outdoor activities?\" - No tool needed; I used the weather data from the previous tool call to provide an analysis and recommendations\n\nSo in summary, I made one tool call to get the Seattle weather information, and then used that data to answer your follow-up question about outdoor activities.",
            "additional_info": {}
          }
        }
      ]
    }
  ]
}
```

Verified that this works for both Bedrock and OpenAI. Additionally verified that [these test requests](https://github.com/jiapingzeng/agui-testsuite/tree/main/src/tests) work for AG-UI agent as it now uses the same Messages processing as the unified interface.

Note that conversational (not AG-UI) agents registered via new agent interface does not work with execute stream. Can fix in the future. Opened issue: https://github.com/opensearch-project/ml-commons/issues/4604

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/agent/OpenaiV1ChatCompletionsModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/OpenaiV1ChatCompletionsModelProvider.java
@@ -24,6 +24,7 @@ import org.opensearch.ml.common.input.execute.agent.ContentBlock;
 import org.opensearch.ml.common.input.execute.agent.ImageContent;
 import org.opensearch.ml.common.input.execute.agent.Message;
 import org.opensearch.ml.common.input.execute.agent.SourceType;
+import org.opensearch.ml.common.input.execute.agent.ToolCall;
 import org.opensearch.ml.common.model.ModelProvider;
 import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
 import org.opensearch.ml.common.utils.ToolUtils;
@@ -80,6 +81,12 @@ public class OpenaiV1ChatCompletionsModelProvider extends ModelProvider {
         "{\"type\":\"image_url\",\"image_url\":{\"url\":\"${parameters.image_data}\"}}";
 
     private static final String MESSAGE_TEMPLATE = "{\"role\":\"${parameters.msg_role}\",\"content\":[${parameters.msg_content_array}]}";
+
+    private static final String MESSAGE_WITH_TOOL_CALLS_TEMPLATE =
+        "{\"role\":\"${parameters.msg_role}\",\"content\":[${parameters.msg_content_array}],\"tool_calls\":[${parameters.tool_calls_array}]}";
+
+    private static final String MESSAGE_WITH_TOOL_CALL_ID_TEMPLATE =
+        "{\"role\":\"${parameters.msg_role}\",\"content\":[${parameters.msg_content_array}],\"tool_call_id\":\"${parameters.tool_call_id}\"}";
 
     private static final String OPENAI_REASONING_EFFORT = "reasoning_effort";
 
@@ -247,7 +254,7 @@ public class OpenaiV1ChatCompletionsModelProvider extends ModelProvider {
 
     /**
      * Builds messages array using templates for OpenAI Chat Completions API.
-     * Converts messages to conversation history format.
+     * Converts messages to conversation history format, handling tool calls and results.
      */
     private String buildMessagesArray(List<Message> messages, MLAgentType type) {
         if (messages == null || messages.isEmpty()) {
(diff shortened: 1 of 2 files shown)


mingshl:
can you give example of image?

ylwu-amzn:
```
"arguments": "{\"location\": \"Seattle, WA\"}"
```
Can we use json object ?

jiapingzeng:
> can you give example of image?

AG-UI request with image: https://github.com/jiapingzeng/agui-testsuite/blob/main/src/tests/image-input.test.ts

LLM response:

> I can see a square icon with a light blue/gray border. Inside, there's a red background with a yellow circular badge or button in the center. The yellow circle contains the text "TEST" written in capital letters.
This appears to be a test badge or indicator icon, possibly used in software development or testing environments to mark test data, test modes, or test environments.

```
data: {"type":"RUN_STARTED","timestamp":1770319324404,"threadId":"thread-image-1","runId":"run-image-1"}

data: {"type":"TEXT_MESSAGE_START","timestamp":1770319328298,"messageId":"msg_392859133943708","role":"assistant"}

data: {"type":"TEXT_MESSAGE_CONTENT","timestamp":1770319328298,"messageId":"msg_392859133943708","delta":"I can"}

data: {"type":"TEXT_MESSAGE_CONTENT","timestamp":1770319328386,"messageId":"msg_392859133943708","delta":" see a square"}

data: {"type":"TEXT_MESSAGE_CONTENT","timestamp":1770319328426,"messageId":"msg_392859133943708","delta":" icon"}

data: {"type":"TEXT_MESSAGE_CONTENT","timestamp":1770319328463,"messageId":"msg_392859133943708","delta":" with a light blue"}

...

data: {"type":"TEXT_MESSAGE_END","timestamp":1770319330475,"messageId":"msg_392859133943708"}

data: {"type":"RUN_FINISHED","timestamp":1770319330475,"threadId":"thread-image-1","runId":"run-image-1"}


```

jiapingzeng:
> "arguments": "{\"location\": \"Seattle, WA\"}"
Can we use json object ?

`"arguments"` here is tool arguments generated by LLM which is always a string. We could accept JSON object but I don't think that would make sense, since tool argument does not have to be JSON.

OpenAI example where `arguments` is string: https://platform.openai.com/docs/guides/function-calling#handling-function-calls

common/src/main/java/org/opensearch/ml/common/agent/BedrockConverseModelProvider.java
```diff
@@ -266,23 +267,114 @@ private String buildMessagesArray(List<Message> messages, MLAgentType type) {
...
 
-            String contentArray = buildContentArrayFromBlocks(message.getContent(), type);
+            // Convert "tool" role to "user" for Bedrock compatibility
+            String role = "tool".equalsIgnoreCase(message.getRole()) ? "user" : message.getRole();
+
             Map<String, Object> msgParams = new HashMap<>();
-            msgParams.put("msg_role", message.getRole());
-            msgParams.put("msg_content_array", contentArray);
+            msgParams.put("msg_role", role);
+            msgParams.put("msg_content_array", contentArray.toString());
```
mingshl, line 324: and this change also make the content array for bedrock models to return as sting?>?

mingshl: it seems the data type casting is within agent execution, it's not impacting the agent input nor response. Then it should be fine

mingshl (approved):

common/src/main/java/org/opensearch/ml/common/agui/AGUIInputConverter.java
```diff
@@ -286,56 +333,73 @@ private static String extractFormat(String mimeType) {
...
 
-        return toolResults;
+            if (lastTextBlock != null) {
+                String originalText = lastTextBlock.getText();
+                String newText = contextBuilder.toString() + originalText;
+                lastTextBlock.setText(newText);
+                log.debug("AG-UI: Appended context to latest user message");
+            } else {
+                // should not happen as user message has to have content
+                log.warn("No text content block found in latest user message, skipping context appending");
```
mingshl, line 398: overall it makes sense, but if this is when the model predict result is malformed, should we throw exception>?

jiapingzeng: This else block happens when the request from frontend is malformed; I think it is ok to only log warning here as this part handles appending context to question only. 

If the input is truly malformed and the user message does not have content, it will fail validation in AgentInputProcessor: https://github.com/jiapingzeng/ml-commons/blob/10dba85eab5afb88c47a4a4580e7ba564b40fb68/common/src/main/java/org/opensearch/ml/common/input/execute/agent/AgentInputProcessor.java#L89-L94

PR #4599, opened 2026-02-04
Fix gemini function calling in tool failure case

rithinpullela:
### Description
Gemini expects a JSON in tool response. While the `supply` method in `GeminiV1BetaGenerateContentFunctionCalling.java` handles it, in error cases the supply method is bypassed by the `runTool` function. This change to the template ensures that we always supply the tool response in JSON format. 

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/GeminiV1BetaGenerateContentFunctionCalling.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/GeminiV1BetaGenerateContentFunctionCalling.java
@@ -73,7 +73,7 @@ public class GeminiV1BetaGenerateContentFunctionCalling implements FunctionCalli
         params
             .put(
                 INTERACTION_TEMPLATE_TOOL_RESPONSE,
-                "{\"role\":\"user\",\"parts\":[{\"functionResponse\":{\"name\":\"${_interactions.tool_call_id}\",\"response\":${_interactions.tool_response}}}]}"
+                "{\"role\":\"user\",\"parts\":[{\"functionResponse\":{\"name\":\"${_interactions.tool_call_id}\",\"response\":{\"text\":\"${_interactions.tool_response}\"}}}]}"
             );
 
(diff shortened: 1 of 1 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/GeminiV1BetaGenerateContentFunctionCalling.java
```diff
@@ -73,7 +73,7 @@ public void configure(Map<String, String> params) {
         params
             .put(
                 INTERACTION_TEMPLATE_TOOL_RESPONSE,
-                "{\"role\":\"user\",\"parts\":[{\"functionResponse\":{\"name\":\"${_interactions.tool_call_id}\",\"response\":${_interactions.tool_response}}}]}"
+                "{\"role\":\"user\",\"parts\":[{\"functionResponse\":{\"name\":\"${_interactions.tool_call_id}\",\"response\":{\"text\":\"${_interactions.tool_response}\"}}}]}"
```
mingshl, line 76: I used ai and it says it's a valid json string. 

was the IT able to replicate running with gemini interface?

ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/GeminiV1BetaGenerateContentFunctionCalling.java
```diff
@@ -73,7 +73,7 @@ public void configure(Map<String, String> params) {
...
             );
```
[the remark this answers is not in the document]
rithinpullela: We don't need to do this, because the tool interaction template that we edited is only used in the exception case of `runTool` method and in a non-exceptional case the template is never used and we rely on the supply method directly

ml-algorithms/src/main/java/org/opensearch/ml/engine/function_calling/GeminiV1BetaGenerateContentFunctionCalling.java
rithinpullela: We do not have ITs for Gemini models yet, I tested on my local though

mingshl (approved):

PR #4602, opened 2026-02-05
Add validation to context managers

mingshl:
### Description
There are several fixes to context managements:

1. add more validations when creating context management names, not allowing empty space, capital letters, and also not exceeding 50 char
2. use global variables for "context_management_name" 
3. change MLResourceNotFoundException to OpenSearchStatusException with NOT FOUND Status, as MLResourceNotFoundException will throw 500 error. should be 4xx error when context management name not found. 
4. there are multiple PRs changing the conversational agent runner, the nextStepListener in MLChantAgentRunner changes was reverted in consequence PRs modifying this class. Moved back to position to save tool response to memory
5. Enhanced exception message: remove the wording "template" in the exception messages, as the template service is an internal class and it might not make sense to users. 
6. address the client context stash issue for https://github.com/opensearch-project/ml-commons/issues/4603

### Related Issues

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
+++ b/common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -39,6 +40,8 @@ public class ContextManagementTemplate implements ToXContentObject, Writeable {
     public static final String CREATED_TIME_FIELD = "created_time";
     public static final String LAST_MODIFIED_FIELD = "last_modified";
     public static final String CREATED_BY_FIELD = "created_by";
+
+    private static final Pattern VALID_NAME_PATTERN = Pattern.compile("^[a-z0-9_-]{1,49}$");
 
     /**
      * Unique name for the context management template
@@ -242,36 +245,7 @@ public class ContextManagementTemplate implements ToXContentObject, Writeable {
      * Validate only the template name
      */
     public boolean isValidName() {
-        if (name == null || name.trim().isEmpty()) {
-            return false;
-        }
-
-        // Name must not contain spaces
-        if (name.contains(" ")) {
-            return false;
-        }
-
-        // Name must not contain capital letters
-        if (!name.equals(name.toLowerCase())) {
-            return false;
-        }
-
-        // Name must not contain control characters
-        if (name.indexOf('\r') >= 0 || name.indexOf('\n') >= 0 || name.chars().anyMatch(ch -> ch < 32)) {
-            return false;
-        }
-
-        // Name must only contain lowercase letters, numbers, hyphens, and underscores
-        if (!name.matches("^[a-z0-9_-]+$")) {
-            return false;
-        }
-
-        // Name length must be less than 50 characters
-        if (name.length() >= 50) {
-            return false;
-        }
-
-        return true;
+        return name != null && VALID_NAME_PATTERN.matcher(name).matches();
     }
 
     /**
--- a/plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
@@ -51,7 +51,7 @@ public class ContextManagementTemplateService {
 
     private static final int DEFAULT_MAX_TEMPLATES = 1000;
     private static final String INVALID_CONTEXT_MANAGEMENT_NAME_ERROR =
-        "Invalid context management name: must contain only lowercase letters, numbers, hyphens, and underscores, and must be less than 50 characters";
+        "Invalid context management name: must contain only lowercase letters, numbers, hyphens, and underscores, and must be between 1 and 49 characters";
     private static final String INVALID_CONTEXT_MANAGEMENT_CONFIG_ERROR = "Invalid context management configuration";
     private static final String CONTEXT_MANAGEMENT_NAME_NULL_ERROR = "Context management name cannot be null, empty, or whitespace";
     private static final String CONTEXT_MANAGEMENT_NOT_FOUND_ERROR = "Context management not found: ";
--- a/plugin/src/test/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateServiceTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateServiceTests.java
@@ -729,7 +729,7 @@ public class ContextManagementTemplateServiceTests extends OpenSearchTestCase {
 
         ArgumentCaptor<Exception> exceptionCaptor = ArgumentCaptor.forClass(Exception.class);
         verify(listener).onFailure(exceptionCaptor.capture());
-        assertTrue(exceptionCaptor.getValue().getMessage().contains("less than 50 characters"));
+        assertTrue(exceptionCaptor.getValue().getMessage().contains("between 1 and 49 characters"));
     }
 
     @Test


common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
```diff
@@ -239,6 +239,21 @@ public boolean isValid() {
             return false;
         }
 
+        // Name must not contain spaces
+        if (name.contains(" ")) {
+            return false;
+        }
+
+        // Name must not contain capital letters
+        if (!name.equals(name.toLowerCase())) {
```
b4sjoo: Consider change to toLowerCase(Locale.ROOT)

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
```diff
@@ -163,21 +163,20 @@ public void getTemplate(String templateName, ActionListener<ContextManagementTem
...
                     } catch (Exception e) {
-                        log.error("Failed to parse context management template: {}", templateName, e);
+                        log.error("Failed to parse context management: {}", templateName, e);
                         wrappedListener.onFailure(e);
                     }
                 }, exception -> {
                     if (exception instanceof IndexNotFoundException) {
-                        wrappedListener
-                            .onFailure(new MLResourceNotFoundException("Context management template not found: " + templateName));
+                        wrappedListener.onFailure(new IllegalArgumentException("Context management not found: " + templateName));
```
b4sjoo: Why `IllegalArgumentException` instead of `MLResourceNotFoundException` here?

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
```diff
@@ -239,6 +239,21 @@ public boolean isValid() {
             return false;
         }
 
+        // Name must not contain spaces
```
b4sjoo: Consider add some name validation logic like https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java#L244 to avoid control characters (\n...) or other special non-alphanumeric characters (#@$...)

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
```diff
@@ -246,7 +245,7 @@ public void listTemplates(int from, int size, ActionListener<List<ContextManagem
     public void deleteTemplate(String templateName, ActionListener<Boolean> listener) {
         try {
             if (templateName == null || templateName.trim().isEmpty()) {
-                listener.onFailure(new IllegalArgumentException("Template name cannot be null, empty, or whitespace"));
+                listener.onFailure(new IllegalArgumentException("context management name cannot be null, empty, or whitespace"));
```
b4sjoo: Nit: Sometimes you use upper case and sometimes you use lowercase on message, consider change all to error message to capital? i.e. Context management name cannot be...

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
mingshl: MLResourceNotFoundException is throwing 500 error

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
mingshl: I thought about special characters in the names but the name are stored as string fields in the system index. Should be fine to contain special characters

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
b4sjoo: What about `OpenSearchStatusException` with NOT_FOUND RestStatus?

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
b4sjoo: But still under this case control characters will cause issue

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -770,6 +770,7 @@ private static void runTool(
                         // TODO: support multiple tool calls at the same time so that multiple
                         // LLMMessages can be generated here
                         interactions.add(llmMessages.getFirst().getResponse());
+                        nextStepListener.onResponse(outputResponseAfterHook);
```
b4sjoo, line 773: The function-calling branch in `runTool` passes already-filtered output `outputResponseAfterHook` to `nextStepListener.onResponse()`, but the outer loop at line 581 applies `filterToolOutput()` again. This can double-encode JSON strings and break JSONPath output_filter or produce extra-quoted output like `"value"` becomes `"\"value\""`

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
mingshl: Locale.ROOT is unnecessary here since we are not performing case conversion.
Uppercase detection is done via Character.isUpperCase, which is Unicode-aware and locale-independent

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
mingshl: This is better. I will change to OpenSearchStatusException.

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
```diff
@@ -239,6 +239,21 @@ public boolean isValid() {
...
+        }
+
+        // Name length must be less than 50 characters
+        if (name.length() >= 50) {
+            return false;
+        }
+
         // Allow null hooks (no context management) but not empty hooks map (misconfiguration)
         if (hooks != null) {
             if (hooks.isEmpty()) {
```
b4sjoo, line 258: `ContextManagementTemplateService` uses a name-specific error message when template.isValid() fails, but isValid() can fail for multiple reasons:                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                              
 isValid() checks (ContextManagementTemplate.java:237-276):                                                                                                                                                                                                                                                                                                   
 1. Name is null/empty (line 238)                                                                                                                                                                                                                                                                                                                             
 2. Name contains spaces (line 243)                                                                                                                                                                                                                                                                                                                           
 3. Name contains capital letters (line 248)                                                                                                                                                                                                                                                                                                                  
 4. Name length >= 50 (line 253)                                                                                                                                                                                                                                                                                                                              
 5. Hooks map is empty (line 259) ← Not a name issue                                                                                                                                                                                                                                                                                                     
 6. ContextManagerConfig type is null/empty (line 267) ← Not a name issue                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                              
But the error message (line 51-52) only mentions names: "Invalid context management name: must not contain spaces or capital letters, and must be less than 50 characters." This misleads users when validation fails due to hooks or config issues. Consider make the error message either more specific or more generic.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
mingshl: the runTool is a void method, which doesn't pass over the filtered output.

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
mingshl: separated the valid name check and the valid configuration check to be more specific. Thanks for the recommendations. It's more readable messages in this way

plugin/src/main/java/org/opensearch/ml/action/contextmanagement/ContextManagementTemplateService.java
mingshl: changed to all capital first letter.

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
mingshl: added the check to not allow special characters

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagementTemplate.java
```diff
@@ -235,10 +235,50 @@ private static Map<String, List<ContextManagerConfig>> parseHooks(XContentParser
      * Validate the template configuration
      */
     public boolean isValid() {
+        return isValidName() && isValidHooksAndConfigs();
+    }
+
+    /**
+     * Validate only the template name
+     */
+    public boolean isValidName() {
```
ylwu-amzn, line 247: This method can be simplified. The regex already handles spaces, capital letters, control characters, and empty strings.

```
private static final Pattern VALID_NAME_PATTERN = Pattern.compile("^[a-z0-9_-]{1,49}$");

public boolean isValidName() {
    return name != null && VALID_NAME_PATTERN.matcher(name).matches();
}
```

mingshl: right, saving more lines [consolidate test](https://github.com/opensearch-project/ml-commons/pull/4602/commits/22cae5b46cc3f2af4427ca7236c306e7455d6f9d)

PR #4564, opened 2026-01-26
Enhancement: Enable remote conversational agentic memory through REST HTTP call

b4sjoo:
## Description
  ### Summary                                                                                                                                                                                                                                                                   
  This PR introduces support for remote agentic conversation memory, enabling agents to store and retrieve conversational memory across OpenSearch clusters via connector-defined REST APIs. This is a significant architectural enhancement that decouples memory storage from the local cluster.                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                
  ### Key Changes                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
  #### 1. New Memory Interface Architecture                                                                                                                                                                                                                                     
  - **New `Memory` interface** (`common/src/main/java/org/opensearch/ml/common/memory/Memory.java`): A generic memory interface with type parameters for message type, save response, and update response                                                                       
  - **New `MLMemoryType` enum**: Defines three memory types:                                                                                                                                                                                                                    
    - `CONVERSATION_INDEX` - Legacy conversation index memory                                                                                                                                                                                                                   
    - `AGENTIC_MEMORY` - Local memory container-backed storage                                                                                                                                                                                                                  
    - `REMOTE_AGENTIC_MEMORY` - Remote cluster memory via connectors                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                
  #### 2. AgenticConversationMemory (Local)                                                                                                                                                                                                                                     
  - New implementation storing conversations in Memory Container using transport actions                                                                                                                                                                                        
  - Uses `TransportCreateSessionAction` and `TransportAddMemoriesAction` for operations                                                                                                                                                                                         
  - Stores structured data in `structured_data_blob` field                                                                                                                                                                                                                      
  - Supports save, update, getMessages, and getTraces operations                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
  #### 3. RemoteAgenticConversationMemory (Remote)                                                                                                                                                                                                                              
  - **~1,400 lines of new code** enabling cross-cluster memory storage                                                                                                                                                                                                          
  - Backed by connector-defined REST APIs (supports HTTP and AWS connectors)                                                                                                                                                                                                    
  - Supports AWS Signature V4 authentication for AOSS/ES endpoints                                                                                                                                                                                                              
  - Implements full memory operations: save, update, getMessages, getTraces                                                                                                                                                                                                     
  - Factory pattern for memory creation with endpoint validation                                                                                                                                                                                                                
[...]

PR #4608, opened 2026-02-05
Fix: Restore Thread Context in MLAgentExecutor properly to fix Memory Access

rithinpullela:
### Description

After #4564, Get Memory for Admin failed due to missing thread context (user/tenant info). The commit inadvertently change the code this way:

Before the commit we restored the context after performing `sdkClient.getDataObjectAsync()` and hence all the functions inside `sdkClient.getDataObjectAsync().whenComplete()` had access to thread context
```java
    sdkClient
        .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor("opensearch_ml_general"))
        .whenComplete((response, throwable) -> {
            context.restore();  // ← Restore context BEFORE doing anything else
            log.debug("Completed Get Agent Request, Agent id:{}", agentId);
            // ... rest of the code uses listener directly
            listener.onFailure(...);
        });
}
```

After the commit we used a Wrapped Listener which does restore after the whole try block is done. This way the code inside `sdkClient.getDataObjectAsync().whenComplete()` does not have the thread context

``` java
    ActionListener<Output> wrappedListener = ActionListener.runBefore(listener, context::restore);
    sdkClient
        .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor("opensearch_ml_general"))
        .whenComplete((response, throwable) -> {
            // ❌ NO context.restore() here anymore!
            log.debug("Completed Get Agent Request, Agent id:{}", agentId);
            // ... code uses wrappedListener
            wrappedListener.onFailure(...);
        });
}
// Context restored after this
```

This broke the Memory Factory related code inside the `whenComplete()` which uses the thread context data like user information to index into the relevant indices.

### Fix:

The below fix restores the context in the `whenComplete` similar to the implementation before the breaking change

I have verified the change in my local security enabled cluster and found that this change fixes the memory access.

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
@@ -263,7 +263,7 @@ public class MLAgentExecutor implements Executable, SettingsChangeListener {
                                                 ? requestParameters.get(MEMORY_CONFIGURATION_FIELD)
                                                 : null;
                                             if (usesRemoteMemory || !Strings.isNullOrEmpty(memoryConfig)) {
-                                                wrappedListener
+                                                listener
                                                     .onFailure(
                                                         new OpenSearchStatusException(
                                                             ML_COMMONS_REMOTE_AGENTIC_MEMORY_DISABLED_MESSAGE,


rithinpullela:
@coderabbitai can you review this again please?

ylwu-amzn:
CI failed

```
/__w/ml-commons/ml-commons/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java:264: error: cannot find symbol
                                                  wrappedListener
                                                  ^
    symbol:   variable wrappedListener
    location: class MLAgentExecutor
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
ylwu-amzn: Compiling error from this line. This `wrappedListener` not defined.

rithinpullela: Rebase has caused this issue, will push again

mingshl:
@rithin-pullela-aws the CI failed

```
> Task :opensearch-ml-plugin:test


REPRODUCE WITH: gradlew ':opensearch-ml-plugin:test' --tests 'org.opensearch.ml.utils.IndexUtilsTests.testGetNumberOfDocumentsInIndex_SearchQuery' -Dtests.seed=BDF34C7F5B832061 -Dtests.security.manager=false -Dtests.locale=de-BE -Dtests.timezone=Africa/Cairo -Druntime.java=21
IndexUtilsTests > testGetNumberOfDocumentsInIndex_SearchQuery FAILED
    java.lang.AssertionError: expected:<20> but was:<9>
        at __randomizedtesting.SeedInfo.seed([BDF34C7F5B832061]:0)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opensearch.ml.utils.IndexUtilsTests.lambda$testGetNumberOfDocumentsInIndex_SearchQuery$4(IndexUtilsTests.java:91)
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
```diff
@@ -200,19 +200,19 @@ public void execute(Input input, ActionListener<Output> listener, TransportChann
 
         if (MLIndicesHandler.doesMultiTenantIndexExist(clusterService, mlFeatureEnabledSetting.isMultiTenancyEnabled(), ML_AGENT_INDEX)) {
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-                ActionListener<Output> wrappedListener = ActionListener.runBefore(listener, context::restore);
                 sdkClient
                     .getDataObjectAsync(getDataObjectRequest, client.threadPool().executor("opensearch_ml_general"))
                     .whenComplete((response, throwable) -> {
+                        context.restore();
```
b4sjoo, line 206: https://github.com/opensearch-project/ml-commons/pull/4564#discussion_r2727495834 I did the code change because of AI got some concern on the block in async condition, can you check?

rithinpullela: I believe the bot's comment is wrong. Because Async call backs execute on a different thread. Try with resources does restore the context on the main thread but we use Context.restore to make that context available on the async thread. Hence restoring context should be okay and in fact needed for proper context propagation.

Hailong-am: do we know what's the thread name used for processing response? 

I would recommend do the same thing as https://github.com/opensearch-project/ml-commons/pull/4531/changes for async case

b4sjoo:
Can you check similar changes on 
1. https://github.com/opensearch-project/ml-commons/pull/4564/files#diff-3f217e7c358a6edfa84e7c413451578443589313530398529f5f70f02d442a40R88
2. https://github.com/opensearch-project/ml-commons/pull/4564/files#diff-96eb7849848f54c625cbb0c0f81e63f28076f4ff0f4aebe9b6f64c5223557745R900
3. https://github.com/opensearch-project/ml-commons/pull/4564/files#diff-a1dfa757350b7f557935822a8ab645550bfa2dcf8585e79fd1b6c9e0ebba3491R98
To see if we also need to fix those?

b4sjoo (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAgentExecutor.java
rithinpullela: Thanks for the comment @Hailong-am. I agree that the pattern you suggested is safer since we are using try with resources.

I believe OpenSearch handles the async call and assigns a new thread to handle the response. That is where the context.restore() and all other subsequent calls are handled inside the whenComplete block.

So Thread 1: Stashes context and executes the get call

Thread 2 assigned by Opensearch to handle the response:
- First stashes whatever is stored on it before handling our request. 
- Then restores whatever context is passed by us -> This is our thread 1 context which is nothing as we stashed it.
- Now our code is executed -> Which restores the previously stashed context
- After exec, this thread gets back its original context irrespective of if we cleaned it or not

This is why the `context.restore()` has worked for us till now. This pattern is observed in multiple places before in our code base

But I agree in terms of defensive coding we need to use the pattern you suggested.

Hailong-am: make sense, my only concern is we restore context into response procesing thread and didn't clean it up. so I'm curious about the thread name

PR #4656, opened 2026-02-21
Fix numeric type preservation in ML inference query template substitution

dhrubo-os:
### Description

This PR fixes multiple CI test failures caused by recent infrastructure changes.

---

#### Fix 1: ML inference range query rewrite integration test

**Root Cause**: The Bedrock connector was configured with `pre_process_function` and `post_process_function`. The post-process function transforms the raw Bedrock response into a `ModelTensor` with `dataAsMap=null`, making `embedding.length()` unevaluable via JsonPath. The entire embedding array was being substituted into the range query instead of its length.

**Fix** (aligned with [PR #4657](https://github.com/opensearch-project/ml-commons/pull/4657)):
- Remove `pre_process_function` and `post_process_function` from the Bedrock connector in the test so the raw response is available as `dataAsMap`
- Use `embedding.length()` directly to get the embedding dimension as an integer
- Use `diary_embedding_size_int` (integer field) in the range query

No changes to `MLInferenceSearchRequestProcessor` — `StringSubstitutor` works correctly as-is.

---

#### Fix 2: Cohere connector model update

**Root Cause**: `command-a-03-2025` only supports the Cohere v2 Chat API but the connector uses the v1 API format, causing `NO_VALID_RESPONSE_GENERATED` errors. `command-r` was removed September 2025.

**Fix**: Updated `COHERE_CONNECTOR_BLUEPRINT` model to `command-r-08-2024` which supports the v1 Chat API.

---

#### Fix 3: bc-fips on unit test classpath in FIPS mode

**Root Cause**: In FIPS mode, `bc-fips` is excluded from all Gradle configurations to prevent jar hell when the plugin is installed alongside OpenSearch core. Unit tests don't run inside OpenSearch, so `bc-fips` is missing at test runtime, causing `NoClassDefFoundError` in `MLSyncUpCronTests`.

**Fix**: Added `bc-fips` to the unit test task classpath via a detached configuration (bypasses `configurations.all` exclusion without affecting the plugin bundle).

---

### Check List

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @b4sjoo @dhrubo-os @mingshl @jngz-es @model-collapse @rbhavna @ylwu-amzn @zane-neo @Zhangxunmt @austintlee @HenryL27 @sam-herman @xinyual @pyek-bot
+*   @b4sjoo @dhrubo-os @mingshl @jngz-es @model-collapse @rbhavna @ylwu-amzn @zane-neo @Zhangxunmt @austintlee @HenryL27 @sam-herman @xinyual @pyek-bot @rithin-pullela-aws
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -14,6 +14,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Sicheng Song         | [b4sjoo](https://github.com/b4sjoo)                 | Amazon    |
 | Mingshi Liu          | [mingshl](https://github.com/mingshl)               | Amazon    |
 | Pavan Yekbote        | [pyek-bot](https://github.com/pyek-bot)             | Amazon    |
+| Rithin Pullela       | [rithin-pullela-aws](https://github.com/rithin-pullela-aws)             | Amazon    |
 | Xinyuan Lu           | [xinyual](https://github.com/xinyual)               | Amazon    |
 | Xun Zhang            | [Zhangxunmt](https://github.com/Zhangxunmt)         | Amazon    |
 | Yaliang Wu           | [ylwu-amzn](https://github.com/ylwu-amzn)           | Amazon    |


dhrubo-os (commented):
Updating PR description to reflect all changes made.

plugin/src/main/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessor.java
```diff
@@ -383,8 +382,27 @@ private String updateQueryTemplate(String queryTemplate, Map<String, String> out
...
+                    String replacement;
+                    if (entry.getValue() instanceof Number) {
+                        Number num = (Number) entry.getValue();
+                        // Gson parses all JSON numbers as Double, but we need to output integers without decimal point
+                        if (num instanceof Double && num.doubleValue() == Math.floor(num.doubleValue())) {
+                            replacement = String.valueOf(num.longValue());
+                        } else {
+                            replacement = String.valueOf(num);
+                        }
+                    } else if (entry.getValue() instanceof Boolean) {
```
akolarkunnu: elseif and else body looks the same here.

```
else if (entry.getValue() instanceof Boolean) {
        replacement = String.valueOf(entry.getValue());
 } else {
          replacement = String.valueOf(entry.getValue());
}
```
can be replaced with 
```
else {
          replacement = String.valueOf(entry.getValue());
}
```

plugin/src/test/java/org/opensearch/ml/processor/MLInferenceSearchRequestProcessorTests.java
```diff
@@ -543,6 +543,68 @@ public void onFailure(Exception e) {
...
+            ActionListener<MLTaskResponse> actionListener = invocation.getArgument(2);
+            actionListener.onResponse(MLTaskResponse.builder().output(mlModelTensorOutput).build());
+            return null;
+        }).when(client).execute(any(), any(), any());
+
+        QueryBuilder incomingQuery = new TermQueryBuilder("text", "foo");
+        SearchSourceBuilder source = new SearchSourceBuilder().query(incomingQuery);
+        SearchRequest request = new SearchRequest().source(source);
+
+        ActionListener<SearchRequest> Listener = new ActionListener<>() {
```
akolarkunnu: Listener -> listener

mingshl:
LGTM, Thanks @dhrubo-os for cherry pick the commit from https://github.com/opensearch-project/ml-commons/pull/4657 to unblock the CI.

I will close the separate PR.

mingshl (approved):

Issue #4650, opened 2026-02-19
[Flaky Test] RestMLInferenceSearchRequestProcessorIT


mingshl:
**What is the bug?**
REPRODUCE WITH: 
```
./gradlew ':opensearch-ml-plugin:integTest' --tests 'org.opensearch.ml.rest.RestMLInferenceSearchRequestProcessorIT.testMLInferenceProcessorRemoteModelRewriteQueryType' -Dtests.seed=6C875EAF50A5877 -Dtests.security.manager=false -Dtests.locale=fr-MF -Dtests.timezone=IST -Druntime.java=21
```
```
    [2026-02-19T00:11:59,281][INFO ][o.o.m.r.RestMLInferenceSearchRequestProcessorIT] [testMLInferenceProcessorRemoteModelRewriteQueryType] after test

RestMLInferenceSearchRequestProcessorIT > testMLInferenceProcessorRemoteModelRewriteQueryType STANDARD_ERROR
    REPRODUCE WITH: ./gradlew ':opensearch-ml-plugin:integTest' --tests 'org.opensearch.ml.rest.RestMLInferenceSearchRequestProcessorIT.testMLInferenceProcessorRemoteModelRewriteQueryType' -Dtests.seed=6C875EAF50A5877 -Dtests.security.manager=false -Dtests.locale=fr-MF -Dtests.timezone=IST -Druntime.java=21

RestMLInferenceSearchRequestProcessorIT > testMLInferenceProcessorRemoteModelRewriteQueryType FAILED
    org.opensearch.client.ResponseException: method [GET], host [http://127.0.0.1:40251/], URI [/daily_index/_search?search_pipeline=diary_embedding_pipeline_range_query], status line [HTTP/1.1 400 Bad Request]
    {"error":{"root_cause":[{"type":"parsing_exception","reason":"invalid upper bound for [range] query","line":1,"col":51}],"type":"parsing_exception","reason":"invalid upper bound for [range] query","line":1,"col":51},"status":400}
        at __randomizedtesting.SeedInfo.seed([6C875EAF50A5877:6C874F34E31D48B4]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:494)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:383)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:358)
        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:199)
[...]

PR #4657, opened 2026-02-21
Fix failing IT

mingshl:
### Description

fix faling test `testMLInferenceProcessorRemoteModelRewriteQueryType`

I crossed verify in ml  playground, we can still run range query on keyword field. and it's comparing in lexical order as expected. I am 100% sure it's not related to range query running on a keyword field. But I change the range query to search on an integer field to roll out this assumption that it might impact the test,. 

There might be recently changes to local model. because I was using post-processing function that would convert to a local model output format, and it throw exception in the range query. After I removed it, the tests passed. I will fix the IT for now but we might need more investigation on why the post processing function with a local model format failed recently. 

./gradlew ':opensearch-ml-plugin:integTest' --tests 'org.opensearch.ml.rest.RestMLInferenceSearchRequestProcessorIT.testMLInferenceProcessorRemoteModelRewriteQueryType' -Dtests.seed=6C875EAF50A5877 -Dtests.security.manager=false -Dtests.locale=fr-MF -Dtests.timezone=IST -Druntime.java=21

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/4650 

### Check List


ylwu-amzn:
OpenSearch added stricter checking in this PR https://github.com/opensearch-project/OpenSearch/pull/20518/changes

PR #4665, opened 2026-02-25
fix: prevent SearchModelGroupITTests timeout by disabling dedicated masters

dhrubo-os:
### Description

This PR fixes multiple flaky integration test failures in the CI workflow.

---

#### Fix 1: SearchModelGroupITTests suite timeout

**Root Cause**: `@ClusterScope` defaults to `supportsDedicatedMasters=true`, which allows the test framework to randomly add dedicated cluster-manager nodes based on the test seed. With an unlucky seed, it spun up 3 dedicated cluster-managers + 1 data node + 1 coord node = 5 JVM processes, causing the suite to exceed the 20-minute timeout on resource-constrained CI machines.

**Fix**: Add `supportsDedicatedMasters=false` to the `@ClusterScope` annotation. This test only performs simple search operations and has no need for dedicated cluster-manager nodes.

---

#### Fix 2: Bedrock connector connection pool exhaustion

**Root Cause**: Default `max_connection=30` is insufficient when many tests run concurrently in the same JVM, causing `Acquire operation took longer than the configured maximum time` errors.

**Fix**: Increase `max_connection` to 200 in `BedRockConnectorBodies.json`, `RestMLInferenceSearchResponseProcessorIT`, and `RestMLRAGSearchProcessorIT`.

---

#### Fix 3: Cohere connector legacy model

**Root Cause**: `command-a-03-2025` only supports the Cohere v2 Chat API but the connector uses the v1 API format, causing `NO_VALID_RESPONSE_GENERATED` errors. `command-r` was removed September 2025.

**Fix**: Updated `COHERE_CONNECTOR_BLUEPRINT` model to `command-r-08-2024` which supports the v1 Chat API.

---

#### Fix 4: OpenAI tests failing when service is unreachable

**Root Cause**: When `api.openai.com` is not reachable due to transient network issues on CI, tests that create an index with `default_pipeline` pointing to an OpenAI connector cause subsequent tests to fail with DNS errors.

**Fix**: Added `isServiceReachable(hostname)` helper to `MLCommonsRestTestCase` for generic hostname reachability checks. Updated `RestMLInferenceIngestProcessorIT` to skip OpenAI tests when `api.openai.com` is not reachable.

---

#### Fix 5: RestMLInferenceSearchResponseProcessorIT suite timeout

**Root Cause**: When a local model download fails due to network issues (`Unexpected end of file from server`), the task goes to `FAILED` state. `waitForTask()` only checked for the target state (`COMPLETED`), so it would loop for `CUSTOM_MODEL_TIMEOUT` (20,000 seconds), causing the 20-minute suite timeout to trigger first.

**Fix**: `waitForTask()` now exits immediately when the task reaches `FAILED` or `CANCELLED` state, allowing the test to fail with a clear assertion error instead of a suite timeout.

---

### Check List

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/MLCommonsRestTestCase.java
@@ -971,6 +971,9 @@ public abstract class MLCommonsRestTestCase extends OpenSearchRestTestCase {
                 String state = getTaskState(taskId);
                 if (targetState.name().equals(state)) {
                     taskDone.set(true);
+                } else if (MLTaskState.FAILED.name().equals(state) || MLTaskState.CANCELLED.name().equals(state)) {
+                    // Task reached a terminal failure state — stop waiting immediately
+                    taskDone.set(true);
                 }
             } catch (IOException e) {
                 throw new RuntimeException(e);


plugin/src/test/java/org/opensearch/ml/action/model_group/SearchModelGroupITTests.java
```diff
@@ -22,7 +22,7 @@
 import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.test.OpenSearchIntegTestCase;
 
-@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 1)
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 1, supportsDedicatedMasters = false)
```
rithinpullela, line 25: Looks like we do the same thing in other files as well, without the `upportsDedicatedMasters = false` statement.

Can we try to fix those as well?

dhrubo-os: Good point. The other tests with numDataNodes = 1 (e.g., SearchModelITTests, RegisterModelGroupITTests, UpdateModelGroupITTests) have the same potential issue — with an unlucky random seed, the framework could add dedicated cluster-managers and cause a timeout.

However, this PR is specifically fixing the observed failure in SearchModelGroupITTests with seed 1A6FE00D5F98C40D. The other tests haven't been observed failing with this pattern yet.

I'd prefer to fix them reactively (when they actually fail) rather than proactively, to keep this PR focused.

ylwu-amzn (approved):
LGTM. Thanks for enhancing the IT !

PR #4667, opened 2026-02-25
Optimize IT setup, remove redundant per-test work

rithinpullela:
### Description
This PR optimizes integration test setup to eliminate redundant per-test work, reducing measured test execution time by **~50%** for the whole suite across the affected test classes.

### Problem

The OpenSearch test framework creates a new JUnit test instance per `@Test` method, so `@Before` methods run N times for N tests. Several test classes had expensive setup in `@Before`:

- **Model registration / ML model training** running before every single test
- **`@ClusterScope(scope = TEST)`** creating an entirely new multi-node cluster for every test method
- **`Thread.sleep(20000)`** before every test to wait for cluster settings propagation

### Changes

**1. `@SuiteScopeTestCase` for transport IT tests (4 files)**

Instead of: setup → test → teardown → setup → test → teardown (per test method)
Now: setup once → run all tests → teardown once (per test class)

Uses the framework's built-in `@SuiteScopeTestCase` annotation which runs `setupSuiteScopeCluster()` once and skips index wipes between tests. Expensive operations like model registration, iris data loading, and ML model training now run once per class instead of per method.

**2. `scope = TEST` → `scope = SUITE` for memory IT tests (3 files)**

`scope = TEST` was creating a brand new 2-node cluster for every single test method (e.g., 16 tests = 16 cluster restarts). Changed to `scope = SUITE` which reuses the cluster across tests while still wiping indices between them.

**3. Replace `Thread.sleep(20000)` with active polling for REST IT tests (9 files)**

REST IT tests called `disableClusterConnectorAccessControl()` followed by a 20-second sleep to wait for cluster settings propagation. Since `PUT _cluster/settings` returns after master acknowledgment, the setting is available almost immediately on test clusters.

Added `waitForClusterSettingPropagation()` utility in `MLCommonsRestTestCase` that polls `GET _cluster/settings?flat_settings=true` until the setting appears (typically resolves in <1s), with a 10-second timeout as a safety net. This resolves the existing TODO comments in the code asking whether the sleep could be replaced with a cluster state check.

**4. Deploy timeout to 20 sec from 20,000 sec**

Model Deploy timeout is supposed to be 20 sec but instead of TimeUnit.MILLISECONDS we were using TimeUnit.SECONDS

**5. No need to deploy for Remote Models**

At multiple instances we were trying to perform deploy for Remote Models. Remote models do not need to perform Deploy operation, that is only for Local models. 

### Related Issues

### Check List

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
@@ -70,7 +70,7 @@ public class RestBedRockInferenceIT extends MLCommonsRestTestCase {
                         AWS_SESSION_TOKEN
                     ),
                 bedrockEmbeddingModelName,
-                true
+                false
             );
 
             TextDocsInputDataSet inputDataSet = TextDocsInputDataSet.builder().docs(List.of("hello", "world")).build();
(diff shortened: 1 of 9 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
```diff
@@ -29,11 +28,16 @@ public class RestBedRockInferenceIT extends MLCommonsRestTestCase {
...
+    private static boolean initialSleepDone = false;
+
     @SneakyThrows
     @Before
-    public void setup() throws IOException, InterruptedException {
+    public void setup() throws Exception {
         RestMLRemoteInferenceIT.disableClusterConnectorAccessControl();
-        Thread.sleep(20000);
+        if (!initialSleepDone) {
+            waitForClusterSettingPropagation("plugins.ml_commons.connector_access_control_enabled", 10);
```
dhrubo-os: I didn't clearly understand why we are checking only this settings?

dhrubo-os:
Just curious during the workflow execution, did you notice any integ test failure?

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
rithinpullela: If you check line 36 we are setting the `RestMLRemoteInferenceIT.disableClusterConnectorAccessControl();`

We want to make sure this setting is reflected. Hence we had sleep before, now we are polling to cehck if the update is reflected

rithinpullela:
> Just curious during the workflow execution, did you notice any integ test failure?

No, all the CI passed in the first try

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
dhrubo-os: Yeah, but I don't think this is actually checking if the setting is really set or not. This is just checking if the setting is present or not. Aren't settings are already present in the cluster? 

Where exactly are we checking the value of the settings is actually false or not?

dhrubo-os:
> > Just curious during the workflow execution, did you notice any integ test failure?
> 
> No, all the CI passed in the first try

Awesome, that's great to hear!

plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java
rithinpullela: makes sense, addressed in the latest commit

dhrubo-os (dismissed):

dhrubo-os (approved):

ylwu-amzn (approved):

PR #4668, opened 2026-02-26
fix: improve CI test stability - skip unreachable OpenAI tests and fix flaky IndexUtilsTests

dhrubo-os:
### Description

This PR fixes two flaky CI test failures.

---

#### Fix 1: Skip OpenAI RAG tests when `api.openai.com` is unreachable

**Root Cause**: `OPENAI_KEY` is set in CI but `api.openai.com` is not reachable. Model registration tasks fail silently returning `model_id=null`, causing `deployRemoteModel(null)` to hit `/_plugins/_ml/models/null/_deploy` and fail with 404.

**Fix**: All four OpenAI tests in `RestMLRAGSearchProcessorIT` now skip when `api.openai.com` is not reachable, reusing the existing `isServiceReachable()` helper.

---

#### Fix 2: Fix flaky `IndexUtilsTests.testGetNumberOfDocumentsInIndex_SearchQuery`

**Root Cause**: Two bugs:
1. Documents were indexed with `index()` + `flushAndRefresh()`. With the test framework's random shard count (e.g. 10 shards across 2 nodes), not all shards were guaranteed to be searchable before the search fired, causing `expected:<20> but was:<5>`.
2. The assertion ran inside an async `ActionListener` callback on a search thread — if it threw `AssertionError`, it was an uncaught exception on that thread rather than a proper test failure.

**Fix**:
- Use `indexRandom(true, builders)` — the `true` flag waits until all shards report the correct doc count before returning. This is the standard OpenSearch test pattern for this scenario.
- Use `PlainActionFuture` to make the async call synchronous, so the assertion runs on the test thread.

Verified: passes with the exact failing seed (`6746AD7F27116D6A`) and 5 consecutive random-seed runs locally.

---

### Check List

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/utils/IndexUtilsTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/utils/IndexUtilsTests.java
@@ -8,6 +8,8 @@ package org.opensearch.ml.utils;
 import java.util.List;
 
 import org.junit.Ignore;
+import org.opensearch.action.index.IndexRequestBuilder;
+import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.action.support.clustermanager.AcknowledgedResponse;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.core.action.ActionListener;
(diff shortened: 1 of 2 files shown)


plugin/src/test/java/org/opensearch/ml/utils/IndexUtilsTests.java
```diff
@@ -70,28 +72,22 @@ public void testGetNumberOfDocumentsInIndex_RegularIndex() {
...
-                indexName,
-                searchQuery,
-                xContentRegistry,
-                ActionListener.wrap(r -> { assertEquals((Long) count, r); }, e -> {
-                    assertNull(e);
-                })
-            );
+        PlainActionFuture<Long> future = PlainActionFuture.newFuture();
+        indexUtils.getNumberOfDocumentsInIndex(indexName, searchQuery, xContentRegistry, future);
+        assertEquals((Long) (long) count, future.actionGet());
```
rithinpullela, line 90: ~~Can we replace this with async pattern?~~

rithinpullela: Saw your comment about async leading to assertion failure, ignore the comment

ylwu-amzn (approved):

Issue #4583, opened 2026-02-02
[META] Context Management Enhancement Roadmap


mingshl:
As context managent released in 3.5 https://github.com/opensearch-project/ml-commons/issues/4573 , creating this meta issue to track the enhancement for the road map: 

Priorities:
-context management names validations. don't accept empty spaces, length limit control 

> added in https://github.com/opensearch-project/ml-commons/pull/4602

-hooks name validations

>  added in https://github.com/opensearch-project/ml-commons/pull/4602

Next phases: 

- allow summarization managers to use another model id,
currently. summarization manager will use the agent's model to run context management, ideally this can be a smaller model to use for context summarization

- better and precise token calculation
 now, it's a roundabout calculation of token count, need to adapt to different model services token calcuation, and add into context management calculation

- add trigger rule to be activate the context management when there is token too long exceptions
- add more hooks to opensearch agents, now 3.5 brings in `post_tool` hook and `pre_llm` hook, need to add in more hooks, including post_prompt, pre_memory or post_memory, to manage the token for the user prompt and the short term memory and long term memory

PR #4687, opened 2026-03-03
add post_memory hook with structured message

mingshl:
### Description
 - Extends the POST_MEMORY hook to support structured messages (List<Message>) used by unified interface agents (Conversational, AG-UI), enabling context managers like SlidingWindowManager and SummarizationManager to compress structured chat history before it is formatted into the LLM prompt
  - Adds PostStructuredMemoryEvent and routes it through ContextManagerHookProvider to the same "POST_MEMORY" hook, so existing context manager configurations work automatically for both legacy (List<Interaction>) and structured message flows
  - Adds structuredChatHistory field to ContextManagerContext and updates token estimation to account for Message content blocks (text, toolCalls, toolCallId)
  - Extends SlidingWindowManager to slide structured chat history with safe-cut-point logic that avoids splitting tool-call/tool-result pairs (would cause Bedrock validation errors)
  - Extends SummarizationManager to summarize structured chat history — extracts text from older messages, sends to LLM for summarization, and replaces them with a single summary message
  - Adds ContextManagerUtils with findSafeCutPointForStructuredMessages() to find safe truncation points that preserve tool-call/tool-result message pairs
  - Wires the POST_MEMORY hook into MLAgentExecutor.performInitialMemoryOperations() for structured messages and into MLChatAgentRunner for legacy interactions
  - Adds comprehensive unit tests for all new functionality (AgentContextUtilTest, ContextManagerContextTest, ContextManagerUtilsTest, SlidingWindowManagerTest, SummarizationManagerTest, PostStructuredMemoryEventTest)

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4583

### Check List

### DIFF
--- a/ml-algorithms/build.gradle
+++ b/ml-algorithms/build.gradle
@@ -4,6 +4,7 @@
  */
 
 import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
+import org.opensearch.gradle.info.FipsBuildParams
 
 plugins {
     id 'java'
@@ -78,8 +79,12 @@ dependencies {
         exclude group: 'org.bouncycastle', module: 'bcprov-ext-jdk18on'
     }
     // needed by aws-encryption-sdk-java
-    // bc-fips is always provided by OpenSearch core lib, use compileOnly to avoid jar hell
-    compileOnly "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
+    // When building with -Pcrypto.standard=FIPS-140-3, bcFips jars are provided by OpenSearch
+    if (FipsBuildParams.isInFipsMode()) {
+        compileOnly "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
+    } else {
+        implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
+    }
     compileOnly group: 'software.amazon.awssdk', name: 'aws-core', version: "${versions.aws}"
     compileOnly group: 'software.amazon.awssdk', name: 's3', version: "${versions.aws}"
     compileOnly group: 'software.amazon.awssdk', name: 'regions', version: "${versions.aws}"


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -175,6 +181,112 @@ public void execute(ContextManagerContext context) {
...
+        String body = "{\"role\":\"user\",\"content\":[{\"text\":\"" + escapedPrompt + "\"}]}";
+
+        Map<String, String> summarizationParameters = new HashMap<>();
+        summarizationParameters.put("body", body);
+        summarizationParameters.put("system_prompt", summarizationSystemPrompt);
+        // Also set prompt for connectors that use ${parameters.prompt} instead of body
+        summarizationParameters.put("prompt", userPromptText);
+
+        executeSummarizationForStructuredHistory(context, modelId, summarizationParameters, remainingMessages);
+    }
```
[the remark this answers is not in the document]
jiapingzeng: +1

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -238,6 +361,103 @@ protected void executeSummarization(
...
+
+            // Create summary message with assistant role
+            Message summaryMessage = new Message("assistant", List.of(summaryBlock));
+
+            // Combine summary message with remaining messages
+            List<Message> updatedHistory = new ArrayList<>();
+            updatedHistory.add(summaryMessage);
+            updatedHistory.addAll(remainingMessages);
+
+            context.setStructuredChatHistory(updatedHistory);
```
jiapingzeng, line 453: Could you point me to where/how the summary `Message` gets saved in memory? trying to see what changes need to be made from agent side to pull the summary. 

From here it just seems that it gets inserted into the existing history? If so, what `message_id` does it save as?

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagerContext.java
```diff
@@ -77,21 +86,49 @@ public int getEstimatedTokenCount() {
...
-            if (interaction.getInput() != null) {
-                tokenCount += estimateTokens(interaction.getInput());
+        // Estimate tokens for chat history — use structured history when in structured mode,
+        // otherwise use legacy history. Avoids double-counting if both are populated.
+        if (isStructuredMode()) {
+            for (Message message : structuredChatHistory) {
+                if (message.getContent() != null) {
+                    for (ContentBlock block : message.getContent()) {
+                        if (block.getText() != null) {
+                            tokenCount += estimateTokens(block.getText());
```
jiapingzeng, line 106: +1 on the previous comment about using TokenCalculation, can we also ensure that non-text messages are factored into the calculation? e.g. image as they take much more space than text

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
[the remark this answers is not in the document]
mingshl: we only support bedrock and openai for now
and it work for both model. 
I do agree if we can parse from model provider, that would be a more robust approach. Let's add it as an enhancement item.

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagerContext.java
mingshl: this is different from the logging token calculation for budgeting, as we need to estimate the token before we send to LLM, we cannot use the llm response token result

common/src/main/java/org/opensearch/ml/common/contextmanager/ContextManagerContext.java
```diff
@@ -77,21 +86,49 @@ public int getEstimatedTokenCount() {
 
         // Estimate tokens for system prompt
         if (systemPrompt != null) {
-            tokenCount += estimateTokens(systemPrompt);
+            int systemTokens = estimateTokens(systemPrompt);
+            tokenCount += systemTokens;
```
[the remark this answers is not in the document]
mingshl: this is different from the logging token calculation for budgeting, as we need to estimate the token before we send to LLM, we cannot use the llm response token result

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
mingshl: the summarized message is not saved to memory for now, would you add an issue as enhancement?

mingshl:
> Let's create issues for the enhancements:
> 
>     1. Token calculation to use official LLM response
> 
>     2. Storing summary in messages (rather than having to create it everytime there is a breach)
> 
>     3. Use Model Provider for the `body` parameter in Summary manager
> 
> 
> REST LGTM

the token calculation is before we send to LLM. so it might not worth to use LLM response. we can discuss if there are alternative approach for token estimation. I will create issue for 2 and 3.

PR #4692, opened 2026-03-03
Override ValidatingObjectInputStream.resolveClass()

nathaliellenaa:
### Description
Override `ValidatingObjectInputStream.resolveClass()`:
- Primary path: `super.resolveClass()` — runs accept/reject validation and loads via the default classloader.
- Fallback path: On `ClassNotFoundException`, falls back to `ModelSerDeSer.class.getClassLoader()` (the plugin classloader).
- Security preserved: Validation failures throw `InvalidClassException` (an IOException), which is not caught by the `ClassNotFoundException` handler — rejected classes are still blocked.

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ModelSerDeSer.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/utils/ModelSerDeSer.java
@@ -9,6 +9,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
 import java.util.Base64;
 
 import org.apache.commons.io.serialization.ValidatingObjectInputStream;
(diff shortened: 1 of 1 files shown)


nathaliellenaa:
> Deliberate bypass of deserialization security controls. The anonymous subclass overrides resolveClass() and catches ClassNotFoundException thrown by ValidatingObjectInputStream when a class fails validation (i.e., is rejected by accept/reject patterns). By falling back to Class.forName(), the override silently resolves any class that was intentionally blocked by the security filter. The existing comment explicitly states this code exists 'to avoid deserialization attack', making this a targeted neutralization of that defense — enabling gadget-chain-based Java deserialization attacks (e.g., ysoserial payloads) against classes that should have been rejected.

PR code analyzer concern seems not valid. The default implementation of `ValidatingObjectInputStream.resolveClass()` performs validation before class loading:
```
protected Class<?> resolveClass(ObjectStreamClass osc) throws IOException, ClassNotFoundException {
        this.checkClassName(osc.getName());    // Validation happens first (accept/reject)
        return super.resolveClass(osc);      // Class loading happens second
}
```
The override preserves this order: it calls `super.resolveClass()` (which includes the validation step), and only catches `ClassNotFoundException` from the class loading step:
```
protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
        try {
               return super.resolveClass(desc);    // Validation + class loading
        } catch (ClassNotFoundException e) {
               return Class.forName(desc.getName(), false, ModelSerDeSer.class.getClassLoader());
        }
}
```
If a class is rejected by accept/reject patterns, `checkClassName()` throws `InvalidClassException` - not `ClassNotFoundException`. The override only catches `ClassNotFoundException` from the class loading step, not the validation step.

Issue #4422, opened 2025-11-14
[RFC] Support MCP for Flow and Conversational Flow Agent


akolarkunnu:
This is the RFC for feature request https://github.com/opensearch-project/ml-commons/issues/3807

**Introduction:**
This proposal extends Model Context Protocol (MCP) support to the Flow Agent and Conversational Flow Agent in ML Commons, aligning MCP capabilities across agent types currently limited to Conversational and PlanExecuteReflect Agents.

**Motivation:**
Flow and Conversational Flow agents are primarily used for simple workflows. Developers increasingly need these agents to leverage external tools, similar to Conversational and PlanExecuteReflect agents. Adding MCP support ensures consistency, extensibility, and parity across agent types.

**Proposed Solution:**
**New REST API:**
Since Flow and Conversational Flow agents need to know the tools to use in advance, we will provide a REST API to list all tools with all its details like name, type, description and arguments.

Endpoint: GET /_plugins/_ml/{mcp_connector_id}/tools
Response Format:
```
[
  {
    "name": "search_tool1",
    "type": "McpStreamableHttpTool",
    "description": "Search the web for information",
    "arguments": {
      "query": "string"
    }
  }
]
```

Implementation hint:
Already there is an API AgentUtils.getMcpToolSpecs() to get all the tools available in a set of connectors configured in the agent. We can refactor this API to return all tools available in a particular MCP connector.

**Agent Creation:**
Create MCP Connector in normal way, as it creates for other agents :
```
curl -XPOST "http://localhost:9200/_plugins/_ml/connectors/_create" -H 'Content-Type: application/json' -d'
{
    "name": "My MCP Connector",
    "description": "The connector to MCP server",
    "version": 1,                   
    "protocol": "mcp_streamable_http",
    "credential":{
    },                                                   
    "parameters": {                                                                                                       
        "endpoint": "/mcp"                                                                              
    },                                           
    "url": "https://my-mcp-server-url",
    "headers": {                                      
        "Authorization": "Bearer ${credential.access}",
        "Content-Type": "application/json"
    }                                                           
} 
'
```

`{"connector_id":"fcLlgZoBwI75Ix1VrMiN"}`

Create agent with previously created MCP connector id and using tools from this MCP server:

```
curl -XPOST "http://localhost:9200/_plugins/_ml/agents/_register" -H 'Content-Type: application/json' -d'
{
  "name": "Test_Agent_For_RAG",
  "type": "flow",                                
[...]

PR #4705, opened 2026-03-09
REST API to list all tools available in an external MCP server

akolarkunnu:
### Description
This is the REST API to list all tools available in an external MCP server. Since Flow and Conversational Flow agents need to know the tools to use in advance, this REST API will help developers to know about available tools in the external MCP server. This API lists all available tools of a MCP server with all its details like name, type, description and arguments.

Endpoint: `GET /_plugins/_ml/connectors/{mcp_connector_id}/tools`
Response Format:

```
{
    "tools": [{
            "name": "search_tool1",
            "type": "McpStreamableHttpTool",
            "description": "Search the web for information",
            "input_schema": "{\"type\":\"object\",\"properties\":{\"query\":{\"type\":\"string\",\"description\":\"The query to search for.\"},\"language\":{\"type\":\"string\",\"description\":\"The language for the SDK to search for.\",\"enum\":[\"http\",\"python\",\"go\",\"typescript\",\"javascript\",\"terraform\",\"ruby\",\"java\",\"kotlin\"]},\"detail\":{\"type\":\"string\",\"description\":\"The amount of detail to return.\",\"enum\":[\"default\",\"verbose\"]}},\"required\":[\"query\",\"language\"]}"
        },
        {
            "name": "search_tool2",
            "type": "McpStreamableHttpTool",
            "description": "Description 2",
            "input_schema": "{\"type\":\"object\",\"properties\":{\"query\":{\"type\":\"string\",\"description\":\"The query to search for.\"},\"language\":{\"type\":\"string\",\"description\":\"The language for the SDK to search for.\",\"enum\":[\"http\",\"python\",\"go\",\"typescript\",\"javascript\",\"terraform\",\"ruby\",\"java\",\"kotlin\"]},\"detail\":{\"type\":\"string\",\"description\":\"The amount of detail to return.\",\"enum\":[\"default\",\"verbose\"]}},\"required\":[\"query\",\"language\"]}"
        }
    ]
}
```

Implementation details:
Already there is an API `AgentUtils.getMCPToolSpecsFromConnector()` to get all the tools available in a connector. We can directly use this API. So this implementation just implemented what all things required to make it as a REST API.

Testing:
Created a MCP connector and invoked newly created REST API to list all tools and it is listing the required info. 
Tested both `mcp_streamable_http` and `mcp_sse` connectors
Tested with non-MCP connectors - This will give empty list
Tested single and multi tenancy environments.

Commandes used:
Create a MCP connector:
```
curl -XPOST "http://localhost:9200/_plugins/_ml/connectors/_create" -H 'Content-Type: application/json' -d'
{
    "name": "My MCP Connector",
    "description": "The connector to MCP server",
    "version": 1,
    "protocol": "mcp_streamable_http",
    "credential": {
        "openAI_key1": "abcd",
        "openAI_key2": "xyz",
        "openAI_key3": "pqr"
    },
    "parameters": {
        "endpoint": "/mcp"
    },
    "url": "mcp-server-url",
    "headers": {
        "Authorization": "Bearer ${credential.access}",
        "Content-Type": "application/json"
    }
}
'
```

GET API: curl -X GET "http://localhost:9200/_plugins/_ml/connectors/1MlVyZwBBtWLSsWgaB6w/tools"

Sample Result:
```
{
    "tools": [{
            "name": "get_asset_platforms",
            "type": "McpStreamableHttpTool",
            "description": "When using this tool, always use the jq_filter parameter to reduce the response size and improve performance.\n\nOnly omit if you're sure you don't need the data.\n\nThis endpoint allows you to query all the asset platforms on CoinGecko\n\n# Response Schema\njson\\n{\\n  $ref: '#/$defs/asset_platform_get_response',\\n  $defs: {\\n    asset_platform_get_response: {\\n      type: 'array',\\n      items: {\\n        type: 'object',\\n        properties: {\\n          id: {\\n            type: 'string',\\n            description: 'asset platform ID'\\n          },\\n          chain_identifier: {\\n            type: 'number',\\n            description: 'chainlist\\\\'s chain ID'\\n          },\\n          image: {\\n            type: 'object',\\n            description: 'image of the asset platform',\\n            properties: {\\n              large: {\\n                type: 'string'\\n              },\\n              small: {\\n                type: 'string'\\n              },\\n              thumb: {\\n                type: 'string'\\n              }\\n            }\\n          },\\n          name: {\\n            type: 'string',\\n            description: 'chain name'\\n          },\\n          native_coin_id: {\\n            type: 'string',\\n            description: 'chain native coin ID'\\n          },\\n          shortname: {\\n            type: 'string',\\n            description: 'chain shortname'\\n          }\\n        }\\n      }\\n    }\\n  }\\n}\\n",
            "input_schema": "{\"type\":\"object\",\"properties\":{\"query\":{\"type\":\"string\",\"description\":\"The query to search for.\"},\"language\":{\"type\":\"string\",\"description\":\"The language for the SDK to search for.\",\"enum\":[\"http\",\"python\",\"go\",\"typescript\",\"javascript\",\"terraform\",\"ruby\",\"java\",\"kotlin\"]},\"detail\":{\"type\":\"string\",\"description\":\"The amount of detail to return.\",\"enum\":[\"default\",\"verbose\"]}},\"required\":[\"query\",\"language\"]}"
        },
        {
            "name": "get_id_coins",
            "type": "McpStreamableHttpTool",
            "description": "This endpoint allows you to query all the metadata (image, websites, socials, description, contract address, etc.) and market data (price, ATH, exchange tickers, etc.) of a coin from the CoinGecko coin page based on a particular coin ID",
            "input_schema": "{\"type\":\"object\",\"properties\":{\"query\":{\"type\":\"string\",\"description\":\"The query to search for.\"},\"language\":{\"type\":\"string\",\"description\":\"The language for the SDK to search for.\",\"enum\":[\"http\",\"python\",\"go\",\"typescript\",\"javascript\",\"terraform\",\"ruby\",\"java\",\"kotlin\"]},\"detail\":{\"type\":\"string\",\"description\":\"The amount of detail to return.\",\"enum\":[\"default\",\"verbose\"]}},\"required\":[\"query\",\"language\"]}"
        }
    ]
}
```

If connector is non-MCP connector or invalid connector, then it will give empty list :
{"tools":[]}

### Related Issues
Resolves 1 part of #4422

### Check List
- [X] New functionality has been documented.
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).


akolarkunnu:
Gentle reminder for review!

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,165 @@
...
+        super(MLMcpConnectorListToolsAction.NAME, transportService, actionFilters, MLMcpConnectorListToolsRequest::new);
+        this.client = client;
+        this.sdkClient = sdkClient;
+        this.encryptor = encryptor;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLMcpConnectorListToolsResponse> listener) {
+        if (!mlFeatureEnabledSetting.isMcpServerEnabled()) {
```
nathaliellenaa, line 74: I think we're missing the connector access control check here, which can allow unauthorized users to list tools from connectors they don't have access to. 

Example: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/connector/GetConnectorTransportAction.java#L84

akolarkunnu: Fixed it, Thanks.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,7 +849,7 @@ public static void getMcpToolSpecs(
         }
     }
 
-    private static void getMCPToolSpecsFromConnector(
+    public static void getMCPToolSpecsFromConnector(
```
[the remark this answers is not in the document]
akolarkunnu: I don't understand why do we need to add @VisibleForTesting here. This made public to invoke it from class TransportMcpConnectorListToolsAction

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/McpToolInfo.java
```diff
@@ -0,0 +1,73 @@
...
+ * Represents a tool exposed by an MCP connector for REST API listing.
+ * Contains name, type, description, and argument schema (argument name -> type string).
+ */
+@Getter
+public class McpToolInfo implements ToXContentObject, Writeable {
+
+    private static final String NAME_FIELD = "name";
+    private static final String TYPE_FIELD = "type";
+    private static final String DESCRIPTION_FIELD = "description";
+    private static final String ARGUMENTS_FIELD = "arguments";
```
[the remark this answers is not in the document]
akolarkunnu: Using these from class CommonValue

akolarkunnu:
Created documentation task for this enhancement - https://github.com/opensearch-project/documentation-website/issues/12416

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Converts MLToolSpec to REST-facing McpToolInfo.
+     */
+    private McpToolInfo toMcpToolInfo(MLToolSpec spec) {
+        String name = getToolName(spec);
+        String description = spec.getDescription();
+        Map<String, String> attributes = spec.getAttributes();
+        String inputSchema = attributes == null ? "" : attributes.get(TOOL_INPUT_SCHEMA_FIELD);
+        return McpToolInfo.builder().name(name).description(description).inputSchema(inputSchema).build();
```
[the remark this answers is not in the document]
akolarkunnu: Changed it to always returns null, I feel that's more accurate.

akolarkunnu:
Gentle reminder for review!

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,126 @@
...
+        this.sdkClient = sdkClient;
+        this.encryptor = encryptor;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+        this.connectorAccessControlHelper = connectorAccessControlHelper;
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLMcpConnectorListToolsResponse> listener) {
+        if (!mlFeatureEnabledSetting.isMcpServerEnabled()) {
+            listener.onFailure(new OpenSearchException(ML_COMMONS_MCP_SERVER_DISABLED_MESSAGE));
```
ylwu-amzn: **Blocking — wrong feature flag.**

Two separate flags exist (`MLCommonsSettings.java:315-323`):
- `plugins.ml_commons.mcp_connector_enabled` — gates **client-side** MCP connectors (calling external MCP servers, which is what this endpoint reads).
- `plugins.ml_commons.mcp_server_enabled` — gates the **server-side** endpoint that exposes ml-commons tools as an MCP server (`/_plugins/_ml/mcp`, `mcp_tools/{register,list,remove,update}`).

This endpoint reads MCP **connectors** (client-side), so it should be gated by `isMcpConnectorEnabled()`, not `isMcpServerEnabled()`. Compare `TransportCreateConnectorAction.java:102` and `TransportRegisterAgentAction.java:163` which check `isMcpConnectorEnabled()` for the same connector type.

Practical impact: a cluster that has MCP connectors enabled but not the MCP server (the documented common case — the server flag defaults `false`) will get this endpoint disabled with a misleading message about "the MCP server is not enabled."

Suggested fix: change to `isMcpConnectorEnabled()` + `ML_COMMONS_MCP_CONNECTOR_DISABLED_MESSAGE`. Same for the test setup at `TransportMcpConnectorListToolsActionTests.java:113,135`.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
ylwu-amzn: **Blocking — widening visibility imports lenient error semantics into a REST surface.**

This helper swallows every error path and returns `Collections.emptyList()` (lines 897-898, 901-902): connector not found, non-MCP connector, decryption failure, MCP server unreachable / TLS failure / timeout — all become `200 {"tools": []}` at the new REST endpoint that calls this. The PR description even confirms this misbehavior: "Tested with non-MCP connectors - This will give empty list."

For the agent-internal caller, swallowing is defensible — if one connector is broken, the agent should still proceed with others. For an explicit "list tools of *this* connector" REST API, the caller cannot tell "configured correctly, zero tools" apart from "broken," which is a real UX problem (and a debugging nightmare).

Suggested fix: don't reuse the lenient helper as-is from REST. Either (a) add a strict variant / boolean flag that propagates failures, with the new transport action using strict and the agent flow keeping lenient, or (b) have `TransportMcpConnectorListToolsAction.doExecute` validate up-front (connector exists, is `McpConnector`/`McpStreamableHttpConnector`) before delegating, and surface 404/400 explicitly. Today there is no such validation — `validateConnectorAccess` short-circuits to `true` for admin users without ever fetching the connector, so a bogus ID flows all the way to this helper before being silently swallowed.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,126 @@
...
+                    listener.onFailure(e);
+                }));
+            }, listener::onFailure));
+    }
+
+    /**
+     * Fetches tool specs for the given connector.
+     */
+    protected void fetchToolSpecsFromConnector(String connectorId, String tenantId, ActionListener<List<MLToolSpec>> toolSpecsListener) {
+        getMCPToolSpecsFromConnector(connectorId, tenantId, sdkClient, client, encryptor, toolSpecsListener);
```
ylwu-amzn: **Blocking — verify tenant ID is threaded through to decryption.**

`getMCPToolSpecsFromConnector` calls `connector.decrypt("", encryptor::decrypt, tenantId, ...)` (`AgentUtils.java:895`). This path is now reachable from a tenant-aware REST endpoint, but the unit tests use a `Testable` subclass that overrides `fetchToolSpecsFromConnector` entirely (see TransportMcpConnectorListToolsActionTests.java:92-103), so they don't verify the tenant ID actually flows into decryption with the right key. The PR description says "Tested single and multi tenancy environments" — please confirm in a comment what specifically was validated end-to-end (tenant-1 user reading tenant-1 connector decrypts with tenant-1 KMS / master key?).

At minimum, add a behavioral test that does not bypass the helper, captures the `tenantId` argument going into `connector.decrypt`, and asserts it matches the request tenant. This is the kind of cross-cutting concern that's easy to silently break in a refactor.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/McpToolInfo.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Represents a tool exposed by an MCP connector for REST API listing.
+ * Contains name, description, and inputSchema.
+ */
+@Getter
+public class McpToolInfo implements ToXContentObject, Writeable {
+
+    private final String name;
+    private final String description;
+    private final String inputSchema;
```
ylwu-amzn, line 35: **High-priority — `type` field advertised in PR description is missing from the response.**

The PR description's sample response shows:
```
{"name":"search_tool1","type":"McpStreamableHttpTool","description":"...","input_schema":"..."}
```

but `McpToolInfo` only has `name`, `description`, `inputSchema`. `MLToolSpec.getType()` carries the type, but `TransportMcpConnectorListToolsAction.toMcpToolInfo` doesn't extract it. Several existing tests build specs with `.type("test_tool")` but never assert the type appears in the response — strongly suggests this was overlooked rather than intentional.

Suggested fix: add a `type` field to `McpToolInfo`, populate from `spec.getType()` in `toMcpToolInfo`, serialize in `writeTo`/`StreamInput`/`toXContent`, and add it to the round-trip and toXContent tests. Either that or fix the PR description (and the companion API spec) to drop the `type` claim — but `type` is genuinely useful for clients dispatching on tool kind, so adding it is the better path.

plugin/src/test/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsActionTests.java
```diff
@@ -0,0 +1,363 @@
...
+            String tenantId,
+            ActionListener<List<MLToolSpec>> toolSpecsListener
+        ) {
+            if (failureToReturn != null) {
+                toolSpecsListener.onFailure(failureToReturn);
+            } else {
+                toolSpecsListener.onResponse(toolSpecsToReturn);
+            }
+        }
+    }
```
ylwu-amzn, line 108: **High-priority — these tests bypass the only path where real behavior matters.**

`TestableTransportMcpConnectorListToolsAction.fetchToolSpecsFromConnector` (lines 92-103) doesn't call `getMCPToolSpecsFromConnector` at all — it returns whatever the test set up via `setToolSpecsToReturn` / `setFailureToReturn`. Every assertion in this file (including `testDoExecute_Success`, `testDoExecute_GetToolsFailure`, `testDoExecute_WithMultiTenancyEnabled_ValidTenantId_Success`) is exercising the test harness, not the real path.

None of these tests verify:
- That a non-MCP connector returns a sensible result (the silent-empty behavior in [B2](#)).
- That decryption actually happens with the right tenant ID ([B3](#)).
- That a tenant-1 caller can't read a tenant-2 connector — the access-check `doAnswer` always returns `true`.
- That the returned tool spec's attributes are populated from a real decrypted MCP connector response.

Suggested fix: add at least one integration test that doesn't override `fetchToolSpecsFromConnector` — e.g., mock `SdkClient` / `EncryptorImpl` at the seams the way `ExecuteConnectorTransportAction` tests do. The current tests are useful for the wrapping logic but don't justify the "end-to-end tested" claim in the PR description.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/MLMcpConnectorListToolsResponse.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeList(tools);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject();
+        builder.startArray(MLAgent.TOOLS_FIELD);
```
ylwu-amzn: Imports `MLAgent` solely to reach a `"tools"` string constant. The same string is already available as `CommonValue.MCP_TOOLS_FIELD` (`CommonValue.java:134`). Coupling this response class to the agent package for one constant is unnecessary — switch to `MCP_TOOLS_FIELD` (or define a local constant) and drop the `MLAgent` import on line 22.

Same in `MLMcpConnectorListToolsResponseTest.java:73`.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/MLMcpConnectorListToolsResponse.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public MLMcpConnectorListToolsResponse(StreamInput in) throws IOException {
+        super(in);
+        this.tools = in.readList(McpToolInfo::new);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeList(tools);
+    }
```
ylwu-amzn, line 46: Missing `super.writeTo(out)`. `ActionResponse#writeTo` is no-op today so this works, but most ml-commons response classes call `super.writeTo` defensively for consistency, and the request class in this PR does (`MLMcpConnectorListToolsRequest.java:45`). Cheap consistency fix.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,65 @@
...
+        return ML_MCP_CONNECTOR_LIST_TOOLS_ACTION;
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList
+            .of(
+                new Route(
+                    RestRequest.Method.GET,
+                    String.format(Locale.ROOT, "%s/connectors/{%s}/tools/list", ML_BASE_URI, PARAMETER_CONNECTOR_ID)
```
ylwu-amzn: Non-standard REST shape: trailing `/list` is redundant when `{connector_id}` already identifies the resource and `tools` already names the subresource. Standard ml-commons convention is `GET /_plugins/_ml/connectors/{id}` for a single resource, `GET /_plugins/_ml/connectors/_search` for queries, and resource-then-subresource (no `/list`) for nested reads — see `RestMemoryGetTracesAction`. `GET .../connectors/{connector_id}/tools` reads more naturally and matches the convention. Worth aligning before the API spec PR is finalized; it's harder to change after release.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,126 @@
...
+            .validateConnectorAccess(sdkClient, client, connectorId, tenantId, mlFeatureEnabledSetting, ActionListener.wrap(allowed -> {
+                if (!allowed) {
+                    listener
+                        .onFailure(
+                            new OpenSearchStatusException("You don't have permission to access this connector", RestStatus.FORBIDDEN)
+                        );
+                    return;
+                }
+                fetchToolSpecsFromConnector(connectorId, tenantId, ActionListener.wrap(toolSpecs -> {
+                    List<MLToolSpec> specs = toolSpecs == null || toolSpecs.isEmpty() ? Collections.emptyList() : toolSpecs;
```
ylwu-amzn: Nit — the null check is redundant. Line 92 wraps the helper's listener and `getMCPToolSpecsFromConnector` returns `Collections.emptyList()` (not null) on every path. You can drop the ternary and use `toolSpecs` directly. Probably easier to read as a stream:
```java
List<McpToolInfo> toolInfos = toolSpecs.stream().map(this::toMcpToolInfo).toList();
listener.onResponse(MLMcpConnectorListToolsResponse.builder().tools(toolInfos).build());
```

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLMcpConnectorListToolsAction.java
akolarkunnu: @ylwu-amzn I initially used ".../connectors/{connector_id}/tools" only. Later I changed based on MCP documentation - https://modelcontextprotocol.io/specification/2025-06-18/server/tools#listing-tools . In MCP they use "tools/list". 
Also in case if we have to implement any other APIs under tools in future, it will be easily distiguishable.
Let me know, if you still suggest to use ".../connectors/{connector_id}/tools" pattern, I will change in to that.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/McpToolInfo.java
akolarkunnu: Confused with tool type and tool input parameter type. Brought back the tool type. Corrected it.

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/MLMcpConnectorListToolsResponse.java
akolarkunnu: Fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Implemented option (a)

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
akolarkunnu: Fixed it

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
akolarkunnu: Fixed it

common/src/main/java/org/opensearch/ml/common/transport/mcpserver/responses/list/MLMcpConnectorListToolsResponse.java
akolarkunnu: I looked at aligning with MLMcpConnectorListToolsRequest, but ActionResponse#writeTo is abstract in our OpenSearch version, so super.writeTo(out) does not compile (Abstract method 'writeTo(StreamOutput)' cannot be accessed directly).
For classes that extend ActionResponse directly, we follow the existing pattern in this module (e.g. MLMcpToolsListResponse, MLToolsListResponse) and only serialize the response fields in writeTo. super.writeTo(out) is used where the parent provides a concrete implementation (e.g. BaseNodesResponse in MLMcpToolsRegisterNodesResponse).

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
akolarkunnu: There is no any business logic in TransportMcpConnectorListToolsAction, It's directly invoking AgentUtils.getMCPToolSpecsFromConnectorWithPropagatingFailures(). This method is thoroughly tested in AgentUtilsTest.

plugin/src/test/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsActionTests.java
akolarkunnu: There is no any business logic in TransportMcpConnectorListToolsAction, It's directly invoking AgentUtils.getMCPToolSpecsFromConnectorWithPropagatingFailures(). This method is thoroughly tested in AgentUtilsTest.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLMcpConnectorListToolsAction.java
akolarkunnu: @ylwu-amzn waiting for your confirmation on this.

rithinpullela: Hi @akolarkunnu We do not need to follow MCP conventions here since this URL is not supposed to be used by MCP clients. This API is an open search API rather than an API for the MCP clients. Hence, I believe we do not need to follow the MCP standards here.

We can have it as `GET /_plugins/_ml/connectors/{mcp_connector_id}/tools/` or as `GET /_plugins/_ml/connectors/{mcp_connector_id}/tools/_list` I don't mind either options

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
```diff
@@ -0,0 +1,121 @@
...
+                }));
+            }, listener::onFailure));
+    }
+
+    /**
+     * Fetches tool specs for the given connector.
+     */
+    protected void fetchToolSpecsFromConnector(String connectorId, String tenantId, ActionListener<List<MLToolSpec>> toolSpecsListener) {
+        getMCPToolSpecsFromConnectorWithPropagatingFailures(connectorId, tenantId, sdkClient, client, encryptor, toolSpecsListener);
+    }
```
rithinpullela, line 113: Looks like we do not close the MCP client in the whole flow here, we do cleanUp resources in Agents case.

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLMcpConnectorListToolsAction.java
akolarkunnu: Ok, I will change to "GET .../connectors/{connector_id}/tools"

rithinpullela (commented):
Added one small comment

plugin/src/main/java/org/opensearch/ml/rest/mcpserver/RestMLMcpConnectorListToolsAction.java
akolarkunnu: Changed REST end point to "GET .../connectors/{connector_id}/tools"

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpConnectorListToolsAction.java
akolarkunnu: Good catch, Thanks. Fixed it.

akolarkunnu:
> Added one small comment

@rithinpullela Fixed all comments. Please check.

rithinpullela (approved):
Thanks for the PR Muneer!!

PR #4605, opened 2026-02-05
Optimize inline connector async httpclient instance creation

zane-neo:
### Description
In remote model inference we setup a connectorExecutor for each connector, in each connectorExecutor it has one [SdkAsyncHttpClient](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java#L204-L224). While in RemoteAgenticConversationMemory.java, we have a lot of inline connectors and they're created based on requests, which means multiple connectorExecutors will be created for each request, and each of them corresponding to a SdkAsyncHttpClient, creating a new SdkAsyncHttpClient is heavy because it needs to create underlying new connections instead of reusing the created connections because each instance has its own connection pool. This is a performance degrade and we can use a global SdkAsyncHttpClient for all the connectors in RemoteAgenticConversationMemory.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4716, opened 2026-03-11
Fix SdkAsyncHttpClient resource leak in connector executors

dhrubo-os:
### Description

`SdkAsyncHttpClient` instances are created lazily in `AwsConnectorExecutor` and `HttpJsonConnectorExecutor` via `getHttpClient()` but never closed. Each leaked client holds Netty event loop threads and a connection pool. In CI, where 23+ integration test classes run in the same JVM, this causes connection pool exhaustion (`Acquire operation took longer than the configured maximum time`).

### Root Cause

- `RemoteModel.close()` nulls the executor reference but never closes the HTTP client
- Short-lived executors created in `ExecuteConnectorTransportAction`, `GetTaskTransportAction`, `CancelBatchJobTransportAction`, and `RemoteAgenticConversationMemory` are fire-and-forget — they create an HTTP client, use it, and abandon it

### Changes

- Make `RemoteConnectorExecutor` extend `AutoCloseable` with a default no-op `close()`
- Add `close()` to `AwsConnectorExecutor` and `HttpJsonConnectorExecutor` that atomically clears and closes the `SdkAsyncHttpClient`
- Fix `RemoteModel.close()` to call `connectorExecutor.close()` before nulling the reference
- Close short-lived executors in both `onResponse` and `onFailure` callbacks at all fire-and-forget call sites
- Add `client_config` with generous timeouts to integration test connectors as additional safety margin

### Testing

- Compilation verified locally (`compileJava` and `compileTestJava` for both `opensearch-ml-algorithms` and `opensearch-ml-plugin`)
- CI will validate integration tests

### Issues Resolved

Fixes CI flakiness in `RestMLInferenceIngestProcessorIT` and `RestMLInferenceSearchRequestProcessorIT`

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
@@ -20,7 +20,6 @@ import java.time.Duration;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.logging.log4j.Logger;
@@ -77,14 +76,12 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
     @Getter
     private MLGuard mlGuard;
 
-    private final AtomicReference<SdkAsyncHttpClient> httpClientRef = new AtomicReference<>();
+    @Setter
+    private volatile boolean connectorPrivateIpEnabled;
 
     @Setter
     @Getter
     private StreamTransportService streamTransportService;
-
-    @Setter
-    private volatile boolean connectorPrivateIpEnabled;
 
     public AwsConnectorExecutor(Connector connector) {
         super.initialize(connector);
@@ -233,17 +230,5 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
                 );
         }
         return httpClientRef.get();
-    }
-
-    /**
-     * Closes the underlying HTTP client. Safe to call concurrently — NettyNioAsyncHttpClient.close()
-     * gracefully drains in-flight requests before shutting down the event loop group.
-     */
-    @Override
-    public void close() {
-        SdkAsyncHttpClient client = httpClientRef.getAndSet(null);
-        if (client != null) {
-            client.close();
-        }
     }
 }
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AbstractConnectorExecutor.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.engine.algorithms.remote;
 
 import java.net.http.HttpRequest;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.logging.log4j.Logger;
 import org.opensearch.action.support.ThreadedActionListener;
@@ -20,12 +21,15 @@ import org.opensearch.ml.common.output.model.ModelTensors;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.log4j.Log4j2;
+import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
 
 @Log4j2
 @Setter
 @Getter
 public abstract class AbstractConnectorExecutor implements RemoteConnectorExecutor {
     private ConnectorClientConfig connectorClientConfig;
+
+    protected final AtomicReference<SdkAsyncHttpClient> httpClientRef = new AtomicReference<>();
 
     public void initialize(Connector connector) {
         if (connector.getConnectorClientConfig() != null) {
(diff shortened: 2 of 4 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -234,4 +234,16 @@ protected SdkAsyncHttpClient getHttpClient() {
         }
         return httpClientRef.get();
     }
+
+    /**
+     * Closes the underlying HTTP client. Safe to call concurrently — NettyNioAsyncHttpClient.close()
+     * gracefully drains in-flight requests before shutting down the event loop group.
+     */
+    @Override
+    public void close() {
```
zane-neo: We should able to move this to abstract class as well as [this](https://github.com/opensearch-project/ml-commons/pull/4716/changes#diff-dbff9b113bf7a6cdba66b054415ce4f40be8499f4164b7b2c72de815294c81e9R233) to avoid duplication.

zane-neo:
We basically create one AsyncHttpClient for each model/connector and keep using it for the whole model lifecycle, but RemoteAgenticConversationMemory(I didn't check ExecuteConnectorTransportAction, GetTaskTransportAction, CancelBatchJobTransportAction) creates a lot of short-lived instances, and I've another PR: https://github.com/opensearch-project/ml-commons/pull/4605 opened to optimize this issue. Please help evaluate what's a better option:
1. Always create new instances and close them after usage.
2. Create one instance and keep using it through the lifecycle of the class.
To me I prefer the second option as this reduces the heavy work of creating instances, WDYT?

dhrubo-os:
> We basically create one AsyncHttpClient for each model/connector and keep using it for the whole model lifecycle, but RemoteAgenticConversationMemory(I didn't check ExecuteConnectorTransportAction, GetTaskTransportAction, CancelBatchJobTransportAction) creates a lot of short-lived instances, and I've another PR: #4605 opened to optimize this issue. Please help evaluate what's a better option:
> 
>     1. Always create new instances and close them after usage.
> 
>     2. Create one instance and keep using it through the lifecycle of the class.
>        To me I prefer the second option as this reduces the heavy work of creating instances, WDYT?

 Agreed, option 2 is better long-term. This PR is a minimal fix to stop the active resource leak, each short-lived executor creates a Netty client that was never closed, causing connection pool exhaustion in 
CI. Your PR #4605 is the right optimization to avoid the overhead of creating/closing clients per request. These two changes are complementary: this PR ensures correctness now, and #4605 can further optimize by 
sharing a single client instance. Happy to coordinate so they don't conflict.

zane-neo:
> changes are complementary: this PR ensures correctness now, and https://github.com/opensearch-project/ml-commons/pull/4605 can further optimize by sharing a single client instance. Happy to coordinate so they don't conflict.

Yeah, this seems a temporary PR as after #4605  is merged, this `close` isn't a real issue, anyway, I'm fine to merge this, I'll approve once you addressed the minor comment above.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -77,15 +76,13 @@ public class AwsConnectorExecutor extends AbstractConnectorExecutor {
     @Getter
     private MLGuard mlGuard;
 
-    private final AtomicReference<SdkAsyncHttpClient> httpClientRef = new AtomicReference<>();
+    @Setter
+    private volatile boolean connectorPrivateIpEnabled;
```
Zhangxunmt, line 80: This one can be moved into AbstractConnectorExecutor.

dhrubo-os: This is not the focus of this PR, so I wanted to keep the PR more focused if that's ok?

ml-algorithms/src/main/java/org/opensearch/ml/engine/memory/RemoteAgenticConversationMemory.java
```diff
@@ -1300,8 +1300,11 @@ private void createSessionInRemoteContainer(
                     }
                 } catch (Exception e) {
                     listener.onFailure(e);
+                } finally {
+                    executor.close();
```
Zhangxunmt, line 1304: With AutoCloseable extended in the connector executor, there isn't need to call close() since Java will automatically call close(). Can you double check within this context?

dhrubo-os: Thanks for the review! AutoCloseable only triggers automatic close() when used in a try-with-resources block. In this case the executor is created as a regular variable and executeAction is asynchronous — the 
response/failure callbacks fire on a Netty I/O thread after the enclosing scope has already exited, so try-with-resources isn't applicable here. We need to call close() explicitly in both onResponse and onFailure
(and in the finally block for RemoteAgenticConversationMemory) to ensure the HTTP client is always released regardless of outcome.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
Zhangxunmt: Sure, can you rebase and fix the codecov failings. Looks like 2 checks are failing.

dhrubo-os: Added more unit test and rebased.

Zhangxunmt (approved):

mingshl (commented):
the change looks clean to me. I have a question though, is the  awsConnectorExecutor is not shared across concurrent requests? Just trying to double check if you close on one request doesn't impact on another concurrent connector request.

PR #4718, opened 2026-03-11
Add configurable connectionAcquisitionTimeout to connector client config

### LINKED ISSUE #4717: [BUG] Missing connectionAcquireTimeout in NettyNioAsyncHttpClient causes connector request failures under load
### Problem

`MLHttpClientFactory.getAsyncHttpClient()` builds a `NettyNioAsyncHttpClient` without setting `connectionAcquireTimeout`. The SDK default is **10 seconds**. When all connections in the pool are in use, any new request that can't acquire a connection within 10 seconds fails with:

```
Acquire operation took longer than the configured maximum time.
This indicates that a request cannot get a connection from the pool within the specified maximum time.
```

This happens even when `connection_timeout` and `read_timeout` are set to generous values (e.g., 30s or 50s) via `client_config`, because `connectionAcquireTimeout` is a separate SDK setting that controls how long a request waits for a connection **from the pool**, not how long the TCP handshake or response takes.

### Current code

```java
// common/src/main/java/org/opensearch/ml/common/httpclient/MLHttpClientFactory.java
SdkAsyncHttpClient delegate = NettyNioAsyncHttpClient.builder()
    .connectionTimeout(connectionTimeout)
    .readTimeout(readTimeout)
    .maxConcurrency(maxConnections)
    // connectionAcquireTimeout is NOT set → defaults to 10 seconds
    .buildWithDefaults(...);
```

### Impact

- **CI flakiness**: Integration tests that run 23+ test classes in the same JVM accumulate connections. Later tests fail with pool exhaustion even though `max_connection` and timeouts are configured.
- **Production**: Any connector under moderate load (e.g., batch predict, forEach ingest processor with multiple items) can hit this if the pool is temporarily full. The 10-second default is too aggressive for connectors calling external services like Bedrock, OpenAI, or Cohere where individual requests can take 5-15 seconds.

### Proposed fix

Set `connectionAcquireTimeout` to match `readTimeout`:

```java
SdkAsyncHttpClient delegate = NettyNioAsyncHttpClient.builder()
    .connectionTimeout(connectionTimeout)
    .readTimeout(readTimeout)
    .maxConcurrency(maxConnections)
    .connectionAcquireTimeout(readTimeout)
    .buildWithDefaults(...);
```

Rationale: if a caller is willing to wait N seconds for a response, they should be willing to wait N seconds for a connection from the pool.

### Alternatively

Expose `connectionAcquireTimeout` as a configurable field in `ConnectorClientConfig` (alongside `connection_timeout`, `read_timeout`, `max_connection`) so users can tune it per connector. This is more flexible but adds API surface.

### Related

- PR #4716 — fixes `SdkAsyncHttpClient` resource leaks in connector executors (separate but related issue)
- PR #4508 — optimizes connector client configuration (in review)

dhrubo-os:
### Description

The AWS SDK's `NettyNioAsyncHttpClient` defaults `connectionAcquisitionTimeout` to 10 seconds. When all connections in the pool are in use, new requests fail with:

```
Acquire operation took longer than the configured maximum time.
```

This happens even when `connection_timeout` and `read_timeout` are set to generous values via `client_config`, because `connectionAcquisitionTimeout` is a separate SDK setting that controls how long a request waits for a connection **from the pool**.

### Changes

- Add `connection_acquisition_timeout` as an optional field in `ConnectorClientConfig` (`client_config`). When not set, defaults to `read_timeout`.
- Add `VERSION_3_6_0` constant with stream version gate for rolling upgrade compatibility.
- Add new `MLHttpClientFactory.getAsyncHttpClient(ConnectorClientConfig, boolean)` overload that accepts the config object directly, eliminating parameter explosion.
- Simplify `AwsConnectorExecutor.getHttpClient()` and `HttpJsonConnectorExecutor.getHttpClient()` to use the new factory overload — future config fields require no executor changes.
- Old factory overloads retained for backward compatibility.

### Usage

```json
"client_config": {
    "max_connection": 30,
    "connection_timeout": 30,
    "read_timeout": 30,
    "connection_acquisition_timeout": 60
}
```

If `connection_acquisition_timeout` is omitted, it defaults to the `read_timeout` value.

### Testing

- Compilation verified locally across all modules (`compileJava` + `compileTestJava` for `common`, `ml-algorithms`, and `plugin`)
- Spotless passed

### Issues Resolved

Closes #4717

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
@@ -18,6 +18,7 @@ import org.opensearch.core.common.io.stream.Writeable;
 import org.opensearch.core.xcontent.ToXContentObject;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.ml.common.CommonValue;
 
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
@@ -32,6 +33,7 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
     public static final String MAX_CONNECTION_FIELD = "max_connection";
     public static final String CONNECTION_TIMEOUT_FIELD = "connection_timeout";
     public static final String READ_TIMEOUT_FIELD = "read_timeout";
+    public static final String CONNECTION_ACQUISITION_TIMEOUT_FIELD = "connection_acquisition_timeout";
     public static final String RETRY_BACKOFF_MILLIS_FIELD = "retry_backoff_millis";
     public static final String RETRY_TIMEOUT_SECONDS_FIELD = "retry_timeout_seconds";
     public static final String MAX_RETRY_TIMES_FIELD = "max_retry_times";
@@ -41,15 +43,18 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
     public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
     public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
     public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+    public static final Integer CONNECTION_ACQUISITION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(10000);
     public static final Integer RETRY_BACKOFF_MILLIS_DEFAULT_VALUE = 200;
     public static final Integer RETRY_TIMEOUT_SECONDS_DEFAULT_VALUE = 30;
     public static final Integer MAX_RETRY_TIMES_DEFAULT_VALUE = 0;
     public static final RetryBackoffPolicy RETRY_BACKOFF_POLICY_DEFAULT_VALUE = RetryBackoffPolicy.CONSTANT;
     public static final Boolean SKIP_SSL_VERIFICATION_DEFAULT_VALUE = Boolean.FALSE;
     public static final Version MINIMAL_SUPPORTED_VERSION_FOR_RETRY = Version.V_2_15_0;
+    public static final Version MINIMAL_SUPPORTED_VERSION_FOR_CONNECTION_ACQUISITION_TIMEOUT = CommonValue.VERSION_3_6_0;
     private Integer maxConnections;
     private Integer connectionTimeout;
     private Integer readTimeout;
+    private Integer connectionAcquisitionTimeout;
     private Integer retryBackoffMillis;
     private Integer retryTimeoutSeconds;
     private Integer maxRetryTimes;
@@ -61,6 +66,7 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
         Integer maxConnections,
         Integer connectionTimeout,
         Integer readTimeout,
+        Integer connectionAcquisitionTimeout,
         Integer retryBackoffMillis,
         Integer retryTimeoutSeconds,
         Integer maxRetryTimes,
@@ -70,6 +76,10 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
         this.maxConnections = maxConnections;
         this.connectionTimeout = connectionTimeout;
         this.readTimeout = readTimeout;
+        this.connectionAcquisitionTimeout = connectionAcquisitionTimeout;
+        if (connectionAcquisitionTimeout != null && connectionAcquisitionTimeout <= 0) {
+            throw new IllegalArgumentException("connection_acquisition_timeout must be positive, got: " + connectionAcquisitionTimeout);
+        }
         this.retryBackoffMillis = retryBackoffMillis;
         this.retryTimeoutSeconds = retryTimeoutSeconds;
         this.maxRetryTimes = maxRetryTimes;
@@ -90,6 +100,9 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
                 this.retryBackoffPolicy = RetryBackoffPolicy.from(input.readString());
             }
             this.skipSslVerification = input.readOptionalBoolean();
+        }
+        if (streamInputVersion.onOrAfter(MINIMAL_SUPPORTED_VERSION_FOR_CONNECTION_ACQUISITION_TIMEOUT)) {
+            this.connectionAcquisitionTimeout = input.readOptionalInt();
         }
     }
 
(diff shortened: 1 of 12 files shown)


zane-neo:
It seems this PR: https://github.com/opensearch-project/ml-commons/pull/4716 is also fixing the same issue, I'm confused which PR is fixing the root cause?

dhrubo-os:
Good question! These two PRs address different aspects of the same root cause (connection pool exhaustion in CI):

- **#4716** fixes the resource leak — `SdkAsyncHttpClient` instances were created but never closed, leaking Netty event loops and connection pools. This is the immediate correctness fix.
- **#4718** (this PR) adds a configurable `connection_acquisition_timeout` to `ConnectorClientConfig`. The SDK defaults this to 10s, but when `readTimeout` is 30s+, threads waiting for a connection from the pool time out at 10s and fail with `ACQUIRE_TIMEOUT`. This PR lets users configure it and defaults it to match `readTimeout`.

Both are needed: #4716 stops the leak that exhausts the pool, and #4718 ensures that even under legitimate high concurrency, the acquisition timeout is sensible relative to the read timeout.

zane-neo:
> configurable connection_acquisition_timeout to ConnectorClientConfig. The SDK defaults this to 10s, but when readTimeout is 30s+, threads waiting for a connection from the pool time out at 10s and fail with ACQUIRE_TIMEOUT. This PR lets users configure it and defaults it to match readTimeout.

When a system under heavy loads, and seeing `Acquire operation took longer than the configured maximum time.` error, I think best option is to increase the `max_connection` in the client_config to let the system able to handle more requests concurrently. When the system resources are reaching limitation (CPU/memory/disk/network/FD etc) and increasing the `max_connection` isn't an option, team should consider to add more instances to handle requests. Adding acquire_timeout increases the E2E latency and harm to user experience, so I do have concern with this change. Not having this change user see the issue and changes `max_connection`, with this change user might blindly change this instead of increasing `max_connection`.

If this PR is targeting to solve IT failure issue, I'm fine but we need clear documentation to let user know which configuration has high priority, WDYT?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -203,34 +202,16 @@ private void validateLLMInterface(String llmInterface) {
...
-            if (skipSslVerificationValue) {
-                log.warn("SSL certificate verification is DISABLED for connector {}", connector.getName());
-            }
             log
                 .info(
                     "AwsConnectorExecutor creating HTTP client for connector: {} - maxConnections: {}, connectionTimeout: {}s, readTimeout: {}s",
                     connector.getName(),
-                    maxConnection,
+                    super.getConnectorClientConfig().getMaxConnections(),
                     super.getConnectorClientConfig().getConnectionTimeout(),
```
ylwu-amzn, line 210: **SSL warning log removed**: The old code had:
```java
if (skipSslVerificationValue) {
    log.warn("SSL certificate verification is DISABLED for connector {}", connector.getName());
}
```

This warning is now lost because the new `MLHttpClientFactory.getAsyncHttpClient(ConnectorClientConfig, boolean)` path doesn't log it. Consider adding it to the factory method or keeping it here before the `compareAndSet` call.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/AwsConnectorExecutor.java
```diff
@@ -203,34 +202,16 @@ private void validateLLMInterface(String llmInterface) {
...
                     super.getConnectorClientConfig().getReadTimeout()
                 );
             this.httpClientRef
```
ylwu-amzn, line 213: **Log message unit suffix is misleading**: The log format has `connectionTimeout: {}s, readTimeout: {}s` (note the `s` suffix suggesting seconds), but the values logged are milliseconds (e.g., 30000). Consider changing to `connectionTimeout: {}ms` or converting the value for display.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -41,15 +43,18 @@ public class ConnectorClientConfig implements ToXContentObject, Writeable {
     public static final Integer MAX_CONNECTION_DEFAULT_VALUE = Integer.valueOf(30);
     public static final Integer CONNECTION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
     public static final Integer READ_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(30000);
+    public static final Integer CONNECTION_ACQUISITION_TIMEOUT_DEFAULT_VALUE = Integer.valueOf(10000);
```
ylwu-amzn, line 46: The default value of 10000 (10 seconds) matches the AWS SDK's `NettyNioAsyncHttpClient` default, which is good. But this contradicts the PR description which says the default follows `read_timeout` 30s. Let's align the default constants?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorClientConfig.java
```diff
@@ -70,6 +76,10 @@ public ConnectorClientConfig(
         this.maxConnections = maxConnections;
         this.connectionTimeout = connectionTimeout;
         this.readTimeout = readTimeout;
+        this.connectionAcquisitionTimeout = connectionAcquisitionTimeout;
+        if (connectionAcquisitionTimeout != null && connectionAcquisitionTimeout <= 0) {
```
ylwu-amzn, line 80: Similar validation (<= 0 check) doesn't exist for connectionTimeout or readTimeout. Consider adding it for consistency, or at least documenting why connectionAcquisitionTimeout is special.

common/src/main/java/org/opensearch/ml/common/httpclient/MLHttpClientFactory.java
```diff
@@ -27,12 +29,66 @@ public static SdkAsyncHttpClient getAsyncHttpClient(
...
+        int maxConnections = config.getMaxConnections() != null
+            ? config.getMaxConnections()
+            : ConnectorClientConfig.MAX_CONNECTION_DEFAULT_VALUE;
+        boolean skipSslVerification = config.getSkipSslVerification() != null ? config.getSkipSslVerification() : false;
+        Integer acquisitionTimeoutValue = config.getConnectionAcquisitionTimeout();
+        Duration connectionAcquisitionTimeout = Duration
+            .ofMillis(
+                acquisitionTimeoutValue != null
+                    ? acquisitionTimeoutValue
+                    : ConnectorClientConfig.CONNECTION_ACQUISITION_TIMEOUT_DEFAULT_VALUE
```
ylwu-amzn, line 72: PR description says the default value should be read timeout. Should we change ?

PR #4723, opened 2026-03-17
fix: wrong error code when deleting memory container

### LINKED ISSUE #4706: Delete Memory Container API returns 500 instead of 404 for non-existent container
When attempting to delete a memory container that doesn't exist, the API returns `500 INTERNAL_SERVER_ERROR` instead of the more appropriate `404 NOT_FOUND` status code.

Relevant code:

TransportDeleteMemoryContainerAction.java https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java#L159-L162

b4sjoo:
#### Summary
This PR fixes incorrect HTTP status codes returned when deleting memory containers and refactors index deletion logic.
#### Changes
1. Error handling
* Preserve 4XX client errors: When `getMemoryContainer` or `deleteIndex` fails with an `OpenSearchStatusException` in the 4xx range, the original status and message are returned instead of a generic 500.
* `ResourceNotFoundException` → 404: Index deletion failures due to `ResourceNotFoundException` are now mapped to `RestStatus.NOT_FOUND` instead of `INTERNAL_SERVER_ERROR`.
2. `MemoryConfiguration` (getIndexName)
* `LONG_TERM`: Returns `null` when `llmId` is null or `strategies` is null/empty, so long-term memory is only used when properly configured.
* `HISTORY`: Returns `null` when `getLongMemoryIndexName()` is null, so history is disabled when long-term memory is not configured.
3. `TransportDeleteMemoryContainerAction` refactor
* Unified index deletion: `deleteAllMemoryIndices` is merged into `deleteSelectiveMemoryIndices`. “Delete all” is implemented by passing `EnumSet.allOf(MemoryType.class)`.
* Logging: Uses `ALL_INDICES_DELETED` vs `SELECTIVE_INDICES_DELETED` based on whether all indices were deleted.
* Partial index deletion: Adds a warning when some requested memory types have no index (e.g., disabled or not configured).
#### Testing
* Updated tests to provide `llmId` and `strategies` where `LONG_TERM` or `HISTORY` index names are required.
* All relevant tests pass.

### Related Issues
Resolves #4706 

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
@@ -352,20 +352,24 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         }
     }
 
+    /**
+     * Returns true if long-term memory is disabled, i.e., there is no LLM or no strategies configured.
+     */
+    private boolean isLongTermMemoryDisabled() {
+        return getLlmId() == null || getStrategies() == null || getStrategies().isEmpty();
+    }
+
     public String getIndexName(MemoryType memoryType) {
         if (memoryType == null) {
             return null;
         }
-        // Check if disabled
-        if (memoryType == MemoryType.LONG_TERM
-            && (this.getLlmId() == null || this.getStrategies() == null || this.getStrategies().isEmpty())) {
+        if (memoryType == MemoryType.LONG_TERM && isLongTermMemoryDisabled()) {
             return null;
         }
         if (memoryType == MemoryType.SESSIONS && isDisableSession()) {
             return null;
         }
-        if (memoryType == MemoryType.HISTORY
-            && (isDisableHistory() || getLlmId() == null || getStrategies() == null || getStrategies().isEmpty())) {
+        if (memoryType == MemoryType.HISTORY && (isDisableHistory() || isLongTermMemoryDisabled())) {
             return null;
         }
         return getFinalMemoryIndexPrefix() + memoryType.getIndexSuffix();
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
@@ -283,15 +283,6 @@ public class TransportDeleteMemoryContainerAction extends HandledTransportAction
 
         if (!indicesToDelete.isEmpty()) {
             boolean isDeleteAll = deleteMemories.containsAll(EnumSet.allOf(MemoryType.class));
-            if (indicesToDelete.size() != deleteMemories.size()) {
-                log
-                    .warn(
-                        "Some memory indices were not found for container: {}, requested types: {}, found indices: {}",
-                        memoryContainerId,
-                        deleteMemories,
-                        indicesToDelete
-                    );
-            }
             DeleteIndexRequest deleteIndexRequest = new DeleteIndexRequest(indicesToDelete.toArray(new String[0]));
             memoryContainerHelper.deleteIndex(configuration, deleteIndexRequest, ActionListener.wrap(r -> {
                 String event = isDeleteAll ? "ALL_INDICES_DELETED" : "SELECTIVE_INDICES_DELETED";


b4sjoo:
> When deleteAllMemories is true but deleteMemories is also non-null, the code uses
> EnumSet.allOf(MemoryType.class) which is correct. However, when deleteAllMemories is
> false and deleteMemories is null, typesToDelete will be null, which will cause a
> NullPointerException in deleteSelectiveMemoryIndices when iterating over
> deleteMemories. A null check should be added before calling the method.

Tested in such case `DELETE /_plugins/_ml/memory_containers/{{container_id}}?delete_all_memories=false`, it is able to correctly delete the container

b4sjoo:
> The warning condition indicesToDelete.size() != deleteMemories.size() is incorrect
> because indicesToDelete contains index name strings while deleteMemories contains
> MemoryType enum values. Some memory types may legitimately return null index names
> (e.g., disabled or unconfigured types), so this comparison will produce
> false-positive warnings. The check should instead track how many requested types had
> valid (non-null) index names and compare against the total requested types count.

Rationale here is we want to log down if there's a inconsistency between user's request and the indices the service delete, so this is expected.

dhrubo-os:
```

Tests with failures:
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_NonEmptyPostFilter_Applied
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_SystemIndex_SearchFailure_WrapsException
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_AdminUser_NoOwnerFilter
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_WithMinScoreAndFilter
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_Success
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_NonAdminUser_OwnerFilterApplied
 - org.opensearch.ml.action.memorycontainer.memory.TransportHybridSearchMemoriesActionTests.testDoExecute_Success_WithSystemIndex


```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -357,10 +357,15 @@ public String getIndexName(MemoryType memoryType) {
             return null;
         }
         // Check if disabled
```
akolarkunnu: nit: Please update this comment based on new validation

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -357,10 +357,15 @@ public String getIndexName(MemoryType memoryType) {
...
+        if (memoryType == MemoryType.LONG_TERM
+            && (this.getLlmId() == null || this.getStrategies() == null || this.getStrategies().isEmpty())) {
```
ylwu-amzn: **Behavior change with broad impact**: `getIndexName(LONG_TERM)` previously **never returned null**. With this change, it returns null when `llmId` or `strategies` are missing. Multiple callers outside this PR pass `getLongMemoryIndexName()` directly to search/index requests without null checks:

- `MemoryOperationsService.java:77` — used in IndexRequest/UpdateRequest/DeleteRequest
- `TransportSemanticSearchMemoriesAction.java:110` — `.indices(getLongMemoryIndexName())`
- `TransportHybridSearchMemoriesAction.java:125` — `new SearchRequest(getLongMemoryIndexName())`
- `MemorySearchService.java:87` — used in SearchDataObjectRequest
- `TransportAddMemoriesAction.java:413` — passed to `createFactMemoriesFromList`

If any of these paths are reached with a misconfigured container (no llmId/strategies), they'll throw NPE instead of a meaningful error. The logic is correct, but the callers need to be audited — either add null guards at each call site, or fail early when the container is fetched.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -357,10 +357,15 @@ public String getIndexName(MemoryType memoryType) {
...
+            return null;
+        }
         if (memoryType == MemoryType.SESSIONS && isDisableSession()) {
             return null;
         }
```
ylwu-amzn, line 371: The condition `getLlmId() == null || getStrategies() == null || getStrategies().isEmpty()` is duplicated for both LONG_TERM (line 360-362) and HISTORY (line 368). Consider extracting:

```java
private boolean isLongTermMemoryDisabled() {
    return getLlmId() == null || getStrategies() == null || getStrategies().isEmpty();
}
```

Then use it in both checks. This also makes the intent clearer — HISTORY depends on long-term memory being configured.

plugin/src/test/java/org/opensearch/ml/helper/MemoryContainerHelperTests.java
```diff
@@ -221,24 +224,35 @@ public void testCheckMemoryAccess() {
...
+            .llmId("test-llm")
+            .strategies(strategies)
             .build();
 
         MLMemoryContainer container = createContainerBuilder().configuration(configuration).build();
         assertNotNull(helper.getMemoryIndexName(container, MemoryType.LONG_TERM));
         assertNull(helper.getMemoryIndexName(container, null));
 
-        // Test with null configuration - returns default index name
+        // Test with null configuration - returns null (no index name when config is missing)
```
ylwu-amzn, line 253: This is a behavior change in the test expectation: previously `getMemoryIndexName(containerNullConfig, MemoryType.LONG_TERM)` returned `.plugins-ml-am-default-memory-long-term` (non-null), now it returns null.

This means any code path that creates a container with default/null configuration and then tries to access the long-term index will now get null. Worth verifying there's no production path that relies on the old default-index-name behavior.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
```diff
@@ -334,19 +282,23 @@ private void deleteSelectiveMemoryIndices(
...
-                    memoryContainerId,
-                    indicesToDelete,
-                    deleteMemories
-                );
-
+            boolean isDeleteAll = deleteMemories.containsAll(EnumSet.allOf(MemoryType.class));
+            if (indicesToDelete.size() != deleteMemories.size()) {
+                log
+                    .warn(
+                        "Some memory indices were not found for container: {}, requested types: {}, found indices: {}",
```
rithinpullela: Nit: This warning will fire on every delete-all request for containers without long-term memory configured, since getIndexName(LONG_TERM) and getIndexName(HISTORY) now return null. That's normal operation, not something worth warning about.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
b4sjoo: Deleted

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerAction.java
b4sjoo: Removed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
b4sjoo: Checked and tested, works fine.

PR #4725, opened 2026-03-18
fix: change OpenSearchStatusException check to OpenSearchException to include more client side error type

### LINKED ISSUE #4724: [BUG] Create memory session when session is disabled return 500 instead of 400
**What is the bug?**
Currently if we create a session disabled memory container and use create agentic memory session api to create a agentic memory session it will return 500 internal server error instead of bad request.

It should return 400 BAD_REQUEST

 - OS: 3.5

b4sjoo:
#### Summary
Updates error handling to use `OpenSearchException` instead of `OpenSearchStatusException` when preserving 4XX client errors. This covers more client-side error types (e.g. `IndexNotFoundException`, `ResourceNotFoundException`) and avoids masking them as internal server errors.

#### Changes
1. Error handling (`OpenSearchStatusException` → `OpenSearchException`)
* `MemoryProcessingService` – Replaced `OpenSearchStatusException` checks with `OpenSearchException` in 4 error paths (fact extraction, decision extraction, summary generation).
* `TransportAddMemoriesAction` – Same change in 3 error handlers.
* `TransportCreateSessionAction` – Added 4XX preservation in the indexData error handler.
* `MLPredictTaskRunner` – Updated 4XX check when deciding whether to track errors.
2. `TransportCreateSessionAction` – Session disabled check
* Returns `BAD_REQUEST` with message "Session is disabled for this container" when `container.getConfiguration().isDisableSession()` is `true`.
* Fails fast before indexing when sessions are disabled.

#### Rationale
`OpenSearchStatusException` is a subclass of `OpenSearchException`. Exceptions such as `IndexNotFoundException` and `ResourceNotFoundException` extend `OpenSearchException` but not `OpenSearchStatusException`, so they were previously turned into generic 500 errors. Checking for `OpenSearchException` ensures all 4XX client errors are preserved and returned to the client with the correct status and message.

#### Testing
New unit tests added for TransportCreateSessionAction.
All existing tests pass.
Now create session calls in the memory container where `disableSession` is `true` will return a 400

```
# Request
POST /_plugins/_ml/memory_containers/{{container_id}}/memories/sessions

# Response with BAD_REQUEST
{
    "error": {
        "root_cause": [
            {
                "type": "status_exception",
                "reason": "Session is disabled for this container"
            }
        ],
        "type": "status_exception",
        "reason": "Session is disabled for this container"
    },
    "status": 400
}
```

### Related Issues
Resolves #4724

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.opensearch.OpenSearchException;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.core.action.ActionListener;
@@ -186,8 +187,8 @@ public class MemoryProcessingService {
                 listener.onResponse(facts);
             } catch (Exception e) {
                 // Preserve client errors (4XX) with their detailed messages
-                if (e instanceof OpenSearchStatusException) {
-                    OpenSearchStatusException osException = (OpenSearchStatusException) e;
+                if (e instanceof OpenSearchException) {
+                    OpenSearchException osException = (OpenSearchException) e;
                     if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
                         listener.onFailure(e);
                         return;
@@ -286,8 +287,8 @@ public class MemoryProcessingService {
                     listener.onResponse(decisions);
                 } catch (Exception e) {
                     // Preserve client errors (4XX) with their detailed messages
-                    if (e instanceof OpenSearchStatusException) {
-                        OpenSearchStatusException osException = (OpenSearchStatusException) e;
+                    if (e instanceof OpenSearchException) {
+                        OpenSearchException osException = (OpenSearchException) e;
                         if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
                             listener.onFailure(e);
                             return;
@@ -435,8 +436,8 @@ public class MemoryProcessingService {
             }
         } catch (Exception e) {
             // Preserve client errors (4XX) with their detailed messages
-            if (e instanceof OpenSearchStatusException) {
-                OpenSearchStatusException osException = (OpenSearchStatusException) e;
+            if (e instanceof OpenSearchException) {
+                OpenSearchException osException = (OpenSearchException) e;
                 if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
                     throw osException;
                 }
@@ -488,8 +489,8 @@ public class MemoryProcessingService {
                         listener.onResponse(summary);
                     } catch (Exception e) {
                         // Preserve client errors (4XX) with their detailed messages
-                        if (e instanceof OpenSearchStatusException) {
-                            OpenSearchStatusException osException = (OpenSearchStatusException) e;
+                        if (e instanceof OpenSearchException) {
+                            OpenSearchException osException = (OpenSearchException) e;
                             if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
                                 listener.onFailure(e);
                                 return;
--- a/plugin/src/main/java/org/opensearch/ml/action/session/TransportCreateSessionAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/session/TransportCreateSessionAction.java
@@ -11,6 +11,7 @@ import java.io.IOException;
 import java.time.Instant;
 
 import org.apache.commons.lang3.StringUtils;
+import org.opensearch.OpenSearchException;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.DocWriteRequest;
 import org.opensearch.action.index.IndexRequest;
@@ -98,6 +99,10 @@ public class TransportCreateSessionAction extends HandledTransportAction<MLCreat
                     );
                 return;
             }
+            if (container.getConfiguration().isDisableSession()) {
+                actionListener.onFailure(new OpenSearchStatusException("Session is disabled for this container", RestStatus.BAD_REQUEST));
+                return;
+            }
             createNewSession(input, container, user, tenantId, actionListener);
         }, actionListener::onFailure));
     }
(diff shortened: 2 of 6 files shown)


dhrubo-os:
I didn't clearly understand how is that really helping here?

b4sjoo:
> I didn't clearly understand how is that really helping here?

Check OpenSearch Repo, you will find `ResourceNotFoundException` is `OpenSearchException` instead of `OpenSearchStatusException`, that's why previously in many cases it is not caught correctly and converted into a 500 `InternalServerError`
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/ResourceNotFoundException.java

b4sjoo:
In the meantime in my opinion in the future all `MLResourceNotFoundException` should be converted to OpenSearch `ResourceNotFoundException` because current `MLResourceNotFoundException` is not set a RestStatus code with `404`

b4sjoo:
> OpenSearchException is the base class for many exceptions including server-side
> errors. Some subclasses of OpenSearchException may not have a meaningful HTTP status
> and could return unexpected values. Additionally, OpenSearchStatusException is
> already a subclass of OpenSearchException, so the change is correct in scope, but
> you should verify that all OpenSearchException subclasses reliably implement
> status() to avoid NullPointerException when calling
> osException.status().getStatus().

Checked [OpenSearchException](https://github.com/opensearch-project/OpenSearch/blob/f46ce23c24b53fd98dcfd9c4fce0a41cb565cda3/libs/core/src/main/java/org/opensearch/OpenSearchException.java#L304), by default it will be 500, so it's not nullable.

> OpenSearchException.status() can potentially return null for certain subclasses,
> which would cause a NullPointerException when calling status.getStatus(). A null
> check should be added before dereferencing the status object.

Same.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -186,8 +187,8 @@ public void extractFactsFromConversation(
                 listener.onResponse(facts);
             } catch (Exception e) {
                 // Preserve client errors (4XX) with their detailed messages
-                if (e instanceof OpenSearchStatusException) {
-                    OpenSearchStatusException osException = (OpenSearchStatusException) e;
+                if (e instanceof OpenSearchException) {
```
ylwu-amzn, line 190: This same 4XX-preservation pattern now appears **7+ times** across `MemoryProcessingService`, `TransportAddMemoriesAction`, and `TransportCreateSessionAction`:

```java
if (e instanceof OpenSearchException) {
    OpenSearchException osException = (OpenSearchException) e;
    if (osException.status().getStatus() >= 400 && osException.status().getStatus() < 500) {
        listener.onFailure(e);
        return;
    }
}
```

Consider extracting to a utility method (e.g., `isClientError(Exception e)`) to reduce duplication and ensure future error handlers stay consistent. `AgentUtils.extractStatusCode` already does something similar.

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -740,8 +740,8 @@ boolean shouldTrackRemoteFailure(Exception e) {
         }
 
         // Don't track any 4xx client errors (user/configuration issues)
-        if (e instanceof OpenSearchStatusException) {
-            RestStatus status = ((OpenSearchStatusException) e).status();
+        if (e instanceof OpenSearchException) {
```
ylwu-amzn, line 743: Note: `MLInferenceSearchResponseProcessor.java:232` has the same `instanceof OpenSearchStatusException` pattern for 4XX checks but is not updated in this PR. Should it be included for consistency, or is that intentionally out of scope?

plugin/src/main/java/org/opensearch/ml/action/session/TransportCreateSessionAction.java
```diff
@@ -98,6 +99,10 @@ protected void doExecute(Task task, MLCreateSessionRequest request, ActionListen
                     );
                 return;
             }
+            if (container.getConfiguration().isDisableSession()) {
```
ylwu-amzn, line 102: Clean fail-fast. Good that this returns before `createNewSession` to avoid unnecessary work.

plugin/src/test/java/org/opensearch/ml/action/session/TransportCreateSessionActionTests.java
```diff
@@ -349,6 +352,99 @@ public void testDoExecute_AccessDenied() {
...
+            return null;
+        }).when(memoryContainerHelper).indexData(any(), any(), any());
+
+        // Execute
+        transportCreateSessionAction.doExecute(task, request, actionListener);
+
+        // Verify - 4XX errors are preserved and passed through (not converted to INTERNAL_SERVER_ERROR)
+        ArgumentCaptor<Exception> exceptionCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(exceptionCaptor.capture());
+
```
ylwu-amzn, line 432: Good test — this specifically uses `IndexNotFoundException` which extends `OpenSearchException` but NOT `OpenSearchStatusException`, directly validating the fix. The assertion pattern with the type check and status extraction is thorough.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
```diff
@@ -435,8 +436,8 @@ private List<MemoryDecision> parseMemoryDecisions(String llmResultPath, MLTaskRe
             }
         } catch (Exception e) {
             // Preserve client errors (4XX) with their detailed messages
```
dhrubo-os, line 438: The current pattern preserves client errors but wraps all 5XX in a generic "Internal server error", which loses the original status code and message.

Could you please remind me again why we wanted this in the first place? Is this due to security concern?

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
dhrubo-os: +1 on this nit

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
b4sjoo: Yes, it is due to security concern.

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
b4sjoo: Intentionally out of scope because it is not wrapping 500 error message.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
dhrubo-os: What was the security concern?

b4sjoo: By security team they found the error stack trace for 5xx can reveal the internal logic

Issue #1380, opened 2023-09-25
[FEATURE]Improve test coverage for ConnectorUtils.java class


dhrubo-os:
We need to improve test coverage in our repo.

This issue will target only to improve test coverage for MLEngineClassLoader.java class

From this url we can see where test coverage is missing for this class: https://app.codecov.io/gh/opensearch-project/ml-commons/blob/2.x/ml-algorithms%2Fsrc%2Fmain%2Fjava%2Forg%2Fopensearch%2Fml%2Fengine%2Falgorithms%2Fremote%2FConnectorUtils.java

RishabBha0428:
Hi Dhrubo, I would like to work in this issue!

dhrubo-os:
Assigned to you.

dhrubo-os:
@RishabBha0428 any update on this?

cjrolo:
Yes, I can pick this one up! Assign to me.

dhrubo-os:
> Yes, I can pick this one up! Assign to me.

can you please comment on that issue?

PR #4734, opened 2026-03-23
Increase ConnectorUtils test coverage

cjrolo:
### Description
This is a focused test-improvement change set for issue #1380 that substantially raises ConnectorUtils coverage by replacing placeholder/smoke tests with behavior-driven assertions and adding missing branch-path tests across input preprocessing, output postprocessing, MLGuard validation (null/pass/fail paths), AWS request signing, SDK/streaming request construction, provider-specific createConnectorAction routing, and Bedrock Nova payload cleanup edge cases

Improvement:

INSTRUCTION:  96.10% (+14.64)
BRANCH: 92.57% (+16.22)
LINE: 96.97% (+14.39)
COMPLEXITY: 89.25% (+23.66)
METHOD: 94.44% (+11.11)
CLASS: 100.00% (no change)

I found that increasing it further would be going into a rabbit hole with additional complexity due to mock-heavy scripting/processor/connector requirements.

### Related Issues
Resolves #[1380]

### Check List
- [X] New functionality has been documented.


cjrolo:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Fix potentially incorrect Content-Length assertion
> **This test asserts that a connector-provided `Content-Length: 999` is preserved even though the actual payload is 16 bytes. If `buildSdkRequest` correctly computes the content length from the payload, this test would fail or assert incorrect behavior. Verify whether the production code actually preserves a pre-set `Content-Length` header, and if not, update the expected value to `"16"` to match the actual payload size.**
> 
> [ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtilsTest.java [1509-1519]](https://github.com/opensearch-project/ml-commons/pull/4734/files#diff-fb80ed1c16edb8a3c96a788fbede483222477394a58e21de3cf7d958c1494282R1509-R1519)
> 
> ```diff
> -public void buildSdkRequest_PreservesExistingContentLengthAndType() {
> +public void buildSdkRequest_PreservesExistingContentType() {
>      Connector connector = mock(Connector.class);
>      when(connector.getActionEndpoint(PREDICT.name(), Map.of())).thenReturn(OPENAI_EMBEDDINGS_URL);
>      when(connector.getParameters()).thenReturn(Map.of("model", "text-embedding-3-small"));
> -    when(connector.getDecryptedHeaders()).thenReturn(Map.of("Content-Type", "text/plain", "Content-Length", "999"));
> +    when(connector.getDecryptedHeaders()).thenReturn(Map.of("Content-Type", "text/plain"));
>  
>      SdkHttpFullRequest request = ConnectorUtils.buildSdkRequest(PREDICT.name(), connector, Map.of(), "{\"input\":\"test\"}", SdkHttpMethod.POST);
>  
>      assertEquals("text/plain", request.firstMatchingHeader("Content-Type").orElse(null));
> -    assertEquals("999", request.firstMatchingHeader("Content-Length").orElse(null));
> +    assertEquals("16", request.firstMatchingHeader("Content-Length").orElse(null));
>  }
> ```
> 
> Suggestion importance[1-10]: 6
> __
> 
> Why: This is a valid concern — if `buildSdkRequest` computes `Content-Length` from the actual payload rather than preserving a pre-set header value of `"999"`, the test would be asserting incorrect behavior. The suggestion correctly identifies this potential issue and proposes using the actual payload size `"16"` instead.
> 
> Low
> Fix mismatch between mock output and expected values
> **The test verifies that the result parameters contain escaped values, but the mock `TemplateScript` returns the raw (unescaped) values from `params` directly via `gson.toJson`. The actual `ConnectorUtils.processInput` may apply escaping before passing to the template, so the expected values should match what the mock actually returns — the raw processed text docs — not the escaped versions. Verify whether `escapeJson(processTextDoc(...))` matches what the mock script returns, or adjust the mock to reflect the real escaping behavior.**
> 
> [ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtilsTest.java [786-791]](https://github.com/opensearch-project/ml-commons/pull/4734/files#diff-fb80ed1c16edb8a3c96a788fbede483222477394a58e21de3cf7d958c1494282R786-R791)
> 
> ```diff
> -String expectedQuery = escapeJson(processTextDoc("query \"with quotes\""));
> -String expectedFirstDoc = escapeJson(processTextDoc("doc1 with \"quotes\""));
> -String expectedSecondDoc = escapeJson(processTextDoc("doc2 with \n newlines"));
> +String expectedQuery = processTextDoc("query \"with quotes\"");
> +String expectedFirstDoc = processTextDoc("doc1 with \"quotes\"");
> +String expectedSecondDoc = processTextDoc("doc2 with \n newlines");
>  assertEquals(expectedQuery, result.getParameters().get("query"));
>  assertEquals(expectedFirstDoc, result.getParameters().get("first"));
>  assertEquals(expectedSecondDoc, result.getParameters().get("second"));
> ```
> 
> Suggestion importance[1-10]: 5
> __
> 
> Why: The mock `TemplateScript` returns raw values from `params` directly, but the test expects `escapeJson(processTextDoc(...))` values. Whether this is correct depends on whether `ConnectorUtils.processInput` passes pre-escaped values to the template or raw values. This is a valid concern about test correctness, but without seeing the production code it's uncertain which side needs adjustment.
> 
> Low
> General	
> Add URI assertion to strengthen charset test
> **The character `é` in ISO-8859-1 encoding is 1 byte, so the assertion is correct for that charset. However, if the default charset or the system charset is UTF-8, `é` would be 2 bytes. The test should explicitly verify that the charset parameter is being respected by also asserting the URI or adding a more robust check, since the assertion of `"1"` is fragile and depends on the charset being correctly applied.**
> 
> [ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtilsTest.java [1121]](https://github.com/opensearch-project/ml-commons/pull/4734/files#diff-fb80ed1c16edb8a3c96a788fbede483222477394a58e21de3cf7d958c1494282R1121-R1121)
> 
> ```diff
> +assertNotNull(request);
> +assertEquals(OPENAI_EMBEDDINGS_URL, request.getUri().toString());
> +// "é" in ISO-8859-1 is 1 byte
>  assertEquals("1", request.firstMatchingHeader("Content-Length").orElse(null));
> ```
> 
> Suggestion importance[1-10]: 3
> __
> 
> Why: The suggestion to add a URI assertion is a minor improvement to test robustness. The `improved_code` adds an `assertNotNull` and URI check but doesn't change the core `Content-Length` assertion, making it a low-impact enhancement.
> 
> Low

Adopting recomended changes except the first one. Because current production logic intentionally preserves an existing Content-Length header if provided and changing expectation would test the wrong behavior for the current implementation.

PR #4766, opened 2026-04-01
Support tool description override in MCP connector

### LINKED ISSUE #4714: [FEATURE] Support tool description override in MCP connector

When using the standalone MCP connector (`mcp_sse` / `mcp_streamable_http`), tool descriptions are fetched directly from the remote MCP server with no way to customize them. This is frustrating because:

- Remote MCP server descriptions may be too generic, too verbose, or not well-suited for the agent's specific use case
- Users cannot tailor tool descriptions to improve agent tool selection accuracy without modifying the remote MCP server itself
- Different agents using the same MCP connector may need different descriptions for the same tools

Allow users to override tool descriptions in the `mcp_connectors` agent configuration. For example:

```json
{
  "mcp_connectors": [
    {
      "mcp_connector_id": "xxx",
      "tool_filters": ["search_*"],
      "tool_descriptions": {
        "search_index": "Search the product catalog index for relevant items based on user query"
      }
    }
  ]
}
```

Apply the description override in `AgentUtils.getMcpToolSpecs()` after tools are fetched from the remote MCP server. For each tool in the returned `MLToolSpec` list, if a matching entry exists in `tool_descriptions`, replace its description with the user-provided value.

Key files involved:
- `AgentUtils.java` — `getMcpToolSpecs()` method where tool specs are aggregated
- `McpConnectorExecutor.java` — `getMcpToolSpecs()` where tool metadata is extracted from remote server

- Modifying the remote MCP server's tool descriptions directly — not always feasible, especially for third-party MCP servers
- Using the built-in MCP server tools (`/_plugins/_ml/mcp/tools/_register`) which already support description override — but this requires a different architecture and doesn't work for standalone remote MCP servers

Currently the description flow is: **Remote MCP Server → `ListToolsResult` → `MLToolSpec.description`** with no interception point. The only processing applied is `StringUtils.processTextDoc()` for whitespace normalization.

Related existing capabilities:
- `tool_filters` already supports filtering which tools are exposed via regex patterns — `tool_descriptions` would be a natural companion
- The built-in MCP server tools already support description override at registration and update time

akolarkunnu:
### Description

Added support for MCP connector-level tool description overrides via tool_descriptions in mcp_connectors config.
Implemented override handling in AgentUtils.getMcpToolSpecs() for both:
 - connectors with tool_filters
 - connectors without tool_filters
Blank override values  and non-map tool_descriptions values are ignored safely.

Usage of tool_descriptions field:

```
"parameters": {
        "_llm_interface": "openai/v1/chat/completions",
        "mcp_connectors":[
            {
                "mcp_connector_id": "UFrJSZ0BiSRlpfWZM9Vs",
                "tool_filters": [
                       "^get_asset_platforms$"
                  ],
                 "tool_descriptions": [
                        { "Tool1": "Overriding Description1" },
                         { "another_tool": "Overriding Description2" }
                  ]
           }
       ]
      }
```

Examples:
Single tool_descriptions entry with filter:

```
curl -XPOST "http://localhost:9200/_plugins/_ml/agents/_register" -H 'Content-Type: application/json' -H 'x-tenant-id: 1234567' -d'
{
    "name": "No cache Claude",
    "type": "conversational",
    "description": "Use this for Agentic Search",
    "llm": {                                                                   
        "model_id": "TlrISZ0BiSRlpfWZntVd",
        "parameters": {
            "max_iteration": 15
        }              
    },                                                   
    "memory": {                                                                                                           
        "type": "conversation_index"                                                                    
    },                                           
    "parameters": {
        "_llm_interface": "openai/v1/chat/completions",
        "mcp_connectors":[
            {                             
                "mcp_connector_id": "UFrJSZ0BiSRlpfWZM9Vs",
                "tool_filters": [
          "^get_asset_platforms$"
        ],
                "tool_descriptions": [
        "get_asset_platforms": "Search product catalog with ranking hints"
      ]
            }
        ]
    },
    "tools": [
        {
            "type": "QueryPlanningTool"
        },
        {
            "type": "ListIndexTool"
        },
        {
            "type": "IndexMappingTool"
        }
    ],
    "app_type": "os_chat"
}

```
Multiple tool_descriptions entry without filter:

```
curl -XPOST "http://localhost:9200/_plugins/_ml/agents/_register" -H 'Content-Type: application/json' -H 'x-tenant-id: 1234567' -d'
{
    "name": "No cache Claude",
    "type": "conversational",
    "description": "Use this for Agentic Search",
    "llm": {                                                                   
        "model_id": "TlrISZ0BiSRlpfWZntVd",
        "parameters": {
            "max_iteration": 15
        }              
    },                     
    "memory": {
        "type": "conversation_index"
    },
    "parameters": {
        "_llm_interface": "openai/v1/chat/completions",
        "mcp_connectors":[
            {
                "mcp_connector_id": "UFrJSZ0BiSRlpfWZM9Vs",
                "tool_descriptions": [
                      {"get_asset_platforms": "Get the asset for different platforms"},
                      {"get_id_coins": "Get the different coins for different countries"},
                      {"get_list_coins_categories_categories": "Get the different coins for different categories"}
              ]
            }
        ]
    },
    "tools": [
        {
            "type": "QueryPlanningTool"
        },
        {
            "type": "ListIndexTool"
        },
        {
            "type": "IndexMappingTool"
        }
    ],
    "app_type": "os_chat"
}
'
```
Tested Scenarios
- Single tool_description with filter
- Single tool_description without filter
- Multiple tool_descriptions with filter
- Multiple tool_descriptions without filter
- Multiple tool_descriptions with non-existing tool description - it should ignore that entry
- Without any tool_descriptions entry

Tests Added
testGetMcpToolSpecs_ToolDescriptionOverrideAppliedWithoutFilter - Applies override when no filters are configured.
testGetMcpToolSpecs_ToolDescriptionOverrideAndWithFilter - Applies override to tools selected by tool_filters.
testGetMcpToolSpecs_OverrideMultipleToolDescriptions - verifies multiple (3) matching entries in tool_descriptions correctly override all corresponding MCP tool descriptions.
testGetMcpToolSpecs_DuplicateToolDescriptionKey_logsWarning_AndUsesLatestValue - Verifies duplicate tool_descriptions keys log a warning and the last value wins.
testGetMcpToolSpecs_IgnoreOverrideForNonExistingThirdTool - verifies extra override entries for non-existent tools are ignored, while existing tools are still overridden correctly.
testGetMcpToolSpecs_IgnoreOverrideForFilteredOutTool - Verifies an override for a tool that exists on the connector but is excluded by tool_filters logs the “filtered out” warning and is not applied.
testGetMcpToolSpecs_ToolDescriptionOverride_WithNull_Retain_OriginalDescription - Verifies a null override value does not change the tool’s original description.
testGetMcpToolSpecs_ToolDescriptionNonStringValue_logsWarning_AndIgnored - Verifies non-string override values log a warning and are ignored (tool keeps original description).
testGetMcpToolSpecs_ToolDescriptionOverrideBlankValueIgnored - Keeps original description when override value is blank.
testGetMcpToolSpecs_ToolDescriptionOverrideNonMapIgnored - Ignores invalid/non-map tool_descriptions config.
testGetMcpToolSpecs_NullToolSpecElement_WithAndWithoutFilter - Verifies null tool-spec entries don’t crash processing—preserved without filters, skipped when tool_filters are applied.

### Related Issues
Resolves #4714

### Check List


akolarkunnu:
@pyek-bot suggested to use array of objects to configure descriptions.

Eg:
```
"tool_descriptions": [
  { "get_asset_platforms": "Get the asset for different platforms" },
  { "another_tool": "Description for another tool" }
]
```

nathaliellenaa:
Thanks for raising this PR @akolarkunnu! One high level question: what happened if user passes `null` as the description, for example 
```
"tool_descriptions": {
    "get_asset_platforms": null
}
```
I assume it will be ignored and preserve the original description?

akolarkunnu:
> > @pyek-bot suggested to use array of objects to configure descriptions.
> > Eg:
> > ```
> > "tool_descriptions": [
> >   { "get_asset_platforms": "Get the asset for different platforms" },
> >   { "another_tool": "Description for another tool" }
> > ]
> > ```
> 
> Has this been addressed in this PR?

Comments are not addressed. I will work on these comments from next onwards. Thanks for your review.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +854,28 @@ public static void getMcpToolSpecs(
...
 
+    private static Map<String, String> toStringMap(Object rawMap) {
+        if (!(rawMap instanceof Map<?, ?>)) {
+            return Collections.emptyMap();
+        }
+        return ((Map<?, ?>) rawMap)
+            .entrySet()
+            .stream()
+            .filter(entry -> entry.getKey() != null && entry.getValue() != null)
+            .collect(Collectors.toMap(entry -> entry.getKey().toString(), entry -> entry.getValue().toString()));
```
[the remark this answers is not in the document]
akolarkunnu: Added a warning in this scenario. Also added a new unit test case to test this scenario - testGetMcpToolSpecs_ToolDescriptionNonStringValue_logsWarning_AndSuccess()

akolarkunnu:
> Thanks for raising this PR @akolarkunnu! One high level question: what happened if user passes `null` as the description, for example
> 
> ```
> "tool_descriptions": {
>     "get_asset_platforms": null
> }
> ```
> 
> I assume it will be ignored and preserve the original description?

@nathaliellenaa Yes, that's correct. null will be ignored and preserve the original description. I added a new unit test case to test this scenario - testGetMcpToolSpecs_ToolDescriptionOverride_WithNull_Retain_OriginalDescription()

akolarkunnu:
> > @pyek-bot suggested to use array of objects to configure descriptions.
> > Eg:
> > ```
> > "tool_descriptions": [
> >   { "get_asset_platforms": "Get the asset for different platforms" },
> >   { "another_tool": "Description for another tool" }
> > ]
> > ```
> 
> Has this been addressed in this PR?

@pyek-bot Addressed this comment now.

akolarkunnu:
> Thanks for the PR @akolarkunnu !
> 
> 1. What happens if user provides an override for a tool that doesn't exist? It would be silently ignored?
> 2. Could you create a documentation issue to add these details?

1. Yes it will silently ignore and add a warning log now for all tools which are configured for description override, but not present in the corresponding MCP connector.
2. I create the doc task - https://github.com/opensearch-project/documentation-website/issues/12370

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -814,25 +816,29 @@ public static void getMcpToolSpecs(
...
+            Map<String, String> toolDescriptionOverrides = toStringMap(mcpConnectorConfig.get(TOOL_DESCRIPTIONS_FIELD));
 
             getMCPToolSpecsFromConnector(connectorId, tenantId, sdkClient, client, encryptor, ActionListener.wrap(mcpToolspecs -> {
                 List<MLToolSpec> filteredTools;
                 if (toolFilters == null || toolFilters.isEmpty()) {
-                    filteredTools = mcpToolspecs;
+                    filteredTools = mcpToolspecs
+                        .stream()
+                        .map(toolSpec -> applyToolDescriptionOverride(toolSpec, toolDescriptionOverrides))
+                        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
akolarkunnu: ok, filtered null elements.

akolarkunnu:
Gentle reminder for review!

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -814,25 +816,30 @@ public static void getMcpToolSpecs(
         for (Map<String, Object> mcpConnectorConfig : mcpConnectorConfigs) {
             String connectorId = (String) mcpConnectorConfig.get(MCP_CONNECTOR_ID_FIELD);
             List<String> toolFilters = (List<String>) mcpConnectorConfig.get(TOOL_FILTERS_FIELD);
+            Map<String, String> toolDescriptionOverrides = toStringMap(mcpConnectorConfig.get(TOOL_DESCRIPTIONS_FIELD));
```
ylwu-amzn, line 819: **Blocking — schema diverges from issue #4714.**

Issue #4714 specifies a flat `Map<String, String>`:

```json
"tool_descriptions": {
    "search_index": "Search the product catalog..."
}
```

but this PR implements `List<Map<String, String>>` (list of single-entry maps):

```json
"tool_descriptions": [
    {"search_index": "Search the product catalog..."}
]
```

Effects:
- A user reading the issue and trying the natural schema gets a **silent no-op** — `toStringMap` (line 893-895) returns empty for any non-`List<?>` input, with no warning logged.
- The PR description's example #2 is actually **invalid JSON** because of this confusion (`"tool_descriptions": [ "tool": "value" ]` — list elements can't have keys). Only example #3 is correctly formatted.
- The list-of-single-entry-maps form has no upside vs a flat map: there's no semantic ordering, duplicate keys silently overwrite (see H2 below), and clients have to write an awkward shape to do what `Map<String, String>` does naturally.

Suggested fix: accept the flat `Map<?, ?>` shape per the issue spec. If you also want backward-compat with anyone who already adopted the list shape, branch on the runtime type in `toStringMap` (`Map` → iterate directly, `List` → keep current behavior, anything else → log a `warn`). Simpler: just match the issue spec since nothing has shipped yet.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -814,25 +816,30 @@ public static void getMcpToolSpecs(
...
+                        .filter(Objects::nonNull)
+                        .map(toolSpec -> applyToolDescriptionOverride(toolSpec, toolDescriptionOverrides))
+                        .collect(Collectors.toList());
                 } else {
                     filteredTools = new ArrayList<>();
                     List<Pattern> compiledPatterns = toolFilters.stream().map(Pattern::compile).collect(Collectors.toList());
 
                     for (MLToolSpec toolSpec : mcpToolspecs) {
                         for (Pattern pattern : compiledPatterns) {
                             if (pattern.matcher(toolSpec.getName()).matches()) {
```
ylwu-amzn: **Blocking — NPE on null tool spec, only on the with-filter path.**

Compare to the no-filter branch at line 824-828, which guards with `.filter(Objects::nonNull)` before mapping. This branch iterates `mcpToolspecs` directly and calls `toolSpec.getName()` without a null check — if any element is null, it NPEs here.

Either the no-filter branch's null guard is unnecessary (and should be removed), or this branch needs the same protection. Today the code is asymmetric in a way that hides the NPE behind whether `tool_filters` is configured.

The PR description claims a `testGetMcpToolSpecs_NoFilterWithNullToolSpecElement` test exists, but I couldn't find it in `AgentUtilsTest.java` — and it would only cover the no-filter side anyway. Suggested fix: add `if (toolSpec == null) continue;` at the top of the outer for-loop, and add a behavioral test that exercises the with-filter path with a null entry.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +856,78 @@ public static void getMcpToolSpecs(
...
+            .filter(Objects::nonNull)
+            .collect(Collectors.toSet());
+        for (String overrideToolName : toolDescriptionOverrides.keySet()) {
+            if (!knownToolNames.contains(overrideToolName)) {
+                log
+                    .warn(
+                        "MCP connector [{}]: tool_descriptions override for [{}] does not match any tool from this connector; override is ignored",
+                        connectorId,
+                        overrideToolName
+                    );
```
ylwu-amzn: **High-priority — warning misleads when an override targets a *filtered-out* tool, not a missing one.**

The check is against `filteredTools` (post-filter), so if a user has:

```json
"tool_filters": ["^Filter.*"],
"tool_descriptions": [{"NonFilterTool": "x"}]
```

and `NonFilterTool` *exists* on the connector but is filtered out, the user gets the warning "tool_descriptions override for [NonFilterTool] does not match any tool from this connector; override is ignored." The tool exists — it was just filtered. The user will reasonably assume the tool name is wrong and waste time chasing the wrong fix.

Suggested fix: distinguish two cases by also passing the pre-filter `mcpToolspecs` into this helper:
- override key in connector tools but not in `filteredTools` → warn "override ignored: tool was filtered out by tool_filters"
- override key in neither → keep the current message

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +856,78 @@ public static void getMcpToolSpecs(
...
+                        String keyStr = key.toString();
+                        if (!(value instanceof String)) {
+                            log
+                                .warn(
+                                    "Tool description override for tool [{}] is non-string type [{}]; coercing with toString() may hide configuration errors",
+                                    keyStr,
+                                    value.getClass().getName()
+                                );
+                        }
+                        merged.put(keyStr, value.toString());
```
ylwu-amzn: **High-priority — duplicate override keys silently overwrite, no warning.**

With config `[{"Tool1": "first"}, {"Tool1": "second"}]`, the user gets `Tool1 -> "second"` with nothing logged. Easy to introduce by accident in a long config list and impossible to debug from runtime behavior alone.

Suggested fix: detect collision before `put`:
```java
String prev = merged.put(keyStr, value.toString());
if (prev != null && !prev.equals(value.toString())) {
    log.warn("Duplicate tool_descriptions entry for [{}]: previous value [{}] overridden by [{}]", keyStr, prev, value);
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +856,78 @@ public static void getMcpToolSpecs(
...
+                    if (key != null && value != null) {
+                        String keyStr = key.toString();
+                        if (!(value instanceof String)) {
+                            log
+                                .warn(
+                                    "Tool description override for tool [{}] is non-string type [{}]; coercing with toString() may hide configuration errors",
+                                    keyStr,
+                                    value.getClass().getName()
+                                );
+                        }
```
ylwu-amzn: Non-string values warn-and-coerce, which produces nonsense descriptions. A user supplying `{"ToolA": true}` ends up with `description = "true"` for ToolA — the test `testGetMcpToolSpecs_ToolDescriptionNonStringValue_logsWarning_AndSuccess` asserts this on purpose, but it's a surprising design choice: the LLM sees a tool described as the literal string `"true"`, which degrades tool selection quality.

Suggested: skip the entry entirely (don't `put`) when the value isn't a `String`. The warning is already there; just don't follow it with bad data. Tests would need a small update.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +856,78 @@ public static void getMcpToolSpecs(
...
+        }
+    }
+
+    private static Map<String, String> toStringMap(Object raw) {
+        if (!(raw instanceof List<?> list)) {
+            return Collections.emptyMap();
+        }
+        Map<String, String> merged = new HashMap<>();
+        for (Object item : list) {
+            if ((item instanceof Map<?, ?> map)) {
```
ylwu-amzn: Nit — the outer parens around `(item instanceof Map<?, ?> map)` are redundant. Compare line 893 above which has the cleaner `if (!(raw instanceof List<?> list))`.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -849,6 +856,78 @@ public static void getMcpToolSpecs(
         }
     }
 
+    /**
+     * Warns when {@code tool_descriptions} contains keys that do not match any tool returned for the connector.
+     * Skips all work when WARN is not enabled, including scanning {@code filteredTools}.
+     */
```
ylwu-amzn, line 876: Nit — "Skips all work when WARN is not enabled, including scanning `filteredTools`" reads like a hot-path optimization, but the work is trivial (collect names of typically <10 tools, check N override keys against a set). The early-exit guard on `log.isWarnEnabled()` is fine, but the framing overstates it. Drop the second sentence or remove the doc comment entirely — the body is self-explanatory.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: I added a comment in the issue https://github.com/opensearch-project/ml-commons/issues/4714 that we changed the configuration structure and we will be documenting this feature as part of https://github.com/opensearch-project/documentation-website/issues/12370

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Dropped the second sentence.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Removed extra paranthesis

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Fixed and changed the test case testGetMcpToolSpecs_ToolDescriptionNonStringValue_logsWarning_AndSuccess according to new behavior.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Fixed and added a test case testGetMcpToolSpecs_DuplicateToolDescriptionKey_logsWarning_AndUsesLatestValue for duplicate override keys.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Fixed and added a new test case testGetMcpToolSpecs_IgnoreOverrideForFilteredOutTool for this scenario.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Fixed it.
Added back the test case and renamed it to testGetMcpToolSpecs_NullToolSpecElement_withAndWithoutFilter. Also added the use case of with filter also.

ylwu-amzn (approved):

PR #4779, opened 2026-04-04
Fail fast when tool is missing input_schema in function calling

ohltyler:
### Description

When a conversational agent builds the function calling payload (e.g., for Bedrock Converse or OpenAI), each tool must have an `input_schema` attribute so the LLM knows what parameters to pass. If `input_schema` is missing, the `${tool.attributes.input_schema}` placeholder is left unresolved, producing invalid JSON that the LLM provider rejects with an opaque `Invalid payload` error.

### Fix

1. **`AgentUtils.addToolsToFunctionCalling()`** — Throw a clear `IllegalArgumentException` when a tool is missing `input_schema`, naming the tool and the missing attribute. This fails fast with an actionable error message instead of silent downstream failures.
2. **`MLChatAgentRunner.runReAct()`** — Wrap `constructLLMPrompt` in try/catch so the synchronous `IllegalArgumentException` is routed to `listener.onFailure(e)`. Without this, the V1 chat agent path would throw the exception but the HTTP request would hang indefinitely (the exception was not propagated to the action listener). The V2 runner (`MLChatAgentRunnerV2.executeLLMCall`) already has the equivalent try/catch and propagates correctly.

Tools should define `input_schema` either:
1. In the tool implementation (as built-in ml-commons tools and [skills plugin tools](https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java) already do)
2. At agent registration time via the tool's attributes override

### Testing

**Unit tests**
- `AgentUtilsTest`: two cases covering null attributes and non-null attributes missing `input_schema` — both verify the `IllegalArgumentException` is thrown with the correct tool name and attribute name. Existing `testAddToolsToFunctionCalling` updated to include `input_schema` in mock tool attributes.
- `MLChatAgentRunnerTest.testRunWithToolMissingInputSchemaInFunctionCalling_PropagatesFailureToListener`: V1 chat runner — tool missing `input_schema` + `TOOL_TEMPLATE` set → `agentActionListener.onFailure` is invoked with the expected `IllegalArgumentException`.
- `MLChatAgentRunnerV2Test.testExecuteAgentLogic_ToolMissingInputSchema_PropagatesExceptionToListener`: V2 chat runner — same contract, also verifies the LLM is never invoked.

**End-to-end live repro** (Bedrock Claude Sonnet 4, `type: "conversational"`, one `AgentTool` registered without `input_schema` in attributes):

```
curl -X POST http://localhost:9200/_plugins/_ml/agents/<id>/_execute \
     -H 'Content-Type: application/json' \
     -d '{"parameters":{"question":"hello"}}'
```

Response (HTTP 400 in < 1 second):
```json
{
  "status": 400,
  "error": {
    "type": "IllegalArgumentException",
    "reason": "Invalid Request",
    "details": "Tool [MyAgentTool] is missing [input_schema] in its attributes. All tools used with function calling must define an input_schema."
  }
}
```

Without the `runReAct` try/catch added in this PR, the same request hangs indefinitely even though the exception is thrown — because the V1 chat runner did not propagate synchronous exceptions from `constructLLMPrompt` to its action listener.

### Issues Resolved

N/A - discovered during investigation of VectorDBTool support for agentic search

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -240,6 +240,9 @@ public class AgentUtils {
             toolParams.put(NAME, StringEscapeUtils.escapeJson(tool.getName()));
             toolParams.put(DESCRIPTION, StringEscapeUtils.escapeJson(tool.getDescription()));
             Map<String, ?> attributes = tool.getAttributes();
+            if (attributes == null || !attributes.containsKey("input_schema")) {
+                toolParams.putIfAbsent("attributes.input_schema", "{\"type\":\"object\",\"properties\":{}}");
+            }
             if (attributes != null) {
                 for (String key : attributes.keySet()) {
                     toolParams.put("attributes." + key, attributes.get(key));
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -154,7 +154,6 @@ public class AgentUtils {
     private static final String NAME = "name";
     private static final String DESCRIPTION = "description";
     private static final String INPUT_SCHEMA = "input_schema";
-    private static final String DEFAULT_INPUT_SCHEMA = "{\"type\":\"object\",\"properties\":{}}";
     private static final Pattern ADDITIONAL_PROPERTIES_PATTERN = Pattern
         .compile(",\\s*\"additionalProperties\"\\s*:\\s*(?:false|true)", Pattern.CASE_INSENSITIVE);
     public static final String AGENT_LLM_MODEL_ID = "agent_llm_model_id";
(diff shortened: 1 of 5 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -240,6 +240,9 @@ public static String addToolsToFunctionCalling(
             toolParams.put(NAME, StringEscapeUtils.escapeJson(tool.getName()));
             toolParams.put(DESCRIPTION, StringEscapeUtils.escapeJson(tool.getDescription()));
             Map<String, ?> attributes = tool.getAttributes();
+            if (attributes == null || !attributes.containsKey("input_schema")) {
+                toolParams.put("attributes.input_schema", "{\"type\":\"object\",\"properties\":{}}");
```
owaiskazi19: Can we create constants for both key and value

ohltyler: Done — extracted both into `INPUT_SCHEMA` and `DEFAULT_INPUT_SCHEMA` constants, and replaced all 5 occurrences of the `"input_schema"` string literal in the file.

rithinpullela (dismissed):

ohltyler:
> This can lead to silent failures where the agent attempts to trigger the tool using the "default" input schema but might fail as the tool expects a different input schema. This failure would never be surfaced to the user.
> 
> I agree that some validation could be added and error messages cleaned up.
> 
> With that said, why not just set the input_schema during agent register? You can override the input_schema today, example: https://github.com/Jon-AtAWS/opensearch-ml-quickstart/blob/5bf3bdbd0fcb71842be7a3f91c7e72894fd626bb/examples/agent_tools.py#L49

@pyek-bot thought about this more, I agree that the silent failure is also not ideal. I've updated the PR to just throw an exception if the input schema is missing. I separately added default input schemas for the tools that we were using when testing agentic search with the vectordb tool here: https://github.com/opensearch-project/skills/pull/722

Setting input schema during register is always an option, but not a good user experience and very error prone.

ohltyler:
@pyek-bot I think the cleaner soln is to just ensure there are defaults and not have ML Commons take on the responsibility of that, so I'll close this PR. I'll comment on your issue and do an audit on other tools that may need to be covered with valid defaults.

Issue #4704, opened 2026-03-09
[FEATURE] Support Optional Un-encrypted Credentials in ML-Commons Connectors


zane-neo:
# Support Optional Un-encrypted Credentials in ML-Commons Connectors

## Summary
Currently, ml-commons requires all connector credentials to be encrypted using `EncryptorImpl` with AWS Encryption SDK. This creates challenges in certain use cases where encryption is unnecessary, redundant, or problematic. This issue proposes adding support for optionally storing credentials in plaintext, controlled by a cluster-level setting (`plugins.ml_commons.allow_plaintext_credentials`, default: `false`) to ensure security by default.

### Architecture Overview
The encryption system in ml-commons currently works as follows:

1. **EncryptorImpl** (`ml-algorithms/src/main/java/org/opensearch/ml/engine/encryptor/EncryptorImpl.java`)
   - Implements the `Encryptor` interface
   - Uses AWS Encryption SDK with AES/GCM/NOPADDING
   - Generates and manages master keys per tenant
   - Master keys are cached with configurable TTL (default: from cluster settings)
   - Auto-initializes master keys in `.ml-config` index if not present

2. **Connector Classes** (AbstractConnector, HttpConnector, AwsConnector, etc.)
   - Store credentials in `Map<String, String> credential` field
   - Implement async `encrypt()` and `decrypt()` methods with `ActionListener`
   - Use `TriConsumer` to call `EncryptorImpl.encrypt()` when credentials are stored/updated
   - Use `TriConsumer` to call `EncryptorImpl.decrypt()` when credentials are needed for requests
   - Decrypted credentials stored temporarily in `decryptedCredential` field
   - Encryption/decryption processes full credential list asynchronously

3. **Usage Flow (Asynchronous)**
   ```
   Create/Update Connector → encrypt(List<credentials>, tenantId, ActionListener) → Store in index
   Use Connector → decrypt(List<credentials>, tenantId, ActionListener) → Execute request → Clear decryptedCredential
   ```

### Code References
- **Encryption/Decryption**: `EncryptorImpl.java:156-201` (async with ActionListener)
- **Master Key Management**: `EncryptorImpl.java:218-248` (getOrInitMasterKey method)
- **Connector Encryption**: `AbstractConnector.java:166-225` (encrypt/decrypt with TriConsumer)
- **Encryptor Interface**: `Encryptor.java:12-47` (async API)

## Proposed Solution

Add support for **optional plaintext credentials** with:
[...]

PR #4782, opened 2026-04-07
support plain text credentials

zane-neo:
### Description
Currently, ml-commons requires all connector credentials to be encrypted using EncryptorImpl with AWS Encryption SDK. This creates challenges in certain use cases where encryption is unnecessary, redundant, or problematic. This issue proposes adding support for optionally storing credentials in plaintext, controlled by a cluster-level setting (plugins.ml_commons.allow_plaintext_credentials, default: false) to ensure security by default.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4704

### Testing
Below is an example request body for different testing cases:
```
{
    "name": "Cohere Connector: embedding",
    "description": "The connector to cohere embedding model",
    "version": 1,
    "protocol": "http",
    "parameters": {
      "model_name": "embed-english-v3.0"
    },
    "credential": {
      "encrypted": "xxx",
      "cohere_key": "YOUR_COHERE_KEY"
    },
    "actions": [
      {
        "action_type": "predict",
        "method": "POST",
        "url": "https://api.cohere.com/v2/embed",
        "headers": {
          "Authorization": "Bearer ${credential.cohere_key}"
        },
        "request_body": "{ \"texts\": ${parameters.texts}, \"truncate\": \"END\", \"model\": \"${parameters.model_name}\", \"embedding_types\": [\"float\", \"int8\", \"uint8\", \"binary\", \"ubinary\"], \"input_type\": \"classification\"}",
        "pre_process_function": "connector.pre_process.cohere.embedding",
        "post_process_function": "connector.post_process.cohere_v2.embedding.float"
      }
    ]
}
```
#### Allow plain text credentials not enabled
Keep the cluster setting: `plugins.ml_commons.allow_plaintext_credentials` unchanged and try to create a connector with `encrypted` to false like below:
```
...
  "credential": {
      "encrypted": "false",
      "cohere_key": "YOUR_COHERE_KEY"
    },
...
```
The result should like below:
```
{
    "error": {
        "root_cause": [
            {
                "reason": "Plaintext credentials are not allowed. Please set cluster setting 'plugins.ml_commons.allow_plaintext_credentials' to true to enable plaintext credentials storage."
            }
        ],
        "reason": "Plaintext credentials are not allowed. Please set cluster setting 'plugins.ml_commons.allow_plaintext_credentials' to true to enable plaintext credentials storage."
    },
    "status": 400
}
```
Remove the key `encrypted` or change the key's value to true can get successful response.

#### Allow plain text credentials enabled
Use plain text credentials with request body like below:
```
...
  "credential": {
      "encrypted": "false",
      "cohere_key": "YOUR_COHERE_KEY"
    },
...
```
Response like below:
```
{
    "connector_id": "6VzuZp0B-y9qd6CWXP-L"
}
```
If you have permission to check the raw documents you will see:
```
...
"_source": {
    "name": "Cohere Connector: embedding",
     ...
    "credential": {
        "encrypted": "false",
        "cohere_key": "YOUR_COHERE_KEY"
    },
...
```

Use encrypted credentials with request body like below:
```
...
  "credential": {
      "cohere_key": "YOUR_COHERE_KEY"
    },
...
```
Response like below:
```
{
    "connector_id": "6lzvZp0B-y9qd6CWZv95"
}
```
If you have permission to read raw documents you will see:
```
...
"_source": {
  "name": "Cohere Connector: embedding",
  ...
  "credential": {
      "cohere_key": "AgV4djTpfnlByArLgWmAo7Y30CBFX9k/jB5ZirpeeccveYcAXwABABVhd3MtY3J5cHRvLXB1YmxpYy1rZXkAREExZlFTRFNtbHlTVnNCSEpWdHA0TDBZV0FQblp1eWJTS0NGL3FVVHhPdlk5VEp1ODhyejVjL2djS0VsUGJsSWVFQT09AAEABkN1c3RvbQAUAAAAgAAAAAxeqnZKVagWpsqAHm4AMASNOiiwECY1Rfvj2gHrqHKZ8v3y/k6kYl36YlISczDq8//Ry4wrbSroT0RAG/y6CwIAABAAL8BD5fT5HlQkSCvFtJKbgb8zZKn4cLR+TWhFQbg3WJT8m5TmM6Sx++aM5+GjRXjF/////wAAAAEAAAAAAAAAAAAAAAEAAAAoVbbRKUon+0+E7KxbnkLF1le/N2x+EKhcym8gAvHfsSWWUqjU097BWzDMYe9kHobdrDv5mXEcfJwAZzBlAjEAisLXWdOCzHbQra0X0u91DZGB4V3iWySEvHMxmvZ4QRZWc/tZkuR5xWBKAwMj01e8AjB33WjvCQ9kwEu7lUf/k8faOx26D4dTYRjA4baTbn+Czl5Z5MXPUv11hDFzsXZVlHM="
  },
...
```
#### Allow plain text credentials changed to disabled
If allow plain text credentials are set to enabled and registered with several connectors with plain text credentials and   the setting are changed back to disabled after that, then the plain text credentials should still work.
*Predict model*
```
POST /_plugins/_ml/models/BmuyZp0BfDnPLd6M65JO/_predict
{
    "parameters": {
        "texts": ["Say this is a test", "world"]
    }
}
```
Response:
```
{
    "inference_results": [
        {
            "output": [
                {
                    "name": "sentence_embedding",
                    "data_type": "FLOAT32",
                    "shape": [
                        1024
                    ],
                    "data": [
                        -0.015640259,
                        -0.0048446655,
                        -0.02684021,
                        -0.020339966,
                        -0.014762878,
                        0.004890442,
                        -0.003250122,
                        ...
                 ]
             }
          ]
      }
   ]
}
```
#### Update connector when allow plain text credentials are set to false
*Update Failed*
```
POST /_plugins/_ml/connectors/AGuyZp0BfDnPLd6MaJLr
{
    "credential": {
        "encrypted": "false",
        "cohere_key": "YOUR_COHERE_KEY"
    },
    "description": "Change the connector credentials to use plain text credential"
}
```
Response:
```
{
    "error": {
        "root_cause": [
            {
                "reason": "Plaintext credentials are not allowed. Please set cluster setting 'plugins.ml_commons.allow_plaintext_credentials' to true to enable plaintext credentials storage."
            }
        ],
        "reason": "Plaintext credentials are not allowed. Please set cluster setting 'plugins.ml_commons.allow_plaintext_credentials' to true to enable plaintext credentials storage."
    },
    "status": 400
}
```
*Update succeed*
```
POST /_plugins/_ml/connectors/AGuyZp0BfDnPLd6MaJLr
{
    "credential": {
        "cohere_key": "YOUR_COHERE_KEY"
    },
    "description": "Change the connector credentials to use plain text credential"
}
```
Response:
```
{
    "_id": "AGuyZp0BfDnPLd6MaJLr",
    "_version": 2,
    "result": "updated",
    "forced_refresh": true,
    "_shards": {
        "total": 1,
        "successful": 1,
        "failed": 0
    },
    "_seq_no": 5,
    "_primary_term": 3
}
```

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
@@ -61,10 +61,6 @@ public abstract class AbstractConnector implements Connector {
     @Setter
     protected Map<String, String> decryptedCredential;
 
-    // Indicates whether credentials should be encrypted. Default is true.
-    // This flag is extracted from the credentials map and stored separately.
-    protected Boolean isEncrypted = true;
-
     protected List<ConnectorAction> actions;
 
     @Setter
@@ -167,6 +163,14 @@ public abstract class AbstractConnector implements Connector {
         return predictEndpoint;
     }
 
+    /**
+     * Check if credentials are configured for plaintext storage (not encrypted).
+     * Returns true if the credential map contains "encrypted" = "false".
+     */
+    public boolean isPlaintextCredentials() {
+        return credential != null && "false".equalsIgnoreCase(credential.get(ENCRYPTED_FIELD));
+    }
+
     @Override
     public void encrypt(
         TriConsumer<List<String>, String, ActionListener<List<String>>> function,
@@ -177,8 +181,8 @@ public abstract class AbstractConnector implements Connector {
             listener.onResponse(true);
             return;
         }
-        // Skip encryption if isEncrypted is false (plaintext mode)
-        if (Boolean.FALSE.equals(isEncrypted)) {
+        // Skip encryption if credentials are configured for plaintext storage
+        if (isPlaintextCredentials()) {
             log.warn("Connector credentials are configured for plaintext storage (not encrypted)");
             listener.onResponse(true);
             return;
@@ -186,6 +190,10 @@ public abstract class AbstractConnector implements Connector {
         List<String> orderedEncryptKeys = new ArrayList<>();
         List<String> orderedToEncrypt = new ArrayList<>();
         for (String key : credential.keySet()) {
+            // Skip the encrypted flag itself
+            if (ENCRYPTED_FIELD.equals(key)) {
+                continue;
+            }
             orderedEncryptKeys.add(key);
             orderedToEncrypt.add(credential.get(key));
         }
@@ -215,10 +223,12 @@ public abstract class AbstractConnector implements Connector {
             listener.onResponse(true);
             return;
         }
-        // Skip decryption if isEncrypted is false (plaintext mode)
-        if (Boolean.FALSE.equals(isEncrypted)) {
+        // Skip decryption if credentials are configured for plaintext storage
+        if (isPlaintextCredentials()) {
             log.debug("Connector credentials are in plaintext mode, using as-is");
             decryptedCredential = new HashMap<>(credential);
+            // Remove the encrypted flag from decrypted credentials
+            decryptedCredential.remove(ENCRYPTED_FIELD);
             this.decryptedHeaders = createDecryptedHeaders(getAllHeaders(action));
             listener.onResponse(true);
             return;
@@ -226,6 +236,10 @@ public abstract class AbstractConnector implements Connector {
         List<String> orderedDecryptKeys = new ArrayList<>();
         List<String> orderedToDecrypt = new ArrayList<>();
         for (Map.Entry<String, String> entry : credential.entrySet()) {
+            // Skip the encrypted flag itself
+            if (ENCRYPTED_FIELD.equals(entry.getKey())) {
+                continue;
+            }
             orderedDecryptKeys.add(entry.getKey());
             orderedToDecrypt.add(entry.getValue());
         }
--- a/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
@@ -48,6 +48,7 @@ public abstract class AbstractConnector implements Connector {
     public static final String OWNER_FIELD = "owner";
     public static final String ACCESS_FIELD = "access";
     public static final String CLIENT_CONFIG_FIELD = "client_config";
+    public static final String ENCRYPTED_FIELD = "encrypted";
 
     protected String name;
     protected String description;
@@ -59,6 +60,10 @@ public abstract class AbstractConnector implements Connector {
     protected Map<String, String> decryptedHeaders;
     @Setter
     protected Map<String, String> decryptedCredential;
+
+    // Indicates whether credentials should be encrypted. Default is true.
+    // This flag is extracted from the credentials map and stored separately.
+    protected Boolean isEncrypted = true;
 
     protected List<ConnectorAction> actions;
 
@@ -172,6 +177,12 @@ public abstract class AbstractConnector implements Connector {
             listener.onResponse(true);
             return;
         }
+        // Skip encryption if isEncrypted is false (plaintext mode)
+        if (Boolean.FALSE.equals(isEncrypted)) {
+            log.warn("Connector credentials are configured for plaintext storage (not encrypted)");
+            listener.onResponse(true);
+            return;
+        }
         List<String> orderedEncryptKeys = new ArrayList<>();
         List<String> orderedToEncrypt = new ArrayList<>();
         for (String key : credential.keySet()) {
@@ -200,6 +211,14 @@ public abstract class AbstractConnector implements Connector {
         ActionListener<Boolean> listener
     ) {
         if (credential == null || credential.isEmpty()) {
+            this.decryptedHeaders = createDecryptedHeaders(getAllHeaders(action));
+            listener.onResponse(true);
+            return;
+        }
+        // Skip decryption if isEncrypted is false (plaintext mode)
+        if (Boolean.FALSE.equals(isEncrypted)) {
+            log.debug("Connector credentials are in plaintext mode, using as-is");
+            decryptedCredential = new HashMap<>(credential);
             this.decryptedHeaders = createDecryptedHeaders(getAllHeaders(action));
             listener.onResponse(true);
             return;
--- a/common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
+++ b/common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
@@ -539,4 +539,17 @@ public final class MLCommonsSettings {
         .boolSetting(ML_PLUGIN_SETTING_PREFIX + "ag_ui_enabled", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
     public static final String ML_COMMONS_AG_UI_DISABLED_MESSAGE =
         "The AG-UI agent feature is not enabled. To enable, please update the setting " + ML_COMMONS_AG_UI_ENABLED.getKey();
+
+    /**
+     * Controls whether plaintext (unencrypted) credentials are allowed in connectors.
+     *
+     * When false (default for open source): All credentials are always encrypted regardless of connector configuration.
+     * When true (default for managed service): Connectors can specify encryption preference via the "encrypted" flag in credentials.
+     *
+     * Default: false (ensures security by default in open source)
+     *
+     * This is a dynamic setting that can be changed at runtime via cluster settings API.
+     */
+    public static final Setting<Boolean> ML_COMMONS_ALLOW_PLAINTEXT_CREDENTIALS = Setting
+        .boolSetting(ML_PLUGIN_SETTING_PREFIX + "allow_plaintext_credentials", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
 }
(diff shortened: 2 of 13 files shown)


plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -121,6 +122,32 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
                         boolean hasPermission = connectorAccessControlHelper.validateConnectorAccess(client, connector);
                         if (hasPermission) {
                             connector.update(mlUpdateConnectorAction.getUpdateContent());
+
+                            // Enforce cluster-level plaintext credentials policy
+                            if (connector instanceof AbstractConnector) {
+                                AbstractConnector abstractConnector = (AbstractConnector) connector;
+                                if (abstractConnector.isPlaintextCredentials()) {
```
ylwu-amzn, line 129: Likely bug: this check runs after `connector.update(...)`, but `HttpConnector.update()` only overwrites `this.credential` when `updateContent.getCredential()` is non-null and non-empty. So for an existing connector that was created with `encrypted:"false"` while the setting was enabled, a later partial update that doesn't touch credentials (e.g., description-only) will still have `credential.get("encrypted") == "false"` — and if an operator has since flipped `allow_plaintext_credentials` back to false, this code will reject that description-only update.

Suggested fix: only enforce the policy when the *incoming* update is actually asking for plaintext storage. Roughly:

```java
Map<String, String> incomingCred = mlUpdateConnectorAction.getUpdateContent().getCredential();
boolean incomingIsPlaintext = incomingCred != null && "false".equalsIgnoreCase(incomingCred.get(AbstractConnector.ENCRYPTED_FIELD));
if (incomingIsPlaintext && !mlFeatureEnabledSetting.isPlaintextCredentialsAllowed()) {
    listener.onFailure(new IllegalArgumentException("Plaintext credentials are not allowed..."));
    return;
}
```

Please also add an IT case: create a plaintext connector with the setting enabled, toggle the setting off, then PUT a description-only update — should succeed. The current `testDynamicSettingChange` only exercises create.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -204,9 +223,23 @@ public void decrypt(
             listener.onResponse(true);
             return;
         }
+        // Skip decryption if credentials are configured for plaintext storage
+        if (isPlaintextCredentials()) {
+            log.debug("Connector credentials are in plaintext mode, using as-is");
```
ylwu-amzn, line 228: encrypt method use `warn` (line 186), should we keep consistent?

Also, both the encrypt warn (line 186) and this message don't include the connector name / ID. That makes the logs much less useful for post-hoc analysis.

plugin/src/main/java/org/opensearch/ml/action/connector/TransportCreateConnectorAction.java
```diff
@@ -119,6 +120,31 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLCrea
...
+            // Enforce cluster-level plaintext credentials policy
+            if (connector instanceof AbstractConnector) {
+                AbstractConnector abstractConnector = (AbstractConnector) connector;
+                if (abstractConnector.isPlaintextCredentials()) {
+                    if (!mlFeatureEnabledSetting.isPlaintextCredentialsAllowed()) {
+                        // Cluster setting does not allow plaintext credentials
+                        listener
+                            .onFailure(
+                                new IllegalArgumentException(
+                                    "Plaintext credentials are not allowed. Please set cluster setting 'plugins.ml_commons.allow_plaintext_credentials' to true to enable plaintext credentials storage."
```
ylwu-amzn, line 132: The error message literally instructs the operator how to disable the security control (`set cluster setting '...' to true`). That's an arguable UX choice — it's friendly but also lowers the friction for an operator under pressure to "just make it work" during an incident.

Consider softening: something like `"Plaintext credentials are not enabled for this cluster. See the ml-commons documentation for the allow_plaintext_credentials setting and its security implications."` This keeps the link to the remediation path but forces a second step where the security implications are visible.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -162,6 +163,14 @@ public String getActionEndpoint(String action, Map<String, String> parameters) {
         return predictEndpoint;
     }
 
+    /**
+     * Check if credentials are configured for plaintext storage (not encrypted).
+     * Returns true if the credential map contains "encrypted" = "false".
+     */
+    public boolean isPlaintextCredentials() {
```
ylwu-amzn, line 170: Minor: `credential.get(ENCRYPTED_FIELD)` treats the value as a `String`, and the comparison is `"false".equalsIgnoreCase(...)`. What happens if a user sends JSON `"encrypted": false` (boolean, not string)?

Because the credential map is declared as `Map<String, String>`, Jackson should coerce or fail during parse — please verify which it is, and document in the javadoc. If it silently coerces boolean-false to the string `"false"`, great. If it fails parse with a confusing error, a clearer upstream validation message would save users debugging time.

Same question for any value other than the exact strings `"true"` / `"false"` (`"FALSE"` works via ignoreCase, but what about `"0"`, `"no"`, `""`? Right now anything that is not `"false"` (ci) falls back to encrypted mode, which is safe-by-default — worth stating that explicitly in the javadoc).

plugin/src/test/java/org/opensearch/ml/rest/RestMLPlaintextCredentialsIT.java
```diff
@@ -0,0 +1,296 @@
...
+            fail("Expected exception after disabling setting again");
+        } catch (ResponseException e) {
+            assertTrue(
+                "Error should mention plaintext credentials",
+                TestHelper.httpEntityToString(e.getResponse().getEntity()).contains("Plaintext credentials are not allowed")
+            );
+        }
+
+        // Clean up
+        deleteConnector(connectorId);
```
ylwu-amzn, line 244: Missing coverage for the scenarios the PR description claims to support:

1. After toggling the setting back to disabled, a previously-created plaintext connector should still be usable for predict (the PR description says this works). Add an IT case.
2. After toggling the setting back to disabled, a description-only update on an existing plaintext connector should succeed (see the bug I flagged on `UpdateConnectorTransportAction.java:129` — this will currently fail, which is a regression).
3. No coverage for MCP connectors (`McpConnector`, `McpStreamableHttpConnector`) which also extend `AbstractConnector` and inherit this behavior.
4. No coverage for a credential key literally named `encrypted` (edge case, but important given the field-name collision risk).

At least (1) and (2) should block merge since they back up core claims in the PR description.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -539,4 +539,17 @@ private MLCommonsSettings() {}
...
+     * Controls whether plaintext (unencrypted) credentials are allowed in connectors.
+     *
+     * When false (default for open source): All credentials are always encrypted regardless of connector configuration.
+     * When true (default for managed service): Connectors can specify encryption preference via the "encrypted" flag in credentials.
+     *
+     * Default: false (ensures security by default in open source)
+     *
+     * This is a dynamic setting that can be changed at runtime via cluster settings API.
+     */
+    public static final Setting<Boolean> ML_COMMONS_ALLOW_PLAINTEXT_CREDENTIALS = Setting
```
ylwu-amzn, line 553: Documentation responsibility: Please open the companion documentation PR and link it here

plugin/src/main/java/org/opensearch/ml/action/connector/TransportCreateConnectorAction.java
```diff
@@ -119,6 +120,31 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLCrea
...
 
+            // Enforce cluster-level plaintext credentials policy
+            if (connector instanceof AbstractConnector) {
+                AbstractConnector abstractConnector = (AbstractConnector) connector;
+                if (abstractConnector.isPlaintextCredentials()) {
+                    if (!mlFeatureEnabledSetting.isPlaintextCredentialsAllowed()) {
+                        // Cluster setting does not allow plaintext credentials
+                        listener
+                            .onFailure(
+                                new IllegalArgumentException(
```
ylwu-amzn, line 131: This will return 5xx internal server error ? This should not be server side error, it's user's configuration.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -162,6 +163,14 @@ public String getActionEndpoint(String action, Map<String, String> parameters) {
...
+        return credential != null && "false".equalsIgnoreCase(credential.get(ENCRYPTED_FIELD));
```
ylwu-amzn, line 171: This method uses the connector's encrypted flag only — not the dynamic setting. 
If an admin enables the setting, creates plaintext connectors, then disables the setting, those existing connectors continue to bypass encryption forever.

zane-neo:
Closing this PR, refer to issue for more details.

PR #4785, opened 2026-04-07
Use GetMappingsRequest instead of GetIndexRequest in QueryPlanningTool

owaiskazi19:
### Description

`QueryPlanningTool` uses `GetIndexRequest` to retrieve index mappings for query planning, but it only accesses `.mappings()` from the response. `GetIndexRequest` requires the `indices:admin/get` permission, which also exposes index settings, aliases, and defaults — none of which QPT uses.

This PR switches to `GetMappingsRequest` / `GetMappingsResponse`, which only requires `indices:admin/mappings/get`. This:

- Follows the principle of least privilege
- Aligns with how `knn_full_access` and `query_assistant_access` roles already handle mapping retrieval
- Enables the `ml_full_access` role to support Agentic Search (QueryPlanningTool) with the narrower `indices:admin/mappings/get` permission instead of the broader `indices:admin/get`

### Related Issue

Resolves #4775

### Changes

**`QueryPlanningTool.java`:**
- `GetIndexRequest` → `GetMappingsRequest`
- `GetIndexResponse` → `GetMappingsResponse`
- `client.admin().indices().getIndex()` → `client.admin().indices().getMappings()`
- Response handling unchanged — `.mappings()` returns the same `Map<String, MappingMetadata>`

**`QueryPlanningToolTests.java`:**
- All mocks updated to use `GetMappingsResponse`
- All `indicesAdminClient.getIndex()` → `indicesAdminClient.getMappings()`

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
  - N/A (internal API change, no user-facing doc change)
- [x] Commits are signed off as per the DCO using `--signoff`

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java
@@ -28,8 +28,8 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.text.StringSubstitutor;
 import org.opensearch.action.admin.cluster.storedscripts.GetStoredScriptRequest;
-import org.opensearch.action.admin.indices.get.GetIndexRequest;
-import org.opensearch.action.admin.indices.get.GetIndexResponse;
+import org.opensearch.action.admin.indices.mapping.get.GetMappingsRequest;
+import org.opensearch.action.admin.indices.mapping.get.GetMappingsResponse;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.IndicesOptions;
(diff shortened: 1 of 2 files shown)


rithinpullela (approved):
Thanks for the PR Owais!

mingshl:
the change make sense, only the mapping is used for query planning. The only thing that comes out of my mind is to double check if there is any conflicts with indexInsightTool? I think people could use indexInsightTool with queryPlanningTool. 

I think it should be fine, if LLM call indexInsightTool first, which generate index insight, then use queryPlanningTool to generate a suitable query.  This is my theory, but have you test this case?

owaiskazi19:
> I think it should be fine, if LLM call indexInsightTool first, which generate index insight, then use queryPlanningTool to generate a suitable query. This is my theory, but have you test this case?

Index Insight's [StatisticalDataTask](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/indexInsight/StatisticalDataTask.java#L122) already uses GetMappingsRequest independently. Both the tools are independent of each other.

mingshl (approved):

mingshl:
> > I think it should be fine, if LLM call indexInsightTool first, which generate index insight, then use queryPlanningTool to generate a suitable query. This is my theory, but have you test this case?
> 
> Index Insight's [StatisticalDataTask](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/indexInsight/StatisticalDataTask.java#L122) already uses GetMappingsRequest independently. Both the tools are independent of each other.

make sense. I took a peak at the IndexInsightTool and it seems there are similar operations, like get index mapping, search few sample documents, make a model inference should think about extracting the similar operations into a utility class when more and more repetitive codes are merged.

PR #4658, opened 2026-02-22
[FEATURE] Semantic and Hybrid Search APIs for Long-Term Memory Retrieval

dhrubo-os:
### Description

Add two new simplified search APIs for agentic memory long-term memory retrieval, closing the competitive gap with Amazon Bedrock AgentCore Memory, mem0, and LangChain.

**Problem:** Currently, searching long-term memories requires users to manually generate embeddings, construct complex k-NN queries, and manually combine keyword + vector searches. Every major competitor provides simple text-based search with automatic embedding.

**Solution:** Two new REST API endpoints that abstract away the complexity.

Resolves #4647

---

#### 1. Semantic Search

```
POST/GET /_plugins/_ml/memory_containers/{memory_container_id}/memories/long-term/_semantic_search
```

**Request body:**
```json
{
  "query": "programming languages for data science",
  "k": 10,
  "namespace": {"user_id": "alice"},
  "tags": {"topic": "programming"},
  "min_score": 0.6,
  "filter": {"range": {"created_time": {"gte": 1700000000000}}}
}
```

| Parameter | Type | Required | Default | Description |
|-----------|------|----------|---------|-------------|
| `query` | String | Yes | — | Plain text search query |
| `k` | Integer | No | 10 | Number of results to return |
| `namespace` | Object | No | — | Filter by namespace fields (e.g., `{"user_id": "alice"}`) |
| `tags` | Object | No | — | Filter by tag fields (e.g., `{"topic": "food"}`) |
| `min_score` | Float | No | — | Minimum relevance score threshold |
| `filter` | Object | No | — | Generic OpenSearch query DSL filter |

**How it works:**
- Constructs `neural` query (or `neural_sparse` for SPARSE_ENCODING) via `wrapperQuery()` — same proven pattern as existing memory consolidation in `MemorySearchQueryBuilder.buildFactSearchQuery()`
- Wraps in `bool` query with namespace/tag/owner/container `filter` clauses plus optional generic `filter`
- Executes via `memoryContainerHelper.searchData()`
- Excludes `memory_embedding` field from response

#### 2. Hybrid Search

```
POST/GET /_plugins/_ml/memory_containers/{memory_container_id}/memories/long-term/_hybrid_search
```

**Request body:**
```json
{
  "query": "programming languages for data science",
  "k": 10,
  "namespace": {"user_id": "alice"},
  "tags": {"topic": "programming"},
  "min_score": 0.3,
  "filter": {"range": {"created_time": {"gte": 1700000000000}}},
[...]

PR #4798, opened 2026-04-22
[FIX] Simplify USER_PREFERENCE prompt to produce plain sentences

### LINKED ISSUE #4780: silently discards strategy results when LLM returns malformed JSON
### Description

When `infer: true` is used with Agentic Memory and the configured LLM returns malformed JSON for a strategy's fact extraction, the result is silently discarded. No error is returned in the API response, no warning is logged, and the caller has no way to know that a strategy failed. The API returns 200 and other strategies (e.g., SEMANTIC, SUMMARY) succeed normally.

This was observed with the `USER_PREFERENCE` strategy, whose internal prompt requires a complex output format (`"Preference sentence. Context: <why/how>. Categories: cat1,cat2"`). Claude Haiku 3 v2 produces invalid JSON for this format, while Haiku 4.5 and Sonnet 4.6 handle it correctly. However, the core issue is not model-specific — any LLM that produces malformed JSON will trigger the same silent failure.

### Steps to Reproduce

1. Create a memory container with USER_PREFERENCE strategy and a smaller LLM (e.g., Claude Haiku 3 v2):

```json
POST /_plugins/_ml/memory_containers/_create
{
  "name": "test",
  "configuration": {
    "embedding_model_id": "<embedding-model>",
    "embedding_model_type": "TEXT_EMBEDDING",
    "embedding_dimension": 384,
    "llm_id": "<claude-haiku-3-v2-model>",
    "disable_session": true,
    "strategies": [
      { "type": "SEMANTIC", "namespace": ["customer_id"] },
      { "type": "USER_PREFERENCE", "namespace": ["customer_id"] },
      { "type": "SUMMARY", "namespace": ["customer_id"] }
    ]
  }
}
```

2. Add a memory with `infer: true` containing clear user preferences.

3. API returns 200. SEMANTIC and SUMMARY produce long-term memories. USER_PREFERENCE produces 0 memories with no error.

### Expected Behavior

When a strategy's LLM output fails to parse, the system should:
- Log a WARNING with the strategy type and the raw LLM output
- Ideally, include a `warnings` array in the API response indicating which strategies failed

### Actual Behavior

- API returns 200 with `working_memory_id` — appears fully successful
- SEMANTIC: ✅ memories created
- SUMMARY: ✅ memories created
- USER_PREFERENCE: ❌ 0 memories, no error, no warning

The only way to discover the failure is to search long-term memory and notice the missing strategy type, or to read DEBUG-level OpenSearch logs and find the malformed JSON in the LLM response.

### Root Cause (from DEBUG logs)

The LLM returns malformed JSON where the array bracket is misplaced:

```json
{"facts":["I prefer phone calls...Categories: communication,time_preference"],
"I am not a morning person...Categories: time_preference",
"Wife Sarah...Categories: relationships,decision_making"]}
```

The `Context:` / `Categories:` format within each fact string contains commas that confuse smaller models into breaking the JSON array structure.

### Test Results by Model

| Model | SEMANTIC | SUMMARY | USER_PREFERENCE |
|-------|----------|---------|-----------------|
| Claude Haiku 3 v2 | ✅ | ✅ | ❌ silent failure |
| Claude Haiku 4.5 | ✅ | ✅ | ✅ |
| Claude Sonnet 4.6 | ✅ | ✅ | ✅ |

### Suggested Fixes

1. **Log a WARNING on JSON parse failure** — Include strategy type and raw LLM output. This is the minimum fix.
2. **Return parse failures in the API response** — Add a `warnings` field so callers can detect partial failures.
3. **Consider simplifying the USER_PREFERENCE prompt** — The `Context:` / `Categories:` format is more complex than SEMANTIC's plain facts format, making it more prone to JSON errors with smaller models.

### Environment

- OpenSearch: 3.6.0 (Docker, `opensearchstaging/opensearch:3.6.0`)
- Failing LLM: Claude Haiku 3 v2 via Bedrock Converse API
- Working LLMs: Claude Haiku 4.5, Claude Sonnet 4.6 via Bedrock Converse API
- Embedding: Local pretrained all-MiniLM-L12-v2

dhrubo-os:
### Description

Simplify the USER_PREFERENCE extraction prompt to produce plain self-contained sentences instead of the structured `Context:/Categories:` format. Also improve error logging when JSON parsing fails to include the strategy type.

**Before:**
```
"Bob prefers phone calls over email. Context: wife Sarah is hard of hearing. Categories: communication,time_preference"
```

**After:**
```
"Bob always prefers phone calls over email because his wife Sarah is hard of hearing."
```

### Motivation

1. **Fixes silent JSON parse failures with smaller LLMs** — The `Context: <why/how>. Categories: cat1,cat2` format contains commas inside JSON string values that confuse smaller models (e.g., Claude Haiku 3) into producing malformed JSON. When this happens, USER_PREFERENCE results are silently discarded with no error returned to the caller. Plain sentences eliminate this failure mode.

2. **Improves search quality** — The `Context:` and `Categories:` boilerplate text was stored in the `memory` field alongside the actual preference, diluting embedding quality for semantic search and creating false keyword matches for BM25/hybrid search. Plain sentences produce cleaner embeddings.

3. **No downstream consumers** — The Context and Categories metadata is not consumed by any code path today — no filtering by category, no surfacing of context to users.

### Testing

Validated on LoCoMo benchmark (1,540 questions, 10 conversations) with Claude Sonnet 4.6. USER_PREFERENCE extraction count actually increased slightly (685 vs 678 baseline).

### Issues Resolved

Resolves #4780

### Check List
- [x] New functionality includes testing
- [x] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
@@ -197,9 +197,8 @@ public class MemoryContainerConstants {
             Respond NOW with ONE LINE of valid JSON ONLY exactly as {"facts":["fact1","fact2",...]}. No extra text, no code fences, no newlines or tabs, no spaces after commas or colons.""";
 
     // JSON enforcement message for user preference extraction
-    public static final String USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE =
-        """
-            Return ONLY ONE LINE of valid JSON exactly as {"facts":["<Preference sentence>. Context: <why/how>. Categories: <cat1,cat2>"]}. Begin with { and end with }. No extra text.""";
+    public static final String USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE = """
+        Return ONLY ONE LINE of valid JSON exactly as {"facts":["<fact sentence>"]}. Begin with { and end with }. No extra text.""";
 
     public static final String USER_PREFERENCE_FACTS_EXTRACTION_PROMPT =
         """
@@ -216,12 +215,14 @@ public class MemoryContainerConstants {
             </EXTRACT>
 
             <STYLE & RULES>
-            • One sentence per preference; merge related details; no duplicates; preserve user wording and numbers; avoid relative time; keep each fact < 350 chars.
-            • Format: "Preference sentence. Context: <why/how>. Categories: cat1,cat2"
+            • One self-contained sentence per preference. Include the subject's name and enough context to be meaningful on its own.
+            • Merge closely related details into the same sentence. No duplicates.
+            • Preserve user wording, names, numbers, and units. Avoid relative time references; use absolute dates when available.
+            • Keep each fact under 350 characters.
             </STYLE & RULES>
 
             <OUTPUT>
-            Return ONLY one minified JSON object exactly as {"facts":["Preference sentence. Context: <why/how>. Categories: cat1,cat2"]}. If none, return {"facts":[]}. The first character MUST be '{' and the last MUST be '}'. No preambles, explanations, code fences, XML, or other text.
+            Return ONLY one minified JSON object exactly as {"facts":["fact sentence"]}. If none, return {"facts":[]}. The first character MUST be '{' and the last MUST be '}'. No preambles, explanations, code fences, XML, or other text.
             </OUTPUT>""";
 
     public static final String SUMMARY_FACTS_EXTRACTION_PROMPT =
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingService.java
@@ -356,7 +356,7 @@ public class MemoryProcessingService {
                             }
                         }
                     } catch (IOException e) {
-                        log.error("Failed to extract content from dataMap", e);
+                        log.error("Failed to parse {} strategy LLM response as JSON", strategy.getType(), e);
                         throw new OpenSearchStatusException("Internal server error", RestStatus.INTERNAL_SERVER_ERROR);
                     }
                 }
--- a/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingServiceTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingServiceTests.java
@@ -1027,8 +1027,8 @@ public class MemoryProcessingServiceTests {
         assertTrue("Should be role-based", prompt.contains("USER PREFERENCE EXTRACTOR"));
 
         // Verify key requirements
-        assertTrue("Should have character limit", prompt.contains("< 350 chars"));
-        assertTrue("Should specify context format", prompt.contains("Context: <why/how>"));
+        assertTrue("Should have character limit", prompt.contains("350 characters"));
+        assertTrue("Should specify self-contained sentences", prompt.contains("self-contained sentence"));
 
         // Verify old problematic format is removed
         assertFalse("Should not use pipe delimiters", prompt.contains("preference | context:"));
@@ -1039,7 +1039,7 @@ public class MemoryProcessingServiceTests {
         // Test that enforcement message matches the new format
         String enforcement = USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE;
 
-        assertTrue("Should specify natural language format", enforcement.contains("Context: <why/how>. Categories:"));
+        assertTrue("Should specify fact sentence format", enforcement.contains("fact sentence"));
         assertFalse("Should not use old pipe format", enforcement.contains("preference | context:"));
     }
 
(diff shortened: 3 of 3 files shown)


common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -216,12 +215,14 @@ public class MemoryContainerConstants {
             </EXTRACT>
 
             <STYLE & RULES>
-            • One sentence per preference; merge related details; no duplicates; preserve user wording and numbers; avoid relative time; keep each fact < 350 chars.
-            • Format: "Preference sentence. Context: <why/how>. Categories: cat1,cat2"
+            • One self-contained sentence per preference. Include the subject's name and enough context to be meaningful on its own.
+            • Merge closely related details into the same sentence. No duplicates.
+            • Preserve user wording, names, numbers, and units. Include relative time references when no absolute date is available.
```
ylwu-amzn: Use relative time may cause confusion later. For example, if user say "2 weeks ago", after 1 week, the relative time becomes wrong.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -197,9 +197,8 @@ public class MemoryContainerConstants {
             Respond NOW with ONE LINE of valid JSON ONLY exactly as {"facts":["fact1","fact2",...]}. No extra text, no code fences, no newlines or tabs, no spaces after commas or colons.""";
 
     // JSON enforcement message for user preference extraction
-    public static final String USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE =
-        """
-            Return ONLY ONE LINE of valid JSON exactly as {"facts":["<Preference sentence>. Context: <why/how>. Categories: <cat1,cat2>"]}. Begin with { and end with }. No extra text.""";
+    public static final String USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE = """
+        Return ONLY ONE LINE of valid JSON exactly as {"facts":["<fact sentence>"]}. Begin with { and end with }. No extra text.""";
```
ylwu-amzn, line 201: Nit: the sibling constant `JSON_ENFORCEMENT_MESSAGE` directly above uses
```java
public static final String JSON_ENFORCEMENT_MESSAGE =
    """
        Respond NOW with ...""";
```
but this one puts `"""` on the same line as `=` with shallower content indentation. Matching the surrounding style would be a small readability win and also keeps the Java text-block incidental-whitespace stripping rules identical between the two constants.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -197,9 +197,8 @@ public class MemoryContainerConstants {
             Respond NOW with ONE LINE of valid JSON ONLY exactly as {"facts":["fact1","fact2",...]}. No extra text, no code fences, no newlines or tabs, no spaces after commas or colons.""";
 
     // JSON enforcement message for user preference extraction
-    public static final String USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE =
-        """
-            Return ONLY ONE LINE of valid JSON exactly as {"facts":["<Preference sentence>. Context: <why/how>. Categories: <cat1,cat2>"]}. Begin with { and end with }. No extra text.""";
```
ylwu-amzn, line 202: Why remove Context/Categories ? Will this break current code ?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
dhrubo-os, line 201: Fixed — now matches the sibling `JSON_ENFORCEMENT_MESSAGE` style.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
dhrubo-os, line 202: No code path consumes Context or Categories today — there's no filtering by category and no surfacing of context metadata to users or agents. The entire string (including the metadata) gets stored in the `memory` field and embedded as-is, which actually hurts search quality by diluting the embedding with boilerplate text.

The main motivation is #4780 — the `Context: <why/how>. Categories: cat1,cat2` format causes smaller LLMs (e.g., Claude Haiku) to produce malformed JSON because the commas inside `Categories: cat1,cat2` confuse the model into breaking the JSON array structure. With plain sentences, Haiku 4.5 successfully extracts all USER_PREFERENCE memories (verified in testing).

This won't break existing code — the `memory` field is a free-text string, and downstream code only reads it as-is. No parser expects the `Context:/Categories:` structure.

dhrubo-os (commented):
Thanks for the review Yaliang!

dhrubo-os:
Correction on the relative time comment: You're right @ylwu-amzn — this was indeed a change introduced in this PR, not from #4658. I've fixed it — the prompt now says "Avoid relative time references; use absolute dates when available", consistent with the original behavior. Thanks for catching this!

rithinpullela (approved):
LGTM!

One small suggestion for the future: Instead of making the LLMs return in a JSON format asking them to make a tool call results in less errors. The models are trained on tool calls and hence they are really good at using tools rather than responding in a particular format. I saw the same behavior with models like Opus 4.5

dhrubo-os:
> LGTM!
> 
> One small suggestion for the future: Instead of making the LLMs return in a JSON format asking them to make a tool call results in less errors. The models are trained on tool calls and hence they are really good at using tools rather than responding in a particular format. I saw the same behavior with models like Opus 4.5

Good Idea, created an issue as an placeholder for this: https://github.com/opensearch-project/ml-commons/issues/4799

ylwu-amzn (approved):

PR #4800, opened 2026-04-23
Simplify memory container creation with inline model config

dhrubo-os:
### Description

Reduces memory container creation from **5 API calls to 1** by supporting inline model specifications.

**Before:** 5 API calls (register embedding connector, register embedding model, register LLM connector, register LLM model, create container)

**After:** Single API call with inline `embedding_model` and `llm`:

#### What the user sends:
```json
POST /_plugins/_ml/memory_containers/_create
{
  "name": "my-memory",
  "configuration": {
    "embedding_model": {
      "model_id": "amazon.titan-embed-text-v2:0",
      "model_provider": "bedrock/embedding",
      "credential": { "access_key": "...", "secret_key": "...", "session_token": "..." }
    },
    "llm": {
      "model_id": "us.anthropic.claude-sonnet-4-6",
      "model_provider": "bedrock/converse",
      "credential": { "access_key": "...", "secret_key": "...", "session_token": "..." }
    },
    "strategies": [
      { "type": "SEMANTIC", "namespace": ["user_id"] },
      { "type": "USER_PREFERENCE", "namespace": ["user_id"] }
    ]
  }
}
```

#### What gets stored in the container index:
The inline specs are **never stored**. The system auto-creates the models, then stores the resulting IDs in the container document — identical to the existing format:
```json
{
  "name": "my-memory",
  "configuration": {
    "embedding_model_id": "auto_created_abc123",
    "embedding_model_type": "TEXT_EMBEDDING",
    "embedding_dimension": 1024,
    "llm_id": "auto_created_def456",
    "strategies": [
      { "type": "SEMANTIC", "namespace": ["user_id"], "id": "semantic_a1b2c3d4", "enabled": true },
      { "type": "USER_PREFERENCE", "namespace": ["user_id"], "id": "user_preference_e5f6g7h8", "enabled": true }
    ],
    "index_prefix": "...",
    "parameters": { "llm_result_path": "$.output.message.content[0].text" }
  },
  "owner": { ... },
  "created_time": "...",
  "last_updated_time": "..."
}
```

This means:
- Full backward compatibility — existing `embedding_model_id`/`llm_id` format still works
- Model IDs are visible in the container document for credential rotation via `PUT /_plugins/_ml/models/<id>`
- No new storage format, no migration needed
- Embedding type (dense/sparse) and dimension auto-detected from known model IDs
- `llm_result_path` auto-set per LLM provider (Bedrock, OpenAI, Gemini)

### Supported Providers

| Embedding | LLM |
|-----------|-----|
| `bedrock/embedding` (Titan, Cohere) | `bedrock/converse` |
| `openai/v1/embeddings` | `openai/v1/chat/completions` |
| | `gemini/v1beta/generatecontent` |

### E2E Verified

All combos tested against Docker cluster (OpenSearch 3.6.0):

| Embedding | LLM | Infer | Semantic Search |
|-----------|-----|-------|-----------------|
| Bedrock Titan v2 | Bedrock Claude | ✅ | ✅ |
| Bedrock Cohere v3 | Bedrock Claude | ✅ | ✅ |
| OpenAI text-embedding-3-small | Bedrock Claude | ✅ | ✅ |
| OpenAI text-embedding-3-small | OpenAI gpt-4o-mini | ✅ | ✅ |
| OpenAI text-embedding-3-small | Gemini 2.0 flash | ✅ | ✅ |

### Known Limitations (Phase 1)
- No rollback if LLM model creation fails after embedding model was created (same gap as agent revamp)
- Orphaned models not auto-cleaned on container deletion (same as agent revamp)

### Check List
- [x] Commits are signed per the DCO using --signoff.

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.agent;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.MLAgentType;
+import org.opensearch.ml.common.connector.AwsConnector;
+import org.opensearch.ml.common.connector.Connector;
+import org.opensearch.ml.common.connector.ConnectorAction;
+import org.opensearch.ml.common.connector.ConnectorClientConfig;
+import org.opensearch.ml.common.connector.ConnectorProtocols;
+import org.opensearch.ml.common.input.execute.agent.ContentBlock;
+import org.opensearch.ml.common.input.execute.agent.Message;
+import org.opensearch.ml.common.model.ModelProvider;
+import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
+
+/**
+ * Model provider for Bedrock Embedding API (Titan, Cohere).
+ * Uses the /invoke endpoint (not /converse) with inputText-based request body.
+ */
+public class BedrockEmbeddingModelProvider extends ModelProvider {
+
+    private static final String DEFAULT_REGION = "us-east-1";
+
+    private static final String REQUEST_BODY_TEMPLATE =
+        "{ \"inputText\": \"${parameters.inputText}\", \"dimensions\": ${parameters.dimensions},"
+            + " \"normalize\": ${parameters.normalize}, \"embeddingTypes\": ${parameters.embeddingTypes} }";
+
+    /**
+     * Known Bedrock embedding models with their type and default dimension.
+     */
+    public static final Map<String, EmbeddingModelInfo> KNOWN_MODELS = Map.of(
+        "amazon.titan-embed-text-v2:0", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+        "amazon.titan-embed-text-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536),
+        "amazon.titan-embed-image-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+        "cohere.embed-english-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+        "cohere.embed-multilingual-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)
+    );
+
+    @Override
+    public Connector createConnector(String modelId, Map<String, String> credential, Map<String, String> modelParameters) {
+        EmbeddingModelInfo info = KNOWN_MODELS.get(modelId);
+
+        Map<String, String> parameters = new HashMap<>();
+        parameters.put("region", DEFAULT_REGION);
+        parameters.put("service_name", "bedrock");
+        parameters.put("model", modelId);
+        parameters.put("dimensions", String.valueOf(info != null ? info.dimension : 1024));
+        parameters.put("normalize", "true");
+        parameters.put("embeddingTypes", "[\"float\"]");
+
+        if (modelParameters != null) {
+            parameters.putAll(modelParameters);
+        }
+
+        Map<String, String> headers = new HashMap<>();
+        headers.put("content-type", "application/json");
+        headers.put("x-amz-content-sha256", "required");
+
+        ConnectorAction predictAction = ConnectorAction
+            .builder()
+            .actionType(ConnectorAction.ActionType.PREDICT)
+            .method("POST")
+            .url("https://bedrock-runtime.${parameters.region}.amazonaws.com/model/${parameters.model}/invoke")
+            .headers(headers)
+            .requestBody(REQUEST_BODY_TEMPLATE)
+            .preProcessFunction("connector.pre_process.bedrock.embedding")
+            .postProcessFunction("connector.post_process.bedrock.embedding")
+            .build();
+
+        ConnectorClientConfig connectorClientConfig = new ConnectorClientConfig();
+        connectorClientConfig.setMaxRetryTimes(3);
+
+        return AwsConnector
+            .awsConnectorBuilder()
+            .name("Auto-generated Bedrock Embedding connector")
+            .description("Auto-generated connector for Bedrock Embedding API")
+            .version("1")
+            .protocol(ConnectorProtocols.AWS_SIGV4)
+            .parameters(parameters)
+            .credential(credential != null ? credential : new HashMap<>())
+            .actions(List.of(predictAction))
+            .connectorClientConfig(connectorClientConfig)
+            .build();
+    }
+
+    @Override
+    public MLRegisterModelInput createModelInput(String modelName, Connector connector, Map<String, String> modelParameters) {
+        EmbeddingModelInfo info = KNOWN_MODELS.get(modelName);
+        FunctionName functionName = info != null ? info.functionName : FunctionName.TEXT_EMBEDDING;
+
+        return MLRegisterModelInput
+            .builder()
+            .functionName(functionName)
+            .modelName("Auto-generated embedding model for " + modelName)
+            .description("Auto-generated embedding model for memory container")
+            .connector(connector)
+            .build();
+    }
+
+    @Override
+    public String getLLMInterface() {
+        return null; // Embedding models don't have an LLM interface
+    }
+
+    // Embedding models don't support agent input mapping — these are LLM-only operations
+    @Override
+    public Map<String, String> mapTextInput(String text, MLAgentType type) {
+        throw new UnsupportedOperationException("Embedding model providers do not support agent input mapping");
+    }
+
+    @Override
+    public Map<String, String> mapContentBlocks(List<ContentBlock> contentBlocks, MLAgentType type) {
+        throw new UnsupportedOperationException("Embedding model providers do not support agent input mapping");
+    }
+
+    @Override
+    public Map<String, String> mapMessages(List<Message> messages, MLAgentType type) {
+        throw new UnsupportedOperationException("Embedding model providers do not support agent input mapping");
+    }
+
+    @Override
+    public String extractMessageFromResponse(Map<String, ?> responseData) {
+        throw new UnsupportedOperationException("Embedding model providers do not support response extraction");
+    }
+
+    @Override
+    public Message parseToUnifiedMessage(String json) {
+        throw new UnsupportedOperationException("Embedding model providers do not support message parsing");
+    }
+
+    /**
+     * Look up embedding model info from known models registry.
+     * @param modelId the Bedrock model ID
+     * @return EmbeddingModelInfo if known, null otherwise
+     */
+    public static EmbeddingModelInfo getModelInfo(String modelId) {
+        return KNOWN_MODELS.get(modelId);
+    }
+
+    /**
+     * Info about a known embedding model.
+     */
+    public static class EmbeddingModelInfo {
+        public final FunctionName functionName;
+        public final int dimension;
+
+        public EmbeddingModelInfo(FunctionName functionName, int dimension) {
+            this.functionName = functionName;
+            this.dimension = dimension;
+        }
+    }
+}
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -29,9 +29,12 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
 
     private static final String DEFAULT_REGION = "us-east-1";
 
-    private static final String REQUEST_BODY_TEMPLATE =
+    private static final String TITAN_REQUEST_BODY =
         "{ \"inputText\": \"${parameters.inputText}\", \"dimensions\": ${parameters.dimensions},"
             + " \"normalize\": ${parameters.normalize}, \"embeddingTypes\": ${parameters.embeddingTypes} }";
+
+    private static final String COHERE_REQUEST_BODY =
+        "{ \"texts\": ${parameters.texts}, \"input_type\": \"${parameters.input_type}\" }";
 
     /**
      * Known Bedrock embedding models with their type and default dimension.
@@ -47,14 +50,30 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
     @Override
     public Connector createConnector(String modelId, Map<String, String> credential, Map<String, String> modelParameters) {
         EmbeddingModelInfo info = KNOWN_MODELS.get(modelId);
+        boolean isCohere = modelId != null && modelId.startsWith("cohere.");
 
         Map<String, String> parameters = new HashMap<>();
         parameters.put("region", DEFAULT_REGION);
         parameters.put("service_name", "bedrock");
         parameters.put("model", modelId);
-        parameters.put("dimensions", String.valueOf(info != null ? info.dimension : 1024));
-        parameters.put("normalize", "true");
-        parameters.put("embeddingTypes", "[\"float\"]");
+
+        String requestBody;
+        String preProcess;
+        String postProcess;
+
+        if (isCohere) {
+            parameters.put("input_type", "search_document");
+            requestBody = COHERE_REQUEST_BODY;
+            preProcess = "connector.pre_process.cohere.embedding";
+            postProcess = "connector.post_process.cohere.embedding";
+        } else {
+            parameters.put("dimensions", String.valueOf(info != null ? info.dimension : 1024));
+            parameters.put("normalize", "true");
+            parameters.put("embeddingTypes", "[\"float\"]");
+            requestBody = TITAN_REQUEST_BODY;
+            preProcess = "connector.pre_process.bedrock.embedding";
+            postProcess = "connector.post_process.bedrock.embedding";
+        }
 
         if (modelParameters != null) {
             parameters.putAll(modelParameters);
@@ -70,9 +89,9 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
             .method("POST")
             .url("https://bedrock-runtime.${parameters.region}.amazonaws.com/model/${parameters.model}/invoke")
             .headers(headers)
-            .requestBody(REQUEST_BODY_TEMPLATE)
-            .preProcessFunction("connector.pre_process.bedrock.embedding")
-            .postProcessFunction("connector.post_process.bedrock.embedding")
+            .requestBody(requestBody)
+            .preProcessFunction(preProcess)
+            .postProcessFunction(postProcess)
             .build();
 
         ConnectorClientConfig connectorClientConfig = new ConnectorClientConfig();
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -93,12 +93,9 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
 
     @Override
     public MLRegisterModelInput createModelInput(String modelName, Connector connector, Map<String, String> modelParameters) {
-        EmbeddingModelInfo info = KNOWN_MODELS.get(modelName);
-        FunctionName functionName = info != null ? info.functionName : FunctionName.TEXT_EMBEDDING;
-
         return MLRegisterModelInput
             .builder()
-            .functionName(functionName)
+            .functionName(FunctionName.REMOTE)
             .modelName("Auto-generated embedding model for " + modelName)
             .description("Auto-generated embedding model for memory container")
             .connector(connector)
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -41,17 +41,12 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
      * Known Bedrock embedding models with their type and default dimension.
      */
     public static final Map<String, EmbeddingModelInfo> KNOWN_MODELS = Map
-        .of(
-            "amazon.titan-embed-text-v2:0",
-            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-            "amazon.titan-embed-text-v1",
-            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536),
-            "amazon.titan-embed-image-v1",
-            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-            "cohere.embed-english-v3",
-            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-            "cohere.embed-multilingual-v3",
-            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)
+        .ofEntries(
+            Map.entry("amazon.titan-embed-text-v2:0", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)),
+            Map.entry("amazon.titan-embed-text-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536)),
+            Map.entry("amazon.titan-embed-image-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)),
+            Map.entry("cohere.embed-english-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)),
+            Map.entry("cohere.embed-multilingual-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024))
         );
 
     @Override
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -57,7 +57,7 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
     @Override
     public Connector createConnector(String modelId, Map<String, String> credential, Map<String, String> modelParameters) {
         EmbeddingModelInfo info = KNOWN_MODELS.get(modelId);
-        boolean isCohere = modelId != null && modelId.startsWith("cohere.");
+        boolean isCohere = info != null && modelId.startsWith("cohere.embed");
 
         Map<String, String> parameters = new HashMap<>();
         parameters.put("region", DEFAULT_REGION);
@@ -74,14 +74,12 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
             preProcess = "connector.pre_process.cohere.embedding";
             postProcess = "connector.post_process.cohere.embedding";
         } else {
-            if (info == null) {
-                log
-                    .warn(
-                        "Unknown Bedrock embedding model: {}. Defaulting to dimension 1024. " + "Override via model_parameters if needed.",
-                        modelId
-                    );
+            if (info == null && (modelParameters == null || !modelParameters.containsKey("dimensions"))) {
+                throw new IllegalArgumentException(
+                    "Unknown Bedrock embedding model: " + modelId + ". Please provide 'dimensions' in model_parameters."
+                );
             }
-            parameters.put("dimensions", String.valueOf(info != null ? info.dimension : 1024));
+            parameters.put("dimensions", String.valueOf(info != null ? info.dimension() : 1024));
             parameters.put("normalize", "true");
             parameters.put("embeddingTypes", "[\"float\"]");
             requestBody = TITAN_REQUEST_BODY;
@@ -94,10 +92,7 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
         }
 
         // Validate region to prevent SSRF via URL injection
-        String region = parameters.get("region");
-        if (region != null && !REGION_PATTERN.matcher(region).matches()) {
-            throw new IllegalArgumentException("Invalid AWS region format: " + region);
-        }
+        MemoryContainerConstants.requireValidAwsRegion(parameters.get("region"));
 
         Map<String, String> headers = new HashMap<>();
         headers.put("content-type", "application/json");
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -18,17 +18,17 @@ import org.opensearch.ml.common.connector.ConnectorClientConfig;
 import org.opensearch.ml.common.connector.ConnectorProtocols;
 import org.opensearch.ml.common.input.execute.agent.ContentBlock;
 import org.opensearch.ml.common.input.execute.agent.Message;
+import org.opensearch.ml.common.memorycontainer.MemoryContainerConstants;
 import org.opensearch.ml.common.model.ModelProvider;
 import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
 
-/**
- * Model provider for Bedrock Embedding API (Titan, Cohere).
- * Uses the /invoke endpoint (not /converse) with inputText-based request body.
- */
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
 public class BedrockEmbeddingModelProvider extends ModelProvider {
 
-    private static final String DEFAULT_REGION = "us-east-1";
-    private static final java.util.regex.Pattern REGION_PATTERN = java.util.regex.Pattern.compile("^[a-z]{2}(-[a-z]+-\\d+)?$");
+    private static final String DEFAULT_REGION = MemoryContainerConstants.DEFAULT_AWS_REGION;
+    private static final java.util.regex.Pattern REGION_PATTERN = MemoryContainerConstants.AWS_REGION_PATTERN;
 
     private static final String TITAN_REQUEST_BODY =
         "{ \"inputText\": \"${parameters.inputText}\", \"dimensions\": ${parameters.dimensions},"
@@ -73,6 +73,13 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
             preProcess = "connector.pre_process.cohere.embedding";
             postProcess = "connector.post_process.cohere.embedding";
         } else {
+            if (info == null) {
+                log
+                    .warn(
+                        "Unknown Bedrock embedding model: {}. Defaulting to dimension 1024. " + "Override via model_parameters if needed.",
+                        modelId
+                    );
+            }
             parameters.put("dimensions", String.valueOf(info != null ? info.dimension : 1024));
             parameters.put("normalize", "true");
             parameters.put("embeddingTypes", "[\"float\"]");
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -18,6 +18,7 @@ import org.opensearch.ml.common.connector.ConnectorClientConfig;
 import org.opensearch.ml.common.connector.ConnectorProtocols;
 import org.opensearch.ml.common.input.execute.agent.ContentBlock;
 import org.opensearch.ml.common.input.execute.agent.Message;
+import org.opensearch.ml.common.memorycontainer.EmbeddingModelInfo;
 import org.opensearch.ml.common.memorycontainer.MemoryContainerConstants;
 import org.opensearch.ml.common.model.ModelProvider;
 import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
@@ -180,16 +181,4 @@ public class BedrockEmbeddingModelProvider extends ModelProvider {
         return KNOWN_MODELS.get(modelId);
     }
 
-    /**
-     * Info about a known embedding model.
-     */
-    public static class EmbeddingModelInfo {
-        public final FunctionName functionName;
-        public final int dimension;
-
-        public EmbeddingModelInfo(FunctionName functionName, int dimension) {
-            this.functionName = functionName;
-            this.dimension = dimension;
-        }
-    }
 }
--- a/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
+++ b/common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
@@ -28,24 +28,30 @@ import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
 public class BedrockEmbeddingModelProvider extends ModelProvider {
 
     private static final String DEFAULT_REGION = "us-east-1";
+    private static final java.util.regex.Pattern REGION_PATTERN = java.util.regex.Pattern.compile("^[a-z]{2}(-[a-z]+-\\d+)?$");
 
     private static final String TITAN_REQUEST_BODY =
         "{ \"inputText\": \"${parameters.inputText}\", \"dimensions\": ${parameters.dimensions},"
             + " \"normalize\": ${parameters.normalize}, \"embeddingTypes\": ${parameters.embeddingTypes} }";
 
-    private static final String COHERE_REQUEST_BODY =
-        "{ \"texts\": ${parameters.texts}, \"input_type\": \"${parameters.input_type}\" }";
+    private static final String COHERE_REQUEST_BODY = "{ \"texts\": ${parameters.texts}, \"input_type\": \"${parameters.input_type}\" }";
 
     /**
      * Known Bedrock embedding models with their type and default dimension.
      */
-    public static final Map<String, EmbeddingModelInfo> KNOWN_MODELS = Map.of(
-        "amazon.titan-embed-text-v2:0", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-        "amazon.titan-embed-text-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536),
-        "amazon.titan-embed-image-v1", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-        "cohere.embed-english-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
-        "cohere.embed-multilingual-v3", new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)
-    );
+    public static final Map<String, EmbeddingModelInfo> KNOWN_MODELS = Map
+        .of(
+            "amazon.titan-embed-text-v2:0",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "amazon.titan-embed-text-v1",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536),
+            "amazon.titan-embed-image-v1",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "cohere.embed-english-v3",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "cohere.embed-multilingual-v3",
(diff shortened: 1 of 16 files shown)


common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryModelService.java
```diff
@@ -0,0 +1,212 @@
...
+        if (modelSpec.getModelParameters() != null) {
+            parameters.putAll(modelSpec.getModelParameters());
+        }
+
+        if (provider.equals("bedrock/converse")) {
+            parameters.putIfAbsent("region", DEFAULT_REGION);
+            parameters.put("service_name", "bedrock");
+            // Validate region to prevent SSRF via URL injection
+            String region = parameters.get("region");
+            if (!REGION_PATTERN.matcher(region).matches()) {
```
ylwu-amzn: Inconsistency with `BedrockEmbeddingModelProvider.java:98`: that file guards with `if (region != null && !REGION_PATTERN.matcher(region).matches())`; this one doesn't null-check. Currently safe because `putIfAbsent("region", DEFAULT_REGION)` runs on line 90 and DEFAULT_REGION is non-null, but a caller passing `parameters.put("region", null)` (since user modelParameters are merged before this) would NPE.

Easy fix — match the embedding provider's pattern:

```java
if (region == null || !REGION_PATTERN.matcher(region).matches()) {
    throw new IllegalArgumentException("Invalid AWS region format: " + region);
}
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -99,10 +108,14 @@ public MemoryConfiguration(
...
         boolean disableSession,
         boolean useSystemIndex,
-        String tenantId
+        String tenantId,
+        MLAgentModelSpec embeddingModelSpec,
+        MLAgentModelSpec llmSpec
     ) {
-        // Validate first
-        validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
+        // Skip validation when inline model specs are present (models not yet created)
```
ylwu-amzn: Skipping `validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize)` here is fine because the deferred validation runs later in `proceedWithContainerCreation` once the inline models have been created and their IDs populated. But this invariant is *not* tested — if someone removes or reorders the deferred validation, the skip here becomes a silent bypass.

Please add a unit test in `MemoryConfigurationTests` that:
1. Builds a `MemoryConfiguration` with an inline spec AND bad fields (e.g., invalid `maxInferSize`, or unsupported `embeddingModelType`) that would normally be rejected by `validateInputs`.
2. Asserts construction succeeds (inline path).
3. After simulating model creation (setting `embeddingModelId`/`llmId` and clearing the inline specs), calls whatever validation entry point `proceedWithContainerCreation` uses and asserts it still fails.

This locks in that the skip is a *deferral*, not a *bypass*.

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerActionTests.java
```diff
@@ -1956,4 +1957,258 @@ private void mockSuccessfulLLMValidation() {
             return null;
         }).when(mlModelManager).getModel(eq("test-embedding-model"), any());
     }
+
+    @Test
+    public void testDoExecute_withInlineEmbeddingModel() {
```
ylwu-amzn: Missing test cases (adding inline to call out, would rather these be here than a follow-up):

1. **Inline spec credentials do not land in stored container document.** Parse an inline spec with a distinctive credential (e.g., `"secret_key": "TESTSECRETKEY-XYZ"`), run creation, then `toXContent` the container and assert `"TESTSECRETKEY-XYZ"` does not appear in the serialized bytes. This is the PR description's core security claim — please lock it in.
2. **Orphan-on-partial-failure current behavior.** Stub LLM create to fail after embedding succeeds; assert whatever the current behavior is (even if it's "embedding model is orphaned"). This makes any future cleanup change visible in the diff.
3. **modelParameters cannot override service_name / model / region.** (If you adopt the allowlist suggestion I left on BedrockEmbeddingModelProvider.java:93.)
4. **Malicious modelId is rejected.** (If you adopt the modelId charset validation I suggested.)

The existing `testDoExecute_inlineModelCreationFailure` covers the unhappy-path error propagation but not any of the above.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -86,6 +87,14 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
     private boolean useSystemIndex = true;
     private String tenantId;
 
+    // Transient fields for inline model creation (not stored in container document)
+    public static final String EMBEDDING_MODEL_SPEC_FIELD = "embedding_model";
+    public static final String LLM_SPEC_FIELD = "llm";
+    @EqualsAndHashCode.Exclude
+    private transient MLAgentModelSpec embeddingModelSpec;
+    @EqualsAndHashCode.Exclude
+    private transient MLAgentModelSpec llmSpec;
```
ylwu-amzn: **Blocking — wire serialization gap.**

These fields are `transient` but that only affects Java serialization, not OpenSearch streaming. `writeTo(StreamOutput)` (~line 189-218) and `MemoryConfiguration(StreamInput)` (~line 165-186) don't read/write them. On a multi-node cluster, if this transport action is dispatched from node A to node B, the inline specs are dropped in flight and the request arrives on node B with `embeddingModelSpec == null` / `llmSpec == null`. The action then falls through to the old "must provide IDs" path and returns a confusing error — or worse, proceeds partially.

E2E verified against a single-node Docker wouldn't catch this. Please:
- Write both specs in `writeTo`/`StreamInput` constructor (see `MLAgentModelSpec.java:57-83` for the pattern).
- Gate with a cluster `Version` check since this is a BWC change to the transport protocol.
- Add a round-trip test: `BytesStreamOutput` → `StreamInput` → assert specs survive.

Alternatively, resolve specs → IDs in the REST handler before dispatch so the transport request never carries specs.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -99,10 +108,14 @@ public MemoryConfiguration(
...
+        String tenantId,
+        MLAgentModelSpec embeddingModelSpec,
+        MLAgentModelSpec llmSpec
     ) {
-        // Validate first
-        validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
+        // Skip validation when inline model specs are present (models not yet created)
+        if (embeddingModelSpec == null && llmSpec == null) {
+            validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
+        }
```
ylwu-amzn, line 118: This skips **all** input validation — including `maxInferSize > 10` — when either inline spec is present. A user passing `{embedding_model: {...}, max_infer_size: 999999}` gets past the constructor. `MemoryConfiguration.validate()` (line ~455) only re-runs embedding validation, not `validateMaxInferSize`. Either:
- Only skip the embedding-pair validation (keep `validateMaxInferSize` always on), or
- Call `this.validate()` after `createInlineModels` fills in the IDs, and make sure `validate()` covers every invariant.

The comment says "models not yet created" but nothing re-validates once IDs are assigned — please make the re-validation step explicit.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +103,106 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+        if (config != null && (config.hasInlineEmbeddingModel() || config.hasInlineLlm())) {
+            createInlineModels(
+                config,
+                tenantId,
+                ActionListener.wrap(v -> { proceedWithContainerCreation(input, user, tenantId, listener); }, listener::onFailure)
+            );
+        } else {
+            proceedWithContainerCreation(input, user, tenantId, listener);
+        }
+    }
```
ylwu-amzn, line 134: **Blocking — orphan models on validation failure.**

Model auto-creation happens here, but `proceedWithContainerCreation → validateConfiguration` runs *after* both models are registered. Failure modes that leak orphans:
- User supplies strategy with empty `namespace` → `MemoryStrategy.validate` throws → both models already created.
- User supplies unknown `embedding_model.model_id` (e.g. typo `amazon.titan-embed-text-v3`) → `detectEmbeddingType` returns null → `embeddingModelType` stays null → `validateStrategiesRequireModels` fails → orphans.
- `validateSharedIndexCompatibility` detects a prefix collision → orphans.

The PR description only mentions the "LLM create fails after embedding created" case, but many user-fixable errors also leak. Please:
1. Run the cheap synchronous validations (strategy namespace, strategy-requires-models) **before** `createInlineModels`.
2. Add a best-effort cleanup in the failure path — call `MLDeleteModelAction` for any IDs created in this request.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +103,106 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                .execute(
+                    MLRegisterModelAction.INSTANCE,
+                    modelRequest,
+                    ActionListener.wrap(response -> { listener.onResponse(response.getModelId()); }, e -> {
+                        log.error("Failed to auto-create model for memory container: {}", modelSpec.getModelId(), e);
+                        listener
+                            .onFailure(
+                                new OpenSearchStatusException("Failed to create model: " + e.getMessage(), RestStatus.INTERNAL_SERVER_ERROR)
+                            );
+                    })
```
ylwu-amzn, line 193: **Credential leak via error message.**

`"Failed to create model: " + e.getMessage()` is returned as the 500 response body. For AWS/HTTP connector failures the message can include signed URLs, request IDs, partial auth headers, or echoed request bodies (some providers echo inputs back).

Suggested: keep the server-side `log.error(..., e)` as is, but return a generic `"Failed to create model"` to the client. If you want client-visible diagnostics, explicitly whitelist safe exception types (e.g., `IllegalArgumentException` from validation).

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +103,106 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                log.info("Auto-created LLM model: {}", llmModelId);
+                listener.onResponse(null);
+            }, listener::onFailure));
+        } else {
+            listener.onResponse(null);
+        }
+    }
+
+    private void createModelFromSpec(
+        org.opensearch.ml.common.agent.MLAgentModelSpec modelSpec,
```
ylwu-amzn: Use a regular `import org.opensearch.ml.common.agent.MLAgentModelSpec;` — fully-qualified names in method signatures are inconsistent with the rest of the file (and make it harder to scan). Same pattern appears in several `client.execute(...)` call sites in the test file.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryModelService.java
```diff
@@ -0,0 +1,212 @@
...
+            protocol = ConnectorProtocols.HTTP;
+            headers.put("Authorization", "Bearer ${credential.openAI_key}");
+        } else if (provider.equals("gemini/v1beta/generatecontent")) {
+            url = "https://generativelanguage.googleapis.com/v1beta/models/${parameters.model}:generateContent";
+            requestBody = GEMINI_MEMORY_TEMPLATE;
+            protocol = ConnectorProtocols.HTTP;
+            headers.put("x-goog-api-key", "${credential.gemini_api_key}");
+        } else {
+            throw new IllegalArgumentException("Unsupported LLM provider for memory: " + modelSpec.getModelProvider());
+        }
```
ylwu-amzn: **Blocking — duplicates existing provider dispatch.**

The codebase already has `ModelProviderType` (enum) and `ModelProviderFactory` for exactly this routing (see `AgentModelService.java:30` — same problem, solved via the factory). Comparing raw strings here means:
- Maintenance burden: a new provider now needs to be added in three places (enum, factory, this switch).
- Semantic drift: `.equals` on a `.toLowerCase()`'d string vs `ModelProviderType.from()` which is `equalsIgnoreCase` — both work but differ.
- Bypasses the validation and error message already built into `ModelProviderType.from()`.

Suggested shape:
```java
ModelProviderType type = ModelProviderType.from(modelSpec.getModelProvider());
return switch (type) {
    case BEDROCK_CONVERSE -> buildBedrockConverseMemoryLlm(modelSpec);
    case OPENAI_V1_CHAT_COMPLETIONS -> buildOpenAIMemoryLlm(modelSpec);
    case GEMINI_V1BETA_GENERATE_CONTENT -> buildGeminiMemoryLlm(modelSpec);
    default -> throw new IllegalArgumentException("Unsupported LLM provider for memory: " + modelSpec.getModelProvider());
};
```

Even cleaner: add a `ModelProvider.createMemoryLlmConnector()` method so the memory-specific template lives next to the rest of the provider-specific code and can't drift.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryModelService.java
ylwu-amzn: Asymmetry with `BedrockEmbeddingModelProvider.java:98`, which null-checks `region` before matching. Here `putIfAbsent` guarantees non-null today so no NPE, but the inconsistent code is a trap for future edits.

Also, this identical block (region extraction + regex match + throw) is duplicated between two files. Please extract a helper:
```java
// MemoryContainerConstants.java
public static void requireValidAwsRegion(String region) {
    if (region == null || !AWS_REGION_PATTERN.matcher(region).matches()) {
        throw new IllegalArgumentException("Invalid AWS region format: " + region);
    }
}
```

Also worth asking: `BedrockConverseModelProvider.java:85-107` (used by the existing agent revamp) does **not** validate region. Should it? Currently the rule is "it depends which auto-generator you went through."

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryModelService.java
```diff
@@ -0,0 +1,212 @@
...
+        String provider = modelProvider.toLowerCase();
+        if (provider.equals("bedrock/converse")) {
+            return "$.output.message.content[0].text";
+        } else if (provider.equals("openai/v1/chat/completions")) {
+            return "$.choices[0].message.content";
+        } else if (provider.equals("gemini/v1beta/generatecontent")) {
+            return "$.candidates[0].content.parts[0].text";
+        }
+        return null;
+    }
```
ylwu-amzn, line 202: Same raw-string switch pattern as `createMemoryLlmInput` above — please route through `ModelProviderType` for a single source of truth on the provider-name strings.

ylwu-amzn (commented):
Thanks for the big UX win — collapsing 5 API calls into 1 is meaningful, and I like that the on-disk container format is unchanged. I've left inline comments on specific lines. The three I'd consider blocking are:

1. Inline specs aren't written to `StreamInput/StreamOutput` — will silently fail on multi-node clusters when the transport action lands on a non-coordinator node.
2. Validation runs *after* model auto-creation, so several user-fixable errors (unknown embedding model, strategy namespace, shared-index conflict) leak orphan models.
3. `MemoryModelService` re-implements provider dispatch via raw string `if/else` instead of going through the existing `ModelProviderType` / `ModelProviderFactory` abstraction.

Other comments are smaller — security (credential leak in error), tests that wouldn't catch the Gemini auth regression if it re-appeared, and some style/consistency items.

common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
```diff
@@ -0,0 +1,184 @@
...
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "amazon.titan-embed-text-v1",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1536),
+            "amazon.titan-embed-image-v1",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "cohere.embed-english-v3",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024),
+            "cohere.embed-multilingual-v3",
+            new EmbeddingModelInfo(FunctionName.TEXT_EMBEDDING, 1024)
+        );
```
ylwu-amzn, line 49: `Map.of(...)` caps at 10 entries. The current 5 are fine, but the next PR adding `amazon.titan-embed-text-v3` plus a couple more Cohere variants will hit that silently-painful limit (compile error, but confusing). Switch to `Map.ofEntries(Map.entry(...), ...)` or a static initializer now to avoid the churn.

common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Override
+    public Connector createConnector(String modelId, Map<String, String> credential, Map<String, String> modelParameters) {
+        EmbeddingModelInfo info = KNOWN_MODELS.get(modelId);
+        boolean isCohere = modelId != null && modelId.startsWith("cohere.");
```
ylwu-amzn: `startsWith("cohere.")` is too loose — will match any future Bedrock-namespaced Cohere model, not just embedding ones. If Cohere releases a rerank/chat model on Bedrock with the `cohere.` prefix, it will silently be treated as an embedding model. Consider explicit per-model handling or an enum `switch` over known IDs.

common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
```diff
@@ -0,0 +1,184 @@
...
+            preProcess = "connector.pre_process.cohere.embedding";
+            postProcess = "connector.post_process.cohere.embedding";
+        } else {
+            if (info == null) {
+                log
+                    .warn(
+                        "Unknown Bedrock embedding model: {}. Defaulting to dimension 1024. " + "Override via model_parameters if needed.",
+                        modelId
+                    );
+            }
```
ylwu-amzn: Silently defaulting unknown models to `dimension=1024` will produce embeddings at the model's actual dimension (often 512, 768, 3072) into a knn_vector field configured for 1024. Depending on the provider that either throws at index time or silently truncates/pads.

Either:
- Require `dimension` in `model_parameters` when the model isn't in `KNOWN_MODELS`, or
- Promote this warning to a hard error.

A warning log that nobody sees until semantic search returns nonsense is worse than an upfront failure.

common/src/main/java/org/opensearch/ml/common/agent/OpenaiEmbeddingModelProvider.java
```diff
@@ -0,0 +1,126 @@
...
+import org.opensearch.ml.common.input.execute.agent.ContentBlock;
+import org.opensearch.ml.common.input.execute.agent.Message;
+import org.opensearch.ml.common.memorycontainer.EmbeddingModelInfo;
+import org.opensearch.ml.common.model.ModelProvider;
+import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
+
+/**
+ * Model provider for OpenAI Embedding API (text-embedding-3-small, text-embedding-3-large, etc.)
+ */
+public class OpenaiEmbeddingModelProvider extends ModelProvider {
```
ylwu-amzn, line 31: Missing `@Log4j2` (sibling `BedrockEmbeddingModelProvider` has it). Related: Bedrock logs a warning for unknown model IDs (line 78-83 there) but the OpenAI equivalent silently omits the dimension parameter. Please add the parallel warning here so the behavior matches.

common/src/main/java/org/opensearch/ml/common/memorycontainer/EmbeddingModelInfo.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class EmbeddingModelInfo {
+    public final FunctionName functionName;
+    public final int dimension;
+
+    public EmbeddingModelInfo(FunctionName functionName, int dimension) {
+        this.functionName = functionName;
+        this.dimension = dimension;
+    }
+}
```
ylwu-amzn: Immutable two-field value holder — this is what records are for. Codebase already uses Java 21 features. Consider:
```java
public record EmbeddingModelInfo(FunctionName functionName, int dimension) {}
```
Call sites change from `info.dimension` to `info.dimension()` — small diff, but avoids the public-mutable-looking `public final` field style and gives you equals/hashCode/toString for free.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -10,6 +10,10 @@
  */
 public class MemoryContainerConstants {
 
+    // AWS region constants for inline model creation
+    public static final String DEFAULT_AWS_REGION = "us-east-1";
+    public static final java.util.regex.Pattern AWS_REGION_PATTERN = java.util.regex.Pattern.compile("^[a-z]{2}(-[a-z]+-\\d+)?$");
```
ylwu-amzn: Please use a regular `import java.util.regex.Pattern;` — fully-qualified name in field declarations is inconsistent with the rest of the codebase. Same thing is done in `MemoryModelService.java:46`.

common/src/test/java/org/opensearch/ml/common/memorycontainer/MemoryModelServiceTest.java
```diff
@@ -0,0 +1,195 @@
...
+            .builder()
+            .modelId("gemini-2.0-flash")
+            .modelProvider("gemini/v1beta/generatecontent")
+            .credential(Map.of("gemini_api_key", "test"))
+            .build();
+
+        MLRegisterModelInput input = MemoryModelService.createModelFromSpec(spec, true);
+        assertNotNull(input);
+        assertEquals(FunctionName.REMOTE, input.getFunctionName());
+    }
```
ylwu-amzn, line 149: These three `memoryLlm_*` tests only assert `assertNotNull(input)` and `FunctionName.REMOTE`. They wouldn't catch:
- Wrong URL template.
- Missing `${parameters.system_prompt}` / `${parameters.user_prompt}` in request body.
- Missing `Authorization: Bearer ${credential.openAI_key}` header for OpenAI.
- Missing `x-goog-api-key` header for Gemini (this was literally the bug fixed in commit 342475f6 on this branch — it would silently regress).

Suggested extra assertions:
```java
Connector c = input.getConnector();
ConnectorAction action = c.getActions().get(0);
assertTrue(action.getHeaders().containsKey("x-goog-api-key"));
assertFalse(action.getUrl().contains("key="));
assertTrue(action.getRequestBody().contains("${parameters.system_prompt}"));
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +103,106 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+            createModelFromSpec(config.getLlmSpec(), true, ActionListener.wrap(llmModelId -> {
+                config.setLlmId(llmModelId);
+                setLlmResultPathFromProvider(config);
+                log.info("Auto-created LLM model: {}", llmModelId);
+                listener.onResponse(null);
+            }, listener::onFailure));
+        } else {
+            listener.onResponse(null);
+        }
+    }
```
ylwu-amzn, line 175: The embedding-then-LLM callback-nesting is already ~40 lines deep. If a third inline resource is ever added (reranker? vector store connector?), this becomes hard to read and hard to add cleanup logic to (see the orphan-model concern). Consider `GroupedActionListener` or a small sequential composer (`ActionListener.runAfter` or an explicit state machine) so adding the next dependency is a one-line diff.

dhrubo-os (commented):
Thanks Yaliang for the thorough review — all 16 comments addressed. Every fix has been E2E verified against a 4-node Docker cluster with all provider combos (Bedrock Titan+Claude, Cohere+Claude, OpenAI embed+OpenAI LLM, OpenAI embed+Bedrock LLM). Pre-validation also verified: `maxInferSize: 999` correctly returns 400 before any models are created.

Summary of fixes:
- **Blocking #1 (serialization):** Added `writeTo`/`StreamInput` for specs + fixed `maxInferSize` preservation bug
- **Blocking #2 (orphans):** Pre-validation before model creation + best-effort cleanup on failure
- **Blocking #3 (dispatch):** Refactored to `ModelProviderType` switch throughout
- All other items: credential leak fix, unknown model hard error, Cohere detection tightened, region helper extracted, record type, Map.ofEntries, @Log4j2, FQN imports, test assertions, callback flattening

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -99,18 +108,24 @@ public MemoryConfiguration(
...
```
dhrubo-os, line 118: Good catch — fixed. `validateNonModelInputs()` now explicitly rejects conflicting configurations:
```java
if (embeddingModelSpec != null && embeddingModelId != null)
    throw "Cannot specify both 'embedding_model' (inline) and 'embedding_model_id'";
if (llmSpec != null && llmId != null)
    throw "Cannot specify both 'llm' (inline) and 'llm_id'";
```
This runs before any model creation, so no orphans.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +104,144 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                listener.onFailure(e);
+                return;
+            }
+
+            createInlineModels(
+                config,
+                tenantId,
+                ActionListener.wrap(v -> { proceedWithContainerCreation(input, user, tenantId, listener); }, e -> {
+                    // Best-effort cleanup of any models created before failure
+                    cleanupOrphanModels(config);
```
dhrubo-os: If cleanup fails, it's logged server-side via `log.warn("Failed to cleanup orphan model", e)`. The user still receives the original error that caused the failure (e.g., validation error, LLM creation failure). We don't surface the cleanup failure to the user because:
1. The original error is what they need to fix
2. The orphan model IDs are visible in the model index — user can delete manually if needed
3. Adding cleanup failure details to the error response would be confusing (two errors in one response)

The model IDs are also logged at INFO level when created, so they're traceable.

common/src/main/java/org/opensearch/ml/common/input/execute/agent/ModelProviderType.java
```diff
@@ -13,6 +13,8 @@
  */
 public enum ModelProviderType {
     BEDROCK_CONVERSE("bedrock/converse"),
+    BEDROCK_EMBEDDING("bedrock/embedding"),
```
dhrubo-os, line 16: No — `BEDROCK_EMBEDDING` and `OPENAI_EMBEDDING` are for memory container inline model creation only. They throw `UnsupportedOperationException` for all agent operations (`mapTextInput`, `mapMessages`, `extractMessageFromResponse`, etc.). The agent framework uses `BEDROCK_CONVERSE`, `OPENAI_V1_CHAT_COMPLETIONS`, and `GEMINI_V1BETA_GENERATE_CONTENT` for LLMs. The embedding providers are only invoked by `MemoryModelService.createModelFromSpec()`.

common/src/main/java/org/opensearch/ml/common/input/execute/agent/ModelProviderType.java
```diff
@@ -13,6 +13,8 @@
...
+    OPENAI_EMBEDDING("openai/v1/embeddings"),
```
dhrubo-os, line 17: The `default -> throw` case already exists in `ModelProviderFactory` before this PR — it's a pre-existing pattern. Our change only adds two new enum values (`BEDROCK_EMBEDDING`, `OPENAI_EMBEDDING`) and registers them in the switch. The `default` branch does catch any unhandled enum values at runtime. Agree that compile-time exhaustiveness would be better, but that's a broader refactor of the existing factory pattern — not in scope for this PR.

dhrubo-os (commented):
Thanks for the review! All comments addressed.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,144 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                log.warn("Failed to cleanup orphan embedding model: {}", config.getEmbeddingModelId(), e);
+            }
+        }
+        if (config.getLlmId() != null && config.hasInlineLlm()) {
+            log.info("Cleaning up orphan LLM model: {}", config.getLlmId());
+            try {
+                client
+                    .execute(
+                        MLModelDeleteAction.INSTANCE,
+                        new MLModelDeleteRequest(config.getLlmId(), null),
```
ylwu-amzn: **High-priority — cleanup drops tenant context.**

Both `MLModelDeleteRequest` calls pass `null` for the tenantId, but the auto-created models were registered under the request's tenant. On multi-tenant clusters this will either fail the delete (leaving the orphan we were trying to clean up) or — worse — succeed against a different tenant's scope.

**Fix:** pass `config.getTenantId()` instead of `null`:

```java
new MLModelDeleteRequest(config.getEmbeddingModelId(), config.getTenantId())
...
new MLModelDeleteRequest(config.getLlmId(), config.getTenantId())
```

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerActionTests.java
```diff
@@ -1956,4 +1957,308 @@ private void mockSuccessfulLLMValidation() {
...
+                eq(org.opensearch.ml.common.transport.register.MLRegisterModelAction.INSTANCE),
+                any(org.opensearch.ml.common.transport.register.MLRegisterModelRequest.class),
+                any(ActionListener.class)
+            );
+
+        action.doExecute(null, request, actionListener);
+
+        ArgumentCaptor<Exception> captor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(captor.capture());
+        assertTrue(captor.getValue().getMessage().contains("Failed to create model"));
```
ylwu-amzn, line 2199: **High-priority — cleanup path has no test coverage.**

This test verifies that model-creation failure surfaces to the listener, but doesn't verify `MLModelDeleteAction.INSTANCE` was actually invoked. More importantly, the real regression risk is the "embedding succeeds → LLM fails" case: the embedding model is registered and must be cleaned up as an orphan. That path isn't exercised anywhere.

Please add a test that:
1. Mocks the first `MLRegisterModelAction` call to succeed (returning an embedding model ID)
2. Mocks the second `MLRegisterModelAction` call to fail
3. `verify(client).execute(eq(MLModelDeleteAction.INSTANCE), argThat(req -> req.getModelId().equals("auto-embed-id")), any())`

This would also catch the tenantId bug I flagged on the cleanup request.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,144 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                config,
+                tenantId,
+                ActionListener.wrap(v -> { proceedWithContainerCreation(input, user, tenantId, listener); }, e -> {
+                    // Best-effort cleanup of any models created before failure
+                    cleanupOrphanModels(config);
+                    listener.onFailure(e);
+                })
+            );
+        } else {
+            proceedWithContainerCreation(input, user, tenantId, listener);
```
ylwu-amzn: **High-priority — orphan window still exists after successful inline creation.**

`cleanupOrphanModels` is only invoked when the inline-model creation chain itself fails. Once we return success into `proceedWithContainerCreation`, any failure in the downstream steps leaves both auto-created models as orphans with no cleanup:
- `validateConfiguration` (strategy validation, shared-index conflict, model-validator failure)
- `initMemoryContainerIndex` (cluster/index failure)
- `indexMemoryContainer` (SDK failure)
- `createMemoryDataIndices` (any of the 4 data-index creations)

The PR description calls this a Phase 1 limitation — fine as a release decision — but the cleanup helper now exists. Would be ~5 lines to wire it into the downstream `onFailure` paths of `proceedWithContainerCreation` so the Phase 2 follow-up isn't a new implementation. If you prefer to defer, please leave a `// TODO:` at this site so the cleanup extension point is discoverable.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -287,6 +324,12 @@ public static MemoryConfiguration parse(XContentParser parser) throws IOExceptio
                 case LLM_ID_FIELD:
                     llmId = parser.text();
                     break;
+                case EMBEDDING_MODEL_SPEC_FIELD:
+                    embeddingModelSpec = MLAgentModelSpec.parse(parser);
+                    break;
+                case LLM_SPEC_FIELD:
+                    llmSpec = MLAgentModelSpec.parse(parser);
+                    break;
```
ylwu-amzn, line 332: Pre-existing bug surfaced by this PR: the switch has no `case TENANT_ID_FIELD`. `toXContent` writes `tenant_id` (line 285-287), the constructor accepts it, and the builder chain at line 394 wires it through — but `parse()` silently drops any `tenant_id` in the request body (falls through to `parser.skipChildren()` at the `default:` branch).

Not introduced by this PR, but inline model creation makes multi-tenancy observable (models get registered per-tenant), so worth fixing while you're already touching this switch. Add:

```java
case TENANT_ID_FIELD:
    tenantId = parser.text();
    break;
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,144 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                            .onFailure(
+                                e instanceof IllegalArgumentException
+                                    ? e
+                                    : new OpenSearchStatusException("Failed to create model", RestStatus.INTERNAL_SERVER_ERROR)
+                            );
+                    })
+                );
+        } catch (Exception e) {
+            log.error("Failed to build model registration input: {}", modelSpec.getModelId(), e);
+            listener.onFailure(new IllegalArgumentException("Invalid model specification: " + e.getMessage()));
```
ylwu-amzn: `catch (Exception e)` with `e.getMessage()` interpolated into the 400 response. For `IllegalArgumentException` from `validateModelSpec` the messages are user-facing and fine, but for any other runtime exception (NPE from a mistyped builder, `IllegalStateException` from connector build) the message may contain internal detail.

Safer split:

```java
} catch (IllegalArgumentException e) {
    listener.onFailure(e);  // user-facing messages pass through
} catch (Exception e) {
    log.error("Failed to build model registration input: {}", modelSpec.getModelId(), e);
    listener.onFailure(new OpenSearchStatusException(
        "Failed to build model registration input", RestStatus.INTERNAL_SERVER_ERROR));
}
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -341,9 +392,62 @@ public static MemoryConfiguration parse(XContentParser parser) throws IOExceptio
...
+        }
+        if (llmSpec != null && llmId != null) {
+            throw new IllegalArgumentException("Cannot specify both 'llm' (inline) and 'llm_id'. Use one or the other.");
+        }
+        if (maxInferSize != null && maxInferSize > 10) {
+            throw new IllegalArgumentException(MAX_INFER_SIZE_LIMIT_ERROR);
+        }
+        if (strategies != null) {
+            for (MemoryStrategy strategy : strategies) {
+                MemoryStrategy.validate(strategy);
```
ylwu-amzn, line 422: Strategy validation is now run twice: once here in `validateNonModelInputs()` (pre-model-creation) and again in `TransportCreateMemoryContainerAction.validateConfiguration` after the models are registered. Not a bug — the validation is idempotent — but it would be cleaner to skip the second pass when `hasInlineEmbeddingModel()` or `hasInlineLlm()` was true, since we already validated before any models were created.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryModelService.java
```diff
@@ -0,0 +1,216 @@
...
+
+    private static final String OPENAI_MEMORY_TEMPLATE =
+        "{\"model\":\"${parameters.model}\",\"messages\":[{\"role\":\"system\",\"content\":\"${parameters.system_prompt}\"},"
+            + "{\"role\":\"user\",\"content\":\"${parameters.user_prompt}\"}]}";
+
+    private static final String GEMINI_MEMORY_TEMPLATE = "{\"system_instruction\":{\"parts\":[{\"text\":\"${parameters.system_prompt}\"}]},"
+        + "\"contents\":[{\"role\":\"user\",\"parts\":[{\"text\":\"${parameters.user_prompt}\"}]}]}";
+
+    private static final String DEFAULT_REGION = MemoryContainerConstants.DEFAULT_AWS_REGION;
+    private static final java.util.regex.Pattern REGION_PATTERN = MemoryContainerConstants.AWS_REGION_PATTERN;
```
ylwu-amzn: Dead field. `REGION_PATTERN` is declared but never referenced — all region validation goes through `MemoryContainerConstants.requireValidAwsRegion()` (line 93). Same dead field in `BedrockEmbeddingModelProvider.java:32`. Please remove both.

common/src/main/java/org/opensearch/ml/common/agent/BedrockEmbeddingModelProvider.java
```diff
@@ -0,0 +1,174 @@
...
+import org.opensearch.ml.common.model.ModelProvider;
+import org.opensearch.ml.common.transport.register.MLRegisterModelInput;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class BedrockEmbeddingModelProvider extends ModelProvider {
+
+    private static final String DEFAULT_REGION = MemoryContainerConstants.DEFAULT_AWS_REGION;
+    private static final java.util.regex.Pattern REGION_PATTERN = MemoryContainerConstants.AWS_REGION_PATTERN;
```
ylwu-amzn: Dead field — `REGION_PATTERN` is declared but never referenced (validation goes through `MemoryContainerConstants.requireValidAwsRegion()` at line 90). Same dead field exists in `MemoryModelService.java:48`. Remove both.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -341,9 +392,62 @@ public static MemoryConfiguration parse(XContentParser parser) throws IOExceptio
...
+            .embeddingModelSpec(embeddingModelSpec)
+            .llmSpec(llmSpec)
             .build();
     }
 
+    /**
+     * Returns true if inline embedding model spec is provided (needs auto-creation).
+     */
+    /**
+     * Validates inputs that don't depend on model creation (maxInferSize, strategies, index prefix).
```
ylwu-amzn: Two back-to-back Javadoc blocks — the first one (`Returns true if inline embedding model spec...`) is orphaned; it belongs to `hasInlineEmbeddingModel()` at line 440, not `validateNonModelInputs()`. Delete lines 400-402 so `validateNonModelInputs` keeps just its own docstring.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,147 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                            )
+                    );
+            } catch (Exception e) {
+                log.warn("Failed to cleanup orphan embedding model: {}", config.getEmbeddingModelId(), e);
+            }
+        }
+        if (config.getLlmId() != null && config.hasInlineLlm()) {
+            log.info("Cleaning up orphan LLM model: {}", config.getLlmId());
+            try {
+                client
```
dhrubo-os, line 224: Fixed — now passes `config.getTenantId()` instead of `null` for both delete requests.

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerActionTests.java
```diff
@@ -1956,4 +1957,401 @@ private void mockSuccessfulLLMValidation() {
...
```
dhrubo-os, line 2199: Added test `testDoExecute_embeddingSucceeds_llmFails_cleanupInvoked` — mocks first register call to succeed, second to fail, then verifies `MLModelDeleteAction.INSTANCE` was invoked for cleanup.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,147 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                // TODO: Wire cleanupOrphanModels into proceedWithContainerCreation failure paths
+                // (validateConfiguration, initMemoryContainerIndex, indexMemoryContainer, createMemoryDataIndices)
+                // to clean up auto-created models if downstream steps fail.
+                proceedWithContainerCreation(input, user, tenantId, listener);
+            }, e -> {
+                // Best-effort cleanup of any models created before failure
+                cleanupOrphanModels(config);
+                listener.onFailure(e);
+            }));
+        } else {
```
dhrubo-os, line 131: Added TODO at the call site. The cleanup helper exists and is ~5 lines to wire in, but touching `proceedWithContainerCreation`'s existing failure paths risks scope creep in this PR. The TODO makes the extension point discoverable.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
dhrubo-os, line 332: Fixed — added `case TENANT_ID_FIELD: tenantId = parser.textOrNull(); break;` to the parse switch. Good catch on the pre-existing bug.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -98,6 +107,147 @@ protected void doExecute(Task task, MLCreateMemoryContainerRequest request, Acti
...
+                        listener
+                            .onFailure(
+                                e instanceof IllegalArgumentException
+                                    ? e
+                                    : new OpenSearchStatusException("Failed to create model", RestStatus.INTERNAL_SERVER_ERROR)
+                            );
+                    })
+                );
+        } catch (IllegalArgumentException e) {
+            listener.onFailure(e);
```
dhrubo-os, line 196: Fixed — split into `catch (IllegalArgumentException e)` (passes through) and `catch (Exception e)` (generic message to client, full exception logged server-side).

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -341,9 +395,62 @@ public static MemoryConfiguration parse(XContentParser parser) throws IOExceptio
...
+     * Called before inline model creation to catch user errors early and avoid orphan models.
+     */
+    public void validateNonModelInputs() {
+        // Reject conflicting configurations
+        if (embeddingModelSpec != null && embeddingModelId != null) {
+            throw new IllegalArgumentException(
+                "Cannot specify both 'embedding_model' (inline) and 'embedding_model_id'. Use one or the other."
+            );
```
dhrubo-os, line 412: Acknowledged — the double validation is idempotent and harmless. The pre-validation catches errors before model creation (avoiding orphans), and the post-validation in `validateConfiguration` covers the full set including model-dependent checks. Skipping the second pass would require threading state about whether pre-validation ran, which adds complexity for no functional benefit.

dhrubo-os (commented):
All 6 comments addressed. Thanks for the thorough second pass — the tenantId null in cleanup was a real bug.

ylwu-amzn (approved):

PR #4803, opened 2026-04-28
feat: Add MCP connector and MCP server metrics

rithinpullela:
### Description

Adds OTel-backed metrics for MCP connectors (client-side — used by agents to call external MCP servers) and the ml-commons MCP server (server-side — exposes ml-commons tools over MCP). Follows the metrics pattern established in #4221 (`MLAdoptionMetricsCounter` / `MLOperationalMetricsCounter`).

#### New metrics

**Adoption (periodic scan in `MLStatsJobProcessor`):**
| Metric | Tags |
|---|---|
| `ml.commons.MCP_CONNECTOR_COUNT` | `protocol` (`mcp_sse` / `mcp_streamable_http`) |
| `ml.commons.MCP_SERVER_REGISTERED_TOOL_COUNT` | `tool_type` |

**Operational (emitted at call sites):**
| Metric | Tags |
|---|---|
| `ml.commons.MCP_CONNECTOR_TOOL_INVOCATION_COUNT` / `_LATENCY` | `protocol`, `status` |
| `ml.commons.MCP_SERVER_TOOL_CALL_COUNT` / `_LATENCY` | `tool_type`, `status` |
| `ml.commons.MCP_SERVER_REQUEST_COUNT` | `method`, `status` |

`cluster_name` is auto-injected by `AbstractMLMetricsCounter`, so **"# domains using MCP connector/server"** is derived upstream from `COUNT(DISTINCT cluster_name)` — same pattern as the existing `MODEL_COUNT`/`AGENT_COUNT` metrics.

The `method` tag on `MCP_SERVER_REQUEST_COUNT` captures the JSON-RPC method (`initialize`, `tools/list`, `tools/call`, `notifications/initialized`, or `unknown` on parse failure). Cardinality is bounded by the MCP spec's fixed method set; the tag makes the difference between "clients browsing the catalog" and "clients actually executing tools" visible.

#### Module move

`AbstractMLMetricsCounter` and `MetricType` are moved from `plugin/` to `common/` (same package path). Reason: `ml-algorithms` hosts `McpSseTool` / `McpStreamableHttpTool` and can't depend on `plugin`. Existing `MLAdoptionMetricsCounter` / `MLOperationalMetricsCounter` remain in `plugin/` untouched.

#### Bug fix included

Removes the `if (mlFeatureEnabledSetting.isMetricCollectionEnabled())` gate around counter **initialization** in `MachineLearningPlugin`. That guard only skipped `initialize()` — every runtime call site (e.g. `MLPredictTaskRunner`, and now the MCP call sites) calls `getInstance()` unconditionally. On the default install where `plugins.ml_commons.metrics_collection_enabled=false` (a `Final` setting), the counters were never initialized, so `getInstance()` threw `IllegalStateException` inside request handlers and broke `POST /_plugins/_ml/mcp` entirely.

Reproduced live: first `tools/list` call on a stock cluster returned `JSON_RPC_INTERNAL_ERROR` with `"MLMcpServerMetricsCounter is not initialized"`. Fix: always `initialize()` the counters; emission is already gated inside `incrementCounter` / `recordHistogram` by the same feature flag, so no behavior change when metrics are off. Added `test_doExecute_succeedsWhenMetricCollectionDisabled` as a regression guard. This latent bug also existed for the pre-existing `MLOperationalMetricsCounter` in the predict path; the fix covers both.

### Verification

- Unit tests added/extended: counter scaffolding (`MLMcpMetricsCountersTests` in `common`), tool invocation (`McpSseToolTests`, `McpStreamableHttpToolTests`), server request (`TransportMcpServerActionTests`), tool-call wrapping (`McpToolsHelperTests`), scan collectors (`MLStatsJobProcessorTests`).
- Manually verified on a local cluster (`./gradlew :opensearch-ml-plugin:run`): registered a tool, fired `tools/list` and `tools/call`, observed correct emission with expected tags in both metrics-enabled and metrics-disabled configurations.

### Related Issues
N/A

### Check List


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
```diff
@@ -58,19 +62,35 @@ public McpSseTool(McpSyncClient mcpSyncClient) {
...
+    private void recordInvocation(long startNanos, String status) {
+        double latencyMs = (System.nanoTime() - startNanos) / 1_000_000.0;
+        Tags tags = Tags.create().addTag("protocol", MCP_SSE).addTag("status", status);
+        MLMcpConnectorMetricsCounter.getInstance().incrementCounter(McpConnectorMetric.MCP_CONNECTOR_TOOL_INVOCATION_COUNT, tags);
+        MLMcpConnectorMetricsCounter
+            .getInstance()
+            .recordHistogram(
+                McpConnectorMetric.MCP_CONNECTOR_TOOL_INVOCATION_LATENCY,
+                latencyMs,
+                Tags.create().addTag("protocol", MCP_SSE)
```
nathaliellenaa: The counter is tagged with protocol + status, but the histogram only carries protocol. This means the counter can be sliced by success vs failure, but the histogram cannot. Is this intentional?

plugin/src/main/java/org/opensearch/ml/jobs/processors/MLStatsJobProcessor.java
```diff
@@ -250,4 +259,65 @@ Tags addTagIfExists(Map<String, ?> sourceTagsMap, String sourceKey, String targe
...
+    void collectMcpConnectorMetrics() {
+        if (!clusterService.state().metadata().indices().containsKey(ML_CONNECTOR_INDEX)) {
+            log.info("Skipping MCP connector metrics collection - ML connector index not found");
+            return;
+        }
+
+        SearchRequest searchRequest = new SearchRequest(ML_CONNECTOR_INDEX);
+        SearchSourceBuilder source = new SearchSourceBuilder();
+        source.query(QueryBuilders.termsQuery("protocol", MCP_SSE, MCP_STREAMABLE_HTTP));
+        source.size(BATCH_SIZE);
```
nathaliellenaa, line 275: If there are more MCP connectors than `BATCH_SIZE` (which is 10,000 based on existing code), the extra ones are silently undercounted. I found the existing `collectModelAndAgentMetrics` method also has the same architecture, should we add a warning log when totalHits > BATCH_SIZE so operators notice

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
ylwu-amzn: **High-priority — histogram drops the `status` tag.**

The counter at line 84 carries both `protocol` and `status`, but this histogram only carries `protocol`. Operators can compute call success rate, but cannot compute p50/p99 latency split by status — and that's usually the more interesting question, since fast validation failures and slow timeouts have very different latency profiles. Mixing them obscures regressions.

Suggested fix: pass the same `tags` object built on line 83 (or a new one with the same fields) to both the counter and the histogram. `cluster_name` auto-injection in `AbstractMLMetricsCounter` handles itself.

Same issue at `McpStreamableHttpTool.java:91` and `McpToolsHelper.java:118`.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpStreamableHttpTool.java
```diff
@@ -65,13 +70,28 @@ public <T> void run(Map<String, String> originalParameters, ActionListener<T> li
...
+    private void recordInvocation(long startNanos, String status) {
+        double latencyMs = (System.nanoTime() - startNanos) / 1_000_000.0;
+        Tags tags = Tags.create().addTag("protocol", MCP_STREAMABLE_HTTP).addTag("status", status);
+        MLMcpConnectorMetricsCounter.getInstance().incrementCounter(McpConnectorMetric.MCP_CONNECTOR_TOOL_INVOCATION_COUNT, tags);
+        MLMcpConnectorMetricsCounter
+            .getInstance()
+            .recordHistogram(
+                McpConnectorMetric.MCP_CONNECTOR_TOOL_INVOCATION_LATENCY,
+                latencyMs,
+                Tags.create().addTag("protocol", MCP_STREAMABLE_HTTP)
```
ylwu-amzn: **High-priority — same `status`-tag-dropped issue as in `McpSseTool.java:90`.** See the comment there for the rationale and suggested fix.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/McpToolsHelper.java
```diff
@@ -90,20 +94,30 @@ public McpStatelessServerFeatures.AsyncToolSpecification createToolSpecification
...
         );
     }
 
+    private static void recordToolCall(String toolType, String status, long startNanos) {
+        double latencyMs = (System.nanoTime() - startNanos) / 1_000_000.0;
+        Tags countTags = Tags.create().addTag("tool_type", toolType).addTag("status", status);
+        MLMcpServerMetricsCounter.getInstance().incrementCounter(McpServerMetric.MCP_SERVER_TOOL_CALL_COUNT, countTags);
+        MLMcpServerMetricsCounter
+            .getInstance()
+            .recordHistogram(McpServerMetric.MCP_SERVER_TOOL_CALL_LATENCY, latencyMs, Tags.create().addTag("tool_type", toolType));
```
ylwu-amzn: **High-priority — histogram drops the `status` tag.**

Same issue as `McpSseTool.java:90` / `McpStreamableHttpTool.java:91`: the counter on line 115 has `tool_type` + `status`, but this histogram only has `tool_type`. Without `status`, success-vs-failure latency cannot be separated in dashboards.

Suggested fix: pass `countTags` (or an equivalent) to `recordHistogram` here too.

plugin/src/main/java/org/opensearch/ml/action/mcpserver/TransportMcpServerAction.java
```diff
@@ -96,21 +103,40 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<MLMcpS
...
+        }
+        if (message instanceof McpSchema.JSONRPCNotification) {
+            return ((McpSchema.JSONRPCNotification) message).method();
+        }
+        return "unknown";
+    }
+
+    private static void recordRequest(String method, String status) {
+        Tags tags = Tags.create().addTag("method", method).addTag("status", status);
+        MLMcpServerMetricsCounter.getInstance().incrementCounter(McpServerMetric.MCP_SERVER_REQUEST_COUNT, tags);
```
ylwu-amzn, line 157: **High-priority — `method` tag is unbounded client input.**

The PR description claims "Cardinality is bounded by the MCP spec's fixed method set," but `extractMethod` (line 125-133) returns whatever string the client sent in the JSON-RPC `method` field. A buggy or hostile client can send any string (typos, version probes, intentional abuse), and each unique string becomes a new time series in the OTel backend. For exporters with per-series cost or memory pressure (Prometheus/Cortex/etc.), this is a slow leak that won't show up in unit tests — and there's no allowlist applied at this point in the request path either.

Suggested fix: validate `method` against the actually-served method set (`initialize`, `notifications/initialized`, `tools/list`, `tools/call`, `ping`, `notifications/cancelled`, etc. — whatever ml-commons supports) and collapse anything else to `other`. Same shape as the existing `"unknown"` fallback for parse errors.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
```diff
@@ -58,19 +62,35 @@ public McpSseTool(McpSyncClient mcpSyncClient) {
...
+            recordInvocation(startNanos, "success");
             listener.onResponse((T) resultJson);
         } catch (Exception e) {
+            recordInvocation(startNanos, "failure");
             log.error("Failed to call MCP tool: {}", this.getName(), e);
             listener.onFailure(e);
         }
     }
 
+    private void recordInvocation(long startNanos, String status) {
```
ylwu-amzn: Medium — this `recordInvocation` is byte-for-byte identical to `McpStreamableHttpTool.recordInvocation` except for the protocol constant. Worth extracting to a small static helper in this package, e.g. `McpToolMetrics.recordConnectorInvocation(String protocol, long startNanos, String status)`. Keeps call sites readable and gives one place to update if the tag schema evolves (e.g. when fixing the histogram-status issue above). Not blocking.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MLStatsJobProcessor.java
```diff
@@ -250,4 +259,65 @@ Tags addTagIfExists(Map<String, ?> sourceTagsMap, String sourceKey, String targe
...
+        searchRequest.source(source);
+
+        client.search(searchRequest, new ActionListener<SearchResponse>() {
+            @Override
+            public void onResponse(SearchResponse searchResponse) {
+                for (SearchHit hit : searchResponse.getHits()) {
+                    Object protocolObj = hit.getSourceAsMap().get("protocol");
+                    String protocol = protocolObj == null ? "unknown" : protocolObj.toString();
```
ylwu-amzn: Medium — the `"unknown"` fallback is unreachable here. The search query at line 272 (`termsQuery("protocol", MCP_SSE, MCP_STREAMABLE_HTTP)`) only returns hits where `protocol` is one of those two values, so `protocolObj` is never null at this point.

Either drop the fallback (let an NPE surface a real indexing bug) or keep it as defensive but add a one-line comment noting that. The same pattern is genuinely defensive in `collectMcpServerRegisteredToolMetrics` line 311 because that scan is unfiltered — only the connector scan is over-defensive.

plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java
```diff
@@ -948,10 +950,12 @@ public Collection<Object> createComponents(
...
-        if (mlFeatureEnabledSetting.isMetricCollectionEnabled()) {
-            MLOperationalMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
-            MLAdoptionMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
-        }
+        // Always initialize counters so call sites can getInstance() safely. Emission itself is
+        // gated inside AbstractMLMetricsCounter by MLFeatureEnabledSetting#isMetricCollectionEnabled().
+        MLOperationalMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
+        MLAdoptionMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
+        MLMcpConnectorMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
+        MLMcpServerMetricsCounter.initialize(clusterService.getClusterName().toString(), metricsRegistry, mlFeatureEnabledSetting);
```
ylwu-amzn: Nit — `clusterService.getClusterName().toString()` is repeated four times. A local `String clusterName = clusterService.getClusterName().toString();` above this block reads better and avoids drift if a fifth counter shows up. Not blocking.

common/src/test/java/org/opensearch/ml/stats/otel/counters/MLMcpMetricsCountersTests.java
```diff
@@ -0,0 +1,53 @@
...
+        assertThrows(IllegalStateException.class, MLMcpConnectorMetricsCounter::getInstance);
+        assertThrows(IllegalStateException.class, MLMcpServerMetricsCounter::getInstance);
+    }
+
+    @Test
+    public void initializeThenGetInstanceReturnsSingleton() {
+        MLMcpConnectorMetricsCounter.initialize("c", registry, featureFlag);
+        MLMcpServerMetricsCounter.initialize("c", registry, featureFlag);
+        assertNotNull(MLMcpConnectorMetricsCounter.getInstance());
+        assertNotNull(MLMcpServerMetricsCounter.getInstance());
```
ylwu-amzn: Nit — these `assertNotNull` checks pass as long as `initialize` assigns *something*. A stronger assertion would be that two consecutive `getInstance()` calls return the same reference (singleton invariant), or that `incrementCounter` on the returned instance routes to the mock registry's counter (exercises the abstract base in this module, since JaCoCo is per-module as the PR comment notes).

Low-priority — call-site tests in other modules already exercise the routing path.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
rithinpullela: Will change histogram to have both

PR #4805, opened 2026-04-28
add tenantId in summarization manager

jiapingzeng:
### Description
Add `tenantId` in summarization manager's request so that it can call the model correctly in multi-tenant environment

### Related Issues

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.engine.algorithms.contextmanager;
 
 import static java.lang.Math.min;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
 import static org.opensearch.ml.common.FunctionName.REMOTE;
 import static org.opensearch.ml.common.utils.StringUtils.processTextDoc;
 import static org.opensearch.ml.engine.algorithms.agent.AgentUtils.LLM_RESPONSE_FILTER;
@@ -287,6 +288,14 @@ public class SummarizationManager implements ContextManager {
         return modelId;
     }
 
+    private String resolveTenantId(ContextManagerContext context) {
+        Map<String, String> parameters = context.getParameters();
+        if (parameters != null) {
+            return parameters.get(TENANT_ID_FIELD);
+        }
+        return null;
+    }
+
     protected void executeSummarization(
         ContextManagerContext context,
         String modelId,
@@ -305,8 +314,16 @@ public class SummarizationManager implements ContextManager {
             // Create ML input
             MLInput mlInput = MLInput.builder().algorithm(REMOTE).inputDataset(inputDataset).build();
 
+            // Propagate tenant_id from context so the prediction request passes multi-tenancy validation
+            String tenantId = resolveTenantId(context);
+
             // Create prediction request
-            MLPredictionTaskRequest request = MLPredictionTaskRequest.builder().modelId(modelId).mlInput(mlInput).build();
+            MLPredictionTaskRequest request = MLPredictionTaskRequest
+                .builder()
+                .modelId(modelId)
+                .mlInput(mlInput)
+                .tenantId(tenantId)
+                .build();
 
             // Execute prediction
             ActionListener<MLTaskResponse> listener = ActionListener.wrap(response -> {
@@ -378,8 +395,16 @@ public class SummarizationManager implements ContextManager {
             // Create ML input
             MLInput mlInput = MLInput.builder().algorithm(REMOTE).inputDataset(inputDataset).build();
 
+            // Propagate tenant_id from context so the prediction request passes multi-tenancy validation
+            String tenantId = resolveTenantId(context);
+
             // Create prediction request
-            MLPredictionTaskRequest request = MLPredictionTaskRequest.builder().modelId(modelId).mlInput(mlInput).build();
+            MLPredictionTaskRequest request = MLPredictionTaskRequest
+                .builder()
+                .modelId(modelId)
+                .mlInput(mlInput)
+                .tenantId(tenantId)
+                .build();
 
             // Execute prediction
             ActionListener<MLTaskResponse> listener = ActionListener.wrap(response -> {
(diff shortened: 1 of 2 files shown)


mingshl (approved):

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -287,6 +288,14 @@ private String resolveModelId(ContextManagerContext context) {
         return modelId;
     }
 
+    private String resolveTenantId(ContextManagerContext context) {
+        Map<String, String> parameters = context.getParameters();
+        if (parameters != null) {
+            return parameters.get(TENANT_ID_FIELD);
+        }
+        return null;
+    }
```
ylwu-amzn, line 297: `ContextManagerContext` already exposes a generic `getParameter(String key)` helper (see `ContextManagerContext.java:242`). You could skip this private helper entirely and call it at the use sites:

```java
String tenantId = (String) context.getParameter(TENANT_ID_FIELD);
```

Minor cast because `getParameter` returns `Object`, but avoids adding a second null-safe accessor that does the same job as the existing one (and as the nearby `resolveModelId` on line 280).

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -305,8 +314,16 @@ protected void executeSummarization(
...
+            String tenantId = resolveTenantId(context);
+
             // Create prediction request
-            MLPredictionTaskRequest request = MLPredictionTaskRequest.builder().modelId(modelId).mlInput(mlInput).build();
+            MLPredictionTaskRequest request = MLPredictionTaskRequest
+                .builder()
+                .modelId(modelId)
+                .mlInput(mlInput)
+                .tenantId(tenantId)
+                .build();
```
ylwu-amzn, line 326: This exact four-line pattern — `resolveTenantId` + builder with `tenantId(tenantId)` — is duplicated in `executeSummarizationForStructuredHistory` at line 398-407. Consider a small helper like `buildPredictionRequest(modelId, mlInput, context)` on this class so adding a future field (session_id? role?) doesn't silently drift between the two paths. Not blocking.

Also, the comment says "passes multi-tenancy validation" — the real reason tenant_id matters is routing correctness (without it, requests either get rejected or run against the wrong tenant's resources). Tightening the comment to say *why* helps future readers decide whether this is critical path or defence-in-depth.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -378,8 +395,16 @@ protected void executeSummarizationForStructuredHistory(
...
```
ylwu-amzn, line 407: Same duplication as the earlier site (line 317-326). If you introduce a helper there, apply it here too. The symmetry between the two paths is what matters — a single-site fix would leave this one still open to the bug.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManagerTest.java
```diff
@@ -608,6 +620,122 @@ public void testResolveModelIdFromConfig() {
...
+
+        addToolInteractionsToContext(20);
+
+        manager.execute(context);
+
+        ArgumentCaptor<MLPredictionTaskRequest> requestCaptor = ArgumentCaptor.forClass(MLPredictionTaskRequest.class);
+        verify(client).execute(any(ActionType.class), requestCaptor.capture(), any(ActionListener.class));
+
+        MLPredictionTaskRequest capturedRequest = requestCaptor.getValue();
+        Assert.assertNull(capturedRequest.getTenantId());
```
ylwu-amzn, line 694: This `assertNull` pins down the current behavior (no silent fallback), which is good. One strengthening suggestion: if someone later adds a "default tenant" fallback in `resolveTenantId`, a `null`→`"default"` change might pass or fail this test depending on the default value. Adding a comment above the assertion like `// Non-multi-tenant callers must not get a synthesized tenant id` makes the intent explicit.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManagerTest.java
```diff
@@ -608,6 +620,122 @@ public void testResolveModelIdFromConfig() {
...
+
+        manager.execute(context);
+
+        // Capture the request passed to client.execute()
+        ArgumentCaptor<MLPredictionTaskRequest> requestCaptor = ArgumentCaptor.forClass(MLPredictionTaskRequest.class);
+        verify(client).execute(any(ActionType.class), requestCaptor.capture(), any(ActionListener.class));
+
+        MLPredictionTaskRequest capturedRequest = requestCaptor.getValue();
+        Assert.assertEquals(expectedTenantId, capturedRequest.getTenantId());
+    }
```
ylwu-amzn, line 737: The three new tests share ~15 lines of identical mock setup (ThreadPool + ExecutorService + doAnswer-to-run-inline). Consider extracting to a `@Before`-populated helper or a small utility method — the test class is now over 750 lines and any future test exercising `client.execute` will want the same fixture.

Separately: these tests verify that `request.getTenantId()` carries the right value, which is exactly the right shape of unit test for this fix. They do **not** verify end-to-end that the multi-tenant path actually succeeds. If there's an existing multi-tenant integration test harness, adding one summarization case there would close the loop. Out of scope for this PR but worth flagging as follow-up.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/contextmanager/SummarizationManager.java
```diff
@@ -6,6 +6,7 @@
 package org.opensearch.ml.engine.algorithms.contextmanager;
 
 import static java.lang.Math.min;
+import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
```
ylwu-amzn, line 9: Nit: while you're here, the PR description has a placeholder `Resolves #[Issue number to be closed when this PR is merged]` still in it — replace with the actual issue or remove the line so the `Resolves` keyword doesn't confuse GitHub's auto-close.

PR #4808, opened 2026-04-30
feat: Add opt-in settings to enforce unique Agent and Agentic Memory Container names

rithinpullela:
### Description

Adds two opt-in, dynamic cluster settings, both default `false` for backward compatibility:
- `plugins.ml_commons.agent_name_uniqueness_enabled`
- `plugins.ml_commons.agentic_memory_name_uniqueness_enabled`

When enabled, registering an Agent or creating an Agentic Memory Container with a name that already exists in the same tenant is rejected with HTTP 409 CONFLICT. Default off preserves existing behavior.

**Semantics: best-effort, not atomic.** This matches the existing pattern in `MLModelGroupManager.validateUniqueModelGroupName` (term query on `name.keyword + tenantId`, then unconditional index write) and inherits the same race window 1:1 — two concurrent registrations with the same name can both observe zero hits and both succeed. Closing the race would require a sidecar lock index + `OpType.CREATE` on a deterministic `_id`; that hardening should be applied uniformly across model groups, agents, and memory containers and is intentionally left for a follow-up.

**What this PR adds beyond the model-group precedent:**
- Opt-in via dynamic cluster setting (model group enforces unconditionally).
- HTTP 409 instead of 400 `IllegalArgumentException`.
- Tenant validation runs **before** the uniqueness search, so in multi-tenant mode a missing tenant fails fast with 403 rather than leaking cross-tenant existence via a 409.
- For `PLAN_EXECUTE_AND_REFLECT` agents without a pre-existing `executor_agent_id`, the auto-created `"<name> (ReAct)"` executor agent's name is also validated so the derived agent can't collide with an existing one.
- `IndexNotFoundException` from the search is treated as "no duplicate possible" so first-ever registrations on a fresh cluster work.

### Manual verification

Verified end-to-end on a single-node cluster (`./gradlew :opensearch-ml-plugin:run`):

| # | Scenario | Result |
|---|----------|--------|
| 1 | Both flags surface via `_cluster/settings?include_defaults=true` as `false` | ✅ |
| 2 | Agent: flag OFF → register same name twice | ✅ both 200 (BWC preserved) |
| 3 | Agent: flag ON → register duplicate name | ✅ 409 with clear error |
| 4 | Agent: flag ON → register unique name | ✅ 200 |
| 5 | Agent: flip flag OFF → duplicate allowed (dynamic) | ✅ 200 |
| 6 | Memory container: flag OFF → create twice | ✅ both 200 |
| 7 | Memory container: flag ON → duplicate | ✅ 409 |
| 8 | Memory container: flag ON → unique | ✅ 200 |
| 9 | Memory container: flip flag OFF → duplicate | ✅ 200 |
| 10 | Cross-independence: flags gate only their own resource | ✅ |

### Related Issues
N/A

### Check List
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). _(No REST API contract change — only new cluster settings.)_

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
@@ -10,11 +10,9 @@ import static org.opensearch.ml.common.settings.MLCommonsSettings.ML_COMMONS_AGE
 
 import java.time.Instant;
 
-import org.opensearch.ExceptionsHelper;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.DocWriteResponse;
 import org.opensearch.action.index.IndexResponse;
-import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.action.support.WriteRequest;
@@ -23,9 +21,6 @@ import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.rest.RestStatus;
-import org.opensearch.index.IndexNotFoundException;
-import org.opensearch.index.query.BoolQueryBuilder;
-import org.opensearch.index.query.TermQueryBuilder;
 import org.opensearch.ml.common.memorycontainer.MLMemoryContainer;
 import org.opensearch.ml.common.memorycontainer.MemoryConfiguration;
 import org.opensearch.ml.common.memorycontainer.MemoryStrategy;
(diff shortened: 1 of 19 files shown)


plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -288,6 +306,69 @@ private void indexMemoryContainer(MLMemoryContainer container, ActionListener<St
         }
     }
 
+    /**
+     * When {@code plugins.ml_commons.agentic_memory_name_uniqueness_enabled} is enabled, reject
+     * creation if a memory container with the same name already exists in the same tenant. When the
+     * setting is disabled (default), this is a no-op so existing clusters remain backward compatible.
+     */
+    private void validateMemoryContainerNameUniqueness(String name, String tenantId, ActionListener<Void> listener) {
```
dhrubo-os, line 308: [nit] The checkAgentNameAvailable and validateMemoryContainerNameUniqueness methods have nearly identical logic (build query, search, check hits, return 409 or pass). There's about 50 lines of duplicated code 
between the two transport actions. This could be extracted into a shared utility

dhrubo-os:
what about update api?

PR #4816, opened 2026-05-07
Fix MCP agent execution hang after MCP SDK 1.1.1 bump

### LINKED ISSUE #4815: [BUG] Agent execution with MCP tools hangs indefinitely on latest code base
**What is the bug?**
Agent execution with MCP tools hangs indefinitely on latest code base.
This is working if we remove MCP connector from agent creation command.
Also it is working fine with OpenSearch 3.6(last released version)

Agent execution log with MCP connector:
Log with MCP:
[2026-05-06T20:29:08,345][INFO ][o.o.m.m.i.ConversationMetaIndex] [integTest-0] Successfully created the memory with id : PgbM_Z0BRhW8BFy3w6Gr
[2026-05-06T20:29:08,383][INFO ][o.o.m.m.a.c.CreateInteractionTransportAction] [integTest-0] Updating the memory PgbM_Z0BRhW8BFy3w6Gr after the message PwbM_Z0BRhW8BFy3w6HU is created
[2026-05-06T20:29:08,383][INFO ][o.o.m.m.i.InteractionsIndex] [integTest-0] Successfully created the message with id : PwbM_Z0BRhW8BFy3w6HU
[2026-05-06T20:29:08,416][INFO ][o.o.m.e.a.a.MLAgentExecutor] [integTest-0] Created parent interaction ID: PwbM_Z0BRhW8BFy3w6HU
[2026-05-06T20:29:11,125][INFO ][o.o.m.e.i.MLIndicesHandler] [integTest-0] Using initIndexWithoutThreadContext method to create index: .plugins-ml-model
[2026-05-06T20:29:11,125][INFO ][o.o.r.m.c.i.LocalClusterIndicesClient] [integTest-0] Searching [.plugins-ml-model] (edited)

Agent execution log without MCP connector:
[2026-05-06T20:28:19,310][INFO ][o.o.m.m.i.ConversationMetaIndex] [integTest-0] Successfully created the memory with id : OQbM_Z0BRhW8BFy3BKEd
[2026-05-06T20:28:19,351][INFO ][o.o.m.m.a.c.CreateInteractionTransportAction] [integTest-0] Updating the memory OQbM_Z0BRhW8BFy3BKEd after the message OgbM_Z0BRhW8BFy3BKFL is created
[2026-05-06T20:28:19,351][INFO ][o.o.m.m.i.InteractionsIndex] [integTest-0] Successfully created the message with id : OgbM_Z0BRhW8BFy3BKFL
[2026-05-06T20:28:19,388][INFO ][o.o.m.e.a.a.MLAgentExecutor] [integTest-0] Created parent interaction ID: OgbM_Z0BRhW8BFy3BKFL
[2026-05-06T20:28:19,390][INFO ][o.o.m.e.a.a.MLChatAgentRunner] [integTest-0] Starting chat agent execution. agentId=NQbK_Z0BRhW8BFy3vKE1, tenantId=null
[2026-05-06T20:28:21,103][INFO ][o.o.m.e.i.MLIndicesHandler] [integTest-0] Using initIndexWithoutThreadContext method to create index: .plugins-ml-model
[2026-05-06T20:28:21,104][INFO ][o.o.r.m.c.i.LocalClusterIndicesClient] [integTest-0] Searching [.plugins-ml-model]
[2026-05-06T20:28:21,109][INFO ][o.o.r.m.c.i.LocalClusterIndicesClient] [integTest-0] Search returned 1 hits hits
[2026-05-06T20:28:21,524][INFO ][o.o.m.m.a.c.CreateInteractionTransportAction] [integTest-0] Updating the memory OQbM_Z0BRhW8BFy3BKEd after the message OwbM_Z0BRhW8BFy3DKHC is created
[2026-05-06T20:28:21,528][INFO ][o.o.m.m.i.InteractionsIndex] [integTest-0] Successfully created the message with id : OwbM_Z0BRhW8BFy3DKHC
[2026-05-06T20:28:21,637][INFO ][o.o.m.m.a.c.UpdateInteractionTransportAction] [integTest-0] Successfully updated the Message with ID: OgbM_Z0BRhW8BFy3BKFL
[2026-05-06T20:28:21,637][INFO ][o.o.m.e.a.a.AgentUtils  ] [integTest-0] {"tenantId":null,"model_usage":{"model_name":"My OpenAI model: gpt-5 Test1","model_url":"removed-url-from-here","call_count":1,"total_tokens":666,"output_tokens":18,"model_id":"KgbG_Z0BRhW8BFy3GKEn","reasoning_tokens":0,"input_tokens":648,"cache_read_input_tokens":0},"eventTime":1778079501637}
[2026-05-06T20:28:21,640][INFO ][o.o.m.e.a.a.MLAgentExecutor] [integTest-0] Agent execution completed successfully. agentType=conversational, agentId=NQbK_Z0BRhW8BFy3vKE1, tenantId=null, latencyMs=2252

- Create MCP connector
- Register a model 
- Create a Conversational agent with above created MCP connector and model.
- Execute above created agent. - This will be in indefinite hung stage.

Execute the agent and give result to agent query

 - OS: MacOSx
 - Version OpenSearch 3.7
 - Plugins - MLCommons plugin

rithinpullela:
## Summary

Fixes opensearch-project/ml-commons#4815 — agents wired to MCP connectors hang indefinitely on latest main but work on 3.6 and on main without MCP.

## Root cause

Commit c9a4bd8636 (#4795) bumped the MCP SDK from 0.12.1 to 1.1.1. The new SDK uses `ServiceLoader` at two points:
1. `McpJsonMapperSupplier` — during `HttpClient*Transport.Builder.build()`.
2. `JsonSchemaValidatorSupplier` — during `McpClient.SyncSpec.build()`.

`ServiceLoader` uses the thread context classloader, which in OpenSearch server threads is the **core** classloader, not the **plugin** classloader where `mcp-json-jackson3` lives. Both lookups throw `ServiceConfigurationError` — an `Error`, not an `Exception` — so the existing `catch (Exception)` in `AgentUtils.getMCPToolSpecsFromConnector` did not catch it. The per-connector `AtomicInteger` counter in `getMcpToolSpecs` never reached zero → the final listener was never invoked → agent hung forever at `getMcpToolSpecs`, right before `Starting chat agent execution`.

## Fix

1. **`McpStreamableHttpConnectorExecutor` / `McpConnectorExecutor`** — bypass `ServiceLoader` entirely by passing explicit suppliers:
   - `.jsonMapper(new JacksonMcpJsonMapperSupplier().get())` on the transport builder.
   - `.jsonSchemaValidator(new JacksonJsonSchemaValidatorSupplier().get())` on the `McpClient.sync(transport)` chain.

2. **`AgentUtils.getMCPToolSpecsFromConnector`** — broaden the outer `catch (Exception)` to `catch (Throwable)` so future `Error`s always complete the listener rather than stranding the chain.

3. **`AgentUtils.getMcpToolSpecs`** — restructure the async counter so the decrement runs inside a `try/finally` on every path (success, filter exception, handled failure, synchronous `Throwable` from `getMCPToolSpecsFromConnector`). Extracted the decrement+complete logic into a `completeIfLast` runnable. This makes the final listener guaranteed to fire once per call — one bad connector can no longer hang the agent even if it throws an unexpected `Error`.

4. **`MLChatAgentRunner.processUnifiedTools`** — add `log.warn(..., e)` on the MCP-failure branch so future MCP failures are observable instead of silent.

## Test plan

- [x] Reproduced the hang locally against latest main (agent stuck indefinitely at `getMcpToolSpecs`, ServiceConfigurationError in logs, async counter never decremented).
- [x] Applied fix, rebuilt plugin, restarted local integTest cluster.
- [x] Registered Bedrock model + MCP streamable-http connector (`http://localhost:9900/mcp`) + chat agent with MCP.
- [x] Executed agent end-to-end: returned in ~11s with real MCP tool output (index list from `cat_indexes` tool). No `ServiceConfigurationError` in cluster logs.
- [x] Verified MCP client initialization log: `Server response with Protocol: 2025-11-25, Capabilities: ToolCapabilities[listChanged=false]`.

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -810,13 +810,6 @@ public class AgentUtils {
         AtomicInteger remainingConnectors = new AtomicInteger(mcpConnectorConfigs.size());
         List<MLToolSpec> finalToolSpecs = Collections.synchronizedList(new ArrayList<>());
 
-        // Guarantees final listener fires exactly once; every code path decrements via try/finally.
-        Runnable completeIfLast = () -> {
-            if (remainingConnectors.decrementAndGet() == 0) {
-                finalListener.onResponse(finalToolSpecs);
-            }
-        };
-
         // We make multiple Async calls in for loop, which happen in parallel
         for (Map<String, Object> mcpConnectorConfig : mcpConnectorConfigs) {
             String connectorId = (String) mcpConnectorConfig.get(MCP_CONNECTOR_ID_FIELD);
@@ -846,17 +839,28 @@ public class AgentUtils {
                     } catch (Throwable t) {
                         log.error("Error post-processing MCP tool specs for connector: " + connectorId, t);
                     } finally {
-                        completeIfLast.run();
+                        completeIfLast(remainingConnectors, finalToolSpecs, finalListener);
                     }
                 }, e -> {
                     log.error("Error processing connector: " + connectorId, e);
-                    completeIfLast.run();
+                    completeIfLast(remainingConnectors, finalToolSpecs, finalListener);
                 }));
             } catch (Throwable t) {
                 // Catch synchronous throws (including Errors) so one bad connector can't strand the counter.
                 log.error("Synchronous failure initiating MCP tool spec lookup for connector: " + connectorId, t);
-                completeIfLast.run();
+                completeIfLast(remainingConnectors, finalToolSpecs, finalListener);
             }
+        }
+    }
+
+    // Guarantees the final listener fires exactly once; every code path decrements via try/finally.
+    private static void completeIfLast(
+        AtomicInteger remainingConnectors,
+        List<MLToolSpec> finalToolSpecs,
+        ActionListener<List<MLToolSpec>> finalListener
+    ) {
+        if (remainingConnectors.decrementAndGet() == 0) {
+            finalListener.onResponse(finalToolSpecs);
         }
     }
 
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
@@ -810,42 +810,53 @@ public class AgentUtils {
         AtomicInteger remainingConnectors = new AtomicInteger(mcpConnectorConfigs.size());
         List<MLToolSpec> finalToolSpecs = Collections.synchronizedList(new ArrayList<>());
 
+        // Guarantees final listener fires exactly once; every code path decrements via try/finally.
+        Runnable completeIfLast = () -> {
+            if (remainingConnectors.decrementAndGet() == 0) {
+                finalListener.onResponse(finalToolSpecs);
+            }
+        };
+
         // We make multiple Async calls in for loop, which happen in parallel
         for (Map<String, Object> mcpConnectorConfig : mcpConnectorConfigs) {
             String connectorId = (String) mcpConnectorConfig.get(MCP_CONNECTOR_ID_FIELD);
             List<String> toolFilters = (List<String>) mcpConnectorConfig.get(TOOL_FILTERS_FIELD);
 
-            getMCPToolSpecsFromConnector(connectorId, tenantId, sdkClient, client, encryptor, ActionListener.wrap(mcpToolspecs -> {
-                List<MLToolSpec> filteredTools;
-                if (toolFilters == null || toolFilters.isEmpty()) {
-                    filteredTools = mcpToolspecs;
-                } else {
-                    filteredTools = new ArrayList<>();
-                    List<Pattern> compiledPatterns = toolFilters.stream().map(Pattern::compile).collect(Collectors.toList());
+            try {
+                getMCPToolSpecsFromConnector(connectorId, tenantId, sdkClient, client, encryptor, ActionListener.wrap(mcpToolspecs -> {
+                    try {
+                        List<MLToolSpec> filteredTools;
+                        if (toolFilters == null || toolFilters.isEmpty()) {
+                            filteredTools = mcpToolspecs;
+                        } else {
+                            filteredTools = new ArrayList<>();
+                            List<Pattern> compiledPatterns = toolFilters.stream().map(Pattern::compile).collect(Collectors.toList());
 
-                    for (MLToolSpec toolSpec : mcpToolspecs) {
-                        for (Pattern pattern : compiledPatterns) {
-                            if (pattern.matcher(toolSpec.getName()).matches()) {
-                                filteredTools.add(toolSpec);
-                                break;
+                            for (MLToolSpec toolSpec : mcpToolspecs) {
+                                for (Pattern pattern : compiledPatterns) {
+                                    if (pattern.matcher(toolSpec.getName()).matches()) {
+                                        filteredTools.add(toolSpec);
+                                        break;
+                                    }
+                                }
                             }
                         }
+
+                        finalToolSpecs.addAll(filteredTools);
+                    } catch (Throwable t) {
+                        log.error("Error post-processing MCP tool specs for connector: " + connectorId, t);
+                    } finally {
+                        completeIfLast.run();
                     }
-                }
-
-                finalToolSpecs.addAll(filteredTools);
-
-                // If this is the last connector, send the final response
-                if (remainingConnectors.decrementAndGet() == 0) {
-                    finalListener.onResponse(finalToolSpecs);
-                }
-            }, e -> {
-                log.error("Error processing connector: " + connectorId, e);
-                // Even on error, we need to check if this is the last connector
-                if (remainingConnectors.decrementAndGet() == 0) {
-                    finalListener.onResponse(finalToolSpecs);
-                }
-            }));
+                }, e -> {
+                    log.error("Error processing connector: " + connectorId, e);
+                    completeIfLast.run();
+                }));
+            } catch (Throwable t) {
+                // Catch synchronous throws (including Errors) so one bad connector can't strand the counter.
+                log.error("Synchronous failure initiating MCP tool spec lookup for connector: " + connectorId, t);
+                completeIfLast.run();
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


akolarkunnu (commented):
There are two failures in the CI from the relevant area:

McpConnectorExecutorTest > getMcpToolSpecs_returnsExpectedSpecs FAILED
    org.opensearch.ml.common.exception.MLException at McpConnectorExecutorTest.java:77
        Caused by: java.lang.NullPointerException at McpConnectorExecutorTest.java:77

McpStreamableHttpConnectorExecutorTest > getMcpToolSpecs_returnsExpectedSpecs FAILED
    org.opensearch.ml.common.exception.MLException at McpStreamableHttpConnectorExecutorTest.java:77
        Caused by: java.lang.NullPointerException at McpStreamableHttpConnectorExecutorTest.java:77

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -1695,7 +1695,7 @@ private void processUnifiedTools(
             processUnifiedToolsWithBackend(mlAgent, params, listener, memory, functionCalling, frontendTools, backendToolsMap, toolSpecMap);
         }, e -> {
             // Even if MCP tools fail, continue with base backend tools
-
+            log.warn("Failed to load MCP tools, continuing with backend tools only", e);
             Map<String, Tool> backendToolsMap = new HashMap<>();
```
akolarkunnu: Not introduced by this PR, duplicate code, could move to single method and invoke from different places :
```
Map<String, Tool> backendToolsMap = new HashMap<>();
            Map<String, MLToolSpec> toolSpecMap = new HashMap<>();
            createTools(toolFactories, params, backendToolSpecs, backendToolsMap, toolSpecMap, mlAgent);

            processUnifiedToolsWithBackend(mlAgent, params, listener, memory, functionCalling, frontendTools, backendToolsMap, toolSpecMap);
        }));
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -810,42 +810,53 @@ public static void getMcpToolSpecs(
         AtomicInteger remainingConnectors = new AtomicInteger(mcpConnectorConfigs.size());
         List<MLToolSpec> finalToolSpecs = Collections.synchronizedList(new ArrayList<>());
 
+        // Guarantees final listener fires exactly once; every code path decrements via try/finally.
+        Runnable completeIfLast = () -> {
```
akolarkunnu: Why do we need a runnable here, any specific reason ?
A simple private method will be good enough?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -810,42 +810,53 @@ public static void getMcpToolSpecs(
...
+                        completeIfLast.run();
+                    }
+                }, e -> {
+                    log.error("Error processing connector: " + connectorId, e);
+                    completeIfLast.run();
+                }));
+            } catch (Throwable t) {
+                // Catch synchronous throws (including Errors) so one bad connector can't strand the counter.
+                log.error("Synchronous failure initiating MCP tool spec lookup for connector: " + connectorId, t);
+                completeIfLast.run();
```
akolarkunnu: Problem with this approach is that, this kind of errors won't be noticing until some one looks to logs.
For eg: The current issue also won't be noticing by us until the user know there is a mismatch in expected result or check the logs explicitly.

It might be ok, since we were doing same thing for failure scenario:
```
e -> {
                    log.error("Error processing connector: " + connectorId, e);
                    completeIfLast.run();
                }));
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
rithinpullela: Kept it as runnable and not a private method because 
1. It is only used locally in this function
2. It only needs local variables and hence the caller sites are cleaner,  turning it into a private static would mean passing all three at every call site

Wouldn't mind switching if you prefer using a simple private method.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
rithinpullela: This is a wider question, probably out of scope of this PR. When the MCP connector fails, we generally do not fail the agent. Our design choice was to not add the MCP connector tools to the agent and we just log a error or warning line. 

The other option is to throw an exception when we cannot add an MCP connector successfully. I would like to defer that broader discussion to a different issue and keep this PR scope to fix the failing MCP connectors

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
akolarkunnu: Syntactically nothing wrong here. But, it's misleading. The javadoc of Runnable says with "The Runnable interface should be implemented by any class whose instances are intended to be executed by a thread. 

When an object implementing interface Runnable is used to create a thread, starting the thread causes the object's run method to be called in that separately executing thread."

And here it's nothing to do with thread.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentUtils.java
```diff
@@ -815,37 +815,52 @@ public static void getMcpToolSpecs(
...
-                } else {
-                    filteredTools = new ArrayList<>();
-                    List<Pattern> compiledPatterns = toolFilters.stream().map(Pattern::compile).collect(Collectors.toList());
-
-                    for (MLToolSpec toolSpec : mcpToolspecs) {
-                        for (Pattern pattern : compiledPatterns) {
-                            if (pattern.matcher(toolSpec.getName()).matches()) {
-                                filteredTools.add(toolSpec);
-                                break;
+            try {
```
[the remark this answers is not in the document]
rithinpullela: The outer try covers the synchronous call to `getMCPToolSpe`csFromConnector. Adding a finally there would call `completeIfLast` in addition to the success listener doing it asynchronously later, this will be a double decrement. The three existing decrement sites (success listener's finally, error listener, sync-throw catch) cover all paths exactly once each

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutor.java
```diff
@@ -88,21 +90,23 @@ public List<MLToolSpec> getMcpToolSpecs() {
                 getMcpRequestHeaders(builder);
             };
 
-            // Create transport
+            // Explicit jsonMapper() bypasses MCP SDK's ServiceLoader lookup, which fails under plugin classloader isolation.
             McpClientTransport transport = HttpClientSseClientTransport
                 .builder(mcpServerUrl)
+                .jsonMapper(new JacksonMcpJsonMapperSupplier().get())
```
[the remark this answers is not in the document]
rithinpullela: Addressed

nathaliellenaa (approved):

ylwu-amzn (approved):

PR #4817, opened 2026-05-07
Dynamic Header Support in ML Connectors

nathaliellenaa:
### Description
Enables runtime parameter substitution in connector headers using `${parameters.*}` placeholders, allowing per-request dynamic values (e.g., transaction IDs, request IDs, correlation IDs) to be passed in HTTP headers to external model services. 

Previously, connector headers only supported:
  - Static values defined at connector creation time
  - Credentials from AWS Secrets Manager via ${credential.secretArn.*}
  
This limitation prevented use cases where customers need to pass unique, per-request metadata in headers for tracing, debugging, and observability.

### Core Implementation
  
  1. Modified `AbstractConnector.createDecryptedHeaders()`
    - Removed parameter substitution at initialization time
    - Now only substitutes `${credential.*}` placeholders
    - Preserves `${parameters.*}` placeholders for runtime substitution
  
  2. Added `AbstractConnector.substituteHeadersWithRuntimeParameters()`
    - New method for per-request header substitution
    - Takes merged parameters (connector defaults + runtime from `_predict`)
    - Returns headers with `${parameters.*}` replaced with runtime values
  
  3. Added `Connector.getHeadersWithRuntimeParameters()` interface method
    - Clean abstraction for runtime header substitution
    - Default implementation returns static headers (backward compatible)
  
  4. Updated `ConnectorUtils.buildSdkRequest()` and `buildOKHttpStreamingRequest()`
    - Call runtime header substitution before building HTTP requests
    - Validate substituted headers for security issues

### Security Validations
  
  5. Added `RestActionUtils.validateHeaderSecurity()`
    - Blocks authorization-related headers from using `${parameters.*}`:
      - Authorization, Proxy-Authorization, Cookie
      - X-API-Key, X-Auth-Token, X-Auth-Header
      - Host, X-Forwarded-Host
    - Only allows `${credential.secretArn.*}` for sensitive headers
    - Enforced at connector creation/update time
  
  6. Added `ConnectorUtils.validateSubstitutedHeaders()`
    - Runtime validation of substituted header values:
      - Rejects CRLF injection (\r\n characters)
      - Rejects control characters (0x00-0x1F except tab)
      - Enforces per-header size limit (8KB)
      - Enforces total headers size limit (64KB)
  
  7. Updated Transport Actions
    - `TransportCreateConnectorAction`: Validate headers at creation
    - `TransportRegisterModelAction`: Validate inline connector headers
    - `UpdateConnectorTransportAction`: Validate headers at update
    - `UpdateModelTransportAction`: Validate inline connector headers at update

### Examples
  
**Before (Static Headers Only)**
```
 {
    ...
    "parameters": {
        "test_id": "default"
    },
    "actions": [{
        "headers": {
            "test_id": "${parameters.test_id}"
        }
    }]
 }
  
POST /_plugins/_ml/models/{model_id}/_predict
{
  "parameters": {
      "test_id": "id-12345"
  }
}
  
/ Result: Header sent = "default" ❌
```
  
**After (Dynamic Headers)**
```
// Same connector configuration
  
POST /_plugins/_ml/models/{model_id}/_predict
{
  "parameters": {
      "test_id": "id-12345"
  }
}
  
  // Result: Header sent = "id-12345" ✅
```

**Security: Blocked Headers**
```
{
    "actions": [{
        "headers": {
            "Authorization": "${parameters.token}" // ❌ Rejected
        }
    }]
}
  
{
    "error": {
        "root_cause": [
            {
                "reason": "Header 'Authorization' cannot use ${parameters.*} placeholders. Use ${credential.*} instead for sensitive values."
            }
        ],
        ...
    },
    "status": 400
}
```

### Related Issues

### Check List


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -527,4 +537,34 @@ private static void removeIfNull(JsonObject parent, String fieldName) {
...
+        int totalSize = 0;
+        for (Map.Entry<String, String> entry : headers.entrySet()) {
+            String value = entry.getValue();
+            if (value == null)
+                continue;
+
+            // Check for header injection
+            if (value.contains("\r") || value.contains("\n")) {
+                throw new IllegalArgumentException("Header value contains invalid characters (CR/LF)");
+            }
```
ylwu-amzn: **High-priority — control-character validation is narrower than the PR description claims.**

The PR description says: *"Rejects control characters (0x00-0x1F except tab)"*. The code only rejects `\r` and `\n`. Other control chars (0x00 NUL, 0x07 BEL, 0x1B ESC, …) pass through.

NUL embedded in a header value can confuse downstream HTTP libraries (some treat it as C-string terminator, some don't); ESC sequences in values that get echoed into operator logs enable terminal-control / log-injection attacks. Either implement what the description says — e.g.:

```java
for (int i = 0; i < value.length(); i++) {
    char c = value.charAt(i);
    if (c < 0x20 && c != '\t') {
        throw new IllegalArgumentException("Header value contains invalid control character");
    }
}
```

…or update the PR description to say only CR/LF are blocked. I'd pick the former — it's a one-loop change and matches the stated intent.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -377,6 +379,10 @@ public static SdkHttpFullRequest buildSdkRequest(
             .uri(uri)
             .contentStreamProvider(requestBody.contentStreamProvider());
         Map<String, String> headers = connector.getDecryptedHeaders();
+        if (connector instanceof AbstractConnector) {
+            headers = ((AbstractConnector) connector).substituteHeadersWithRuntimeParameters(headers, parameters);
+            validateSubstitutedHeaders(headers);
+        }
```
ylwu-amzn: **High-priority — `instanceof`-gated validation skips security checks for any non-`AbstractConnector` impl.**

The PR description says a `Connector.getHeadersWithRuntimeParameters()` interface method was added with a default impl that returns static headers. There is no such method on `Connector` — dispatch is via this `instanceof AbstractConnector` cast. Today every concrete connector extends `AbstractConnector`, so it works, but:

1. Any future `Connector` impl that doesn't extend `AbstractConnector` silently gets *no* substitution **and** *no* `validateSubstitutedHeaders` call. The CRLF / 8KB / 64KB checks are gated on the same `instanceof`, which is the wrong default for a security-relevant check.
2. The description and the code disagree, which makes review and audit harder.

Two cleaner shapes:
- **Add the interface method** as the description says: `default Map<String,String> getHeadersWithRuntimeParameters(Map<String,String> runtimeParameters) { return getDecryptedHeaders(); }` on `Connector`, override in `AbstractConnector` to do the substitution. Then call it unconditionally here.
- Or at minimum, hoist `validateSubstitutedHeaders(headers)` out of the `if` so it runs for *every* non-null `headers` map regardless of connector type.

Same issue on the streaming path at line 423-426.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -377,6 +379,10 @@ public static SdkHttpFullRequest buildSdkRequest(
             .uri(uri)
             .contentStreamProvider(requestBody.contentStreamProvider());
         Map<String, String> headers = connector.getDecryptedHeaders();
```
ylwu-amzn: **Medium — MCP connectors aren't covered by the new substitution path; UX is now inconsistent.**

`McpConnectorExecutor:82` and `McpStreamableHttpConnectorExecutor:85` read headers via `connector.getDecryptedHeaders()` directly and never call `substituteHeadersWithRuntimeParameters`. Pre-PR, that was internally consistent because `createDecryptedHeaders` did do `${parameters.*}` substitution at init time. Post-PR, the parent class no longer does that, but the MCP overrides also don't substitute, so MCP connectors ship `${parameters.foo}` as a literal placeholder in headers.

Meanwhile `validateHeaderSecurity` runs on *all* connectors at create time regardless of protocol — so an MCP connector with `Authorization: ${parameters.token}` is rejected up front, but `X-Trace-Id: ${parameters.trace_id}` is accepted and then quietly broken at predict time. Confusing for users.

Two reasonable resolutions:
- Apply `substituteHeadersWithRuntimeParameters` inside the MCP executors too (pass the merged parameters in).
- Or reject any `${parameters.*}` in MCP-protocol headers at create time with a clearer error ("dynamic headers are not yet supported for MCP connectors").

Either way please document the asymmetry in the connector docs.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -384,4 +387,40 @@ public static void putMcpRequestHeaders(RestRequest request, Client client) {
...
+     * @throws IllegalArgumentException if validation fails
+     */
+    public static void validateHeaderSecurity(Map<String, String> headers) {
+        if (headers == null || headers.isEmpty()) {
+            return;
+        }
+
+        // Headers that should NOT allow ${parameters.*} substitution
+        Set<String> blockedHeaders = Set
+            .of("authorization", "proxy-authorization", "cookie", "x-api-key", "x-auth-token", "x-auth-header", "host", "x-forwarded-host");
```
ylwu-amzn: **Medium — denylist of blocked header names is incomplete and unextensible.**

A few defense-in-depth observations on the blocked-headers list:

- It's a denylist, so any new auth-bearing header (custom `X-Tenant-Auth`, `X-Service-Token`, future auth schemes) is allowed by default. That's the wrong default for a security policy.
- `X-Forwarded-For` isn't on the list — runtime control of that lets a tenant spoof its source IP to a downstream model service that authorizes by IP.
- No way for an operator to extend this set without forking the plugin.

Suggestions, in increasing effort:
1. At minimum, add `x-forwarded-for` and a comment explaining the intended scope ("this is auth-bearing headers we know about; not exhaustive").
2. Cluster setting like `plugins.ml_commons.connector.dynamic_header.blocked_header_names` that appends to the built-in set.
3. Invert to an allowlist of header *name patterns* that may use `${parameters.*}` (e.g., `X-Request-Id`, `X-Trace-*`, `X-Correlation-Id`).

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Added the validation
```
for (int i = 0; i < value.length(); i++) {
    char c = value.charAt(i);
    if (c < 0x20 && c != '\t') {
        throw new IllegalArgumentException("Header value contains invalid control character");
    }
}
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Added interface method `getHeadersWithRuntimeParameters` and call it unconditionally

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
nathaliellenaa: Added `x-forwarded-for` and 7 other headers (16 total) covering IP spoofing and SSRF vectors.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Implemented the second option to reject any `${parameters.*}` in MCP connector headers at creation time. 

MCP connectors only perform one-time tool discovery during agent registration, not per-request invocations. Headers are applied once when calling `getMcpToolSpecs()` to fetch available tools. Since runtime parameters don't exist at registration time and are never used for subsequent agent executions, allowing `${parameters.*}` would create a misleading user experience where headers appear to support dynamic values but actually remain static or unsubstituted.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -376,7 +377,8 @@ public static SdkHttpFullRequest buildSdkRequest(
             .method(method)
             .uri(uri)
             .contentStreamProvider(requestBody.contentStreamProvider());
-        Map<String, String> headers = connector.getDecryptedHeaders();
+        Map<String, String> headers = connector.getHeadersWithRuntimeParameters(parameters);
+        validateSubstitutedHeaders(headers);
```
ylwu-amzn, line 381: **Blocking — deny-list isn't enforced at request time.**

`RestActionUtils.validateConnectorHeaders` runs only at the create/update transport actions. A connector that already exists on the cluster (created before this PR, restored from a snapshot, or written via a direct index call) can still have `${parameters.*}` in `Authorization`/`Cookie`/`Host`/`X-Forwarded-For`/etc.

Before this PR, `${parameters.*}` in headers was substituted at decrypt time using only `connector.parameters`, so the `_predict` caller could not influence those values. After this PR, `parameters` passed in here is the merged map (connector defaults overlaid by the `_predict` body), so a caller doing `{"parameters":{"token":"attacker-value"}}` will now overwrite the `Authorization` template on a pre-existing vulnerable connector. That's a privilege escalation relative to prior behavior.

Mitigation: re-enforce the deny-list at request time — either re-run `validateConnectorHeaders(connector.getDecryptedHeaders(), connector.getProtocol())` here / in `getHeadersWithRuntimeParameters`, or short-circuit substitution for blocked-list headers.

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
```diff
@@ -86,13 +86,27 @@ protected Map<String, String> createDecryptedHeaders(Map<String, String> headers
...
+        if (headers == null || headers.isEmpty()) {
+            return headers;
+        }
+
+        Map<String, String> substitutedHeaders = new HashMap<>();
+        StringSubstitutor substitutor = new StringSubstitutor(runtimeParameters, "${parameters.", "}");
+
+        for (Map.Entry<String, String> entry : headers.entrySet()) {
+            String value = entry.getValue();
+            if (value != null && value.contains("${parameters.")) {
```
ylwu-amzn, line 130: **Blocking — unresolved `${parameters.*}` is silently sent as the literal header value.**

When the runtime parameter is missing, the placeholder string passes through `substitutor.replace(...)` untouched and ships as-is to the remote service. `Connector.validatePayload` already throws `IllegalArgumentException` for the same case in payloads — the header path should match.

This also interacts badly with the `Content-Type`/`Content-Length` defaulting in `ConnectorUtils.buildSdkRequest` (~lines 387–392): `matchingHeaders("Content-Type").isEmpty()` returns false when the literal placeholder is present, so the JSON default-fallback never kicks in and the request goes out with `Content-Type: ${parameters.content_type}`.

Suggest: after the substitution loop, scan the resulting map for any value still matching `\$\{parameters\.[^}]+\}` and throw, mirroring `validatePayload`.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -121,6 +124,12 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
                         boolean hasPermission = connectorAccessControlHelper.validateConnectorAccess(client, connector);
                         if (hasPermission) {
                             connector.update(mlUpdateConnectorAction.getUpdateContent());
+                            if (connector.getActions() != null) {
+                                for (ConnectorAction action : connector.getActions()) {
+                                    Map<String, String> headers = action.getHeaders();
+                                    RestActionUtils.validateConnectorHeaders(headers, connector.getProtocol());
+                                }
+                            }
```
ylwu-amzn: **High-priority — pre-existing connectors with grandfathered bad headers can no longer be updated.**

This validates *all* actions on the connector after `connector.update(...)`, not just actions whose headers were changed by this update. A user with a connector created before this PR that has e.g. `Authorization: Bearer ${parameters.token}` will now hit `IllegalArgumentException` on every unrelated update (renaming, changing the description, adjusting an unrelated action), with no clean recovery path short of delete-and-recreate.

Same shape in `UpdateModelTransportAction.java:344`. At minimum, log a warning on legacy bad headers instead of failing; ideally only validate actions whose headers actually changed in this update.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtilsTest.java
```diff
@@ -1347,4 +1347,136 @@ public void processOutput_WithProcessorChain_StringOutput() throws IOException {
         assertEquals(1, tensors.getMlModelTensors().size());
         assertEquals("response", tensors.getMlModelTensors().get(0).getName());
     }
+
+    @Test
+    public void testValidateSubstitutedHeaders_ValidHeaders() {
```
ylwu-amzn, line 1352: **Test coverage gap.** The new `validateSubstitutedHeaders` helper is well-covered, but no test exercises the integration path: a real (non-mock) connector with `X-Request-ID: ${parameters.request_id}` → `_predict` with `request_id=abc` → `buildSdkRequest` → the resulting `SdkHttpFullRequest` has the substituted value.

The existing `buildSdkRequest_*` tests in this file mock `connector.getDecryptedHeaders()` (lines 1070, 1094, 1115, 1135), so they bypass `getHeadersWithRuntimeParameters` entirely. Add at least one end-to-end test that asserts the substitution actually flows into the request the SDK sends. (Same gap exists for `buildOKHttpStreamingRequest`.)

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -527,4 +530,37 @@ private static void removeIfNull(JsonObject parent, String fieldName) {
...
+        int totalSize = 0;
+        for (Map.Entry<String, String> entry : headers.entrySet()) {
+            String value = entry.getValue();
+            if (value == null)
+                continue;
+
+            // Check for header injection
+            for (int i = 0; i < value.length(); i++) {
+                char c = value.charAt(i);
+                if (c < 0x20 && c != '\t') {
```
ylwu-amzn: Minor: `c < 0x20 && c != '\t'` catches NUL through US but misses DEL (`0x7F`), which RFC 7230 also excludes from `field-value` and which some HTTP stacks handle oddly. Cheap to harden: `(c < 0x20 && c != '\t') || c == 0x7F`.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -527,4 +530,37 @@ private static void removeIfNull(JsonObject parent, String fieldName) {
...
+                    throw new IllegalArgumentException("Header value contains invalid control character");
+                }
+            }
+
+            // Check per-header size
+            if (value.length() > 8192) {
+                throw new IllegalArgumentException("Header value exceeds 8KB limit");
+            }
+
+            totalSize += entry.getKey().length() + value.length();
```
ylwu-amzn: Inconsistency: the per-header check on line 554 limits `value.length() > 8192` (value-only), but `totalSize` here sums `key.length() + value.length()`. Either include the key in the per-header check, or document that the 8KB limit is value-only while the 64KB limit is key+value combined.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -384,4 +410,50 @@ public static void putMcpRequestHeaders(RestRequest request, Client client) {
...
+     * @param headers Connector headers to validate
+     * @param connectorProtocol The connector protocol (e.g., "http", "mcp_sse", "mcp_streamable_http")
+     * @throws IllegalArgumentException if validation fails
+     */
+    public static void validateConnectorHeaders(Map<String, String> headers, String connectorProtocol) {
+        if (headers == null || headers.isEmpty()) {
+            return;
+        }
+
+        boolean isMcpProtocol = connectorProtocol != null && connectorProtocol.toLowerCase(Locale.ROOT).startsWith("mcp");
```
ylwu-amzn: Nit: `connectorProtocol.toLowerCase(Locale.ROOT).startsWith("mcp")` will silently apply the dynamic-header restriction to any future `mcp_*` protocol. If a future MCP variant should support dynamic headers, this string-prefix match is hard to find and easy to forget.

Consider an explicit set built from `ConnectorProtocols.MCP_SSE`/`MCP_STREAMABLE_HTTP`, e.g. `Set.of(MCP_SSE, MCP_STREAMABLE_HTTP).contains(connectorProtocol.toLowerCase(...))`.

common/src/test/java/org/opensearch/ml/common/connector/HttpConnectorTest.java
```diff
@@ -683,4 +683,126 @@ public void testFindAction_MultipleActionsWithSameType() {
...
+        Map<String, String> headers = new HashMap<>();
+        headers.put("X-Request-ID", "${parameters.request_id}");
+
+        Map<String, String> runtimeParameters = new HashMap<>();
+        runtimeParameters.put("request_id", "default-value");
+
+        Map<String, String> result = connector.substituteHeadersWithRuntimeParameters(headers, runtimeParameters);
+
+        Assert.assertEquals("default-value", result.get("X-Request-ID"));
+    }
```
ylwu-amzn: Nit: this test is named `_WithDefaultValue` but doesn't exercise any "default value" concept — it just passes `request_id=default-value` as a runtime parameter, which is functionally identical to `_WithRuntimeValue` above. Either rename, or expand to test the connector-default fallback (which is the actually interesting case: `RemoteConnectorExecutor.preparePayloadAndInvoke` merges `connector.parameters` into the runtime parameter map, so a connector default should fill in when `_predict` doesn't pass the param).

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Re-enforced at request time in `getHeadersWithRuntimeParameters`

common/src/main/java/org/opensearch/ml/common/connector/AbstractConnector.java
nathaliellenaa: Fixed it by throwing error for unresolved `${parameters.*}`

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
nathaliellenaa: Fixed it by only validate headers if actions were modified in the update request

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
nathaliellenaa: Included the key + value in both per-header check and total size

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
nathaliellenaa: Created explicit set
```
private static final Set<String> MCP_PROTOCOLS = Set.of(ConnectorProtocols.MCP_SSE, ConnectorProtocols.MCP_STREAMABLE_HTTP);
```

ylwu-amzn (approved):

PR #4820, opened 2026-05-11
Feature/agent level token limit

### LINKED ISSUE #4728: [FEATURE] Agent-level Token Limit parameter

Currently, there is no way to enforce a hard limit on the total number of tokens consumed during an agent execution in ML Commons.

The only available control is `parameters.max_iteration`, which indirectly limits usage by restricting the number of steps an agent can take. However, this is not sufficient because token consumption can vary significantly between iterations depending on the task, prompt size, and intermediate outputs. As a result, token usage is unpredictable and cannot be reliably bounded.

While `parameters.max_tokens` can be configured at the model level (`/_plugins/_ml/models`), this setting does not apply to agent executions as a whole.

This becomes a critical limitation for:
- Cost control and billing predictability
- Multi-tenant environments
- Production safety and resource governance

Without a way to cap total token usage per agent execution, it is difficult to safely operate agents in cost-sensitive or large-scale environments.

A new parameter at the agent level:

```json
{
  "parameters": {
    "max_tokens": <max_tokens_for_agent_execution>
  }
}
```

With the recent addition of cumulative token tracking across multiple LLM calls, this feature can be implemented in a straightforward way.

The agent execution loop can dynamically adjust the `max_tokens` / `maxTokens` passed to each LLM call based on the remaining token budget:

```json
{
  "parameters": {
    "max_tokens": "<max_tokens_for_agent_execution> - <totalTokensConsumedSoFar>"
  }
}
```

This would:
- Enforce a hard upper bound on total token usage
- Work seamlessly across multiple LLM calls within a single agent execution
- Provide deterministic and predictable cost control

If the remaining token budget reaches zero, the agent execution can terminate gracefully with an appropriate error or partial result.

The only current alternative is `parameters.max_iteration`, but it is too imprecise to be useful for controlling token usage:
- Token consumption per iteration varies widely
- There is no direct correlation between iterations and cost
- It cannot guarantee an upper bound on total tokens

**Additional context**

Recent PR introducing cumulative token counting across LLM calls:
https://github.com/opensearch-project/ml-commons/pull/4683/files

GugaGlonti:
### Description
This PR introduces agent-level token budget support for agent executions using `parameters.max_tokens`.

Before this change, agent executions did not enforce a cumulative token budget across LLM calls. This PR adds token usage aggregation and budget enforcement for V1 chat, V2 chat, and Plan-Execute-Reflect agents.

The implementation:
- Tracks cumulative token usage across agent LLM calls.
- Treats positive `parameters.max_tokens` as the total agent execution budget.
- Stops agent execution when the cumulative token budget is exhausted.
- Clamps outgoing LLM request parameters to the remaining token budget before each LLM call.
- Preserves smaller model-level per-call output limits when present.
- Applies provider-compatible output limit fields:
  - `max_tokens`
  - `maxTokens`
  - `maxOutputTokens`
- Preserves existing token usage response behavior, including sub-agent usage merging for Plan-Execute-Reflect.

### Related Issues
Resolves #4728

### Check List

### Execute Agent request with parameter `"max_tokens": 10000`

```json
POST _plugins/_ml/agents/8Tp1JZ4Bd954a9UWJGmU/_execute?async=true
{
  "parameters": {
    "question": "How many flights from Beijing to Seattle?",
    "include_token_usage": true,
    "max_tokens": 10000
  }
}
```

### Response: ML-Task terminating after detecting token-overexhaustion

```
{
  "task_type": "AGENT_EXECUTION",
  "function_name": "AGENT",
  "state": "COMPLETED",
  "worker_node": [
    "dLXu2EpiREqk64AL7dAC3w"
  ],
  "create_time": 1778745110465,
  "last_update_time": 1778745122576,
  "is_async": true,
  "response": {
    "memory_id": "KDp4JZ4Bd954a9UWb2rC",
    "executor_agent_memory_id": "Ljp4JZ4Bd954a9UWeGrr",
    "inference_results": [
      {
        "output": [
          {
            "result": "KDp4JZ4Bd954a9UWb2rC",
            "name": "memory_id"
          },
          {
            "result": "KTp4JZ4Bd954a9UWb2rX",
            "name": "parent_interaction_id"
          },
          {
            "result": "Ljp4JZ4Bd954a9UWeGrr",
            "name": "executor_agent_memory_id"
          },
          {
            "result": "Rjp4JZ4Bd954a9UWmmqm",
            "name": "executor_agent_parent_interaction_id"
          },
          {
            "name": "response",
            "dataAsMap": {
              "response": """Agent execution stopped: token budget exhausted (10197/10000 tokens used). Completed steps: [
<step-1>
Use ListIndexTool to identify all available indices in the OpenSearch cluster without filtering to discover what indices exist
</step-1>
, 
<step-1-result>
## Available Indices in OpenSearch Cluster

I've successfully retrieved all indices from the OpenSearch cluster. Here are all 16 available indices:

### System/Plugin Indices:
1. **`.plugins-ml-memory-message`** - green, open, 15 docs (4 deleted)
2. **`.plugins-ml-model-group`** - green, open, 5 docs (5 deleted)
3. **`.kibana_-152937574_admintenant_1`** - green, open, 2 docs
4. **`.plugins-ml-memory-meta`** - green, open, 8 docs (15 deleted)
5. **`.ql-datasources`** - green, open, 0 docs
6. **`.plugins-ml-agent`** - green, open, 4 docs (6 deleted)
7. **`.plugins-ml-task`** - green, open, 11 docs (11 deleted)
8. **`.plugins-ml-connector`** - green, open, 6 docs
9. **`.opendistro_security`** - green, open, 9 docs
10. **`.kibana_1`** - green, open, 0 docs
11. **`.plugins-ml-config`** - green, open, 1 doc
12. **`.plugins-ml-model`** - green, open, 5 docs (2 deleted)
13. **`.plugins-ml-context-management-templates`** - green, open, 3 docs

### User/Application Indices:
14. **`kb-sops-canonical`** - yellow, open, 0 docs (1 replica)
15. **`security-auditlog-2026.05.14`** - green, open, 162 docs
16. **`top_queries-2026.05.14-84928`** - green, open, 21 docs (9 deleted)

**Summary:** The cluster contains mostly plugin and system indices (13), with 3 user-facing indices. Most indices are in green health status, with one index (`kb-sops-canonical`) in yellow status. Total documents across all indices: 269 documents.
<step-traces>

</step-traces>

</step-1-result>
, 
<step-2>
Use IndexMappingTool with index pattern '*' to examine field mappings across all indices and identify which index contains flight data with origin and destination fields
</step-2>
, 
<step-2-result>
Agent execution stopped: token budget exhausted (2131/1934 tokens used)
<step-traces>

</step-traces>

</step-2-result>
]"""
            }
          },
          {
            "name": "token_usage",
            "dataAsMap": {
              "per_turn_usage": [
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "total_tokens": 1395,
                  "output_tokens": 149,
                  "turn": 1,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 1246,
                  "cache_read_input_tokens": 0
                },
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "total_tokens": 1574,
                  "output_tokens": 55,
                  "turn": 2,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 1519,
                  "cache_read_input_tokens": 0
                },
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "total_tokens": 2975,
                  "output_tokens": 518,
                  "turn": 3,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 2457,
                  "cache_read_input_tokens": 0
                },
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "total_tokens": 2122,
                  "output_tokens": 123,
                  "turn": 4,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 1999,
                  "cache_read_input_tokens": 0
                },
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "total_tokens": 2131,
                  "output_tokens": 59,
                  "turn": 5,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 2072,
                  "cache_read_input_tokens": 0
                }
              ],
              "per_model_usage": [
                {
                  "model_name": "Bedrock Claude Sonnet model",
                  "model_url": "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-haiku-4-5-20251001-v1:0/converse",
                  "call_count": 5,
                  "total_tokens": 10197,
                  "output_tokens": 904,
                  "model_id": "7Dp0JZ4Bd954a9UW_mkX",
                  "cache_creation_input_tokens": 0,
                  "input_tokens": 9293,
                  "cache_read_input_tokens": 0
                }
              ]
            }
          }
        ]
      }
    ],
    "parent_interaction_id": "KTp4JZ4Bd954a9UWb2rX",
    "executor_agent_parent_interaction_id": "Lzp4JZ4Bd954a9UWeGr-"
  }
}
```


ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -593,6 +650,9 @@ private void executePlanningLoop(
                 }
                 // Mark sub-agent so its token tracker suppresses logging (parent logs merged totals)
                 reactParams.put(AgentTokenTracker.IS_SUB_AGENT_FIELD, "true");
+                if (tokenBudget.isLimited()) {
+                    reactParams.put(AgentTokenBudget.MAX_TOKENS_FIELD, Long.toString(tokenBudget.remaining(tokenTracker)));
```
rithinpullela: **C1 (Critical) — sub-agent budget bypass.**

When `tokenBudget.remaining(tokenTracker)` is `0` (parent at its cap but execution didn't terminate on the prior `isExhausted` check at line 497), this writes `max_tokens="0"` into `reactParams`. The child then calls `AgentTokenBudget.fromExecutionParams`, which at `AgentTokenBudget.java:45` requires `value > 0` and otherwise returns `unlimited()` — so the child runs **with no budget at all**.

The guard `tokenBudget.isLimited()` only protects against the unlimited case; it doesn't protect against the zero-remaining case.

Suggested fix (smallest change): change the parser to treat `0` as exhausted rather than unlimited:
```java
// AgentTokenBudget.java:44-47
long value = Long.parseLong(maxTokensStr);
if (value > 0) {
    return new AgentTokenBudget(value);
}
if (value == 0) {
    return new AgentTokenBudget(0L); // hard stop; isExhausted always true
}
```
Alternatively, guard here: `if (tokenBudget.isLimited() && tokenBudget.remaining(tokenTracker) > 0)` and short-circuit the sub-agent dispatch when remaining is 0.

rithinpullela (commented):
Below is a review generated by AI with my oversight. 

Thanks for the PR — the core design is clean, backwards compatibility is preserved when `max_tokens` is unset, and the three exhaustion-check sites per runner make the post-call overrun bounded.

I reviewed the diff against `main` (base `be86ad045`, head `af3fc6b00`) and left inline comments for the findings below. One of them (C1) is a genuine sub-agent budget bypass and is the main thing I'd like addressed before merge.

### Critical
- **C1** — PER sub-agent receives `max_tokens="0"` when parent's remaining budget is 0; the child parses `0` as *unlimited*, bypassing the budget. See inline comment on `MLPlanExecuteAndReflectAgentRunner.java:654`.

### Important
- **I1** — Dead `getMaxTokensBudget()` helpers (×3); inline on each.
- **I2** — `fromExecutionParams` fails open on every parse error (silent for 0/negative/overflow; WARN-then-unlimited for non-numeric). Inverted default for a safety feature. See `AgentTokenBudget.java:45`.
- **I3** — V2 emits `stop_reason="max_tokens"`; V1 and PER don't expose any structured stop reason. Callers must string-match. See `MLChatAgentRunnerV2.java:56`.
- **I6** — `AgentTokenTracker.mergeSubAgentUsage` has unchecked casts (pre-existing lines 166, 171-172, 184; not in this PR's diff). The class already has a defensive `getLongFromMap` pattern at line 198. See inline on the new `getCumulativeTotalTokens` for a suggested symmetric helper. A malformed or adversarial sub-agent response would trigger `ClassCastException` with no try/catch in the planner-loop stream.
- **I8** — Significant duplication of the "check `isExhausted` → build exhausted message → short-circuit" block: ~4 copies in `MLChatAgentRunner` (lines 594, 865, 1509, 1606), ~5 in `MLPlanExecuteAndReflectAgentRunner` (lines 497, 582, 756, 1080, 1122), ~2 in `MLChatAgentRunnerV2` (179, 243). Each site is a maintenance hazard — e.g., adding a stop-reason field (I3) requires updating every copy. Extract a helper per runner family.

### Scope note (C2)
The budget applies only to the agent's direct LLM calls and PER sub-agents. LLM calls made via `AgentTool` / `MLModelTool` are *not* tracked and *not* bounded. This is probably the right scope given how rarely those tools are used, but it's worth stating explicitly in the PR description and in `AgentTokenBudget` Javadoc so operators don't assume a harder cap than the feature provides.

### Nits (non-blocking)
- `MLChatAgentRunnerV2.java:56` — `STOP_REASON_MAX_TOKENS = "max_tokens"` literal collides with the parameter key of the same name. Consider `"budget_exhausted"`.
- `AgentTokenBudget.java:178` — reaches back into `MLChatAgentRunner.LLM_INTERFACE` constant; a shared `AgentUtils` constant would avoid the circular knowledge.
- `AgentTokenBudget.java:184-188` — `contains("bedrock")` / `contains("gemini")` can misroute custom interface strings like `"my-gemini-proxy"`. Consider allow-listing.
- `AgentTokenTracker.getCumulativeTotalTokens()` could be package-private — only consumed inside the agent package.

Review was done with multiple parallel passes and cross-verified before posting. Happy to clarify any of these.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,192 @@
...
+    static AgentTokenBudget fromExecutionParams(Map<String, String> params) {
+        if (params == null) {
+            return unlimited();
+        }
+
+        String maxTokensStr = params.get(MAX_TOKENS_FIELD);
+        if (maxTokensStr != null) {
+            try {
+                long value = Long.parseLong(maxTokensStr);
+                if (value > 0) {
```
rithinpullela: **I2 (Important) — fails open on all parse errors.**

Current behavior of `fromExecutionParams`:
- `null` / missing → `unlimited()` (silent, correct)
- **negative** → `unlimited()` (silent, no log)
- **zero** → `unlimited()` (silent, no log — and see C1 for the bypass this enables)
- **non-numeric** (`"abc"`, `"100.5"`, `"1e10"`, `" 100 "` with whitespace) → WARN log, then `unlimited()`
- **Long overflow** (`"9999999999999999999"`) → `NumberFormatException`, WARN log, `unlimited()`

For a cost/safety feature this default is inverted: a user who typed `max_tokens=100.5` clearly intended to cap something and now has no cap. Recommend rejecting the request (throw `IllegalArgumentException`) on explicit parse failure or non-positive input. Keep fail-open only when the key is genuinely absent.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
```diff
@@ -51,6 +53,7 @@ public class MLChatAgentRunnerV2 extends AbstractV2AgentRunner {
 
     // Chat-specific constants
     private static final String STOP_REASON_MAX_ITERATIONS = "max_iterations";
+    private static final String STOP_REASON_MAX_TOKENS = "max_tokens";
```
rithinpullela: **I3 (Important) — inconsistent stop-reason surface across runners.**

V2 threads `STOP_REASON_MAX_TOKENS` through to `AgentLogicResult` (programmatically detectable). V1 (`MLChatAgentRunner`) and PER (`MLPlanExecuteAndReflectAgentRunner`) have no structured stop-reason — callers must string-match the exhausted message to distinguish "budget exhausted" from a normal `end_turn`.

Consider adding a `stop_reason` field to V1's `sendFinalAnswer` and PER's `saveAndReturnFinalResult` (additive change in the output `ModelTensor` dataAsMap, fully backwards compatible).

Minor nit on the constant itself: `STOP_REASON_MAX_TOKENS = "max_tokens"` collides lexically with the parameter key of the same name (`AgentTokenBudget.MAX_TOKENS_FIELD`). Consider `"budget_exhausted"` or `"token_budget"` to avoid confusion for downstream consumers.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -691,6 +751,25 @@ private void executePlanningLoop(
...
+                    addSteps(completedSteps, allParams, COMPLETED_STEPS_FIELD);
+
+                    if (tokenBudget.isExhausted(tokenTracker)) {
+                        boolean includeTokenUsage = Boolean.parseBoolean(allParams.getOrDefault(AgentUtils.INCLUDE_TOKEN_USAGE, "false"));
+                        saveAndReturnFinalResult(
+                            memory,
+                            parentInteractionId,
+                            allParams.get(EXECUTOR_AGENT_MEMORY_ID_FIELD),
+                            allParams.get(EXECUTOR_AGENT_PARENT_INTERACTION_ID_FIELD),
+                            tokenBudget.exhaustedMessage(tokenTracker) + ". Completed steps: " + completedSteps,
```
rithinpullela: Same `List.toString()` issue as flagged at line 504. Also at lines 596 and 1081.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AbstractV2AgentRunner.java
```diff
@@ -501,6 +501,17 @@ protected int getMaxIterations(Map<String, String> params) {
...
     }
 
+    /**
+     * Get agent-level token budget from parameters.
+     * Returns -1 if no budget is configured (unlimited).
+     *
+     * @param params Execution parameters (merged agent + runtime params)
+     * @return Maximum total tokens allowed for the agent execution, or -1 for unlimited
+     */
+    protected long getMaxTokensBudget(Map<String, String> params) {
```
rithinpullela: **I1 (Important) — dead code.**

This method is never called in production. Full-tree grep shows the only callers are 5 tests in `AbstractV2AgentRunnerTest.java` (lines 257, 270, 282, 295, 308). All real budget logic uses `AgentTokenBudget.fromExecutionParams(params)` directly.

Two identical dead methods also exist at `MLChatAgentRunner.java:1132` and `MLPlanExecuteAndReflectAgentRunner.java:1046`. Recommend removing all three along with their tests — the tests provide false coverage confidence for code that's unreachable in production.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -1061,6 +1122,17 @@ public static void saveTraceData(
...
     }
 
+    /**
+     * Get agent-level token budget from parameters.
+     * Returns -1 if no budget is configured (unlimited).
+     *
+     * @param params Execution parameters (merged agent + runtime params)
+     * @return Maximum total tokens allowed for the agent execution, or -1 for unlimited
+     */
+    private long getMaxTokensBudget(Map<String, String> params) {
```
rithinpullela: **I1** — dead code, same as `AbstractV2AgentRunner.java:511`. Not called in production; only test callers exist. Recommend removing.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -958,6 +1036,17 @@ Map<String, Object> getTaskUpdates() {
...
     }
 
+    /**
+     * Get agent-level token budget from parameters.
+     * Returns -1 if no budget is configured (unlimited).
+     *
+     * @param params Execution parameters (merged agent + runtime params)
+     * @return Maximum total tokens allowed for the agent execution, or -1 for unlimited
+     */
+    private long getMaxTokensBudget(Map<String, String> params) {
```
rithinpullela: **I1** — dead code, same as `AbstractV2AgentRunner.java:511`. Not called in production. Recommend removing.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
```diff
@@ -212,6 +212,24 @@ public boolean hasUsage() {
         return !perTurnUsage.isEmpty();
     }
 
+    /**
+     * Returns the cumulative total tokens consumed across all recorded turns.
+     * Used for enforcing agent-level token budgets.
+     *
+     * @return Total tokens consumed, or 0 if no usage has been recorded
+     */
+    public long getCumulativeTotalTokens() {
```
rithinpullela, line 256: **I6 (Important) — unchecked casts in the adjacent `mergeSubAgentUsage` method (pre-existing, not in this diff).**

Leaving this here because this PR expands use of `AgentTokenTracker` in adversarial paths (sub-agent merging at `MLPlanExecuteAndReflectAgentRunner.java:689`), which makes the defensive surface more important.

The class already has the right pattern at line 198:
```java
private static Long getLongFromMap(Map<String, Object> map, String key) {
    Object value = map.get(key);
    if (value instanceof Number) {
        return ((Number) value).longValue();
    }
    return null;
}
```

But `mergeSubAgentUsage` uses unguarded casts at lines 166, 171-172, and 184:
```java
String modelId = (String) subModel.get(MODEL_ID);                         // line 166
String modelUrl = (String) subModel.getOrDefault(MODEL_URL, modelId);     // line 171
String modelName = (String) subModel.getOrDefault(MODEL_NAME, modelId);   // line 172
int subCallCount = subModel.containsKey(CALL_COUNT)
    ? ((Number) subModel.get(CALL_COUNT)).intValue() : 1;                 // line 184
```

If a malformed sub-agent response includes non-String / non-Number values, this throws `ClassCastException`, which is not caught by the tensor-stream `forEach` at `MLPlanExecuteAndReflectAgentRunner.java:677` — surfaces as an ungraceful 500.

Suggested follow-up: add symmetric `getStringFromMap(map, key, fallback)` and `getIntFromMap(map, key, fallback)` helpers next to `getLongFromMap`, and use them at 166, 171-172, 184. Also consider guarding the element cast at line 152.

Happy to split this out into a separate cleanup PR if preferred.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,192 @@
...
+    private AgentTokenBudget(long maxTokens) {
+        this.maxTokens = maxTokens;
+    }
+
+    static AgentTokenBudget fromExecutionParams(Map<String, String> params) {
+        if (params == null) {
+            return unlimited();
+        }
+
+        String maxTokensStr = params.get(MAX_TOKENS_FIELD);
```
rithinpullela, line 40: **Nit on parameter-name consistency** — `fromExecutionParams` reads only `MAX_TOKENS_FIELD` (`"max_tokens"`), but `applyRemainingToParams` downstream reads all three field names (`max_tokens`, `maxTokens`, `maxOutputTokens`) via `OUTPUT_LIMIT_FIELDS` / `smallestPositiveOutputLimit`.

So a Bedrock user who sets `maxTokens=1000` at the agent level gets `UNLIMITED` here (no budget), yet that same value will be picked up as a per-call cap later. Probably document that the budget key is always `max_tokens` regardless of provider, or accept all three on the input side too.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,192 @@
...
+
+        String normalized = llmInterface.trim().toLowerCase(Locale.ROOT);
+        if (normalized.startsWith(LLM_INTERFACE_BEDROCK_CONVERSE) || (normalized.contains("bedrock") && normalized.contains("converse"))) {
+            return BEDROCK_MAX_TOKENS_FIELD;
+        }
+        if (normalized.startsWith(LLM_INTERFACE_GEMINI_V1BETA_GENERATE_CONTENT) || normalized.contains("gemini")) {
+            return GEMINI_MAX_OUTPUT_TOKENS_FIELD;
+        }
+        return MAX_TOKENS_FIELD;
+    }
```
ylwu-amzn: **Blocking — provider detection silently fails open.**

Two concerns:

1. `LLM_INTERFACE_BEDROCK_CONVERSE = "bedrock/converse"` is already a strict prefix of every legitimate Bedrock-converse interface (`bedrock/converse/claude`, `bedrock/converse/deepseek_r1`, fully-qualified model IDs). The `contains("bedrock") && contains("converse")` fallback adds no real coverage but silently accepts anything containing both substrings (incl. user-supplied junk). Same for `contains("gemini")`.

2. **The bigger issue:** when `_llm_interface` is missing from `targetParams`, this returns `MAX_TOKENS_FIELD` regardless of provider. For Bedrock-converse connectors that read `${parameters.maxTokens}` from their request template, the budget clamp will be written to a field the connector ignores — so the per-call cap is silently *not* enforced. Only the post-call exhaustion check stops the loop, which defeats half of the budget mechanism.

Fix options (any of):
- Detect provider from `mlAgent.getModel().getModelProvider()` instead of an interface string.
- Drop the substring fallback and fail loudly when the interface is unrecognized.
- Write all three candidate field names (`max_tokens`, `maxTokens`, `maxOutputTokens`) — overhead is trivial and removes the brittle string match entirely.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,192 @@
...
+        if (value == null) {
+            return null;
+        }
+        try {
+            long parsed = Long.parseLong(value);
+            return parsed > 0 ? parsed : null;
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
```
ylwu-amzn, line 180: **Blocking — `0` is treated as if the field were absent, then overwritten.**

`parsePositiveLong` returns `null` for any non-positive value, so an existing `max_tokens=0` (which several connectors interpret as "no per-call cap") is treated as missing — and `applyRemainingToParams` then overwrites it with `remaining`. That changes the connector's behavior for users who explicitly set `0`.

If the intent is "ignore non-positive values", document it; otherwise treat `0` as "force unlimited per-call" and skip the clamp. Today the behavior is implicit and surprising.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        long effectiveLimit = remaining;
+        Long existingLimit = smallestPositiveOutputLimit(targetParams, llmSpecParams);
+        if (existingLimit != null) {
+            effectiveLimit = Math.min(existingLimit, remaining);
+        }
+
+        targetParams.put(outputLimitField(targetParams), Long.toString(effectiveLimit));
+    }
```
ylwu-amzn, line 116: Note: when `existingLimit` (the smallest positive across `targetParams`/`llmSpecParams`) wins, this overwrites `targetParams[outputLimitField]`. The test `testApplyRemainingToParams_PreservesSmallerExistingLimit` asserts the smaller value ends up in params — which works, but the name "preserves" is misleading: we're rewriting the user's `targetParams.max_tokens=80` with `30` (or with `min(existing, remaining)`). Consider renaming the test or adding a comment explaining the semantic.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
```diff
@@ -215,6 +240,33 @@ private void executeLLMCall(
...
+                            .onResponse(
+                                new AgentLogicResult(
+                                    resultMessage,
+                                    STOP_REASON_MAX_TOKENS,
+                                    accumulatedTokenUsage[0],
+                                    toolInteractionMessages
+                                )
+                            );
+                        return;
+                    }
```
ylwu-amzn, line 266: **Blocking — persisted memory left in inconsistent state.**

`messages.add(assistantMessage)` ran on line 241, *before* this exhaustion check. So if the LLM returned `tool_calls` and the budget is exhausted here, the local `messages` list (and therefore the persisted memory built from it on the parent's success path) contains a tool-calling assistant turn followed by — nothing. No tool result, no follow-up.

The next time this conversation is loaded, the model sees a dangling `tool_use` block. Anthropic-style providers return a 400 ("each `tool_use` must be followed by a `tool_result`"). OpenAI is more lenient but still surprising.

Fix options:
- Move the budget check above `messages.add(assistantMessage)` (line 241) so the orphan turn is never added.
- Drop the trailing tool-calling turn from `messages` on this branch.
- Synthesize a `tool_result` content block with the budget-exhausted text so the conversation is well-formed.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
ylwu-amzn: **High-priority — stop-reason name collides with provider-side stop reasons.**

`STOP_REASON_MAX_TOKENS = "max_tokens"` overlaps with the upstream Bedrock provider's own `max_tokens` stop reason (see `BedrockStreamingHandler.java:497`). Any downstream code keying off `stopReason` to discriminate "model hit its own max_tokens" vs "agent budget exhausted" will be unable to tell them apart.

Suggest `agent_max_tokens_budget` or `budget_exhausted` for the agent-side reason.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
```diff
@@ -160,13 +166,31 @@ private void executeLLMCall(
...
+                .onResponse(
+                    new AgentLogicResult(
+                        createBudgetExhaustedMessage(tokenBudget.exhaustedMessage(consumedBeforeCall)),
+                        STOP_REASON_MAX_TOKENS,
+                        accumulatedTokenUsage[0],
+                        toolInteractionMessages
+                    )
+                );
+            return;
+        }
```
ylwu-amzn, line 192: **High-priority — `parseToolInteractionsForPersistence` can throw and mask the real cause.**

Both the pre-call (here) and post-call (~L252) exhaustion branches call `parseToolInteractionsForPersistence(toolInteractionJsonList, modelProvider)`, which parses JSON. If parsing fails on any entry, the entire budget-exhaustion path will throw and surface as `onFailure` to the caller, masking the actual signal (budget exhausted) with a parse error.

Wrap in try/catch + log; on this path we want to return the partial result rather than fail the run.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -593,6 +650,9 @@ private void executePlanningLoop(
                 }
                 // Mark sub-agent so its token tracker suppresses logging (parent logs merged totals)
                 reactParams.put(AgentTokenTracker.IS_SUB_AGENT_FIELD, "true");
+                if (tokenBudget.isLimited()) {
+                    reactParams.put(AgentTokenBudget.MAX_TOKENS_FIELD, Long.toString(tokenBudget.remaining(tokenTracker)));
+                }
```
ylwu-amzn, line 676: **Blocking — sub-agent token usage isn't merged back on failure.**

The parent passes `tokenBudget.remaining(tokenTracker)` as the sub-agent's budget here. The sub-agent then constructs its own `AgentTokenBudget`/`AgentTokenTracker` from zero — fine. But the parent only learns the sub-agent's actual usage on the success path (line 689 — `tokenTracker.mergeSubAgentUsage(...)` inside the tensor switch). If the sub-agent fails partway through (the wrapping `MLExecuteTaskAction` errors out after one or more LLM calls), the parent never sees those tokens, and on the next planning iteration `tokenBudget.isExhausted(tokenTracker)` will report "plenty left" while the user has already been billed.

Mitigations:
- Have the sub-agent emit a `token_usage` tensor on its failure path too.
- Or have the parent subtract the budget passed to the sub-agent regardless of result, treating it as a sunk cost.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -512,6 +544,7 @@ private void executePlanningLoop(
 
         }
 
+        tokenBudget.applyRemainingToParams(allParams, llm.getParameters(), tokenTracker);
```
ylwu-amzn: **High-priority — mutating shared `allParams` across iterations.**

This writes `max_tokens` (or `maxTokens`/`maxOutputTokens`) directly into `allParams`, which is the long-lived map shared across the entire planning loop. After iteration 1, `allParams.max_tokens` is no longer the user's `100` budget — it's the per-call clamp from the previous iteration.

In this code path the `AgentTokenBudget` is created once (line 309) and reused, so the bug is dormant. But it's a footgun for any future refactor that re-derives the budget from `allParams` later. Consider writing the clamp to a per-request `new HashMap<>(allParams)` rather than mutating `allParams` in place.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -958,6 +1036,17 @@ Map<String, Object> getTaskUpdates() {
...
+    /**
+     * Get agent-level token budget from parameters.
+     * Returns -1 if no budget is configured (unlimited).
+     *
+     * @param params Execution parameters (merged agent + runtime params)
+     * @return Maximum total tokens allowed for the agent execution, or -1 for unlimited
+     */
+    private long getMaxTokensBudget(Map<String, String> params) {
+        return AgentTokenBudget.fromExecutionParams(params).getMaxTokens();
+    }
```
ylwu-amzn: **Blocking — dead code.**

`getMaxTokensBudget` is `private` here and never called from anywhere in `MLPlanExecuteAndReflectAgentRunner` (callers use the `AgentTokenBudget` instance directly). Same dead method also added in `MLChatAgentRunner.java:1132`. Delete both — `AbstractV2AgentRunner.getMaxTokensBudget` is the only one with real callers/tests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
ylwu-amzn: **Blocking — dead code.** Mirrors the same unused method in `MLPlanExecuteAndReflectAgentRunner.java:1046`. No callers in the production code; please delete.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -1433,6 +1506,12 @@ private void handleMaxIterationsReached(
             );
         }, listener::onFailure);
 
+        if (tokenBudget.isExhausted(tokenTracker)) {
+            String fallbackResponse = buildMaxIterationsFallbackResponse(maxIterations, lastThought);
+            responseListener.onResponse(fallbackResponse);
+            return;
+        }
```
ylwu-amzn, line 1600: **Blocking — wrong message when budget exhausted at max-iterations time.**

When both "max iterations reached" and "budget exhausted" trigger on the same final iteration, this returns the max-iterations fallback string ("Max iterations reached…") even though the budget is the actual cause. Every other budget-exhaustion site in the PR returns `tokenBudget.exhaustedMessage(...)`, so the user-visible reason is now inconsistent across paths.

Use `tokenBudget.exhaustedMessage(tokenTracker)` here, or compose both reasons ("Max iterations reached AND token budget exhausted (X/Y)").

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -474,7 +489,24 @@ private void executePlanningLoop(
...
+            return;
+        }
+        if (tokenBudget.isExhausted(tokenTracker)) {
+            boolean includeTokenUsage = Boolean.parseBoolean(allParams.getOrDefault(AgentUtils.INCLUDE_TOKEN_USAGE, "false"));
+            saveAndReturnFinalResult(
+                memory,
+                parentInteractionId,
+                allParams.get(EXECUTOR_AGENT_MEMORY_ID_FIELD),
+                allParams.get(EXECUTOR_AGENT_PARENT_INTERACTION_ID_FIELD),
+                tokenBudget.exhaustedMessage(tokenTracker) + ". Completed steps: " + completedSteps,
```
ylwu-amzn: **High-priority — `List.toString()` leaks `[a, b, c]` into user-visible output.**

`completedSteps` is a `List<String>`, and `". Completed steps: " + completedSteps` invokes `List.toString()`, producing `"[step1, step2, step3]"` with the surrounding brackets and `, ` separators. That renders as garbage in chat UIs.

Same pattern appears at lines ~596, ~763, ~1081, ~1123. Either format as a numbered list (`String.join("\n", completedSteps)`) or move the steps into trace tensors / additionalInfo and keep the headline message clean.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
```diff
@@ -212,6 +212,24 @@ public boolean hasUsage() {
...
+        long total = 0;
+        for (ModelUsageAggregation agg : perModelUsage.values()) {
+            TokenUsage usage = agg.getAggregatedUsage();
+            Long effectiveTotal = usage.getEffectiveTotalTokens();
+            if (effectiveTotal != null) {
+                total += effectiveTotal;
+            }
+        }
+        return total;
```
ylwu-amzn, line 262: Minor: `getEffectiveTotalTokens()` returns `null` when only one of `inputTokens`/`outputTokens` is set, so this method silently treats partial data as zero. For providers that report only input tokens (rare), the budget would never trigger.

Consider summing `(inputTokens != null ? inputTokens : 0) + (outputTokens != null ? outputTokens : 0)` when at least one is non-null. Not blocking — just worth flagging since the budget assumes `getEffectiveTotalTokens` reflects reality.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -413,7 +425,8 @@ private void setToolsAndRunAgent(
         String conversationId,
         ActionListener<Object> finalListener,
         AgentTokenTracker tokenTracker,
-        FunctionCalling functionCalling
+        FunctionCalling functionCalling,
+        AgentTokenBudget tokenBudget
```
ylwu-amzn, line 429: Nit: `setupMemoryAndRun`, `setToolsAndRunAgent`, and `executePlanningLoop` are now 9–11 parameters each after this PR. Consider extracting a small `PerExecutionContext` (carrying `tokenTracker`, `tokenBudget`, `functionCalling`, `memory`, `conversationId`) before signature creep gets worse. Not blocking.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
GugaGlonti: Fixed. Explicit invalid/non-positive max_tokens now fails closed, and PER short-circuits before dispatching a sub-agent when the remaining budget is 0.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
GugaGlonti: Fixed. Missing max_tokens remains unlimited, but explicit invalid, zero, negative, overflow, decimal, scientific notation, and whitespace-invalid values now throw IllegalArgumentException.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
GugaGlonti: Fixed. Agent budget exhaustion now uses budget_exhausted, and V1/PER budget-exhausted responses include stop_reason in dataAsMap.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AbstractV2AgentRunner.java
GugaGlonti: Fixed. Removed the unused helpers from AbstractV2, V1, and PER, along with the false-coverage tests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
GugaGlonti: Fixed. Removed the unused helpers from AbstractV2, V1, and PER, along with the false-coverage tests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
GugaGlonti: Fixed. Removed the unused helpers from AbstractV2, V1, and PER, along with the false-coverage tests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
GugaGlonti: Fixed. Added defensive map/list handling plus typed string/int/long helpers so malformed sub-agent usage data is ignored instead of throwing ClassCastException.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
GugaGlonti: Intentional. The agent-level budget key remains parameters.max_tokens only. maxTokens and maxOutputTokens are provider request/output-limit fields, not alternate budget inputs. Downstream clamping now writes all three provider-compatible fields.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
GugaGlonti: Fixed. Removed provider/interface detection for output-limit clamping. The effective clamp is now written to max_tokens, maxTokens, and maxOutputTokens.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
GugaGlonti: Intentional. Non-positive output-limit values are not treated as valid smaller caps; an active agent-level budget still clamps the outgoing request to the remaining budget.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
GugaGlonti: Fixed. Renamed the test to describe clamping to the smaller effective limit.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
GugaGlonti: Fixed. V2 now checks budget exhaustion before adding the assistant tool-call message to messages, so exhausted tool-call turns are not persisted without tool results.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
GugaGlonti: Fixed. Replaced agent-side "max_tokens" with "budget_exhausted".

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
GugaGlonti: Fixed. Removed the unused helpers from AbstractV2, V1, and PER, along with the false-coverage tests.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
GugaGlonti: Fixed. If V1 hits max iterations and budget exhaustion together, it now returns the budget-exhausted message and structured stop_reason=budget_exhausted.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
GugaGlonti: Fixed. Completed steps are now newline-joined instead of using Java List.toString() formatting.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
GugaGlonti: Fixed. getCumulativeTotalTokens() now counts input-only or output-only token data when total_tokens is absent.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
GugaGlonti: Acknowledged. I left the PerExecutionContext extraction out of this PR to avoid expanding the scope beyond the review fixes.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
rithinpullela: Looks like this issue has not been addressed yet, can you address this? Should be a small fix

rithinpullela (commented):
Thanks for addressing all the comments @GugaGlonti, the PR looks good to me to merge except for `parseToolInteractionsForPersistence` part in `MLChatAgentRunnerV2.java`, can you please take a look

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -1455,10 +1621,17 @@ private void handleMaxIterationsReached(
...
+            functionCalling,
+            tokenBudget
         );
     }
 
+    private String buildMaxIterationsFallbackResponse(int maxIterations, AtomicReference<String> lastThought) {
+        return (lastThought.get() != null && !lastThought.get().isEmpty() && !"null".equals(lastThought.get()))
+            ? String.format("%s. Last thought: %s", String.format(MAX_ITERATIONS_MESSAGE, maxIterations), lastThought.get())
+            : String.format(MAX_ITERATIONS_MESSAGE, maxIterations);
+    }
```
ylwu-amzn, line 1633: **Blocking — dead helper, never called.**

This private method is defined but has zero call sites. The same expression is still inlined at lines 1614-1619 inside the `e -> { ... }` block of `generateLLMSummary`'s listener.

Either replace the inline expression at 1614-1619 with `responseListener.onResponse(buildMaxIterationsFallbackResponse(maxIterations, lastThought));`, or delete this helper. As-is, it's dead code that will rot.

```
$ grep -n buildMaxIterationsFallbackResponse MLChatAgentRunner.java
1629:    private String buildMaxIterationsFallbackResponse(...)
```
(only the definition; no callers.)

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -717,12 +816,16 @@ private void executePlanningLoop(
...
+                        tokenBudget
                     );
                 }, e -> {
                     String agentId = allParams.getOrDefault(AGENT_ID_LOG_FIELD, "unknown");
                     String tenantIdLog = allParams.get(TENANT_ID_FIELD);
                     log.error("Failed to execute ReAct agent. agentId={}, tenantId={}", agentId, tenantIdLog, e);
+                    if (tokenBudget.isLimited()) {
+                        tokenTracker.recordReservedTokens(reActAgentId, subAgentBudget);
+                    }
                     finalListener.onFailure(e);
```
ylwu-amzn, line 829: **High-priority — `recordReservedTokens` mutation is dead before `onFailure`.**

`recordReservedTokens` mutates `tokenTracker`, but the very next line invokes `finalListener.onFailure(e)`, terminating the agent run. Nothing else reads the tracker on this path, so the reservation is silently discarded.

Two possible intents — neither matches the current code:

1. **Pre-reservation against runaway retries**: then the reservation must happen *before* dispatching the executor at line 691 (`client.execute(MLExecuteTaskAction.INSTANCE, ...)`) and be debited back when the actual sub-agent usage merges in via `mergeSubAgentUsage` (line 710). As written, it's executed too late to matter.
2. **Telemetry on the failed call**: then the merged `tokenTracker.toOutputMap()` needs to actually flow into the failure response — but `finalListener.onFailure(e)` carries only the exception.

Please clarify intent and adjust accordingly, or delete the call. There's also no test for this path — `MLPlanExecuteAndReflectAgentRunnerTest` does not exercise the executor `onFailure` branch.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunnerV2.java
```diff
@@ -213,6 +237,34 @@ private void executeLLMCall(
...
+                            .onResponse(
+                                new AgentLogicResult(
+                                    resultMessage,
+                                    AgentTokenBudget.STOP_REASON_BUDGET_EXHAUSTED,
+                                    accumulatedTokenUsage[0],
+                                    toolInteractionMessages
+                                )
+                            );
+                        return;
+                    }
```
ylwu-amzn, line 266: **High-priority — `stop_reason` mis-classifies natural finish on the over-budget call.**

When the LLM returns a complete, tool-call-free response that just happens to push cumulative usage past the budget, this branch (correctly) surfaces the model's actual final answer (`assistantMessage`) — but tags the result `stop_reason=budget_exhausted`. From the caller's perspective this is indistinguishable from "we cut you off mid-thought," but in this case the model finished cleanly.

Consider:
```java
String stopReason = (toolCalls == null || toolCalls.isEmpty())
    ? STOP_REASON_END_TURN
    : AgentTokenBudget.STOP_REASON_BUDGET_EXHAUSTED;
```

If `budget_exhausted` is desired regardless (e.g., to signal "do not continue"), document the semantic explicitly. The existing `testExecuteAgentLogic_TokenBudgetExhausted_StopsExecution` only exercises the tool-call path, so this natural-finish branch has no direct coverage.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -474,7 +489,25 @@ private void executePlanningLoop(
...
+                tokenBudget.exhaustedMessage(tokenTracker) + ". Completed steps:\n" + formatCompletedSteps(completedSteps),
+                null,
+                finalListener,
+                tokenTracker,
+                allParams.get(TENANT_ID_FIELD),
+                includeTokenUsage,
+                AgentTokenBudget.STOP_REASON_BUDGET_EXHAUSTED
             );
             return;
         }
```
ylwu-amzn, line 513: **Medium — duplicated boilerplate (4 sites in this file).**

The same 11-line "budget exhausted → saveAndReturnFinalResult" block appears four times in this file: here (497-513), 582-605, 659-674, and 778-792. Each computes the same `includeTokenUsage` boolean, the same message string `tokenBudget.exhaustedMessage(...) + ". Completed steps:\n" + formatCompletedSteps(completedSteps)`, and calls `saveAndReturnFinalResult` with `STOP_REASON_BUDGET_EXHAUSTED`.

Extract a helper:

```java
private void returnBudgetExhausted(
    Memory memory, String parentInteractionId, Map<String, String> allParams,
    String exhaustedMessage, List<String> completedSteps,
    AgentTokenTracker tokenTracker, ActionListener<Object> finalListener
) {
    boolean includeTokenUsage = Boolean.parseBoolean(allParams.getOrDefault(AgentUtils.INCLUDE_TOKEN_USAGE, "false"));
    saveAndReturnFinalResult(
        memory, parentInteractionId,
        allParams.get(EXECUTOR_AGENT_MEMORY_ID_FIELD),
        allParams.get(EXECUTOR_AGENT_PARENT_INTERACTION_ID_FIELD),
        exhaustedMessage + ". Completed steps:\n" + formatCompletedSteps(completedSteps),
        null, finalListener, tokenTracker,
        allParams.get(TENANT_ID_FIELD), includeTokenUsage,
        AgentTokenBudget.STOP_REASON_BUDGET_EXHAUSTED
    );
}
```

Drops ~40 LOC and removes the risk of one site drifting from the others.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -923,6 +1065,15 @@ void saveAndReturnFinalResult(
...
     }
 
+    private static Map<String, Object> responseDataMap(String finalResult, String stopReason) {
+        Map<String, Object> dataAsMap = new HashMap<>();
+        dataAsMap.put(RESPONSE_FIELD, finalResult);
+        if (stopReason != null) {
+            dataAsMap.put("stop_reason", stopReason);
```
ylwu-amzn, line 1072: **Medium — magic string.**

`AgentTokenBudget.STOP_REASON_BUDGET_EXHAUSTED` is a constant, but the field name `"stop_reason"` is hard-coded both here and at `MLChatAgentRunner.java:1524`. Promote to a constant (e.g., `AgentTokenBudget.STOP_REASON_FIELD = "stop_reason"`) so a future rename doesn't drift between the two callers.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java
```diff
@@ -1386,6 +1516,16 @@ public static void returnFinalResponse(
         }
     }
 
+    private static Map<String, Object> responseDataMap(String finalAnswer, Map<String, Object> additionalInfo, String stopReason) {
+        Map<String, Object> dataAsMap = new HashMap<>();
+        dataAsMap.put("response", finalAnswer);
+        dataAsMap.put(ADDITIONAL_INFO_FIELD, additionalInfo);
+        if (stopReason != null) {
+            dataAsMap.put("stop_reason", stopReason);
```
ylwu-amzn, line 1524: **Medium — magic string (paired with `MLPlanExecuteAndReflectAgentRunner.java:1072`).**

Same `"stop_reason"` literal appears in both files. Promote to a constant and reference it from both call sites.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenBudget.java
```diff
@@ -0,0 +1,182 @@
...
+        Long existingLimit = smallestPositiveOutputLimit(targetParams, llmSpecParams);
+        if (existingLimit != null) {
+            effectiveLimit = Math.min(existingLimit, remaining);
+        }
+
+        String effectiveLimitString = Long.toString(effectiveLimit);
+        for (String field : OUTPUT_LIMIT_FIELDS) {
+            targetParams.put(field, effectiveLimitString);
+        }
+    }
```
ylwu-amzn, line 116: **Medium — input-prompt tokens are not pre-checked, only output-cap is clamped.**

This only writes the three output-cap fields (`max_tokens`, `maxTokens`, `maxOutputTokens`). The user-facing `max_tokens` is documented as "the total agent execution budget" (input + output across all calls), but on the next LLM call we cannot know the input-prompt token count until after the response. A large prompt can therefore push cumulative usage well past the budget before the post-call check at `MLChatAgentRunner.java:594` catches it.

This isn't broken — the post-call check still fires — but the user could see *substantially* more than `max_tokens` actually charged on a single oversized prompt. Please document this semantic explicitly in the class doc, e.g.:

> `max_tokens` is enforced post-hoc per LLM call; a single in-flight call can exceed the remaining budget by up to its input prompt size. Output limits are clamped pre-call.

(The PR description should probably say this too.)

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLPlanExecuteAndReflectAgentRunner.java
```diff
@@ -691,6 +772,26 @@ private void executePlanningLoop(
                                 )
                         );
 
+                    addSteps(completedSteps, allParams, COMPLETED_STEPS_FIELD);
```
ylwu-amzn, line 775: **Medium — `addSteps` reordering deserves a comment.**

This `addSteps(...)` call was previously *after* `saveTraceData(...)`; the PR moves it before both `saveTraceData` and the new budget-check so that `formatCompletedSteps(completedSteps)` in the budget-exhausted message includes the step that just finished.

Functionally appears safe (no caller of `saveTraceData` reads `allParams[COMPLETED_STEPS_FIELD]`), but please add a one-line comment documenting why this must precede the budget check — otherwise a future cleanup is likely to reorder it back. Something like:

```java
// Must precede the budget check below so the just-finished step is included in the exhaust message.
addSteps(completedSteps, allParams, COMPLETED_STEPS_FIELD);
```

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
```diff
@@ -212,6 +247,34 @@ public boolean hasUsage() {
...
+        if (totalTokens != null) {
+            return totalTokens;
+        }
+        Long inputTokens = usage.getInputTokens();
+        Long outputTokens = usage.getOutputTokens();
+        if (inputTokens != null || outputTokens != null) {
+            return (inputTokens != null ? inputTokens : 0L) + (outputTokens != null ? outputTokens : 0L);
+        }
+        return 0L;
+    }
```
ylwu-amzn, line 276: **Style/nit — possible duplication of `TokenUsage.getEffectiveTotalTokens()`.**

`AgentTokenBudget.consumedTokens(TokenUsage)` already delegates to `usage.getEffectiveTotalTokens()` (per `AgentTokenBudget.java:142`). If `TokenUsage.getEffectiveTotalTokens()` does the same fallback (`totalTokens` → `input + output` → `0`), this private method is duplicated logic — just call it directly and remove this helper. If it does *not* do the same thing, the divergence is itself a bug worth resolving.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/AgentTokenTracker.java
```diff
@@ -195,6 +202,34 @@ public void mergeSubAgentUsage(Map<String, Object> tokenUsageMap) {
         }
     }
 
+    public void recordReservedTokens(String modelId, long tokenCount) {
+        if (tokenCount <= 0) {
+            return;
+        }
+        String effectiveModelId = modelId != null ? modelId : "unknown";
```
ylwu-amzn, line 209: **Nit — magic string `"unknown"`.**

If this constant appears anywhere else in the codebase (likely it does — telemetry pipelines often have a canonical "unknown model" sentinel), promote to a shared constant. Also worth pausing on: is silently dropping the reservation (when modelId is null) a better choice than emitting an `"unknown"` per-model row that downstream consumers may not expect? Tied to H1 — this method's correct call-site is unclear.

PR #4824, opened 2026-05-17
feat: Add structured output (constrained decoding) support for agentic memory fact extraction

### LINKED ISSUE #4799: [FEATURE] Support structured output / constrained decoding for agentic memory fact extraction
### Problem

Agentic memory fact extraction currently relies on prompt engineering to get LLMs to return valid JSON (`{"facts": ["fact1", "fact2"]}`). This is fragile — smaller models produce malformed JSON (see #4780), and even capable models occasionally fail. We mitigate this with JSON enforcement messages and simplified prompts (#4798), but the fundamental problem remains: we're asking the model to produce structured output via natural language instructions rather than enforcing it at the API layer.

### Proposal

Add support for **structured output / constrained decoding** in the ML connector and agentic memory extraction pipeline. When the underlying LLM provider supports it, use the provider's native structured output mechanism to guarantee valid JSON responses. Fall back to prompt-based extraction for providers that don't support it.

### Provider Landscape

Every major LLM provider now supports some form of constrained decoding:

| Provider | Mechanism | API Parameter | Constrained Decoding |
|----------|-----------|---------------|---------------------|
| **AWS Bedrock (Converse)** | JSON Schema output format | `outputConfig.textFormat.type: "json_schema"` | ✅ Yes |
| **AWS Bedrock (Converse)** | Strict tool use | `toolSpec.strict: true` | ✅ Yes |
| **OpenAI** | Structured outputs | `response_format: { type: "json_schema", json_schema: {...} }` | ✅ Yes |
| **Azure OpenAI** | Structured outputs | Same as OpenAI (model-dependent) | ✅ Yes (GPT-4o+) |
| **Google Gemini / Vertex AI** | Controlled generation | `response_schema` + `responseMimeType: "application/json"` | ✅ Yes |
| **Ollama (local models)** | JSON schema format | `format: { schema }` (uses GBNF grammars via llama.cpp) | ✅ Yes (v0.5+) |
| **Mistral** | JSON mode | `response_format: { type: "json_object" }` | ⚠️ JSON mode only (no schema enforcement) |
| **Cohere** | Response format | `response_format: { type: "json_object", schema: {...} }` | ✅ Yes |

This is no longer a niche feature — it's an industry standard. The fact extraction JSON schema is simple and stable:

```json
{
  "type": "object",
  "properties": {
    "facts": {
      "type": "array",
      "items": { "type": "string" }
    }
  },
  "required": ["facts"],
  "additionalProperties": false
}
```

### Benefits

1. **Eliminates JSON parse failures entirely** — Constrained decoding guarantees the output conforms to the schema at the token generation level. No more silent strategy failures.
2. **Removes prompt engineering overhead** — The JSON enforcement messages (`USER_PREFERENCE_JSON_ENFORCEMENT_MESSAGE`, `JSON_ENFORCEMENT_MESSAGE`) and format instructions in extraction prompts become unnecessary, saving tokens per extraction call.
3. **Enables smaller/cheaper models** — With guaranteed JSON output, customers can confidently use cheaper models (Haiku, GPT-4o-mini, local Ollama models) for extraction without worrying about parse failures.
4. **Improves extraction quality** — The model can focus entirely on extracting good facts rather than splitting attention between content quality and format compliance.

### Design Considerations

1. **Connector-level abstraction** — The structured output parameter varies by provider (`outputConfig` for Bedrock, `response_format` for OpenAI, `format` for Ollama). The connector layer should abstract this, similar to how it already abstracts auth (`aws_sigv4` vs `http`) and request/response formats.

2. **Graceful fallback** — If the connector doesn't support structured output (e.g., a custom HTTP connector to an internal LLM), fall back to the current prompt-based approach with JSON enforcement messages.

3. **Schema passed at request time** — The extraction schema should be passed as part of the predict request, not hardcoded in the connector. This allows different schemas for different use cases (fact extraction vs memory decisions).

### References

- #4780 — Silent JSON parse failures with smaller LLMs (the motivating bug)
- #4798 — Simplified USER_PREFERENCE prompt (immediate mitigation)
- [AWS Blog: Structured outputs on Amazon Bedrock](https://aws.amazon.com/blogs/machine-learning/structured-outputs-on-amazon-bedrock-schema-compliant-ai-responses/) (Feb 2026)
- [OpenAI: Introducing Structured Outputs](https://openai.com/index/introducing-structured-outputs-in-the-api/)
- [Ollama: Structured Outputs](https://ollama.com/blog/structured-outputs)
- Suggestion from @rithinpullela in #4798 review

ehavener:
### Description
Adds structured output (constrained decoding) support for agentic memory fact extraction. When a connector's predict action has `supports_structured_output: true`, `MemoryContainerHelper` resolves the provider from the connector URL and injects a provider-specific JSON schema into fact-extraction requests. `HttpConnector` reads `_*_json` / `_*_additions_json` parameters and injects them as top-level fields in the outgoing request body, so the provider enforces output structure at the token level rather than relying on prompt instructions.

**Implemented providers:** OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere v2, Google Gemini/Vertex AI, Amazon Bedrock Converse

#### How it works

1. **`ConnectorAction`** — adds a `supportsStructuredOutput` boolean field (default `false`); wire-serialized for clusters ≥ 3.7.0 only.
2. **`HttpConnector.injectStructuredOutputParams`** — reads `_<X>_json` / `_<X>_additions_json` parameters and injects them as top-level fields in the outgoing request body.
3. **`MemoryContainerHelper.getStructuredOutputParameters`** — async lookup that resolves the model's connector, checks the flag, and returns the provider-specific schema parameters. Returns empty map on any failure; callers fall back to prompt enforcement.
4. **`MemoryProcessingService`** — merges structured output params into the predict request when available, skipping the prompt enforcement sentence. Falls back gracefully on lookup failure.

#### Manual testing
- [x] Ollama : A smoke test via a direct _predict call with _response_format_json confirmed that HttpConnector.injectStructuredOutputParams correctly reads the _*_json naming convention and injects response_format into the Ollama request body. Ollama then enforces the schema at the token level, returning {"facts": ["The user prefers dark mode."]} instead of free-forming the structure.

- [x] Bedrock Converse (claude-sonnet-4-6)

### Related Issues
Resolves [#4799](https://github.com/opensearch-project/ml-commons/issues/4799)

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/pull/1130).

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -7,6 +7,7 @@ package org.opensearch.ml.common.connector;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.VERSION_3_5_0;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -43,6 +44,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String REQUEST_BODY_FIELD = "request_body";
     public static final String ACTION_PRE_PROCESS_FUNCTION = "pre_process_function";
     public static final String ACTION_POST_PROCESS_FUNCTION = "post_process_function";
+    public static final String SUPPORTS_STRUCTURED_OUTPUT_FIELD = "supports_structured_output";
     public static final String OPENAI = "openai";
     public static final String COHERE = "cohere";
     public static final String BEDROCK = "bedrock";
@@ -70,6 +72,20 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     private String requestBody;
     private String preProcessFunction;
     private String postProcessFunction;
+    private boolean supportsStructuredOutput;
+
+    public ConnectorAction(
+        ActionType actionType,
+        String name,
+        String method,
+        String url,
+        Map<String, String> headers,
+        String requestBody,
+        String preProcessFunction,
+        String postProcessFunction
+    ) {
+        this(actionType, name, method, url, headers, requestBody, preProcessFunction, postProcessFunction, false);
+    }
 
     @Builder(toBuilder = true)
     public ConnectorAction(
@@ -80,7 +96,8 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         Map<String, String> headers,
         String requestBody,
         String preProcessFunction,
-        String postProcessFunction
+        String postProcessFunction,
+        boolean supportsStructuredOutput
     ) {
         if (actionType == null) {
             throw new IllegalArgumentException("action type can't be null");
@@ -101,6 +118,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         this.requestBody = requestBody;
         this.preProcessFunction = preProcessFunction;
         this.postProcessFunction = postProcessFunction;
+        this.supportsStructuredOutput = supportsStructuredOutput;
     }
 
     public ConnectorAction(StreamInput input) throws IOException {
@@ -115,6 +133,9 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         this.postProcessFunction = input.readOptionalString();
         if (input.getVersion().onOrAfter(VERSION_3_5_0)) {
             this.name = input.readOptionalString();
+        }
+        if (input.getVersion().onOrAfter(VERSION_3_7_0)) {
+            this.supportsStructuredOutput = input.readBoolean();
         }
     }
 
@@ -134,6 +155,9 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         out.writeOptionalString(postProcessFunction);
         if (out.getVersion().onOrAfter(VERSION_3_5_0)) {
             out.writeOptionalString(name);
+        }
+        if (out.getVersion().onOrAfter(VERSION_3_7_0)) {
+            out.writeBoolean(supportsStructuredOutput);
         }
     }
 
@@ -164,6 +188,9 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         if (name != null) {
             builder.field(NAME_FIELD, name);
         }
+        if (supportsStructuredOutput) {
+            builder.field(SUPPORTS_STRUCTURED_OUTPUT_FIELD, true);
+        }
         return builder.endObject();
     }
 
@@ -181,6 +208,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
         String requestBody = null;
         String preProcessFunction = null;
         String postProcessFunction = null;
+        boolean supportsStructuredOutput = false;
 
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
@@ -212,6 +240,9 @@ public class ConnectorAction implements ToXContentObject, Writeable {
                 case ACTION_POST_PROCESS_FUNCTION:
                     postProcessFunction = parser.text();
                     break;
+                case SUPPORTS_STRUCTURED_OUTPUT_FIELD:
+                    supportsStructuredOutput = parser.booleanValue();
+                    break;
                 default:
                     parser.skipChildren();
                     break;
@@ -227,6 +258,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
             .requestBody(requestBody)
             .preProcessFunction(preProcessFunction)
             .postProcessFunction(postProcessFunction)
+            .supportsStructuredOutput(supportsStructuredOutput)
             .build();
     }
 
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -69,9 +69,7 @@ public class HttpConnector extends AbstractConnector {
     static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
         .of(
             "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
-            "generationConfig", // Google Gemini
-            "tools",            // Bedrock Converse (future)
-            "tool_choice"       // Bedrock Converse (future)
+            "generationConfig"  // Google Gemini
         );
 
     // TODO: add RequestConfig like request time out,
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -378,7 +378,7 @@ public class HttpConnector extends AbstractConnector {
                     payload = jsonObject.toString();
                 }
             }
-            if (parameters != null && isJson(payload)) {
+            if (parameters != null && isJson(payload) && connectorAction.get().isSupportsStructuredOutput()) {
                 payload = injectStructuredOutputParams(parameters, payload);
             }
             return (T) payload;
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -69,7 +69,8 @@ public class HttpConnector extends AbstractConnector {
     static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
         .of(
             "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
-            "generationConfig"  // Google Gemini
+            "generationConfig", // Google Gemini
+            "toolConfig"        // Amazon Bedrock Converse
         );
 
     // TODO: add RequestConfig like request time out,
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -385,56 +385,59 @@ public class HttpConnector extends AbstractConnector {
         return (T) parameters.get("http_body");
     }
 
-    // Convention: parameters named _<fieldname>_additions_json are merged into an existing top-level
-    // JSON object; parameters named _<fieldname>_json are injected as a new top-level JSON object.
-    // Both conventions require the value to be a valid JSON object string and the target field name
-    // must be in STRUCTURED_OUTPUT_ALLOWED_FIELDS to prevent callers from overriding provider control fields.
+    // Convention: parameters named _<fieldname>_json inject/replace a top-level JSON object field;
+    // parameters named _<fieldname>_additions_json merge into an existing top-level JSON object field.
+    // Both conventions require: (a) the payload is a JSON object (array payloads are left untouched),
+    // (b) a non-empty, allowlisted field name, and (c) a valid JSON object value string.
+    // Replacements (_X_json) are applied before merges (_X_additions_json) so behaviour is
+    // deterministic when both are present for the same field.
+    // Note: matched parameters are read but not removed from the map.
     private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
+        JsonElement parsed = JsonParser.parseString(payload);
+        if (!parsed.isJsonObject()) return payload;
+        JsonObject body = parsed.getAsJsonObject();
+        boolean modified = false;
+
+        // Pass 1: _<field>_json — inject or replace a top-level field
         for (Map.Entry<String, String> entry : parameters.entrySet()) {
-            String key = entry.getKey();
-            if (key.startsWith("_") && key.endsWith("_additions_json")) {
-                String fieldName = key.substring(1, key.length() - "_additions_json".length());
-                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
-                    payload = mergeIntoJsonObject(payload, fieldName, entry.getValue());
-                }
-            } else if (key.startsWith("_") && key.endsWith("_json")) {
-                String fieldName = key.substring(1, key.length() - "_json".length());
-                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
-                    payload = injectTopLevelJsonField(payload, fieldName, entry.getValue());
-                }
-            }
+            if (entry.getKey().endsWith("_additions_json")) continue;
+            String fieldName = allowedFieldName(entry.getKey(), "_json");
+            JsonObject value = asJsonObject(entry.getValue());
+            if (fieldName == null || value == null) continue;
+            body.add(fieldName, value);
+            modified = true;
         }
-        return payload;
+
+        // Pass 2: _<field>_additions_json — merge into an existing top-level object
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            String fieldName = allowedFieldName(entry.getKey(), "_additions_json");
+            JsonObject additions = asJsonObject(entry.getValue());
+            if (fieldName == null || additions == null) continue;
+            JsonObject target = body.has(fieldName) && body.get(fieldName).isJsonObject()
+                ? body.getAsJsonObject(fieldName)
+                : new JsonObject();
+            additions.entrySet().forEach(e -> target.add(e.getKey(), e.getValue()));
+            body.add(fieldName, target);
+            modified = true;
+        }
+
+        return modified ? body.toString() : payload;
     }
 
-    private String injectTopLevelJsonField(String payload, String fieldName, String valueStr) {
-        if (valueStr == null || !isJson(valueStr)) {
-            return payload;
-        }
-        JsonElement valueElement = JsonParser.parseString(valueStr);
-        if (!valueElement.isJsonObject()) {
-            return payload;
-        }
-        JsonObject jsonObject = JsonParser.parseString(payload).getAsJsonObject();
-        jsonObject.add(fieldName, valueElement.getAsJsonObject());
-        return jsonObject.toString();
+    /** Returns the allowlisted field name encoded in a _&lt;field&gt;_&lt;suffix&gt; key, or null if invalid. */
+    private static String allowedFieldName(String key, String suffix) {
+        if (!key.startsWith("_") || !key.endsWith(suffix)) return null;
+        int end = key.length() - suffix.length();
+        if (end <= 1) return null;
+        String fieldName = key.substring(1, end);
+        return STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName) ? fieldName : null;
     }
 
-    private String mergeIntoJsonObject(String payload, String fieldName, String additionsStr) {
-        if (additionsStr == null || !isJson(additionsStr)) {
-            return payload;
-        }
-        JsonElement additionsElement = JsonParser.parseString(additionsStr);
-        if (!additionsElement.isJsonObject()) {
-            return payload;
-        }
-        JsonObject jsonObject = JsonParser.parseString(payload).getAsJsonObject();
-        JsonObject target = jsonObject.has(fieldName) && jsonObject.get(fieldName).isJsonObject()
-            ? jsonObject.getAsJsonObject(fieldName)
-            : new JsonObject();
-        additionsElement.getAsJsonObject().entrySet().forEach(e -> target.add(e.getKey(), e.getValue()));
-        jsonObject.add(fieldName, target);
-        return jsonObject.toString();
+    /** Parses a JSON string as an object, returning null if absent, invalid, or not an object. */
+    private static JsonObject asJsonObject(String json) {
+        if (json == null || !isJson(json)) return null;
+        JsonElement el = JsonParser.parseString(json);
+        return el.isJsonObject() ? el.getAsJsonObject() : null;
     }
 
     private boolean neededStreamParameterInPayload(Map<String, String> parameters) {
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -61,6 +62,17 @@ public class HttpConnector extends AbstractConnector {
     public static final String REGION_FIELD = "region";
     // TODO: move the AgentUtils class from algorithm module to common module
     public static final String LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS = "openai/v1/chat/completions";
+
+    // Allowlist of top-level JSON fields that _<field>_json / _<field>_additions_json parameters may inject.
+    // Restricts injection to structured-output fields only; prevents callers from overriding
+    // provider control fields such as "model", "stream", or "max_tokens".
+    static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
+        .of(
+            "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
+            "generationConfig", // Google Gemini
+            "tools",            // Bedrock Converse (future)
+            "tool_choice"       // Bedrock Converse (future)
+        );
 
     // TODO: add RequestConfig like request time out,
 
@@ -377,16 +389,21 @@ public class HttpConnector extends AbstractConnector {
 
     // Convention: parameters named _<fieldname>_additions_json are merged into an existing top-level
     // JSON object; parameters named _<fieldname>_json are injected as a new top-level JSON object.
-    // Both conventions require the value to be a valid JSON object string.
+    // Both conventions require the value to be a valid JSON object string and the target field name
+    // must be in STRUCTURED_OUTPUT_ALLOWED_FIELDS to prevent callers from overriding provider control fields.
     private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
         for (Map.Entry<String, String> entry : parameters.entrySet()) {
             String key = entry.getKey();
             if (key.startsWith("_") && key.endsWith("_additions_json")) {
                 String fieldName = key.substring(1, key.length() - "_additions_json".length());
-                payload = mergeIntoJsonObject(payload, fieldName, entry.getValue());
+                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
+                    payload = mergeIntoJsonObject(payload, fieldName, entry.getValue());
+                }
             } else if (key.startsWith("_") && key.endsWith("_json")) {
                 String fieldName = key.substring(1, key.length() - "_json".length());
-                payload = injectTopLevelJsonField(payload, fieldName, entry.getValue());
+                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
+                    payload = injectTopLevelJsonField(payload, fieldName, entry.getValue());
+                }
             }
         }
         return payload;
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -395,16 +395,19 @@ public class HttpConnector extends AbstractConnector {
     // Note: matched parameters are read but not removed from the map.
     private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
         JsonElement parsed = JsonParser.parseString(payload);
-        if (!parsed.isJsonObject()) return payload;
+        if (!parsed.isJsonObject())
+            return payload;
         JsonObject body = parsed.getAsJsonObject();
         boolean modified = false;
 
         // Pass 1: _<field>_json — inject or replace a top-level field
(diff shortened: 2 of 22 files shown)


common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -366,11 +377,66 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
                     payload = jsonObject.toString();
                 }
             }
+            if (parameters != null && isJson(payload)) {
+                payload = injectStructuredOutputParams(parameters, payload);
```
ylwu-amzn: **Blocking — `getAsJsonObject()` will throw on top-level JSON arrays.**

`isJson(payload)` returns `true` for both objects *and* arrays (it parses with `gson.fromJson(json, Object.class)`). The two helpers below (`injectTopLevelJsonField`, `mergeIntoJsonObject`) then unconditionally call `JsonParser.parseString(payload).getAsJsonObject()`, which throws `IllegalStateException: Not a JSON Object: [...]` if the payload is an array.

No current connector in this PR ships an array request body, but `injectStructuredOutputParams` is now in the generic `createPayload` path that runs for every HTTP connector — this is a latent foot-gun.

**Suggested fix:** narrow the precondition to require an object specifically, e.g. add a small `isJsonObject(payload)` check (parse once, verify `.isJsonObject()`), or do the parse-and-check at the top of `injectStructuredOutputParams` and bail early.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -366,11 +377,66 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+    private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            String key = entry.getKey();
+            if (key.startsWith("_") && key.endsWith("_additions_json")) {
+                String fieldName = key.substring(1, key.length() - "_additions_json".length());
+                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
+                    payload = mergeIntoJsonObject(payload, fieldName, entry.getValue());
+                }
+            } else if (key.startsWith("_") && key.endsWith("_json")) {
+                String fieldName = key.substring(1, key.length() - "_json".length());
```
ylwu-amzn: **Blocking — pathological keys throw `StringIndexOutOfBoundsException`.**

A parameter named exactly `"_additions_json"` (length 15) passes `startsWith("_") && endsWith("_additions_json")`, then this slice runs:

```java
key.substring(1, key.length() - "_additions_json".length())
// = "_additions_json".substring(1, 0) → throws StringIndexOutOfBoundsException
```

Same for a key of exactly `"_json"` on line 401. Neither matches the documented schema, but parameter names come from user-defined connector parameters, so a malformed key crashes the predict call.

**Suggested fix:** require a minimum key length before slicing, or extract the field name with a regex like `^_([^_].*)_(?:additions_)?json$` and skip if the group is empty.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -366,11 +377,66 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+                if (STRUCTURED_OUTPUT_ALLOWED_FIELDS.contains(fieldName)) {
+                    payload = injectTopLevelJsonField(payload, fieldName, entry.getValue());
+                }
+            }
+        }
+        return payload;
+    }
```
ylwu-amzn: **Blocking — non-deterministic ordering when both `_X_json` and `_X_additions_json` are present for the same field.**

`parameters` is a `Map<String, String>` (almost always `HashMap` in practice). If a caller sets both `_response_format_json` (replace) and `_response_format_additions_json` (merge) for the same allowlisted field, the final payload depends on `HashMap` iteration order — which is unstable across JDK versions and JIT-hostile to debug. None of the current providers exercise this combination, but the API surface allows it and there's no documented precedence.

**Suggested fix:** decide on precedence (e.g. apply all `_<X>_json` replacements first, then all `_<X>_additions_json` merges; or reject the combination outright), enforce it by sorting/grouping the matching entries before applying, and document it in the method-level comment.

ylwu-amzn: **Nit — re-parses the payload on every matching parameter.**

For each match, `injectTopLevelJsonField` / `mergeIntoJsonObject` parses the (mutated) payload string back into a `JsonObject` and re-serializes. With one or two structured-output parameters this is fine, but as a future-proofing nit: parse once at the top of the method, mutate the `JsonObject` in place across all matches, then `toString()` once at the end. Keeps the helpers free of the parse/serialize round-trip.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -61,6 +63,15 @@ public class HttpConnector extends AbstractConnector {
...
     public static final String LLM_INTERFACE_OPENAI_V1_CHAT_COMPLETIONS = "openai/v1/chat/completions";
 
+    // Allowlist of top-level JSON fields that _<field>_json / _<field>_additions_json parameters may inject.
+    // Restricts injection to structured-output fields only; prevents callers from overriding
+    // provider control fields such as "model", "stream", or "max_tokens".
+    static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
+        .of(
+            "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
+            "generationConfig"  // Google Gemini
+        );
```
ylwu-amzn, line 76: Future-maintenance note: the allowlist here, the per-provider schema constants in `MemoryContainerConstants`, and the URL → parameter-name mapping in `MemoryContainerHelper.schemaForUrl` are coupled by convention only. Adding a 7th provider means coordinated edits across three files with no compile-time safety net.

Not blocking, but worth considering a small registry/enum that bundles `(URL predicate, injection-parameter name, schema constant, allowlisted field)` so additions are mechanical and discoverable. Could be a follow-up.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +507,77 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+            return Map.of("_response_format_json", FACTS_EXTRACTION_COHERE_RESPONSE_FORMAT_JSON);
+        }
+        // "/v1/chat/completions" is the de-facto path for OpenAI-compatible servers (Ollama, vLLM,
+        // LM Studio, etc.) that do not contain "openai" or "deepseek" in their URL. Matching on the
+        // path alone is intentional: supports_structured_output must be explicitly set to true on the
+        // connector action, which requires admin access, so an unintended match only results in an
+        // extra response_format field being sent — the upstream server will ignore or reject it safely.
+        if (lower.contains("openai") || lower.contains("deepseek") || lower.contains("/v1/chat/completions")) {
+            return Map.of("_response_format_json", FACTS_EXTRACTION_OPENAI_RESPONSE_FORMAT_JSON);
+        }
```
ylwu-amzn: **Medium — substring matching is broader than the comment claims, and the "falls back gracefully" promise has limits.**

`lower.contains("openai")` matches OpenAI, Azure (intentional), but also any internal proxy/path/CNAME containing the substring. The comment defends this by noting `supports_structured_output` must be explicitly set by an admin and that mismatched schemas are usually no-ops.

Two concerns:
1. If a future provider needs a different schema and happens to have "openai" in a docs URL or hostname, this would silently mis-route.
2. "Falls back gracefully" isn't quite right when the upstream provider returns 4xx — that surfaces as a failed predict, not a fallback. Worth tightening the comment to say so.

**Suggested:** parse the URL host + path and match on tokens rather than naive `contains`, or at minimum factor the URL fragments into named constants so the precedence (Cohere v2 → Gemini → OpenAI/DeepSeek/`/v1/chat/completions`) is explicit.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +507,77 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+        if (connector.getActions() == null) {
+            return Map.of();
+        }
+        return connector
+            .getActions()
+            .stream()
+            .filter(a -> ConnectorAction.ActionType.PREDICT.equals(a.getActionType()) && a.isSupportsStructuredOutput())
+            .findFirst()
+            .map(a -> schemaForUrl(a.getUrl()))
+            .orElse(Map.of());
```
ylwu-amzn, line 580: Medium / observation: `findFirst()` over PREDICT actions is non-deterministic when a connector has multiple named PREDICT actions (the codebase supports this — see `HttpConnector.findAction`). For fact-extraction it's effectively always one action, but the assumption is implicit. A short comment noting "first matching PREDICT action wins; assumes one PREDICT per connector for fact extraction" would prevent surprises if someone configures multiple.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +507,77 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+                } else if (mlModel.getConnectorId() != null) {
+                    modelManager
+                        .getConnector(
+                            mlModel.getConnectorId(),
+                            null,
+                            ActionListener.wrap(c -> listener.onResponse(schemaForConnector(c)), e -> {
+                                log.warn("Failed to fetch connector {} for structured output detection", mlModel.getConnectorId(), e);
+                                listener.onResponse(Map.of());
+                            })
+                        );
```
ylwu-amzn, line 558: Question — when `mlModel.getConnector()` is null and we fall through to `modelManager.getConnector(connectorId, null, ...)`, the secondary async call is dispatched *after* `runBefore(context::restore)` has already restored the original (user) context.

If `getConnector` requires system-context permissions, this could fail in security-enabled clusters with non-admin users. Other methods in this file stash once at the top level and rely on inner methods to stash again — does `modelManager.getConnector` do that internally? If so, a one-line code comment to that effect would help; if not, the secondary call should be wrapped in its own stashed context.

Flagging as a question, not a known bug — I didn't trace `getConnector`'s implementation.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -200,6 +200,68 @@ public class MemoryContainerConstants {
...
+                "type": "object",
+                "properties": {
+                    "facts": {
+                        "type": "array",
+                        "items": {"type": "string"}
+                    }
+                },
+                "required": ["facts"]
+            }
+        }""";
```
ylwu-amzn, line 268: Nit — the OpenAI schema (above) sets `"strict": true` and `"additionalProperties": false`, locking the response to exactly `{facts: [string]}`. The Gemini `responseSchema` here doesn't have an equivalent constraint, so Gemini can in principle return extra keys. The downstream parser only reads `facts`, so it's not a correctness bug, but worth aligning for consistency (and to keep the constrained-decoding promise tight).

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/MemoryProcessingServiceTests.java
```diff
@@ -1105,6 +1112,167 @@ public void testMultiTurnConversationHandling() {
...
+            RemoteInferenceInputDataSet dataset = (RemoteInferenceInputDataSet) request.getMlInput().getInputDataset();
+            Map<String, String> parameters = dataset.getParameters();
+
+            assertFalse(
+                "_response_format_json must NOT be present for prompt-based connector",
+                parameters.containsKey("_response_format_json")
+            );
+            assertTrue(
+                "JSON enforcement message MUST be in user_prompt for prompt-based connector",
+                parameters.get("user_prompt").contains("Respond NOW with ONE LINE of valid JSON ONLY")
```
ylwu-amzn: Nit (also applies to lines 1002, 1139, 1256): asserting via the raw literal `"Respond NOW with ONE LINE of valid JSON ONLY"` duplicates a substring of `JSON_ENFORCEMENT_MESSAGE`. If the wording ever changes, four tests break at once.

Reference `JSON_ENFORCEMENT_MESSAGE` directly (it's already imported), or extract a sentinel substring constant. Minor.

docs/remote_inference_blueprints/open_ai_connector_chat_blueprint.md
```diff
@@ -20,6 +20,7 @@ POST /_plugins/_ml/connectors/_create
       "action_type": "predict",
       "method": "POST",
       "url": "https://${parameters.endpoint}/v1/chat/completions",
+      "supports_structured_output": true,
```
ylwu-amzn, line 25: Doc gap (also affects the Azure / DeepSeek / Ollama blueprint diffs in this PR): the `"supports_structured_output": true` line is added with no explanation. New users copy-pasting will not know what this does, when to set it, or what happens with currently-unimplemented providers (Anthropic, Bedrock Converse).

`docs/tutorials/remote_inference.md` only got 1 line of churn in this PR, so the explanation isn't there yet either. Suggest either a one-line inline comment in each blueprint or a dedicated section in `remote_inference.md` that's cross-referenced from each blueprint.

Separately: Cohere v2 and Gemini are now supported but their blueprints (if present in the repo) didn't get the `supports_structured_output: true` flag for consistency.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -366,11 +377,66 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+            }
             return (T) payload;
         }
         return (T) parameters.get("http_body");
     }
 
+    // Convention: parameters named _<fieldname>_additions_json are merged into an existing top-level
+    // JSON object; parameters named _<fieldname>_json are injected as a new top-level JSON object.
+    // Both conventions require the value to be a valid JSON object string and the target field name
+    // must be in STRUCTURED_OUTPUT_ALLOWED_FIELDS to prevent callers from overriding provider control fields.
```
ylwu-amzn: Nit — drift between PR description and code. The PR description says:

> `HttpConnector.injectStructuredOutputParams` — strips `_<X>_json` / `_<X>_additions_json` parameters and injects them as top-level fields...

But the implementation reads the parameters and *doesn't* remove them from the map. They linger and remain visible to anything downstream that consumes `parameters` (logging, metrics, downstream processors).

Probably fine to keep them in the map (mutating a caller-owned map is its own can of worms), but please update the PR description / method-level comment so future readers don't go hunting for the strip step.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
ehavener: I checked the implementation of getConnector. It opens its own stashContext() before the async index read and restores it inside the whenComplete callback before notifying the listener: 
`try (ThreadContext.StoredContext ctx = client.threadPool().getThreadContext().stashContext()) {`

This is consistent with how it's called elsewhere in MLModelManager e.g.
```
getConnector(mlModel.getConnectorId(), mlModel.getTenantId(), ActionListener.wrap(connector -> {
    mlModel.setConnector(connector);
    setupParamsAndPredictable(modelId, mlModel, initModelActionListener);
    log.info("Completed setting connector {} in the model {}", mlModel.getConnectorId(), modelId);
}, wrappedListener::onFailure));
```

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -212,6 +240,9 @@ public static ConnectorAction parse(XContentParser parser) throws IOException {
                 case ACTION_POST_PROCESS_FUNCTION:
                     postProcessFunction = parser.text();
                     break;
+                case SUPPORTS_STRUCTURED_OUTPUT_FIELD:
+                    supportsStructuredOutput = parser.booleanValue();
+                    break;
                 default:
```
ylwu-amzn, line 246: **High-priority — downgrade BWC: the new field is silently dropped on XContent parse by 3.6 nodes.**

The wire-stream BWC at line 138 is correctly gated on `VERSION_3_7_0` and tested. But connectors are also persisted as JSON to `.plugins-ml-connector` and re-parsed on read. After a 3.7 → 3.6 rolling downgrade, a 3.6 node reading a connector doc with `"supports_structured_output": true` falls through this `default: parser.skipChildren()` branch and silently drops the flag — structured output stops working with no log line.

Not a blocker (downgrades are the cluster admin's responsibility), but worth:
- A release-note line in `release-notes/opensearch-ml-commons.release-notes-3.7.0.0.md` calling out that downgrade clears `supports_structured_output`.
- Optionally `log.debug("Skipping unknown connector_action field: {}", fieldName)` on the default case so this is at least visible in logs.

For reference: this is purely an XContent (index) BWC concern — the wire-stream path at lines 134-139 is fine.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +530,125 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+        // and other OpenAI-compatible servers whose hostnames contain neither "openai" nor "deepseek".
+        // supports_structured_output must be explicitly set to true on the connector action
+        // (requires admin access), so an unintended match only results in an extra response_format
+        // field being sent. Note: if the upstream provider rejects the extra field with a 4xx,
+        // that surfaces as a failed predict call, not a silent fallback.
+        if (hostHasSegment(host, URL_TOKEN_OPENAI)
+            || pathHasSegment(path, URL_TOKEN_OPENAI)
+            || hostHasSegment(host, URL_TOKEN_DEEPSEEK)
+            || path.contains(URL_TOKEN_OPENAI_COMPAT_PATH)) {
+            return Map.of("_response_format_json", FACTS_EXTRACTION_OPENAI_RESPONSE_FORMAT_JSON);
```
ylwu-amzn, line 631: **High-priority — schema-leak via misconfigured connector URL (defense-in-depth).**

This routes `supports_structured_output: true` connectors to the OpenAI schema based on token-equality matching. An admin who sets the flag on a connector pointing at `https://openai.attacker.com/v1/chat/completions` (or any URL where the host contains the literal label `openai`, or the path contains the literal segment `/v1/chat/completions`) will inject the OpenAI schema into the outbound request body.

The inline comment above correctly notes the threat model ("supports_structured_output must be explicitly set to true... requires admin access, so an unintended match only results in an extra response_format field being sent"). The schema doesn't contain secrets, so this isn't a credential-exposure issue. But it's still a small information disclosure to whatever endpoint the connector points at.

Cheap defense-in-depth: cross-reference the resolved URL against `plugins.ml_commons.trusted_connector_endpoints_regex` *before* deriving the provider. If an admin sets the flag on a connector pointing at a non-allowlisted host, fail fast at predict time rather than silently leak. This dovetails with the work in PR #4826.

Alternatively, tighten matching to full-host equality for the public providers: `host.equals("api.openai.com")` etc. That removes the `openai.attacker.com` class of false matches entirely.

Not blocking since both connector creation and the flag are admin-only.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -380,11 +392,76 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+        // Pass 2: _<field>_additions_json — merge into an existing top-level object
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            String fieldName = allowedFieldName(entry.getKey(), "_additions_json");
+            JsonObject additions = asJsonObject(entry.getValue());
+            if (fieldName == null || additions == null)
+                continue;
+            JsonObject target = body.has(fieldName) && body.get(fieldName).isJsonObject()
+                ? body.getAsJsonObject(fieldName)
+                : new JsonObject();
+            additions.entrySet().forEach(e -> target.add(e.getKey(), e.getValue()));
```
ylwu-amzn, line 438: **High-priority — `_<field>_additions_json` "merge" overwrites existing keys; the JavaDoc says "merge into" which is ambiguous.**

`JsonObject.add(key, value)` *replaces* when the key already exists. So if a connector author has `generationConfig: {"temperature": 0.7, "maxOutputTokens": 1024}` in the request body template, and the additions are `{"temperature": 0.99, "responseMimeType": "application/json"}`, the user's `temperature: 0.7` is silently overridden by `0.99`.

The current `FACTS_EXTRACTION_GEMINI_GENERATION_CONFIG_JSON` constant doesn't include any keys that overlap with what a sensible connector author would pre-populate, so it's not a problem today. But future schemas could collide silently.

Two options:
1. Update the JavaDoc on lines 403-409 to be explicit: "On key collision, addition values *replace* existing values in the target object — the operation is not strictly additive." This is the lowest-friction option.
2. Change semantics to genuinely additive (`if (!target.has(key)) target.add(key, value)`) and force schema authors to use `_<field>_json` (full replace) when they need overrides. More disruptive but truer to the name `_additions_`.

A test like `createPayload_AdditionsJson_DoesNotOverrideExistingKey` would also pin the chosen behavior.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +530,125 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+                    "_structured_output_result_path",
+                    BEDROCK_STRUCTURED_OUTPUT_RESULT_PATH
+                );
+        }
+        if (hostHasSegment(host, URL_TOKEN_GOOGLEAPIS)) {
+            return Map.of("_generationConfig_additions_json", FACTS_EXTRACTION_GEMINI_GENERATION_CONFIG_JSON);
+        }
+        // Cohere structured output (json_schema type) requires the v2 Chat API (/v2/chat).
+        // The legacy /v1/chat endpoint only supports json_object and ignores json_schema.
+        if (hostHasSegment(host, URL_TOKEN_COHERE) && pathHasSegment(path, URL_TOKEN_V2_PATH_SEGMENT)) {
```
ylwu-amzn, line 614: Cohere v2 path matching accepts more URLs than just chat. `pathHasSegment(path, URL_TOKEN_V2_PATH_SEGMENT)` matches any URL with a `/v2/` segment, so `https://api.cohere.com/v2/embed`, `/v2/rerank`, etc., all classify as "Cohere v2 chat" and receive a `response_format` field that those endpoints don't understand. Cohere will likely 400, surfacing as a confusing fact-extraction failure for users who configured a non-chat Cohere connector with `supports_structured_output: true`.

Fix: tighten the path check to require both `v2` and `chat` segments, or check for the literal substring `/v2/chat` (similar to how `URL_TOKEN_OPENAI_COMPAT_PATH` is used for the OpenAI-compatible endpoints below).

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -380,11 +392,76 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
...
+    private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
+        JsonElement parsed = JsonParser.parseString(payload);
+        if (!parsed.isJsonObject())
+            return payload;
+        JsonObject body = parsed.getAsJsonObject();
+        boolean modified = false;
+
+        // Pass 1: _<field>_json — inject or replace a top-level field
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            if (entry.getKey().endsWith("_additions_json"))
```
ylwu-amzn, line 419: Defense-in-depth, but fragile: this `endsWith("_additions_json")` skip is what prevents Pass 1 from also processing `_<field>_additions_json` keys. Without it, Pass 1 would call `allowedFieldName("_response_format_additions_json", "_json")` which slices to `"response_format_additions"` — not in the allowlist, so it returns `null` and gets skipped anyway. So the skip is redundant *today* given the allowlist, but if someone removes this line during a refactor (thinking it's redundant) and *also* extends the allowlist later (e.g., adds `"response_format_additions"` for some reason), the bug surface widens silently.

Suggest either:
- A short comment explaining the layering: `// Without this skip, allowlist still prevents Pass 1 from processing *_additions_json keys, but explicit is safer.`
- Or refactor the loop to classify each key once into one of {replace, merge, ignore} so the dispatch is structural rather than implicit.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -63,6 +65,16 @@ public class HttpConnector extends AbstractConnector {
...
 
+    // Allowlist of top-level JSON fields that _<field>_json / _<field>_additions_json parameters may inject.
+    // Restricts injection to structured-output fields only; prevents callers from overriding
+    // provider control fields such as "model", "stream", or "max_tokens".
+    static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
+        .of(
+            "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
+            "generationConfig", // Google Gemini
+            "toolConfig"        // Amazon Bedrock Converse
+        );
```
ylwu-amzn, line 76: The 3-element allowlist here is the wire-layer enforcement. The corresponding constants in `MemoryContainerConstants.FACTS_EXTRACTION_*` (`OPENAI`, `COHERE`, `GEMINI`, `BEDROCK_CONVERSE`) are tightly coupled to these field names — adding a new provider requires extending both files in lockstep, otherwise the schema constant exists but the wire layer silently drops the injection.

Cheap fix: add a one-liner cross-reference here, e.g.:
```java
// Keep in sync with MemoryContainerConstants.FACTS_EXTRACTION_*: each constant's top-level
// field name must appear in this allowlist to actually reach the outgoing request body.
```
…and a similar pointer in `MemoryContainerConstants` near the schema constants.

common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -380,11 +392,76 @@ public <T> T createPayload(String action, Map<String, String> parameters) {
                     payload = jsonObject.toString();
                 }
             }
+            if (parameters != null && isJson(payload) && connectorAction.get().isSupportsStructuredOutput()) {
+                payload = injectStructuredOutputParams(parameters, payload);
```
ylwu-amzn, line 396: Confirmed gate is correct: `connectorAction.get().isSupportsStructuredOutput()` is the admin-only flag, so a non-admin user cannot bypass the allowlist by passing `_response_format_json` directly in `_predict`. The only way the injection runs is if an admin set the flag on the action at create/update time. Worth a unit test that verifies a connector with the flag set to `false` ignores `_response_format_json` even when present in parameters — couldn't find that test in the new HttpConnectorTest cases. It's the inverse of `_DefaultsFalse` and explicit about the gate.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -497,6 +530,125 @@ public void countContainersWithPrefix(String indexPrefix, String tenantId, Actio
...
+                listener.onResponse(Map.of());
+            }), context::restore));
+        }
+    }
+
+    private Map<String, String> schemaForConnector(Connector connector) {
+        if (connector.getActions() == null) {
+            return Map.of();
+        }
+        // first matching PREDICT action wins; assumes one PREDICT per connector for fact extraction
```
ylwu-amzn, line 573: Nit: the comment correctly notes "first matching PREDICT action wins; assumes one PREDICT per connector for fact extraction." If a connector has multiple named PREDICT actions (per the action-name framework that landed in 3.5.0), only the first one with `supports_structured_output: true` is considered. That's fine for agentic memory's current single-action assumption. Worth converting the comment into a `// TODO:` if multi-action structured output is on the roadmap, so it's grep-able.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -76,15 +88,36 @@
...
+    private static final String URL_TOKEN_DEEPSEEK = "deepseek";
+    // Multi-segment path — used with path.contains() on a query-stripped path, which is safe.
+    private static final String URL_TOKEN_OPENAI_COMPAT_PATH = "/v1/chat/completions";
+    private static final String URL_TOKEN_AMAZONAWS = "amazonaws";
+    private static final String URL_TOKEN_CONVERSE_SEGMENT = "converse";
+
+    // Captures (1) host and (2) path from a URL, stopping before '?' or '#'.
+    // Handles connector template URLs such as https://${parameters.endpoint}/openai/...
+    // where the host may be a placeholder — in that case path matching takes over.
+    private static final Pattern URL_PARTS = Pattern.compile("^[^:]+://([^/?#]*)(/?[^?#]*)");
```
ylwu-amzn, line 108: Edge case: this regex captures the entire URL authority into group 1, including userinfo. So `https://user:openai@attacker.example.com/path` gives `host = "user:openai@attacker.example.com"`. After `host.split("\\.", -1)`, you get `["user:openai@attacker", "example", "com"]` — none of those equals the bare token `"openai"`, so the current implementation is safe. But this isn't tested anywhere I could find. Worth one targeted test (`testGetStructuredOutputParameters_UrlWithUserinfo_ReturnsEmptyMap` or similar) to pin the behavior so a future regex change doesn't accidentally introduce a host-confusion bug.

ylwu-amzn (approved):
No blocking issues. Approving to catch the 3.7 release; please file a follow-up PR for the inline feedback.

PR #4836, opened 2026-05-27
Bound AgentTool recursion depth to prevent self-referencing chains

rithinpullela:
## Summary
- Add an `_agent_call_depth` counter that AgentTool stamps on its outbound `AgentMLInput` and reads back on entry, capping the chain at one nested AgentTool hop (`MAX_AGENT_CALL_DEPTH = 1`).
- Reject deeper chains with `OpenSearchStatusException(BAD_REQUEST)` so the failure surfaces as HTTP 400 (configuration error) rather than 500.
- Cover the bound with unit tests in `AgentToolTests` (initial stamp, blocks at the cap, blocks past the cap, ignores garbage values) and an end-to-end IT in `RestMLFlowAgentIT` that registers a flow agent, rewires its AgentTool to point at itself, and asserts the execute call fails with the recursion-depth message.

## Test plan
- [x] `./gradlew :opensearch-ml-algorithms:test --tests AgentToolTests`
- [x] `./gradlew :opensearch-ml-algorithms:test --tests AbstractV2AgentRunnerTest` (no V2 path changes after revert; ensures no regression)
- [ ] `./gradlew :opensearch-ml-plugin:integTest --tests RestMLFlowAgentIT`

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
@@ -9,8 +9,10 @@ import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
 
 import java.util.Map;
 
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.execute.agent.AgentMLInput;
@@ -37,12 +39,9 @@ import lombok.extern.log4j.Log4j2;
 public class AgentTool implements Tool {
     public static final String TYPE = "AgentTool";
 
-    // Reserved framework parameter that tracks how many AgentTool hops we have made
-    // along the current execution chain. It is read+incremented on each hop so that
-    // self-loops and chained recursion (A->A, A->B->A, A->B->C->...) are bounded.
+    // Tracks AgentTool hops on the current chain to bound nested agent recursion.
     public static final String AGENT_CALL_DEPTH_FIELD = "_agent_call_depth";
-    // The root agent may call one nested agent via AgentTool, but that nested agent must
-    // not start another AgentTool chain. So depth 0 -> 1 is allowed and depth >= 1 is blocked.
+    // Allows depth 0 -> 1; a nested AgentTool call (depth >= 1) is blocked.
     public static final int MAX_AGENT_CALL_DEPTH = 1;
 
     private final Client client;
@@ -86,18 +85,18 @@ public class AgentTool implements Tool {
             if (currentDepth >= MAX_AGENT_CALL_DEPTH) {
                 listener
                     .onFailure(
-                        new IllegalStateException(
+                        new OpenSearchStatusException(
                             "AgentTool recursion depth exceeded the maximum allowed ("
                                 + MAX_AGENT_CALL_DEPTH
-                                + "). Aborting nested agent execution to prevent infinite recursion."
+                                + "). Aborting nested agent execution to prevent infinite recursion.",
+                            RestStatus.BAD_REQUEST
                         )
                     );
                 return;
             }
 
             Map<String, String> extractedParameters = ToolUtils.extractInputParameters(parameters, attributes);
-            // Stamp incremented depth on the parameters that travel to the nested agent so
-            // it is observable by the next AgentTool invocation in the chain.
+            // Propagate incremented depth to the nested agent so subsequent hops can see it.
             extractedParameters.put(AGENT_CALL_DEPTH_FIELD, String.valueOf(currentDepth + 1));
             String tenantId = parameters.get(TENANT_ID_FIELD);
             AgentMLInput agentMLInput = AgentMLInput
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
@@ -36,6 +36,15 @@ import lombok.extern.log4j.Log4j2;
 @ToolAnnotation(AgentTool.TYPE)
 public class AgentTool implements Tool {
     public static final String TYPE = "AgentTool";
+
+    // Reserved framework parameter that tracks how many AgentTool hops we have made
+    // along the current execution chain. It is read+incremented on each hop so that
+    // self-loops and chained recursion (A->A, A->B->A, A->B->C->...) are bounded.
+    public static final String AGENT_CALL_DEPTH_FIELD = "_agent_call_depth";
+    // The root agent may call one nested agent via AgentTool, but that nested agent must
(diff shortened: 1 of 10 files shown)


ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -72,7 +80,24 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                                + "). Aborting nested agent execution to prevent infinite recursion.",
+                            RestStatus.BAD_REQUEST
+                        )
+                    );
+                return;
+            }
+
             Map<String, String> extractedParameters = ToolUtils.extractInputParameters(parameters, attributes);
+            // Propagate incremented depth to the nested agent so subsequent hops can see it.
+            extractedParameters.put(AGENT_CALL_DEPTH_FIELD, String.valueOf(currentDepth + 1));
```
ylwu-amzn, line 103: **High-priority — Plan-Execute-Reflect bypasses this guard.**

`MLPlanExecuteAndReflectAgentRunner.java:580-608` builds the inner ReAct sub-agent's `reactParams` field-by-field and dispatches via `client.execute(MLExecuteTaskAction.INSTANCE, ...)` directly — it does **not** go through `AgentTool`. So `_agent_call_depth` from `allParams` is never forwarded into `reactParams`, and AgentTool inside the inner ReAct agent starts at depth=0 again. A Plan-Execute agent A whose ReAct subagent uses an AgentTool→A (or →B→A) re-enters with a fresh budget every iteration.

The PR title ("Bound AgentTool recursion depth to prevent self-referencing chains") implies this is closed; today it isn't, for one of the four agent types.

Suggested fix — one-liner around line 595 of `MLPlanExecuteAndReflectAgentRunner`:
```java
if (allParams.containsKey(AgentTool.AGENT_CALL_DEPTH_FIELD)) {
    reactParams.put(AgentTool.AGENT_CALL_DEPTH_FIELD, allParams.get(AgentTool.AGENT_CALL_DEPTH_FIELD));
}
```
Alternatively, scope this PR to FLOW / CONVERSATIONAL / CONVERSATIONAL_FLOW explicitly in the description and file a follow-up for Plan-Execute.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -36,6 +38,12 @@
 @ToolAnnotation(AgentTool.TYPE)
 public class AgentTool implements Tool {
     public static final String TYPE = "AgentTool";
+
+    // Tracks AgentTool hops on the current chain to bound nested agent recursion.
+    public static final String AGENT_CALL_DEPTH_FIELD = "_agent_call_depth";
+    // Allows depth 0 -> 1; a nested AgentTool call (depth >= 1) is blocked.
+    public static final int MAX_AGENT_CALL_DEPTH = 1;
```
ylwu-amzn: Low — `MAX_AGENT_CALL_DEPTH = 1` is hard-coded. That's a defensible default but it preempts legitimate two-deep flows (orchestrator → router-agent → leaf-agent). Customers needing depth=2 can't get it without a recompile.

Consider exposing this as a cluster setting (e.g. `plugins.ml_commons.agent.max_call_depth`, default 1, range 1–5) following the pattern at `MLCommonsSettings.java:357`. Out of scope for this PR if you'd rather ship the bound first and configurability later, but worth a follow-up issue.

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -72,7 +80,24 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+            int currentDepth = readDepth(parameters);
+            if (currentDepth >= MAX_AGENT_CALL_DEPTH) {
+                listener
+                    .onFailure(
+                        new OpenSearchStatusException(
+                            "AgentTool recursion depth exceeded the maximum allowed ("
+                                + MAX_AGENT_CALL_DEPTH
+                                + "). Aborting nested agent execution to prevent infinite recursion.",
+                            RestStatus.BAD_REQUEST
+                        )
```
ylwu-amzn, line 96: Low — error message wording.

"Aborting nested agent execution to prevent infinite recursion" reads like an internal-state log line. A user-facing 400 should explain *what they did wrong*, e.g.:

> "AgentTool nested call depth exceeded the maximum (1). Check that your agent's tools do not transitively reference the agent itself."

Note: `RestMLFlowAgentIT.java:69` asserts on `body.contains("AgentTool recursion depth")`, so any rewording must keep that substring (or the test must be updated in lockstep).

ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/AgentTool.java
```diff
@@ -95,6 +120,24 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+            return 0;
+        }
+        String raw = parameters.get(AGENT_CALL_DEPTH_FIELD);
+        if (raw == null || raw.isBlank()) {
+            return 0;
+        }
+        try {
+            int depth = Integer.parseInt(raw);
+            // Defensive: a negative value would let an attacker reset the counter.
+            return Math.max(depth, 0);
```
ylwu-amzn, line 137: Low — the comment on the next line says "Defensive: a negative value would let an attacker reset the counter" and the code does `Math.max(depth, 0)`. There's a unit test for `"not-a-number"` but none for `"-5"` (or any negative numeric). One-liner test in `AgentToolTests` to lock in the documented defense:

```java
parameters.put(AGENT_CALL_DEPTH_FIELD, "-5");
tool.run(parameters, listener);
// assert chain still starts at "1" (negative clamped to 0, then incremented)
```

plugin/src/test/java/org/opensearch/ml/rest/RestMLFlowAgentIT.java
```diff
@@ -50,6 +52,23 @@ public void testAgentListIndexTool() throws IOException {
...
+        assertNotNull(agentId);
+
+        // Rewire the AgentTool to point at the agent itself, so executing it tries to recurse.
+        Response updateResponse = updateAgentToolToTargetSelf(agentId);
+        assertEquals(200, updateResponse.getStatusLine().getStatusCode());
+
+        // Executing must fail fast with our recursion-depth guard rather than recursing forever.
+        ResponseException ex = expectThrows(ResponseException.class, () -> executeAgent(agentId, Map.of("question", "\"hi\"")));
+        String body = EntityUtils.toString(ex.getResponse().getEntity());
+        assertTrue("Expected recursion-depth error, got: " + body, body.contains("AgentTool recursion depth"));
```
ylwu-amzn: Low — robustness of the assertion.

The test only verifies the body contains `"AgentTool recursion depth"`. If something else fails the execute call before reaching the depth guard (e.g. the `placeholder-will-be-replaced` agent_id rejection from a future validator, a 404 from get-agent, etc.), the test could pass on the wrong path or fail with a misleading message.

Two small hardenings:
1. Also assert the HTTP status is 400 (the new `BAD_REQUEST`), not 404/500: `assertEquals(400, ex.getResponse().getStatusLine().getStatusCode());`
2. The PR description marks this IT as unverified (`[ ] ./gradlew :opensearch-ml-plugin:integTest --tests RestMLFlowAgentIT`). Worth a local run before merge — the placeholder-id register path can be brittle if `MLAgentUpdateInput.parse` ever tightens up validation.

PR #4851, opened 2026-06-09
Fix flaky IT: testChatAgentWithMcpStreamableHttpConnector, add timeout

rithinpullela:
## Summary
- Fix flaky `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector` (two independent flake sources)

**Flake 1 — Bedrock 400 "The provided request is not valid" (registration race)**
- `TransportMcpToolsRegisterOnNodesAction.registerToolsOnNode` uses `Flux...subscribe()` (fire-and-forget) and returns success before the reactive `addTool` Mono completes, so HTTP 200 from `/_plugins/_ml/mcp/tools/_register` does not guarantee the tool is servable
- Worse, `/_plugins/_ml/mcp/tools/_list` reads the `.plugins-ml-mcp-tools` system index, while the MCP server answers `tools/list` from a per-node in-memory registry (synced from the index only every 10s) — so polling `/_list` (earlier commits here) can pass while the tool is still not servable
- The agent's MCP client then discovers zero tools, and `BedrockConverseFunctionCalling` renders `tool_configs` as `, "toolConfig": {"tools": []}` — verified empirically against the Converse API that an empty tools array returns exactly this generic 400 (all other malformed request shapes return specific error messages)
- Fix: poll the real `/_plugins/_ml/mcp` endpoint with a JSON-RPC `tools/list` call (30s timeout, 500ms cadence) — the same call the agent's MCP client makes — before continuing setup

**Flake 2 — LLM answers with a timeout apology instead of the index list**
- The MCP connector loops back into the same cluster over HTTP; under CI load the default 30s read timeout is too tight and the loopback tool call intermittently times out
- Fix: set `client_config` connection/read timeouts to 120s on the MCP connector

## Test plan
- [x] Ran the IT 4 times locally with valid AWS Bedrock creds against the earlier revision — 4/4 pass (1m27s first cold run, ~25s on warm cluster)
- [ ] Confirm the updated poll + timeout pass on CI

## Out of scope (follow-up)
The deeper product fixes: `TransportMcpToolsRegisterOnNodesAction:114-144` should await the Mono before returning success; agents that resolve zero MCP tools silently continue and send Bedrock an empty toolConfig (an opaque 400 for any real user hitting the same race) — should fail loudly or omit `toolConfig`. This PR only stabilizes the IT.

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
@@ -78,16 +78,18 @@ public class RestChatAgentWithMcpConnectorIT extends MLCommonsRestTestCase {
         assertEquals(200, registerResponse.getStatusLine().getStatusCode());
 
         // MCP per-node tool registration is fire-and-forget, so HTTP 200 doesn't guarantee
-        // the tool is visible yet. Poll /_list until ListIndexTool shows up (15s max).
-        Thread.sleep(500);
-        long deadline = System.currentTimeMillis() + 15_000;
-        while (System.currentTimeMillis() < deadline) {
-            Response listResponse = TestHelper
-                .makeRequest(client(), "GET", "/_plugins/_ml/mcp/tools/_list", null, "", null);
+        // the tool is visible yet. Poll /_list until ListIndexTool shows up (20 x 500ms = 10s max).
+        boolean toolVisible = false;
+        for (int attempt = 0; attempt < 20; attempt++) {
+            Thread.sleep(500);
+            Response listResponse = TestHelper.makeRequest(client(), "GET", "/_plugins/_ml/mcp/tools/_list", null, "", null);
             if (TestHelper.httpEntityToString(listResponse.getEntity()).contains("ListIndexTool")) {
+                toolVisible = true;
                 break;
             }
-            Thread.sleep(500);
+        }
+        if (!toolVisible) {
+            throw new AssertionError("ListIndexTool did not become visible in /_plugins/_ml/mcp/tools/_list within 10s");
         }
 
         ingestIrisData(irisIndex);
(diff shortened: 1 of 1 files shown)


plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
```diff
@@ -77,6 +77,21 @@ public void setup() throws IOException, ParseException, InterruptedException {
             );
         assertEquals(200, registerResponse.getStatusLine().getStatusCode());
 
+        // MCP per-node tool registration is fire-and-forget, so HTTP 200 doesn't guarantee
+        // the tool is visible yet. Poll /_list until ListIndexTool shows up (20 x 500ms = 10s max).
+        boolean toolVisible = false;
+        for (int attempt = 0; attempt < 20; attempt++) {
```
akolarkunnu: nit - The bug description says 15 seconds, but in code it's 10 seconds

akolarkunnu: There is OpenSearchTestCase.assertBusy() OpenSearchTestCase.assertBusyWithFixedSleepTime() APIs to handle this kind of scenarios. Code will be more neat and clean with that APIs.

rithinpullela: Done — switched to OpenSearchTestCase.assertBusyWithFixedSleepTime, which keeps the fixed 500ms cadence (assertBusy uses exponential backoff) and throws on timeout, so the manual loop and flag are gone.

rithinpullela: Fixed — the wait is now 15s (TimeValue.timeValueSeconds(15)), matching the bug description.

PR #4858, opened 2026-06-14
Follow up for feat: Add structured output (constrained decoding) support for agentic memory fact extraction #4824

ehavener:
### Description
Applies follow up feedback comments on #4799 which was merged into OpenSearch 3.7

- High-priority — downgrade BWC: the new field is silently dropped on XContent parse by 3.6 nodes.
- Reminder comments to keep all of these schema constant variables defined in lockstep (URL predicate, injection-parameter name, schema constant, allowlisted field)
- Test that verifies a connector with the flag set to false ignores _response_format_json even when present in parameters
- Addresses: _<field>_additions_json "merge" overwrites existing keys; the JavaDoc says "merge into" which is ambiguous. Update test createPayload_AdditionsJson_DoesNotOverrideExistingKey
- Update testGetStructuredOutputParameters_ProviderTokenInUserinfo_ReturnsEmptyMap to test that URL_PARTS regex excludes userInfo for security
- Add TODO to update schemaForConnector return if multi-action connectors is implemented
- Tighten schemaForUrl cohere path check to require both v2 and chat segments, or check for the literal substring /v2/chat
- Tighten schemaForUrl openai host matching to: host.equals("api.openai.com") for security

### Related Issues
Original issue: [[FEATURE] Support structured output / constrained decoding for agentic memory fact extraction #4799](https://github.com/opensearch-project/ml-commons/issues/4799)

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -244,6 +244,7 @@ public class ConnectorAction implements ToXContentObject, Writeable {
                     supportsStructuredOutput = parser.booleanValue();
                     break;
                 default:
+                    logger.debug("Skipping unknown connector action field: {}", fieldName);
                     parser.skipChildren();
                     break;
             }
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -440,7 +440,10 @@ public class HttpConnector extends AbstractConnector {
             JsonObject target = body.has(fieldName) && body.get(fieldName).isJsonObject()
                 ? body.getAsJsonObject(fieldName)
                 : new JsonObject();
-            additions.entrySet().forEach(e -> { if (!target.has(e.getKey())) target.add(e.getKey(), e.getValue()); });
+            additions.entrySet().forEach(e -> {
+                if (!target.has(e.getKey()))
+                    target.add(e.getKey(), e.getValue());
+            });
             body.add(fieldName, target);
             modified = true;
         }
--- a/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
@@ -68,6 +68,8 @@ public class HttpConnector extends AbstractConnector {
     // Allowlist of top-level JSON fields that _<field>_json / _<field>_additions_json parameters may inject.
     // Restricts injection to structured-output fields only; prevents callers from overriding
     // provider control fields such as "model", "stream", or "max_tokens".
+    // Each entry corresponds to a FACTS_EXTRACTION_* constant in MemoryContainerConstants.
+    // Any new field added here must have a matching constant wired into MemoryContainerHelper.schemaForUrl.
     static final Set<String> STRUCTURED_OUTPUT_ALLOWED_FIELDS = Set
         .of(
             "response_format",  // OpenAI, Azure OpenAI, DeepSeek, Ollama, Cohere
@@ -406,6 +408,8 @@ public class HttpConnector extends AbstractConnector {
     // (b) a non-empty, allowlisted field name, and (c) a valid JSON object value string.
     // Replacements (_X_json) are applied before merges (_X_additions_json) so behaviour is
     // deterministic when both are present for the same field.
+    // Collision semantics: keys already present in the target object are preserved — the operation
+    // is strictly additive. Use _<field>_json (full replacement) to override existing keys.
     // Note: matched parameters are read but not removed from the map.
     private String injectStructuredOutputParams(Map<String, String> parameters, String payload) {
         JsonElement parsed = JsonParser.parseString(payload);
@@ -415,6 +419,7 @@ public class HttpConnector extends AbstractConnector {
         boolean modified = false;
 
         // Pass 1: _<field>_json — inject or replace a top-level field
+        // Without this skip, allowlist still prevents Pass 1 from processing *_additions_json keys, but explicit is safer.
         for (Map.Entry<String, String> entry : parameters.entrySet()) {
             if (entry.getKey().endsWith("_additions_json"))
                 continue;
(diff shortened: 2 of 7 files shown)


common/src/main/java/org/opensearch/ml/common/connector/HttpConnector.java
```diff
@@ -435,7 +440,10 @@ private String injectStructuredOutputParams(Map<String, String> parameters, Stri
             JsonObject target = body.has(fieldName) && body.get(fieldName).isJsonObject()
                 ? body.getAsJsonObject(fieldName)
                 : new JsonObject();
-            additions.entrySet().forEach(e -> target.add(e.getKey(), e.getValue()));
+            additions.entrySet().forEach(e -> {
+                if (!target.has(e.getKey()))
+                    target.add(e.getKey(), e.getValue());
+            });
```
ylwu-amzn: Additive-only merge can leave the payload internally inconsistent. For Gemini, `responseSchema` gets injected while a pinned `generationConfig.responseMimeType` (say `text/plain`) is preserved — Gemini rejects that combination, so fact extraction fails with a 400 and nothing logs why. The `_<field>_json` escape hatch the comment above suggests isn't reachable here either: `schemaForUrl` only ever emits `_generationConfig_additions_json`.

Suggest making it all-or-nothing per field — if any key collides, skip the injection for that field and `log.warn`, so the reason structured output didn't apply is visible.

plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java
```diff
@@ -611,22 +620,24 @@ private Map<String, String> schemaForUrl(String url) {
...
-        // (requires admin access), so an unintended match only results in an extra response_format
-        // field being sent. Note: if the upstream provider rejects the extra field with a 4xx,
-        // that surfaces as a failed predict call, not a silent fallback.
-        if (hostHasSegment(host, URL_TOKEN_OPENAI)
+        // and other OpenAI-compatible servers regardless of hostname.
+        if (host.equals(URL_HOST_OPENAI_API)
             || pathHasSegment(path, URL_TOKEN_OPENAI)
-            || hostHasSegment(host, URL_TOKEN_DEEPSEEK)
+            || host.equals(URL_HOST_DEEPSEEK_API)
             || path.contains(URL_TOKEN_OPENAI_COMPAT_PATH)) {
```
ylwu-amzn: The equality check doesn't close the label-confusion case: `https://openai.attacker.com/v1/chat/completions` still matches via `path.contains(URL_TOKEN_OPENAI_COMPAT_PATH)`, and the new test only covers `/api/chat`. Conversely, real `api.openai.com` / `api.deepseek.com` URLs match on that same path, so both equality checks are near-dead code (and `api.openai.com:443` now misses).

The bigger loss is the comment this hunk deleted — admin-only flag, mis-match only appends one field to an already admin-configured URL, provider 4xx surfaces as a failed predict rather than a silent fallback. That rationale is accurate; the new "closes the attack" framing isn't. Please keep the old one.

plugin/src/test/java/org/opensearch/ml/helper/MemoryContainerHelperTests.java
```diff
@@ -1107,6 +1125,63 @@ public void testGetStructuredOutputParameters_NoMatchingPredictAction_ReturnsEmp
...
+        assertTrue(captor.getValue().isEmpty());
+    }
+
+    public void testGetStructuredOutputParameters_ProviderTokenInUserinfo_ReturnsEmptyMap() {
+        // URL_PARTS captures the full authority (userinfo + host) in group 1. A provider token
+        // that appears only in the userinfo (e.g. "openai" in "user:openai@attacker.example.com")
+        // must not trigger a match. After splitting group 1 on '.', the userinfo segment is
+        // "user:openai@attacker" — which does not equal the bare token "openai" — so the host
+        // confusion attack is blocked. This test pins that behaviour so a future regex change
+        // does not accidentally introduce a host-confusion vulnerability.
```
ylwu-amzn: This passes because the URL matches no provider, not because userinfo is excluded — `URL_PARTS` group(1) includes userinfo, so the property the comment claims doesn't hold for the providers still using label matching:

`https://x.googleapis.com@attacker.example.com/v1/models/m:generateContent` → host = `x.googleapis.com@attacker.example.com`, `hostHasSegment(host, "googleapis")` is **true** → routes as Gemini. Same shape works for cohere (`api.cohere.com@...` + `/v2/chat`) and amazonaws.

Either strip userinfo before matching (`host.substring(host.indexOf('@') + 1)`), or use a URL of that shape here and assert the real behaviour.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -244,6 +244,7 @@ public static ConnectorAction parse(XContentParser parser) throws IOException {
                     supportsStructuredOutput = parser.booleanValue();
                     break;
                 default:
+                    logger.debug("Skipping unknown connector action field: {}", fieldName);
```
ylwu-amzn: 3.6 nodes run 3.6's parser, so this never fires on the downgrade path — it only helps a 3.7 node reading a field from a newer version. Fine as a diagnostic, but the release note is the actual mitigation for that bullet.

release-notes/opensearch-ml-commons.release-notes-3.7.0.0.md
```diff
@@ -12,6 +12,7 @@ Compatible with OpenSearch and OpenSearch Dashboards version 3.7.0
 
 * Add private IP security controls and ReDoS protection for ML connectors ([#4818](https://github.com/opensearch-project/ml-commons/pull/4818))
 * Add structured output (constrained decoding) support for agentic memory fact extraction ([#4824](https://github.com/opensearch-project/ml-commons/pull/4824))
+* **Downgrade note**: rolling back to 3.6 silently clears `supports_structured_output` on connector actions; structured output injection falls back to prompt enforcement until the cluster is upgraded again ([#4824](https://github.com/opensearch-project/ml-commons/pull/4824))
```
ylwu-amzn: `main` is 3.8.0-SNAPSHOT and the `3.7` branch is still taking backports — this line isn't there, so the people reading the 3.7 release notes won't see the warning. Worth backporting to `3.7` (it also links #4824 rather than this PR).

Issue #4727, opened 2026-03-19
[RFC] Working Memory Retention for Agentic Memory


b4sjoo:
## Summary

This RFC proposes a 2-stage retention and cleanup mechanism for working memory in agentic memory containers.

- **Stage 1** combines **query-time expiry filtering** with **opportunistic cleanup on write** to deliver immediate correctness with minimal infrastructure changes.
- **Stage 2** adds a **Job Scheduler periodic cleanup** to durably reclaim expired documents across all containers, including idle ones.

Together the two stages provide a complete retention lifecycle: expired working memory is hidden immediately, cleaned up best-effort on active containers, and guaranteed to be reclaimed by a background job.

## Problem Statement

Working memory in agentic memory containers currently grows without bound. There is no retention policy, no expiry filtering on reads, and no scheduled cleanup.

This causes three problems:

1. **Unbounded storage growth** — every working memory document persists indefinitely unless manually deleted.
2. **Stale data in query results** — expired working memory remains visible to read APIs, polluting context windows and degrading agent quality.
3. **Operational burden** — cleanup depends on ad hoc manual intervention with no centralized control or observability.

We need a retention mechanism that hides expired data from reads, eventually deletes it from storage, supports per-container configuration, and can be rolled out incrementally.

## Goals

- Add a per-container retention configuration for working memory.
- Make expired working memory invisible to all read APIs.
- Physically delete expired documents over time.
- Keep the initial implementation simple and low risk.
- Reuse existing infrastructure (delete-by-query, Job Scheduler) where practical.

## Non-Goals

- Retention for long-term memory, session memory, or history memory.
- Multi-tenancy isolation in this iteration.
- Re-architecting working memory into a different index layout or introducing index partitioning.
- Per-container cleanup cadence configuration.

## Current State

Today in the agentic memory system:

- Working memory documents carry `created_time` and `last_updated_time` fields. Retention is evaluated against `last_updated_time`.
- Each memory container stores working memory in a dedicated index (or shared index scoped by `memory_container_id`).
- Delete-by-query is already supported in the memory code path via `TransportDeleteMemoriesByQueryAction`.
- There is **no** retention-aware filtering in any memory read API.
- There is **no** scheduled cleanup for expired working memory.
- The only way to remove working memory is explicit manual deletion.

### Per-container configuration

Add a new field to memory container configuration:

| Field | Type | Default | Description |
|---|---|---|---|
[...]

PR #4860, opened 2026-06-16
Add RetentionRule data model and retention_policy field to MemoryConf... (PR1/PR5)

AirFunnyB:
Title: Add RetentionRule data model for memory retention lifecycle

### Description
Introduces the `RetentionRule` data model class and integrates
  `retention_policy`
  into `MemoryConfiguration` for the memory retention lifecycle feature.

  - Adds `RetentionRule` with fields: `session_ttl`, `history_ttl`,
  `working_memory_ttl`, `max_sessions`
  - Adds `retention_policy` field to `MemoryConfiguration` with full
  serialization/deserialization
  - Updates `ml_memory_container` index mapping to include retention
  policy fields
  - Unit tests for RetentionRule and updated MemoryConfiguration tests

### Related Issues
Related to RFC #4727

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.common.memorycontainer;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_AGENTIC_MEMORY_SYSTEM_INDEX_PREFIX;
 import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DEFAULT_MEMORY_INDEX_PREFIX;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DIMENSION_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DISABLE_HISTORY_FIELD;
@@ -64,7 +65,6 @@ import lombok.extern.log4j.Log4j2;
  */
 @Getter
 @Setter
-@Builder
 @EqualsAndHashCode
 @Log4j2
 public class MemoryConfiguration implements ToXContentObject, Writeable {
@@ -73,22 +73,18 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
     private String embeddingModelId;
     private String llmId;
     private Integer dimension;
-    @Builder.Default
-    private Integer maxInferSize = MAX_INFER_SIZE_DEFAULT_VALUE;
+    private Integer maxInferSize;
     private List<MemoryStrategy> strategies;
-    @Builder.Default
-    private Map<String, Map<String, Object>> indexSettings = new HashMap<>();
-    @Builder.Default
-    private Map<String, Object> parameters = new HashMap<>();
-    @Builder.Default
-    private boolean disableHistory = false;
-    @Builder.Default
-    private boolean disableSession = false;
-    @Builder.Default
-    private boolean useSystemIndex = true;
+    private Map<String, Map<String, Object>> indexSettings;
+    private Map<String, Object> parameters;
+    private boolean disableHistory;
+    private boolean disableSession;
+    private boolean useSystemIndex;
     private String tenantId;
     private Map<MemoryType, RetentionRule> retentionPolicy;
+    private transient boolean retentionPolicyExplicitlyNull = false;
 
+    @Builder
     public MemoryConfiguration(
         String indexPrefix,
         FunctionName embeddingModelType,
@@ -99,17 +95,23 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         List<MemoryStrategy> strategies,
         Map<String, Map<String, Object>> indexSettings,
         Map<String, Object> parameters,
-        boolean disableHistory,
-        boolean disableSession,
-        boolean useSystemIndex,
+        Boolean disableHistory,
+        Boolean disableSession,
+        Boolean useSystemIndex,
         String tenantId,
         Map<MemoryType, RetentionRule> retentionPolicy
     ) {
+        // Apply defaults for Boolean parameters
+        boolean effectiveUseSystemIndex = (useSystemIndex != null) ? useSystemIndex : true;
+        boolean effectiveDisableHistory = (disableHistory != null) ? disableHistory : false;
+        boolean effectiveDisableSession = (disableSession != null) ? disableSession : false;
+
         // Validate first
         validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
+        validateRetentionPolicy(retentionPolicy);
 
         // Assign values after validation
-        this.indexPrefix = buildIndexPrefix(indexPrefix, useSystemIndex);
+        this.indexPrefix = buildIndexPrefix(indexPrefix, effectiveUseSystemIndex);
         this.embeddingModelType = embeddingModelType;
         this.embeddingModelId = embeddingModelId;
         this.llmId = llmId;
@@ -127,9 +129,9 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         if (parameters != null && !parameters.isEmpty()) {
             this.parameters.putAll(parameters);
         }
-        this.disableHistory = disableHistory;
-        this.disableSession = disableSession;
-        this.useSystemIndex = useSystemIndex;
+        this.disableHistory = effectiveDisableHistory;
+        this.disableSession = effectiveDisableSession;
+        this.useSystemIndex = effectiveUseSystemIndex;
         this.tenantId = tenantId;
         this.retentionPolicy = retentionPolicy;
     }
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -168,6 +173,15 @@ public MemoryConfiguration(StreamInput input) throws IOException {
         this.disableSession = input.readBoolean();
         this.useSystemIndex = input.readBoolean();
         this.tenantId = input.readOptionalString();
+        if (input.readBoolean()) {
```
nathaliellenaa: ### [blocker] Unguarded stream read — MemoryConfiguration.java (StreamInput constructor)

`MemoryConfiguration` gets sent between nodes over the wire (it's deserialized via new `MemoryConfiguration(in)` in `MLCreateMemoryContainerInput`, `MLUpdateMemoryContainerInput`, and `MLMemoryContainer)`. During an upgrade, a 3.7 node and a 3.6 node talk to each other.

The new code always reads/writes the extra retentionPolicy bytes, but an older node doesn't know they exist. So:
- **New node reads from old node:** the extra bytes aren't there → it reads into the next field's data → garbage or EOFException.
- **New node writes to old node:** the old node never reads those bytes → everything after gets misaligned → corruption.

**Suggested fix:**
Only read `retentionPolicy` when both nodes are new enough, using a version check (example: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/McpConnector.java#L217)

```
Version streamInputVersion = input.getVersion();
if (streamInputVersion.onOrAfter(VERSION_3_x_0) && input.readBoolean()) { ... }
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -200,6 +214,16 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeBoolean(disableSession);
         out.writeBoolean(useSystemIndex);
         out.writeOptionalString(tenantId);
+        if (retentionPolicy != null && !retentionPolicy.isEmpty()) {
```
nathaliellenaa: ### [blocker] Unguarded stream write — MemoryConfiguration.java (writeTo)

Same issue as above, need to add version check (example: https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/connector/McpConnector.java#L275)

common/src/main/java/org/opensearch/ml/common/memorycontainer/RetentionRule.java
```diff
@@ -0,0 +1,101 @@
...
+    private final Integer maxCount;
+
+    public RetentionRule(Integer retentionDays, Integer maxCount) {
+        if (retentionDays != null && retentionDays <= 0) {
+            throw new IllegalArgumentException("retention_days must be a positive integer or null");
+        }
+        if (maxCount != null && maxCount <= 0) {
+            throw new IllegalArgumentException("max_count must be a positive integer or null");
+        }
+        this.retentionDays = retentionDays;
```
nathaliellenaa, line 41: `new RetentionRule(null, null)` is accepted and serializes to {}. Is this intentional?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -535,6 +608,27 @@ public void update(MemoryConfiguration updateContent) {
...
+            }
+            for (Map.Entry<MemoryType, RetentionRule> entry : updateContent.getRetentionPolicy().entrySet()) {
+                MemoryType type = entry.getKey();
+                RetentionRule incomingRule = entry.getValue();
+                RetentionRule existingRule = this.retentionPolicy.get(type);
+
+                if (existingRule == null) {
+                    this.retentionPolicy.put(type, incomingRule);
+                } else {
+                    Integer mergedDays = incomingRule.getRetentionDays() != null
```
nathaliellenaa: Just to confirm — is the "no retention_days for history" rule applied correctly on the update path? I see it enforced in `parseRetentionPolicy`, but not in `update()`. Through the current API this looks fine since the config goes through parse() first, so I just want to make sure update can't end up with `retention_days` set on history.

nathaliellenaa (commented):
Great start @AirFunnyB! Left some comments

common/src/main/java/org/opensearch/ml/common/memorycontainer/RetentionRule.java
AirFunnyB: Yes. Both nulls means no active constraints

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Fixed — wrapped both the read and write with input.getVersion().onOrAfter(VERSION_3_7_0) / out.getVersion().onOrAfter(VERSION_3_7_0). Older nodes will never see or expect the extra bytes. Fixed in the next push.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Good catch — you're right that it wasn't enforced on the update path. I've added validateRetentionPolicy() in update() before the merge, so it now rejects retention_days on history regardless of entry point. Fixed in the next push.

common/src/main/resources/index-mappings/ml_memory_container.json
nathaliellenaa: We should bump this version to 2 so the new mapping will be applied on upgraded clusters

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -168,6 +175,17 @@ public MemoryConfiguration(StreamInput input) throws IOException {
         this.disableSession = input.readBoolean();
         this.useSystemIndex = input.readBoolean();
         this.tenantId = input.readOptionalString();
+        if (input.getVersion().onOrAfter(VERSION_3_7_0)) {
```
nathaliellenaa: Since 3.7.0 is already released, we should gate it against the unreleased version 3.8.0

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -71,21 +73,18 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
...
     private String llmId;
     private Integer dimension;
-    @Builder.Default
-    private Integer maxInferSize = MAX_INFER_SIZE_DEFAULT_VALUE;
+    private Integer maxInferSize;
     private List<MemoryStrategy> strategies;
-    @Builder.Default
-    private Map<String, Map<String, Object>> indexSettings = new HashMap<>();
-    @Builder.Default
-    private Map<String, Object> parameters = new HashMap<>();
```
nathaliellenaa, line 80: Was dropping the `@Builder.Default` initializers intentional? 

Removing them means `indexSettings`/`parameters` come back as null (not empty) when a config is rebuilt from a stream with no maps set, and since `writeTo/toXContent` call `.isEmpty()` without a null check, the next serialization will NPE on the normal transport path. If null is meaningful for your update logic we should null-guard those checks; otherwise restore the initializers

common/src/main/resources/index-mappings/ml_memory_container.json
AirFunnyB: Fixed.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Changed to VERSION_3_8_0.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Yeah, not intentional. It broke when I moved @Builder to the constructor. Turns out @Builder.Default doesn't work with a constructor-level @Builder so it got dropped, and I forgot to put the init back. I just added the field initializers (new ArrayList<>() / new HashMap<>()) so they won't be null no matter how the object gets built. Fixed now.

nathaliellenaa: Sounds good, thanks for explaining

nathaliellenaa (approved):
Thanks for addressing the comments @AirFunnyB! LGTM

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -71,21 +73,18 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
...
-    @Builder.Default
-    private boolean disableSession = false;
-    @Builder.Default
-    private boolean useSystemIndex = true;
+    private boolean disableHistory;
+    private boolean disableSession;
+    private boolean useSystemIndex;
     private String tenantId;
+    private Map<MemoryType, RetentionRule> retentionPolicy;
+    private transient boolean retentionPolicyExplicitlyNull = false;
```
mingshl, line 85: what does it mean by explicitly Null?

mingshl (approved):

mingshl (approved):

mingshl:
you need at least one linux and one window CI passed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: It distinguishes "user sent retention_policy: null" (remove the policy) from "user didn't include the field" (leave it unchanged). Both parse to null in Java, so this flag tracks which case it is.

PR #4862, opened 2026-06-18
Add pinned field to memory models for retention policy eviction exemption (PR2/PR5)

AirFunnyB:
### Description
Adds a boolean `pinned` field to `MLMemorySession`,`MLLongTermMemory`, and `MLMemoryHistory` models. Pinned memories are exempt from retention policy eviction, allowing users to protect important memories from automatic cleanup.
- Adds pinned field (default null, treated as unpinned by eviction logic) to all three memory model classes
  - Full serialization/deserialization support (XContent and StreamInput/StreamOutput)
  - Updates index mappings for `ml_memory_sessions`, `ml_memory_long_term`, and `ml_memory_long_term_history`
  - Unit tests for pinned field behavior across all models
  
### Related Issues
Part of the memory retention lifecycle feature (RFC #4727). 
Depends on #4860 (RetentionRule data model).

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
@@ -6,6 +6,7 @@
 package org.opensearch.ml.common.memorycontainer;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.CREATED_TIME_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.LAST_UPDATED_TIME_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEMORY_CONTAINER_ID_FIELD;
@@ -102,7 +103,9 @@ public class MLLongTermMemory implements ToXContentObject, Writeable {
         this.ownerId = in.readOptionalString();
         this.memoryContainerId = in.readOptionalString();
         this.strategyId = in.readOptionalString();
-        this.pinned = in.readOptionalBoolean();
+        if (in.getVersion().onOrAfter(VERSION_3_7_0)) {
+            this.pinned = in.readOptionalBoolean();
+        }
         // Note: memoryEmbedding is not serialized in StreamInput/Output as it's typically handled separately
     }
 
@@ -127,7 +130,9 @@ public class MLLongTermMemory implements ToXContentObject, Writeable {
         out.writeOptionalString(ownerId);
         out.writeOptionalString(memoryContainerId);
         out.writeOptionalString(strategyId);
-        out.writeOptionalBoolean(pinned);
+        if (out.getVersion().onOrAfter(VERSION_3_7_0)) {
+            out.writeOptionalBoolean(pinned);
+        }
         // Note: memoryEmbedding is not serialized in StreamInput/Output as it's typically handled separately
     }
 
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
@@ -6,7 +6,7 @@
 package org.opensearch.ml.common.memorycontainer;
 
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
-import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_8_0;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.CREATED_TIME_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.LAST_UPDATED_TIME_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.MEMORY_CONTAINER_ID_FIELD;
@@ -103,7 +103,7 @@ public class MLLongTermMemory implements ToXContentObject, Writeable {
         this.ownerId = in.readOptionalString();
         this.memoryContainerId = in.readOptionalString();
         this.strategyId = in.readOptionalString();
-        if (in.getVersion().onOrAfter(VERSION_3_7_0)) {
+        if (in.getVersion().onOrAfter(VERSION_3_8_0)) {
             this.pinned = in.readOptionalBoolean();
         }
         // Note: memoryEmbedding is not serialized in StreamInput/Output as it's typically handled separately
@@ -130,7 +130,7 @@ public class MLLongTermMemory implements ToXContentObject, Writeable {
         out.writeOptionalString(ownerId);
         out.writeOptionalString(memoryContainerId);
         out.writeOptionalString(strategyId);
-        if (out.getVersion().onOrAfter(VERSION_3_7_0)) {
+        if (out.getVersion().onOrAfter(VERSION_3_8_0)) {
             out.writeOptionalBoolean(pinned);
         }
         // Note: memoryEmbedding is not serialized in StreamInput/Output as it's typically handled separately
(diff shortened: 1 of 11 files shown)


AirFunnyB:
Addressing Code-Diff-Analyzer Findings

  1. MLMemoryHistory.java line 139 (medium) — createdTime serialization reorder:

  This reorder is intentional — it fixes a bug where writeTo() wrote namespace → 
  tags → createdTime but the StreamInput constructor read createdTime → namespace
  → tags. Since MLMemoryHistory is new code introduced in this feature branch
  (never released), there are no existing nodes with the old ordering. The fix
  aligns writeTo() with the constructor so stream round-trips work correctly. A
  version guard is not needed because no prior version has ever serialized this
  class.

  2. MemoryConfiguration.java line 91 (low) — retentionPolicyExplicitlyNull 
  transient field:

  This was reviewed and accepted in PR1 (#4860). The transient field is necessary
  to support partial update semantics where absent ≠ null.

AirFunnyB:
CI is expected to fail until PR1 (#4860) is merged into the feature branch — this PR depends on the RetentionRule data model and constants introduced there.

common/src/main/resources/index-mappings/ml_memory_long_term.json
nathaliellenaa: We should bump this version to 2

common/src/main/resources/index-mappings/ml_memory_long_term_history.json
nathaliellenaa: Same here, bump this version to 2

common/src/main/resources/index-mappings/ml_memory_sessions.json
nathaliellenaa: Bump version to 2

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
```diff
@@ -58,6 +60,7 @@ public class MLLongTermMemory implements ToXContentObject, Writeable {
     private String ownerId;
     private String memoryContainerId;
     private String strategyId;
+    private Boolean pinned;
```
nathaliellenaa, line 63: The PR description says` pinned` defaults to `false`, but it's actually nullable and stays null when not set — so unpinned memories don't store a pinned field at all. That's fine as long as the eviction logic counts "missing" as not-pinned; just flag it so a `pinned = false` filter doesn't accidentally skip them. (Same for MLMemoryHistory and MLMemorySession.)

nathaliellenaa (commented):
@AirFunnyB Overall looks good, added some comment. Don't forget to apply spotless `./gradlew spotlessApply`

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLLongTermMemory.java
AirFunnyB: Good catch. The field's nullable on purpose — null just means it was never explicitly pinned. The eviction query uses must_not: {term: {pinned: true}}, so it matches docs where pinned is null, absent, or false, which means missing-field docs never get skipped by accident. I've updated the PR description to say "default null (treated as unpinned)" instead of "default false".

nathaliellenaa (approved):
Thanks @AirFunnyB , LGTM!

AirFunnyB:
> Thanks @AirFunnyB , LGTM!

Awesome will reach out to @mingshl to have her approve merging

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -148,6 +148,14 @@ public class MemoryContainerConstants {
         "Backend role contains invalid characters. Only alphanumeric characters and :+=,.@-_/ are allowed: %s";
     public static final String BACKEND_ROLE_EMPTY_ERROR = "Backend role cannot be empty or blank";
 
+    // Pinned field
+    public static final String PINNED_FIELD = "pinned";
+
+    // Retention policy field names
+    public static final String RETENTION_POLICY_FIELD = "retention_policy";
```
mingshl, line 155: are these in used? I saw the declaration but don't see them in the later codes

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
```diff
@@ -57,6 +59,7 @@ public class MLMemoryHistory implements ToXContentObject, Writeable {
     private Instant createdTime;
     private String tenantId;
     private String error;
+    private Boolean pinned;
```
mingshl, line 62: is pinned field only available for MemoryHistory? How about working memory?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
AirFunnyB: Yes, both are used in later PRs

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
```diff
@@ -131,6 +139,7 @@ public void writeTo(StreamOutput out) throws IOException {
         } else {
             out.writeBoolean(false);
         }
+        out.writeOptionalInstant(createdTime);
```
mingshl: would you restore this order?  this change is unintentional right?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
mingshl: remove this and introduce in later PR to be more readable pls

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
AirFunnyB: It does not make sense to have working memories with out the context of the session. You can pin the session if you want to keep that context.

mingshl:
For design wise, can you explain why pinned field are  available for MLMemorySession,MLLongTermMemory, and MLMemoryHistory models. Can you explain what other types are not available, for example, workingMemory, and the reasons behind?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
AirFunnyB: Restored — createdTime is now written after namespace/tags in commit 0d3658d. Thank you!

mingshl:
there is dependency issue, try to run local ./gradlew build to see if you can reproduce the issue

```


* What went wrong:
Could not determine the dependencies of task ':opensearch-ml-plugin:dependencyLicenses'.
> Failed to query the value of task ':opensearch-ml-plugin:dependencyLicenses' property 'dependencies'.

   > Could not resolve all dependencies for configuration ':opensearch-ml-plugin:runtimeClasspath'.
Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.
      > Could not resolve com.networknt:json-schema-validator:3.0.2.

        Required by:
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
            project ':opensearch-ml-plugin'

         > Could not resolve com.networknt:json-schema-validator:3.0.2.
For more on this, please refer to https://docs.gradle.org/9.4.1/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
            > Could not get resource 'https://ci.opensearch.org/ci/dbc/snapshots/maven/com/networknt/json-schema-validator/3.0.2/json-schema-validator-3.0.2.pom'.
               > Could not GET 'https://ci.opensearch.org/ci/dbc/snapshots/maven/com/networknt/json-schema-validator/3.0.2/json-schema-validator-3.0.2.pom'. Received status code 503 from server: Service Unavailable
> There are 85 more failures with identical causes.

```

AirFunnyB:
Session, Long-Term Memory, and History are evaluated directly by the retention job, so they need pinned to let customers exempt specific documents from eviction.

Working Memory is never evaluated directly — it's only deleted via session cascade. To preserve a conversation, you pin the session instead, which protects all its messages. Pinning individual messages would risk conversations with gaps, or information with out full context which breaks agent reasoning.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MLMemoryHistory.java
mingshl: can you explain how to pin the session if working memory doesn't have pin field?

AirFunnyB: To pin a session (and preserve/pin its working memory), you use the update-memory API on the session document:

PUT /_plugins/_ml/memory_containers/{container_id}/memories/sessions/{session_id}
{ "pinned": true }

The retention job skips pinned sessions during eviction, and since working memory is only deleted via session cascade, all messages in that session are protected. You can see the session pinning in MLMemorySession.java in this PR.

mingshl (approved):

PR #4878, opened 2026-06-26
Add LLM judgment connector blueprints

RexyZhang2001:
### Description
Adds connector blueprints under `docs/remote_inference_blueprints/llm_judgment/` for the Search Relevance Workbench (SRW) LLM-as-a-judge flow, which generates relevance judgments by calling a chat model registered in ml-commons.

SRW is provider-agnostic on the wire: it sends provider-neutral `system_prompt` and `user_prompt` parameters and reads the model text back from a neutral top-level `response` field. Each blueprint absorbs one provider's request/response differences by mapping those neutral parameters into the provider's `request_body` and copying the model text into the `response` field via a `post_process_function`. This is the same connector-blueprint mechanism neural search uses to support many embedding providers without provider-specific plugin code.

Blueprints included:
- `openai_chat_blueprint.md` — OpenAI Chat Completions
- `azure_openai_chat_blueprint.md` — Azure OpenAI Chat
- `deepseek_chat_blueprint.md` — DeepSeek (OpenAI-compatible)
- `ollama_chat_blueprint.md` — Ollama and other local OpenAI-compatible servers (vLLM, llama.cpp)
- `google_gemini_chat_blueprint.md` — Google Gemini `generateContent` (API key passed in the `x-goog-api-key` header, since `${credential.*}` is substituted into headers but not the URL)
- `anthropic_claude_bedrock_blueprint.md` — Anthropic Claude on Amazon Bedrock (native `invoke`)
- `bedrock_converse_blueprint.md` — generic Amazon Bedrock Converse API (one format for any chat model on Bedrock)
- `README.md` — index and the neutral request/response contract

Each blueprint follows the existing `remote_inference_blueprints` format (numbered create-connector / register-and-deploy / test-inference steps with sample responses). Every blueprint's `request_body` and `post_process_function` was validated end-to-end against a local OpenSearch cluster — HTTP providers via a mock endpoint, and Claude (native invoke) and Bedrock Converse against real Amazon Bedrock.

### Related Issues
https://github.com/opensearch-project/search-relevance/issues/509

### Related PR
https://github.com/opensearch-project/search-relevance/pull/515

### Check List
- [x] New functionality has been documented.
- [x] Commits are signed per the DCO using `--signoff`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

Issue #50, opened 2021-06-25
Search model by model id (task id) instead of getting by model id.


spbjss:
When the predict tried to get the model from index by model id (returned to client by train), it will fail, because the model has auto generated  _id, the "get" will try to match the model id with the _id which will never succeed.

Instead of getting the model by _id, we could search the model by the model id (taskId). For example:
curl -XGET 'http://localhost:9200/.os_ml_model_result/_search?q=taskId:052cde39-3ec2-4b51-ad76-fd92db60fb0e'

PR #4882, opened 2026-06-29
fix: Retry transient HTTP errors from any remote service

RexyZhang2001:
#### Problem

Currently, connector retries only trigger on `RemoteConnectorThrottlingException`, which is produced solely from AWS's `x-amzn-ErrorType` response header.  However, other providers (OpenAI, Gemini, Cohere, Anthropic, DeepSeek, etc.) signal transient failures with plain HTTP status codes, not that AWS-specific exception, so their throttling and server errors are never retried.

| Provider | Transient failure signal | Retried before? | Retried now? |
|---|---|---|---|
| AWS (Bedrock/SageMaker) | `RemoteConnectorThrottlingException` (header) | ✅ | ✅ |
| OpenAI, Gemini, Cohere, Anthropic, DeepSeek, etc | HTTP `429` | ❌ | ✅ |
| Any remote service | HTTP `5xx` | ❌ | ✅ |

#### Change

The retry decision is extracted into a new `isRetryable(Throwable)` helper method. Instead of only recognizing the AWS-specific `RemoteConnectorThrottlingException`, it now also inspects the HTTP status code on `OpenSearchStatusException` — the common exception type produced by the shared `MLSdkAsyncHttpResponseHandler` that all connectors (AWS and generic HTTP) use.

| Condition | Retried? | Example scenario |
|---|---|---|
| `RemoteConnectorThrottlingException` | ✅ | AWS SageMaker/Bedrock header-based throttling |
| HTTP `429` (Too Many Requests) | ✅ | OpenAI / Gemini / Cohere rate-limiting |
| HTTP `5xx` (Server Error) | ✅ | Transient outage, gateway timeout, connection timeout |

Non-transient failures (`4xx` client errors, unknown status codes) are not retried — the request itself is wrong or unsupported, and repeating it would not help.

Local per-model and per-user rate limiting remains unaffected: those checks throw before the retry wrapper is created, so they never enter the retry path.

#### Testing

Unit tests cover each branch of `isRetryable`: throttling, `429`, `5xx`, non-retryable `4xx`, and `null` status.

### Related Issues
https://github.com/opensearch-project/search-relevance/issues/510

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
@@ -419,7 +419,7 @@ public interface RemoteConnectorExecutor extends AutoCloseable {
         public boolean shouldRetry(Exception e) {
             Throwable cause = ExceptionsHelper.unwrapCause(e);
             Integer maxRetryTimes = args.connectionExecutor.getConnectorClientConfig().getMaxRetryTimes();
-            boolean shouldRetry = cause instanceof RemoteConnectorThrottlingException;
+            boolean shouldRetry = isRetryable(cause);
             if (++retryTimes > maxRetryTimes && maxRetryTimes != -1) {
                 shouldRetry = false;
             }
@@ -429,6 +429,22 @@ public interface RemoteConnectorExecutor extends AutoCloseable {
                     .debug(String.format(Locale.ROOT, "The %d-th retry for invoke remote model", retryTimes), e);
             }
             return shouldRetry;
+        }
+
+        private static boolean isRetryable(Throwable cause) {
+            // AWS header-based throttling.
+            if (cause instanceof RemoteConnectorThrottlingException) {
+                return true;
+            }
+            // Transient HTTP failures from any remote service: 429 and 5xx.
+            if (cause instanceof OpenSearchStatusException) {
+                RestStatus status = ((OpenSearchStatusException) cause).status();
+                if (status == null) {
+                    return false;
+                }
+                return status == RestStatus.TOO_MANY_REQUESTS || status.getStatus() >= 500;
+            }
+            return false;
         }
     }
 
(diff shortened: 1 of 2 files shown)


nathaliellenaa (approved):
LGTM!

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/RemoteConnectorExecutor.java
```diff
@@ -430,6 +430,22 @@ public boolean shouldRetry(Exception e) {
             }
             return shouldRetry;
         }
+
+        private static boolean isRetryable(Throwable cause) {
+            // AWS header-based throttling.
+            if (cause instanceof RemoteConnectorThrottlingException) {
```
mingshl, line 436: for this example, when bedrock throttles , and we keep retrying in all failing requests, the throttling might be tight. if isRetryable, how many retries it would take?

mingshl: If a user has 100 documents going through an ingest pipeline with ml_inference, and Bedrock starts throttling at document #50, the remaining 50 docs will all retry, potentially 3x each = 150 additional requests hitting an already throttled endpoint. the throttled condition might last even longer in this case

mingshl: do we have Exponential backoff with jitter applied when retry?

RexyZhang2001: Thanks for the comments. For the Bedrock example, this PR doesn't change anything. Bedrock throttling was already retried before this PR, via the AWS-specific `RemoteConnectorThrottlingException` (detected from the `x-amzn-ErrorType` response header).

What this PR adds is retry for other providers — OpenAI, Gemini, Cohere, Anthropic, DeepSeek — that previously had no retry at all for 429/5xx errors. So the "how many retries" question applies to them, not Bedrock. For those: retry count is capped by `max_retry_times` (default 0, meaning off unless a user turns it on in blueprint), and the recommended backoff (exponential_full_jitter) spaces retries out with random delays so they don't all hit the service at once, which is recommended in the blueprints used for LLM judgment in search-relevance:
https://github.com/opensearch-project/ml-commons/pull/4878

rithinpullela (approved):

PR #4894, opened 2026-07-03
Add Gemini embedding connector blueprint

cjrolo:
## Summary

- Adds a pre-configured blueprint for Gemini text embeddings using the `embedContent` API.
- Lists the Gemini chat and embedding blueprints in the standard blueprints index.
- Documents that Gemini API keys are passed through the `x-goog-api-key` header.

The chat connector authentication fix originally included in this PR was merged separately in #4874. This branch has been rebased onto `main`, so that implementation is no longer duplicated here.

## Sources

- Gemini embeddings API: https://ai.google.dev/api/embeddings
- Gemini embeddings guide: https://ai.google.dev/gemini-api/docs/embeddings
- Gemini API authentication: https://ai.google.dev/gemini-api/docs/api-key

## Test plan

- [x] Rebased onto current `main` and verified the diff contains no duplicate auth implementation from #4874.
- [x] `git diff --check`
- [x] `./gradlew spotlessCheck`
- [ ] Manually create the embedding connector with a real Gemini API key and confirm `_predict` returns a float vector through `response_filter`.

Docs-only change; no production code is modified.

### DIFF
--- a/docs/remote_inference_blueprints/google_gemini_connector_chat_blueprint.md
+++ b/docs/remote_inference_blueprints/google_gemini_connector_chat_blueprint.md
@@ -21,6 +21,8 @@ PUT /_cluster/settings
 
 The Gemini API uses `contents` (not `messages`) with `role` and `parts`. `generationConfig` is optional and can be passed per-request if needed.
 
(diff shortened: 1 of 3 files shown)


cjrolo:
Update to match recent Gemini updates. Removed from Draft

PR #4866, opened 2026-06-22
Add retention policy and pinned field to memory   container REST APIs (PR3/PR5)

AirFunnyB:
## Description
  Wires `retention_policy` into Create/Update
  MemoryContainer endpoints and `pinned` field into
  Create/Update Memory (AddMemories/UpdateMemory)
  endpoints with input validation.

  - Create MemoryContainer: accepts and persists
  `retention_policy`
  - Update MemoryContainer: supports partial update of
  `retention_policy`
  - AddMemories: accepts `pinned` field on new memories
  - UpdateMemory: allows toggling `pinned` on existing
  memories
  - Input validation and unit tests for all endpoints

  ## Related Issues
  Part of the memory retention lifecycle feature (RFC
  #4859).
  Depends on #4860 (RetentionRule data model) and #4862
  (pinned field).

  ## Check List
  - [x] New functionality includes testing.
  - [ ] New functionality has been documented.
  - [ ] API changes companion pull request created.
  - [x] Commits are signed per the DCO using --signoff.
  - [ ] Public documentation issue/PR created.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4896, opened 2026-07-06
test: Retry integration tests on transient remote-service 5xx errors

rithinpullela:
### Description
Integration tests that call real remote inference endpoints (Bedrock, OpenAI, Cohere) intermittently fail CI when the *external service* has a transient blip — not because of any bug in ml-commons.

A triage of 93 failed CI jobs over 21 days found two recurring availability signatures:
- **3×** HTTP 503 wrapped in `CommonValue.REMOTE_SERVICE_ERROR` ("Error from remote service: ... Service Unavailable")
- **6×** HTTP 500 from `MLSdkAsyncHttpResponseHandler` ("Error communicating with remote model: Read timed out" / connection reset)

Each occurrence fails the whole CI run and typically triggers a full re-run, wasting ~35 minutes per retry.

This change adds a **retry** at `TestHelper.makeRequest` — the shared choke point for these IT calls. When a request fails with one of the two narrowly-matched remote-unavailability signatures, it is retried (up to 3 times, linear backoff of 3s/6s/9s) to ride out a momentary outage. Because these blips are transient, a retry usually lets the test run for real.

**Retry-then-fail — genuine regressions are never masked:**
- Only the two matched signatures are retried; everything else (4xx, unwrapped 5xx, assertion failures) fails immediately on the first attempt.
- If the remote stays unavailable across *every* attempt, the original `ResponseException` is re-thrown and the test **fails normally** — we do not skip. This means a bad change submitted while a remote is down still surfaces as a failure.
- 503s must carry the remote-service wrapper or a Service Unavailable phrasing; 500s must be the remote-connectivity wrapper with a timeout/connection-reset cause.

Example CI failure this addresses: https://github.com/opensearch-project/ml-commons/actions/runs/28430472024/job/85193234881

### Related Issues
Related to the CI flakiness addressed in #4851 (different failure class: that PR fixes a test-specific race; this one retries genuine transient remote-service unavailability across the suite).

### Check List
- [x] New functionality includes testing. (Change is itself test infrastructure; verified compile + spotless + existing ITs unaffected)

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/utils/TestHelper.java
+++ b/plugin/src/test/java/org/opensearch/ml/utils/TestHelper.java
@@ -41,21 +41,25 @@ import java.util.TreeSet;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.hc.core5.http.io.entity.StringEntity;
 import org.apache.logging.log4j.util.Strings;
 import org.apache.lucene.search.TotalHits;
+import org.junit.AssumptionViolatedException;
 import org.opensearch.Version;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.search.ShardSearchFailure;
 import org.opensearch.client.Request;
 import org.opensearch.client.RequestOptions;
 import org.opensearch.client.Response;
+import org.opensearch.client.ResponseException;
 import org.opensearch.client.RestClient;
 import org.opensearch.client.WarningsHandler;
 import org.opensearch.cluster.ClusterName;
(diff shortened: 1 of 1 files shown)


AirFunnyB:
+1 — been hitting the exact same CI flakiness on #4866 this is a great fix!

One thing I noticed: the NPE from MLAgentExecutor (inputDataSet is null) comes back as a 500 but doesn't have the "Error communicating with remote model" wrapper, so it won't match the patterns here. Might be worth a broader 500 catch for that case?

plugin/src/test/java/org/opensearch/ml/utils/TestHelper.java
```diff
@@ -201,9 +205,53 @@ public static Response makeRequest(
...
+        try {
+            body = EntityUtils.toString(e.getResponse().getEntity());
+        } catch (Exception ignored) {
+            return;
+        }
+        if (body == null) {
+            return;
+        }
+        if (status == 503 && REMOTE_SERVICE_ERROR_PATTERN.matcher(body).find()) {
+            throw new AssumptionViolatedException("Remote service unavailable (503), skipping test: " + body);
```
jiapingzeng: Could this result in false positives? Let's say some bad code change is submitted when the remote service is down, would CIs pass as a result?

rithinpullela:
> +1 — been hitting the exact same CI flakiness on #4866 this is a great fix!
> 
> One thing I noticed: the NPE from MLAgentExecutor (inputDataSet is null) comes back as a 500 but doesn't have the "Error communicating with remote model" wrapper, so it won't match the patterns here. Might be worth a broader 500 catch for that case?

The errors such as NPE from MLAgentExecutor are errors originating from our code, so these should be caught on CI. The goal of this PR is to skip tests which would fail because of a remote service(Bedrock or OpenAI) which we have no control over

plugin/src/test/java/org/opensearch/ml/utils/TestHelper.java
rithinpullela: Yes, the CIs would pass in that case. But since we hit 5xx transiently, I believe we will have a high probability of catching the bug in one of the 4 CIs

jiapingzeng (approved):

Issue #4889, opened 2026-06-30
[RFC] Server-Side Batch Inference for ML Commons


bzhangam:
## Summary

Introduce batch inference in the **ML Commons server** to maximize model throughput for high-throughput workloads. Remote model endpoints (e.g., SageMaker, Bedrock, Cohere) throttle based on the rate of requests they receive, so sustaining throughput requires packing more inference work into fewer invocations.

The proposal is delivered in two phases, both implemented on the ML Commons server so that **all callers (every processor, every plugin, direct API) benefit** and configuration is read directly from model metadata with no client-side caching:

- **Phase 1 — Size-based batching (ingest):** The server splits a single incoming multi-text predict request into sub-batches bounded by byte size and text count. This solves the "fixed `batch_size` is hard to tune and can overflow request size" problem for ingest. **No queue, no cross-request state.**
- **Phase 2 — Queue-based cross-request batching (search):** The server coalesces inference requests from *multiple concurrent callers* into shared batches. This is the only way to batch the **search path**, where dense-model inference happens per query and cannot be grouped through a bulk API. Phase 2 differentiates ingest vs. search traffic by using **separate model IDs per path** (simpler, no new server-side machinery) with a two-way door to a server-side `requestType` signal if needed later. Phase 2 is gated on benchmarks proving a throughput gain with added query latency bounded by `flush_timeout_ms` (default 10ms for search).

The two phases compose cleanly: Phase 2's queue accumulates requests and hands each flush to the same size-based splitter from Phase 1, which guarantees every dispatched batch respects the model's limits.

**Related:**
- [neural-search#1761](https://github.com/opensearch-project/neural-search/issues/1761) — Size-based batching proposal
- [neural-search#1762](https://github.com/opensearch-project/neural-search/pull/1762) — Size-based batching PR
- [neural-search#598](https://github.com/opensearch-project/neural-search/issues/598) — Original batch ingestion RFC
- [OpenSearch#12457](https://github.com/opensearch-project/OpenSearch/issues/12457) — OpenSearch core batching RFC

---

## Motivation

Remote model endpoints throttle based on how many requests they receive within a time window. To sustain throughput we must reduce the *number* of invocations by packing more inference work into each request. This applies to both the ingest and search paths, but the constraints — and therefore the right solution — differ between them.

### Problem 1: Ingest path — count-based batching is hard to tune and can overflow request size

[...]

PR #4898, opened 2026-07-10
Add server-side size-based batch inference splitting for ingest predict requests

RexyZhang2001:
### Description

Implements Phase 1 of the server-side batch inference(#4889): size-based batching for the ingest path.

When a remote model is configured with `batch_inference_config`, the ML Commons server splits a single multi-text predict request into sub-batches bounded by item count (`max_items_per_request`) and UTF-8 byte size (`max_bytes_per_request`), dispatches them concurrently against the model, and reassembles the results into one response identical in shape to an un-split call. This solves the "a fixed `batch_size` is hard to tune and can overflow the request size limit" problem without any client-side changes, a queue, or cross-request state.

**File organization**

```
+------------------------------------------------------------------------+
| Config layer -- what a model's split limits are                        |
|                                                                        |
|BatchInferenceConfig      the two size limits, persisted on the model   |
|MLModel                   holds the config on the model object          |
|MLRegisterModelInput      lets users set the config at register time    |
|MLModelManager            persists the config into the model index      |
+------------------------------------------------------------------------+
                          | read by
                          v
+------------------------------------------------------------------------+
| Execution layer -- splitting, dispatch, and reassembly                 |
| (org.opensearch.ml.batch)                                              |
|                                                                        |
|BatchItem                 one unit of work: payload + byte size         |
|BatchableInput            per-type strategy: toItems / merge / combine  |
|TextDocsBatchableInput    the text-docs implementation                  |
|BatchableInputRegistry    input type -> handler lookup                  |
|SizeBasedBatchSplitter    packs items into size-legal sub-batches       |
|SingleBatchInvoker        how one sub-batch is actually invoked         |
|BatchInferenceExecutor    orchestrates split -> dispatch -> merge;      |
|                          retries 429/503 with backoff, fails the whole |
|                          request on any other error                    |
+------------------------------------------------------------------------+
                          | wired into
                          v
+------------------------------------------------------------------------+
| Integration -- the real predict path                                   |
|                                                                        |
|MLPredictTaskRunner       reads the model's config from cache, routes   |
|                          through the executor when batching applies,   |
|                          otherwise takes the existing single-call path |
|                          unchanged                                     |
+------------------------------------------------------------------------+
```

**Request flow**

```
predict request arrives
   -> MLPredictTaskRunner reads the model's batch_inference_config from cache
   -> not configured, or already fits within limits  ->  single call, exactly as before
   -> configured AND exceeds limits:
        BatchableInputRegistry finds the handler for this input type (text docs)
        handler.toItems           -> break the request into items with byte sizes
        SizeBasedBatchSplitter    -> pack items into N size-legal sub-batches
        for each sub-batch: handler.merge -> invoke the model (429/503 retried with backoff)
        all N responses collected, in original order
        handler.combine           -> flatten into one response, shaped like an un-split call
   -> listener notified once, response returned to caller
```

**BWC**
`BatchInferenceConfig` holds the two size limits (`max_items_per_request`, default 512; `max_bytes_per_request`, default -1/disabled), is persisted on `MLModel`, and is settable at model registration. A model without `batch_inference_config` never enters this flow and behaves exactly as today.

### Related Issues

Relates to #4889 (RFC: Server-Side Batch Inference for ML Commons). 
This PR implements Phase 1 (size-based batching for ingest) only

### Benchmark Result
https://github.com/opensearch-project/ml-commons/issues/4889#issuecomment-5398072424

### Testing

**Unit tests** (118 tests, all passing): `BatchInferenceConfigTest`, `SizeBasedBatchSplitterTests`, `TextDocsBatchableInputTests`, `BatchInferenceExecutorTests`, plus existing `MLModelTests`, `MLRegisterModelInputTest`, and `MLPredictTaskRunnerTests` re-verified for regressions.

**Manual end-to-end testing**

Tested end-to-end on a live single-node OpenSearch cluster (`./gradlew run`, build `3.8.0-SNAPSHOT`) using a real remote model, **Amazon Bedrock `cohere.embed-english-v3`**.

Server-side batching splits a single predict request into several smaller model calls and then reassembles the results into one response. The feature is correct only if a caller cannot tell whether the request was split internally. To verify this, every scenario registers two models that share the same Bedrock connector:

- **Batched model** — has `batch_inference_config` set, so requests are split.
- **Control model** — has no config, so the request runs as a single call (the existing behavior).

For the same input, the batched model must produce a result that is indistinguishable from the control model on three properties:

1. **Count** — N input documents return exactly N embeddings (none dropped or duplicated).
2. **Shape** — the response is a single group of N embeddings, exactly like an un-split call.
3. **Order** — the embedding at position *i* is the embedding of input document *i*. This is checked by comparing the batched result to the control result one position at a time using cosine similarity, where `1.0` means the two vectors are identical and `~0` means they are unrelated. A value close to `1.0` at every position means each embedding stayed matched to its original document even though it was produced in a different sub-batch.

Each request was sent as:

```json
POST /_plugins/_ml/_predict/text_embedding/<model_id>
{
  "text_docs": [ ... ],
  "return_number": true,
}
```

The `batch_inference_config` field was also confirmed to persist through the full stack: it is accepted at registration, stored in the `.plugins-ml-model` index, cached on the node, and returned by `GET /_plugins/_ml/models/<id>`.

**Summary of scenarios**

| # | Scenario | Config | Input | Server split | Response | Outcome |
|---|----------|--------|-------|--------------|----------|---------|
| 1 | Count ceiling | `max_items_per_request = 3` | 10 documents | 10 items → 4 sub-batches | 1 group, 10 embeddings | Order preserved across 4 calls |
| 2 | Byte ceiling (adaptive) | `max_bytes_per_request = 40` | 7 mixed-size documents | 7 items → 3 sub-batches | 1 group, 7 embeddings | Batches adapt to text size, no overflow |
| 3 | Oversize document | `max_bytes_per_request = 40` | 3 documents (1 over the limit) | 3 items → 3 sub-batches | 1 group, 3 embeddings | Oversize doc isolated, not dropped |
| 4 | Under limits (passthrough) | `max_items_per_request = 100`, `max_bytes_per_request = 100000` | 4 documents | no split | 1 group, 4 embeddings | Behaves exactly as before, no overhead |

**Scenario 1 — Count ceiling**

Config `max_items_per_request = 3`, input of 10 short documents.

```
Server log: [BatchInferenceExecutor] Size-based batching: split 10 items into 4 sub-batches
Response:   1 group, 10 embeddings (1024-d)

Cosine similarity vs. the control model, per position:
  item01 0.99988   item02 0.99995   item03 0.99991   item04 0.99996   item05 0.99971
  item06 0.99962   item07 0.99976   item08 0.99977   item09 0.99989   item10 0.99997
  positions below 0.99: 0
```

The 10 documents were split into 4 Bedrock calls (3 + 3 + 3 + 1), as shown by the log line. The response is still a single group of 10 embeddings, identical in shape to an un-split call. Comparing each batched embedding to the control embedding at the same position gives a cosine similarity of roughly 1.0 for all 10 positions, with zero positions below 0.99, confirming every embedding remained matched to its original document across the 4 separate calls.

**Scenario 2 — Byte ceiling (adaptive batching)**

Config `max_bytes_per_request = 40` (no binding item count), input of 7 documents of different lengths.

```
Server log: [BatchInferenceExecutor] Size-based batching: split 7 items into 3 sub-batches
Response:   1 group, 7 embeddings (1024-d)
```

With no binding item-count limit, batches are formed by total UTF-8 byte size. The 7 documents of varying length pack into 3 byte-bounded sub-batches: short documents are grouped together while longer ones form smaller batches. This adapts automatically to document size, which a fixed count-based `batch_size` cannot do. All 7 embeddings are returned in a single group, confirming no sub-batch exceeded the 40-byte limit and no input was lost.

**Scenario 3 — Oversize document isolated**

Config `max_bytes_per_request = 40`, input of 3 documents where one is much larger than 40 bytes.

```
Server log: [BatchInferenceExecutor] Size-based batching: split 3 items into 3 sub-batches
Response:   1 group, 3 embeddings (1024-d)
```

The oversized document cannot share a byte-bounded batch with any neighbor, so the splitter places it in a sub-batch of its own rather than dropping or truncating it. The request still succeeds and returns all 3 embeddings — the case where a fixed count-based batch could previously fail with a "request too large" error.

**Scenario 4 — Under limits (passthrough)**

Config `max_items_per_request = 100`, `max_bytes_per_request = 100000`, input of 4 small documents.

```
Server log: (no "Size-based batching" line)
Response:   1 group, 4 embeddings (1024-d)
```

The request already fits within both limits, so no split occurs. The absence of the split log line confirms a single model call, identical to a model with no `batch_inference_config` — no added overhead when batching isn't needed.

**Note on cosine similarity values**

Cosine similarity is approximately `0.9999` rather than exactly `1.0` because Bedrock/Cohere returns marginally different float values depending on how many texts share a single `invoke` call — a property of the endpoint, not the splitting logic. A real ordering bug would produce a value near `0` at the affected positions; every position being close to `1.0` confirms the order is correct.

### Check List


common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
```diff
@@ -0,0 +1,108 @@
...
+ * (count ceiling) and max_bytes_per_request (UTF-8 byte ceiling, -1 to disable).
+ */
+@Getter
+public class BatchInferenceConfig implements ToXContentObject, Writeable {
+
+    public static final String MAX_ITEMS_PER_REQUEST_FIELD = "max_items_per_request";
+    public static final String MAX_BYTES_PER_REQUEST_FIELD = "max_bytes_per_request";
+
+    // Default when the field is omitted; callers should set their model's documented batch limit.
+    public static final int DEFAULT_MAX_ITEMS_PER_REQUEST = 512;
```
bzhangam: why default 512 items and no limit on bytes size? should we also have a default bytes size e.g. 1KB?

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public BatchInferenceConfig(StreamInput in) throws IOException {
+        this.maxItemsPerRequest = in.readInt();
+        this.maxBytesPerRequest = in.readLong();
+    }
+
+    private void validate() {
+        if (maxItemsPerRequest < 1) {
+            throw new IllegalArgumentException(MAX_ITEMS_PER_REQUEST_FIELD + " must be a positive integer, but got " + maxItemsPerRequest);
```
bzhangam: We also should support -1 as no limit.

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+        }
+        ModelTensors combined = ModelTensors.builder().mlModelTensors(tensors).build();
+        return ModelTensorOutput.builder().mlModelOutputs(List.of(combined)).build();
+    }
+
+    private TextDocsInputDataSet asTextDocs(MLInput input) {
+        if (input == null || !(input.getInputDataset() instanceof TextDocsInputDataSet)) {
+            throw new IllegalArgumentException("TextDocsBatchableInput requires a TextDocsInputDataSet");
```
bzhangam: Why not use the same error as the below one?

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+     * True only when there is a config, a handler for the input type, and the request needs more than
+     * one sub-batch. Otherwise the caller should invoke the model directly.
+     */
+    public boolean shouldBatch(MLInput input, BatchInferenceConfig config) {
+        if (config == null) {
+            return false;
+        }
+        BatchableInput handler = registry.get(input);
+        if (handler == null) {
+            return false;
```
bzhangam: Why we silently skip the batch here? If the config is defined then the user will expect the predict to the model should be batched properly. If the input type is not supported we should throw an error to let user know the batch config is not working for the model.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        return splitter.split(handler.toItems(input), config).size() > 1;
```
bzhangam: Why we want to split it now and later in the execute we will split it again? Let's avoid this and split just one time. Maybe there is no need to have shouldBatch. We always run execute but fall back to invoker.invoke(input, listener); when no batch config or only one batch.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+        final int total = batches.size();
+        if (log.isDebugEnabled()) {
+            int items = 0;
+            for (List<BatchItem> b : batches) {
+                items += b.size();
+            }
+            log.debug("Size-based batching: split {} items into {} sub-batches", items, total);
+        }
+        final AtomicReferenceArray<MLOutput> results = new AtomicReferenceArray<>(total);
+        final AtomicInteger remaining = new AtomicInteger(total);
```
bzhangam: Why we need two atomic to count when the work is completed? I think one is enough just need to check if it's 0. Just remember to decrement it in error case.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+        for (int i = 0; i < total; i++) {
+            final int index = i;
+            final MLInput subInput = handler.merge(input, batches.get(i));
+            final ActionListener<MLOutput> subListener = ActionListener.wrap(output -> {
+                results.set(index, output);
+                if (remaining.decrementAndGet() == 0 && completed.compareAndSet(false, true)) {
+                    try {
+                        listener.onResponse(handler.combine(toList(results)));
+                    } catch (Exception combineError) {
+                        listener.onFailure(combineError);
```
bzhangam: How do we combine the error? We simply do listener.onFailure(error);

plugin/src/main/java/org/opensearch/ml/batch/BatchSplitter.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public List<List<BatchItem>> split(List<BatchItem> items, BatchInferenceConfig config) {
+        if (items == null || items.isEmpty()) {
+            throw new IllegalArgumentException("Cannot split an empty item list");
+        }
+        if (config == null) {
+            throw new IllegalArgumentException("BatchInferenceConfig must not be null");
+        }
+
+        int maxItems = config.getMaxItemsPerRequest();
```
bzhangam, line 33: why we assume the item limit is always enabled?

plugin/src/main/java/org/opensearch/ml/batch/SizeBasedBatchSplitter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+import org.opensearch.ml.common.model.BatchInferenceConfig;
+
+/**
+ * Greedily packs ordered items into sub-batches that respect the count and (optional) byte ceilings.
+ * Preserves order, never drops an item, and gives an over-sized item its own sub-batch rather than
+ * splitting it.
+ */
+public class SizeBasedBatchSplitter {
```
bzhangam: "Size-based" implies byte-size only, but it actually splits on two dimensions: item count (max_items_per_request) and byte size
  (max_bytes_per_request). A reader seeing SizeBasedBatchSplitter would reasonably assume it's only about payload size in bytes.

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public MLInput merge(MLInput source, List<BatchItem> items) {
+        List<String> docs = new ArrayList<>(items.size());
+        for (BatchItem item : items) {
+            docs.add((String) item.getPayload());
+        }
+        TextDocsInputDataSet subDataSet = asTextDocs(source).toBuilder().docs(docs).build();
+        return source.toBuilder().inputDataset(subDataSet).build();
```
bzhangam, line 53: Worth to document. It's an implicit assumption worth documenting — that MLAlgoParams must have per-request (not per-item or batch-relative) semantics for batching to be transparent. If a future parameter violates that assumption, the BatchableInput handler for that type would need to handle it explicitly.

heemin32:
1. Can we only use byte size? Is there a model which limit the number of request in a single batch?
2. Can we enable this feature by default with reasonable default value? What would be the risk of it?
3. What is the current behavior when the requests go above count limit and byte limit each?

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
RexyZhang2001: I was referring to Configuration section in the [RFC](https://github.com/opensearch-project/ml-commons/issues/4889), which specifically mentioned we should support `max_bytes_per_request` as -1 by default. Do we need to change it or keep it as it is? We could revisit in the RFC, so this PR doesn't diverge from it

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
RexyZhang2001: make sense, will improve this part

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Yes, need to throw a clear error naming the unsupported input type

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Make sense, we could remove `shouldBatch` , so `execute()` now decides internally (no config / no handler / single sub-batch → pass through directly) and computes the split exactly once.

plugin/src/main/java/org/opensearch/ml/batch/BatchSplitter.java
RexyZhang2001: will address together with the comment above

plugin/src/main/java/org/opensearch/ml/batch/SizeBasedBatchSplitter.java
RexyZhang2001: Could rename it to `BatchSplitter`

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
RexyZhang2001: will add document

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
RexyZhang2001: These two throw exceptions for different reasons：
- `asTextDocs` validates the `MLInput` passed into the handler — a caller handing us the wrong input type is a bad argument, so ·IllegalArgumentException·.
- `asTensorOutpu` validates the `MLOutput` the model returned — that's not a method argument, it's an internal invariant ("a batchable text-docs call should always come back as `ModelTensorOutput`"), so `IllegalStateException`.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Yes, one counter is enough. 
We could keep a single remaining counter that decrements on both success and failure, and only decide what to do when it hits zero (rather than completing eagerly on first success). By then every sub-batch has settled, so there are no null slots to trip over.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: We could collect failures into a per-request list as they come in, and at completion: if there were no failures, `combine()` and respond; if there was exactly one, surface it as-is (preserves its type/status, e.g. a 429); if there were several, wrap them into one error with the rest attached via `addSuppressed()` so nothing is silently dropped.

RexyZhang2001:
> 1. Can we only use byte size? Is there a model which limit the number of request in a single batch?
> 2. Can we enable this feature by default with reasonable default value? What would be the risk of it?
> 3. What is the current behavior when the requests go above count limit and byte limit each?


1. I think we need both. Byte size handles "payload too large", but some model impose a hard item-count cap that's independent of size. Cohere's embed API rejects more than 96 texts per call. OpenAI's caps at 2048, no matter how short each text is. 

2. If we batch by default, a model that makes one call today would silently start making several concurrent calls after an upgrade. The user changed nothing and didn't opt in, but their latency, billing, and throttling behavior all change. It is better to keep it user opt-in for now.

3. 
currently: If the request is too big, the model rejects it and returns an error, and that error comes back as an exception. The request fails silently truncated or dropped.

this PR: The splitter packs greedily and starts a new sub-batch as soon as adding the next item would exceed **either** ceiling 
    - Over the count limit → cut once the current sub-batch already holds `max_items_per_request` items.
    - Over the byte limit → cut once adding the next item would push the running byte total past `max_bytes_per_request`.
    - An item that's larger than the byte limit all by itself gets its own sub-batch (guarded by `!current.isEmpty()`), rather than being dropped, rejected, or split mid-item.

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
bzhangam: After thinking twice I feel a better UX can be we require at least one limit provided by the user. For the limit not provided we default to -1 to not enforce that limit.

Since different models can have different limit I feel it's better to require the user to provide it rather than we default to some value. And it's also possible the user only wants to use one limit and in that case for the limit they don't provide we simply default to no limit. I can update the RFC to reflect this change if there is no concern about this proposal.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
bzhangam: When will we call `addSuppressed`? I think here the combineError is just the exception for one call. Not sure why you call it combineError and also don't know when we combine the error from different calls.

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
bzhangam: I mean the error message. The below one is more clear about what wrong type we got. Easier for debug when needed.

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
RexyZhang2001: @heemin32 suggested we could consider setting the default value for each model in the blueprint. That way we have default values when creating connectors, and the user can customize them while creating model as implemented in the PR. 
logic graph:
```
  Blueprint pre-fills defaults          User optionally sets
  on the Connector                      batch_inference_config on the Model
          │                                       │
          └──────────────────┬────────────────────┘
                             ▼
        For each limit (max_items, max_bytes) independently:
        ┌─────────────────────────────────────────────────┐
        │  user set it?  ─yes─►  use user's value           │
        │       │no                                         │
        │  connector default? ─yes─►  use connector default │
        │       │no                                         │
        │       └─────────────────►  -1 (no limit)          │
        └─────────────────────────────────────────────────┘
                             ▼
              Validate the resolved config:
              both == -1  ─►  throw "at least one limit required"
              otherwise   ─►  proceed to split

```
Each limit is resolved independently through the chain **user value** → **connector default** → **-1**

plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java
RexyZhang2001: right, could improve it

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Sorry, It's something I implemented locally as the second commit, haven't pushed it yet. 
It works like this: 
as each sub-batch settles, we collect any failure into a per-request list, and we only decide what to do once all sub-batches have finished (tracked by a single counter hitting zero). At that point:
- no failures → merge the successful outputs and respond,
- exactly one failure → surface it as-is, so its type/status is preserved (e.g. a 429),
- multiple failures → wrap them into one error with the rest attached, so nothing gets silently dropped.

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
RexyZhang2001: After discussion, I think the current design which requires the user to specify sizes in two separate places is redundant and not great for customer experience.

Example: a user has 1000 documents to ingest and sets batch_size = 100 during ingestion, so the documents are split into 10 chunks of 100. Each chunk then goes to the model, where the user is asked to separately set `max_items`/`max_bytes `at model registration.

- If the user sets these too generously (e.g. max_items = 512, max_bytes = 400000), each chunk of 100 already fits comfortably within those limits, so this feature effectively does nothing — the user configured it, but it never engages.
- The user doesn't set `batch_inference_config` at all, and it defaults to the `max_items`/`max_bytes` shipped in that model's blueprint. But the blueprint default can just as easily be too generous relative to `batch_size`.  It's still a fixed number, not one that's aware of how the user is chunking on the ingest side. Same outcome: the feature is present, but never engaged.

Either way, the user ends up answering a size question twice, once for batch_size, once for the model, without a clear signal for why the second one matters, or what value makes sense. And that second number is genuinely hard to get right: it varies by provider, by model, even by model version, and there's no easy way for a user to test and discover the "correct" value for their specific endpoint.

heemin32: batch_inference_config serves a different purpose than batch_size. Consider a case where an unusually large document gets ingested from time to time. The customer won't adjust batch_size for individual document sizes. batch_inference_config acts as a safeguard to keep requests from exceeding the model's limit. On top of that, batch_size only applies to the ingest processor. How do we handle batched requests coming from search when they exceed the model limit?

> And that second number is genuinely hard to get right: it varies by provider, by model, even by model version, and there's no easy way for a user to test and discover the "correct" value for their specific endpoint.

 Why is that? I'd argue this second number is actually easier to get right. You can just increase the request count (or size per call) until you hit the model's limit, and you've discovered the correct value in one pass.

RexyZhang2001: > Why is that? I'd argue this second number is actually easier to get right. You can just increase the request count (or size per call) until you hit the model's limit, and you've discovered the correct value in one pass.

I can think of these:
1. Each model has different versions. For example, Cohere has v3 and v4, and each has different size limitations. This can apply to other models as they keep rolling out new versions—their limits tend to improve. We may need to constantly update these values in blueprints. 
2. We could manage to test out the max-item-per-request count, but the max-byte-per-request may be hard to test. Most models now use tokens instead of bytes as the measurement method.
3. Some models have three limitations that can restrict the model, such as OpenAi which has `number of items` (2048), `tokens per item` (8192), and `total tokens per request` (300k). Only having two limitations is not enough.

bzhangam: 1. It's hard to have the max limit in the blueprints since different model version may share the same blueprint but with different limit. I think we can provide some general good/well known limit in blueprint and ensure those limit should under the model actual limit. If we are not sure we simply don't provide a recommended limit in the blueprint.
2. Using token size is more accurate but it's tricky to calculate the token size from our local since different model can have different algorithm to calculate the token size. Probably we can use a generally good algorithm to calculate the approximate token size and support the token size based limit. But will need to call out we do approximate not actual token size calculation.
3. We can introduce a new limit for size per item since the implementation is not hard. User can choose which limit they want to use for their model.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+                } catch (Exception outputMergeError) {
+                    listener.onFailure(outputMergeError);
+                }
+                return;
+            }
+        }
+    }
+
+    private void onSubBatchFailure(AtomicInteger state, Exception error, ActionListener<MLOutput> listener) {
+        while (true) {
```
bzhangam: why we want to hold a thread and stay in this while loop when one sub batch fails? Why not simply store its failure and release the thread? If the failed sub batch happens to be the last finished one then we just call listener with all the failure info we collected from the failed sub batches.

bzhangam: You can refer to this implementation: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L421

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private List<MLOutput> toList(AtomicReferenceArray<MLOutput> array) {
+        List<MLOutput> ordered = new ArrayList<>(array.length());
+        for (int i = 0; i < array.length(); i++) {
+            ordered.add(array.get(i));
+        }
+        return ordered;
+    }
+
+    private void invokeWithRetry(SingleBatchInvoker invoker, MLInput subInput, int attempt, ActionListener<MLOutput> listener) {
```
bzhangam: Why we need to implement the retry here? Why not implement the retry in the invoker itself so that easier to reuse it? Also wonder we don't have existing retry for the invoker?

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -586,7 +615,32 @@ private void runPredict(
...
+                        // If the model defines batch inference limits and the request exceeds them, split it into
+                        // size-bounded sub-batches, run each against the model, and merge the results in input order.
+                        // Output validation and task completion still run once, on the merged response. Requests that
+                        // are not configured for batching, or already fit within the limits, take the single-call path.
+                        if (mlTask.getTaskType().equals(MLTaskType.BATCH_PREDICTION)) {
+                            predictor.asyncPredict(mlInput, trackPredictDurationListener, channel); // with listener
+                        } else {
+                            BatchInferenceConfig batchInferenceConfig = getBatchInferenceConfig(modelId);
+                            final Predictable batchPredictor = predictor;
+                            batchInferenceExecutor
```
bzhangam, line 618: why not simply pass the mlInput, trackPredictDurationListener, channel and the predictor to the batchInferenceExecutor.execute to have the detail in the executor? No need to surface the detail in the runner.

As I commented above the batchPredictor.asyncPredict has no retry support so we need to implement it again in the executor?

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
```diff
@@ -10,6 +10,7 @@
 import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
 import static org.opensearch.ml.common.CommonValue.VERSION_2_19_0;
 import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_8_0;
```
heemin32: We might want to wait for VERSION_3_9_0 to be available and switch it to 3_9_0 as 3_8_0 has just been released already.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+
+        final List<List<BatchItem>> batches;
+        try {
+            batches = splitter.split(handler.toItems(input), config);
+        } catch (Exception e) {
+            listener.onFailure(e);
+            return;
+        }
+
+        if (batches.size() == 1) {
```
heemin32, line 82: Even for batch size 1, can we just call `dispatchBatches(input, handler, batches, invoker, listener);`? We can move this logic inside `dispatchBatches` to simplify the code.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
heemin32: It won't hold the thread for long. Anyway, `while(true)` still feels a bit risky, so I'd rather avoid it.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
heemin32: Or the connector already has retry logic? https://github.com/opensearch-project/ml-commons/pull/4882/changes

plugin/src/main/java/org/opensearch/ml/batch/BatchSplitter.java
```diff
@@ -0,0 +1,58 @@
...
+            throw new IllegalArgumentException("BatchInferenceConfig must not be null");
+        }
+
+        boolean itemLimitEnabled = config.isItemLimitEnabled();
+        int maxItems = config.getMaxItemsPerRequest();
+        boolean byteLimitEnabled = config.isByteLimitEnabled();
+        long maxBytes = config.getMaxBytesPerRequest();
+
+        List<List<BatchItem>> batches = new ArrayList<>();
+        List<BatchItem> current = new ArrayList<>();
```
heemin32: Either we could pre-calculate the size or use `LinkedList<>`?

heemin32:
@RexyZhang2001 Could we do a benchmark to see if there are any big degradation in terms of throughput. For example, with 10M documents to ingest, baseline: 1000 batch size, variant1: 2000 batch size which will trigger split by 2, variant2: 3000 batch size which will trigger split by 3.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Avoided holding the thread and changed it to storing the succeeded sub-batches and failed sub-batches.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Right, we could simply use the retry logic on the connector side. Implementing another retry could cause nested retry.

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
RexyZhang2001: Simplied this

common/src/main/java/org/opensearch/ml/common/transport/register/MLRegisterModelInput.java
RexyZhang2001: Changed to 3_9_0

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Moved one batch-size case into `dispatchBatches`

plugin/src/main/java/org/opensearch/ml/batch/BatchSplitter.java
RexyZhang2001: Used arrayList for pre-calculation, please check again.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+        final List<Exception> failures = Collections.synchronizedList(new ArrayList<>());
+        // Decremented once per sub-batch, on success and on failure alike, so exactly one callback sees
+        // zero and completes the listener.
+        final AtomicInteger remaining = new AtomicInteger(total);
+
+        for (int i = 0; i < total; i++) {
+            final int index = i;
+            final List<BatchItem> batch = batches.get(i);
+            final ActionListener<MLOutput> subListener = ActionListener.wrap(output -> {
+                results.set(index, output);
```
heemin32: If results.set(index, output) or failures.add(error) throws an exception, remaining.decrementAndGet() never runs, the counter never reaches zero, and listener is never called — leaving the caller hanging indefinitely.


Wrap the body of each lambda so that any unexpected exception still decrements the counter and triggers completion:

```
  final ActionListener<MLOutput> subListener = ActionListener.wrap(output -> {
      try {
          results.set(index, output);
      } finally {
          if (remaining.decrementAndGet() == 0) {
              complete(failures, results, handler, listener);
          }
      }
  }, error -> {
      try {
          failures.add(error);
      } finally {
          if (remaining.decrementAndGet() == 0) {
              complete(failures, results, handler, listener);
          }
      }
  });
```


You should also consider whether complete(...) itself can throw — if it does, the exception propagates out of the finally block and the caller's listener still won't be called cleanly. Wrapping complete(...) in a try/catch that calls listener.onFailure(...) directly would close that gap too.

RexyZhang2001: Makes sense; added exception handling

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
```diff
@@ -586,7 +610,16 @@ private void runPredict(
                                 // recordPredictMetrics(modelId, durationInMs, output, internalListener);
                             }
                         }, e -> handlePredictFailure(mlTask, internalListener, e, shouldTrackRemoteFailure(e), modelId, actionName));
-                        predictor.asyncPredict(mlInput, trackPredictDurationListener, channel); // with listener
+                        // If the model defines batch inference limits and the request exceeds them, split it into
+                        // size-bounded sub-batches, run each against the model, and merge the results in input order.
+                        // Output validation and task completion still run once, on the merged response. Requests that
+                        // are not configured for batching, or already fit within the limits, take the single-call path.
+                        if (mlTask.getTaskType().equals(MLTaskType.BATCH_PREDICTION)) {
```
bzhangam, line 615: Why we can rely on the MLTaskType.BATCH_PREDICTION to decide if we should use batchInferenceExecutor?

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,217 @@
...
+        );
+    }
+
+    /**
+     * Splits into sub-batches only when there is a config, a handler for the input type, and the request
+     * needs more than one sub-batch. A request runs as a single call when the model has no config or
+     * already fits within the limits, and fails when the model has a config but the input type has no
+     * handler, rather than being sent unsplit.
+     */
+    public void execute(MLInput input, BatchInferenceConfig config, SingleBatchInvoker invoker, ActionListener<MLOutput> listener) {
```
bzhangam: If this function is only invoked in this class better to make it private and call it doExecute. Same pattern used across OpenSearch.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,217 @@
...
+            return failures.get(0);
+        }
+        StringBuilder message = new StringBuilder().append(failures.size()).append(" of the sub-batches failed: ");
+        for (int i = 0; i < failures.size(); i++) {
+            if (i > 0) {
+                message.append("; ");
+            }
+            message.append(failures.get(i).getMessage());
+        }
+        Exception combined = new OpenSearchStatusException(message.toString(), RestStatus.INTERNAL_SERVER_ERROR);
```
bzhangam: I think we should not simply use INTERNAL_SERVER_ERROR. We should decide the status based on the exception of each failed sub-batch. If all sub-batch failed due to 403 then this should also be a 403. Only if when there is an error with status >= 500 we can use INTERNAL_SERVER_ERROR.

plugin/src/main/java/org/opensearch/ml/batch/SingleBatchInvoker.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.ml.common.input.MLInput;
+import org.opensearch.ml.common.output.MLOutput;
+
+/**
+ * Runs inference for one sub-batch, decoupling the batching logic from how a model call is made.
+ */
+@FunctionalInterface
+public interface SingleBatchInvoker {
```
bzhangam: I feel it's kind weird that we define the interface here but implement it in BatchInferenceExecutor as a function pass to another function. Wound't it be cleaner we implement it here and BatchInferenceExecutor simply use it?

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
RexyZhang2001: The two "batch" things here are unrelated. `MLTaskType.BATCH_PREDICTION` is offline batch inference, the request came in through `POST /_plugins/_ml/models/{id}/_batch_predict`. That has nothing to do with our `batch_inference_config` approach.

It can only be one of two values here: `MLTaskType.BATCH_PREDICTION` and `MLTaskType.PREDICTION`. `taskType` is assigned in exactly one place in this class. — the `taskType(actionType.equals(ActionType.BATCH_PREDICT) ? MLTaskType.BATCH_PREDICTION : MLTaskType.PREDICTION).inputType(inputDataType)` in `executeTask` picks either `BATCH_PREDICTION` or `PREDICTION`. The other `MLTaskType` values belong to the other runners (`MLTrainingTaskRunner`, `MLExecuteTaskRunner`, etc.) and never reach here.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: renamed it

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Referred to this method in https://github.com/opensearch-project/OpenSearch/blob/3eed58bc083142a1beb8f21102cd8785c0513aeb/server/src/main/java/org/opensearch/action/search/SearchPhaseExecutionException.java#L107. It now takes the first failure's status and only escalate if some sub-batch hit a 5xx. So all-403 gives 403, 403 + 429 gives 403, and 403 + 502 gives 502.

plugin/src/main/java/org/opensearch/ml/batch/SingleBatchInvoker.java
RexyZhang2001: moved the implementation into `SingleBatchInvoker`

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
bzhangam: ok let's add some comment to explain this otherwise it's confusing. From the the task type we cannot tell it's for offline inference and we cannot apply the batch inference config we define. All the rate limit and batch strategy is handled by the remote service when we use the offline batch mode.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+        );
+    }
+
+    /**
+     * Splits into sub-batches only when there is a config, a handler for the input type, and the request
+     * needs more than one sub-batch. A request runs as a single call when the model has no config or
+     * already fits within the limits, and fails when the model has a config but the input type has no
+     * handler, rather than being sent unsplit.
+     */
+    public void doExecute(MLInput input, BatchInferenceConfig config, SingleBatchInvoker invoker, ActionListener<MLOutput> listener) {
```
bzhangam: why still public method?

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+        MLInput input,
+        BatchInferenceConfig config,
+        Predictable predictor,
+        TransportChannel channel,
+        ActionListener<MLTaskResponse> listener
+    ) {
+        doExecute(
+            input,
+            config,
+            SingleBatchInvoker.forPredictor(predictor, channel),
```
bzhangam: After thinking twice I feel the `SingleBatchInvoker` abstraction is redundant — we already have the `Predictable` interface which is exactly the thing that knows how to call the remote model. Introducing a second concept ("invoker") that maps 1:1 to `predictor.asyncPredict()` adds indirection without a clear benefit.

I'd suggest passing the `Predictable` and `TransportChannel` directly into the executor:

```java
batchInferenceExecutor.execute(mlInput, batchInferenceConfig, predictor, channel, listener);
```

And inside the executor's loop, just call:

```java
predictor.asyncPredict(subInput, subListener, channel);
```

This avoids introducing a new concept when an existing one already covers the same responsibility. If we ever need to swap in a different invocation strategy in the future, we can refactor at that point — but today there's only one implementation, so the extra layer just makes readers wonder "how is an invoker different from a predictor?"

bzhangam (approved):

plugin/src/main/java/org/opensearch/ml/task/MLPredictTaskRunner.java
RexyZhang2001: Added in new commit

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Removed this method in favor of removing `SingleBatchInvoker`

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Changed in new commit

heemin32 (approved):

RexyZhang2001:
Benchmark result: https://github.com/opensearch-project/ml-commons/issues/4889#issuecomment-5398072424

jiapingzeng:
@RexyZhang2001 could you improve test coverage? CI is failing with
```
[ant:jacocoReport] Rule violated for class org.opensearch.ml.batch.BatchableInputRegistry: branches covered ratio is 0.5, but expected minimum is 0.7
[ant:jacocoReport] Rule violated for class org.opensearch.ml.batch.BatchItem: branches covered ratio is 0.5, but expected minimum is 0.7

FAILURE: Build failed with an exception.

> Task :opensearch-ml-search-processors:jacocoTestCoverageVerification
> Task :opensearch-ml-search-processors:check
> Task :opensearch-ml-search-processors:build

```

jiapingzeng (approved):
LGTM!

PR #4903, opened 2026-07-13
Fix flaky CI: httpbin dependency, MCP tool sync race, SearchModelGroupITTests hang

### LINKED ISSUE #4639: [Flaky Test] SearchModelGroupITTests.test_term_search timeout
**What is the bug?**
This term query should not timeout, creating an issue to track this

> Task :opensearch-ml-plugin:test

REPRODUCE WITH: gradlew ':opensearch-ml-plugin:test' --tests 'org.opensearch.ml.action.model_group.SearchModelGroupITTests.test_term_search' -Dtests.seed=3CF6E3AE518F8566 -Dtests.security.manager=false -Dtests.locale=ur-PK -Dtests.timezone=EAT -Druntime.java=21

SearchModelGroupITTests > test_term_search FAILED
    java.lang.Exception: Test abandoned because suite timeout was reached.
        at __randomizedtesting.SeedInfo.seed([3CF6E3AE518F8566]:0)

REPRODUCE WITH: gradlew ':opensearch-ml-plugin:test' --tests 'org.opensearch.ml.action.model_group.SearchModelGroupITTests.test_term_search' -Dtests.seed=3CF6E3AE518F8566 -Dtests.security.manager=false -Dtests.locale=ur-PK -Dtests.timezone=EAT -Druntime.java=21
SearchModelGroupITTests > classMethod FAILED

    java.lang.Exception: Suite timeout exceeded (>= 1800000 msec).

 - Plugins

rithinpullela:
## Summary

Fixes three recurring CI failures identified across all failed `Build and Test ml-commons` runs from 2026-06-18 to 2026-07-10.

- **`testRuntimeParameterSubstitutionInHeaders`** — Replace the live httpbin.org dependency with a local loopback echo server. The test only needs header-echo semantics to verify `${parameters.*}` substitution; a flaky public service is unnecessary. Skipped on the dockerized leg where the containerized node cannot reach the test-JVM loopback.

- **`SearchModelGroupITTests` suite-timeout hang** — Five unit test classes install mock `ResourceSharingClient` on the JVM-global singleton without cleanup. A leaked feature-enabled mock causes `SearchModelGroupTransportAction` to take the resource-authz branch, calling `getAccessibleResourceIds()` on an unstubbed mock whose no-op never invokes the `ActionListener`. This hangs `actionGet()` until the 20–30 min suite timeout. Fix: add `tearDown()` resets and bound all `actionGet()` calls with a 30s timeout.

- **`testChatAgentWithMcpStreamableHttpConnector`** — `@Ignore` the test. The MCP tool-registration race (fire-and-forget addTool + round-robin readiness poll) is not fixable test-side; requires product changes to make `addTool` await the sync chain.

### Related Issues

Resolves https://github.com/opensearch-project/ml-commons/issues/4639

### Check List
- [x] New functionality includes testing (test-only change).

### DIFF
--- a/plugin/src/test/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportActionTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportActionTests.java
@@ -132,6 +132,15 @@ public class DeleteModelGroupTransportActionTests extends OpenSearchTestCase {
         ResourceSharingClientAccessor.getInstance().setResourceSharingClient(null);
     }
 
+    @Override
+    public void tearDown() throws Exception {
+        try {
+            ResourceSharingClientAccessor.getInstance().setResourceSharingClient(null);
+        } finally {
+            super.tearDown();
+        }
+    }
+
     @Test
     public void testDeleteModelGroup_Success() throws InterruptedException {
 
--- a/plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
+++ b/plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
@@ -16,6 +16,7 @@ import org.apache.hc.core5.http.message.BasicHeader;
 import org.junit.After;
 import org.junit.Assume;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.opensearch.client.Response;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.ml.common.MLTaskState;
@@ -47,6 +48,7 @@ public class RestChatAgentWithMcpConnectorIT extends MLCommonsRestTestCase {
     private static final String AWS_SESSION_TOKEN = System.getenv("AWS_SESSION_TOKEN");
     private static final String REGION = "us-west-2";
     private static final String MODEL_ID_BEDROCK = "us.anthropic.claude-haiku-4-5-20251001-v1:0";
+    private static final int MCP_TOOLS_SYNC_INTERVAL_SECONDS = 10;
 
     // Random suffix so the LLM can't hallucinate the canonical "iris_data" and pass the assertion
     // without actually calling ListIndexTool through MCP.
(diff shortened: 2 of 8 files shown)


plugin/src/test/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportActionTests.java
```diff
@@ -132,6 +132,15 @@ public void setup() throws IOException {
         ResourceSharingClientAccessor.getInstance().setResourceSharingClient(null);
     }
 
+    @Override
```
mingshl, line 135: good catch!

plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
```diff
@@ -129,6 +141,7 @@ public void teardown() throws IOException {
         deleteIndexWithAdminClient(irisIndex);
     }
 
+    @Ignore("Flaky: MCP tool registration race not fixable test-side; re-enable once addTool awaits sync")
```
mingshl, line 144: I am fine with ignoring the test that is always causing problem, but please document the issue and make sure we get back to fix them when we have a good way around. Please a tracking issue for the product fix. cc @RexyZhang2001 This is the same issue that Rex was asking before

mingshl (approved):
Thanks for making the fix, the failing CI lately were complained by multiple contributors. Thanks for stepping in and help the CI flow through

plugin/src/test/java/org/opensearch/ml/action/model_group/DeleteModelGroupTransportActionTests.java
mingshl: without cleanup is exactly the kind of subtle state leak

plugin/src/test/java/org/opensearch/ml/rest/RestChatAgentWithMcpConnectorIT.java
rithinpullela: Thanks Mingshi, I created an Issue to fix the race condition in the MCP server that we have today: https://github.com/opensearch-project/ml-commons/issues/4905

PR #4900, opened 2026-07-10
Fix flaky CI: httpbin dependency, MCP tool sync race, SearchModelGroupITTests hang (#4639)

rithinpullela:
### Description

Fixes three recurring CI failures (analysis of all failed `Build and Test ml-commons` runs from 2026-06-18 to 2026-07-10).

**1. `RestMLRemoteInferenceIT.testRuntimeParameterSubstitutionInHeaders` — remove the live httpbin.org dependency (8 failing runs)**

The test verifies ml-commons' `${parameters.*}` substitution into outgoing connector headers; httpbin.org/post was only used as an echo mirror to reflect the sent headers back. httpbin.org is a best-effort public service and its outages (503s) failed CI even though the retry added in #4896 was present and exhausted (~4 attempts over 18s vs multi-minute outages).

Replaced with a local loopback `HttpServer` (JDK built-in, no new dependency) that echoes request headers with the same `{"headers": {...}}` contract — the same resolution opensearch-py adopted for the identical problem (opensearch-project/opensearch-py#395, closing opensearch-project/opensearch-py#327), and the same pattern this repo already uses in `MockLLM.java`. The substitution coverage is unchanged; the test is now hermetic. Skipped on the dockerized leg (`tests.clustername=docker-cluster`) where the containerized node cannot reach the test-JVM loopback; the substitution logic is node-identical and fully covered by the testClusters legs.

**2. `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector` — wait for MCP tool sync before executing the agent (9 failing runs)**

On the multi-node leg, MCP tool registration is fire-and-forget and each node loads tools into its in-memory registry on a ~10s sync cycle. The setup readiness poll uses `client()`, which round-robins across nodes, so it can pass as soon as any one node has synced — while the agent's MCP connector is pinned to a single host (`getClusterHosts().get(0)`) that may not have synced yet. The agent then resolves zero tools and sends Bedrock an empty `tools` array, which Bedrock rejects with `400 "The provided request is not valid"`.

[...]

PR #4907, opened 2026-07-13
Serve /_plugins/_ml/mcp on demand from the index instead of a per-node cache

### LINKED ISSUE #4905: [PROPOSAL] Serve /_plugins/_ml/mcp by building the MCP server on demand from the index, instead of a per-node in-memory cache synced by a background thread
## Summary

The MCP server serves `tools/list` and `tools/call` from a per-node in-memory tool registry. Each node keeps that registry warm through two mechanisms: a fire-and-forget broadcast to all nodes on every tool write, and a background thread that re-reads the `.plugins-ml-mcp-tools` index every 10 seconds. This design creates a consistency race and carries ongoing complexity that the feature does not need.

This proposal replaces the cache with on-demand construction: build the stateless MCP server at request time, loading tools directly from the `.plugins-ml-mcp-tools` index, which is already the source of truth. Measured server construction cost is under 0.1 ms for 100 tools. The change removes the background sync job, the cross-node broadcast actions, the in-memory version-tracking map, and one class of partial-failure responses, and it makes a newly registered tool servable on the next request.

## Background: how the MCP server works today

The read path is node-local. A client POSTs JSON-RPC to `/_plugins/_ml/mcp`. `RestMcpServerAction` forwards it to `TransportMcpServerAction`, a `HandledTransportAction` that runs on whichever node the REST layer selected. That node answers `tools/list` and `tools/call` from a single long-lived `McpStatelessAsyncServer` instance held by `McpStatelessServerHolder`. The SDK server keeps its tools in an in-memory list; there is no hook to load tools from the index per request.

Because any node can receive a `/mcp` request, every node must hold the full tool set in memory. Two mechanisms keep each node's copy current:

1. **Broadcast on write.** Register, update, and remove each write to the `.plugins-ml-mcp-tools` index first (with `RefreshPolicy.IMMEDIATE`), then broadcast to all nodes through a `TransportNodesAction` (`MLMcpToolsRegisterOnNodesAction` and siblings). Each node applies the change to its own SDK server and updates a static `IN_MEMORY_MCP_TOOLS` map that tracks tool name to document version.

2. **Background sync every 10 seconds.** `McpStatelessServerHolder.startSyncMcpToolsJob` reschedules `autoLoadAllMcpTools` every 10 seconds (`SYNC_MCP_TOOLS_JOB_INTERVAL = 10`). It re-reads the index, diffs against `IN_MEMORY_MCP_TOOLS` by version, and adds or upgrades changed tools. This is the convergence path for nodes that were down, partitioned, or not yet joined when a broadcast fired.

The index is the source of truth. The in-memory registry is a derived per-node cache. Register, update, and remove validate against the index, write to the index first, and only then touch memory. A comment in `TransportMcpToolsRemoveAction` states the design intent directly: "the failure on memory operation won't have impact on consistency."

## Problem

The cache produces two costs.

**Consistency race.** Registration is fire-and-forget from the caller's point of view, and each node loads tools on a 10-second cycle. The readiness poll in the integration test uses `client()`, which round-robins across nodes, so it can pass as soon as one node has synced. The agent's MCP connector, however, pins to a single host (`getClusterHosts().get(0)`). When that host has not yet synced, the agent resolves zero tools and sends Bedrock an empty `tools` array, which Bedrock rejects with a 400. This is the root cause of the flaky `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector`, which is now disabled with `@Ignore` (see #4900 / #4903). The test-side sync-wait mitigation did not close the window because the sync job reschedules only after each reload completes.

**Ongoing complexity.** The cache requires a dedicated sync thread pool, a background reschedule loop, a static version-tracking map, three `*OnNodesAction` broadcast transport actions, and version-diff reconciliation logic. Register, update, and remove each report a "persisted successfully but failed to register in mcp server memory" partial-failure state that exists only because memory and index can diverge.

## Proposed change

Serve each `/mcp` request from an MCP server built at request time, with tools loaded from the index.

On a `tools/list` or `tools/call` request:

1. Read all tool definitions from `.plugins-ml-mcp-tools` (capped at 1000 by `MAX_TOOL_NUMBER`). Loading the full set on every request keeps the server behavior identical to a real MCP server and stays correct if the MCP spec later allows batched requests that span multiple tools.
2. Build the tool specifications with the existing `McpToolsHelper.createToolSpecification`.
3. Construct the stateless server with the SDK builder's `.tools(list).build()`, which accepts all tools upfront and needs no async `addTool` chain.
4. Handle the JSON-RPC request against that server and return the response.

This removes `McpStatelessServerHolder`'s singleton, the `IN_MEMORY_MCP_TOOLS` map, `startSyncMcpToolsJob`, the sync thread pool, and the three `*OnNodesAction` broadcast actions. Register, update, and remove keep writing to the index and drop their memory-sync step.

Because register writes to the index with `RefreshPolicy.IMMEDIATE`, a per-request load is read-your-writes consistent. The moment `_register` returns, the next `/mcp` request on any node loads the new tool. This removes the race at its source rather than widening the sync window.

## Cost analysis

I decompiled the MCP SDK (`io.modelcontextprotocol.sdk:mcp:1.1.1`) and benchmarked server construction on an Apple M-series laptop, JDK 21, 50,000 iterations after 20,000 warmup iterations.

| Tools | Build server, tool specs reused (us/op) | Parse schemas and build, full cold rebuild (us/op) |
|-------|------------------------------------------|-----------------------------------------------------|
| 5     | 1.2                                      | 7.4                                                 |
| 20    | 3.1                                      | 19.7                                                |
| 50    | 8.2                                      | 44.4                                                |
| 100   | 23.8                                     | 92.1                                                |

The worst case, 100 tools with every JSON schema re-parsed from scratch, is 0.092 ms. The `McpStatelessAsyncServer` constructor allocates a few collections and a map of five handler lambdas, then sets one field on the transport. It starts no threads, opens no sockets, and performs no I/O.

The server object is not the meaningful cost. The added per-request cost is one search against `.plugins-ml-mcp-tools`, an index capped at 1000 documents. That search runs single-digit milliseconds and is dominated by the tool execution and LLM round-trips that gate every agent step. `/mcp` is agent-driven, not a high-QPS endpoint.

## Trade-offs

The serving path gains a dependency on the index that it does not have today. An agent step that calls a tool now reads the full tool set from the index once per request. Reusing a single shared `JsonMapper` and `DefaultJsonSchemaValidator` across requests keeps the build cheap, since both are thread-safe and the validator keeps its schema cache. If index read rate ever becomes a concern, a short-TTL spec cache keyed on document version restores most of the benefit without the background thread.

## History

This design was inherited, not chosen for the current server. The index, cache, and sync arrived with tool persistence (#3874) on the earlier stateful SSE server, where a long-lived server holding client sessions was structurally necessary. The stateless rewrite (#4162) changed the transport and session model but kept the tool-loading model unchanged. No prior discussion evaluated a per-request design.

## Scope

In scope: the `/mcp` serving path, `McpStatelessServerHolder`, the three `*OnNodesAction` broadcast actions, and the memory-sync steps in register, update, and remove.

Out of scope: the `.plugins-ml-mcp-tools` index schema, the register/update/remove REST APIs and their index writes, and the MCP client path (`McpConnectorExecutor`).

## Acceptance criteria

1. A tool registered through `_register` is servable through `tools/list` and `tools/call` on the next request on any node, with no fixed wait.
2. `tools/call` for a registered tool returns the tool result. `tools/call` for an unknown tool returns a JSON-RPC error.
3. The background sync thread pool, `IN_MEMORY_MCP_TOOLS`, and the three `*OnNodesAction` broadcast actions are removed.
4. `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector` is re-enabled and passes on the multi-node leg without a sync-wait.

## Related

- #4639 (SearchModelGroupITTests hang, same flaky-CI cleanup effort)
- #4900 / #4903 (disabled the affected MCP test with `@Ignore`)
- #4162 (stateful to stateless migration)
- #3874 (introduced tool persistence and the in-memory cache)

rithinpullela:
## Summary

Serve the MCP server endpoint (`/_plugins/_ml/mcp`) by building the stateless MCP server **per request** from the `.plugins-ml-mcp-tools` index, which is already the source of truth. This removes the per-node in-memory tool registry, the 10-second background sync job (and its thread pool), and the cross-node register/update/remove broadcast actions.

Resolves #4905.

### Why

The previous design kept each node's MCP server warm through two mechanisms: a fire-and-forget broadcast to all nodes on every tool write, and a background thread re-reading the index every 10 seconds. Because registration was fire-and-forget and each node synced on its own cycle, a freshly registered tool was not immediately servable on the node an MCP client happened to hit. This is the race that made `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector` flaky (it was disabled with `@Ignore` in #4900/#4903).

Building the server per request from the index makes registration **read-your-writes consistent** — register writes to the index with an immediate refresh, so the next request on any node loads the new tool. This removes the race at its root and deletes the cache, sync job, and broadcast machinery. The MCP SDK server object is a lightweight in-memory construction (measured <0.1 ms for 100 tools; no threads, sockets, or I/O), so the only added per-request cost is one search against a small (≤1000 doc) system index.

### Architecture: before and after

**Before** — every node keeps an in-memory copy of the tools, kept warm by a broadcast on each write plus a 10s background sync job. A request is served from whichever node the coordinator lands on, whose cache may be stale until the broadcast or the next sync arrives.

```
  WRITE (register/update/remove)                 SERVE  /_plugins/_ml/mcp
  ------------------------------                 --------------------------
     REST _register                                   MCP client
          |                                                |
          v                                                v
   [ .plugins-ml-mcp-tools ]  <---- 10s sync ----   coordinator node
          |                          job (pull)           |
          | broadcast (OnNodesAction)                     v
          |----------------> node 0  in-mem MCP server (served from cache;
          |----------------> node 1  in-mem MCP server  may be stale until
          |----------------> node 2  in-mem MCP server  broadcast/sync lands)

  State lives in 3 places: the index + every node's cache, reconciled by
  a broadcast and a 10s sync job.
```

**After** — the index is the only state. Each request builds a fresh, throwaway MCP server from the index, so any node serves the current tool set. No cache, no broadcast, no sync job.

```
  WRITE (register/update/remove)                 SERVE  /_plugins/_ml/mcp (any node)
  ------------------------------                 -------------------------------------
     REST _register                                   MCP client
          |                                                |
          v                                                v
   [ .plugins-ml-mcp-tools ]  <----- searchAllTools ---  coordinator node
     (IMMEDIATE refresh)       ----- tools ----------->        |
                                                               v
                                              build stateless MCP server per request
                                                               |
                                                               v
                                              handle tools/list | tools/call, discard

  State lives in ONE place: the index. A tool is servable on any node the
  moment _register returns (read-your-writes).
```

### Changes

- `TransportMcpServerAction` builds a fresh `McpStatelessAsyncServer` per request, loading tools via `McpToolsHelper.searchAllTools` and passing them to the SDK builder's `.tools(...)` upfront. Tools that fail to build (e.g. a tool whose plugin is no longer installed) are skipped and logged, so one bad tool does not fail `tools/list`/`tools/call` for the rest.
- `searchAllTools` treats a missing tools index as an empty list, so `tools/list`/`tools/call` succeed on a cluster with no tools registered.
- Register/Update/Remove write the index and return a local-node response directly, instead of broadcasting to `*OnNodesAction`.
- Deleted: `McpStatelessServerHolder`, the three `Transport*OnNodesAction` and their `MLMcpTools*OnNodesAction` markers, the `opensearch_mcp_tools_sync` thread pool, and the `JSON_RPC_SERVER_NOT_READY_ERROR` constant.
- Re-enabled `RestChatAgentWithMcpConnectorIT.testChatAgentWithMcpStreamableHttpConnector` and removed its sync-wait workaround.

### Testing

- MCP + plugin unit tests pass, including a new test that a tool which fails to build is skipped while healthy tools still serve.
- MCP integration tests pass on an ephemeral cluster, including the full `initialize` → `tools/list` → `tools/call` streaming flow.
- Verified on a live **2-node** cluster (`./gradlew run -PnumNodes=2`): a freshly registered tool is servable immediately on **both** nodes; removal takes effect immediately; empty/missing index returns `{"tools":[]}` (no 500); and a poison doc in the index (a tool type with no factory) does not take down the MCP surface.

### Backward compatibility note for reviewers

This removes the `*OnNodesAction` transport actions outright. During a mixed-version rolling upgrade, an old coordinator node could still write to the index (which succeeds) and then broadcast to an upgraded node that no longer registers that action, surfacing as an error on the write even though the tool was persisted. MCP is experimental and feature-flagged, serving is unaffected throughout, and no data is lost, so this is judged acceptable — but flagging it explicitly in case a no-op compatibility shim for one release is preferred.

### Known trade-off (documented)

Serving now performs one index search per request, and the per-request server build cost scales with the number of registered tools. For MCP's low-QPS, agent-driven usage this is well within budget; a short-TTL spec cache keyed on index state is a straightforward fast-follow if a deployment ever registers very large tool counts.

### Check List


nathaliellenaa (approved):
LGTM! My one concern is the backward compatibility gap during rolling upgrades from removing the `*OnNodesAction` transport actions. An old coordinator broadcasting to an upgraded node would hit an action-not-found error even though the index write succeeded. But as the PR description notes, MCP is experimental and feature-flagged, serving is unaffected, and no data is lost, so this is acceptable. Approved.

jiapingzeng (approved):

PR #4911, opened 2026-07-16
Memory Retention

### LINKED ISSUE #4859: [RFC] Memory Retention Lifecycle for Agentic Memory Containers
# RFC: Memory Retention Lifecycle for Agentic Memory Containers

> **Update (2026-07-16):** This RFC has been revised to reflect team decisions made during implementation. The most significant changes from the original proposal:
> 1. **Retention is opt-in** — no retention occurs unless the user explicitly sets a policy on their container, or a cluster admin explicitly configures default values. Out of the box, nothing is deleted.
> 2. **Cluster defaults are admin-configured and disabled by default** — the four `default_*` settings all ship as `-1` (off). Admins can optionally set them to establish organization-wide baselines. Without admin action, no policies are auto-applied.
> 3. **Session liveness auto-touch ships in v1** — adding working memory to a session now bumps the session's `last_updated_time`, so active conversations extend their session's lifetime.
> 4. **`max_count` eviction sorts by `last_updated_time` (LRU) for sessions and long-term, and by `created_time` (FIFO) for history.**
> 5. **`last_updated_time` is only bumped by content changes** — pinning, tagging, and metadata updates do not extend retention lifetime.
>
> **Update (2026-07-20) — final implementation as shipped in #4914 and #4918.** The following behaviors changed during implementation/review and are marked `[Updated]` throughout:
> 1. **`[Updated]` The feature is disabled by default.** The `plugins.ml_commons.memory.retention_enabled` master switch now defaults to **`false`** (was `true`). An admin must explicitly enable it; while disabled the job short-circuits and the memory-container/pinned APIs reject retention input (see items 6, 8).
> 2. **`[Updated]` `max_count` eviction sort finalized per type:** sessions and long-term evict the **least-recently-updated** (`last_updated_time` ASC — LRU); history evicts by `created_time` ASC (FIFO, append-only); the `disable_session` working-memory TTL ages by `created_time` (hard lifespan cap). This reverses the earlier "`created_time` for all types" decision.
> 3. **`[Updated]` `working_memory_ttl_days` defaults to `-1` (disabled).** Sessionless working memory is kept indefinitely unless an admin sets a positive TTL (was a 30-day default that "could not be disabled").
> 4. **`[Updated]` Lazy session creation.** When a client adds memory under its own `session_id` without calling create-session, the add path idempotently creates a minimal backing session doc (`OpType.CREATE`), so the orphan sweep does not delete live working memory.
> 5. **`[Updated]` First-observation grace period.** The orphan sweep records a per-container `orphan_sweep_baseline_time` and deletes nothing from a container until one full `orphan_ttl_days` window has elapsed since first observation, protecting pre-existing data on the first run.
> 6. **`[Updated]` Count-based ID collection capped at 50,000 per run** to bound heap; large backlogs converge over multiple runs.
> 7. **No session-record backfill for old data** (deliberate) — the system cannot distinguish "never created" from "user deleted it"; the grace period (item 5) handles the transition instead.
> 8. **`[Updated]` `pinned` is gated behind `retention_enabled`** — setting `pinned` while retention is disabled is rejected with HTTP 403 on the update path.
> 9. **`[Updated]` Combined strategy + retention update fix** — updating strategies and `retention_policy` in the same request no longer drops the retention change.
> 10. **Delivered as 2 PRs** — data model + API (#4914) and job + hardening (#4918) — not the original 7-PR stack.

## 1. Problem Statement

The agentic memory framework lacks any automated lifecycle management. Containers grow without bound, causing four issues:

1. **Unbounded storage growth** — no mechanism to reclaim space from obsolete memories
2. **Context-window pollution** — agents retrieve stale/contradictory information, degrading response quality
3. **Missing cascade semantics** — expired sessions would leave orphaned working memory
4. **Manual intervention required** — operators have no automated cleanup

Today there is no deletion job, no TTL field, and no retention configuration anywhere in the codebase.

## 2. Goals & Non-Goals

**Goals:**
- Per-container policy using `retention_days` and/or `max_count` for sessions and long-term memory, and `max_count` for history
- A background Job Scheduler job as the only enforcement mechanism — no read-path changes; minimal write-path additions for session liveness tracking
- Session expiry cascades: working memory deleted first, then the session document, as a unit
- Working memory never deleted on its own (protects conversation integrity)
- Cluster-level working-memory TTL for containers with `disable_session: true`
- Orphan sweep for working memory referencing missing sessions
- `pinned: true` exempts documents from eviction
- **Retention is opt-in** — no retention occurs unless the user explicitly sets a policy, or a cluster admin explicitly configures default values (all defaults ship as `-1`/disabled). An explicit `"retention_policy": null` prevents even admin-configured defaults from being applied to a specific container.
- Backward compatible — existing containers are unaffected unless an admin explicitly configures default retention settings, in which case containers without a policy (and not explicitly opted out) receive those admin-configured defaults via backfill on the next job run

**Non-goals (v1):** query-time filtering (staleness up to one job interval accepted), write-time eviction, LRU/access-based eviction, per-namespace or per-user `max_count`, AOSS Serverless support, cluster-wide admin retention ceiling, dry-run preview endpoint, on-demand purge endpoint.

## 3. Architecture

A 24-hour Job Scheduler job iterates every memory container. Containers with no retention policy are skipped unless a cluster admin has explicitly configured default retention settings (all default settings ship as `-1`/disabled). For each container that has an active policy, the job evaluates three memory types independently: sessions (`retention_days` + `max_count`, skip pinned), long-term memory (same), and history (`max_count` only, skip pinned). Expired sessions trigger a cascade in a strict order: (1) delete working memory where `namespace.session_id` matches expired IDs, (2) delete the session docs — WM-first order is crash-safe and prevents orphans. Special cases: `disable_session: true` containers get a `created_time`-based WM TTL, and a once-per-run orphan sweep cleans up working memory pointing at non-existent sessions.

### 4.1 Schema

New field on `MemoryConfiguration`: a `Map<MemoryType, RetentionRule>` named `retentionPolicy`. Valid keys are SESSIONS, LONG_TERM, HISTORY; WORKING is not valid. JSON keys serialize via `MemoryType.getValue()`: `"sessions"`, `"long-term"`, `"history"`.

`RetentionRule` fields:
- `retention_days` — nullable Integer, must be positive; null disables it; not allowed for HISTORY
- `max_count` — nullable Integer, must be positive; null disables it

Both null means no deletion for that type. When both are set, **OR logic** applies — a document is deleted if either constraint is violated.

`RetentionRule` also carries parse-time flags recording whether each field was present in the JSON request, enabling field-level merge on update (serialization details in the Appendix).

Index mapping changes: a `pinned` boolean field is added to the sessions, long-term, and history mappings (schema_version 1 → 2). The `retention_policy` field is stored on the container document. Working memory indices are not modified.

### 4.2 Defaults (Opt-In, Admin-Configured)

**Retention is opt-in and disabled by default.** Out of the box, the `retention_enabled` master switch is `false`, no retention policy exists on any container, and no data is ever deleted automatically. Retention enforcement only occurs when an admin has set `retention_enabled=true` **and** one of the following holds:
1. A user explicitly sets a `retention_policy` on their container, OR
2. A cluster administrator explicitly configures default retention settings (which are then applied to containers that have no policy of their own).

> **`[Updated]` Auto-applied defaults are double-gated.** Neither the create-time application nor the job backfill runs unless **both** conditions hold: (a) `retention_enabled=true`, and (b) at least one of the four `default_*` settings is > 0. With the shipped defaults (`retention_enabled=false` and all four `default_*` = `-1`), no defaults are ever written to any container. Enabling `retention_enabled` alone is not sufficient — an admin must also set a `default_*` value for defaults to be applied.

**All default settings ship as `-1` (disabled):**

| Setting | Default | Effect when > 0 |
|---|:---:|---|
| `default_session_retention_days` | -1 (off) | Applied as sessions `retention_days` on containers without a policy |
| `default_session_max_count` | -1 (off) | Applied as sessions `max_count` on containers without a policy |
| `default_long_term_max_count` | -1 (off) | Applied as long-term `max_count` on containers without a policy |
| `default_history_max_count` | -1 (off) | Applied as history `max_count` on containers without a policy |

If none of these settings are set to a value greater than zero, containers without an explicit policy are simply skipped by the retention job. No backfill occurs. No data is touched.

**When an admin enables the feature and configures defaults**, auto-application happens in two places (both require `retention_enabled=true`, and both are skipped when multi-tenancy is enabled — see §4.4):
1. **At container creation** — `TransportCreateMemoryContainerAction` writes the admin-configured defaults onto the container document if the user did not provide an explicit `retention_policy` (guarded by `isMemoryRetentionEnabled() && anyConfigured`).
2. **At job execution time (backfill)** — pre-existing containers with no `retention_policy` get admin-configured defaults written onto the container document before processing. The whole job `run()` early-returns when `retention_enabled=false`, so backfill never executes while the feature is disabled.

The backfill only executes if any of the four default settings are greater than zero. If all four are `-1` (or `retention_enabled=false`), no backfill occurs.

> **`[Updated]` "Backfill" here means retention *policy* backfill only — writing admin-configured defaults onto policyless containers. The system deliberately does *not* backfill session *records* for pre-existing working memory: on old data it cannot distinguish "a session was never created" from "the user deleted the session," so manufacturing session docs could resurrect intentionally-deleted state. Instead, the orphan sweep's first-observation grace period (§4.7) gives pre-existing working memory one full `orphan_ttl_days` window to gain a backing session (via the lazy creation in §4.6) before it can be swept.

**Null opt-out:** Setting `"retention_policy": null` explicitly opts a container out of ALL retention. The opt-out is persisted as a literal `"retention_policy": null` field on the container document — distinct from the field being entirely absent, which means "never configured." Because the marker is persisted and round-trips through the container index, the backfill will NOT override an explicit opt-out even if admin defaults are configured. Providing a concrete policy later clears the opt-out (opt back in). Per-field `null` removes/disables that specific constraint.

> Note: opting out of retention disables `retention_days` and `max_count` enforcement only. The orphan sweep and the `disable_session: true` working-memory TTL are separate safety mechanisms that operate independently of the container's `retention_policy`.

**Update semantics** are per-field merge, not per-type replace — omitted types and omitted fields keep current values; explicit `null` removes a field. PUT responses embed the resulting post-merge policy (the update request fetches the `configuration.retention_policy` source, returned in the response's `get` block) so the caller can verify final state without a follow-up GET.

**`[Updated]` Combined strategy + retention updates preserve the policy.** When a single update request modifies `strategies` *and* `retention_policy` together, the update path takes a strategy-rebuild branch that reconstructs the configuration from the merged strategies. That rebuild now explicitly carries the incoming `retention_policy` forward (`finalUpdateConfig.retentionPolicy(...)`), and copies the transient `retentionPolicyExplicitlyNull` opt-out marker separately (it lives outside the builder). Without this, the earlier code silently dropped the retention change whenever strategies were updated in the same request. An explicit `"retention_policy": null` wipe also survives the strategy branch.

**`disable_session: true` containers:** a session policy is stored but has no effect (the job finds zero sessions). The API returns a warning when a session policy is set on such a container. Working memory in these containers is governed by the cluster-level `working_memory_ttl_days` setting instead.

### 4.3 Pinned Memories

A `pinned: true` boolean on session, long-term, and history documents (working memory can't be pinned — pin the parent session instead). The job skips pinned docs; pinned docs are excluded from `max_count` math; if pinned count exceeds `max_count`, the container intentionally grows past the limit and the job logs a WARN to alert operators.

**`[Updated]` `pinned` is gated behind `retention_enabled`.** Because pinning is dead metadata when the retention job is not running, the update-memory path rejects any request whose body contains the `pinned` field while `retention_enabled` is `false`, returning **HTTP 403** (`ML_COMMONS_MEMORY_PINNED_DISABLED_MESSAGE`, which points operators at the cluster setting to enable). The gate keys off the *presence* of the `pinned` key — `pinned: false` is rejected the same as `pinned: true` — and is evaluated before the working-memory-type check. (The add-memory path separately rejects `pinned` on working memory with a 400 regardless of the retention setting; see §5.)

### 4.4 Job Execution Flow

0. **`[Updated]` Feature gate (before the lock):** `run()` first returns immediately if multi-tenancy is enabled, and then if `retention_enabled=false` (logging that the job is disabled via the cluster setting). No lock is acquired and nothing is processed while the feature is off.
1. Acquire a distributed lock via LockService. All logic must be idempotent under concurrent execution (see the Appendix for the lock-release caveat).
2. For each memory container (with or without an explicit policy), resolve per-container index names via `MemoryConfiguration.getIndexName(MemoryType)` for sessions, long-term, history, and working indices. **Critical:** every query must include a `memory_container_id` term filter, because multiple containers can share one physical index.
3. **Backfill (only when the feature is enabled and admin defaults are configured):** the entire job `run()` first early-returns if `retention_enabled=false` (step 0, below), so backfill only runs when the feature is on. Then, if the container has no policy and has not explicitly opted out, AND any admin default setting is > 0, write admin-configured defaults onto the container document before processing. If all default settings are `-1`, this step is skipped entirely. The backfill uses conditional write semantics (skips if a policy already exists), preventing races with concurrent user policy updates.
4. **Identify expired sessions:** if `retention_days` set, search for non-pinned sessions with `last_updated_time` older than the cutoff; if `max_count` set, count non-pinned sessions, and if over the cap, identify the least-recently-updated by **`last_updated_time`** beyond the cap (`[Updated]` — LRU, was `created_time`), appending IDs (deduplicated). **`[Updated]`** ID collection is capped at `COUNT_BASED_ID_CAP` = **50,000** per run: only the oldest 50,000 excess IDs are gathered on any single run, bounding heap; a larger backlog converges over subsequent runs (the ASC sort by `last_updated_time`, tie-broken by `_id`, is deterministic so each run resumes from the oldest remaining docs). The job detects legacy epoch-second timestamps (values < 10 billion) and normalizes them to milliseconds before comparison, preventing false expiration of pre-upgrade data.
5. **Cascade — WM first:** `_delete_by_query` on the working index filtered by container ID and session ID match. The query uses a `should` with `minimumShouldMatch(1)` matching both the top-level `session_id` keyword field (for new documents) and the legacy `namespace.session_id` flat_object path (for pre-upgrade documents). Expired session IDs are batched into groups of 500 per request to stay within the max clause count (default 1024).
6. Bulk-delete the session documents.
7. **Long-term:** `_delete_by_query` for non-pinned docs with `last_updated_time` past the `retention_days` cutoff; for `max_count`, search the least-recently-updated by **`last_updated_time`** beyond cap (`[Updated]` — LRU, was `created_time`; same 50,000-per-run cap as sessions) and bulk delete.
8. **History:** `max_count` only, sorted by `created_time` (append-only FIFO; no `last_updated_time` exists on history). Same 50,000-per-run ID cap.
9. **Working-memory TTL:** `[Updated]` for `disable_session: true` containers *only*, `_delete_by_query` WM where `created_time` is older than `working_memory_ttl_days` — **but this runs only when `working_memory_ttl_days > 0`; it defaults to `-1` (disabled), so sessionless working memory is kept indefinitely unless an admin sets a positive TTL** (see §4.8).
10. **Conditional throttle:** pause `{throttle}` seconds between containers only when deletions occurred in the current container. No-op containers (opted out, no expired data, backfill-only) proceed immediately to the next container, reducing total job runtime from O(containers × throttle) to O(active-containers × throttle).
11. **Orphan sweep** (once per run, containers with sessions enabled): before running the orphan sweep for a container, the job verifies that both the sessions index and working memory index exist. If either is missing, the sweep is skipped for that container (prevents misclassifying all WM as orphaned when sessions simply haven't been created yet). **`[Updated]` First-observation grace period:** the first time the sweep observes a container it stamps a per-container `orphan_sweep_baseline_time` (via a write-once conditional painless script, `retryOnConflict(3)`) and deletes nothing that run; on subsequent runs it defers all orphan deletion until `orphan_sweep_baseline_time + orphan_ttl_days` has elapsed. This gives pre-existing working memory — which predates retention or was written by a path that created no session doc — one full `orphan_ttl_days` window to acquire a backing session before it can be classified as orphaned. The window only ever delays the first deletion; it is never reset. The sweep uses a **paged composite aggregation** on the top-level `session_id` keyword field in the working-memory index mapping to enumerate distinct session IDs efficiently (cost proportional to unique sessions, not total documents). A separate legacy fallback enumerates pre-upgrade documents that lack the top-level field via `search_after` pagination on `namespace.session_id`. Both paths start from a **per-run random keyspace position** and wrap around, so the 50,000 session-ID cap truncates a different slice on each run (probabilistic full coverage). Check session existence via batched `termsQuery("_id", batch)` against the sessions index (1000 per request), then `_delete_by_query` WM whose session no longer exists (subject to a separate per-document `created_time` cutoff of `now − orphan_ttl_days`). Also delete WM lacking a `session_id` entirely, older than `orphan_ttl_days`. Note that live working memory written under a client-supplied `session_id` is protected because the add path now lazily creates a backing session document for it (see §4.6).

**`[Updated]` Eviction sort field:** `max_count` eviction sorts by **`last_updated_time` (LRU) for sessions and long-term**, and by **`created_time` (FIFO) for history** (history has no `last_updated_time`). Rationale: for sessions and long-term, evicting the least-recently-updated keeps actively-used memory and, together with session auto-touch (§4.6), means conversation activity protects a session from count-based eviction. History is append-only, so `created_time` is the natural insertion order. The `disable_session` working-memory TTL (§4.8) is a hard lifespan cap and ages by `created_time`. Time-based `retention_days` also uses `last_updated_time` (sessions and long-term).

**Client usage:** native OpenSearch Client via `client.execute()` with `ThreadContext.stashContext()`, NOT `SdkClient` (which lacks `_delete_by_query` and search-based pagination patterns needed here).

**Multi-tenancy:** the retention job **runs under `plugins.ml_commons.multi_tenancy_enabled=true`** for self-hosted (local-metadata) clusters, tenant-isolated. The job stays a single cluster-wide, system-context janitor: it scans the one global container registry (`matchAllQuery`) and processes **every container regardless of tenant**. Tenant isolation is achieved **per-container, not per-run** — every per-container search/delete is already bounded by `termQuery(memory_container_id, …)`, and because container `_id`s are globally unique with each container belonging to exactly one tenant, filtering by `memory_container_id` transitively confines every read/delete to that container's own tenant. **No `tenant_id` term filter is added**: it would be redundant on sessions/working/history and outright *wrong* on long-term, whose index mapping (`ml_memory_long_term.json`) has no `tenant_id` field — such a filter would match zero long-term docs and silently disable long-term retention. So `memory_container_id`, not `tenant_id`, is the isolation key.

- One scheduled run cleans **all tenants'** data, each container scoped to its own tenant. There is no per-tenant scheduling.
- Default policies **are** auto-applied and explicit `retention_policy` values **are** enforced under local-metadata multi-tenancy (the previous "stored but not enforced" `HeaderWarning` no longer applies to this mode).

**Remote-metadata exception (AOSS / DynamoDB is out of scope):** when a **remote** metadata store is configured (`plugins.ml_commons.remote_metadata_type` ∈ `RemoteOpenSearch` / `AWSDynamoDB` / `AWSOpenSearchService`), the container registry lives outside the local cluster. The native-client job cannot enumerate it, so it would find zero containers and silently clean nothing. In that mode the job **skips with a WARN** (both scheduling in `MLCommonsClusterEventListener` and execution in `run()` gate on `MLCommonsSettings.isRemoteMetadataStore(...)`). Tenant-isolated retention for remote-metadata deployments is left to a future workstream.

**Scalability:** with conditional throttling (only after containers that perform deletions), a run over 10,000 containers where only 50 have expired data takes ~250s of throttle instead of the theoretical maximum. Expected v1 scale is under 1,000 containers. If a run overruns, the next run runs concurrently (safe via idempotent operations). No partial-run resumption — each run processes everything from scratch.

**Non-existent indices:** all searches/`_delete_by_query` use `allowNoIndices=true` and `ignoreUnavailable=true`; missing indices are skipped with a debug log, and the orphan sweep skips containers whose sessions or working memory index doesn't exist.

**max_count overshoot:** between the COUNT query and the BULK DELETE, concurrent writes can push the count above `max_count`. Overshoot is proportional to write throughput during job execution and is corrected on the next run.

### 4.5 Cascade Order Rationale (crash-safety)

- Crash after WM delete, before session delete → next run re-identifies the still-existing sessions, finds no WM, deletes sessions cleanly
- Crash before any deletes → nothing changed
- Reverse order (session first) would be unsafe: session gone means the cascade can never re-run, producing permanent orphans
- Two concurrent job instances → identical expired sets (deterministic ASC sort — `last_updated_time` for sessions/long-term, `created_time` for history — tie-broken by `_id`), overlapping no-op deletes, safe

### 4.6 Session Liveness

> **Changed from the original proposal**, which kept `last_updated_time` static and deferred auto-touch to v2. Team decision (2026-07-10): active conversations should extend session lifetime; the write amplification is acceptable because the session index is small and the alternative (premature eviction of active conversations) is worse.

**Auto-touch on message add (now in v1):** `TransportAddMemoriesAction` bumps the parent session's `last_updated_time` when working memory is added to an existing session. Details:
- Uses `retryOnConflict(3)` to handle concurrent message writes to the same session
- Skipped for newly-created sessions (already fresh) and `disable_session: true` containers
- Best-effort and asynchronous — fires after the add-memory response returns; a failed bump is logged at DEBUG and never fails the request

Note: `[Updated]` auto-touch extends both time-based retention (`retention_days`) and count-based eviction (`max_count`), because `max_count` for sessions now sorts by `last_updated_time` (LRU). Conversation activity therefore protects an active session from *both* forms of eviction; pinning remains the explicit hard exemption.

**`[Updated]` Lazy session creation (backing doc for client-supplied `session_id`):** when a client adds working memory under its own `session_id` (the namespace carries a `session_id`) without having called create-session, and sessions are enabled (`disable_session: false`), `TransportAddMemoriesAction` first upserts a minimal backing session document keyed by that `session_id` using `IndexRequest.opType(OpType.CREATE)`. This is idempotent — if the session already exists the `VersionConflictEngineException` is treated as an expected no-op. The upsert is **chained before** the working-memory write (not fire-and-forget): a genuine failure to create the backing session aborts the add with HTTP 500 so no orphan-eligible working memory is left behind. The backing doc stamps `created_time`/`last_updated_time` so session-level retention still applies. Purpose: without a session document, the orphan sweep cannot distinguish "session never created" from "session deleted" and would eventually evict this live working memory; the backing doc makes the session discoverable. (Sessionless containers and the auto-summarize create-session branch do not take this path.)

Edge case: if a message is written to a session after the job identifies that session for eviction, the auto-touch fires but the session is still deleted in the same job run. This window is extremely narrow (sub-second) and the orphan sweep provides a safety net for any resulting working memory.

**Conditional bump on update (new in v1):** `TransportUpdateMemoryAction` only bumps `last_updated_time` when a **content** field changes. Organizational/metadata changes do not extend retention lifetime:

| Memory Type | Bumps `last_updated_time` | Does NOT bump |
|---|---|---|
| Sessions | `summary` | `pinned`, `metadata`, `additional_info`, `agents` |
| Long-term | `memory` | `tags`, `pinned` |
| Working | `messages`, `binary_data`, `structured_data`, `structured_data_blob` | `metadata`, `tags` |

This prevents pinning or tagging a memory from silently resetting its retention clock.

Session timestamps are stored as epoch milliseconds, consistent with the job's `System.currentTimeMillis()` comparisons (see the Appendix for the precision fix history).

History note: `MLMemoryHistory` records are immutable audit entries with only `created_time`, which is used for count-based sort ordering.

### 4.7 Orphan Sweep Details

A safety net for WM pointing at missing sessions (bugs, crashes, partial writes). **`[Updated]` Per-container first-observation grace:** the sweep stamps `orphan_sweep_baseline_time` the first time it sees a container and defers *all* orphan deletion for that container until one full `orphan_ttl_days` window has passed since that baseline (see §4.4 step 11) — this protects pre-existing/pre-upgrade working memory from being wiped on the first run. The working-memory index mapping includes a top-level `session_id` keyword field (denormalized from the `namespace` flat_object on write) that enables efficient composite aggregation. The sweep uses a paged `CompositeAggregationBuilder` on this field to enumerate distinct session IDs at O(unique sessions) cost. A legacy fallback path covers pre-upgrade documents lacking the top-level field via `search_after` pagination on `namespace.session_id`. Enumeration starts from a per-run random keyspace position and wraps around, capped at 50,000 session IDs — each run's cap truncates a different keyspace slice for probabilistic full coverage. Batched `termsQuery("_id", ...)` existence checks (1000/request), container-ID-filtered deletes, plus a separate pass deleting WM with no `session_id` older than `orphan_ttl_days` (grace period for in-flight writes).

The sweep processes up to 50,000 distinct session IDs per container per run. Due to randomized start-key enumeration, containers exceeding this threshold achieve full coverage probabilistically across multiple runs.

### 4.8 disable_session: true Containers

No session exists to cascade from, so WM is cleaned by an age-based TTL on `created_time` (a hard lifespan cap, not LRU) — the cluster setting `working_memory_ttl_days`.

**`[Updated]` The TTL defaults to `-1` (disabled).** By default, sessionless working memory is kept indefinitely: the job's working-memory TTL pass short-circuits whenever `working_memory_ttl_days <= 0`, so no deletion occurs. An admin must set a positive value (1–365 days) to age it out. `-1` is the "off" sentinel (range −1 to 365). The setting has no per-container override and applies uniformly to all `disable_session: true` containers. (This replaces the earlier design, which defaulted to 30 days and could not be disabled.)

## 5. API Changes

**Modified endpoints:**
- `POST /_plugins/_ml/memory_containers/_create` — accepts optional `retention_policy`; no policy means no retention unless admin defaults are configured; `null` explicitly opts out (blocks even admin defaults)
- `PUT /_plugins/_ml/memory_containers/{id}` — accepts policy updates with field-level merge semantics
- `GET /_plugins/_ml/memory_containers/{id}` — returns the policy
- `POST/PUT .../memories/{type}/{id}` — accepts `pinned` (sessions, long-term, history only)

**Validation rules (400 errors unless noted):**
- `"working"` key in the policy → error explaining WM is deleted when its parent session expires; configure retention on sessions instead
- `retention_days` on history → 400
- `retention_days` or `max_count` ≤ 0 → 400 (zero is rejected — use `null` to disable)
- Both fields null for a type → valid (no deletion)
- Note: valid keys are `"sessions"`, `"long-term"`, `"history"`; the rejected key is `"working"` (not `"working_memory"`)
- **`[Updated]` `retention_policy` supplied while `retention_enabled` is `false`** → **HTTP 403** on create and update (the feature is off; enable it via the cluster setting first)
- **`[Updated]` `pinned` field present in a memory-update request while `retention_enabled` is `false`** → **HTTP 403** (distinct from the 400s above; see §4.3). The add-memory path still rejects `pinned` on working memory with a 400 independently of the retention setting.

**Security:** No new permissions. Policy management reuses container-update authorization; pinning reuses memory-update auth. The job runs with plugin-level system index access, matching existing ml-commons background operations. Operators should note that container-update authorization implicitly grants the ability to set aggressive retention policies that trigger deletion on the next job run.

**Cluster settings:**

| Setting | Default | Purpose |
|---|---|---|
| `plugins.ml_commons.memory.retention_enabled` | **`false`** `[Updated]` | Master opt-in switch — the feature is **disabled by default**. When `false`, the retention job short-circuits *and* the memory-container APIs reject a supplied `retention_policy` (403) and the memory-update API rejects the `pinned` field (403). An admin must set it to `true` to enable retention. Dynamic — takes effect on the next scheduled run without a restart. |
| `plugins.ml_commons.memory.retention_job_interval_hours` | 24 | Run frequency (integer, hours), configurable 1–168. **In 3.8** a change required job re-registration (node restart) to take effect (the persisted schedule was written once at first startup and never reconciled). **In 3.9 this becomes dynamic** — a `PUT _cluster/settings` change is reconciled onto the persisted job document on the elected cluster-manager and rescheduled without a restart. See §9 (3.9). |
| `plugins.ml_commons.memory.retention_job_throttle_seconds` | 5 | Pause between containers (integer, seconds) |
| `plugins.ml_commons.memory.orphan_ttl_days` | 7 | grace period for WM with no session_id |
| `plugins.ml_commons.memory.working_memory_ttl_days` | **`-1` (off)** `[Updated]` | TTL (by `created_time`) for `disable_session: true` containers. Defaults to `-1` — sessionless working memory is kept indefinitely; only a value > 0 (range 1–365) ages it out. |
| `plugins.ml_commons.memory.default_session_retention_days` | -1 (off) | Admin-configured default `retention_days` for sessions. Only applied when > 0. |
| `plugins.ml_commons.memory.default_session_max_count` | -1 (off) | Admin-configured default `max_count` for sessions. Only applied when > 0. |
| `plugins.ml_commons.memory.default_long_term_max_count` | -1 (off) | Admin-configured default `max_count` for long-term. Only applied when > 0. |
| `plugins.ml_commons.memory.default_history_max_count` | -1 (off) | Admin-configured default `max_count` for history. Only applied when > 0. |

**Index naming:** per-container via `getIndexName(MemoryType)`, pattern `.plugins-ml-am-{indexPrefix}-memory-{type}`; shared prefixes mean shared physical indices, hence the mandatory container-ID filter.

**Observability:** Per-container INFO logs report deleted counts for sessions, cascaded working memory, long-term, history, and orphans. A WARN is logged when pinned count exceeds `max_count` for any type. A completion summary logs containers processed, total deleted, duration, and errors. Errors go to WARN/ERROR. **DEBUG-level logging** provides per-batch document IDs for bulk-delete paths (session deletion, max_count eviction) and session-ID batches for cascade/orphan delete-by-query paths. Criteria-only DBQ paths (long-term TTL, working-memory TTL, empty-namespace cleanup) log the index name and cutoff timestamp at DEBUG. No memory content is logged at any level — only IDs, index names, counts, and criteria. Job registration can be verified by checking for the `MEMORY_RETENTION` document in the ML jobs index.

## 6. Risks & Mitigations

- **Admin-configured defaults apply to containers without an explicit policy** → all default settings ship as `-1` (off), so no automatic deletion occurs without explicit admin action; explicit `null` opt-out on individual containers is always preserved by backfill; `[Updated]` the whole feature is off by default (`retention_enabled=false`) and can also be disabled at any time by flipping it back to `false`; documented in release notes
- **Misconfigured retention_days** → merge semantics prevent accidental constraint drops, conservative-value guidance, log monitoring, kill switch
- **Large cascade (10K+ message sessions)** → `_delete_by_query` batching (500–1000 session IDs per request) plus inter-container throttle
- **Expensive orphan sweep** → `search_after` pagination, 50K session-ID cap, batched existence checks, runs once per job
- **Race: new WM during cascade** → cascade filters by session_id (not doc IDs), so writes concurrent with the delete-by-query are caught; writes after it completes are handled by the orphan sweep on the next run
- **Race: session updated during identification** → bulk delete uses collected IDs; worst case is deleting one borderline-expired session
- **Job overruns interval** → concurrent runs safe via idempotent, deterministic operations (ASC sort — `last_updated_time` for sessions/long-term, `created_time` for history — tie-broken by `_id`, produces identical deletion sets)
- **Shared-index cross-contamination** → mandatory `memory_container_id` filter, never unscoped delete-by-query
- **Write amplification from session auto-touch** → one extra small-index update per message; `retryOnConflict(3)`; async/best-effort so it never adds latency or failures to the add-memory path
- **Continuously-refined long-term memories never expire** → memories repeatedly updated by strategy execution (e.g., facts refined on every conversation) may never expire via `retention_days` since each update resets the clock. Use `max_count` as a hard cap for workloads with frequent memory updates
- **Heap pressure on first run** → `[Updated]` bounded by design: count-based ID collection is capped at `COUNT_BASED_ID_CAP` = 50,000 IDs per run (mirrored by the orphan sweep's 50,000 session-ID cap), so a container with a very large expired backlog evicts at most 50,000 per type per run and converges over subsequent runs rather than accumulating an unbounded ID set in heap
- **Typo in max_count (e.g., 1 vs 1000)** → unmitigated in v1; deletions irreversible; snapshot recommended before tightening a policy; the v2 `_retention_preview` endpoint will address this

## 7. Second Iteration (v2 — deferred, no redesign needed)

- **`_retention_preview` endpoint**: dry-run counts of what would be deleted — **`[3.9]` delivered in §9 as `POST .../memory_containers/{id}/_retention/_dry_run`** (cluster-wide variant added).
- **`_purge` endpoint**: immediate single-container enforcement — **`[3.9]` delivered in §9 as the cluster-wide `POST .../memory_containers/_retention/_execute`** (on-demand trigger of the real job).
- **`"default"` keyword**: pass the string `"default"` at any level of the policy to reset that level to current cluster defaults (design complete; parser/resolution deferred)
- **Stats API metrics**: `job_last_success_timestamp`, `total_deleted`, `containers_processed`, `job_duration_ms` (follows `MLStatsJobProcessor` pattern)
- **Cluster-wide retention ceiling**: admin-enforced `max_retention_days_ceiling` / `max_count_ceiling` overrides
- **Policy change audit trail**: history entries on policy create/update/remove for compliance (SOX, HIPAA)
- **Deletion audit trail** — record deleted document IDs and counts as history entries for compliance (SOX, HIPAA) and operational forensics
- **Dedicated permission**: separate authorization for retention policy changes
- **Pinned count visibility**: `pinned_count` per type in GET container responses
- **`max_pinned_per_container` setting**: reject or warn when pinned count exceeds a threshold (v1 provides the WARN log only)

## 8. Design Decisions (Resolved)

The following questions were raised during the RFC process and resolved during implementation:

1. **Default values** — Resolved: there are no built-in defaults. All `default_*` settings ship as `-1` (disabled). Admins choose their own values based on their organization's needs. No data is deleted without explicit configuration.
2. **Opt-in vs opt-out posture** — Resolved: retention is opt-in. Nothing happens unless a user sets a policy on their container or an admin explicitly configures cluster-level defaults. Pre-existing containers are only affected if an admin takes action.
3. **OR semantics** — Resolved: OR logic applies. When both `retention_days` and `max_count` are set, a memory is deleted if either is violated. AND semantics deferred to v2 if needed.
4. **Count-based eviction and activity** — `[Updated]` Resolved: `max_count` evicts by `last_updated_time` (LRU) for sessions and long-term, and by `created_time` (FIFO) for history. Combined with session auto-touch (§4.6), conversation activity keeps an active session out of the eviction set; pinning remains the explicit hard exemption. (This finalizes the earlier "FIFO for all types" position; activity-aware eviction is now the shipped behavior for sessions/long-term, not deferred.)
5. **Multi-tenancy** — Resolved: the job **runs tenant-isolated under `multi_tenancy_enabled=true`** on self-hosted (local-metadata) clusters, using the per-container `memory_container_id` filter as the isolation key (no `tenant_id` term). It **skips with a WARN** only when a remote metadata store (AOSS / DynamoDB) is configured, since the container registry then lives outside the local cluster. See the **Multi-tenancy** subsection under §4.4.
6. **Job cadence** — Resolved: 24h default interval, configurable from 1h to 168h. The inter-container throttle (5s default) only applies between containers that had deletions, keeping total runtime proportional to active containers.
7. **Pinned overflow** — Resolved: pinned documents are never deleted. When pinned count exceeds `max_count`, the container grows past the limit and the job logs a WARN. A hard cap (`max_pinned_per_container`) is deferred to v2.
8. **v2 priorities** — Resolved: `_retention_preview` (dry-run) and `_purge` (on-demand enforcement) are both planned for 3.9, alongside retention receipts for auditability.

## Access Control for Retention Policy (Decision)

**Decision:** Retention policy is a **container-level setting**. Anyone who can access a container can set or edit its retention policy, and that policy applies to **every** memory in that container. There is **no per-user isolation** of retention. Retention stays on the **same access model as every other container config field** — no special gate is invented for it.

**Where the policy lives.** The retention policy is a field on the container's configuration — `MemoryConfiguration.retentionPolicy` (`common/.../memorycontainer/MemoryConfiguration.java:84`, a `Map<MemoryType, RetentionRule>`) — sitting next to the other container settings such as strategy and embedding model. It is not a separate object with its own permission surface; it is part of the container document.

**How edits are gated.** The policy is set and edited exclusively through the normal create/update-container API:
- **Update** (`PUT /_plugins/_ml/memory_containers/{id}`) is access-gated in `TransportUpdateMemoryContainerAction.doExecute()` (`plugin/.../memorycontainer/TransportUpdateMemoryContainerAction.java:131`), which calls `memoryContainerHelper.checkMemoryContainerAccess(user, container)` and returns HTTP 403 if it fails. This is the retention-policy edit path, so retention edits are already access-gated.
- **Create** (`POST /_plugins/_ml/memory_containers/_create`) does not call `checkMemoryContainerAccess` — creation instead *establishes* ownership by stamping the caller as owner (`.owner(user)` from `RestActionUtils.getUserContext(client)`, `TransportCreateMemoryContainerAction.java:216`) along with the request's backend roles. The access model is thus defined at create time and enforced on every subsequent access.
- `checkMemoryContainerAccess` (`plugin/.../helper/MemoryContainerHelper.java:195`) grants access when: security is disabled (`user == null`), the user has the `all_access` (admin) role, the user is the container owner, or the user shares one of the container's allowed/owner backend roles. This is the exact same gate used by every other container-scoped operation (get, delete, update-memory, add-memory, search-memory, create-session).

Because retention reuses this gate, **no new permission** is introduced (consistent with §5 "Security").

**Consistency with `max_count`.** This mirrors the already-settled decision that `max_count` is a **shared, container-wide cap, not a per-user quota** (§2 Non-goals lists "per-namespace or per-user `max_count`" as out of scope; the shared-cap rationale is detailed in DesignDoc.md §6.9). Retention, like `max_count`, is a property of the container, evaluated container-wide.

**Accepted caveat.** In a **shared** container (multiple users behind shared backend roles), any accessor who can edit the policy can tighten it and thereby trigger deletion of **other users'** memories on the next job run. This is an accepted trade-off, not a bug. The escape hatch for callers who need strict isolation is the same one offered for `max_count`: **use a separate container per user**.

**Non-goals (explicitly out of scope for this decision):**
- **Per-user / per-namespace retention** — deferred. Enforcing retention per namespace requires the expensive `flat_object` enumeration of distinct namespace values (the same costly scroll+dedup pattern the orphan sweep uses); not undertaken in v1.
- **Admin-only retention** — explicitly **not** doing this. Users set their own container's retention; retention is not restricted to cluster admins.
- **`tenant_id` / multi-tenancy retention** — a separate workstream, now delivered for self-hosted (local-metadata) clusters: the job runs tenant-isolated under `plugins.ml_commons.multi_tenancy_enabled=true`, with `memory_container_id` (not `tenant_id`) as the isolation key (see the **Multi-tenancy** subsection under §4.4). This is orthogonal to the access-control decision above — *who may edit a policy* (container access) is unchanged. Remote-metadata (AOSS / DynamoDB) multi-tenant retention remains out of scope: the job skips with a WARN when `plugins.ml_commons.remote_metadata_type` names a remote store.

### Verification

Confirmed against the read-only source tree at `/Users/ballewer/OpenSearchMLLocal/ml-commons-livetest`:
- `MemoryConfiguration.retentionPolicy` field — `common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java:84` (plus the persisted opt-out marker `retentionPolicyExplicitlyNull` at :90).
- `checkMemoryContainerAccess(User, MLMemoryContainer)` defined at `plugin/src/main/java/org/opensearch/ml/helper/MemoryContainerHelper.java:195` (owner / `all_access` admin / shared backend-role logic).
- Called from the retention-edit path `TransportUpdateMemoryContainerAction.java:131`.
- Create path does not call the gate; it sets `.owner(user)` at `TransportCreateMemoryContainerAction.java:216`, establishing the access model at creation.
- Multi-tenancy: the job now **runs** under `multi_tenancy_enabled=true` (local metadata); the only remaining `run()` skip is the remote-metadata gate `if (MLCommonsSettings.isRemoteMetadataStore(clusterService.getSettings())) { ...warn...; return; }`, mirrored by the scheduling gate in `MLCommonsClusterEventListener`. Isolation is via the per-container `memory_container_id` filter. (`isRemoteMetadataStore` is defined in `common/.../settings/MLCommonsSettings.java`.)

## 9. Third Iteration (3.9)

`[3.9]` The 3.8 sections above describe the core system. 3.9 adds four capabilities on top of it. None of them change the core deletion semantics — they add configurability, preview, on-demand triggering, and broader deployment support to the existing job.

1. **Dynamic retention job interval.** In 3.8 the persisted schedule doc was written once at first startup and never reconciled, so `retention_job_interval_hours` changes required a node restart (see §5 cluster-settings table). 3.9 fixes this write-once behavior: on startup and on a `PUT _cluster/settings` change, the effective interval (transient > persistent > yaml > default) is reconciled onto the persisted job document and rescheduled live, gated on the elected cluster-manager (single-writer) and behind the same rolling-upgrade guard as job creation (deferred until the jobs index exists or all nodes are on the current version, so a mixed-version cluster is never left yellow).

2. **Dry-run preview API.** `POST /_plugins/_ml/memory_containers/{id}/_retention/_dry_run` (single container) and `POST /_plugins/_ml/memory_containers/_retention/_dry_run` (cluster-wide, returns an object `{ "skipped_count", "truncated", "warning", "results": [...] }`). Returns exactly what the job **would** delete — per-type `would_delete` with `by_reason` breakdowns, `total_would_delete`, `pinned_would_skip`, `effective_policy`, `policy_source`, and `warnings` — performing **zero** deletions. It reuses the job's own `identify*`/count logic (shared `computeDefaultRetentionPolicy()`) so the preview cannot diverge from the real run's time-based decisions; the long-term `max_count` attribution is a close single-run approximation. Cluster-wide traversal is bounded at `MAX_CONTAINERS_PER_DRY_RUN` (1000) with an explicit truncation warning, and containers that fail evaluation are surfaced via `skipped_count` rather than silently dropped. Supersedes the v2 `_retention_preview` name in §7.

3. **On-demand execution API.** `POST /_plugins/_ml/memory_containers/_retention/_execute` triggers the real retention job immediately (cluster-wide) rather than waiting for the interval. It reuses the scheduler's singleton path, returns a status (`triggered` / `already_running` / `retention_disabled` / `remote_metadata_store`), and acknowledges asynchronously without blocking on the delete pipeline. **Admin-only** (`all_access`): since it initiates a cluster-wide deletion run, a lower-privilege caller cannot use it to force early deletion of data they could not otherwise reach. Supersedes the v2 `_purge` name in §7. **Job-guard hardening** ships alongside it: the cluster-wide JobScheduler lock is held across the entire async pipeline (not just kickoff) with periodic renewal, and teardown is fail-safe — a run releases the cluster lock only on its own terminal/error paths, while the per-node in-progress guard is reclaimed independently, so a stuck-guard watchdog can recover a wedged run without ever disturbing a live run's lock. If lock renewal is lost mid-run, the run stops claiming cluster-wide exclusion rather than deleting a lock another node may hold.

4. **Retention under self-hosted multi-tenancy.** The 3.8 job skipped entirely when `multi_tenancy_enabled=true`. 3.9 makes it **run** on self-hosted (local-metadata) clusters, using the per-container `memory_container_id` filter as the isolation key (globally unique, one tenant per container — no `tenant_id` term filter, which would be redundant on sessions/working/history and unsupported on long-term). The job now skips only when a **remote metadata store** (`remote_metadata_type` = AOSS / DynamoDB) is configured, since the container registry then lives outside the local cluster and the native-client scan cannot enumerate it. This finalizes Design Decision §8.5.

**Not in 3.9 (still deferred):** retention receipts / deletion audit trail (§7), `"default"` policy keyword, stats-API metrics, cluster-wide retention ceilings, dedicated retention permission, and per-object receipt lookup. An `SdkClient` migration of the dry-run container reads (for remote-metadata parity) is a possible follow-up but is intentionally out of scope — retention targets self-hosted deployments and the remote-metadata gate handles managed clusters by cleanly refusing to act.

## Related

References RFC #4859 (memory retention lifecycle); this RFC documents the complete implementation as shipped in 3.8.0, plus the in-progress 3.9 additions marked `[3.9]` in §9. `[Updated]` The 3.8 work was delivered as **two PRs** — **#4914** (retention policy data model, `pinned` field, and API wiring) and **#4918** (retention job, orphan sweep, and hardening) — consolidating the originally-proposed 7-PR stack.

## Appendix: Implementation Notes

**Transient parse flags on `RetentionRule`:** the per-field flags (`retentionDaysExplicitlySet`, `maxCountExplicitlySet`) record whether each field was present in the JSON request. They enable field-level merge on update and are not serialized over transport — the merge always runs on the node that parsed the request (`TransportUpdateMemoryContainerAction.doExecute()`), so this is safe. The container-level opt-out marker (`retentionPolicyExplicitlyNull`) is different: it IS persisted (serialized as `"retention_policy": null` on the container document) and transported, so it survives both index round-trips and node-to-node update requests. Providing a concrete policy clears the marker.

**`[Updated]` Strategy + retention combined-update fix:** in `TransportUpdateMemoryContainerAction.doExecute()`, when the request carries a non-empty `strategies` list the code enters a rebuild branch that constructs `finalUpdateConfig` from the merged strategies. That rebuild explicitly copies the incoming `retentionPolicy` (and re-copies the transient `retentionPolicyExplicitlyNull` marker, which is outside the `@Builder`) so a simultaneous retention change is not silently dropped; `currentConfig.update(finalUpdateConfig)` then performs the field-level retention merge into the stored policy. A regression test asserts `retention_policy` is not dropped on a combined strategy+retention update.

**Lock-release caveat:** the distributed lock acquired via LockService releases when `run()` returns, **before** async work completes. All job logic must therefore be idempotent under concurrent execution — this is why deletion sets are deterministic (ASC sort by the type's evaluation field — `last_updated_time` for sessions/long-term, `created_time` for history — tie-broken by `_id`) and all deletes are no-op-safe on re-execution.

**Epoch-seconds fix history:** session `created_time`/`last_updated_time` previously used epoch seconds (`getEpochSecond()`); this was changed to epoch milliseconds (`toEpochMilli()`) for consistency with the job's `System.currentTimeMillis()` comparisons. Mixed precision would make sessions appear to be from 1970 and cause immediate eviction. As a defense-in-depth measure for pre-upgrade documents, the job additionally detects legacy epoch-second timestamps (values < 10 billion) and normalizes them to milliseconds before comparison.

### 3.8 — Core Retention System

The complete retention lifecycle engine — explicit opt-in per container, scheduled deletion, cascade semantics.

- Retention policy model per container (retention_days, max_count per memory type)
- Opt-in semantics: no deletion without explicit policy configuration (user-set or admin-configured defaults, all shipping as disabled)
- Explicit opt-out via `"retention_policy": null`, admin-configurable cluster defaults (all `-1` by default), and `retention_enabled` kill switch
- Scheduled job with throttle, single-flight execution
- Session, long-term, and history retention (time-based + count-based)
- Working memory TTL for session-less containers
- Cascade delete (expired sessions → their working memory)
- Orphan sweep (working memory whose parent session no longer exists)
- Pinning exemption from all retention rules
- Legacy epoch-second timestamp detection and normalization

### 3.9 — Dynamic Interval, Dry-Run, On-Demand Execution, Multi-Tenancy

`[3.9]` Four additive capabilities, described in full in **§9 (Third Iteration)**. In brief:

- **Dynamic retention job interval** — `retention_job_interval_hours` changes take effect live (no restart), reconciled on the elected cluster-manager behind a rolling-upgrade guard.
- **Dry-Run preview API** — `POST .../memory_containers/{id}/_retention/_dry_run` (and cluster-wide) returns what the job would delete, zero mutations, reusing the job's own counting logic.
- **On-demand execution API** — `POST .../memory_containers/_retention/_execute` triggers the real job immediately; ships with pipeline-lock + guard-leak hardening.
- **Retention under self-hosted multi-tenancy** — the job runs (isolated by `memory_container_id`) instead of skipping; skips only for remote metadata stores.

**Deferred beyond 3.9:** retention receipts / deletion audit trail, `"default"` policy keyword, stats-API metrics, retention ceilings, dedicated permission (see §7 and §9 "Not in 3.9").

AirFunnyB:
### Description

Implements the agentic memory retention lifecycle (RFC #4859). Adds a background Job Scheduler job that enforces per-container lifecycle rules (`retention_days`, `max_count`) on sessions, long-term memory, and history — preventing unbounded storage growth and context window pollution.

**Core features:**

- `retention_policy` configuration on memory containers with per-memory-type rules
- Background retention job (default 24h interval) with FIFO eviction of expired/over-count memories
- Session deletion cascades to working memory (working memory first → session — prevents orphans)
- Orphan sweep with 7-day age guard catches working memory pointing to non-existent sessions
- `pinned: true` field exempts individual memories from eviction
- Cluster-level admin defaults auto-apply to containers without a policy (conditional script update, never clobbers user-set policies or explicit-null opt-outs)
- Working memory TTL for session-less containers (`disableSession=true`)
- Epoch-second detection prevents mass-eviction of pre-upgrade legacy data
- Backward compatible: VERSION_3_8_0 guards on all new serialized fields

**Also fixes:** A latent bug in `SearchModelGroupTransportAction.search()` where an empty indices array caused all-indices search — exposed by the new retention job document in `.plugins-ml-jobs`.

### Related Issues

Resolves #4859

### Check List

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

### DIFF
--- a/AUDIT_FIXES.md
+++ b/AUDIT_FIXES.md
@@ -1,0 +1,385 @@
+# Memory Retention RFC — Audit Fixes
+
+These are the confirmed issues from a 20-agent adversarial audit of the
+`feature/memory-retention/rfc-implementation` branch that should be fixed
+before ship.
+
+---
+
+## 1. RetentionRule explicit-null field removal doesn't persist
+
+**Problem:**
+
+When a user sends an update to remove a retention limit (e.g.,
+`{"retention_policy": {"sessions": {"max_count": null}}}`), the in-memory
+merge correctly sets `max_count` to null. But `RetentionRule.toXContent()`
+only emits non-null fields, so the serialized update doc omits `max_count`
+entirely. OpenSearch's partial-update merge then preserves the stored value,
+and the retention job keeps enforcing a cap the user explicitly removed.
+
+**Solution:**
+
+In `RetentionRule.toXContent()`, emit explicit null fields when the field was
+explicitly cleared. We need to propagate the `retentionDaysExplicitlySet` and
+`maxCountExplicitlySet` flags through the merge (or alternatively, always
+replace the entire `retention_policy` object in the update doc rather than
+relying on recursive partial merge).
+
+The simplest fix: after the merge in `MemoryConfiguration.update()`, when
+constructing the merged RetentionRule, preserve the "explicitly set" flags so
+`toXContent()` can emit `builder.nullField(...)` for cleared fields. The
+pattern already exists in the same file for the whole-policy opt-out:
+
+```java
+} else if (retentionPolicyExplicitlyNull) {
+    builder.nullField(RETENTION_POLICY_FIELD);
+}
+```
+
+Apply the same pattern per-field inside `RetentionRule.toXContent()`.
+
+**Files to modify:**
+- `common/src/main/java/org/opensearch/ml/common/memorycontainer/RetentionRule.java`
+  - Add logic to `toXContent()` to emit `nullField` when explicitly cleared
+  - May need to carry the explicitlySet flags through the merge
+- `common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java`
+  - In the retention policy merge (around line 685), ensure the constructed
+    merged RetentionRule preserves explicit-set metadata
+
+---
+
+## 2. ModelGuardrail timeout: set back to 5 seconds
+
+**Problem:**
+
+`ML_GUARDRAIL_TIMEOUT_IN_SECONDS` was raised from 5s to 60s. Combined with
+fail-closed semantics (`isAccepted` defaults to false), a hung guardrail model
+pins every ML predict thread for 60s then rejects — causing thread-pool
+starvation of the `opensearch_ml_predict_remote` pool under moderate load.
+
+**Solution:**
+
+Change the constant back to 5 seconds:
+
+```java
+public static final long ML_GUARDRAIL_TIMEOUT_IN_SECONDS = 5;
+```
+
+**File to modify:**
+- `common/src/main/java/org/opensearch/ml/common/model/ModelGuardrail.java` (line 56)
+
+---
+
+## 3. Session max_count eviction: add count short-circuit
+
+**Problem:**
+
+`identifyCountBasedExpiredSessions` → `identifyCountBasedPage` enumerates ALL
+non-pinned sessions at 100/page (sorted created_time DESC), counting until it
+passes `maxCount`. A container with 100k sessions and max_count=100k issues
+1,000 sequential search requests per run to delete nothing.
+
+The long-term (`executeLongTermMaxCount`) and history paths already have a
+`size(0)/trackTotalHits` pre-count that short-circuits when total <= maxCount.
+The session path is missing this.
+
+**Solution:**
+
+Mirror the pattern from `executeLongTermMaxCount` (line 870):
+
+1. First: do a `size(0), trackTotalHits(true)` count query for non-pinned
+   sessions in this container
+2. If `totalCount <= maxCount`: return empty set immediately (no excess)
+3. If `totalCount > maxCount`: compute `excess = totalCount - maxCount`, then
+   use `collectOldestDocIds(sessionIndex, containerId, CREATED_TIME_FIELD, excess, ...)`
+   to fetch only the excess oldest session IDs
+
+This replaces the current `identifyCountBasedPage` approach entirely for
+sessions, making it consistent with LT and history.
+
+**Files to modify:**
+- `plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java`
+  - Replace `identifyCountBasedExpiredSessions` / `identifyCountBasedPage`
+    with the count-first pattern
+
+---
+
+## 4. Orphan sweep: O(total docs) enumeration → O(unique sessions)
+
+**Problem:**
+
+`enumerateSessionIdsFromWorkingMemory()` pages through every working-memory
+document in the container (at 1000 docs/page) to extract distinct
+`namespace.session_id` values. For 5M working-memory messages across 2,000
+sessions, this is 5,000 sequential search requests.
+
+The `namespace` field is mapped as `flat_object` in `ml_memory_working.json`
+(line 42-44). `flat_object` does NOT support terms aggregations or field
+collapsing, so we cannot simply do `collapse("namespace.session_id")` or a
+composite aggregation.
+
+**Solution — add a top-level `session_id` keyword field:**
+
+The cleanest fix is to denormalize `session_id` into a top-level keyword field
+on working-memory documents, enabling efficient aggregation/collapse.
+
+### Step 1: Add field to index mapping
+
+In `common/src/main/resources/index-mappings/ml_memory_working.json`:
+
+```json
+"session_id": {
+  "type": "keyword"
+}
+```
+
+### Step 2: Populate on write
+
+In `TransportAddMemoriesAction.java`, where working memory documents are
+indexed, the session_id is already available from
+`input.getNamespace().get(SESSION_ID_FIELD)`. Add it as a top-level field in
+the index request source alongside `namespace`.
+
+### Step 3: Use composite/terms aggregation in orphan sweep
+
+Replace `enumerateSessionIdsFromWorkingMemory()` with a composite aggregation:
+
+```java
+CompositeAggregationBuilder agg = new CompositeAggregationBuilder("session_ids",
+    List.of(new TermsValuesSourceBuilder("sid").field("session_id")))
+    .size(ORPHAN_SESSION_BATCH_SIZE);
+```
+
+Page through the composite aggregation using `after_key` — each page returns
+up to 1000 unique session IDs directly, with cost proportional to unique
+sessions, not total documents.
+
+### Step 4: Update queries
+
+The orphan sweep queries (`cascadeDeleteWorkingMemory`,
+`deleteOrphanedWorkingMemory`, etc.) currently filter on
+`namespace.session_id`. Update them to use the top-level `session_id` field
+for better performance (keyword vs flat_object subpath).
+
+### Backward compatibility note
+
+Existing working-memory documents won't have the new `session_id` field. The
+orphan sweep should fall back to the current `namespace.session_id` exists
+query for documents missing the top-level field, or run a one-time backfill.
+Alternatively, accept that pre-upgrade orphans use the old (slower) path until
+they're naturally cleaned up.
+
+**Files to modify:**
+- `common/src/main/resources/index-mappings/ml_memory_working.json`
+- `plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java`
+- `plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java`
+  (enumerateSessionIdsFromWorkingMemory, cascadeDeleteWorkingMemory,
+  deleteOrphanedWorkingMemory)
+
+---
+
+## 5. Orphan sweep: ORPHAN_SESSION_ID_CAP never advances across runs
+
+**Problem:**
+
+The orphan enumeration always starts from `searchAfter=null` sorted `_id ASC`
+and stops at 50k unique session IDs. If the first 50k in _id order are all
+live sessions, orphaned working memory beyond the cap is never examined on any
+subsequent run.
+
+**Solution:**
+
+Randomize the sort key so each run starts from a different position in the
+keyspace. Over multiple runs, all regions are probabilistically covered.
+
+Replace:
+```java
+.sort("_id", SortOrder.ASC)
+```
+
+With a random score sort (or rotate based on run count):
+```java
+.sort(SortBuilders.scriptSort(
+    new Script(ScriptType.INLINE, "painless",
+        "doc['_id'].value.hashCode() ^ params.seed", Map.of("seed", System.currentTimeMillis())),
+    ScriptSortBuilder.ScriptSortType.NUMBER))
+```
+
+Alternatively (simpler): use a hash of (container_id + run_timestamp) as a
+prefix filter or sort tiebreaker so each run covers a different slice.
+
+Also update the log message from "Remaining orphans will be caught on next
+run" to "Remaining orphans will be covered in future runs" since coverage is
+now probabilistic rather than guaranteed-next-run.
+
+If the fix in #4 above (composite aggregation on top-level session_id) is
+implemented, this cap becomes much less likely to be hit since enumeration
+cost drops by orders of magnitude and the cap could be raised significantly.
+
+**Files to modify:**
+- `plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java`
+  (enumerateSessionIdsFromWorkingMemory)
+
+---
+
+## 6. Unconditional 5s throttle between containers → throttle only after deletions
+
+**Problem:**
+
+`scheduleNext()` delays 5 seconds after EVERY container, including containers
+that needed zero work. At 10k containers × 5s = ~14 hours of pure sleep per
+run, potentially exceeding the 24h job interval.
+
+**Solution:**
+
+Only throttle after containers where deletions actually occurred. Containers
+that were skipped (opt-out, no policy, no expired data, backfill-only) should
+proceed immediately to the next container.
+
+Change `processContainer` to track whether any deletions happened and pass
+that to the chain:
+
+```java
+private void processContainer(SearchHit hit, ActionListener<Boolean> listener) {
+    // ... existing logic ...
+    // listener.onResponse(true) when deletions occurred
+    // listener.onResponse(false) when no deletions (skip, opt-out, etc.)
+}
+
+private void processContainerChain(SearchHit[] hits, int index, Object[] nextPageSortValues) {
+    if (index >= hits.length) { ... }
+
+    processContainer(hits[index], ActionListener.wrap(hadDeletions -> {
+        if (hadDeletions) {
+            scheduleNext(hits, index + 1, nextPageSortValues);
+        } else {
+            // No deletions — proceed immediately
+            processContainerChain(hits, index + 1, nextPageSortValues);
+        }
+    }, e -> {
+        log.error(...);
+        processContainerChain(hits, index + 1, nextPageSortValues);
+    }));
+}
+```
+
+This means a run over 10k containers where only 50 have expired data takes
+~250s of throttle instead of ~14 hours.
+
+**Files to modify:**
+- `plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java`
+  (processContainer, processContainerChain, executeRetentionPipeline)
+
+---
+
+## 7. Test coverage gaps (6 items)
+
+These are the critical test gaps that leave data-safety paths unprotected
+against regressions. Each can be a separate test method.
+
+### 7a. applyDefaultRetentionPolicy — painless backfill script
+
+Test that:
+- When cluster defaults are configured (>0) and container has no policy, the
+  conditional script update is issued
+- When the script returns NOOP (user set policy concurrently), the container
+  is skipped (no deletions run)
+- When the script succeeds, executeRetentionPipeline runs with the backfilled
+  policy
+- When all defaults are -1, no update is issued
+
+### 7b. Epoch-second normalization
+
+Test that:
+- A session with `last_updated_time` in epoch seconds (e.g., 1752400000) that
+  is newer than the retention cutoff when normalized is NOT deleted
+- A session with `last_updated_time` in epoch millis that IS older than the
+  cutoff IS deleted
+- A session with missing/unparseable last_updated_time is SKIPPED (not deleted)
+
+### 7c. buildEpochAwareCutoffQuery structure
+
+Assert that the delete-by-query for long-term and working-memory includes both
+range clauses (gte threshold + lt cutoff; lt secondsScaleCutoff) and
+minimumShouldMatch=1. A mutation to plain lt(cutoffMillis) should fail the
+test.
+
+### 7d. retention_policy:null opt-out in processor
+
+Test that a container with `"retention_policy": null` in source triggers no
+searches, no deletes, and no default-policy backfill update.
+
+### 7e. Non-content updates don't bump last_updated_time
+
+Test that updating only `pinned`, `tags`, `metadata`, or `additional_info` on
+a session does NOT include `last_updated_time` in the indexed doc. A mutation
+making `shouldBumpLastUpdatedTime` always return true should fail.
+
+### 7f. Session last_updated_time bump on add-memories
+
+Test that:
+- Adding a message to an existing session issues a client.update with
+  LAST_UPDATED_TIME_FIELD
+- Adding a message to a newly created session does NOT issue the bump (session
+  was just created with current time)
+- The bump failure does not fail the add-memories response
+
+**Files to modify:**
+- `plugin/src/test/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessorTests.java`
+- `plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryActionTests.java`
+- `plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java`
+
+---
+
+## 8. Add document IDs to retention job logs at DEBUG level
+
+**Decision context:**
+
+Retention deletions are recorded in job logs only (no audit index, no records
+written into the agent's history index — that would pollute the prompt
+pipeline and create recursive-retention problems). The existing INFO-level
+aggregate counts stay unchanged. We're adding DEBUG-level lines that show
+WHICH documents were deleted, so an operator investigating "what disappeared"
+can raise the log level and trace specific IDs.
+
+**Rules:**
+
+- INFO lines (aggregate counts) stay exactly as they are
+- NO memory content in logs at any level — IDs and criteria only (content may
+  contain PII/credentials)
+- Guard each new line with `if (log.isDebugEnabled())` (or rely on log4j2
+  parameterized-message laziness for cheap args; use the explicit guard where
+  building the message involves collection formatting)
+- Do NOT add extra search requests just to obtain IDs for logging. Log IDs
+  where the code already holds them; log the deletion criteria where it
+  doesn't (delete-by-query paths).
+
+**Where to add DEBUG lines in `MemoryRetentionJobProcessor.java`:**
+
+ID-based paths (we already hold the exact IDs — log them):
+- `deleteSessionBatch` / session deletion: log the expired session IDs per
+  batch, e.g. `container={} deleting session ids={}`
+- `deleteDocumentsBatchRecursive` (used by long-term max_count and history
+  max_count): log the doc IDs per batch with the index name
+
+Delete-by-query paths driven by known session-ID batches (log the batch):
+- `cascadeDeleteBatch` (working memory cascade): log the session-ID batch
+  driving the DBQ
+- `deleteOrphanBatch` (orphaned working memory): log the orphan session-ID
+  batch
+
+Pure criteria-based delete-by-query paths (no IDs available — log criteria,
+do NOT pre-fetch IDs):
+- `executeLongTermRetentionDays`: log index + cutoffMillis
+- `executeWorkingMemoryTTL`: log index + cutoffMillis
+- `deleteEmptyNamespaceWorkingMemory`: log index + cutoffMillis
+
+Log quantity is not a concern; do not truncate the ID lists.
+
+**Test:** add one test asserting a DEBUG deletion-IDs line is emitted for the
+session-deletion path (e.g., via a log appender capture), and that the INFO
+aggregate line is unchanged.
+
+**Files to modify:**
+- `plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java`
+- `plugin/src/test/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessorTests.java`
--- a/DESIGN_DOC.md
+++ b/DESIGN_DOC.md
@@ -127,7 +127,7 @@ There is no deletion job, no TTL field, and no retention configuration anywhere 
 │    5. History   → delete over-count only, skip pinned               │
 │    6. WM TTL    → (disableSession only) delete old working memory   │
 │    7. Orphan    → sweep orphaned working memory (once per job run)  │
-│    8. Throttle  → pause 5s before next container                    │
+│    8. Throttle  → pause 5s before next container (only if deleted)  │
 │                                                                     │
 │  Deletion: _delete_by_query (throttled, idempotent)                 │
 └─────────────────────────────────────────────────────────────────────┘
@@ -356,6 +356,8 @@ When a customer updates a retention policy via the PUT API, **field-level merge*
 Example: current policy is `"sessions": {"retention_days": 60, "max_count": 500}`. Customer sends `{"sessions": {"max_count": 350}}`. Result is `"sessions": {"retention_days": 60, "max_count": 350}` — `retention_days` is unchanged because it was not in the request.
 
 To remove `retention_days`: send `{"sessions": {"retention_days": null, "max_count": 350}}`.
+
+**Implementation detail:** When a field is explicitly set to null, `RetentionRule.toXContent()` emits `builder.nullField(fieldName)` so the partial-update merge correctly removes the stored value. Without this, omitting a field from the serialized update document would leave the old value intact (OpenSearch's partial merge preserves unmentioned fields). The `retentionDaysExplicitlySet` and `maxCountExplicitlySet` flags track whether the null was explicit (user intent to remove) vs. simply absent (user didn't mention the field — preserve current value).
 
 Types not included in the update request are left unchanged. The PUT response **always returns the full resulting retention_policy** so the developer can verify the final state.
 
@@ -705,9 +707,10 @@ FOR each memory container WHERE retention_policy IS NOT NULL:
     IF expired_session_ids is not empty:
         _delete_by_query working_index WHERE:
             memory_container_id = {id}
-            AND namespace.session_id IN [expired_session_ids]
+            AND (session_id IN [expired_session_ids]    // top-level keyword
+                 OR namespace.session_id IN [expired_session_ids])  // legacy flat_object
         // No pinned check — working memory does not support pinning
-        // NOTE: namespace is a flat_object; use termQuery("namespace.session_id", id)
+        // Uses buildSessionIdMatchQuery(): should(terms(session_id), terms(namespace.session_id)) minimumShouldMatch(1)
     
     // Phase 3: Delete the session documents themselves
     IF expired_session_ids is not empty:
@@ -747,7 +750,10 @@ FOR each memory container WHERE retention_policy IS NOT NULL:
         // working_memory_ttl_days is a cluster setting (default 30d), always present
     
     LOG deletion counts per type (including cascade working memory count)
-    PAUSE {throttle_interval} seconds (use threadPool.schedule(), not Thread.sleep())
+    IF any deletions occurred in this container:
+        PAUSE {throttle_interval} seconds (use threadPool.schedule(), not Thread.sleep())
+    ELSE:
+        Proceed immediately to next container (no throttle for no-op containers)
     // Then invoke next container via recursive callback (see note below)
 
 // Phase 7: Orphan Sweep (runs once per full job, not per-container)
@@ -756,22 +762,31 @@ FOR each memory container WHERE disableSession = false:
     sessions_index = container.configuration.getIndexName(MemoryType.SESSIONS)
     
     // Find working memory pointing to non-existent sessions.
-    // NOTE: namespace is a flat_object. Composite aggregation on flat_object sub-fields
-    // may not work reliably. Use paginated scroll search instead:
-    SCROLL SEARCH working_index WHERE memory_container_id = {id}
-        → extract distinct namespace.session_id values (deduplicate in memory)
-    MULTI-GET those session IDs against sessions_index (batched, 1000 per request)
+    // Uses a paged composite aggregation on the top-level session_id keyword field
+    // (namespace is flat_object and cannot be aggregated). Enumeration starts at a
+    // random keyspace position each run and wraps around, so the 50K cap truncates
+    // a different slice each run for probabilistic full coverage.
+    start_key = randomEnumerationStartKey()
+    COMPOSITE AGGREGATION on session_id field, starting at start_key, wrap around
+        → collect distinct session_id values (up to ORPHAN_SESSION_ID_CAP = 50,000)
+    // Legacy fallback: for pre-upgrade docs without top-level session_id:
+    SEARCH_AFTER on docs WHERE namespace.session_id EXISTS AND session_id NOT EXISTS
+        → extract namespace.session_id values, same random-start wrap-around pattern
+    
+    CHECK session existence via batched termsQuery("_id", batch) (1000 per request)
     orphan_ids = session IDs that returned NOT FOUND
     IF orphan_ids is not empty:
         _delete_by_query working_index WHERE:
             memory_container_id = {id}
-            AND namespace.session_id IN [orphan_ids]
+            AND (session_id IN [orphan_ids] OR namespace.session_id IN [orphan_ids])
     
     // Find working memory with no session_id at all (shouldn't exist, but can
     // from partial writes — grace period allows incomplete writes to be finished)
     _delete_by_query working_index WHERE:
         memory_container_id = {id}
-        AND (namespace.session_id IS NULL OR namespace IS NULL)
+        AND session_id NOT EXISTS
+        AND namespace.session_id NOT EXISTS
+        AND namespace.user_id NOT EXISTS AND namespace.agent_id NOT EXISTS
         AND created_time < (now - orphan_ttl_days)  // configurable, default 7d
         // 7-day grace: gives partially-written documents time to have session_id
         // set in a subsequent update before being treated as orphans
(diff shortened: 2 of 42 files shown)


AirFunnyB:
Splitting into multiple PR's for easier review

PR #4914, opened 2026-07-17
[Memory Retention] Add retention policy data model, pinned field, and API wiring (1/2)

AirFunnyB:
## Description

First of two PRs implementing the memory retention lifecycle (RFC #4859). This PR adds the **data model and API surface** for retention; it does not yet run any eviction. The retention job, cluster defaults, and hardening land in a stacked follow-up PR (2/2).

Scoping this as two PRs keeps each reviewable — this one is purely additive plumbing with no background behavior, so it can merge safely ahead of the job that consumes it.

#### What's included

**Data model (`opensearch-ml-common`)**

- New `RetentionRule` model (`retention_days`, `max_count`), with validation that both must be positive integers or null, plus `Writeable` + XContent parse/serialize.
- `MemoryConfiguration` gains a `retention_policy` field (map of memory type → `RetentionRule`) with parsing, serialization, and equality.
- New `pinned` boolean field on `MLMemorySession`, `MLLongTermMemory`, and `MLMemoryHistory` models — marks a memory as exempt from future eviction.
- New constants: `PINNED_FIELD`, `RETENTION_POLICY_FIELD`, `RETENTION_DAYS_FIELD`, `MAX_COUNT_FIELD`.
- Index mappings updated for `pinned` (session/long-term/history) and `retention_policy` (container).

**API wiring (`opensearch-ml-plugin`)**

- `retention_policy` is accepted and validated on the Create and Update Memory Container transport actions.
- `pinned` is accepted on Add/Update Memory, with validation that it is only allowed for non-working memory types (working memory rejects `pinned` with a 400 — to preserve a conversation, pin the session instead).

**Feature-flag gating (kill switch)**

- Both `retention_policy` and `pinned` are gated behind the `plugins.ml_commons.memory_retention_enabled` cluster setting. When the feature is disabled, setting either is rejected with a `403 FORBIDDEN` and a clear message — keeping the whole retention surface behind a single cluster-level kill switch.

**Update semantics**

- Updating a memory container supports **clearing** `retention_policy` by passing an explicit `null` (mirrors the partial-document update pattern): omitting the field keeps the existing policy, while an explicit `null` removes it.
- A combined strategy + retention update preserves the retention policy instead of dropping it during the strategy rebuild.

#### What's NOT in this PR (coming in 2/2)

- The scheduled retention job (session eviction, long-term/history cleanup, working memory TTL, orphan sweep).
- Cluster-level default retention policies and auto-apply / opt-out behavior.
- Job-related hardening.

No eviction runs as a result of this PR — retention fields are stored and validated only. Existing containers and memories are unaffected (all new fields default to absent/false; no backfill).

### Related Issues

RFC #4859

### Check List

- [x] New functionality has been documented (RFC #4859).
- [x] API changes companion pull request created, if applicable.

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
@@ -289,6 +289,11 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
                 entry.getValue().toXContent(builder, params);
             }
             builder.endObject();
+        } else if (retentionPolicyExplicitlyNull) {
+            // Emit an explicit null so the UpdateRequest.doc() partial merge removes the stored policy
+            // (mirrors the dimension/SPARSE_ENCODING pattern above). Gated on the wipe flag so a plain
+            // policy-less container never gains a spurious retention_policy: null field.
+            builder.nullField(RETENTION_POLICY_FIELD);
         }
         builder.endObject();
         return builder;
@@ -646,7 +651,11 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         // Merge retention policy
         if (updateContent.isRetentionPolicyExplicitlyNull()) {
             this.retentionPolicy = null;
+            // Carry the wipe intent onto this config so toXContent() emits an explicit null and the
+            // partial-document merge (UpdateRequest.doc()) removes the stored policy.
+            this.retentionPolicyExplicitlyNull = true;
         } else if (updateContent.getRetentionPolicy() != null) {
+            this.retentionPolicyExplicitlyNull = false;
             validateRetentionPolicy(updateContent.getRetentionPolicy());
             if (this.retentionPolicy == null) {
                 this.retentionPolicy = new EnumMap<>(MemoryType.class);
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.common.memorycontainer;
 import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.ml.common.CommonValue.ML_AGENTIC_MEMORY_SYSTEM_INDEX_PREFIX;
 import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
+import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DEFAULT_MEMORY_INDEX_PREFIX;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DIMENSION_FIELD;
 import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.DISABLE_HISTORY_FIELD;
@@ -64,7 +65,6 @@ import lombok.extern.log4j.Log4j2;
  */
 @Getter
 @Setter
-@Builder
 @EqualsAndHashCode
 @Log4j2
 public class MemoryConfiguration implements ToXContentObject, Writeable {
@@ -73,22 +73,18 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
     private String embeddingModelId;
     private String llmId;
     private Integer dimension;
-    @Builder.Default
-    private Integer maxInferSize = MAX_INFER_SIZE_DEFAULT_VALUE;
+    private Integer maxInferSize;
     private List<MemoryStrategy> strategies;
-    @Builder.Default
-    private Map<String, Map<String, Object>> indexSettings = new HashMap<>();
-    @Builder.Default
-    private Map<String, Object> parameters = new HashMap<>();
-    @Builder.Default
-    private boolean disableHistory = false;
-    @Builder.Default
-    private boolean disableSession = false;
-    @Builder.Default
-    private boolean useSystemIndex = true;
+    private Map<String, Map<String, Object>> indexSettings;
+    private Map<String, Object> parameters;
+    private boolean disableHistory;
+    private boolean disableSession;
+    private boolean useSystemIndex;
     private String tenantId;
     private Map<MemoryType, RetentionRule> retentionPolicy;
+    private transient boolean retentionPolicyExplicitlyNull = false;
 
+    @Builder
     public MemoryConfiguration(
         String indexPrefix,
         FunctionName embeddingModelType,
@@ -99,17 +95,23 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         List<MemoryStrategy> strategies,
         Map<String, Map<String, Object>> indexSettings,
         Map<String, Object> parameters,
-        boolean disableHistory,
-        boolean disableSession,
-        boolean useSystemIndex,
+        Boolean disableHistory,
+        Boolean disableSession,
+        Boolean useSystemIndex,
         String tenantId,
         Map<MemoryType, RetentionRule> retentionPolicy
     ) {
+        // Apply defaults for Boolean parameters
+        boolean effectiveUseSystemIndex = (useSystemIndex != null) ? useSystemIndex : true;
+        boolean effectiveDisableHistory = (disableHistory != null) ? disableHistory : false;
+        boolean effectiveDisableSession = (disableSession != null) ? disableSession : false;
+
         // Validate first
         validateInputs(embeddingModelType, embeddingModelId, dimension, maxInferSize);
+        validateRetentionPolicy(retentionPolicy);
 
         // Assign values after validation
-        this.indexPrefix = buildIndexPrefix(indexPrefix, useSystemIndex);
+        this.indexPrefix = buildIndexPrefix(indexPrefix, effectiveUseSystemIndex);
         this.embeddingModelType = embeddingModelType;
         this.embeddingModelId = embeddingModelId;
         this.llmId = llmId;
@@ -127,9 +129,9 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         if (parameters != null && !parameters.isEmpty()) {
             this.parameters.putAll(parameters);
         }
-        this.disableHistory = disableHistory;
-        this.disableSession = disableSession;
-        this.useSystemIndex = useSystemIndex;
+        this.disableHistory = effectiveDisableHistory;
+        this.disableSession = effectiveDisableSession;
+        this.useSystemIndex = effectiveUseSystemIndex;
         this.tenantId = tenantId;
         this.retentionPolicy = retentionPolicy;
     }
@@ -173,13 +175,15 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         this.disableSession = input.readBoolean();
         this.useSystemIndex = input.readBoolean();
         this.tenantId = input.readOptionalString();
-        if (input.readBoolean()) {
-            int size = input.readVInt();
-            this.retentionPolicy = new EnumMap<>(MemoryType.class);
-            for (int i = 0; i < size; i++) {
-                MemoryType key = input.readEnum(MemoryType.class);
-                RetentionRule value = new RetentionRule(input);
-                this.retentionPolicy.put(key, value);
+        if (input.getVersion().onOrAfter(VERSION_3_7_0)) {
+            if (input.readBoolean()) {
+                int size = input.readVInt();
+                this.retentionPolicy = new EnumMap<>(MemoryType.class);
+                for (int i = 0; i < size; i++) {
+                    MemoryType key = input.readEnum(MemoryType.class);
+                    RetentionRule value = new RetentionRule(input);
+                    this.retentionPolicy.put(key, value);
+                }
             }
         }
     }
@@ -214,15 +218,17 @@ public class MemoryConfiguration implements ToXContentObject, Writeable {
         out.writeBoolean(disableSession);
         out.writeBoolean(useSystemIndex);
         out.writeOptionalString(tenantId);
-        if (retentionPolicy != null && !retentionPolicy.isEmpty()) {
-            out.writeBoolean(true);
-            out.writeVInt(retentionPolicy.size());
-            for (Map.Entry<MemoryType, RetentionRule> entry : retentionPolicy.entrySet()) {
-                out.writeEnum(entry.getKey());
-                entry.getValue().writeTo(out);
+        if (out.getVersion().onOrAfter(VERSION_3_7_0)) {
+            if (retentionPolicy != null && !retentionPolicy.isEmpty()) {
+                out.writeBoolean(true);
+                out.writeVInt(retentionPolicy.size());
+                for (Map.Entry<MemoryType, RetentionRule> entry : retentionPolicy.entrySet()) {
+                    out.writeEnum(entry.getKey());
+                    entry.getValue().writeTo(out);
+                }
+            } else {
+                out.writeBoolean(false);
             }
-        } else {
-            out.writeBoolean(false);
         }
     }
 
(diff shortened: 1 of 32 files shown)


mingshl:
write the api example in the PR description please

mingshl:
is the feature flag included in this PR?

nathaliellenaa:
> is the feature flag included in this PR?

Can we include the feature flag in this PR? That way, if PR 2 doesn't get merged before the code freeze, we can simply disable the feature as a fallback.

common/src/main/resources/index-mappings/ml_memory_container.json
nathaliellenaa: Need to bump the schema_version here, same for the other mapping files

nathaliellenaa:
Could you add some ITs?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -200,6 +218,18 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeBoolean(disableSession);
         out.writeBoolean(useSystemIndex);
         out.writeOptionalString(tenantId);
+        if (out.getVersion().onOrAfter(VERSION_3_8_0)) {
```
nathaliellenaa, line 222: The map serializes correctly here, but should `retentionPolicyExplicitlyNull` also be written/read? If a `"retention_policy": null ` update ever crosses a node boundary, the flag resets to `false` on the receiving side and `update()` would preserve the old policy instead of wiping it — should we version-gate the flag on `VERSION_3_8_0` too?

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
```diff
@@ -348,6 +362,9 @@ public static MLAddMemoriesInput parse(XContentParser parser, String memoryConta
                 case OWNER_ID_FIELD:
                     ownerId = parser.text();
                     break;
+                case PINNED_FIELD:
+                    pinned = parser.booleanValue();
```
nathaliellenaa: `pinned` is parsed with a bare `parser.booleanValue()`, so` "pinned": null` throws instead of being treated as absent like the model classes.

Suggested fix:
```
pinned = parser.currentToken() == XContentParser.Token.VALUE_NULL ? null : parser.booleanValue();
```

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
```diff
@@ -91,6 +96,7 @@ public MLAddMemoriesInput(
         Map<String, String> tags,
         Map<String, Object> parameters,
         String ownerId,
+        Boolean pinned,
```
nathaliellenaa: Inserting `pinned` mid-list in the all-args constructor breaks any positional caller. Suggest to move `pinned` to the end of the constructor parameter list.

AirFunnyB:
#

> > is the feature flag included in this PR?
> 
> Can we include the feature flag in this PR? That way, if PR 2 doesn't get merged before the code freeze, we can simply disable the feature as a fallback.

Yes added plugins.ml_commons.memory.retention_enabled.  Set default false for this PR. PR2 will be default true. Create/update reject a retention_policy with 403 when it's off, so retention can be disabled cluster-wide as a fallback.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
AirFunnyB: Moved pinned to the end of the constructor (and field order, since Lombok @Builder derives the constructor from it). All callers use the builder.

common/src/main/java/org/opensearch/ml/common/transport/memorycontainer/memory/MLAddMemoriesInput.java
AirFunnyB: Used your suggestion: pinned = parser.currentToken() == VALUE_NULL ? null : parser.booleanValue(). Added parse tests for null and true.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Good catch — now writing/reading the flag inside the VERSION_3_8_0 gate so the null-wipe survives a node boundary. Added a stream round-trip test.

common/src/main/resources/index-mappings/ml_memory_container.json
AirFunnyB: Done.

AirFunnyB:
> Could you add some ITs?

Done added RestMemoryContainerRetentionIT: create→get round-trip, explicit-null wipe on update, working/history validation rejections, and retention-disabled rejection.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -96,16 +95,23 @@ public MemoryConfiguration(
         List<MemoryStrategy> strategies,
         Map<String, Map<String, Object>> indexSettings,
         Map<String, Object> parameters,
-        boolean disableHistory,
-        boolean disableSession,
-        boolean useSystemIndex,
-        String tenantId
+        Boolean disableHistory,
```
mingshl, line 98: boolean and Boolean are different

mingshl: the main thing is Boolean accept null value and you don't want to change them

AirFunnyB: That's deliberate: I moved @Builder onto the constructor (to thread retentionPolicy through validation), and @Builder.Default is ignored when @Builder is on a constructor. So builder().build() leaves an unset primitive boolean at false, which would flip useSystemIndex from its intended true default — and both MLCreateMemoryContainerInput and MLMemoryContainer call bare builder().build(). Boxing the params lets the constructor detect "unset" (null) and re-apply the right default. No serialization or getter change.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
```diff
@@ -118,6 +118,18 @@ protected void doExecute(Task task, MLAddMemoriesRequest request, ActionListener
             return;
         }
 
+        // Reject pinned field — add-memories creates working memory which cannot be pinned
+        if (input.getPinned() != null) {
```
nathaliellenaa, line 122: Should `pinned` also be gated behind `isMemoryRetentionEnabled()` here? Right now `retention_policy` is flag-gated but pinned isn't, so the kill switch only half-disables the feature.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportUpdateMemoryAction.java
```diff
@@ -136,6 +137,20 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<IndexR
                     return;
                 }
 
+                // Reject pinned field for working memory type
+                Map<String, Object> updateContent = updateRequest.getMlUpdateMemoryInput().getUpdateContent();
+                if (memoryType == MemoryType.WORKING && updateContent.containsKey(PINNED_FIELD)) {
```
nathaliellenaa, line 152: Same as above

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
AirFunnyB: You're right that pinned is dead metadata without the retention job reading it, so will make the kill switch cover it too. Added the isMemoryRetentionEnabled() gate in TransportUpdateMemoryAction (403 when disabled), which is where pinned is actually persisted. TransportAddMemoriesAction already rejects pinned unconditionally (working memory can't be pinned), so nothing slips through there. Covered by a new unit test and an IT.

nathaliellenaa (approved):
Thanks for addressing the comments @AirFunnyB , LGTM

jiapingzeng:
@AirFunnyB a few of the ITs from this PR are failing when I run them locally, could you take a look?

```
> ./gradlew :opensearch-ml-plugin:integTest
...
RestMemoryContainerRetentionIT > testPinnedRoundTripOnSession FAILED
    org.opensearch.client.ResponseException: method [PUT], host [http://[::1]:43287], URI [/_plugins/_ml/memory_containers/2x6XgJ8B45QAzN5YLbCj/memories/sessions/pinned-session-1], status line [HTTP/1.1 404 Not Found]
    {"error":{"root_cause":[{"type":"status_exception","reason":"Memory not found"}],"type":"status_exception","reason":"Memory not found"},"status":404}
        at __randomizedtesting.SeedInfo.seed([9BE8C6554C6604F9:656F757DDA9AC637]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:498)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:387)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:362)
        at app//org.opensearch.ml.utils.TestHelper.performRequestWithRemoteRetry(TestHelper.java:216)
        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:207)
        at app//org.opensearch.ml.utils.TestHelper.makeRequest(TestHelper.java:180)
        at app//org.opensearch.ml.rest.RestMemoryContainerRetentionIT.testPinnedRoundTripOnSession(RestMemoryContainerRetentionIT.java:172)
...
RestMemoryContainerRetentionIT > testPinnedRejectedWhenFeatureDisabled FAILED
    java.lang.AssertionError: expected:<403> but was:<404>
        at __randomizedtesting.SeedInfo.seed([9BE8C6554C6604F9:7321D98ACFFD7FBF]:0)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.opensearch.ml.rest.RestMemoryContainerRetentionIT.testPinnedRejectedWhenFeatureDisabled(RestMemoryContainerRetentionIT.java:223)
...
RestMemoryContainerRetentionIT > testUpdateContainerWipesRetentionPolicyWithExplicitNull FAILED
    java.lang.AssertionError: retention_policy should be wiped after explicit null update expected null, but was:<{sessions={retention_days=30.0}}>
        at __randomizedtesting.SeedInfo.seed([9BE8C6554C6604F9:4D60A99EE4311439]:0)
        at org.junit.Assert.failNotNull(Assert.java:756)
        at org.junit.Assert.assertNull(Assert.java:738)
        at org.opensearch.ml.rest.RestMemoryContainerRetentionIT.testUpdateContainerWipesRetentionPolicyWithExplicitNull(RestMemoryContainerRetentionIT.java:90)
```

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -250,6 +282,14 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
         if (tenantId != null) {
             builder.field(TENANT_ID_FIELD, tenantId);
         }
+        if (retentionPolicy != null && !retentionPolicy.isEmpty()) {
```
jiapingzeng, line 285: Seems that we don't handle the case where `retentionPolicy` is explicitly null here which is why `testUpdateContainerWipesRetentionPolicyWithExplicitNull` fails

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
```diff
@@ -535,6 +643,18 @@ public void update(MemoryConfiguration updateContent) {
             // Only update dimension for TEXT_EMBEDDING if provided
             this.dimension = updateContent.getDimension();
         }
+        // Merge retention policy
+        if (updateContent.isRetentionPolicyExplicitlyNull()) {
+            this.retentionPolicy = null;
```
jiapingzeng, line 653: Should we also set `this.retentionPolicyExplicitlyNull = true;` here?

plugin/src/test/java/org/opensearch/ml/rest/RestMemoryContainerRetentionIT.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test
+    public void testPinnedRoundTripOnSession() throws IOException {
+        // Bare container (no LLM/embedding) so sessions work without external credentials.
+        String containerId = createContainer("{ \"name\": \"pinned_it_" + System.currentTimeMillis() + "\", \"configuration\": {} }");
+
+        // Add a memory with a session_id namespace -> creates a session and returns its id.
+        String addBody = "{\n"
+            + "  \"messages\": [{ \"role\": \"user\", \"content\": [{ \"type\": \"text\", \"text\": \"hello\" }] }],\n"
+            + "  \"namespace\": { \"session_id\": \"pinned-session-1\" },\n"
```
jiapingzeng: I don't think a session is actually being created here since we are passing session id manually, which is why it is showing 404 in ITs.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportUpdateMemoryContainerAction.java
jiapingzeng, line 182: I believe right now, if we update strategy and retention at the same time, retention will get silently dropped here. Should we also add `.retentionPolicy(updateConfiguration.getRetentionPolicy())` here? Can we add an IT as well?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Fixed — toXContent() now emits an explicit "retention_policy": null when the wipe flag is set, gated on the flag so a normal policy-less container doesn't gain a spurious null field.

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryConfiguration.java
AirFunnyB: Done. update() now sets retentionPolicyExplicitlyNull = true in the wipe branch (and resets it to false in the non-null branch so a stale flag doesn't carry over into a later normal update).

plugin/src/test/java/org/opensearch/ml/rest/RestMemoryContainerRetentionIT.java
AirFunnyB: You're right. Reworked both pinned ITs to create a real session via POST /memories/sessions first then pin it and verify. Both pass now.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportUpdateMemoryContainerAction.java
AirFunnyB: Confirmed and fixed. Added .retentionPolicy(updateConfiguration.getRetentionPolicy()) plus setRetentionPolicyExplicitlyNull(...) — the flag is a transient field outside the @Builder constructor so it needs the explicit setter. Added a unit test (testUpdateStrategiesAndRetentionTogether_RetentionNotDropped) asserting retention survives the combined path.

AirFunnyB:
Note on the retention kill-switch: when retention_enabled=false, the gate rejects setting a non-empty retention_policy (403) but intentionally still allows an explicit "retention_policy": null wipe. Rationale: if an admin flips the flag off as an emergency stop, operators should be able to clear the stale policies off existing containers without turning the policy back on.

jiapingzeng (approved):

mingshl:
these IT are failing 
```
2868 tests completed, 3 failed, 61 skipped
Tests with failures:
 - org.opensearch.ml.settings.MLFeatureEnabledSettingTests.testToolExecuteSettings
 - org.opensearch.ml.settings.MLFeatureEnabledSettingTests.testMetricCollectionSettings
 - org.opensearch.ml.settings.MLFeatureEnabledSettingTests.testAddListenerAndNotify


```

mingshl (approved):

PR #4918, opened 2026-07-21
[Memory Retention] Add retention job, cluster defaults, and hardening (2/2)

AirFunnyB:
# [Memory Retention] Add retention job, cluster defaults, and hardening (2/2)

## Summary

Adds a scheduled background job (`MemoryRetentionJobProcessor`) that garbage-collects agentic memory. On a fixed interval it walks every memory container and, for each one that has a retention policy, deletes memory that has aged out or exceeded a count cap. It also runs an orphan sweep for dangling working memory.

Retention rules cover three memory types, with different supported dimensions:

| Memory type | Age (`retention_days`) | Count (`max_count`) | Notes |
|---|---|---|---|
| **Sessions** | ✅ | ✅ | Deleting a session cascades to its working memory. |
| **Long-term** | ✅ | ✅ | |
| **History** | ❌ | ✅ | Append-only event log — no age dimension. |
| **Working memory** | — | — | No user policy; deleted only by cascade, orphan sweep, or an opt-in TTL (below). |

**Disabled by default and opt-in.** The kill switch `plugins.ml_commons.memory.retention_enabled` defaults to `false`. While off, the container APIs reject a `retention_policy` and the job exits early — upgrading deletes nothing.

**Stacked on PR1 (#4914)**, which introduced the data model, the `pinned` field, and the per-container `retention_policy` API. Review/merge #4914 first. Implements RFC #4859.

## What gets deleted, and when

For a memory type with a policy, the job computes the set of documents to delete each run. **A document is deleted if it matches _either_ the age rule _or_ the count rule** — the two sets are combined and de-duplicated by document `_id`, so a document caught by both rules is only deleted once, and a document caught by just one rule is still deleted.

**Age rule (`retention_days`)** — delete documents older than `now − retention_days`, measured against `last_updated_time` (last activity, not creation).

**Count rule (`max_count`)** — if the type holds more than `max_count` non-pinned documents, evict the oldest excess (`count − max_count` documents).

Concretely, if a container's session policy is `{retention_days: 30, max_count: 100}` and it holds 150 sessions:

- The age rule selects every session idle for >30 days.
- The count rule selects the 50 least-recently-active sessions.
- The delete set is the **union** of those two, de-duplicated. A session that is both idle >30 days *and* in the oldest 50 is deleted once, not twice.

**Pinned documents are exempt** from both rules (they count toward neither the age scan nor the count total). If pinned documents alone exceed `max_count`, the container can grow past the cap; a warning is logged.

### Setting, changing, and clearing a policy

A `retention_policy` is set per container via the create/update API (#4914). Because rules are nullable-integer fields, `null` — not `-1` or `0` — is how you turn retention off, at two granularities:

- **Clear the whole policy** — send `"retention_policy": null`. The container is marked as an explicit opt-out and Pass 1 skips it entirely. (This wipe is allowed even when the feature flag is off.)
- **Clear one dimension, keep the other** — send that field explicitly as `null`, e.g. `{"sessions": {"retention_days": null, "max_count": 100}}` drops the age rule but keeps the count cap. Fields you *omit* from an update are preserved; only fields sent explicitly as `null` are cleared.

`retention_days` and `max_count` must be a positive integer or `null` — sending `-1` or `0` is rejected with a validation error. `-1` is only the "unset" sentinel at the cluster-settings layer (`default_*`, `working_memory_ttl_days`), where a setting cannot be null; in the per-container policy API, "off" is always `null`.

### Ordering for count-based eviction

Eviction deletes the *oldest* documents first, but "oldest" is measured differently per type:

| Type | Count-eviction sort | Rationale |
|---|---|---|
| **Sessions** | `last_updated_time` (LRU) | Living records — keep the most recently active, drop the stale. |
| **Long-term** | `last_updated_time` (LRU) | Same — mutable records. |
| **History** | `created_time` | Append-only; a row is never updated, so creation time *is* its age. |

Collection (`collectOldestDocIdsPage`) sorts `(timeField ASC, _id ASC)` and pages via `search_after` on that composite key, page size 100, capped at `COUNT_BASED_ID_CAP = 50000` deletions per container per run; any remainder is handled on the next run. The `_id` tiebreaker makes equal timestamps deterministic.

### Working memory

Working memory has no user-facing policy. It is deleted only by:

1. **Cascade** — when its parent session is deleted (sessions stage).
2. **`working_memory_ttl_days` TTL** — but **only** for containers with `disable_session=true`, and **only when an operator sets a positive value** (default `-1` = off, see below).
3. **Orphan sweep** — working memory whose `session_id` has no matching session doc, older than `orphan_ttl_days` (default 7).
4. **Empty-namespace sweep** — working memory with a fully empty namespace, older than `orphan_ttl_days`.

### Entry guards (`run()`)

`run()` returns early, in order, if:

1. **Multi-tenancy** is enabled (`plugins.ml_commons.multi_tenancy_enabled=true`) — retention does not support multi-tenant clusters.
2. **Kill switch** is off (`retention_enabled=false`, the default).
3. **A prior run is still in flight** — an `AtomicBoolean` CAS single-flight guard, cleared on every completion/error path so a crashed run cannot wedge the job.

Interval: `retention_job_interval_hours` (default 24, range 1–168).

### Two passes over the container index

Both passes paginate `ML_MEMORY_CONTAINER_INDEX` by `_id` with `search_after`, page size 100.

**Pass 1 — per-container retention** (`resolveContainersWithPolicies`). For each container:

- **Has a non-empty `retention_policy`** → run the retention pipeline.
- **No policy, but explicitly set `retention_policy: null`** → opted out; skip.
- **No policy, no explicit opt-out** → `applyDefaultRetentionPolicy` seeds a policy from cluster defaults, but only if an admin set at least one `default_*` setting `> 0`. With all defaults at `-1` (unset), nothing is written and the container is skipped. The backfill is a conditional (if-absent) script update, so a concurrent user write always wins.

The pipeline runs four stages strictly in order: **sessions → long-term → history → working-memory TTL**. Between containers that performed deletions, the next is scheduled after `retention_job_throttle_seconds` (default 5, range 1–60) on the `GENERIC` threadpool; containers that deleted nothing chain immediately.

**Pass 2 — orphan sweep** (`executeOrphanSweep`), after Pass 1 completes. Selects only containers that **have a `retention_policy`** and **do not set `disable_session=true`**:

```
.must(existsQuery("configuration.retention_policy"))
.mustNot(termQuery("configuration.disable_session", true))
```

So a stock cluster (no policies) sweeps nothing. The sweep also short-circuits if the working-memory index is absent (nothing to do) or the sessions index is absent (would misclassify all working memory as orphaned).

## Two gates before any deletion

Nothing is deleted unless **both** are satisfied:

1. `retention_enabled=true` (operator flips the kill switch), **and**
2. A per-container `retention_policy` exists — set directly via the #4914 API, or seeded from a cluster default `> 0`.

On a stock cluster with the feature turned on but all defaults at `-1` and no per-container policies, Pass 1 skips every container and Pass 2 matches none: **zero deletions.**

## Hardening details

**Epoch-seconds legacy timestamps.** Documents written before the epoch-millis fix stored timestamps in epoch *seconds*, which would satisfy a raw `lt(cutoffMillis)` range and look expired. Two guards:

- **Sessions `retention_days`** (`identifyTimeBasedExpiredSessions`) uses `rangeQuery.lt(cutoffMillis)` as a pre-filter, then re-verifies each candidate client-side via `normalizeTimestamp` before marking it expired.
- **Long-term `retention_days`, working-memory TTL, and orphan/empty-namespace sweeps** build the range with `buildEpochAwareCutoffQuery`, normalizing at the query level (direct `DeleteByQuery`, no per-doc re-check).

Count-based sorting compares raw stored timestamps with no normalization; this is still order-preserving because epoch-seconds values (`< 1e10`) always sort below epoch-millis values.

**First-observation grace for the orphan sweep.** The first time the sweep sees a container, it stamps an `orphan_sweep_baseline_time` field (write-once conditional script) and defers all orphan deletion for one full `orphan_ttl_days` window instead of deleting immediately. This prevents working memory that predates the feature — or that was written via paths creating no session doc — from being wiped on the first run. Deletion is delayed by one window, never permanently skipped. Adds `orphan_sweep_baseline_time` to the container mapping (`schema_version` → 3).

**Session upsert prevents false orphans.** When a client supplies its own `session_id`, the add path now writes the backing session doc first via an idempotent `CREATE` (swallowing `VersionConflictEngineException`) before indexing working memory, so working memory always has a real parent session. A genuine session-write failure aborts the add rather than creating orphan-eligible working memory. Also fixes a latent NPE on a null `owner_id` when security is disabled.

## Cluster settings

All settings are node-scoped, dynamic, and prefixed `plugins.ml_commons.` (e.g. `plugins.ml_commons.memory.retention_enabled`).

| Setting | Default | Range | Purpose |
|---|---|---|---|
| `memory.retention_enabled` | `false` | boolean | Kill switch. When `false`, container APIs reject a `retention_policy` and the job does nothing. |
| `memory.retention_job_interval_hours` | `24` | 1–168 | How often the job runs. |
| `memory.retention_job_throttle_seconds` | `5` | 1–60 | Pause after a container that deleted data, before scheduling the next. |
| `memory.orphan_ttl_days` | `7` | 1–365 | Age threshold for the orphan/empty-namespace sweeps. |
| `memory.working_memory_ttl_days` | `-1` (off) | -1–365 | TTL for working memory in `disable_session=true` containers. `-1` = off: working memory is kept indefinitely. Set `> 0` to age it out. |
| `memory.default_session_retention_days` | `-1` | -1–3650 | Default session age cutoff. `-1` = unset. |
| `memory.default_session_max_count` | `-1` | -1–1000000 | Default session count cap. `-1` = unset. |
| `memory.default_long_term_max_count` | `-1` | -1–1000000 | Default long-term count cap. `-1` = unset. |
| `memory.default_history_max_count` | `-1` | -1–10000000 | Default history count cap. `-1` = unset. |

The four `default_*` settings only *seed* a policy onto a container that has none and has not opted out. `-1` means "unset — apply no default for this dimension" (not "delete immediately"). A default policy is written only if at least one is `> 0`.

**`working_memory_ttl_days` is off by default.** It is the *only* mechanism that ages out working memory in a `disable_session=true` (sessionless) container — those containers have no sessions, so cascade never fires and the orphan/empty-namespace sweeps (which target `disable_session=false` containers) never touch them. Defaulting to `-1` means a sessionless container keeps its working memory indefinitely, treating it as a durable store. An operator who wants sessionless working memory aged out must set a positive value. To turn it back off after setting one, reset it to `-1` (or send `null` to the cluster settings API, which reverts it to the `-1` default). Session-backed containers ignore this setting entirely; their working memory is governed by cascade and the orphan sweeps.

## Net effect on upgrade

| Cluster state | Result |
|---|---|
| `retention_enabled=false` (default) | Job exits immediately; nothing deleted. |
| `retention_enabled=true`, no per-container policy, all `default_*` = `-1` | Pass 1 skips every container; Pass 2 matches none; nothing deleted. |
| A cluster default or per-container policy is set | Retention applies to that container's **entire history** on the next run — see below. First orphan sweep of each container is stamp-and-skip; client-supplied-session working memory is protected by the session upsert. |

Deletion is opt-in per container, but it is **not** limited to data written after the policy was set. Once a container has a policy, the pipeline scans all of that container's documents (`matchAll` on `containerId`, minus pinned) with no lower age bound, so applying a policy to an existing container deletes its backlog that violates the rules on the next job run — e.g. a container with 10,000 sessions and a new `max_count: 100` policy has the 9,900 least-recently-active sessions evicted. What is *not* retroactive: the policy is never auto-applied on upgrade (an operator must set it or configure a cluster default), and the first orphan sweep of a container is deferred one `orphan_ttl_days` window.


common/src/main/java/org/opensearch/ml/common/model/ModelGuardrail.java
```diff
@@ -53,7 +53,7 @@ public class ModelGuardrail extends Guardrail {
     public static final String MODEL_ID_FIELD = "model_id";
     public static final String RESPONSE_FILTER_FIELD = "response_filter";
     public static final String RESPONSE_VALIDATION_REGEX_FIELD = "response_validation_regex";
-    public static final long ML_GUARDRAIL_TIMEOUT_IN_SECONDS = 60;
+    public static final long ML_GUARDRAIL_TIMEOUT_IN_SECONDS = 5;
```
nathaliellenaa: Just curious, why we change the timeout here?

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -0,0 +1,1723 @@
...
+            client.search(countRequest, ActionListener.wrap(countResponse -> {
+                long totalCount = countResponse.getHits().getTotalHits().value();
+                if (totalCount <= maxCount) {
+                    listener.onResponse(new HashSet<>());
+                    return;
+                }
+
+                int excess = (int) (totalCount - maxCount);
+                // Step 2: search for oldest excess sessions by created_time ASC
+                collectOldestDocIds(sessionIndex, containerId, CREATED_TIME_FIELD, excess, listener);
```
nathaliellenaa: This collects every excess ID (`totalCount - maxCount`) into a `HashSet` before deleting, with no per-run limit (the orphan path caps at 50k). On a large index with a small `maxCount` this holds millions of IDs on the job thread, risking heap exhaustion. Can we add a per-run cap or switch to delete-by-query with `max_docs`?

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -0,0 +1,1723 @@
...
+        DeleteByQueryRequest dbq = new DeleteByQueryRequest(workingIndex);
+        dbq.setIndicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN);
+        // Epoch-aware cutoff: legacy docs store created_time in epoch seconds and would all match
+        // a plain lt(cutoffMillis) range. See normalizeTimestamp().
+        dbq
+            .setQuery(
+                QueryBuilders
+                    .boolQuery()
+                    .must(QueryBuilders.termQuery(MEMORY_CONTAINER_ID_FIELD, containerId))
+                    .must(buildEpochAwareCutoffQuery(CREATED_TIME_FIELD, cutoffMillis))
```
jiapingzeng, line 1230: Seems that the query here is based on created time, should it use last updated time instead?

nathaliellenaa, line 1230: TTL is measured from `created_time`, so a doc created 31 days ago but updated yesterday gets deleted. Is this intended? The long-term path already uses `last_updated_time` (line 886).

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -0,0 +1,1723 @@
...
+                }
+
+                Set<String> existingIds = new HashSet<>();
+                for (SearchHit hit : response.getHits().getHits()) {
+                    existingIds.add(hit.getId());
+                }
+
+                for (String sessionId : batch) {
+                    if (!existingIds.contains(sessionId)) {
+                        allOrphanIds.add(sessionId);
```
nathaliellenaa, line 1793: `add-memories` neither creates nor validates a session doc when the client supplies its own `session_id` ([createNewSessionIfAbsent](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java#L156)), so if `create-session` was never called, that working memory has no backing session doc and the sweep unable to tell "never created" from "deleted", evicts it after `orphan_ttl_days`. Can we lazily-create the session on write, or gate the sweep to IDs that had a session doc?

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -0,0 +1,1723 @@
...
+                    }
+                }
```
jiapingzeng, line 1795: Seems all docs without a matching session doc are considered orphaned, could there be working memory that doesn't have a session doc? Should they be handled separately?

In managed service we always create a session doc but I don't think it is guaranteed in open source.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -0,0 +1,1723 @@
...
+                containerId,
+                workingIndex,
+                cutoffMillis
+            );
+
+        DeleteByQueryRequest dbq = new DeleteByQueryRequest(workingIndex);
+        dbq.setIndicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN);
+        // Epoch-aware cutoff: legacy docs store created_time in epoch seconds and would all match
+        // a plain lt(cutoffMillis) range. See normalizeTimestamp().
+        dbq
```
jiapingzeng, line 1868: Same as above

jiapingzeng (commented):
Overall good job! Added some comments

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Working-memory TTL uses created_time on purpose to guarantee a hard maximum lifespan. If it used last_updated_time, a client that touches a doc periodically would keep resetting its expiry clock, and the doc could live indefinitely — which defeats the point of a working-memory TTL. Measuring from created_time means a working-memory doc lives at most working_memory_ttl_days, regardless of edits.

Long-term memory intentionally does the opposite (last_updated_time, line 886), because long-term memories are meant to be reinforced/updated over time, so "recently updated" is the signal we want to keep them alive. We still use created_time for max_count.

common/src/main/java/org/opensearch/ml/common/model/ModelGuardrail.java
AirFunnyB: Tightened from the conservative 60s I pushed in #4904 to 5s — this is a lightweight guardrail check, not a long-running operation, and holding the job thread for a minute on a hung call defeats the purpose of having a bound. Happy to adjust as any where between 60-5 should work

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Good catch — the deletion was already batched (500/batch) but the ID collection was unbounded, so a large totalCount - maxCount backlog could balloon the heap on the job thread. Fixed by clamping at a single choke point in collectOldestDocIds: COUNT_BASED_ID_CAP = 50_000, mirroring the existing orphan-sweep cap, so all three count-based paths (session/long-term/history) are covered by one change. We evict the oldest 50k per run and log a warn; because the query sorts created_time ASC, _id ASC, each run removes the globally-oldest slice and the backlog converges over subsequent scheduled runs. Added testCollectOldestDocIdsCapsAtCountBasedCap to guard it.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Working-memory TTL uses created_time on purpose to guarantee a hard maximum lifespan. If it used last_updated_time, a client that touches a doc periodically would keep resetting its expiry clock, and the doc could live indefinitely — which defeats the point of a working-memory TTL. Measuring from created_time means a working-memory doc lives at most working_memory_ttl_days, regardless of edits.

Long-term memory intentionally does the opposite (last_updated_time, line 886), because long-term memories are meant to be reinforced/updated over time, so "recently updated" is the signal we want to keep them alive. We still use created_time for max_count.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: > Seems all docs without a matching session doc are considered orphaned, could there be working memory that doesn't have a session doc? Should they be handled separately?
> 
> In managed service we always create a session doc but I don't think it is guaranteed in open source.

> `add-memories` neither creates nor validates a session doc when the client supplies its own `session_id` ([createNewSessionIfAbsent](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java#L156)), so if `create-session` was never called, that working memory has no backing session doc and the sweep unable to tell "never created" from "deleted", evicts it after `orphan_ttl_days`. Can we lazily-create the session on write, or gate the sweep to IDs that had a session doc?

This is a real data-loss bug. If someone adds memories with their own session_id but never actually creates that session, there's no session record for it. The cleanup job can't tell the difference between "this session was never created" and "this session was deleted," so it would wrongly treat the memory as abandoned and delete it after orphan_ttl_days.

I fixed this two ways:

1. Create the session automatically when memory is added. Now, when someone adds memory with their own session_id, we create the backing session record first, before saving the memory. If that fails, the whole request fails instead of silently leaving memory that could get deleted later. So anything written from now on is safe.

2. Give each container a grace period before any cleanup. The first time the cleanup job looks at a container, it just records the date and skips deletion that round. It waits one full orphan_ttl_days window before it will delete anything. This protects memory that was written before this fix shipped — add-memories has been around since 3.2, so there's real old data out there — from being wiped on the very first cleanup after retention is turned on.

On backfilling old data (going back and creating session records for existing memory): I looked into it and decided against it, on purpose. When a user deletes a session, we don't also delete its memory — cleaning that up is exactly what this job is for. And nothing in the data tells us why a session record is missing. So old session-less memory could mean "never created" or "the user deleted this on purpose." A backfill can't tell them apart, so it would end up recreating sessions the user deliberately deleted and making that memory impossible to ever clean up — which is worse than the problem we started with.

What's left (and I'm okay with it): old, pre-existing memory with no session record only gets delayed by the grace period, not protected forever — after one window it'll be cleaned up. This is limited and shrinks over time, since all new memory now creates its session automatically. The clean way to fully solve it later would be to start recording whether a session was actually deleted — but that's a future data-model change, not something a backfill can fix.

common/src/main/java/org/opensearch/ml/common/model/ModelGuardrail.java
mingshl: is the memory change related to model guardrail?

AirFunnyB: No the guardrail is not related to the retention policy. Should I raise as a separate PR

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
jiapingzeng: But if a client is periodically touching a doc, it means they are still using that memory and that we should not delete it, right? 

If we're going with created time for working memory and last updated time for long term memory as a design decision, let's make sure to document it once you work on documentation.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -546,9 +546,9 @@ public void validate(List<String> value) {
             + ML_COMMONS_REMOTE_AGENTIC_MEMORY_ENABLED.getKey();
 
     // Feature flag for memory retention. Gates acceptance of retention_policy on the memory container APIs
-    // and (in a later PR) the memory retention job. Provides a cluster-level kill switch for the feature.
+    // and the memory retention job. Provides a cluster-level kill switch for the feature.
     public static final Setting<Boolean> ML_COMMONS_MEMORY_RETENTION_ENABLED = Setting
-        .boolSetting(ML_PLUGIN_SETTING_PREFIX + "memory.retention_enabled", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
+        .boolSetting(ML_PLUGIN_SETTING_PREFIX + "memory.retention_enabled", true, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
jiapingzeng: Let's keep the feature disabled by default

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Hey yes. WM will remain created_time for the reason mentioned above. Also note WM does not have its own retention policy. LM will be last_updated_time for retention_days but for max_count I'm going to change created_time to last_updated_time for max_count. The rationale behind this is exactly the session liveliness you were talking about — this will sacrifice the simple "oldest created gets deleted first" model, but I'll make the documentation clear on how last_updated_time is updated and how the logic works.

Current set up:

- Working memory → created_time (hard max lifespan on raw data; cleaned up by orphan/empty-namespace sweeps + cascade)
- Sessions & long-term retention_days → last_updated_time (recency / reinforcement)
- Sessions & long-term max_count → last_updated_time (LRU — just changed based on this thread)
- History max_count → created_time (append-only event log)

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
AirFunnyB: ok

common/src/test/java/org/opensearch/ml/common/settings/MLFeatureEnabledSettingTests.java
```diff
@@ -215,12 +215,12 @@ public void testAgenticMemoryCanBeDisabled() {
     }
 
     @Test
-    public void testMemoryRetentionDisabledByDefault() {
+    public void testMemoryRetentionEnabledByDefault() {
```
jiapingzeng: this test needs to be updated too

common/src/test/java/org/opensearch/ml/common/settings/MLFeatureEnabledSettingTests.java
```diff
@@ -233,14 +233,15 @@ public void testMemoryRetentionCanBeDisabled() {
 
     @Test
     public void testMemoryRetentionDynamicUpdate() {
```
jiapingzeng, line 235: Same here

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
```diff
@@ -162,6 +171,40 @@ private MLMemoryContainer buildMemoryContainer(MLCreateMemoryContainerInput inpu
             }
         }
 
+        // Auto-apply admin-configured default retention policy when user did not provide one
+        if (configuration != null && !configuration.isRetentionPolicyExplicitlyNull()) {
```
jiapingzeng: Should we also guard this by the feature flag?

plugin/src/test/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessorTests.java
```diff
@@ -0,0 +1,3025 @@
...
+        }).when(client).search(any(SearchRequest.class), isA(ActionListener.class));
+
+        processor.run();
+
+        // Verify container search is initiated
+        verify(client, atLeastOnce()).search(any(SearchRequest.class), isA(ActionListener.class));
+    }
+
+    @Test
+    public void testRunExecutesWithDefaultSettings() {
```
jiapingzeng, line 229: Need to enable feature flag here

jiapingzeng (commented):
A few more changes needed as a part of feature flag change

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportCreateMemoryContainerAction.java
AirFunnyB: Yes good call!

jiapingzeng (approved):
LGTM!

Issue #4915, opened 2026-07-19
[RFC] GCP Vertex AI Connector Support


tonypiazza:
## Problem Statement

ml-commons ships first-class remote-connector support for AWS: the `aws_sigv4` auth strategy signs each outbound request, and there are ready-made blueprints for Amazon Bedrock and SageMaker under `docs/remote_inference_blueprints/`.

There is no equivalent for **Google Cloud Vertex AI**. Two gaps:

1. **No auth strategy.** The connector framework's auth helpers are AWS-centric (SigV4). The generic `http` protocol can inject `${credential.api_key}` into an `Authorization` header, but Google Cloud APIs authenticate with **OAuth2 access tokens that expire roughly hourly**. A static token pasted into a connector's `credential` block stops working after ~1 hour and must be rotated by hand — not viable for a durable connector.
2. **Limited, workaround-based blueprints.** The one existing Vertex AI blueprint (text embeddings) relies on the manual static-token approach above, and there is no coverage for Gemini text generation, streaming, or batch. See "Current state" below.

This is the same class of problem `aws_sigv4` solves for AWS: turn long-lived credentials into a fresh, per-request authorization automatically. Google needs the analogous strategy.

### Current state (baseline this RFC replaces)

There is one existing Vertex AI blueprint — `docs/remote_inference_blueprints/gcp_vertexai_connector_embedding_blueprint.md` — and it demonstrates the problem rather than solving it. It uses the generic `http` protocol with a **static, manually generated access token** pasted into `credential.vertexAI_token` (via `gcloud auth print-access-token`) and injected through `${credential.vertexAI_token}`. Because the token is short-lived, the blueprint's own **Step 8 ("Update the access token")** instructs users to refresh it by hand and call the update-model API each time it expires — "Applications are responsible for refreshing this token as needed."

That blueprint is effectively the manual-token workaround. This RFC proposes making token minting/refresh a first-class capability so users no longer have to do it themselves. (The separate `google_gemini_connector_chat_blueprint.md` targets the Google AI Studio Gemini API with an `x-goog-api-key` and is a different product/auth model, not Vertex AI.)

## Proposed Solution

Add a new **`google_cloud`** connector protocol that mints and refreshes OAuth2 access tokens on behalf of the connector, mirroring the existing `aws_sigv4` pattern, plus a set of Vertex AI blueprints.

### New components

[...]

PR #4921, opened 2026-07-22
[FEATURE] Add GCP Vertex AI connector via google_cloud auth strategy

tonypiazza:
### Description

Adds a first-class `google_cloud` connector protocol so OpenSearch can call GCP Vertex AI without a manually rotated access token. The plugin mints and refreshes GCP OAuth2 tokens itself, mirroring the existing `aws_sigv4` strategy.

Implements #4915.

### What's included

**Core `google_cloud` connector** (`common` + `ml-algorithms`)
- New `GOOGLE_CLOUD` protocol.
- `GoogleCloudConnector` supporting two credential modes:
  - Service-account key (`private_key` + `client_email`)
  - ADC / Workload Identity (`auth_mode: adc`, no credentials)
- `GoogleCredentialProvider` — wraps `google-auth-library`, mints/refreshes tokens, applies scopes.
- `GoogleConnectorExecutor` — HTTP execution plus per-request bearer-token injection; supports predict, batch, and streaming.
- Streaming via Vertex `streamGenerateContent` (SSE), reusing `HttpStreamingHandler` with a Gemini frame parser.

**Vertex AI batch status/cancel** (shared framework — see note below)
- Adds a Vertex case to the `ConnectorUtils` batch status/cancel URL derivation.
- Propagates the trusted-connector-endpoints regex to the batch task executors (previously unset on those paths).

**Blueprints** — Gemini generation, embeddings, streaming, and batch. The legacy `http`-protocol manual-token embedding blueprint is marked as no longer recommended.

**Dependency** — adds `com.google.auth:google-auth-library-oauth2-http` (ml-algorithms) and bundles `com.google.guava:failureaccess` in the plugin (required at runtime). Flagging for maintainer/security review.

### Testing

- Unit tests across `common`, `ml-algorithms`, and `plugin`; full build and spotless pass.
- Gemini SSE parser unit-tested with real captured Vertex frames.
- Manually verified end-to-end against live Vertex AI (`gemini-2.5-flash`, `text-embedding-004`, `us-central1`):
  - predict — service-account key
  - predict — ADC
  - text embeddings
  - batch lifecycle: submit -> status -> cancel (confirmed `JOB_STATE_CANCELLED`)
  - streaming SSE transport and parsing

### Notes for reviewers

- The batch status/cancel change touches shared, non-Google code (`GetTaskTransportAction`, `CancelBatchJobTransportAction`, `ConnectorUtils`). It is isolated in the second commit. The regex propagation is arguably a latent-bug fix (the setting was silently unset on those paths). Happy to split this into a separate PR if preferred.
- Existing AWS (Bedrock/SageMaker), HTTP, and MCP behavior is unchanged — all changes are additive and the existing suites pass.

### Check List

- [x] New functionality has been documented (blueprints).

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
@@ -49,8 +49,11 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String COHERE = "cohere";
     public static final String BEDROCK = "bedrock";
     public static final String SAGEMAKER = "sagemaker";
+    // Matches the Vertex AI host (e.g. us-central1-aiplatform.googleapis.com) for batch job status/cancel.
+    public static final String VERTEX_AI = "aiplatform";
     public static final String SAGEMAKER_PRE_POST_FUNC_TEXT = "default";
-    public static final List<String> SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES = List.of(SAGEMAKER, OPENAI, BEDROCK, COHERE);
+    public static final List<String> SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES = List
+        .of(SAGEMAKER, OPENAI, BEDROCK, COHERE, VERTEX_AI);
 
     private static final String INBUILT_FUNC_PREFIX = "connector.";
     private static final String PRE_PROCESS_FUNC = "PreProcessFunction";
--- a/common/src/main/java/org/opensearch/ml/common/connector/ConnectorProtocols.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/ConnectorProtocols.java
@@ -12,21 +12,22 @@ public class ConnectorProtocols {
 
     public static final String HTTP = "http";
     public static final String AWS_SIGV4 = "aws_sigv4";
+    public static final String GOOGLE_CLOUD = "google_cloud";
     public static final String MCP_SSE = "mcp_sse";
     public static final String MCP_STREAMABLE_HTTP = "mcp_streamable_http";
 
-    public static final List<String> VALID_PROTOCOLS = Arrays.asList(AWS_SIGV4, HTTP, MCP_SSE, MCP_STREAMABLE_HTTP);
+    public static final List<String> VALID_PROTOCOLS = Arrays.asList(AWS_SIGV4, HTTP, GOOGLE_CLOUD, MCP_SSE, MCP_STREAMABLE_HTTP);
 
     public static void validateProtocol(String protocol) {
         if (protocol == null) {
-            throw new IllegalArgumentException(
-                "Connector protocol is null. Please use one of " + Arrays.toString(VALID_PROTOCOLS.toArray(new String[0]))
-            );
+            throw new IllegalArgumentException("Connector protocol is null. Please use one of " + supportedProtocols());
         }
         if (!VALID_PROTOCOLS.contains(protocol)) {
-            throw new IllegalArgumentException(
-                "Unsupported connector protocol. Please use one of " + Arrays.toString(VALID_PROTOCOLS.toArray(new String[0]))
-            );
+            throw new IllegalArgumentException("Unsupported connector protocol. Please use one of " + supportedProtocols());
         }
+    }
+
+    public static String supportedProtocols() {
+        return "[" + String.join(", ", VALID_PROTOCOLS) + "]";
     }
 }
--- a/common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
+++ b/common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.common.connector;
+
+import static org.opensearch.ml.common.connector.ConnectorProtocols.GOOGLE_CLOUD;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.common.io.stream.BytesStreamOutput;
+import org.opensearch.commons.authuser.User;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.ml.common.AccessMode;
+
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@org.opensearch.ml.common.annotation.Connector(GOOGLE_CLOUD)
+public class GoogleCloudConnector extends HttpConnector {
+
+    public static final String PRIVATE_KEY_FIELD = "private_key";
+    public static final String CLIENT_EMAIL_FIELD = "client_email";
+    public static final String TOKEN_URI_FIELD = "token_uri";
+    public static final String DEFAULT_TOKEN_URI = "https://oauth2.googleapis.com/token";
+
+    public static final String AUTH_MODE_FIELD = "auth_mode";
+    public static final String AUTH_MODE_ADC = "adc";
+    public static final String SCOPES_FIELD = "scopes";
+    public static final String DEFAULT_SCOPE = "https://www.googleapis.com/auth/cloud-platform";
+
+    public static final String AUTHORIZATION_HEADER = "Authorization";
+
+    // Runtime-only bearer token set by the executor before each request. Excluded from
+    // equals/hashCode and serialization: it is transient per-request state, not connector config.
+    @EqualsAndHashCode.Exclude
+    private transient String authorizationValue;
+
+    @Builder(builderMethodName = "googleCloudConnectorBuilder")
+    public GoogleCloudConnector(
+        String name,
+        String description,
+        String version,
+        String protocol,
+        Map<String, String> parameters,
+        Map<String, String> credential,
+        List<ConnectorAction> actions,
+        List<String> backendRoles,
+        AccessMode accessMode,
+        User owner,
+        ConnectorClientConfig connectorClientConfig,
+        String tenantId,
+        String provisionedBy
+    ) {
+        super(
+            name,
+            description,
+            version,
+            protocol,
+            parameters,
+            credential,
+            actions,
+            backendRoles,
+            accessMode,
+            owner,
+            connectorClientConfig,
+            tenantId,
+            provisionedBy
+        );
+        validate();
+    }
+
+    public GoogleCloudConnector(String protocol, XContentParser parser) throws IOException {
+        super(protocol, parser);
+        validate();
+    }
+
+    public GoogleCloudConnector(StreamInput input) throws IOException {
+        super(input);
+        validate();
+    }
+
+    public GoogleCloudConnector(String protocol, StreamInput input) throws IOException {
+        super(protocol, input);
+        // Validation skipped for StreamInput deserialization (mirrors AwsConnector).
+    }
+
+    private boolean isAdcMode() {
+        return parameters != null && AUTH_MODE_ADC.equalsIgnoreCase(parameters.get(AUTH_MODE_FIELD));
+    }
+
+    private void validate() {
+        if (isAdcMode()) {
+            // ADC mode: no service-account credentials expected. Reject a mix.
+            if (credential != null && (credential.containsKey(PRIVATE_KEY_FIELD) || credential.containsKey(CLIENT_EMAIL_FIELD))) {
+                throw new IllegalArgumentException("auth_mode=adc must not include service-account credentials (private_key/client_email)");
+            }
+            return;
+        }
+        // Service-account key mode.
+        if (credential == null || !credential.containsKey(PRIVATE_KEY_FIELD) || !credential.containsKey(CLIENT_EMAIL_FIELD)) {
+            throw new IllegalArgumentException("Missing credential");
+        }
+    }
+
+    @Override
+    public Connector cloneConnector() {
+        try (BytesStreamOutput bytesStreamOutput = new BytesStreamOutput()) {
+            this.writeTo(bytesStreamOutput);
+            StreamInput streamInput = bytesStreamOutput.bytes().streamInput();
+            return new GoogleCloudConnector(streamInput);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public boolean useAdc() {
+        return isAdcMode();
+    }
+
+    public String getPrivateKey() {
+        return decryptedCredential.get(PRIVATE_KEY_FIELD);
+    }
+
+    public String getClientEmail() {
+        return decryptedCredential.get(CLIENT_EMAIL_FIELD);
+    }
+
+    public String getTokenUri() {
+        String uri = decryptedCredential.get(TOKEN_URI_FIELD);
+        return uri != null ? uri : DEFAULT_TOKEN_URI;
+    }
+
+    public String getScopes() {
+        if (parameters == null || parameters.get(SCOPES_FIELD) == null) {
+            return DEFAULT_SCOPE;
+        }
+        return parameters.get(SCOPES_FIELD);
+    }
+
+    /**
+     * Set the bearer token the executor minted for the current request. Applied
+     * programmatically (not via user ${parameters.*} substitution) so the framework's
+     * blocked-dynamic-header guard on Authorization is not bypassed.
+     */
+    public void setAuthorizationValue(String authorizationValue) {
+        this.authorizationValue = authorizationValue;
+    }
+
+    @Override
+    public Map<String, String> getHeadersWithRuntimeParameters(Map<String, String> runtimeParameters) {
+        Map<String, String> headers = super.getHeadersWithRuntimeParameters(runtimeParameters);
+        if (authorizationValue == null) {
+            return headers;
+        }
+        Map<String, String> withAuth = headers == null ? new HashMap<>() : new HashMap<>(headers);
+        withAuth.put(AUTHORIZATION_HEADER, authorizationValue);
+        return withAuth;
+    }
+}
--- a/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
@@ -11,6 +11,7 @@ import static org.opensearch.ml.common.CommonValue.TENANT_ID_FIELD;
 import static org.opensearch.ml.common.CommonValue.VERSION_2_19_0;
 import static org.opensearch.ml.common.CommonValue.VERSION_3_0_0;
 import static org.opensearch.ml.common.CommonValue.VERSION_3_7_0;
+import static org.opensearch.ml.common.connector.ConnectorProtocols.GOOGLE_CLOUD;
 import static org.opensearch.ml.common.connector.ConnectorProtocols.MCP_SSE;
 import static org.opensearch.ml.common.connector.ConnectorProtocols.MCP_STREAMABLE_HTTP;
 import static org.opensearch.ml.common.utils.StringUtils.getParameterMap;
@@ -115,7 +116,9 @@ public class MLCreateConnectorInput implements ToXContentObject, Writeable {
                 throw new IllegalArgumentException("Connector protocol is null");
             }
             boolean isMcpConnector = (protocol.equals(MCP_SSE) || protocol.equals(MCP_STREAMABLE_HTTP));
-            if ((credential == null || credential.isEmpty()) && !isMcpConnector) {
+            // google_cloud connectors may omit credentials when using ADC / Workload Identity.
+            boolean allowsEmptyCredential = isMcpConnector || protocol.equals(GOOGLE_CLOUD);
+            if ((credential == null || credential.isEmpty()) && !allowsEmptyCredential) {
                 throw new IllegalArgumentException("Connector credential is null or empty list");
             }
             if (actions != null) {
--- a/docs/remote_inference_blueprints/gcp_vertexai_batch_blueprint.md
+++ b/docs/remote_inference_blueprints/gcp_vertexai_batch_blueprint.md
@@ -54,20 +54,15 @@ POST /_plugins/_ml/connectors/_create
                 "Content-Type": "application/json"
             },
             "request_body": "{\"displayName\":\"${parameters.job_name}\",\"model\":\"publishers/google/models/${parameters.model}\",\"inputConfig\":{\"instancesFormat\":\"jsonl\",\"gcsSource\":{\"uris\":[\"${parameters.input_uri}\"]}},\"outputConfig\":{\"predictionsFormat\":\"jsonl\",\"gcsDestination\":{\"outputUriPrefix\":\"${parameters.output_uri}\"}}}"
-        },
-        {
-            "action_type": "batch_predict_status",
-            "method": "GET",
-            "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/batchPredictionJobs/${parameters.job_id}"
-        },
-        {
-            "action_type": "cancel_batch_predict",
-            "method": "POST",
-            "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/batchPredictionJobs/${parameters.job_id}:cancel"
         }
     ]
 }
 ```
+
+Only the `batch_predict` action is defined. ml-commons derives the status and cancel calls
+from it automatically (using the Vertex batch job's resource `name` returned at submit time),
+so you do **not** declare separate `batch_predict_status` / `cancel_batch_predict` actions —
+doing so would leave an unresolved `${parameters.job_id}` in the derived URL.
 
 For ADC / Workload Identity mode, set `"auth_mode": "adc"` in `parameters` and leave
 `credential` empty (`{}`).
@@ -90,24 +85,20 @@ POST /_plugins/_ml/models/<MODEL_ID>/_batch_predict
 }
 ```
 
+Submitting returns a `task_id`. Status and cancel are performed against that task via the
+ML task APIs below (not a `_predict/...` route).
+
 ## 5. Check job status
 
 ```json
-POST /_plugins/_ml/models/<MODEL_ID>/_predict/batch_predict_status
-{
-    "parameters": {
-        "job_id": "<BATCH_PREDICTION_JOB_ID>"
-    }
-}
+GET /_plugins/_ml/tasks/<TASK_ID>
 ```
+
+The response includes the refreshed remote job state under `remote_job.state` (e.g.
+`JOB_STATE_PENDING`, `JOB_STATE_RUNNING`, `JOB_STATE_SUCCEEDED`, `JOB_STATE_CANCELLED`).
 
 ## 6. Cancel a job
 
 ```json
-POST /_plugins/_ml/models/<MODEL_ID>/_predict/cancel_batch_predict
-{
-    "parameters": {
-        "job_id": "<BATCH_PREDICTION_JOB_ID>"
-    }
-}
+POST /_plugins/_ml/tasks/<TASK_ID>/_cancel
 ```
--- a/docs/remote_inference_blueprints/gcp_vertexai_batch_blueprint.md
+++ b/docs/remote_inference_blueprints/gcp_vertexai_batch_blueprint.md
@@ -1,0 +1,113 @@
+# GCP Vertex AI Batch Inference Connector Blueprint
+
+This blueprint runs offline batch inference against Vertex AI `batchPredictionJobs` using the
+`google_cloud` connector protocol. OAuth2 tokens are minted and refreshed automatically; no
+`Authorization` header is added by you.
+
+It defines three actions:
+- `batch_predict` — submit a batch prediction job.
+- `batch_predict_status` — poll a job's status.
+- `cancel_batch_predict` — cancel a running job.
+
+Batch input and output are configured in the request body via GCS or BigQuery locations. See
+the [Vertex AI batch prediction docs](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/batch-prediction).
+
+## 1. Add Vertex AI endpoint to trusted URLs
+
+```json
+PUT /_cluster/settings
+{
+    "persistent": {
+        "plugins.ml_commons.trusted_connector_endpoints_regex": [
+            "^https://.*-aiplatform\\.googleapis\\.com/.*$"
+        ]
+    }
+}
+```
+
+## 2. Create the connector
+
+```json
+POST /_plugins/_ml/connectors/_create
+{
+    "name": "GCP Vertex AI Connector: batch inference",
+    "description": "Vertex AI batchPredictionJobs connector",
+    "version": 1,
+    "protocol": "google_cloud",
+    "parameters": {
+        "project_id": "<YOUR_PROJECT_ID>",
+        "location": "us-central1",
+        "model": "gemini-2.5-flash",
+        "scopes": "https://www.googleapis.com/auth/cloud-platform"
+    },
+    "credential": {
+        "private_key": "<YOUR_SERVICE_ACCOUNT_PRIVATE_KEY>",
+        "client_email": "<YOUR_SERVICE_ACCOUNT_CLIENT_EMAIL>",
+        "token_uri": "https://oauth2.googleapis.com/token"
+    },
+    "actions": [
+        {
+            "action_type": "batch_predict",
+            "method": "POST",
+            "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/batchPredictionJobs",
+            "headers": {
+                "Content-Type": "application/json"
+            },
+            "request_body": "{\"displayName\":\"${parameters.job_name}\",\"model\":\"publishers/google/models/${parameters.model}\",\"inputConfig\":{\"instancesFormat\":\"jsonl\",\"gcsSource\":{\"uris\":[\"${parameters.input_uri}\"]}},\"outputConfig\":{\"predictionsFormat\":\"jsonl\",\"gcsDestination\":{\"outputUriPrefix\":\"${parameters.output_uri}\"}}}"
+        },
+        {
+            "action_type": "batch_predict_status",
+            "method": "GET",
+            "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/batchPredictionJobs/${parameters.job_id}"
+        },
+        {
+            "action_type": "cancel_batch_predict",
+            "method": "POST",
+            "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/batchPredictionJobs/${parameters.job_id}:cancel"
+        }
+    ]
+}
+```
+
+For ADC / Workload Identity mode, set `"auth_mode": "adc"` in `parameters` and leave
+`credential` empty (`{}`).
+
+## 3. Register and deploy the model
+
+Register and deploy the model as in the
+[Gemini blueprint](./gcp_vertexai_gemini_blueprint.md) (steps 3–5), using this connector.
+
+## 4. Submit a batch prediction job
+
+```json
+POST /_plugins/_ml/models/<MODEL_ID>/_batch_predict
+{
+    "parameters": {
+        "job_name": "vertex-batch-2026-07-21",
+        "input_uri": "gs://<YOUR_BUCKET>/batch_input.jsonl",
+        "output_uri": "gs://<YOUR_BUCKET>/output/"
+    }
+}
+```
+
+## 5. Check job status
+
+```json
+POST /_plugins/_ml/models/<MODEL_ID>/_predict/batch_predict_status
+{
+    "parameters": {
+        "job_id": "<BATCH_PREDICTION_JOB_ID>"
+    }
+}
+```
+
+## 6. Cancel a job
+
+```json
+POST /_plugins/_ml/models/<MODEL_ID>/_predict/cancel_batch_predict
+{
+    "parameters": {
+        "job_id": "<BATCH_PREDICTION_JOB_ID>"
+    }
+}
+```
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
@@ -478,7 +478,14 @@ public class ConnectorUtils {
                 // Vertex batch responses carry a full resource name
                 // (projects/<num>/locations/<loc>/batchPredictionJobs/<id>). Status/cancel are
                 // addressed as {host}/v1/{name}[:cancel], so build from the endpoint's base URL.
-                String vertexBaseUrl = predictEndpoint.substring(0, predictEndpoint.indexOf("/v1/") + "/v1/".length());
+                int v1Index = predictEndpoint.indexOf("/v1/");
+                if (v1Index < 0) {
+                    throw new IllegalArgumentException(
+                        "Vertex AI batch_predict endpoint must contain a '/v1/' path segment to derive the status/cancel URL, got: "
+                            + predictEndpoint
+                    );
+                }
+                String vertexBaseUrl = predictEndpoint.substring(0, v1Index + "/v1/".length());
                 url = isCancelAction ? vertexBaseUrl + "${parameters.name}:cancel" : vertexBaseUrl + "${parameters.name}";
                 method = isCancelAction ? "POST" : "GET";
                 break;
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
@@ -12,6 +12,7 @@ import static org.opensearch.ml.common.connector.ConnectorAction.BEDROCK;
 import static org.opensearch.ml.common.connector.ConnectorAction.COHERE;
 import static org.opensearch.ml.common.connector.ConnectorAction.OPENAI;
 import static org.opensearch.ml.common.connector.ConnectorAction.SAGEMAKER;
+import static org.opensearch.ml.common.connector.ConnectorAction.VERTEX_AI;
 import static org.opensearch.ml.common.connector.HttpConnector.RESPONSE_FILTER_FIELD;
 import static org.opensearch.ml.common.connector.MLPreProcessFunction.CONVERT_INPUT_TO_JSON_STRING;
 import static org.opensearch.ml.common.connector.MLPreProcessFunction.PROCESS_REMOTE_INFERENCE_INPUT;
@@ -471,6 +472,14 @@ public class ConnectorUtils {
                 url = isCancelAction
                     ? predictEndpoint + "/${parameters.processedJobArn}/stop"
                     : predictEndpoint + "/${parameters.processedJobArn}";
+                method = isCancelAction ? "POST" : "GET";
+                break;
+            case VERTEX_AI:
+                // Vertex batch responses carry a full resource name
+                // (projects/<num>/locations/<loc>/batchPredictionJobs/<id>). Status/cancel are
+                // addressed as {host}/v1/{name}[:cancel], so build from the endpoint's base URL.
+                String vertexBaseUrl = predictEndpoint.substring(0, predictEndpoint.indexOf("/v1/") + "/v1/".length());
+                url = isCancelAction ? vertexBaseUrl + "${parameters.name}:cancel" : vertexBaseUrl + "${parameters.name}";
                 method = isCancelAction ? "POST" : "GET";
                 break;
             default:
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleConnectorExecutor.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleConnectorExecutor.java
@@ -1,0 +1,257 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.engine.algorithms.remote;
+
+import static org.opensearch.ml.common.connector.ConnectorProtocols.GOOGLE_CLOUD;
+import static org.opensearch.ml.engine.algorithms.agent.AgentUtils.LLM_INTERFACE_GEMINI_V1BETA_GENERATE_CONTENT;
+import static org.opensearch.ml.engine.algorithms.agent.MLChatAgentRunner.LLM_INTERFACE;
+import static software.amazon.awssdk.http.SdkHttpMethod.DELETE;
+import static software.amazon.awssdk.http.SdkHttpMethod.GET;
+import static software.amazon.awssdk.http.SdkHttpMethod.POST;
+import static software.amazon.awssdk.http.SdkHttpMethod.PUT;
+
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
+import java.time.Duration;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.regex.Pattern;
+
+import org.apache.commons.text.StringEscapeUtils;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.support.ThreadedActionListener;
+import org.opensearch.common.collect.Tuple;
+import org.opensearch.common.util.TokenBucket;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.ml.common.connector.Connector;
+import org.opensearch.ml.common.connector.GoogleCloudConnector;
+import org.opensearch.ml.common.exception.MLException;
+import org.opensearch.ml.common.httpclient.MLHttpClientFactory;
+import org.opensearch.ml.common.input.MLInput;
+import org.opensearch.ml.common.model.MLGuard;
+import org.opensearch.ml.common.output.model.ModelTensors;
+import org.opensearch.ml.common.transport.MLTaskResponse;
+import org.opensearch.ml.engine.algorithms.remote.streaming.StreamPredictActionListener;
+import org.opensearch.ml.engine.algorithms.remote.streaming.StreamingHandler;
+import org.opensearch.ml.engine.algorithms.remote.streaming.StreamingHandlerFactory;
+import org.opensearch.ml.engine.annotation.ConnectorExecutor;
+import org.opensearch.script.ScriptService;
+import org.opensearch.transport.StreamTransportService;
+import org.opensearch.transport.client.Client;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+import software.amazon.awssdk.core.internal.http.async.SimpleHttpContentPublisher;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.http.async.AsyncExecuteRequest;
+import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+
+@Log4j2
+@ConnectorExecutor(GOOGLE_CLOUD)
+public class GoogleConnectorExecutor extends AbstractConnectorExecutor {
+
+    @Getter
+    private GoogleCloudConnector connector;
+    @Setter
+    @Getter
+    private ScriptService scriptService;
+    @Setter
+    @Getter
+    private TokenBucket rateLimiter;
+    @Setter
+    @Getter
+    private Map<String, TokenBucket> userRateLimiterMap;
+    @Setter
+    @Getter
+    private Client client;
+    @Setter
+    @Getter
+    private MLGuard mlGuard;
+    @Setter
+    private volatile boolean connectorPrivateIpEnabled;
+    @Setter
+    private volatile List<Pattern> connectorTrustedPrivateEndpoints;
+    @Setter
+    private volatile List<Pattern> connectorRestrictedIpPatterns;
+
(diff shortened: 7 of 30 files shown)


tonypiazza:
Thanks — dispositions for the Code-Diff-Analyzer findings on `3d3f0fdc`:

**token_uri SSRF (medium) — fixed** in `d9186f5c6`. `token_uri` is now validated before use: it must be an `https` URL whose host is `oauth2.googleapis.com` or a `*.googleapis.com` endpoint. Non-Google hosts, non-https schemes, the metadata address, and suffix-spoofing lookalikes are rejected, with unit tests covering each case.

**ADC contacts the metadata server (low) — by design, now documented.** ADC / Workload Identity mode intentionally resolves credentials from the environment (including the GCE/GKE metadata server) — that is the purpose of the mode, and the target is fixed by the Google SDK, not user input. It is opt-in via `auth_mode: adc`. The Gemini blueprint now notes that this mode is intended for GCP-hosted nodes.

**New dependencies (high x2) — expected; flagged for maintainer/security review.** `com.google.auth:google-auth-library-oauth2-http:1.23.0` (official Google artifact on Maven Central) and bundling `com.google.guava:failureaccess:1.0.1` in the plugin (required at runtime by guava/google-auth; the prior global exclusion left it absent from the plugin classloader). These are called out in the PR description for maintainer verification of provenance/hash. Happy to provide any additional detail needed for sign-off.

tonypiazza:
@mingshl — this PR implements the `google_cloud` / Vertex AI connector from RFC #4915. It's ready for a look when you have time.

One thing that needs a maintainer decision: the `Code-Diff-Analyzer` check is failing solely on the **mandatory dependency-flagging policy**, not on a defect. The feature necessarily adds `com.google.auth:google-auth-library-oauth2-http` (official Google artifact) and bundles `com.google.guava:failureaccess` in the plugin (required at runtime by guava/google-auth). The one actionable security finding it raised — a `token_uri` SSRF vector — is fixed in `d9186f5c6` with a Google-host allowlist plus tests.

If the dependency additions look acceptable after review, this is the case the analyzer expects to be cleared via the `skip-diff-analyzer` label. Happy to provide provenance/hashes or anything else you need for sign-off. Full disposition of all four findings is in the comment above.

tonypiazza:
## Validation guide — google_cloud / Vertex AI connector

This walks through validating the connector end-to-end against live Vertex AI. It assumes a local build of this branch (`./gradlew run`) or any cluster with this plugin installed.

### Prerequisites (GCP side)

You need a GCP project with Vertex AI and credentials. If you don't have one, a new project with the free trial works; Gemini + text-embedding models are available in `us-central1`.

1. Enable the Vertex AI API:
   ```
   gcloud services enable aiplatform.googleapis.com --project <PROJECT_ID>
   ```
2. Create a service account and grant it Vertex AI access:
   ```
   gcloud iam service-accounts create ml-commons-test --project <PROJECT_ID>
   gcloud projects add-iam-policy-binding <PROJECT_ID> \
     --member "serviceAccount:ml-commons-test@<PROJECT_ID>.iam.gserviceaccount.com" \
     --role roles/aiplatform.user
   gcloud iam service-accounts keys create key.json \
     --iam-account ml-commons-test@<PROJECT_ID>.iam.gserviceaccount.com
   ```
   `key.json` now contains `private_key`, `client_email`, and `token_uri` — the three fields the connector needs.

> Alternatively, on a GCE/GKE node with an attached service account you can skip the key entirely and use ADC mode (Option B below).

### Cluster settings

```
PUT /_cluster/settings
{
  "persistent": {
    "plugins.ml_commons.only_run_on_ml_node": false,
    "plugins.ml_commons.trusted_connector_endpoints_regex": ["^https://.*-aiplatform\\.googleapis\\.com/.*$"]
  }
}
```

### 1. Create the connector

**Option A — service-account key** (paste the three fields from `key.json`):

```
{
  "name": "GCP Vertex AI: Gemini",
  "version": 1,
  "protocol": "google_cloud",
  "parameters": {
    "project_id": "<PROJECT_ID>",
    "location": "us-central1",
    "model": "gemini-2.5-flash"
  },
  "credential": {
    "private_key": "<private_key from key.json>",
    "client_email": "<client_email from key.json>",
    "token_uri": "https://oauth2.googleapis.com/token"
  },
  "actions": [
    {
      "action_type": "predict",
      "method": "POST",
      "url": "https://${parameters.location}-aiplatform.googleapis.com/v1/projects/${parameters.project_id}/locations/${parameters.location}/publishers/google/models/${parameters.model}:generateContent",
      "headers": { "Content-Type": "application/json" },
      "request_body": "{\"contents\":[{\"role\":\"user\",\"parts\":[{\"text\":\"${parameters.prompt}\"}]}]}"
    }
  ]
}
```

**Option B — ADC / Workload Identity** (only on a GCP-hosted node): same as above but drop `private_key`/`client_email` and set `"auth_mode": "adc"` in `parameters` with an empty `"credential": {}`.

### 2. Register + deploy

```
{ "name": "vertex-gemini", "function_name": "remote", "connector_id": "<CONNECTOR_ID>" }

POST /_plugins/_ml/models/<MODEL_ID>/_deploy
```

### 3. Predict

```
POST /_plugins/_ml/models/<MODEL_ID>/_predict
{ "parameters": { "prompt": "Say hello from OpenSearch" } }
```

Expected — HTTP 200 with a Gemini `candidates[0].content.parts[0].text`. The exact text will
vary (model output is non-deterministic); what matters is the 200 status and the response
structure below:

```json
{
  "inference_results": [
    { "output": [ { "name": "response", "dataAsMap": {
      "candidates": [ { "content": { "role": "model",
        "parts": [ { "text": "Hello from OpenSearch! How can I help you today?" } ] }, "finishReason": "STOP" } ],
      "modelVersion": "gemini-2.5-flash"
    } } ], "status_code": 200 }
  ]
}
```

### 4. (Optional) Embeddings

Same connector shape, but model `text-embedding-004`, URL ending `:predict`, and
`"request_body": "{\"instances\":[{\"content\":\"${parameters.prompt}\"}]}"`. A predict
returns `predictions[0].embeddings.values` (768 floats).

### 5. (Optional) Batch inference

Stage a JSONL input in GCS (one `{"request":{"contents":[...]}}` per line), then use a
connector with `batch_predict` action against `.../batchPredictionJobs`. Submit with
`_batch_predict`, poll with `GET /_plugins/_ml/tasks/<task_id>` (shows the remote
`JOB_STATE_*`), cancel with `POST /_plugins/_ml/tasks/<task_id>/_cancel`. See
`gcp_vertexai_batch_blueprint.md` for the full connector body.

### What we validated

We ran all of the above against live Vertex AI (`gemini-2.5-flash`, `text-embedding-004`,
`us-central1`): predict (SA key + ADC) → 200, embeddings → 200 (768 dims), batch lifecycle
submit → status → cancel (confirmed `JOB_STATE_CANCELLED`), and streaming SSE parsing. Happy
to share sanitized request/response transcripts, or to pair on a call, if that's easier than
standing up GCP access.

wanshi-ma:
Hi @tonypiazza, I ran the Batch inference end-to-end against real GCP Vertex AI. Batch submit works and returns a `task_id` and creates a remote job. But I found 2 problems following `gcp_vertexai_batch_blueprint.md`.

Problem 1:

The blueprint tells to check status / cancel with:

```
POST /_plugins/_ml/models/<MODEL_ID>/_predict/batch_predict_status  {"job_id": ...}
POST /_plugins/_ml/models/<MODEL_ID>/_predict/cancel_batch_predict  {"job_id": ...}
```

Both return `HTTP 400 no handler found for uri`.

Problem 2:

Using the OpenSearch Batch Predict route on a connector built from the blueprint:

```
GET /_plugins/_ml/tasks/<task_id>
→ HTTP 400 illegal_argument_exception:
  Illegal character in path at index 131:
  https://us-central1-aiplatform.googleapis.com/v1/projects/<proj>/locations/us-central1/batchPredictionJobs/${parameters.job_id}
```

The URL contains the literal `${parameters.job_id}`.

Could you help take a look?

tonypiazza:
@wanshi-ma thank you for testing this against real GCP — both findings were real bugs in the batch blueprint (the connector code itself was correct). Fixed in the latest commit.

**Problem 1** — the blueprint's steps 5/6 pointed at `POST /_plugins/_ml/models/<id>/_predict/batch_predict_status` (and cancel), which don't exist. Batch status/cancel go through the ML task APIs:
- status: `GET /_plugins/_ml/tasks/<task_id>` (remote state under `remote_job.state`)
- cancel: `POST /_plugins/_ml/tasks/<task_id>/_cancel`

**Problem 2** — the blueprint declared explicit `batch_predict_status` / `cancel_batch_predict` actions using `${parameters.job_id}`. ml-commons derives those calls from the `batch_predict` action (using the Vertex job's resource `name` returned at submit), so the extra actions left an unresolved `${parameters.job_id}` in the URL. The blueprint now declares **only** the `batch_predict` action.

I re-validated end-to-end against live Vertex AI using the connector built verbatim from the corrected blueprint:
- submit → `HTTP 200`, remote `JOB_STATE_PENDING`
- `GET /_plugins/_ml/tasks/<task_id>` → `HTTP 200`, resolved job name `projects/.../batchPredictionJobs/2413485481040805888` (no literal `${parameters.*}`)
- `POST /_plugins/_ml/tasks/<task_id>/_cancel` → `HTTP 200`; Vertex confirms `JOB_STATE_CANCELLED`

Would appreciate a re-check on your side when you have a chance. Thanks again for catching this.

tonypiazza:
Thanks — dispositions for the latest Code-Diff-Analyzer findings:

**`ConnectorUtils` `/v1/` URL splice (medium) — fixed** in `f51cc6fdb`. The Vertex status/cancel URL derivation now checks for the `/v1/` segment and throws a clear `IllegalArgumentException` naming the endpoint instead of producing a malformed URL when it's absent. Unit test added.

**Programmatic `Authorization` header injection (medium) — by design.** The framework guard blocks *user-supplied* `${parameters.*}` substitution into sensitive headers (untrusted request input). Here the bearer token is minted by the plugin from the connector's own stored credentials and set programmatically — it is not user request input. This is the same trust model as the existing `aws_sigv4` path, which likewise sets its authorization/signing headers programmatically rather than via parameter substitution.

**`onGeminiEvent` swallows malformed chunks (low) — intentional, and consistent with the existing handler.** This mirrors the sibling `onOpenAIEvent` in the same class, which also catches and debug-logs unparseable SSE chunks. Streaming responses can contain partial/keepalive frames; skipping a chunk that doesn't match the expected shape (rather than failing the whole stream) is the intended behavior. Happy to align both handlers differently if maintainers prefer, but I kept it consistent with the current OpenAI path.

**New dependencies (high x2) — expected; for maintainer review.** `com.google.auth:google-auth-library-oauth2-http` and the bundled `com.google.guava:failureaccess` are the supply-chain additions this feature requires; both are flagged in the PR description for provenance/hash verification. These are the mandatory dependency-flag findings the analyzer raises regardless of legitimacy — the intended path is a maintainer review + `skip-diff-analyzer` label once the deps are deemed acceptable.

docs/remote_inference_blueprints/gcp_vertexai_batch_blueprint.md
```diff
@@ -0,0 +1,104 @@
...
+
+## 3. Register and deploy the model
+
+Register and deploy the model as in the
+[Gemini blueprint](./gcp_vertexai_gemini_blueprint.md) (steps 3–5), using this connector.
+
+## 4. Submit a batch prediction job
+
+```json
+POST /_plugins/_ml/models/<MODEL_ID>/_batch_predict
```
mingshl, line 94: Trying to understand the workflow, when I have model_input in google uri, I used this OpenSearch batch_predict api to trigger Google model for batch predict and then the output store in google uri. How can OpenSearch ingest these prediction output data?

tonypiazza: @mingshl  This PR covers the *offline batch inference* half of the flow: OpenSearch submits the `batchPredictionJob`, Vertex runs it and writes the predictions to the GCS `output_uri` you specify.

Retrieving those results back into OpenSearch is a separate step, and today it isn't wired for GCP. The existing `_batch_ingestion` API (`TransportBatchIngestionAction`) supports only `s3` and `openai` source types (`@Ingester("s3")` / `@Ingester("openai")`) — there is no `gcs` ingester yet. So after the Vertex job completes, the JSONL predictions sit under your GCS `outputUriPrefix` and would need to be pulled in manually.

I scoped this PR to the connector + batch job lifecycle to keep it reviewable. Closing the ingestion loop is a planned follow-up contribution: a `@Ingester("gcs")` implementation (mirroring `S3DataIngestion`) plus `gcs` source-type handling in `TransportBatchIngestionAction`, so `_batch_ingestion` can read the Vertex predictions from GCS into an index. I'll put that up as a separate PR after this one lands.

jiapingzeng (approved):
LGTM! Thanks for the PR

mingshl:
@tonypiazza we are consulting with security team and see there is any security risks for this feature. Will get you back once we get more clarifications.

tonypiazza:
Thanks @mingshl — understand you need to run it through the security team. To help move that along, here's what's relevant on the security surface:

**Dependencies added** (both official artifacts on Maven Central, verifiable by hash):
- `com.google.auth:google-auth-library-oauth2-http:1.23.0` — Google's official OAuth2 client. Transitively brings `google-auth-library-credentials:1.23.0`, `google-http-client:1.43.3`, and `google-http-client-gson:1.43.3`. (I excluded its `grpc-context` and `guava` transitives to avoid classpath conflicts.)
- `com.google.guava:failureaccess:1.0.1` — a Guava runtime companion, required on the plugin classloader.

**Mitigations already in the PR:**
- **SSRF** — the `token_uri` credential is validated against an allowlist (`https`, `*.googleapis.com` only) before any token refresh, so it can't be pointed at an internal/arbitrary host. The Vertex batch status/cancel URL derivation likewise fails fast if the endpoint is malformed.
- **Header injection** — the OAuth2 bearer token is minted from the connector's own stored credentials and set programmatically; it does **not** go through user `${parameters.*}` substitution, so the framework's blocked-dynamic-header guard is preserved. This mirrors the existing `aws_sigv4` trust model.
- **Endpoint trust** — all outbound calls are still subject to `plugins.ml_commons.trusted_connector_endpoints_regex`.
- **ADC mode** is opt-in (`auth_mode: adc`) and intended for GCP-hosted nodes; documented as such.

Happy to answer specific questions from the security team if needed. I'll hold off on further changes so the diff stays stable during review.

tonypiazza:
**Fixed: completed Vertex batch jobs now reach a terminal state**

While reviewing the batch status handling, I found a gap: a successfully-completed Vertex `batchPredictionJob` reports its status under the `state` field with value `JOB_STATE_SUCCEEDED`, which the shared batch-status settings didn't recognize (`state` wasn't in `remote_job.status_field`, and `JOB_STATE_*` wasn't in the status regexes). The `MLTask` never left `RUNNING`, so `MLBatchTaskUpdateProcessor` would re-poll it indefinitely. (Cancel happened to work because `JOB_STATE_CANCELLED` substring-matches the cancelled regex.)

Fixed with the minimal, backward-compatible change — extending the shared, provider-neutral status settings (no `google_cloud`-specific code path):
- added `state` to `remote_job.status_field`
- added `JOB_STATE_SUCCEEDED` / `JOB_STATE_FAILED` / `JOB_STATE_EXPIRED` to the completed / failed / expired regexes

Verified live against Vertex AI on a clean build: a job driven to `JOB_STATE_SUCCEEDED` now transitions the task to `COMPLETED` using the shipped defaults (no cluster-setting overrides). Unit tests cover all four Vertex terminal states through the real state machine.

**Follow-up RFC (heads-up):** the batch-status mechanism relies on a single global set of status fields + regexes shared across all providers. As more providers are added, that shared vocabulary grows and the collision surface with it. I plan to file a separate RFC proposing a **provider-neutral, per-connector status-mapping contract** — each connector declaring its own status semantics, with the current global settings preserved as the default for backward compatibility. The change here is deliberately the small, compatible fix; the broader design is worth its own discussion.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -399,7 +399,8 @@ public void validate(List<String> value) {
                 .of(
                     "status", // openai, bedrock, cohere
                     "Status",
-                    "TransformJobStatus" // sagemaker
+                    "TransformJobStatus", // sagemaker
+                    "state" // google vertex ai (batchPredictionJobs)
```
ylwu-amzn: **High-priority — this status field is cluster-global, not Vertex-scoped.**

`GetTaskTransportAction.processTaskResponse` (line 550, unchanged) iterates **every** configured status field with no precedence and no connector-type gating:

```java
for (String statusField : remoteJobStatusFields) {
    String statusValue = String.valueOf(remoteJob.get(statusField));
    if (remoteJob.containsKey(statusField)) { updateTaskState(updatedTask, mlTask, statusValue); }
}
```

So for any *existing* non-Google batch connector whose remote status response happens to contain a `state` key — a very common field name — the ML task state is now additionally driven by that value. And because the loop has no precedence, a later field can overwrite the verdict of an earlier one.

Compounding it: `matchesPattern` is `pattern.matcher(input).find()`, i.e. substring semantics on a `CASE_INSENSITIVE` pattern. A `state` value like `"failed_over"` or `"completed_at"` would match.

The batch blueprint (`gcp_vertexai_batch_blueprint.md:103`) advertises this as working "by default — no per-connector configuration is required", which is exactly what makes the blast radius cluster-wide.

Could the field be resolved per-connector / per-remote-server instead? `ConnectorAction.getRemoteServerFromURL` already exists for this kind of dispatch. At minimum, the loop probably wants first-match-wins precedence.

Also: `GetTaskTransportActionTests` mutates the shared test settings to include `state` and then only tests the Google values — there's no test asserting a pre-existing non-Google connector carrying a `state` key is unaffected.

plugin/src/main/java/org/opensearch/ml/action/tasks/GetTaskTransportAction.java
```diff
@@ -356,6 +363,13 @@ private void processRemoteBatchPrediction(
                     .orElse(null)
             );
 
+        // The Vertex AI batch job identifier ("name") is a slash-delimited resource path used
+        // directly in the status/cancel URL; keep it un-escaped so its '/' are not turned into '\/'.
+        if (parameters.containsKey("name")) {
+            String existingNoEscape = parameters.get(NO_ESCAPE_PARAMS);
+            parameters.put(NO_ESCAPE_PARAMS, existingNoEscape == null || existingNoEscape.isBlank() ? "name" : existingNoEscape + ",name");
+        }
```
ylwu-amzn: **High-priority — the guard is generic even though the comment is Vertex-specific.**

The comment says "The Vertex AI batch job identifier (`name`) …", but the condition is `parameters.containsKey("name")` — it fires for **any** batch connector whose `remote_job` map contains a `name` key, including OpenAI / Bedrock / SageMaker jobs.

For those, `name` silently stops being JSON-escaped when substituted into the status/cancel request body. For a value containing `"` or `\` that's a malformed-JSON / injection-shaped change, not a no-op.

Suggest gating on the connector actually being Vertex — this PR already adds `ConnectorAction.VERTEX_AI` and there's `getRemoteServerFromURL` for exactly this dispatch.

Also, this block is byte-identical to `CancelBatchJobTransportAction.java:194-199`; worth factoring into one shared helper.

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
```diff
@@ -182,6 +191,13 @@ private void processRemoteBatchPrediction(MLTask mlTask, ActionListener<MLCancel
                     .orElse(null)
             );
 
+        // The Vertex AI batch job identifier ("name") is a slash-delimited resource path used
+        // directly in the status/cancel URL; keep it un-escaped so its '/' are not turned into '\/'.
+        if (parameters.containsKey("name")) {
+            String existingNoEscape = parameters.get(NO_ESCAPE_PARAMS);
+            parameters.put(NO_ESCAPE_PARAMS, existingNoEscape == null || existingNoEscape.isBlank() ? "name" : existingNoEscape + ",name");
+        }
```
ylwu-amzn: Duplicate of the block in `GetTaskTransportAction.java:366-371` — same generic-`name` concern (see that comment), and worth extracting into a shared helper rather than keeping two byte-identical copies.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
```diff
@@ -473,6 +474,21 @@ public static ConnectorAction createConnectorAction(Connector connector, Connect
...
+                // addressed as {host}/v1/{name}[:cancel], so build from the endpoint's base URL.
+                int v1Index = predictEndpoint.indexOf("/v1/");
+                if (v1Index < 0) {
+                    throw new IllegalArgumentException(
+                        "Vertex AI batch_predict endpoint must contain a '/v1/' path segment to derive the status/cancel URL, got: "
+                            + predictEndpoint
+                    );
+                }
+                String vertexBaseUrl = predictEndpoint.substring(0, v1Index + "/v1/".length());
+                url = isCancelAction ? vertexBaseUrl + "${parameters.name}:cancel" : vertexBaseUrl + "${parameters.name}";
```
ylwu-amzn: **High-priority — two issues with the `/v1/` splice.**

1. **`v1beta1` endpoints are rejected.** Vertex AI publishes `batchPredictionJobs` under both `/v1/` and `/v1beta1/`. A user who configures a `v1beta1` endpoint gets `IllegalArgumentException` here. Matching `/v1` plus an optional `beta<N>` segment (or a small regex) would cover both without loosening much.

2. **The failure comes too late to be useful.** This URL is derived lazily, in `GetTaskTransportAction.executeConnector` / `CancelBatchJobTransportAction.executeConnector` — i.e. **after** `BATCH_PREDICT` already submitted a job to Vertex. So a misconfigured endpoint leaves an orphaned remote job that can be neither polled nor cancelled, while still billing. Validating the `batch_predict` URL at connector-create time (where the other endpoint validation lives) would fail before any money is spent.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleCredentialProvider.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        String scheme = uri.getScheme();
+        String host = uri.getHost();
+        if (!"https".equalsIgnoreCase(scheme) || host == null) {
+            throw new IllegalArgumentException("token_uri must be an https Google endpoint, got: " + tokenUri);
+        }
+        String lowerHost = host.toLowerCase(java.util.Locale.ROOT);
+        boolean allowed = lowerHost.equals("oauth2.googleapis.com")
+            || lowerHost.equals("googleapis.com")
+            || lowerHost.endsWith(".googleapis.com");
```
ylwu-amzn: The SSRF hardening here (025077b7a) is a genuine improvement, but the allowlist is broader than this exchange needs: it accepts *any* `*.googleapis.com` host with *any* path.

The value POSTed to this URI is a signed JWT assertion for the service account, so a connector author can direct that assertion at e.g. `https://<region>-aiplatform.googleapis.com/v1/…` or any other Google API surface rather than the token endpoint. Much smaller than the pre-hardening state, but the only host Google documents for this exchange is `oauth2.googleapis.com`.

Suggest narrowing to an explicit set — `oauth2.googleapis.com`, plus `sts.googleapis.com` if workload-identity federation is in scope.

Note `GoogleCredentialProviderTest.validateTokenUri_acceptsGoogleEndpoints` currently *asserts* `us-central1-aiplatform.googleapis.com` is accepted, so the broad behavior is codified as intended — is that deliberate?

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleCredentialProvider.java
```diff
@@ -0,0 +1,98 @@
...
+     * Build a provider from connector config. Service-account key mode reconstructs
+     * {@link ServiceAccountCredentials} from the decrypted credential fields; ADC mode uses
+     * {@link GoogleCredentials#getApplicationDefault()}.
+     */
+    public static GoogleCredentialProvider fromConnector(GoogleCloudConnector connector) {
+        try {
+            List<String> scopes = Collections.singletonList(connector.getScopes());
+            GoogleCredentials creds;
+            if (connector.useAdc()) {
+                creds = GoogleCredentials.getApplicationDefault().createScoped(scopes);
```
ylwu-amzn, line 45: Question on the security model for ADC mode, since you flagged this area for maintainer review.

`GoogleCredentials.getApplicationDefault()` resolves credentials from `GOOGLE_APPLICATION_CREDENTIALS` (a filesystem read) or the GCE/GKE metadata server at `169.254.169.254`. Both bypass `connector.validateResolvedEndpoint(...)` / `trustedConnectorEndpointsRegex` and the restricted-IP patterns in `MLHttpClientFactory` — by design, since the auth library does that I/O, not the connector.

That's the right implementation for Workload Identity. It does mean any user permitted to create a connector can cause the node to mint tokens for the node's own service account.

Is ADC intended to be gated behind a cluster setting like other privileged features here, or is "the cluster operator opted in by running on GCP with a bound SA" considered sufficient? Either answer is fine — worth stating explicitly in the PR description.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleConnectorExecutor.java
```diff
@@ -0,0 +1,257 @@
...
+    @Override
+    public Logger getLogger() {
+        return log;
+    }
+
+    GoogleCredentialProvider credentialProvider() {
+        if (credentialProvider == null) {
+            credentialProvider = GoogleCredentialProvider.fromConnector(connector);
+        }
+        return credentialProvider;
```
ylwu-amzn: `credentialProvider` is a non-`volatile` field lazily initialized here with no synchronization. Concurrent predicts on the same deployed model can each construct a provider — wasteful, since each builds a `GoogleCredentials` and may trigger its own token fetch — and a partially-published instance is observable in principle.

Either build it eagerly in the constructor (the connector is already available there), or make the field `volatile` with double-checked locking.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
```diff
@@ -243,6 +247,39 @@ private void onOpenAIEvent(String data) {
...
+                }
+
+                String finishReason = extractPath(dataMap, "$.candidates[0].finishReason");
+                if (finishReason != null) {
+                    sendCompletionResponse(isStreamClosed, streamActionListener);
+                }
+            } catch (Exception e) {
+                log.debug("Skipping malformed chunk: {}", data);
+            }
+        }
```
ylwu-amzn, line 292: Three gaps in this branch relative to the OpenAI branch in the same class:

**1. The TTFT metric isn't emitted.** Line 264 logs `"First token received. …"`, but unlike the OpenAI branch (line 354) and `BedrockStreamingHandler.java:378`, there's no `AgentUtils.logTimeToFirstToken(modelId, tenantId, timeToFirstToken)` call. The log line is there; the metric isn't.

**2. No AG-UI handling.** The OpenAI branch threads `isAGUIAgent` through at lines 285-287, 314-315, 322, 358, 383, 415, 481. This branch has none, so AG-UI agents over Vertex will emit raw content responses instead of AG-UI events. If that's intentionally out of scope, a `// TODO` would help.

**3. A safety-blocked prompt never terminates the stream.** Completion is only signalled when `finishReason` is non-null. A blocked response returns `promptFeedback.blockReason` with **no** `candidates` and no `finishReason` — so neither `sendContentResponse` nor `sendCompletionResponse` fires and the client hangs. Same for an error frame. `HttpStreamingHandlerGeminiTest` covers text / final / malformed chunks but not this shape, which is a real Vertex response — worth a test.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
```diff
@@ -115,7 +116,9 @@ public MLCreateConnectorInput(
                 throw new IllegalArgumentException("Connector protocol is null");
             }
             boolean isMcpConnector = (protocol.equals(MCP_SSE) || protocol.equals(MCP_STREAMABLE_HTTP));
-            if ((credential == null || credential.isEmpty()) && !isMcpConnector) {
+            // google_cloud connectors may omit credentials when using ADC / Workload Identity.
+            boolean allowsEmptyCredential = isMcpConnector || protocol.equals(GOOGLE_CLOUD);
+            if ((credential == null || credential.isEmpty()) && !allowsEmptyCredential) {
```
ylwu-amzn: `allowsEmptyCredential` is true for *all* `google_cloud` connectors, not just `auth_mode=adc`. So a service-account connector created with no credential at all passes here and only fails later in `GoogleCloudConnector.validate()`.

The test (`constructorMLCreateConnectorInput_GoogleCloudWithEmptyCredential_ShouldNotThrow`) passes `parameters: Map.of("auth_mode", "adc")` even though production ignores `auth_mode` at this point — which masks the over-broad relaxation. Either check `auth_mode` here too, or drop it from the test so it reflects what the code actually does.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -430,15 +431,15 @@ public void validate(List<String> value) {
...
+            "(expired|timeout|JOB_STATE_EXPIRED)",
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         );
 
     public static final Setting<String> ML_COMMONS_REMOTE_JOB_STATUS_FAILED_REGEX = Setting
         .simpleString(
             ML_PLUGIN_SETTING_PREFIX + "remote_job.status_regex.failed",
-            "(failed)",
+            "(failed|JOB_STATE_FAILED)",
```
ylwu-amzn: `JOB_STATE_FAILED` is a no-op. The regex is compiled `CASE_INSENSITIVE` and matched with `find()` (substring semantics), so the existing `(failed)` alternative already matches `JOB_STATE_FAILED`. Same for `JOB_STATE_EXPIRED` on line 434 vs `(expired)`.

Only `JOB_STATE_SUCCEEDED` was actually needed. Your own test comment makes exactly this argument for `JOB_STATE_CANCELLED` — the same reasoning applies to these two.

common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
```diff
@@ -0,0 +1,170 @@
...
+    public String getPrivateKey() {
+        return decryptedCredential.get(PRIVATE_KEY_FIELD);
+    }
+
+    public String getClientEmail() {
+        return decryptedCredential.get(CLIENT_EMAIL_FIELD);
+    }
+
+    public String getTokenUri() {
+        String uri = decryptedCredential.get(TOKEN_URI_FIELD);
```
ylwu-amzn, line 145: These three dereference `decryptedCredential` unguarded. In ADC mode `AbstractConnector.decrypt` short-circuits on an empty credential map and leaves `decryptedCredential` **null** — so a future caller that reaches these without first checking `isAdcMode()` gets an NPE instead of a clear error. `GoogleCredentialProvider.fromConnector` gets this right today via `connector.useAdc()`, so this is purely defensive.

An `Objects.requireNonNull(decryptedCredential, "connector credentials are not decrypted")` or an explicit `isAdcMode()` guard would be cheap insurance.

common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public GoogleCloudConnector(StreamInput input) throws IOException {
+        super(input);
+        validate();
+    }
+
+    public GoogleCloudConnector(String protocol, StreamInput input) throws IOException {
+        super(protocol, input);
+        // Validation skipped for StreamInput deserialization (mirrors AwsConnector).
```
ylwu-amzn: Minor: the 1-arg `StreamInput` constructor (line 88) validates, this 2-arg one skips validation. It does mirror `AwsConnector`, so it's consistent with precedent — but the asymmetry means a `google_cloud` connector arriving over the wire via this path is never validated. Worth a word in the comment on which path production actually takes.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
```diff
@@ -49,8 +49,11 @@ public class ConnectorAction implements ToXContentObject, Writeable {
     public static final String COHERE = "cohere";
     public static final String BEDROCK = "bedrock";
     public static final String SAGEMAKER = "sagemaker";
+    // Matches the Vertex AI host (e.g. us-central1-aiplatform.googleapis.com) for batch job status/cancel.
+    public static final String VERTEX_AI = "aiplatform";
     public static final String SAGEMAKER_PRE_POST_FUNC_TEXT = "default";
-    public static final List<String> SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES = List.of(SAGEMAKER, OPENAI, BEDROCK, COHERE);
+    public static final List<String> SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES = List
+        .of(SAGEMAKER, OPENAI, BEDROCK, COHERE, VERTEX_AI);
```
ylwu-amzn, line 56: `VERTEX_AI` is added to `SUPPORTED_REMOTE_SERVERS_FOR_DEFAULT_ACTION_TYPES`, but the `validateProcessFunctions` switch (lines 338-362) has no `VERTEX_AI` case — so it silently no-ops for Vertex connectors while SageMaker / OpenAI / Bedrock / Cohere all get pre/post-process-function validation. Intentional, or an oversight?

common/src/main/java/org/opensearch/ml/common/connector/ConnectorProtocols.java
```diff
@@ -12,21 +12,22 @@ public class ConnectorProtocols {
...
-            throw new IllegalArgumentException(
-                "Unsupported connector protocol. Please use one of " + Arrays.toString(VALID_PROTOCOLS.toArray(new String[0]))
-            );
+            throw new IllegalArgumentException("Unsupported connector protocol. Please use one of " + supportedProtocols());
         }
     }
+
+    public static String supportedProtocols() {
+        return "[" + String.join(", ", VALID_PROTOCOLS) + "]";
+    }
```
ylwu-amzn, line 32: Style / test-strength nit: extracting this helper is a readability win, but `ConnectorProtocolsTest`, `MLCreateConnectorInputTests`, `HttpConnectorTest`, `McpConnectorTest` and `McpStreamableHttpConnectorTest` were all switched to assert against `supportedProtocols()` instead of a literal expected list. That makes those assertions tautological — they'd still pass if a protocol were accidentally added or removed. Keeping one literal expected string somewhere would preserve the guard.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
tonypiazza: @ylwu-amzn You're right that adding `state` to the global `remote_job.status_field` is too broad — the status loop applies every configured field to every connector, so a non-Google connector whose response carries a `state` key would be interpreted against it, and with substring matching (`Matcher.find()`) a value like `completed_at` would false-match. That's a cluster-wide side effect.

Rather than gate the global defaults per-server, I'd like to propose a connector-scoped mechanism and get your read on the approach before I implement it. Sketch:

**Optional, first-class `batch_job_status` on the connector:**

```json
"batch_job_status": {
  "field_name": "state",
  "mapping": {
    "JOB_STATE_SUCCEEDED": "COMPLETED",
    "JOB_STATE_FAILED":    "FAILED",
    "JOB_STATE_EXPIRED":   "EXPIRED",
    "JOB_STATE_CANCELLED": "CANCELLED"
  }
}
```

- `field_name` = the flat, top-level status key to read; `mapping` = **exact-match** remote value → `MLTaskState` (exact match, not substring, so `completed_at` can't false-match).
- **Resolution rule:** if a connector declares `batch_job_status`, status is interpreted *exclusively* from the declaration and the global field/regex loop is skipped for it. If it declares nothing, current global behavior is unchanged.
- **BWC:** the global settings revert to their pre-PR defaults (I'd drop the `state` / `JOB_STATE_*` additions) and remain the fallback for undeclared connectors — nothing removed. New connector field is version-guarded on the wire.

Rationale for "flat field + exact map + optional": I checked the batch status shape for OpenAI, Groq, SageMaker, Bedrock, Cohere, Vertex, and OCI work requests — all expose status in a flat, top-level field, and the OpenAI-compatible ones already work with the current defaults (so they'd declare nothing). Vertex/OCI are the divergent cases that declare a mapping. No verified provider nests status, so I'm intentionally not adding nested-path support.

This also lines up with future providers (e.g. Azure AI Foundry) — a new provider becomes a declare-and-go connector instead of another edit to shared status code.

Does this approach look right to you? If so I'll implement it here (reverting the global-defaults change), with tests covering both the declared path and an undeclared connector carrying a `state` key to prove no cross-provider bleed. I'll also address your other comments (the `name` no-escape gating/dedup, the `/v1/` + `v1beta1` and create-time validation points, etc.) — replying to those threads separately.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleCredentialProvider.java
tonypiazza: @ylwu-amzn Narrowed in 6e741068e. `validateTokenUri` now accepts exactly `oauth2.googleapis.com` instead of any `*.googleapis.com` host with any path. That is the only Google surface that legitimately receives the signed service-account JWT during token exchange, so there is no reason to let token_uri point at other Google APIs. ADC/Workload Identity does not go through this method (it resolves via `getApplicationDefault()`), so no `sts.googleapis.com` entry is needed. I also updated the test that had codified the broad behavior; it now asserts other `*.googleapis.com` hosts and the bare apex are rejected.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleConnectorExecutor.java
tonypiazza: @ylwu-amzn Fixed in 6e741068e. `credentialProvider` is now `volatile` with double-checked locking around construction, so concurrent predicts cannot race to build it or observe a partially-constructed instance. I kept it lazy rather than eager-initializing in the constructor: eager init would resolve credentials (ADC hits the metadata server, SA-key reconstructs from the key) for every executor at construction time, and would defeat the test injection path. DCL closes the race while preserving both.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
tonypiazza: @ylwu-amzn This thread raises three separate points. Two are fixed in 9b3b537f7; the AG-UI one I would like to scope as a follow-up.

- **(a) TTFT metric.** Fixed. The Gemini branch now calls `AgentUtils.logTimeToFirstToken(...)` on first token, matching the OpenAI branch. It was previously only logging the timing.
- **(b) AG-UI handling.** Deferring to a follow-up. AG-UI streaming is an agent-framework feature gated on `agent_type == AG_UI`; it only activates when an AG-UI agent drives the stream and is independent of the connector/LLM interface. Plain streaming over google_cloud works today (verified against live Vertex). The missing piece is specifically an AG-UI agent running over a Gemini LLM, which is not wired, tested, or documented anywhere in this PR; there is no user path that reaches it yet. Replicating the ~100 lines of AG-UI event weaving for an unexercised agent+provider pairing would be untested speculative code. It fits naturally with Vertex agent / function-calling support, which is a separate follow-up. I will track it there. Let me know if you would prefer otherwise.
- **(c\) Safety-blocked prompt hangs the stream.** Fixed, and this was a real bug. A prompt rejected by Vertex safety filters arrives as `promptFeedback.blockReason` with no candidates and no `finishReason`, so the completion sentinel (gated on `finishReason`) never fired and the client hung. The handler now detects `blockReason` and sends the completion response. I added a `HttpStreamingHandlerGeminiTest` case with a real blocked-prompt frame. This is the same terminal-state class as the batch-status hang.

common/src/main/java/org/opensearch/ml/common/transport/connector/MLCreateConnectorInput.java
tonypiazza: @ylwu-amzn Fixed in d7c22a68b (earlier batch). The empty-credential relaxation now applies only when `protocol == google_cloud AND parameters.auth_mode == adc`, not to every google_cloud connector. A service-account-key connector with a missing credential now fails at creation. Added a negative test (google_cloud, no auth_mode, empty credential rejected).

common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
tonypiazza: @ylwu-amzn Fixed at the root in cb9352d1e. `AbstractConnector.decrypt` now initializes `decryptedCredential` to an empty map on the credential-less short-circuit, so credential-less connectors (google_cloud ADC, MCP) get a non-null map. This also removes a latent NPE in `MLEngine.getConnectorCredential`'s `putIfAbsent`. Added a test asserting the ADC accessors return null (not NPE) and token_uri falls back to the default.

common/src/main/java/org/opensearch/ml/common/connector/GoogleCloudConnector.java
tonypiazza: @ylwu-amzn Documented in cb9352d1e. I expanded the constructor comment to explain the wire-deserialization path intentionally skips `validate()`: the object was already validated at create time via the builder/XContent constructors, and the aws_sigv4 connector follows the same pattern. No behavior change.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorProtocols.java
tonypiazza: @ylwu-amzn Fixed in cb9352d1e. Added a literal-list guard: `assertEquals("[aws_sigv4, http, google_cloud, mcp_sse, mcp_streamable_http]", supportedProtocols())`. The other tests keep the dynamic helper to avoid duplicating the 5-way list everywhere, but the guard now pins the exact contents.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
tonypiazza: @ylwu-amzn This raises two points; the first is fixed, the second I would like to land with #1.

- **Rejects v1beta1.** Fixed in d7c22a68b. The Vertex status/cancel URL derivation now recognizes `/v1/` and `/v1beta<N>/` via `VERTEX_API_VERSION_PATTERN` (`/v1(beta\d+)?/`), so batchPredictionJobs endpoints published under v1beta1 are no longer rejected. Added a v1beta1 test.
- **Validates too late.** Validating the batch_predict URL at connector-create (so a bad version segment fails before a job is submitted, rather than post-submit with an orphaned billing job) belongs with the create-time validation work in the #1 `batch_job_status` rework, which threads the connector through this path. Deferring this part to land with #1.

plugin/src/main/java/org/opensearch/ml/action/tasks/GetTaskTransportAction.java
tonypiazza: @ylwu-amzn Agreed this is too broad. The guard keys off `parameters.containsKey("name")`, so it silently stops JSON-escaping `name` for every provider, not just Vertex. The clean fix is to Vertex-gate it and extract the shared helper, but that means threading the connector into the status path, which is exactly what the #1 `batch_job_status` rework does. I would like to land this with #1 rather than add a throwaway gate now. Holding on your decision in the #1 thread (discussion_r3722712461).

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
tonypiazza: @ylwu-amzn Same block, duplicated here. This will be extracted into the shared helper and fixed together with the GetTaskTransportAction thread when #1 lands.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
tonypiazza: @ylwu-amzn Correct. The existing `(failed)`/`(expired)` substring patterns already match `JOB_STATE_FAILED`/`JOB_STATE_EXPIRED` (case-insensitive `find()`), so those additions are inert; only `JOB_STATE_SUCCEEDED` was actually needed. This is moot under #1: the batch_job_status design reverts the global-defaults change entirely, so these lines come back out. Will be removed when #1 lands.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/GoogleCredentialProvider.java
tonypiazza: @ylwu-amzn This is by design and opt-in. ADC only activates when the connector is explicitly created with `auth_mode=adc` (the default is service-account key, which does go through `validateTokenUri`). When an operator opts into ADC they are declaring that the node runs in a GCP-hosted environment (GKE/GCE Workload Identity), where resolving credentials from `GOOGLE_APPLICATION_CREDENTIALS` or the metadata server (169.254.169.254) is the intended, standard mechanism; it is how every Google client library authenticates there. The trustedConnectorEndpointsRegex still governs the *inference* endpoint the token is sent to; ADC only governs how the node authenticates itself locally. Treating the metadata server as untrusted would break the standard GCP auth path. Happy to document this trade-off in the blueprint if useful.

common/src/main/java/org/opensearch/ml/common/connector/ConnectorAction.java
tonypiazza: @ylwu-amzn Intentional, but happy to add a case if you would prefer it for symmetry. `validateProcessFunctions` only does anything when `isInBuiltProcessFunction(processFunction)` is true, i.e. the pre/post-process function is a built-in `connector.*` function, and it warns when the built-in does not match the detected remote server. There are no built-in process functions for google_cloud/Vertex (the blueprints use no process function, or a custom one), so a `VERTEX_AI` case would have no built-in token to validate against and would be a no-op today. I left it out rather than add an inert branch, but I can add a placeholder for when Vertex built-in functions exist if you would rather.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
ylwu-amzn: **`batch_job_status`: approach looks good, please implement it here.**

Connector-scoped resolution + exact-match mapping + skip-the-global-loop-when-declared is the right design. It removes the cross-connector interaction rather than narrowing it, and it's BWC by construction. Agreed on skipping nested-path support.

On scope — I'd rather not just revert the global defaults and defer, because that would leave Vertex batch status unmapped and the batch blueprint in this PR wouldn't complete its task lifecycle. So: implement the minimal version here (`field_name` + exact `mapping`, global loop skipped when declared, defaults reverted), with the undeclared-connector-carrying-`state` test you proposed. That unblocks the `name` no-escape and `JOB_STATE_*` threads too.

Create-time `batch_predict` URL validation can be a follow-up — no need to grow this PR further.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
```diff
@@ -243,6 +247,50 @@ private void onOpenAIEvent(String data) {
             }
         }
 
+        private void onGeminiEvent(String data) {
```
ylwu-amzn, line 250: Agreed on not doing the AG-UI event weaving here — but the pairing is reachable today, so it needs a guard rather than nothing.

`MLAGUIAgentRunner:95-105` accepts the Gemini interface via `FunctionCallingFactory` (which has a Gemini case), then unconditionally sets `agent_type=ag_ui` with no interface allowlist. So `isAGUIAgent` is true here, `onGeminiEvent` sends raw text, `convertToAGUIEvent` fails `isJson`, and the stream truncates on the first token with `RunErrorEvent("Unexpected chunk: ...")`.

A few lines fixes it — either reject `ag_ui` + Gemini at agent-execute time, or emit one actionable error here and terminate. Then the follow-up stays purely additive.

akolarkunnu (approved):
Changes looking good to me.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
tonypiazza: @ylwu-amzn Implemented in this PR, scoped as you described. Summary of what landed:

- New first-class connector field `batch_job_status: { field_name, mapping }` (`BatchJobStatusMapping`, 19a7631ae): `field_name` is the flat top-level status key, `mapping` is an EXACT-match remote-value to MLTaskState. Validated at create and on the wire.
- Connector-scoped resolution in `GetTaskTransportAction.processTaskResponse` (0ebc2e46b): when a connector declares `batch_job_status`, status is resolved exclusively from it and the global field-list/regex loop is skipped; undeclared connectors keep the current behavior. Includes the regression test you asked for: an undeclared connector carrying a `state` key is unaffected.
- Reverted the interim global `state` / `JOB_STATE_*` defaults (099e30209), which also resolves the `JOB_STATE_FAILED`/`JOB_STATE_EXPIRED` no-op thread.
- The Vertex batch blueprint now declares the mapping (e0ecd37a2), so Vertex batch status still reaches a terminal state without the reverted defaults.
- Carried the field through the REST create/update path too (`MLCreateConnectorInput` + `HttpConnector.update`, eb304b005).

I validated the full path end-to-end against live Vertex (batch submit through ml-commons, no cluster-setting overrides): `JOB_STATE_SUCCEEDED` transitions the task to COMPLETED and `JOB_STATE_CANCELLED` to CANCELLED, both via the declared mapping. Create-time `batch_predict` URL validation I left as a follow-up per your note.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/streaming/HttpStreamingHandler.java
tonypiazza: @ylwu-amzn You're right, and my earlier reply was wrong on that point. `MLAGUIAgentRunner` accepts the Gemini interface via `FunctionCallingFactory` and then unconditionally sets `agent_type=ag_ui`, so the pairing is reachable and would truncate the stream. I added a guard (21216bd45): `MLAGUIAgentRunner.run` now rejects the Gemini interface up front with a clear "AG-UI streaming is not yet supported for the Gemini (google_cloud) LLM interface" error, before the stream opens, rather than failing mid-stream. Full AG-UI event weaving for Gemini is the separate Vertex-agent follow-up; the guard keeps that purely additive. Test added in MLAGUIAgentRunnerTest.

plugin/src/main/java/org/opensearch/ml/action/tasks/GetTaskTransportAction.java
tonypiazza: @ylwu-amzn Landed in bdfe2316c. The `name` no-escape is now gated on the value being a slash-delimited resource path rather than merely present, and the duplicated block is extracted into a shared `ToolUtils.markResourcePathNameNoEscape`. I gated on the value shape (a Vertex batchPredictionJobs name is always `projects/.../batchPredictionJobs/N`) instead of the protocol, because that is the actual property that requires leaving `/` unescaped in the derived URL: a plain provider job name with no slash is no longer affected, which is the over-broad behavior you flagged. Open to switching to an explicit protocol gate if you'd prefer it.

plugin/src/main/java/org/opensearch/ml/action/tasks/CancelBatchJobTransportAction.java
tonypiazza: @ylwu-amzn Fixed together with the GetTask thread in bdfe2316c: both call sites now use the shared `ToolUtils.markResourcePathNameNoEscape` helper, so the byte-identical duplication is gone.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/ConnectorUtils.java
tonypiazza: @ylwu-amzn Part A (v1beta1 endpoints) is fixed. Part B (validating the batch_predict URL at connector-create so a bad version segment fails before a job is submitted) I'm leaving as the follow-up you agreed to on the batch_job_status thread, since it fits with the create-time validation work rather than this PR.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
tonypiazza: @ylwu-amzn Resolved: the global `state` / `JOB_STATE_*` defaults were reverted entirely (099e30209) as part of moving to the per-connector mechanism, so these no-op regex tokens are gone.

ylwu-amzn (approved):

tonypiazza:
Updating this now that the security review is resolved and the branch has moved.

**Security requirement met.** Per the security team (relayed by @mingshl), the connector is now gated behind an opt-in feature flag `plugins.ml_commons.connector.vertexai_enabled`, disabled by default, following the memory-retention flag pattern. Creating a `google_cloud` connector while it's off returns 403 with a message naming the setting to enable. Verified end to end on a local node (flag off rejects, flag on works). The branch is rebased on latest `main`, mergeable, and the existing approvals from @ylwu-amzn, @jiapingzeng, @akolarkunnu carried over. (The latest push also corrected a test-wiring gap: the new flag needed to be registered in the plugin-module `MLFeatureEnabledSettingTests` cluster settings; builds and the relevant common + plugin suites pass locally.)

Three maintainer-side steps remain to merge, none in the connector code:

1. **Approve the Build-and-Test workflow run.** It reset to "waiting" after the latest push (external-contributor gate) and won't start on its own.

2. **`skip-diff-analyzer` label.** Code-Diff-Analyzer fails only on the expected findings: the `google-auth-library` dependency and the programmatic `Authorization` header (the plugin mints the OAuth2 token from the connector's own credentials and sets it per request, same trust model as `aws_sigv4`). Both were dispositioned in earlier rounds.

3. **A known-flaky integTest.** When Build-and-Test runs, `RestMemoryRetentionJobIntervalIT.testUpdateIntervalRescheduleReflectedInPersistedDoc` (a memory-retention scheduler test added in #4952) may fail on `linux docker`. It has been failing on `main` and other PRs for weeks and touches none of the code here. Deferring to you on how to handle it for merge.

For context: the per-connector `batch_job_status` mechanism is implemented per @ylwu-amzn's guidance, and I validated the full batch lifecycle end to end against live Vertex AI (`JOB_STATE_SUCCEEDED` to COMPLETED and `JOB_STATE_CANCELLED` to CANCELLED via the declared mapping).

PR #4928, opened 2026-07-23
[BUG] Fix retention job interval setting being write-once

AirFunnyB:
### Description

The memory retention background job's schedule was effectively write-once. The job document in `.plugins-ml-jobs` was created with `opType(CREATE)`, and `indexJob` swallowed the resulting `VersionConflictEngineException` as success, so the document was never overwritten. As a result, changing `plugins.ml_commons.memory.retention_job_interval_hours` had no effect — neither a live `PUT _cluster/settings` nor a restart / `settings.yml` value would reschedule the job.

This adds two write paths so the interval takes effect without a restart:

- **`upsertMemoryRetentionJob(int)`** re-indexes the job doc with `opType(INDEX)` (version-bumping, not gated by the `CREATE` latch). The version bump is observed by JobScheduler's `JobSweeper.postIndex`, which deschedules the stale job and reschedules from the fresh `IntervalSchedule`.
- **`reconcileMemoryRetentionJob(int)`** runs on startup and upserts only when the persisted interval differs, so a value from `settings.yml` / a restart is applied without re-anchoring the next run on every boot.

A settings-update consumer wires live `PUT _cluster/settings` changes to the upsert. All new writes to the shared, fixed-id job document are gated on the elected cluster manager (single writer) and mirror the startup path's guards: agentic memory enabled, multi-tenancy disabled, and at least one 3.1+ data node present (rolling-upgrade deferral). The `[1, 168]` bound on the setting is enforced by core before the consumer fires.

### Related Issues

Part of #4859

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
@@ -60,6 +60,44 @@ public class MLCommonsClusterEventListener implements ClusterStateListener {
         this.mlModelAutoReDeployer = mlModelAutoReDeployer;
         this.client = client;
         this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+
+        // Apply live changes to the retention job interval (dynamic PUT _cluster/settings) by upserting the persisted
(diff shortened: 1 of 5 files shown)


AirFunnyB:
Truthfully not a bug just adding an extra feature that is not critical.

PR #4930, opened 2026-07-24
[BUG] Fix update memory container masking 404/403 as 500

AirFunnyB:
### Description

The failure lambda in `TransportUpdateMemoryContainerAction.doExecute()` discarded the exception from `MemoryContainerHelper.getMemoryContainer()` and always emitted a blanket 500. This masked the 404 raised for a missing container and the 403 raised on access denial. Now the original exception is forwarded when it already carries an HTTP status (`OpenSearchException` / `OpenSearchStatusException`); only genuinely unexpected exceptions are wrapped as 500.

**Repro:** `PUT /_plugins/_ml/memory_containers/does-not-exist` returned 500; now returns 404.

### Related Issues

N/A

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
+++ b/plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
@@ -27,6 +27,8 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchException;
+import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.search.ShardSearchFailure;
 import org.opensearch.cluster.node.DiscoveryNode;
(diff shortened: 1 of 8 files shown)


plugin/src/test/java/org/opensearch/ml/action/memorycontainer/TransportDeleteMemoryContainerActionTests.java
```diff
@@ -229,7 +230,12 @@ public void testDeleteMemoryContainer_Failure() {
...
-        assertEquals("Failed to find memory container", argumentCaptor.getValue().getMessage());
+        // A genuine delete failure must NOT be masked as a misleading 404 "Failed to find memory container".
+        // A plain unexpected exception surfaces as a sanitized 500.
+        Exception captured = argumentCaptor.getValue();
+        assertEquals(RestStatus.INTERNAL_SERVER_ERROR, ExceptionsHelper.status(captured));
+        assertTrue(captured instanceof OpenSearchStatusException);
+        assertEquals("Internal server error", captured.getMessage());
     }
 
     public void testDeleteMemoryContainer_IndexNotFoundException() {
```
nathaliellenaa, line 241: Can you add another delete-path test for a non-existent memory container ID, asserting the underlying 404 is surfaced rather than masked as 500.

nathaliellenaa (approved):
Thanks @AirFunnyB , LGTM! Added one non-blocking comment

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -383,4 +385,37 @@ public static void putMcpRequestHeaders(RestRequest request, Client client) {
...
+     * Everything else — including {@code 5xx} {@link OpenSearchException}s whose messages may embed internal
+     * infrastructure detail (node addresses, shard routing, cluster-block reasons) and any other unexpected exception —
+     * is replaced with a generic {@code 500 Internal server error}. This matches the 4xx-only forwarding pattern used by
+     * the sibling failure handlers in the memory-container transport actions and keeps server internals out of client
+     * responses.
+     *
+     * @param e the exception received by the failure handler
+     * @return the exception to pass to {@link ActionListener#onFailure(Exception)}
+     */
+    public static Exception wrapAsStatusException(Exception e) {
```
jiapingzeng, line 409: Glad to see a helper method introduced for this functionality that is commonly used! I see that you have already replaced some of older usages to this helper, wondering if you could also update the other usages to use the helper too?

Example, there are probably a few more similar ones in the codebase: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java#L212-L224

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -383,4 +385,37 @@ public static void putMcpRequestHeaders(RestRequest request, Client client) {
             }
         }
     }
+
+    /**
+     * Wraps an exception for an {@link ActionListener} failure so that the client receives a meaningful HTTP status
+     * instead of a blanket 500, while never leaking server-side error detail.
+     * <p>
+     * The REST layer derives the response status via {@link ExceptionsHelper#status(Throwable)}. This helper forwards
```
jiapingzeng, line 393: Would it make sense to use `ExceptionsHelper.status()` in this helper so that we do not need to handle `IllegalArgumentException` and `OpenSearchException` separately?

Something like:
```java
public static Exception wrapAsStatusException(Exception e) {
    RestStatus status = ExceptionsHelper.status(e);
    if (status.getStatus() >= 400 && status.getStatus() < 500) {
        return e;
    }
    return new OpenSearchStatusException("Internal server error", RestStatus.INTERNAL_SERVER_ERROR);
}
```

jiapingzeng (approved):
LGTM! Small suggestions below.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
AirFunnyB: wrapAsStatusException now uses ExceptionsHelper.status()

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
AirFunnyB: Yup good call out. Applied

PR #4937, opened 2026-07-27
[Memory Retention] Fix yellow cluster during rolling upgrade

AirFunnyB:
## Description

PR #4918 registers the memory-retention job from `MLCommonsClusterEventListener`, which auto-creates the `.plugins-ml-jobs` index. Because it was gated only on `agentic_memory_enabled` (default `true`), the first upgraded node in a rolling upgrade creates the index; its primary lands on the new-version node and the replica can't allocate to older nodes (`NodeVersionAllocationDecider`), so the cluster stays **yellow** until the upgrade finishes — failing the neural-search rolling-upgrade BWC test.

## Fix

Only touch `.plugins-ml-jobs` when it's safe:
- the index already exists (writing a doc can't strand a replica), or
- every node is `>= Version.CURRENT` (`getMinNodeVersion().onOrAfter(CURRENT)`), so a new index's replicas can allocate anywhere.

Mid-upgrade the jobs are **deferred, not skipped** — the state change when the last old node leaves re-runs the listener. Same fix covers the stats-collector job.

Also made retention **opt-in**: now gated on `memory.retention_enabled` (default `false`) too. Enabling it at runtime re-runs the listener (dynamic setting), so no restart needed.

## Testing

`MLCommonsClusterEventListenerTests` rewritten for multi-node clusters, tests passing: mixed-version defers, index-exists still registers, deferred-then-registered, retention-disabled no-op, plus preserved multi-tenancy / agentic-disabled / stats cases.

## Related

RFC #4859. Follows #4914, #4918.

## Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
+++ b/plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
@@ -83,34 +83,38 @@ public class MLCommonsClusterEventListener implements ClusterStateListener {
         }
 
         /*
-         * In version 3.1, a new index `.plugins-ml-jobs` replaces the old `.ml_commons_task_polling_job` index for the job scheduler.
-         * Version 3.1 also introduces a stats collector job that should run at startup if the relevant settings are enabled.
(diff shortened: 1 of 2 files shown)


jiapingzeng (approved):
LGTM and thanks for the quick fix!

Issue #4938, opened 2026-07-27
[BUG] Three minor pre-existing MCP defects found in 3.8 bug bash: empty tool input, unparseable tools-index doc, all-or-nothing bulk remove


rithinpullela:
### What is the bug?

Three defects in the MCP surface, filed together so they can be triaged as one small batch of hardening fixes. All three are minor. Only item 1 is reachable through supported API usage. Items 2 and 3 require either out-of-band index state or a malformed request.

---

### 1. MCP tool invoked with an empty `input` fails with `Unsupported response type`

`McpStreamableHttpTool.run` and `McpSseTool.run` pass the `input` parameter straight to `StringUtils.fromJson(input, "input")`, which accepts only a JSON object or array. An empty string parses to a `MissingNode`, so it falls through to `throw new IllegalArgumentException("Unsupported response type")`.

This matters because for a no-argument MCP tool, the LLM in a chat agent chooses `action_input` itself, and emitting `""` is a natural choice. Observed on an unprompted run: the model emitted `action_input: ""` for a no-arg tool, the tool call failed, and the agent burned an iteration recovering. The same tool succeeds when the model emits `{}`.

Reproduce on a cluster with an MCP connector registered:

```bash
# A flow agent whose MCP tool is configured with an empty input
POST /_plugins/_ml/agents/_register
{
  "name": "empty_input_repro",
  "type": "flow",
  "tools": [{
    "type": "McpStreamableHttpTool",
    "name": "aws___list_regions",
    "description": "x",
    "parameters": {"input": ""}
  }],
  "parameters": {"mcp_connectors": "[{\"mcp_connector_id\":\"<CONNECTOR_ID>\"}]"}
}

POST /_plugins/_ml/agents/<AGENT_ID>/_execute
{"parameters": {"question": "go"}}
```

Actual:

```json
{"status":400,"error":{"type":"IllegalArgumentException","reason":"Invalid Request","details":"Unsupported response type"}}
```

Expected: an empty or blank `input` is treated as no arguments, equivalent to `{}`, and the tool call succeeds.

Suggested fix: default a null, empty, or blank `input` to `{}` before calling `fromJson` in `McpStreamableHttpTool` and `McpSseTool`.

Note the frequency with which a given model emits `""` versus `{}` has not been measured. It was observed once unprompted and once forced, both on Claude Sonnet 4.5.

---

### 2. A single unparseable document in `.plugins-ml-mcp-tools` makes the whole MCP surface serve nothing

[...]

PR #4942, opened 2026-07-28
Harden MCP tool input handling, serving, and error status codes

rithinpullela:
## Description

Fixes three MCP defects reported in #4938. All three predate the 3.8 MCP changes.

### 1. Blank MCP tool input is treated as no arguments

`McpStreamableHttpTool` and `McpSseTool` passed `input` straight to `StringUtils.fromJson`, which accepts only a JSON object or array. A null, empty, or blank input therefore failed with `Unsupported response type`.

This is reachable in ordinary use: for a no-argument MCP tool the LLM chooses `action_input` itself, and emitting `""` is a natural choice. Observed on an unprompted run, where the tool call failed and the agent burned an iteration recovering. The same tool succeeds when the model emits `{}`.

A null, empty, or blank `input` is now equivalent to `{}`. Behaviour for all valid JSON object and array inputs is unchanged.

### 2. One unparseable document no longer takes down the whole MCP surface

`McpToolsHelper.searchAllTools` collected per-document parse errors and then failed the entire search if any document failed. Because `/_plugins/_ml/mcp` builds its tool list from that search, a single bad document broke `tools/list` and `tools/call` for every healthy tool.

This was asymmetric with `TransportMcpServerAction`, which already skips and logs individual tools that fail to *build* while healthy tools continue to serve. `searchAllTools` now skips and logs individual documents that fail to *parse*, matching that behaviour.

The catch covers `RuntimeException` as well as `IOException`. This is load-bearing: a document missing the required `type` field throws `IllegalArgumentException` from `McpToolRegisterInput`, and that, not a malformed-JSON `IOException`, is the failure mode actually reported in the issue. A narrower `catch (IOException)` does not fix the reported bug. The added test `test_searchAllTools_skipsToolMissingType` covers exactly this case, and I verified it fails without the broadened catch.

The leniency is deliberately confined to this serving read path, and there is a comment saying so. The write paths (`_register`, `_update`, `_remove`) continue to use the strict search methods, where a parse failure should still surface as an error.

### 3. Unknown tool name on `_remove` and `_update` returns 400 instead of 500

`POST /_plugins/_ml/mcp/tools/_remove` with a name that does not exist returned HTTP 500. A caller-supplied unknown name is a client error, not a server error, and a 5xx invites client retries that can never succeed.

`_update` carried the identical condition at two sites and is changed alongside `_remove`, so the MCP tool family stays internally consistent. Fixing only one would leave `_remove` at 400 and `_update` at 500 for the same condition.

`BAD_REQUEST` rather than `NOT_FOUND` is deliberate: removal and update are all-or-nothing, so an unknown name makes the whole request unprocessable rather than addressing one missing resource. Consistency within the MCP tool family was weighted above matching the sibling delete-model actions, which use 404.

## Behaviour changes for reviewers

Two user-visible changes worth calling out explicitly:

- `tools/list` and `GET /_plugins/_ml/mcp/tools/_list` may now return a **partial** result when a document in `.plugins-ml-mcp-tools` cannot be parsed. Previously the request failed outright. The skip is logged at `ERROR` with the index name and document id. There is no API field indicating a partial result, so the log is the only signal. This trade was made because one poison document previously caused a total MCP outage for every healthy tool, which is the worse failure mode.
- `_remove` and `_update` return **400 instead of 500** for an unknown tool name. This is technically a breaking change for a client keying on 500, though such a client is keying on a bug. There is recent precedent on this branch: #4930 applied the same class of fix across 9 sites.

Deliberately **not** changed: the all-or-nothing removal and update semantics. Only the status code changed.

## Testing

Unit tests added or updated in `McpToolsHelperTests`, `McpStreamableHttpToolTests`, `McpSseToolTests`, `TransportMcpToolsRemoveActionTests`, and `TransportMcpToolsUpdateActionTests`. Covers null, empty, blank, and valid `{}` input with assertions on the resulting empty argument map; `searchAllTools` with a document missing `type`, with malformed JSON, and with all documents unparseable; and the 400 status on both the none-found and partial-found branches of remove and update.

Each fix was checked to be rollback-sensitive rather than merely passing. For fix 1 I reverted the one-line change in place and confirmed `testRunNoArguments` fails, then restored it. For fix 2 I confirmed the missing-`type` test fails under the previous `catch (IOException)`.

Verified end to end on a live 2-node cluster:

- A document missing `type` injected directly into `.plugins-ml-mcp-tools` no longer breaks `tools/list` on either node; healthy tools remain listable and callable; the malformed-JSON case is also skipped.
- `_remove` and `_update` return 400 for an unknown name and 200 for a valid request, and the tool that does exist is still not removed, confirming the semantics are unchanged.

`:opensearch-ml-algorithms:test --tests "*Mcp*"`, `:opensearch-ml-plugin:test --tests "*mcpserver*"`, `MLFlowAgentRunnerTest`, `MLChatAgentRunnerTest`, and `spotlessJavaCheck` all pass on JDK 21.

## Follow-ups not included here

- Integration test coverage for the 400 status and the poison-document skip. Both are unit-tested only; `RestMcpToolsRemoveActionIT` asserts no status for the not-found case.
- `CallToolResult.isError()` is still discarded by both MCP tools, so a tool-level error from a remote MCP server is returned to the agent as a successful result. Handling it turned out to be a larger change than it appears, because `MLFlowAgentRunner` routes tool failures to a fatal `onFailure`, so throwing would abort an entire flow-agent run on a recoverable per-call error. That needs its own decision on flow-agent semantics and is left out of this PR.

### Check List

### DIFF
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpSseTool.java
@@ -61,7 +61,7 @@ public class McpSseTool implements WithModelTool {
         try {
             Map<String, String> parameters = ToolUtils.extractInputParameters(originalParameters, attributes);
             String input = parameters.get("input");
-            Map<String, Object> inputArgs = StringUtils.fromJson(input, "input");
+            Map<String, Object> inputArgs = StringUtils.fromJson(input == null || input.isBlank() ? "{}" : input, "input");
             McpSchema.CallToolResult result = mcpSyncClient.callTool(new McpSchema.CallToolRequest(this.name, inputArgs));
             String resultJson = StringUtils.toJson(result.content());
             listener.onResponse((T) resultJson);
--- a/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpStreamableHttpTool.java
+++ b/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/McpStreamableHttpTool.java
@@ -60,7 +60,7 @@ public class McpStreamableHttpTool implements WithModelTool {
         try {
             Map<String, String> parameters = ToolUtils.extractInputParameters(originalParameters, attributes);
             String input = parameters.get("input");
-            Map<String, Object> inputArgs = StringUtils.fromJson(input, "input");
+            Map<String, Object> inputArgs = StringUtils.fromJson(input == null || input.isBlank() ? "{}" : input, "input");
             McpSchema.CallToolResult result = mcpSyncClient.callTool(new McpSchema.CallToolRequest(this.name, inputArgs));
             String resultJson = StringUtils.toJson(result.content());
             @SuppressWarnings("unchecked")
--- a/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/McpSseToolTests.java
+++ b/ml-algorithms/src/test/java/org/opensearch/ml/engine/tools/McpSseToolTests.java
@@ -11,6 +11,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.opensearch.ml.common.CommonValue.MCP_SYNC_CLIENT;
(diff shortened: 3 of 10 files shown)


ylwu-amzn:
Confirming this class of defect is still present in 3.9.0 RC3, since this PR is not merged: removing a tool that does not exist returns **HTTP 500**, not 404/400.

```
POST /_plugins/_ml/mcp/tools/_remove
["does_not_exist"]
-> 500  "Unable to remove tools as no tool ... found"
```

Found during the OS 3.9 RC3 bug bash (credit: @mingshl) while verifying #5034. Two adjacent null-payload defects from the same run are filed separately as #5081 (`"name": null` on register, and `"query_text": null` on TEXT_SIMILARITY predict, both 500 instead of 400) — if this PR is going to normalise MCP error statuses, the register-side `parser.text()` on a `VALUE_NULL` belongs in the same pass.

Issue #4925, opened 2026-07-22
[RFC] Agentic search templates: param-schema index + registration for search-template fill


aditya0shah:
> **This is the control-plane half of a two-repo feature.** The feature, *search-template filling* for agentic search, has its canonical design and its query-path (agent-server) implementation in **opensearch-project/opensearch-agent-server#150**. That RFC explains *why* the model fills templates instead of writing DSL and how a live query flows. **This** RFC specifies only the ml-commons control plane: a new system index for per-template metadata, a CRUD API to manage it, and the registration logic that derives that metadata from a stored search template plus its index mapping. No ml-commons Java runs on the query path.

## Summary

Search-template filling lets an LLM fill the parameters of a registered OpenSearch search template (content words, a filter value, a sort field) instead of authoring a full `_search` body; OpenSearch then renders those values into the stored Mustache body. For the model to fill a template safely, it needs a **typed description of that template's parameters**: which params exist, their types, which are required, and what values (e.g. field-name enums) are legal. That description can't be attached to a core `_scripts` template and can't be reliably re-parsed from the body on every query, so ml-commons stores it.

This RFC adds:

1. A dedicated system index, **`.plugins-ml-agentic-search-templates`**, holding one **param-schema** document per template.
2. A **CRUD REST API** (`register`, `get`, `update`, `list`, `delete`) under `/_plugins/_ml/agentic_search_templates` to manage those documents.
3. **Registration logic** that, from just a template name + target index, fetches the stored Mustache body from core `_scripts`, fetches the index mapping, derives a typed param-schema (via a Mustache tag scanner + mapping-derived field enums), pre-flight-renders the body to prove it's valid, and stores the result.

All of this is **registration-time / control-plane** work. 

## Motivation

> Full motivation for template *filling* (the latency and reliability payoff, and why a template should be a contract rather than a hint) is in the canonical agent-server RFC. This section covers only why the metadata belongs in ml-commons.

To fill a template, the agent server builds a typed schema (a Pydantic `FillTemplate` model) whose fields are the template's Mustache parameters. That schema is assembled from three inputs: the template body (param names, types, required-ness), the index mapping (which fields a param may target), and optional human enrichment (value enums, descriptions). Two facts follow:

1. **It can't live on the `_scripts` template.** The Mustache body is owned by opensearch-core and referenced only by id; there's nowhere to attach our metadata.
[...]

PR #4944, opened 2026-07-29
Add agentic search template param-schema derivation

aditya0shah:
### Description

Introduces the metadata layer for agentic search templates: given a stored Mustache search template, derive the typed parameter schema an LLM needs in order to fill it, and persist that schema alongside the template.

The Mustache body itself stays in core `_scripts`; this adds ml-commons metadata beside it, one document per template in a new system index keyed by the template name.

**What's here**

- `MustacheTemplateAnalyzer` walks the template body and infers each parameter's name, type, and whether it is required. A name used only as a section guard is a boolean; a triple-stache is an injected JSON value; a parameter is required only when it is a bare value at root scope never wrapped in its own section, since anything wrapped in a section renders conditionally.
- `AgenticSearchTemplateService` combines that parse with the target index's mapping, scoping field-selector parameters to the mapping's field names so a model can only target a field that exists.
- Registration pre-flight validates by rendering the body twice — all-filled and required-only — through the cluster's own Mustache engine, so a body that cannot render legal JSON fails at registration rather than on first query.
- A schema edit is validated the same way, against the merged schema, and additionally checked for internal consistency: an enum whose values do not fit the declared type, an empty enum, or a non-boolean `required` flag are all rejected. Without this an edit could be accepted and then silently break every subsequent render.

This PR is the derivation and validation core only; the transport and REST surface follows in a separate PR so this can be reviewed on its own. It compiles and tests standalone.

### Related Issues

Related to #4925

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLIndex.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLIndex.java
@@ -7,6 +7,8 @@ package org.opensearch.ml.common;
 
 import static org.opensearch.ml.common.CommonValue.MCP_SESSION_MANAGEMENT_INDEX;
 import static org.opensearch.ml.common.CommonValue.MCP_TOOLS_INDEX;
+import static org.opensearch.ml.common.CommonValue.ML_AGENTIC_SEARCH_TEMPLATES_INDEX;
+import static org.opensearch.ml.common.CommonValue.ML_AGENTIC_SEARCH_TEMPLATES_INDEX_MAPPING_PATH;
 import static org.opensearch.ml.common.CommonValue.ML_AGENT_INDEX;
 import static org.opensearch.ml.common.CommonValue.ML_AGENT_INDEX_MAPPING_PATH;
 import static org.opensearch.ml.common.CommonValue.ML_CONFIG_INDEX;
@@ -56,7 +58,8 @@ public enum MLIndex {
     MCP_TOOLS(MCP_TOOLS_INDEX, false, ML_MCP_TOOLS_INDEX_MAPPING_PATH),
     JOBS(ML_JOBS_INDEX, false, ML_JOBS_INDEX_MAPPING_PATH),
     INDEX_INSIGHT_CONFIG(ML_INDEX_INSIGHT_CONFIG_INDEX, false, ML_INDEX_INSIGHT_CONFIG_INDEX_MAPPING_PATH),
-    INDEX_INSIGHT_STORAGE(ML_INDEX_INSIGHT_STORAGE_INDEX, false, ML_INDEX_INSIGHT_STORAGE_INDEX_MAPPING_PATH);
+    INDEX_INSIGHT_STORAGE(ML_INDEX_INSIGHT_STORAGE_INDEX, false, ML_INDEX_INSIGHT_STORAGE_INDEX_MAPPING_PATH),
+    AGENTIC_SEARCH_TEMPLATES(ML_AGENTIC_SEARCH_TEMPLATES_INDEX, false, ML_AGENTIC_SEARCH_TEMPLATES_INDEX_MAPPING_PATH);
 
     private final String indexName;
     // whether we use an alias for the index
--- a/plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
@@ -1,0 +1,601 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.action.agenticsearch;
+
+import static org.opensearch.common.xcontent.json.JsonXContent.jsonXContent;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.OpenSearchStatusException;
+import org.opensearch.action.admin.cluster.storedscripts.GetStoredScriptRequest;
+import org.opensearch.action.admin.indices.get.GetIndexRequest;
+import org.opensearch.action.admin.indices.get.GetIndexResponse;
+import org.opensearch.action.delete.DeleteRequest;
+import org.opensearch.action.delete.DeleteResponse;
+import org.opensearch.action.get.GetRequest;
+import org.opensearch.action.index.IndexRequest;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.support.IndicesOptions;
+import org.opensearch.action.support.WriteRequest;
+import org.opensearch.action.update.UpdateRequest;
+import org.opensearch.action.update.UpdateResponse;
+import org.opensearch.cluster.metadata.MappingMetadata;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
+import org.opensearch.commons.authuser.User;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.core.xcontent.MediaTypeRegistry;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.IndexNotFoundException;
+import org.opensearch.index.query.MatchAllQueryBuilder;
+import org.opensearch.ml.common.CommonValue;
+import org.opensearch.ml.common.MLIndex;
+import org.opensearch.ml.common.agenticsearch.AgenticSearchTemplate;
+import org.opensearch.ml.engine.indices.MLIndicesHandler;
+import org.opensearch.script.Script;
+import org.opensearch.script.ScriptService;
+import org.opensearch.script.ScriptType;
+import org.opensearch.script.TemplateScript;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.transport.client.Client;
+
+import lombok.extern.log4j.Log4j2;
+
+/**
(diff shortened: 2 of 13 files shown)


jiapingzeng:
Both Copilot comments seem valid, could you take a look?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,601 @@
...
+                }
+                // Expose a text field's keyword sub-field (used for exact/sort).
+                if (field.get("fields") instanceof Map) {
+                    for (String sub : ((Map<String, Object>) field.get("fields")).keySet()) {
+                        out.add(name + "." + sub);
+                    }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
aditya0shah: Good catch — fixed in 4d22928.

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,601 @@
...
+                return value instanceof String;
+        }
+    }
+
+    private AgenticSearchTemplate parse(BytesReference source) throws Exception {
+        XContentParser parser = MediaTypeRegistry.JSON
+            .xContent()
+            .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, source.streamInput());
+        return AgenticSearchTemplate.parse(parser);
+    }
```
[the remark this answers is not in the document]
aditya0shah: Fixed in 4d22928 — wrapped in try-with-resources, matching `renderAndCheckJson` in the same class.

aditya0shah:
Thanks @jiapingzeng — both were valid, addressed in 4d22928.

common/src/main/java/org/opensearch/ml/common/MLIndex.java
```diff
@@ -56,7 +58,8 @@ public enum MLIndex {
     MCP_TOOLS(MCP_TOOLS_INDEX, false, ML_MCP_TOOLS_INDEX_MAPPING_PATH),
     JOBS(ML_JOBS_INDEX, false, ML_JOBS_INDEX_MAPPING_PATH),
     INDEX_INSIGHT_CONFIG(ML_INDEX_INSIGHT_CONFIG_INDEX, false, ML_INDEX_INSIGHT_CONFIG_INDEX_MAPPING_PATH),
-    INDEX_INSIGHT_STORAGE(ML_INDEX_INSIGHT_STORAGE_INDEX, false, ML_INDEX_INSIGHT_STORAGE_INDEX_MAPPING_PATH);
+    INDEX_INSIGHT_STORAGE(ML_INDEX_INSIGHT_STORAGE_INDEX, false, ML_INDEX_INSIGHT_STORAGE_INDEX_MAPPING_PATH),
+    AGENTIC_SEARCH_TEMPLATES(ML_AGENTIC_SEARCH_TEMPLATES_INDEX, false, ML_AGENTIC_SEARCH_TEMPLATES_INDEX_MAPPING_PATH);
```
jiapingzeng, line 62: Should we also register the new index here: https://github.com/opensearch-project/ml-commons/blame/main/plugin/src/main/java/org/opensearch/ml/plugin/MachineLearningPlugin.java#L1558-L1579

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+     *
+     * @param templateId the {@code _scripts} template name (also the doc id)
+     * @param index the target index (for field-name enums)
+     * @param description optional human description
+     * @param user the caller, captured by the transport action before it stashed the
+     *     thread context (reading it here would be too late — the stash below drops the
+     *     security-user transient); may be null when security is disabled
+     * @param listener yields the stored {@link AgenticSearchTemplate}
+     */
+    public void register(String templateId, String index, String description, User user, ActionListener<AgenticSearchTemplate> listener) {
```
jiapingzeng, line 110: Are we able to add tests for these?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+ * the index mapping, (3) derives the param-schema — {@link MustacheTemplateAnalyzer}
+ * for names/types/required, the mapping for field-name enums — and (4) pre-flight
+ * validates by rendering the body twice (all-filled and required-only) through the
+ * cluster's own Mustache engine, so a broken body fails here, not on the customer's
+ * first query. The derived schema is then the customer's editable tuning surface.
+ */
+@Log4j2
+public class AgenticSearchTemplateService {
+
+    private static final int DEFAULT_MAX_TEMPLATES = 1000;
```
jiapingzeng: This doesn't seem to get used anywhere

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/MustacheTemplateAnalyzer.java
```diff
@@ -0,0 +1,240 @@
...
+ *   <tr><td>{@code {{{x}}}} (triple)</td><td>unescaped scalar (often an array/object injected as JSON)</td></tr>
+ * </table>
+ *
+ * The output is the {@code param_schema} map the customer then enriches (value
+ * enums, descriptions) and the field-name enums the index mapping supplies.
+ */
+public class MustacheTemplateAnalyzer {
+
+    /** param-schema entry keys, matching what the agent server reads. */
+    static final String TYPE_KEY = "type";
```
jiapingzeng, line 43: Seems that these constants are defined in both here and `AgenticSearchTemplateService`, could we define them in a central place?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+                        templates.add(parse(hit.getSourceRef()));
+                    }
+                    long total = response.getHits().getTotalHits() != null ? response.getHits().getTotalHits().value() : templates.size();
+                    wrapped.onResponse(new MLListResult(templates, total));
+                } catch (Exception e) {
+                    wrapped.onFailure(e);
+                }
+            }, e -> {
+                if (e instanceof IndexNotFoundException) {
+                    wrapped.onResponse(new MLListResult(new ArrayList<>(), 0));
```
jiapingzeng, line 399: What scenarios can be index be missing here?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+        GetIndexRequest request = new GetIndexRequest().indices(index).indicesOptions(IndicesOptions.strictExpand()).local(false);
+        client.admin().indices().getIndex(request, ActionListener.wrap(response -> {
+            try {
+                listener.onResponse(extractFieldNames(response));
+            } catch (Exception e) {
+                listener.onFailure(e);
+            }
+        }, e -> {
+            if (e instanceof IndexNotFoundException) {
+                listener.onFailure(new OpenSearchStatusException("Index does not exist: " + index, RestStatus.BAD_REQUEST));
```
jiapingzeng, line 298: Why is this one BAD_REQUEST while the other ones are NOT_FOUND?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+                spec.put(SOURCE_KEY, "mapping");
+            }
+        }
+        return schema;
+    }
+
+    /** Heuristic: a param that selects a field (name ends with _field/_by or is "field"). */
+    private static boolean targetsAField(String name) {
+        String n = name.toLowerCase(java.util.Locale.ROOT);
+        return n.equals("field") || n.endsWith("_field") || n.endsWith("_by") || n.equals("sort_by");
```
jiapingzeng: this condition seems very broad, can specific fields be targetted instead?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
```diff
@@ -0,0 +1,616 @@
...
+            }));
+        } catch (Exception e) {
+            listener.onFailure(e);
+        }
+    }
+
+    private void writeTemplatePatch(String templateId, AgenticSearchTemplate patch, ActionListener<UpdateResponse> listener) {
+        patch.setLastUpdatedTime(Instant.now());
+        try {
+            UpdateRequest updateRequest = new UpdateRequest(INDEX, templateId)
```
jiapingzeng, line 502: Could there be some race condition here? Generally for update after get, we want to do something like this: https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/model/MLModelManager.java#L449-L450

jiapingzeng (commented):
Overall looks good! Added some comments.

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/MustacheTemplateAnalyzer.java
aditya0shah: Yes centralized to `MustacheTemplateAnalyzer`

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
aditya0shah: The two BAD_REQUESTs in register are for caller-supplied inputs that don't exist: the _scripts body and the target index. The NOT_FOUNDs are for the template resource itself being absent on get/update/delete.

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
aditya0shah: The system index is created on the first successful register. A list issued before any template has been registered hits a not-yet-created index, so it returns an empty page rather than a 404

jiapingzeng (approved):
LGTM!

PR #4945, opened 2026-07-29
Add CRUD API for agentic search templates

aditya0shah:
### Description

Exposes the param-schema layer over transport and REST so a customer can register a stored search template for filling, then read, list, edit, or remove its schema.

```
POST   /_plugins/_ml/agentic_search_templates
GET    /_plugins/_ml/agentic_search_templates/{template_id}
GET    /_plugins/_ml/agentic_search_templates
PUT    /_plugins/_ml/agentic_search_templates/{template_id}
DELETE /_plugins/_ml/agentic_search_templates/{template_id}
```

Registration takes only the link — the `_scripts` template name, the target index, and an optional description — because the schema is derived server-side rather than supplied by the caller. The update action takes a partial schema so a customer can tune one parameter's description or enum without restating the rest.

Adds the five transport action/request/response triples, their transport handlers, the REST handlers, and the plugin wiring. Gated on the agent framework flag.

**Stacked on #4944** — please review that one first. This branch contains both commits, so the diff below includes #4944's changes; it will shrink to just this PR's commit once the base merges. Draft until then.

### Related Issues

Related to #4925

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/transport/agenticsearch/MLRegisterAgenticSearchTemplateRequest.java
+++ b/common/src/main/java/org/opensearch/ml/common/transport/agenticsearch/MLRegisterAgenticSearchTemplateRequest.java
@@ -11,6 +11,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.Map;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
@@ -26,12 +27,13 @@ import lombok.ToString;
 import lombok.experimental.FieldDefaults;
 
 /**
- * Register a search template for filling. Per the design (§4.5) the customer sends
- * only the link — the {@code template_id} (the existing {@code _scripts} template id),
- * the target index, and a description — and registration derives the param-schema. So
- * this request carries no schema; it is produced server-side from the template body +
- * index mapping. The {@code templateId} is the single identifier: it is the stored
- * script id, the system-index doc id, and what get/update/delete address.
+ * Register a search template for filling. The request carries the {@code templateId}
+ * (the existing {@code _scripts} template id), the target index, and an optional
+ * description. Registration derives the param-schema server-side from the template body
+ * and index mapping. A caller may instead supply {@code paramSchema} directly; when set
+ * it is validated and pre-flight rendered against the template body, then stored without
+ * derivation. The {@code templateId} is also the system-index doc id and the id used by
+ * get, update, and delete.
  */
 @Getter
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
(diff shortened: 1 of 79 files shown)


mingshl:
what is the reason-ale behind the search templates in ml common plugin rather than neural search plugin? this is unlike other ml apis.

aditya0shah:
@mingshl Main reason is neural-search is stateless so this would be the first system index there. That means owning index creation, mappings, upgrade migration and security registration, all for one doc per template. ml-commons already has all of that.

jiapingzeng:
Please gate the new APIs behind a feature flag that is disabled by default. We need security review before we can enable new APIs.

common/src/main/java/org/opensearch/ml/common/transport/agenticsearch/MLRegisterAgenticSearchTemplateRequest.java
```diff
@@ -0,0 +1,91 @@
...
+ * no schema; it is produced server-side from the template body + index mapping.
+ */
+@Getter
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@ToString
+public class MLRegisterAgenticSearchTemplateRequest extends ActionRequest {
+
+    String templateId;
+    String index;
+    String description;
```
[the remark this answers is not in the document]
aditya0shah: Collapsed template_name → template_id. There's a single identifier (the _scripts id = doc id = path param); register no longer implies a separate generated id.

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateAgenticSearchTemplateAction.java
```diff
@@ -0,0 +1,87 @@
...
+        if (templateId == null || templateId.trim().isEmpty()) {
+            throw new IllegalArgumentException("Template id is required");
+        }
+        if (!request.hasContent()) {
+            throw new IllegalArgumentException("Update body is required");
+        }
+
+        XContentParser parser = request.contentParser();
+        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+        AgenticSearchTemplate patch = AgenticSearchTemplate.parse(parser);
```
jiapingzeng: Do we want to allow user to modify all fields of `AgenticSearchTemplate`?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/AgenticSearchTemplateService.java
jiapingzeng, line 175: This is from previous PR but we should do `OpType.CREATE` here so if id exists it does not get replaced.

jiapingzeng (commented):
Overall looks good!

plugin/src/main/java/org/opensearch/ml/rest/RestMLUpdateAgenticSearchTemplateAction.java
aditya0shah: Yes good catch, now only allowing updates to description and param_schema

jiapingzeng (approved):
LGTM!

PR #4952, opened 2026-08-04
[Memory] Retention job dynamic interval fix + dry-run API

AirFunnyB:
Title:
  [Memory] Retention job dynamic interval fix + dry-run API (RFC #4859)

  Description:
  ### Description
  
  Part 1 of 2 (stacked) for the agentic memory retention lifecycle, per RFC #4859.

  **Dynamic retention job interval**
  - Fixes the retention job interval setting (`plugins.ml_commons.memory.retention_job_interval_hours`) being effectively *write-once*: the scheduled job document was
  seeded on first startup and never reconciled, so changing the cluster setting (or `settings.yml` on restart) silently had no effect on the next run.
  - Reconciles on startup by reading the effective interval (cluster settings layered over node settings, honoring OpenSearch precedence) and upserting the persisted job
  doc only when it differs — gated on the elected cluster manager so exactly one node writes the shared doc.
  - Defers retention/stats job scheduling until cluster state is recovered and the jobs index primary shards are active, so a rolling upgrade does not leave the cluster
  yellow. Builds on and preserves the rolling-upgrade guard from #4937.
  
  **Dry-run API**
  - Adds `POST /_plugins/_ml/memory_containers/{memory_container_id}/_retention/_dry_run` (single container) and `POST /_plugins/_ml/memory/_retention/_dry_run`
  (cluster-wide) to preview exactly what the retention job *would* delete — without deleting anything.
  - The dry-run path reuses the real job's eviction-counting logic (shared `computeDefaultRetentionPolicy()`) and performs zero deletes. Response reports `effective_policy`, per-type `would_delete` with `by_reason` breakdowns, `total_would_delete`, `pinned_would_skip`, and `warnings`.
  - Cluster-wide response is an object { "skipped_count": N, "results": [ … ] }, where skipped_count is the number of containers that could not be evaluated (parse/eval error) and were omitted from results, so callers can tell "no deletions" from "not evaluated." The single-container response is a bare result object.

  The existing `agentic_memory_enabled` / `memory_retention_enabled` / multi-tenancy gates are retained unchanged. A follow-up stacked PR (on-demand execute +
  remote-metadata gating) builds on this branch.

  ### Related Issues
  Relates to RFC #4859
  
  ### Check List
  - [ ] API changes companion pull request created.
  - [ ] Public documentation issue/PR created.

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
@@ -135,7 +135,7 @@ public class MemoryContainerConstants {
         + "/{"
         + PARAMETER_MEMORY_CONTAINER_ID
         + "}/_retention/_dry_run";
-    public static final String RETENTION_DRY_RUN_ALL_PATH = "/_plugins/_ml/memory/_retention/_dry_run";
+    public static final String RETENTION_DRY_RUN_ALL_PATH = BASE_MEMORY_CONTAINERS_PATH + "/_retention/_dry_run";
 
     // Memory types are defined in MemoryType enum
     // Memory strategy types are defined in MemoryStrategyType enum
--- a/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
+++ b/common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
@@ -130,6 +130,12 @@ public class MemoryContainerConstants {
     public static final String DELETE_MEMORIES_BY_QUERY_PATH = MEMORIES_PATH + "/{" + PARAMETER_MEMORY_TYPE + "}" + "/_delete_by_query";
     public static final String SEMANTIC_SEARCH_MEMORIES_PATH = MEMORIES_PATH + "/long-term/_semantic_search";
     public static final String HYBRID_SEARCH_MEMORIES_PATH = MEMORIES_PATH + "/long-term/_hybrid_search";
+    // Retention dry-run: preview what the scheduled retention job would delete, without deleting.
+    public static final String RETENTION_DRY_RUN_PATH = BASE_MEMORY_CONTAINERS_PATH
+        + "/{"
+        + PARAMETER_MEMORY_CONTAINER_ID
+        + "}/_retention/_dry_run";
+    public static final String RETENTION_DRY_RUN_ALL_PATH = "/_plugins/_ml/memory/_retention/_dry_run";
 
     // Memory types are defined in MemoryType enum
     // Memory strategy types are defined in MemoryStrategyType enum
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
@@ -1,0 +1,314 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.ml.action.memorycontainer;
+
+import static org.opensearch.common.xcontent.json.JsonXContent.jsonXContent;
+import static org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ml.common.CommonValue.ML_MEMORY_CONTAINER_INDEX;
+import static org.opensearch.ml.common.memorycontainer.MemoryContainerConstants.ORPHAN_SWEEP_BASELINE_TIME_FIELD;
+import static org.opensearch.ml.common.settings.MLCommonsSettings.ML_COMMONS_AGENTIC_MEMORY_DISABLED_MESSAGE;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchStatusException;
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.get.GetRequest;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.action.support.IndicesOptions;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.commons.authuser.User;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.index.IndexNotFoundException;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.ml.common.memorycontainer.MLMemoryContainer;
+import org.opensearch.ml.common.settings.MLFeatureEnabledSetting;
+import org.opensearch.ml.common.transport.memorycontainer.MLMemoryRetentionDryRunAction;
+import org.opensearch.ml.common.transport.memorycontainer.MLMemoryRetentionDryRunRequest;
+import org.opensearch.ml.common.transport.memorycontainer.MLMemoryRetentionDryRunResponse;
+import org.opensearch.ml.common.transport.memorycontainer.MemoryRetentionDryRunResult;
+import org.opensearch.ml.helper.MemoryContainerHelper;
+import org.opensearch.ml.jobs.processors.MemoryRetentionJobProcessor;
+import org.opensearch.ml.utils.RestActionUtils;
+import org.opensearch.ml.utils.TenantAwareHelper;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.SortOrder;
+import org.opensearch.tasks.Task;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.TransportService;
+import org.opensearch.transport.client.Client;
+
+import lombok.AccessLevel;
+import lombok.experimental.FieldDefaults;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * Transport action for the retention dry-run: loads the target container(s), enforces access control, then
+ * delegates the read-only "what would be deleted" computation to {@link MemoryRetentionJobProcessor}. Performs
+ * no deletions.
+ */
+@Log4j2
+@FieldDefaults(level = AccessLevel.PRIVATE)
+public class TransportMemoryRetentionDryRunAction extends HandledTransportAction<ActionRequest, MLMemoryRetentionDryRunResponse> {
+
+    private static final int CONTAINER_PAGE_SIZE = 100;
+
+    final Client client;
+    final ClusterService clusterService;
+    final ThreadPool threadPool;
+    final NamedXContentRegistry xContentRegistry;
+    final MLFeatureEnabledSetting mlFeatureEnabledSetting;
+    final MemoryContainerHelper memoryContainerHelper;
+
+    @Inject
+    public TransportMemoryRetentionDryRunAction(
+        TransportService transportService,
+        ActionFilters actionFilters,
+        Client client,
+        ClusterService clusterService,
+        ThreadPool threadPool,
+        NamedXContentRegistry xContentRegistry,
+        MLFeatureEnabledSetting mlFeatureEnabledSetting,
+        MemoryContainerHelper memoryContainerHelper
+    ) {
+        super(MLMemoryRetentionDryRunAction.NAME, transportService, actionFilters, MLMemoryRetentionDryRunRequest::new);
+        this.client = client;
+        this.clusterService = clusterService;
+        this.threadPool = threadPool;
+        this.xContentRegistry = xContentRegistry;
+        this.mlFeatureEnabledSetting = mlFeatureEnabledSetting;
+        this.memoryContainerHelper = memoryContainerHelper;
+    }
+
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<MLMemoryRetentionDryRunResponse> actionListener) {
+        if (!mlFeatureEnabledSetting.isAgenticMemoryEnabled()) {
+            actionListener.onFailure(new OpenSearchStatusException(ML_COMMONS_AGENTIC_MEMORY_DISABLED_MESSAGE, RestStatus.FORBIDDEN));
+            return;
+        }
+
+        MLMemoryRetentionDryRunRequest dryRunRequest = MLMemoryRetentionDryRunRequest.fromActionRequest(request);
+        String tenantId = dryRunRequest.getTenantId();
+        if (!TenantAwareHelper.validateTenantId(mlFeatureEnabledSetting, tenantId, actionListener)) {
+            return;
+        }
+
+        User user = RestActionUtils.getUserContext(client);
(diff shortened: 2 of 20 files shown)


plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
```diff
@@ -60,6 +61,44 @@ public MLCommonsClusterEventListener(
...
+     * Single-writer gate for memory-retention job writes that mutate the shared, fixed-id job document. Exactly one
+     * node (the elected cluster manager) should ever upsert/reconcile, and only when agentic memory is enabled and
+     * multi-tenancy is disabled (mirroring the startup scheduling path). The 3.1+ data-node check mirrors the startup
+     * loop's rolling-upgrade guard so a live setting change during a mixed-version upgrade does not create the new
+     * {@code .plugins-ml-jobs} index before the cluster is ready for it.
+     */
+    private boolean shouldManageMemoryRetentionJob() {
+        return clusterService.state().nodes().isLocalNodeElectedClusterManager()
+            && hasDataNodeOnOrAfterV31()
+            && mlFeatureEnabledSetting.isAgenticMemoryEnabled()
```
nathaliellenaa, line 88: The interval-change consumer gates on `isAgenticMemoryEnabled()` while the startup path gates on `isMemoryRetentionEnabled()`, so an interval change can schedule a job doc that startup deliberately wouldn't.

Suggested fix: add `&& mlFeatureEnabledSetting.isMemoryRetentionEnabled()` to `shouldManageMemoryRetentionJob()`

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
```diff
@@ -0,0 +1,314 @@
...
+                    wrapped.onFailure(RestActionUtils.wrapAsStatusException(e));
+                }
+            }));
+        } catch (Exception e) {
+            log.error("Failed to run retention dry-run for container {}", containerId, e);
+            actionListener.onFailure(new OpenSearchStatusException("Internal server error", RestStatus.INTERNAL_SERVER_ERROR));
+        }
+    }
+
+    private void executeClusterWide(
```
nathaliellenaa, line 186: The paging loop accumulates into `results` across every page with no bound, and each container triggers a dozen-plus counting searches. On a deployment with many thousands of containers that's both an unbounded response held in heap until paging finishes, and tens of thousands of searches with no backpressure — the real job doing the same traversal paces itself with `ML_COMMONS_MEMORY_RETENTION_JOB_THROTTLE_SECONDS`, which the dry-run doesn't reuse.

Suggested fix: add a `MAX_CONTAINERS_PER_DRY_RUN `bound (e.g. 1000), stop paging there, and include a truncation warning naming how many containers weren't evaluated so it isn't silent.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
```diff
@@ -0,0 +1,321 @@
...
+
+        SearchHit hit = hits[index];
+        String containerId = hit.getId();
+        final MLMemoryContainer container;
+        try {
+            container = parseContainer(hit.getSourceAsString());
+        } catch (Exception e) {
+            log.warn("Skipping container {} in retention dry-run: failed to parse", containerId, e);
+            skipped[0]++;
+            processHitChain(processor, user, tenantId, hits, index + 1, nextPageSort, results, skipped, listener);
```
nathaliellenaa: `processHitChain` advances to the next container by recursing inside the `dryRunContainer` callback — safe with the real (async) client, so under a synchronous/mocked client the stack depth is the total container count, not one page. This is the same shape as the [OldestDocIdsPager](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java#L976) was introduced to fix.

Suggested fix: drive the traversal from an OldestDocIdsPager-style drain() loop (or a plain loop over an index cursor). `countWorkingForSessionsBatch`/`countWorkingForOrphanBatch` share the pattern but are genuinely bounded (≤100 frames from `COUNT_BASED_ID_CAP`/`DELETE_BY_QUERY_BATCH_SIZE`) — worth converting for consistency.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
```diff
@@ -0,0 +1,321 @@
...
+            return;
+        }
+
+        // Tenant + access filtering: silently skip containers the caller cannot see, mirroring the job's own
+        // per-container isolation. A mismatched tenant or denied access simply excludes the container.
+        boolean tenantOk = tenantId == null || container.getTenantId() == null || tenantId.equals(container.getTenantId());
```
nathaliellenaa: The `container.getTenantId() == null` clause means a container with no tenant stamp is visible to every tenant. `TenantAwareHelper.validateTenantResource` — which the single-container path uses at line 141, compares with `Objects.equals`, so it denies that case. Containers created before multi-tenancy was enabled would leak their id, retention policy, and per-type document counts across tenants.

Suggested fix:
```
 boolean tenantOk = !mlFeatureEnabledSetting.isMultiTenancyEnabled() || Objects.equals(tenantId, container.getTenantId());
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
```diff
@@ -0,0 +1,321 @@
...
+    final ThreadPool threadPool;
+    final NamedXContentRegistry xContentRegistry;
+    final MLFeatureEnabledSetting mlFeatureEnabledSetting;
+    final MemoryContainerHelper memoryContainerHelper;
+
+    @Inject
+    public TransportMemoryRetentionDryRunAction(
+        TransportService transportService,
+        ActionFilters actionFilters,
+        Client client,
```
nathaliellenaa, line 86: Container reads here use raw `Client` rather than `SdkClient` like the container CRUD actions do (`TransportSearchMemoryContainerAction`, or `MemoryContainerHelper.getMemoryContainer(id, tenantId, listener)` at MemoryContainerHelper) — which is also how the hand-rolled tenant check at line 267 arose. In a remote-metadata deployment this search returns zero hits, so the API reports an empty array rather than an error, reading as "nothing to clean up." Is this intentional?

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
```diff
@@ -130,6 +130,12 @@ public class MemoryContainerConstants {
     public static final String DELETE_MEMORIES_BY_QUERY_PATH = MEMORIES_PATH + "/{" + PARAMETER_MEMORY_TYPE + "}" + "/_delete_by_query";
     public static final String SEMANTIC_SEARCH_MEMORIES_PATH = MEMORIES_PATH + "/long-term/_semantic_search";
     public static final String HYBRID_SEARCH_MEMORIES_PATH = MEMORIES_PATH + "/long-term/_hybrid_search";
+    // Retention dry-run: preview what the scheduled retention job would delete, without deleting.
+    public static final String RETENTION_DRY_RUN_PATH = BASE_MEMORY_CONTAINERS_PATH
+        + "/{"
+        + PARAMETER_MEMORY_CONTAINER_ID
+        + "}/_retention/_dry_run";
+    public static final String RETENTION_DRY_RUN_ALL_PATH = "/_plugins/_ml/memory/_retention/_dry_run";
```
nathaliellenaa: Should we use the `BASE_MEMORY_CONTAINERS_PATH` since the rest of the features lives inside the memory_containers?

Suggested fix:
```
public static final String RETENTION_DRY_RUN_ALL_PATH = BASE_MEMORY_CONTAINERS_PATH + "/_retention/_dry_run"
```

plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
```diff
@@ -60,6 +61,44 @@ public MLCommonsClusterEventListener(
...
+            && !MLCommonsSettings.ML_COMMONS_MULTI_TENANCY_ENABLED.get(clusterService.getSettings());
+    }
+
+    /**
+     * Whether the cluster contains at least one data node running 3.1 or later. The new {@code .plugins-ml-jobs} index
+     * and its jobs must not be written until this holds, to avoid issues during blue/green and rolling upgrades.
+     */
+    private boolean hasDataNodeOnOrAfterV31() {
```
nathaliellenaa: This returns true if any data node is ≥ 3.1.0, so it gives no rolling-upgrade protection. The startup gate it claims to mirror requires `getMinNodeVersion().onOrAfter(Version.CURRENT)` — i.e. no node older than current. So during a rolling upgrade, an interval change fires the consumer → `upsertMemoryRetentionJob` creates `.plugins-ml-jobs` while older nodes are still present, stranding replicas (cluster yellow) — exactly what #4937 fixed. 

Suggested fix: extract one shared helper (`jobsIndexExists || minNodeVersion.onOrAfter(Version.CURRENT)`) used by both the startup gate and `shouldManageMemoryRetentionJob()`, and add a mixed-version consumer test asserting no upsert.

nathaliellenaa (commented):
Thanks @AirFunnyB , added some comments

plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
AirFunnyB: Fixed

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
AirFunnyB: Yes, this is intentional. The raw Client mirrors how the retention job itself reads the container registry, so the dry-run stays byte-for-byte consistent with what the real run would evaluate — that consistency is the point, and using `SdkClient` only here would risk the preview and the job diverging.

To be explicit about scope: retention is a self-hosted (local-metadata) feature; remote-metadata / managed deployments (AOSS, DynamoDB) are out of scope for it. The follow-up PR adds an `isRemoteMetadataStore` gate so that in remote-metadata mode the job cleanly skips and the dry-run emits an explicit warning rather than silently returning an empty array.

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
AirFunnyB: Fixed

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
AirFunnyB: Fixed

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/TransportMemoryRetentionDryRunAction.java
AirFunnyB: Fixed

plugin/src/main/java/org/opensearch/ml/cluster/MLCommonsClusterEventListener.java
AirFunnyB: Fixed

common/src/main/java/org/opensearch/ml/common/memorycontainer/MemoryContainerConstants.java
AirFunnyB: Fixed

nathaliellenaa (approved):
Thanks for addressing the comments @AirFunnyB , LGTM

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -1885,6 +1908,695 @@ private void deleteEmptyNamespaceWorkingMemory(String containerId, String workin
...
+        ActionListener<MemoryRetentionDryRunResult> listener
+    ) {
+        try {
+            EffectivePolicy effective = resolveEffectivePolicyForDryRun(config);
+            String policySource = effective.source;
+            int ttlDays = clusterService.getClusterSettings().get(ML_COMMONS_MEMORY_WORKING_MEMORY_TTL_DAYS);
+            DryRunContext ctx = new DryRunContext(containerId, policySource, System.currentTimeMillis(), effective.policy, ttlDays);
+
+            if (clusterService.getClusterSettings().get(ML_COMMONS_MULTI_TENANCY_ENABLED)) {
+                ctx.warnings.add("multi-tenancy is enabled; the scheduled retention job does not run, so nothing would be deleted");
```
jiapingzeng, line 2029: Should we also set total would delete to be 0, in addition to this warning? Or just skip the dry run session.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -1885,6 +1908,695 @@ private void deleteEmptyNamespaceWorkingMemory(String containerId, String workin
...
+            }
+            if (!clusterService.getClusterSettings().get(ML_COMMONS_MEMORY_RETENTION_ENABLED)) {
+                ctx.warnings
```
jiapingzeng, line 2036: is just a warning enough here? seems that we will still call `dryRunSessions` on line 2038 in this case even when the feature is disabled.

jiapingzeng:
> plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java: 2590
countMatching() stashes the thread security context (stashContext) before issuing search queries, meaning all dry-run counting searches run with elevated/system-level privileges and bypass document-level security (DLS). This method is reachable by any authenticated user who can trigger a dry-run. Container-level access control is enforced upstream, but counts are returned for documents the calling user may not have DLS access to view directly.

Is this finding valid?

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Done. When multi-tenancy is enabled the dry-run skips all the counting passes, adds the "multi-tenancy is enabled" warning, and returns total_would_delete=0. It no longer falls through and runs the passes.

plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
AirFunnyB: Same treatment for retention_enabled=false: the dry-run now short-circuits, adds the "retention is disabled cluster-wide" warning, and returns 0 without running any passes.

AirFunnyB:
> > plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java: 2590
> > countMatching() stashes the thread security context (stashContext) before issuing search queries, meaning all dry-run counting searches run with elevated/system-level privileges and bypass document-level security (DLS). This method is reachable by any authenticated user who can trigger a dry-run. Container-level access control is enforced upstream, but counts are returned for documents the calling user may not have DLS access to view directly.
> 
> Is this finding valid?

No DLS concern here. The dry-run only ever returns aggregate counts, never document content, and container-level access is enforced upstream before this code runs. In the short-circuited modes it returns 0 without issuing any search at all, so there's nothing for DLS to leak.

AirFunnyB:
@jiapingzeng do you think post my changes this LGTM

jiapingzeng (approved):
LGTM!

PR #4978, opened 2026-08-11
fix memory NPE on security-disabled cluster

jiapingzeng:
### Description
Adding a conversational memory with `infer: true` and no `session_id` fails with a 500 on a security-disabled cluster. The Add Memory API auto-creates a session, but builds the session document with `Map.of(...)`, which throws `NullPointerException` when `owner_id` is null (no authenticated user when security is disabled).

Fix: Replace `Map.of(...)` with a mutable `HashMap` that omits `owner_id` when null — mirroring the null-safe pattern already used in `upsertSessionThenProcess`. Also hardens the path against any other nullable field.

### Steps to reproduce
 
Add a conversational memory **without** a `session_id` in the namespace (so a session is
auto-created), with `infer: true`:
 
### Request
 
```json
POST /_plugins/_ml/memory_containers/<memory_container_id>/memories
{
  "messages": [
    { "role": "user", "content": [ { "text": "I prefer Rust for CLI tools.", "type": "text" } ] }
  ],
  "namespace": { "user_id": "alice" },
  "infer": true,
  "payload_type": "conversational"
}
```
 
### Response
 
```json
{
  "error": {
    "root_cause": [
      { "type": "status_exception", "reason": "Internal server error" }
    ],
    "type": "status_exception",
    "reason": "Internal server error"
  },
  "status": 500
}
```
 
### Server log / stack trace
 
```
[ERROR][o.o.m.a.m.m.TransportAddMemoriesAction] Failed to summarize messages for session creation
java.lang.NullPointerException
    at java.base/java.util.Objects.requireNonNull(Objects.java:220)
    at java.base/java.util.ImmutableCollections$MapN.<init>(ImmutableCollections.java:1437)
    at java.base/java.util.Map.of(Map.java:1485)
    at org.opensearch.ml.action.memorycontainer.memory.TransportAddMemoriesAction.lambda$createNewSessionIfAbsent$0(TransportAddMemoriesAction.java:170)
    at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82)
    at org.opensearch.ml.action.memorycontainer.memory.MemoryProcessingService.lambda$summarizeMessages$0(MemoryProcessingService.java:532)
```

### Related Issues

### Check List

### DIFF
--- a/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
+++ b/plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
@@ -182,24 +182,18 @@ public class TransportAddMemoriesAction extends HandledTransportAction<MLAddMemo
                 // TODO: use LLM to summarize first user message
                 ActionListener<String> summaryListener = ActionListener.wrap(summary -> {
                     Instant now = Instant.now();
-                    indexRequest
-                        .source(
-                            Map
-                                .of(
-                                    OWNER_ID_FIELD,
-                                    input.getOwnerId(),
-                                    MEMORY_CONTAINER_ID_FIELD,
-                                    input.getMemoryContainerId(),
-                                    SUMMARY_FIELD,
-                                    summary,
-                                    NAMESPACE_FIELD,
-                                    input.getNamespace(),
-                                    CREATED_TIME_FIELD,
-                                    now.toEpochMilli(),
-                                    LAST_UPDATED_TIME_FIELD,
-                                    now.toEpochMilli()
-                                )
-                        );
+                    // Build the source with a mutable map: owner_id may be null when security is disabled, and Map.of
+                    // rejects null values (which previously NPE'd this auto-create session path).
+                    Map<String, Object> source = new HashMap<>();
+                    if (input.getOwnerId() != null) {
+                        source.put(OWNER_ID_FIELD, input.getOwnerId());
+                    }
+                    source.put(MEMORY_CONTAINER_ID_FIELD, input.getMemoryContainerId());
+                    source.put(SUMMARY_FIELD, summary);
+                    source.put(NAMESPACE_FIELD, input.getNamespace());
+                    source.put(CREATED_TIME_FIELD, now.toEpochMilli());
+                    source.put(LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+                    indexRequest.source(source);
                     indexRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
                     ActionListener<IndexResponse> responseActionListener = ActionListener.<IndexResponse>wrap(r -> {
                         input.getNamespace().put(SESSION_ID_FIELD, r.getId());
--- a/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java
+++ b/plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java
@@ -699,6 +699,88 @@ public class TransportAddMemoriesActionTests {
     }
 
     @Test
+    public void testDoExecute_SessionCreation_NullOwnerId_SecurityDisabled() {
+        // Regression: on a security-disabled cluster owner_id is null; the auto-create session-source builder
+        // used Map.of(...) which throws NPE on null values. See paste 1786408766.
+        when(mlFeatureEnabledSetting.isAgenticMemoryEnabled()).thenReturn(true);
+
+        MessageInput message = MessageInput.builder().content(createTestContent("Hello")).role("user").build();
+        List<MessageInput> messages = Arrays.asList(message);
+
+        Map<String, String> namespace = new HashMap<>();
+        namespace.put("user_id", "alice");
+
+        MLAddMemoriesInput input = mock(MLAddMemoriesInput.class);
+        when(input.getPinned()).thenReturn(null);
+        when(input.getMemoryContainerId()).thenReturn("container-123");
+        when(input.getMessages()).thenReturn(messages);
+        when(input.isInfer()).thenReturn(true);
+        when(input.getNamespace()).thenReturn(namespace);
+        when(input.getOwnerId()).thenReturn(null); // security disabled -> no authenticated user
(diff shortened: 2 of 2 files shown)


plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
```diff
@@ -182,24 +182,18 @@ private void createNewSessionIfAbsent(
...
+                    }
+                    source.put(MEMORY_CONTAINER_ID_FIELD, input.getMemoryContainerId());
+                    source.put(SUMMARY_FIELD, summary);
+                    source.put(NAMESPACE_FIELD, input.getNamespace());
+                    source.put(CREATED_TIME_FIELD, now.toEpochMilli());
+                    source.put(LAST_UPDATED_TIME_FIELD, now.toEpochMilli());
+                    indexRequest.source(source);
                     indexRequest.setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
                     ActionListener<IndexResponse> responseActionListener = ActionListener.<IndexResponse>wrap(r -> {
                         input.getNamespace().put(SESSION_ID_FIELD, r.getId());
```
ylwu-amzn, line 199: **High-priority — null `namespace` still 500s, now after a write.**

`namespace` is optional (`validate()` doesn't require it, `parse()` leaves it null), so `{"messages":[...], "infer": true}` reaches this branch with `namespace == null`. `Map.of` used to throw before any write; now line 193 accepts the null, the session doc gets indexed, and line 199 NPEs → still a 500, plus an orphaned session doc and a misleading "Failed to index session data" log. Verified locally.

```java
if (input.getNamespace() == null) {
    input.setNamespace(new HashMap<>());
}
```

plugin/src/main/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesAction.java
```diff
@@ -182,24 +182,18 @@ private void createNewSessionIfAbsent(
...
-                                )
-                        );
+                    // Build the source with a mutable map: owner_id may be null when security is disabled, and Map.of
+                    // rejects null values (which previously NPE'd this auto-create session path).
+                    Map<String, Object> source = new HashMap<>();
+                    if (input.getOwnerId() != null) {
+                        source.put(OWNER_ID_FIELD, input.getOwnerId());
+                    }
+                    source.put(MEMORY_CONTAINER_ID_FIELD, input.getMemoryContainerId());
+                    source.put(SUMMARY_FIELD, summary);
```
ylwu-amzn, line 192: Consider `MLMemorySession.builder()...toXContent()` as in `TransportCreateSessionAction.createNewSession`. `toXContent` already null-guards every field, so no per-field checks are needed, and it carries `tenant_id` — which this map omits even though `input.getTenantId()` is read just above, so auto-created sessions differ from API-created ones in multi-tenant setups. Also drops the third copy of the session-doc schema in this file.

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java
```diff
@@ -698,6 +698,88 @@ public void testDoExecute_SuccessfulSessionCreation() {
         verify(actionListener).onResponse(any(MLAddMemoriesResponse.class));
     }
 
+    @Test
+    public void testDoExecute_SessionCreation_NullOwnerId_SecurityDisabled() {
+        // Regression: on a security-disabled cluster owner_id is null; the auto-create session-source builder
+        // used Map.of(...) which throws NPE on null values. See paste 1786408766.
```
ylwu-amzn, line 704: `See paste 1786408766.` is an internal-only link — please use the GitHub issue number instead (the PR description's `Resolves #...` is also unfilled).

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java
```diff
@@ -698,6 +698,88 @@ public void testDoExecute_SuccessfulSessionCreation() {
...
+        verify(actionListener).onResponse(any(MLAddMemoriesResponse.class));
+
+        // The session document source must omit owner_id (rather than storing null).
+        IndexRequest sessionIndexRequest = indexRequestCaptor.getAllValues()
+            .stream()
+            .filter(r -> "session-index".equals(r.index()))
+            .findFirst()
+            .orElseThrow(() -> new AssertionError("No session-index IndexRequest captured"));
+        Map<String, Object> sessionSource = sessionIndexRequest.sourceAsMap();
+        assertFalse("owner_id should be omitted when null", sessionSource.containsKey("owner_id"));
```
ylwu-amzn, line 778: Nothing asserts `owner_id` *is* written when non-null, so an inverted check would pass the whole suite while hiding sessions from their owners (`addOwnerIdFilter`) — the same inversion that exists today in `MemoryOperationsService.createErrorMemoryHistory`. One line in `testDoExecute_SuccessfulSessionCreation` covers it: `assertEquals("user-123", sessionSource.get("owner_id"));`

plugin/src/test/java/org/opensearch/ml/action/memorycontainer/memory/TransportAddMemoriesActionTests.java
```diff
@@ -698,6 +698,88 @@ public void testDoExecute_SuccessfulSessionCreation() {
...
+        List<MessageInput> messages = Arrays.asList(message);
+
+        Map<String, String> namespace = new HashMap<>();
+        namespace.put("user_id", "alice");
+
+        MLAddMemoriesInput input = mock(MLAddMemoriesInput.class);
+        when(input.getPinned()).thenReturn(null);
+        when(input.getMemoryContainerId()).thenReturn("container-123");
+        when(input.getMessages()).thenReturn(messages);
+        when(input.isInfer()).thenReturn(true);
```
ylwu-amzn, line 717: Nit: `false` is enough to reach session creation (see `testDoExecute_SuccessfulSessionCreation`). With `true`, the inline executor also runs `extractLongTermMemory`, where `getStrategies()` is null on the mock and the NPE gets swallowed.

jiapingzeng:
no longer needed as https://github.com/opensearch-project/ml-commons/pull/4973 will fix this bug

PR #4982, opened 2026-08-18
Agentic template schema enrichment

aditya0shah:
### Description

 
  When a search template is registered without a caller-supplied param-schema,
  ml-commons derives one from the Mustache body. That derivation used to produce only
  a name, type, and `required` flag per parameter. This PR adds a register-time pass
  that, **with no LLM**, fills in:

  - a per-parameter **description** (what it does, and its default),
  - a per-parameter **enum** for slots with a fixed set of values,
  - a template-level **description** used for multi-template selection.

  Everything is recovered from the template body. The pass is best-effort: on any
  failure the base derivation stands and registration still succeeds.

  ### Core idea

  Reading the template only tells us a blank named `q` exists, not what it's for. So
  we **fill every blank with a unique tracer value, render the body into real query
  JSON, and see where each tracer lands.** Its position tells us the parameter's role.

  For `{ "query": { "match": { "title": "{{q}}" } }, "size": {{n}} }`, rendering with
  tracers gives `match.title = "<tracer>"` and `size = <tracer>` → `q` is a full-text
  query on `title`; `n` is the result count.

  ### How it works

  `enrichStructurally` runs four steps (`TemplateStructureAnalyzer`):

  1. **Mark** – give each parameter a unique, type-correct tracer so the render stays
     legal JSON: string → `__mk_sentinel_<i>__`, number → a large distinct int, bool →
     `true` (opens its section), array → `[]` (not tracked; a raw-JSON slot has no
     single field). Render through the cluster's Mustache engine (same one
     `_render/template` uses).
  2. **Locate** – walk the rendered JSON and record each tracer's path. Found as a
     value → the slot that parameter fills; found as an object key → a field-selector
     (e.g. the sort key). First occurrence wins.
  3. **Classify** – match the tail of the path against a fixed OpenSearch clause table
     to get the role, target field, and (for ranges) the bound:

     | Where the tracer landed | Role | Description | Enum |
     |---|---|---|---|
     | `multi_match.query` (+ sibling `fields`) | full-text | `Full-text query matched against the <fields> field(s).` | — |
     | `match` / `match_bool_prefix` `.<f>` | full-text | same | — |
     | `match_phrase` / `match_phrase_prefix` `.<f>` | phrase | `Exact phrase to match in the <f> field.` | — |
     | `term.<f>` | filter | `Filter by exact <f> value.` | — |
     | `terms.<f>` | filter | `Filter by one or more <f> values.` | — |
     | `prefix`/`wildcard`/`regexp` `.<f>` | pattern | `Pattern to match against the <f> field.` | — |
     | `fuzzy.<f>` | fuzzy | `Fuzzy match term for the <f> field.` | — |
     | `range.<f>.gte` / `.gt` | range bound | `Lower bound (>= / >) on <f>.` | — |
     | `range.<f>.lte` / `.lt` | range bound | `Upper bound (<= / <) on <f>.` | — |
     | object key under `sort` | sort field | `Field to sort results by.` | index fields |
     | object key elsewhere | field selector | `Field name to target.` | index fields |
     | `sort[].<f>.order` | sort order | `Sort direction[ for the <f> sort].` | `asc, desc` |
     | top-level `size` | result count | `Number of results to return.` | — |
     | top-level `from` | result offset | `Offset of the first result to return.` | — |
     | `…operator` | match operator | `Whether all query terms must match (and) or any may match (or).` | `and, or` |
     | `…zero_terms_query` | zero terms | `Behavior when the analyzed query has no terms.` | `none, all` |
     | `…score_mode` | score mode | `How to combine the scores of matching clauses.` | `multiply, sum, avg, first, max, min` |
     | `…boost_mode` | boost mode | `How to combine the query score with the function score.` | `multiply, replace, sum, avg, max, min` |
     | `…boost` | boost | `Relevance boost weight for this clause.` | — |

     A tracer that can't be located, or a clause not in the table, gets no role and
     keeps its base spec. Bool/array params carry no clause and are described from type
     alone (`Set to true to enable the optional <param> clause.` /
     `A JSON array or object passed as a raw JSON string.`).

  4. **Emit** – build each field from the recovered facts:
     - **description** from the skeleton above (a `title^2` boost suffix is stripped to
       `title`).
     - **default** recovered by rendering a *second* time with only required params;
       an optional slot then shows the body's own fallback (e.g. an inverted section),
       appended as `Defaults to <value> if unset.`
     - **enum** from two sources: field-selectors (named `field`, `sort_by`, or ending
       `_field`) are scoped to the index mapping's leaf field names
       (`source: "mapping"`) so the model can only pick a real field; closed slots
       (sort order, operator, `zero_terms_query`, `score_mode`, `boost_mode`) get their
       fixed OpenSearch value set.
     - **template description** = the recovered roles joined into one line:
       `full-text search over <f>; filters by <f>; range filters on <f>; sortable; …`.

  ### Merge rules

  Enrichment never overrides what's already there:

  - writes a description only when it's empty (caller's is kept);
  - adds an enum only when the param has none (a mapping enum is kept);
  - a caller-supplied template description still wins;
  - never adds or renames params — it only mutates specs the body already declares.

  ### Worked example

  Stored template `_scripts/products_search`:

  ```json
  {
    "query": {
      "bool": {
        "must": [ { "multi_match": { "query": "{{search_text}}", "fields": ["title", "description"] } } ],
        "filter": [
          { "term": { "status": "{{status}}" } },
          { "range": { "price": { "lte": {{max_price}} } } }
        ]
      }
    },
    "sort": [ { "{{sort_by}}": { "order": "{{order}}" } } ],
    "size": {{size}}{{^size}}10{{/size}}
  }
```
  Derived + enriched param-schema:
```json
  {
    "search_text": { "type": "string", "required": true,
      "description": "Full-text query matched against the title and description fields." },
    "status":      { "type": "string", "required": true,
      "description": "Filter by exact status value." },
    "max_price":   { "type": "number", "required": true,
      "description": "Upper bound (<=) on price." },
    "sort_by":     { "type": "string", "required": true, "source": "mapping",
      "enum": ["title", "price", "status", "created_at"],
      "description": "Field to sort results by." },
    "order":       { "type": "string", "required": true, "enum": ["asc", "desc"],
      "description": "Sort direction." },
    "size":        { "type": "number", "required": false,
      "description": "Number of results to return. Defaults to 10 if unset." }
  }
```
  Template-level description:

  `Full-text search over title and description; filters by status; range filters on price; sortable.`

  (sort_by's enum is the index's leaf field names; order gets asc/desc; size
  picks up its 10 default from the required-only render.)

### Check List


plugin/src/main/java/org/opensearch/ml/action/agenticsearch/TemplateStructureAnalyzer.java
```diff
@@ -0,0 +1,637 @@
...
+        if (isRangeBound(last) && "range".equals(prev2)) {
+            return new Facts(ROLE_RANGE_BOUND, fieldList(prev), last);
+        }
+
+        // sort[].<field>.order (field may be a literal name or a field-selector marker)
+        if ("order".equals(last) && keys.contains("sort")) {
+            return new Facts(ROLE_SORT_ORDER, fieldList(prev), null);
+        }
+
+        // Closed-vocabulary option slots.
```
jiapingzeng: Can we also gate these conditions on their parent clause, similar to the range above?

plugin/src/main/java/org/opensearch/ml/action/agenticsearch/TemplateStructureAnalyzer.java
```diff
@@ -0,0 +1,637 @@
...
+            // leak into the description as a field name.
+            if (f instanceof String && !((String) f).startsWith(MARKER_PREFIX)) {
+                out.add(stripBoost((String) f));
+            }
+        }
+        return out;
+    }
+
+    /** Navigate to a (possibly container) node at a key-path, for reading a clause's siblings. */
+    private static Object navigate(Object root, List<Object> path) {
```
jiapingzeng, line 539: this seems almost identical to `valueAt()`, are we able to combine them?

jiapingzeng (approved):
LGTM!

Issue #4990, opened 2026-08-26
[BUG] Model registration fails under security with "Direct self-reference leading to cycle (User[principal])" after Jackson 3.x migration


bzhangam:
**What is the bug?**
`RestActionUtils.isSuperAdminUser(...)` serializes the Security plugin's `org.opensearch.security.user.User` object with a Jackson `ObjectMapper`. That `User` implements `java.security.Principal` and its `getPrincipal()` returns `this` (a direct self-reference) with no `@JsonIgnore`. After ml-commons migrated to Jackson 3.x (`tools.jackson`) in #4795, Jackson 3 throws on this self-reference by default, so the call fails with HTTP 500. Since `isSuperAdminUser` runs on every model registration (via `TransportRegisterModelAction.isSuperAdminUserWrapper`), any model register/deploy on a security-enabled cluster fails whenever the authenticated `User` is present in the thread context.

```
invalid_definition_exception: Direct self-reference leading to cycle
  at [No location information] (through reference chain: org.opensearch.security.user.User["principal"])
status: 500
```

**How can one reproduce the bug?**
1. Start a security-enabled cluster with the ml-commons plugin (built against Jackson 3.x, e.g. `opensearch-ml-plugin:3.9.0.0-SNAPSHOT`).
2. As an authenticated user, call `POST /_plugins/_ml/models/_register` (or register + deploy a model).
3. `isSuperAdminUser` serializes the transient `OPENDISTRO_SECURITY_USER` and hits the self-reference.
4. See the 500 error `Direct self-reference leading to cycle ... org.opensearch.security.user.User["principal"]`.

Reproduced consistently in neural-search security-enabled integration tests (any test that registers a model in setup, e.g. `MMRNeuralQueryTransformerIT`): https://github.com/opensearch-project/neural-search/actions/runs/32916260192/job/98020534485

**What is the expected behavior?**
Model registration succeeds on a security-enabled cluster. `isSuperAdminUser` should determine the user name without failing on the self-referential `User` bean.

**What is your host/environment?**
 - OS: Linux
 - Version: OpenSearch 3.9.0 line; `opensearch-ml-plugin:3.9.0.0-SNAPSHOT` (verified to bundle `jackson-databind-3.2.2.jar`, i.e. Jackson 3.x)
 - Plugins: opensearch-ml-plugin, opensearch-security (security enabled)

**Do you have any screenshots?**
Server-side stack trace:
```
[...]

PR #4994, opened 2026-08-27
fix: Avoid Jackson 3 self-reference cycle in isSuperAdminUser

bzhangam:
### Description
RestActionUtils.isSuperAdminUser serialized the security plugin's User object (a Principal whose getPrincipal() returns this) with a Jackson 3 ObjectMapper. Jackson 3 throws "Direct self-reference leading to cycle" by default, so every model register/deploy on a security-enabled cluster failed with HTTP 500.

The method only needs the user name, and in production the object is always a Principal. Read the name directly via Principal.getName() and skip serializing the bean; fall back to the previous JSON path for non-Principal objects.

Add unit tests that put a real self-referential Principal (the shape of org.opensearch.security.user.User) in the OPENDISTRO_SECURITY_USER transient. These reproduce the exact failure before the fix and pass after it, closing the coverage gap left by the prior Map-based tests.

### Related Issues
https://github.com/opensearch-project/ml-commons/issues/4990

### Check List


plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +254,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
             return AccessController.doPrivileged((PrivilegedExceptionAction<Boolean>) () -> {
-                String userContext = objectMapper.writeValueAsString(userObject);
-                final JsonNode node = objectMapper.readTree(userContext);
-                final String userName = node.get("name").asText();
+                // The security plugin's org.opensearch.security.user.User implements Principal and its
+                // getPrincipal() returns `this` (a direct self-reference with no @JsonIgnore). Serializing
+                // that bean throws "Direct self-reference leading to cycle" under Jackson 3. Since we only
+                // need the user name, read it directly from the Principal and avoid serializing the bean.
+                // See ml-commons issue #4990.
+                final String userName;
```
mingshl, line 262: is this the only call site that serializes the security User with Jackson 3?

bzhangam: Yes — verified across the codebase. isSuperAdminUser is the only place that reads the raw security User object and serializes it with Jackson.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +254,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
         try {
             return AccessController.doPrivileged((PrivilegedExceptionAction<Boolean>) () -> {
-                String userContext = objectMapper.writeValueAsString(userObject);
-                final JsonNode node = objectMapper.readTree(userContext);
-                final String userName = node.get("name").asText();
+                // The security plugin's org.opensearch.security.user.User implements Principal and its
+                // getPrincipal() returns `this` (a direct self-reference with no @JsonIgnore). Serializing
+                // that bean throws "Direct self-reference leading to cycle" under Jackson 3. Since we only
+                // need the user name, read it directly from the Principal and avoid serializing the bean.
+                // See ml-commons issue #4990.
```
ylwu-amzn: security#6460 added `@JsonIgnore` to `getPrincipal()` on Sep 2, so this is no longer accurate. Maybe reword as hardening.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +254,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
+                final String userName;
+                if (userObject instanceof Principal) {
+                    userName = ((Principal) userObject).getName();
```
ylwu-amzn: Nit: `instanceof Principal principal` reads better and is used elsewhere in the repo.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
ylwu-amzn: security#6460 added `@JsonIgnore` to `getPrincipal()` on Sep 2, so this is no longer accurate. Maybe reword as hardening.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
ylwu-amzn: Nit: `instanceof Principal principal` would let you drop the cast on the next line.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +254,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
+                } else {
+                    String userContext = objectMapper.writeValueAsString(userObject);
+                    final JsonNode node = objectMapper.readTree(userContext);
+                    userName = node.get("name").asText();
+                }
```
ylwu-amzn: In production this transient is always a `Principal`, so only the two `Map.of` tests reach this fallback. Worth keeping?

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
bzhangam: sure will update

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
bzhangam: yeah we can do that

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
bzhangam: That's just a defensive fall back. Since we know it's always Principal we can remove it.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +251,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
+                // rather than serializing the bean: it is simpler, and it avoids depending on the bean being
+                // safe to serialize (User#getPrincipal() returns `this`, a self-reference Jackson 3 rejects
+                // by default). See ml-commons issue #4990.
+                if (!(userObject instanceof Principal principal)) {
+                    // Anything else in this transient is a contract violation, not a normal input. Fail loud
+                    // rather than silently deciding this is not an admin on a security-relevant path.
+                    throw new IllegalStateException(
+                        "Expected the security user transient to be a Principal but got " + userObject.getClass().getName()
+                    );
+                }
```
ylwu-amzn: `User` only became a `Principal` in security#6419 (3.9). On security 3.8 it's `Serializable, CustomAttributesAware`, on 2.19 `Serializable, Writeable, CustomAttributesAware` — and ml-commons 3.8 already runs Jackson 3, so this looks backportable. If it lands on 3.8/2.19 this throws for all 13 callers of `isSuperAdminUser` (get/delete/update model, deploy, undeploy, agents, `_stats`, `_profile`), none of which catch it → 500.

Falling back to the JSON read instead of throwing keeps it version-safe. Withdrawing my earlier "worth keeping?" — I'd only considered reachability on `main`.

bzhangam: sure. we can add back the defensive fallback.

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
```diff
@@ -253,9 +251,19 @@ public static boolean isSuperAdminUser(ClusterService clusterService, Client cli
...
+                final String userName = principal.getName();
```
ylwu-amzn: Since you're adding the fallback back — one refinement on *where* the `IllegalStateException` goes. Put it on "couldn't read a name", not on "isn't a `Principal`": the non-`Principal` case is precisely the legitimate security < 3.9 case, while a missing `name` genuinely is a contract violation.

I also checked that the 3.8 bean is actually safe to serialize here, i.e. the fallback works rather than just deferring a different failure: security 3.8 `User` exposes only `getName` / `getRoles` / `getSecurityRoles` / `getRequestedTenant` / `getCustomAttributesMap` / `getPluginName` / `getAuthenticatedBy` plus three `is*` accessors — all plain values and immutable collections, no cycle. `getPrincipal()`, the self-reference, only arrives in 3.9 together with `Subject`.

Worth stating in the comment *why* the branch exists, so a later cleanup doesn't delete the backport safety along with what looks like test-only code.

```suggestion
                if (userObject instanceof Principal principal) {
                    return isAdminDN(principal.getName());
                }
                // security < 3.9 (3.8, 2.19): User is not a Principal, so read the name off the serialized
                // bean. Those versions have no User#getPrincipal(), so there is no self-reference to trip on.
                final JsonNode node = objectMapper.readTree(objectMapper.writeValueAsString(userObject));
                final JsonNode nameNode = node.get("name");
                if (nameNode == null) {
                    // Fail loud rather than silently deciding this is not an admin on a security-relevant path.
                    throw new IllegalStateException("Security user transient has no name: " + userObject.getClass().getName());
                }
                final String userName = nameNode.asText();
```

This keeps the existing `return isAdminDN(userName);` below as the fallback's return path, and keeps `objectMapper` / the `tools.jackson` imports — so the PR description shouldn't claim the Jackson dependency goes away, only that the hot path no longer serializes the bean.

plugin/src/test/java/org/opensearch/ml/utils/RestActionUtilsTests.java
```diff
@@ -328,12 +330,54 @@ public void testIsSuperAdminUser_NotAdmin() {
...
+        ThreadContext threadContext = new ThreadContext(Settings.EMPTY);
+
+        when(clusterService.getSettings())
+            .thenReturn(Settings.builder().putList(RestActionUtils.SECURITY_AUTHCZ_ADMIN_DN, "cn=admin").build());
+        when(client.threadPool()).thenReturn(mock(ThreadPool.class));
+        when(client.threadPool().getThreadContext()).thenReturn(threadContext);
+        threadContext.putTransient(RestActionUtils.OPENDISTRO_SECURITY_USER, Map.of("name", "notAPrincipal"));
+
+        Assert.assertThrows(IllegalStateException.class, () -> RestActionUtils.isSuperAdminUser(clusterService, client));
+    }
```
ylwu-amzn: This test has to flip with the change above: `Map.of("name", "notAPrincipal")` becomes the *supported* fallback input, so it will return `false` rather than throw.

Suggest splitting it in two:

- **fallback success** — a non-`Principal` POJO exposing `getName()` (the shape of security 3.8's `User`) resolves admin / non-admin correctly. That's the path the branch exists for, and a bare `Map` doesn't really stand in for the real bean.
- **fail loud** — an object with no `name` at all (e.g. `Map.of("foo", "bar")`) throws `IllegalStateException`.

bzhangam: sure

plugin/src/main/java/org/opensearch/ml/utils/RestActionUtils.java
bzhangam: ok

ylwu-amzn (approved):

mingshl:
there is a dependency issue

```
Execution failed for task ':opensearch-ml-plugin:compileYamlRestTestJava'.
> Could not resolve all files for configuration ':opensearch-ml-plugin:yamlRestTestCompileClasspath'.
   > Could not resolve org.jspecify:jspecify:1.0.0.
     Required by:
         project ':opensearch-ml-plugin' > org.apache.logging.log4j:log4j-core:2.25.5
         project ':opensearch-ml-plugin' > org.apache.logging.log4j:log4j-core:2.25.5 > org.apache.logging.log4j:log4j-api:2.25.5
      > Conflict found for module 'org.jspecify:jspecify': between versions 1.0.1 and 1.0.0
> There is 1 more failure with an identical cause.


```

bzhangam:
> there is a dependency issue
> 
> ```
> Execution failed for task ':opensearch-ml-plugin:compileYamlRestTestJava'.
> > Could not resolve all files for configuration ':opensearch-ml-plugin:yamlRestTestCompileClasspath'.
>    > Could not resolve org.jspecify:jspecify:1.0.0.
>      Required by:
>          project ':opensearch-ml-plugin' > org.apache.logging.log4j:log4j-core:2.25.5
>          project ':opensearch-ml-plugin' > org.apache.logging.log4j:log4j-core:2.25.5 > org.apache.logging.log4j:log4j-api:2.25.5
>       > Conflict found for module 'org.jspecify:jspecify': between versions 1.0.1 and 1.0.0
> > There is 1 more failure with an identical cause.
> ```

@mingshl I think this issue is not related to this PR change.

mingshl:
thanks for fixing the dependency issue https://github.com/opensearch-project/ml-commons/pull/5023 , rebasing the PR

mingshl (approved):

mingshl:
@jiapingzeng would you trigger the CI, i am not able to approve my own triggered CI

PR #4995, opened 2026-08-27
Fix self-reference cycle when serializing security User in isSuperAdminUser

### LINKED ISSUE #4990: [BUG] Model registration fails under security with "Direct self-reference leading to cycle (User[principal])" after Jackson 3.x migration

adityaanikam:
### Description
RestActionUtils.isSuperAdminUser serializes the Security plugin's User object with Jackson to extract the user name. User implements Principal and its getPrincipal() bean property returns itself with no @JsonIgnore, a direct self-reference that Jackson 2 tolerated but Jackson 3 (adopted in #4795) rejects with InvalidDefinitionException, failing every model registration on a security-enabled cluster.

Reads the name directly via Principal.getName() when the transient user object is a Principal, instead of round-tripping it through the object mapper. Falls back to the existing JSON path for any other object shape.

### Related Issues
Resolves #4990

### Check List
- [ ] API changes companion pull request created.
- [ ] Public documentation issue/PR created.


ylwu-amzn:
Thanks for the fix @adityaanikam, and sorry for the duplicated effort here.

Closing this as a duplicate of #4994, which was opened about two hours earlier the
same day and resolves the same issue (#4990) with the same approach

PR #4996, opened 2026-08-28
Add server-side cross-request batch inference queue for the online predict path

RexyZhang2001:
### Description

Implements Phase 2 of server-side batch inference (#4889): a cross-request batch queue for the online predict path.

When a remote model is configured with `batch_inference_config.queue.enabled`, ML Commons holds incoming single-request predicts from independent concurrent callers to the same model for a short window, coalesces them into shared model calls, and then routes each result back to the caller that submitted it. A flush fires on whichever comes first — the accumulated items reaching the model's size limits, or `flush_timeout_ms` elapsing since the first queued request. Every dispatched call still goes through the Phase 1 size-based splitter, so it respects `max_items_per_request` / `max_bytes_per_request`. This lets a search path that issues one text per request amortize many callers into far fewer model calls, without any client-side batching or shared client state.

Only requests that are safe to merge share a call — same input type and same non-payload parameters (algorithm parameters, result filter). The queue is opt-in per model and off by default; a model without the queue enabled bypasses it entirely and behaves exactly as Phase 1.

**File organization**

```
+------------------------------------------------------------------------+
| Config layer -- whether a model coalesces, and how long it waits       |
|                                                                        |
|BatchQueueConfig          queue settings: enabled + flush_timeout_ms    |
|BatchInferenceConfig      adds an optional nested `queue` block         |
+------------------------------------------------------------------------+
                          | read by
                          v
+------------------------------------------------------------------------+
| Execution layer -- coalesce, flush, dispatch, route back               |
| (org.opensearch.ml.batch)                                              |
|                                                                        |
|QueueEntry                one waiting request: input + listener, plus   |
|                          its items and group key derived once up front |
|ModelBatchQueue           per-model queue: coalesce -> flush on         |
|                          threshold or timer -> split -> dispatch ->    |
|                          route each result back to its caller          |
|ModelBatchQueueManager    owns the per-model queues; creates one lazily |
|                          and replaces it when the model's config changes|
|BatchableInput            adds groupKey (which requests may coalesce)    |
|                          and distribute (inverse of combine)           |
|TextDocsBatchableInput    text-docs groupKey / distribute               |
|BatchItem                 adds source back-pointers for result routing  |
|BatchInferenceRouter      single entry point: queue when enabled,       |
|                          otherwise the Phase 1 size-based splitter     |
+------------------------------------------------------------------------+
                          | wired into
                          v
+------------------------------------------------------------------------+
| Integration -- the real predict path                                   |
|                                                                        |
|MLPredictTaskRunner       online predict routes through the router;     |
|                          offline batch prediction calls the model      |
|                          directly, unchanged                           |
+------------------------------------------------------------------------+
```

**Request flow**

```
online predict request arrives
   -> MLPredictTaskRunner reads the model's batch_inference_config from cache
   -> queue not enabled  ->  Phase 1 size-based splitter, exactly as before
   -> queue enabled:
        ModelBatchQueueManager finds/creates this model's queue
        QueueEntry             decompose input into items + compute group key (once)
        ModelBatchQueue        add to the queue; start the flush_timeout_ms timer
                               if this is the first entry
        flush (threshold hit OR timer fires):
          group the drained entries by (input type + group key)
          for each group: split into size-legal sub-batches (Phase 1 splitter)
                          invoke the model per sub-batch
                          distribute + combine -> each caller's own result, in order
   -> every caller's listener notified once with exactly its own response
```

**BWC**
The new `queue` block is an optional field inside `batch_inference_config` and is wire-gated on the 3.9 transport version, so a mixed-version cluster stays compatible. `queue.enabled` defaults to `false` and `flush_timeout_ms` to `50`. A model without the queue enabled never enters the queue path and behaves exactly as Phase 1.

### Related Issues

Relates to issue #4889 
Builds on PR #4898

### Benchmark Result
https://github.com/opensearch-project/ml-commons/issues/4889#issuecomment-5398072424
Validated on a live 3-node cluster (`./gradlew run -PnumNodes=3`, `3.9.0-SNAPSHOT`) against a real remote model, Amazon Bedrock `cohere.embed-english-v3`. 

- **Coalescing** — 300 concurrent single-text queries made 23 model calls (13.0 texts/call, ~13x fewer) with 0 errors; 30 concurrent coalesced to exactly 10 per node, confirming one queue per node.
- **Flush paths** — a full batch flushes early at the size limit; the remainder flushes on the timer.
- **Result routing** — 15 mutually-unrelated texts replayed concurrently: each response's nearest reference vector was its own text, 15/15, 0 mis-routed.
- **Latency** — worst case (queries sent one at a time, no siblings to batch): p50 222 -> 224 ms, within run-to-run noise.

### Testing

**Unit tests** (all passing): `ModelBatchQueueTests`, `ModelBatchQueueManagerTests`, `BatchInferenceRouterTests`, `BatchQueueConfigTest`, plus `BatchInferenceConfigTest` and `TextDocsBatchableInputTests` extended for the queue config, group key, and result distribution.

### Check List


plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueueManager.java
```diff
@@ -0,0 +1,87 @@
...
+        String groupKey;
+        try {
+            groupKey = handler.groupKey(input);
+        } catch (Exception e) {
+            groupKey = null;
+        }
+        return new QueueEntry(input, listener, predictor, channel, items, groupKey);
+    }
+
+    private boolean sameQueueConfig(BatchInferenceConfig existing, BatchInferenceConfig incoming) {
```
bzhangam: Why not override the equal function of the BatchInferenceConfig?

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+            }, error -> {
+                try {
+                    markFailed(subBatch, entryFailures, error);
+                } finally {
+                    settle();
+                }
+            });
+
+            try {
+                QueueEntry template = group.get(subBatch.get(0).getSourceIndex());
```
bzhangam: Why we call the QueueEntry as template?

plugin/src/main/java/org/opensearch/ml/batch/QueueEntry.java
```diff
@@ -0,0 +1,64 @@
...
+        String groupKey
+    ) {
+        this.input = input;
+        this.listener = listener;
+        this.predictor = predictor;
+        this.channel = channel;
+        this.items = items;
+        this.groupKey = groupKey;
+        if (items == null) {
+            this.itemCount = 1;
```
bzhangam, line 63: Why itemCount 1?

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+        this.registry = registry;
+        this.splitter = splitter;
+        this.threadPool = threadPool;
+    }
+
+    BatchInferenceConfig getConfig() {
+        return config;
+    }
+
+    public void enqueue(QueueEntry entry) {
```
bzhangam, line 89: ConcurrentHashMap<String, ModelBatchQueue> has no removal path, so every model that ever received a queue-enabled predict leaks for the JVM's life (slow leak with transient model IDs). Rather than hooking undeploy/delete/config-change events (easy to miss a path), I'd suggest idle/last-use TTL eviction: stamp
  lastUsedNanos on each enqueue, and a periodic sweep removes queues that are idle beyond a TTL.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+        queue.add(entry);
```
bzhangam: Separately from the idle leak, a hot model has no bound on in-flight queued work. Both overCount/overBytes limits can be disabled, and even when enabled they bound batch size per dispatched call, not outstanding queued work — so if the remote model is slow (throttling/cold start) and arrival rate > drain rate, queue grows until OOM. Suggest a per-node cap enforced in enqueue, rejecting with OpenSearchRejectedExecutionException (429-style backpressure) rather than dropping:

Bytes is the real OOM guard, sized relative to heap and clamped to a range so it self-scales without going nonsensical: effective = clamp(heapMax × fraction, floorBytes, ceilingBytes) — e.g. clamp(heap × 1%, 64MB, 512MB). 4GB heap → floored to 64MB; 32GB → 320MB; 128GB → capped at 512MB. I'd make this a node-wide shared budget (one reserve/release counter in the manager) since heap is a node resource, and make fraction/floor/ceiling dynamic settings so it's tunable from load tests without a restart.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+        pendingEntries.incrementAndGet();
```
bzhangam: Why not make pendingEntries, pendingItems and pendingBytes one object so there calculation is atomic? In current implementation there can be case:
Thread 1: pendingEntries.incrementAndGet();
Thread 2: fire flush;
Thread 2: pendingItems.addAndGet(-drainedItems); // can be transient negative number which is not accurate

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+            dispatch(batch);
+        }
+        if (pendingEntries.get() > 0) {
+            scheduleTimer();
+        }
+    }
+
+    // Returns null if another flush is already draining.
+    private List<QueueEntry> drain() {
+        if (!flushing.compareAndSet(false, true)) {
```
bzhangam: should we say draining rather than flushing?

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -0,0 +1,366 @@
...
+                }, TimeValue.timeValueMillis(flushTimeoutMs), REMOTE_PREDICT_THREAD_POOL);
+            } catch (Exception e) {
+                timerScheduled.set(false);
+                log.warn("Failed to schedule batch flush timer for model {}; will retry on next enqueue", modelId, e);
+            }
+        }
+    }
+
+    void flush() {
+        List<QueueEntry> batch = drain();
```
bzhangam, line 162: After drain if there is timer should we reset it so we don't have the case:
timer scheduled [E1]
over flush [E1, E2, E3]
new entry [E4] 
timer flush [E4]

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueueManager.java
RexyZhang2001: Added `@EqualsAndHashCode` to `BatchQueueConfig` and dropped `sameQueueConfig`.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Renamed to firstSourceEntry meaning the source request of the sub-batch's first item

plugin/src/main/java/org/opensearch/ml/batch/QueueEntry.java
RexyZhang2001: It's an entry whose input type has no batch handler; it carries no items but still counts as one pending request so it's drained and rejected promptly. Added a comment in the code for explanation

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Added idle-TTL eviction

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Added a node-wide shared budget

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Replaced the three atomics with a single `AtomicReference<Totals>` record updated atomically, so no transient-negative reads.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Renamed to draining

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Fixed in new commit

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -44,41 +46,61 @@ public class ModelBatchQueue {
...
+                        "Batch inference queue memory budget is exhausted for model " + modelId + "; retry after backoff"
+                    )
+                );
+            return;
+        }
         queue.add(entry);
-        pendingEntries.incrementAndGet();
-        long items = pendingItems.addAndGet(entry.getItemCount());
-        long bytes = pendingBytes.addAndGet(entry.getByteSize());
+        Totals current = totals.updateAndGet(t -> t.plus(entry));
```
bzhangam: `enqueue` → `totals` update is not atomic with the queue add, which can strand entries and skew the budget.

In `enqueue()`:
```java
queue.add(entry);                                          // (1)
Totals current = totals.updateAndGet(t -> t.plus(entry));  // (2)
```
`drain()` snapshots `totals.getAndSet(ZERO)` and then polls exactly `snapshot.entries()` times. If a concurrent flush runs between (1) and (2), the entry is already in `queue` but not yet in `totals`, so:
- it may be polled and dispatched even though its bytes/count were never accounted, and `budget.release(drainedBytes)` then releases bytes that were reserved but never counted → budget drift; or
- it goes unpolled, and after `getAndSet(ZERO)` nothing re-arms the timer for it unless another enqueue happens to arrive.

Could we make the queue add and the totals update effectively atomic (single source of truth), or change `drain()` to poll until empty and derive the drained totals from what was actually polled rather than trusting the snapshot count?

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueueManager.java
```diff
@@ -78,10 +110,16 @@ private QueueEntry toEntry(MLInput input, ActionListener<MLTaskResponse> listene
         return new QueueEntry(input, listener, predictor, channel, items, groupKey);
     }
 
-    private boolean sameQueueConfig(BatchInferenceConfig existing, BatchInferenceConfig incoming) {
-        return existing.getMaxItemsPerRequest() == incoming.getMaxItemsPerRequest()
-            && existing.getMaxBytesPerRequest() == incoming.getMaxBytesPerRequest()
-            && existing.getQueue().getFlushTimeoutMs() == incoming.getQueue().getFlushTimeoutMs()
-            && existing.getQueue().isEnabled() == incoming.getQueue().isEnabled();
+    void evictIdleQueues() {
```
bzhangam, line 141: **Eviction vs. in-flight enqueue has the same window.**

`evictIdleQueues()` removes a queue when `isIdle()` (`entries()==0 && !draining`). A caller that already resolved the `ModelBatchQueue` via `queueFor()` can `enqueue()` onto it while the sweep removes it from the map, leaving the entry on an orphaned queue whose only flush path is its own timer. Same root cause as above issue — tightening the enqueue/totals ordering should also close this.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
```diff
@@ -44,41 +46,61 @@ public class ModelBatchQueue {
...
+        return lastUsedNanos;
+    }
+
+    boolean isIdle() {
+        return totals.get().entries() == 0 && !draining.get();
+    }
+
     public void enqueue(QueueEntry entry) {
+        lastUsedNanos = System.nanoTime();
+        if (!budget.tryReserve(entry.getByteSize())) {
```
bzhangam: **Byte accounting under-counts, which weakens the OOM guard.**
- `QueueEntry.byteSize` for `items == null` entries is `0`, so a flood of unsupported-input-type requests reserves nothing while still retaining `MLInput`/channel/listener until flush rejects them. Suggest charging a non-zero nominal size.
- For supported entries, `byteSize` is payload-only; it ignores the retained `MLInput` graph, `TransportChannel`, the copied `List<BatchItem>`, and per-entry overhead. Since this is an OOM guard, under-counting is the dangerous direction — a small fixed per-entry overhead added to `byteSize` would make the "% of heap" budget reflect real retained heap.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -156,6 +157,40 @@ private MLCommonsSettings() {}
             Setting.Property.Dynamic
         );
 
+    public static final Setting<Double> ML_COMMONS_BATCH_QUEUE_MEMORY_FRACTION = Setting
+        .doubleSetting(
+            ML_PLUGIN_SETTING_PREFIX + "batch_queue.memory_fraction",
+            0.01,
+            0.0,
+            1.0,
```
bzhangam: Should bound the ceil? It doesn't make sense we try to use all memory for queue

RexyZhang2001: Added the ceiling as 0.1

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Fixed by separating model payload size from estimated retained heap size.

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueueManager.java
RexyZhang2001: Fixed by performing queue resolution and admission within the same per-model `ConcurrentHashMap.compute` operation used by idle eviction

plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java
RexyZhang2001: Fixed by guarding queue mutation and totals with the same state lock.

bzhangam (approved):

jiapingzeng:
@RexyZhang2001 could you run spotless apply? Otherwise the PR looks good

RexyZhang2001:
@jiapingzeng Please help trigger the CI check again, thanks!

jiapingzeng (approved):
LGTM

PR #5000, opened 2026-09-03
Fix deletion of remote models stored without additional_config

### LINKED ISSUE #4999: [BUG] Cannot delete a model that was created with missing additional_config
**What is the bug?**
I am unable to delete a model that seems to have been created without an `additional_config` param. 

For context, we have a workflow that builds models that we use with semantic data type on our indices. This workflow accidentally ran against a cluster that was on OpenSearch 2.19 (before semantic data type was introduced). The cluster has since been upgraded to OpenSearch 3.5, and when trying to delete the model, we run into the following error:

```
> DELETE /_plugins/_ml/models/UquOWaABkVxAkscWvcWk
{
  "error": {
    "root_cause": [
      {
        "reason": "Space type must be provided in additional_config for remote text embedding model"
      }
    ],
    "reason": "Space type must be provided in additional_config for remote text embedding model"
  },
  "status": 400
}
```

The model has the following configuration:
```
      {
        "_index": ".plugins-ml-model",
        "_id": "UquOWaABkVxAkscWvcWk",
        "_version": 10,
        "_seq_no": 3026,
        "_primary_term": 1,
        "_score": 1,
        "_source": {
          "last_deployed_time": 1788208922100,
          "model_version": "19",
          "created_time": 1788208922020,
          "deploy_to_all_nodes": true,
          "is_hidden": false,
          "description": "MXS TEXT_EMBEDDING model for semantic fields (cosinesimil)",
          "model_state": "DEPLOY_FAILED",
          "planning_worker_node_count": 3,
          "model_id": "UquOWaABkVxAkscWvcWk",
          "rate_limiter": {
            "unit": "SECONDS",
            "limit": "200"
          },
          "model_config": {
            "model_type": "TEXT_EMBEDDING",
            "embedding_dimension": 768,
            "framework_type": "SENTENCE_TRANSFORMERS"
          },
          "auto_redeploy_retry_times": 1,
          "last_updated_time": 1788287922711,
          "name": "mxs_model_semantic_text_embedding_jina_v2_cosinesimil",
          "connector_id": "39OOWaABdEXvP4wovbAJ",
          "current_worker_node_count": 0,
          "model_group_id": "jAtl-5cBVjorKMYw3i93",
          "planning_worker_nodes": [
            "WF6-ldeRTuu7WxAjUELngw",
            "HzSngL_ISPmrU8WrfMj-Xw",
            "pgpIbgrlR-ODLpKQR_jHAA"
          ],
          "algorithm": "REMOTE"
        }
      }
```

It seems as though, additionally, I am unable to update the model in place:
```
> PUT /_plugins/_ml/models/UquOWaABkVxAkscWvcWk
{
  "name": "mxs_model_semantic_text_embedding_jina_v2_cosinesimil_old"
}

{
  "error": {
    "root_cause": [
      {
        "type": "status_exception",
        "reason": "Failed to find model to update with the provided model id: UquOWaABkVxAkscWvcWk"
      }
    ],
    "type": "status_exception",
    "reason": "Failed to find model to update with the provided model id: UquOWaABkVxAkscWvcWk"
  },
  "status": 404
}
```

This seems to be very similar to another issue that was closed without a resolution: https://github.com/opensearch-project/ml-commons/issues/3965 

Unable to reproduce

The DELETE command should go through cleanly, regardless of whether or not the model is misconfigured

 - OS: 3.5 (AWS Managed)

Ystk-hsn:
### Description
Remote text embedding models registered before the `space_type` requirement was introduced (#3786) have no `additional_config` in their stored document. `RemoteModelConfig` runs registration-time validation from its constructor, so reading such a document back from the model index fails with `IllegalArgumentException: Space type must be provided in additional_config for remote text embedding model`. The same parse call (`MLModel.parse`) sits in the delete / get / predict / deploy paths, so such a model cannot be deleted, updated, or undeployed (matches the `_undeploy` report in the issue).

This change keeps registration-time validation unchanged and skips it only when reading a stored document back:
- Added a `RemoteModelConfig.parse(parser, boolean validate)` overload. The existing `parse(parser)` still validates, so the user input paths (`MLRegisterModelInput` / `MLRegisterModelMetaInput`) behave exactly as before.
- `MLModel.parse` (stored documents) now uses the lenient overload.

### Testing
Regression tests at two levels, both reproduce the issue (fail before the fix, pass after):
- `MLModelTests.parseRemoteModel_legacyStoredDocWithoutSpaceType` — parse level, uses the stored document shape from the issue report
- `DeleteModelTransportActionTests.testDeleteLegacyRemoteModel_withoutSpaceType_Success` — action level

Existing `RemoteModelConfigTests` validation tests pass unchanged, confirming registration-time behavior is unaffected. Full `common` / `plugin` / `ml-algorithms` test suites pass locally.

### Notes for reviewers
- The `StreamInput` constructor already skips this validation, so this makes XContent deserialization of stored documents consistent with transport deserialization.
- One behavioral note: reading back a stored remote text embedding model no longer fails when `additional_config` is missing. Consumers that need `space_type` still need to handle its absence for pre-#3786 documents, as they did before that requirement existed.

### Related Issues
Resolves #4999

### Check List

### DIFF
--- a/common/src/main/java/org/opensearch/ml/common/MLModel.java
+++ b/common/src/main/java/org/opensearch/ml/common/MLModel.java
@@ -757,7 +757,9 @@ public class MLModel implements ToXContentObject {
                     } else if (FunctionName.TEXT_EMBEDDING.name().equals(algorithmName)) {
                         modelConfig = TextEmbeddingModelConfig.parse(parser);
                     } else if (FunctionName.REMOTE.name().equals(algorithmName)) {
-                        modelConfig = RemoteModelConfig.parse(parser);
+                        // Stored documents may predate the space_type requirement (issue #4999);
+                        // registration-time validation lives in MLRegisterModelInput / MLRegisterModelMetaInput.
+                        modelConfig = RemoteModelConfig.parse(parser, false);
                     } else {
                         modelConfig = BaseModelConfig.parse(parser);
                     }
--- a/common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
@@ -48,6 +48,20 @@ public class RemoteModelConfig extends BaseModelConfig {
         Integer modelMaxLength,
         Map<String, Object> additionalConfig
     ) {
+        this(modelType, embeddingDimension, frameworkType, allConfig, poolingMode, normalizeResult, modelMaxLength, additionalConfig, true);
+    }
+
+    private RemoteModelConfig(
+        String modelType,
+        Integer embeddingDimension,
+        BaseModelConfig.FrameworkType frameworkType,
+        String allConfig,
+        BaseModelConfig.PoolingMode poolingMode,
+        boolean normalizeResult,
+        Integer modelMaxLength,
+        Map<String, Object> additionalConfig,
+        boolean validate
+    ) {
         super(
             modelType,
             allConfig,
@@ -60,11 +74,21 @@ public class RemoteModelConfig extends BaseModelConfig {
             null,
             null
         );
-        validateNoDuplicateKeys(allConfig, additionalConfig);
-        validateTextEmbeddingConfig();
+        if (validate) {
+            validateNoDuplicateKeys(allConfig, additionalConfig);
+            validateTextEmbeddingConfig();
+        }
     }
 
     public static RemoteModelConfig parse(XContentParser parser) throws IOException {
+        return parse(parser, true);
+    }
+
+    /**
+     * Parses a remote model config. Pass {@code validate = false} when reading back a stored
+     * model document, which may predate the space_type requirement (issue #4999).
+     */
+    public static RemoteModelConfig parse(XContentParser parser, boolean validate) throws IOException {
         String modelType = null;
         Integer embeddingDimension = null;
         BaseModelConfig.FrameworkType frameworkType = null;
@@ -117,7 +141,8 @@ public class RemoteModelConfig extends BaseModelConfig {
             poolingMode,
             normalizeResult,
             modelMaxLength,
-            additionalConfig
+            additionalConfig,
+            validate
         );
     }
 
--- a/common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
+++ b/common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
@@ -75,7 +75,6 @@ public class RemoteModelConfig extends BaseModelConfig {
             null
         );
         if (validate) {
-            validateNoDuplicateKeys(allConfig, additionalConfig);
             validateTextEmbeddingConfig();
         }
     }
--- a/common/src/test/java/org/opensearch/ml/common/MLModelTests.java
+++ b/common/src/test/java/org/opensearch/ml/common/MLModelTests.java
@@ -8,6 +8,7 @@ package org.opensearch.ml.common;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.opensearch.common.xcontent.json.JsonXContent.jsonXContent;
 import static org.opensearch.core.xcontent.ToXContent.EMPTY_PARAMS;
 
 import java.io.IOException;
@@ -22,9 +23,11 @@ import org.json.JSONObject;
 import org.junit.Before;
 import org.junit.Test;
 import org.opensearch.common.io.stream.BytesStreamOutput;
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.authuser.User;
 import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.ml.common.connector.Connector;
@@ -602,5 +605,33 @@ public class MLModelTests {
         streamInput.setVersion(CommonValue.VERSION_3_5_0);
         MLModel deserialized = new MLModel(streamInput);
         assertNull(deserialized.getProvisionedBy());
+    }
+
+    // Issue #4999: documents stored before the space_type requirement (#3786) must still parse back.
+    @Test
+    public void parseRemoteModel_legacyStoredDocWithoutSpaceType() throws IOException {
+        String legacySource = "{"
+            + "\"name\":\"legacy_remote_text_embedding_model\","
+            + "\"algorithm\":\"REMOTE\","
+            + "\"model_version\":\"19\","
+            + "\"model_state\":\"DEPLOY_FAILED\","
+            + "\"connector_id\":\"test_connector_id\","
+            + "\"model_group_id\":\"test_model_group_id\","
+            + "\"model_config\":{"
+            + "\"model_type\":\"TEXT_EMBEDDING\","
+            + "\"embedding_dimension\":768,"
+            + "\"framework_type\":\"SENTENCE_TRANSFORMERS\""
+            + "}"
+            + "}";
+        try (
+            XContentParser parser = jsonXContent.createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, legacySource)
+        ) {
+            parser.nextToken();
+            MLModel parsed = MLModel.parse(parser, FunctionName.REMOTE.name());
+            assertEquals("legacy_remote_text_embedding_model", parsed.getName());
+            assertNotNull(parsed.getModelConfig());
+            assertEquals("TEXT_EMBEDDING", parsed.getModelConfig().getModelType());
+            assertNull(((BaseModelConfig) parsed.getModelConfig()).getAdditionalConfig());
+        }
     }
 }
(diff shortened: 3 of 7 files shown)


common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
```diff
@@ -60,11 +74,21 @@ public RemoteModelConfig(
             null,
             null
         );
-        validateNoDuplicateKeys(allConfig, additionalConfig);
-        validateTextEmbeddingConfig();
+        if (validate) {
+            validateNoDuplicateKeys(allConfig, additionalConfig);
+            validateTextEmbeddingConfig();
+        }
```
ylwu-amzn: **High-priority — this guard does not do what it looks like it does.**

`super(...)` on line 65 reaches `BaseModelConfig`'s constructor, which already calls `validateNoDuplicateKeys(allConfig, additionalConfig)` unconditionally (`BaseModelConfig.java:85`). So line 78 is a redundant second invocation, and gating it behind `validate` has no effect — duplicate-key validation still runs on the lenient read path.

Verified by probe: `RemoteModelConfig.parse(parser, false)` on

```json
{"model_type":"text_embedding","embedding_dimension":768,
 "framework_type":"SENTENCE_TRANSFORMERS",
 "all_config":"{\"space_type\":\"l2\"}",
 "additional_config":{"space_type":"cosinesimil"}}
```

still throws `Duplicate keys found in both all_config and additional_config: space_type`.

Two consequences:

1. Suggest dropping line 78 so the guard says what it actually does. (The double-call is pre-existing, not introduced here — but it's now load-bearing for readability.)
2. It means the Code-Diff-Analyzer's low finding above is factually wrong: the only check actually bypassed is `validateTextEmbeddingConfig`. Worth noting so nobody changes the code in response to it.

Either way this isn't a live BWC gap — `additional_config` was itself introduced by #3786, so pre-3.1 documents can't hold a collision and `validateNoDuplicateKeys` early-returns on a null map.

**Related question, not a request for this PR:** should the read path be lenient about duplicate keys too? A document that acquires a collision *after* creation would brick the model exactly as #4999 describes. `_update` sends a partial doc (`MLUpdateModelInput.toXContent:176-177`) and OpenSearch merges nested objects recursively, `MLUpdateModelInput` parses `model_config` via `BaseModelConfig.parse` (`:310`), and `UpdateModelTransportAction.validateModelConfig` (`:246-278`) never checks for cross-field collisions — so a request setting only `all_config` validates against a null `additionalConfig` and early-returns. I have not traced the final `UpdateRequest` construction, so treat this as a question rather than a claim.

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
```diff
@@ -60,11 +74,21 @@ public RemoteModelConfig(
...
     }
 
     public static RemoteModelConfig parse(XContentParser parser) throws IOException {
+        return parse(parser, true);
+    }
```
ylwu-amzn: **High-priority — nothing pins the validating overload, so the PR's central claim is untested.**

"Registration-time validation is unchanged" now rests entirely on this line defaulting to `true`. But every validation test in `RemoteModelConfigTests` goes through the **builder**, not through `parse`:

- `textEmbedding_MissingEmbeddingDimension` (`:86`) — builder
- `textEmbedding_MissingFrameworkType` (`:93`) — builder
- `textEmbedding_MissingSpaceType` (`:100`) — builder
- `parse()` (`:129`) — parses a **valid** config

So changing this line to `return parse(parser, false);` would leave the whole suite green while silently disabling validation on the real user-input paths (`MLRegisterModelInput.java:510,678`, `MLRegisterModelMetaInput.java:442`). Two overloads that differ only by a boolean is exactly the case that needs a negative test at the seam.

Could you add a parse-level negative case — feed the legacy document shape to `RemoteModelConfig.parse(parser)` and assert it still throws `Space type must be provided in additional_config for remote text embedding model`? That pins the default and makes the "user paths behave exactly as before" claim self-enforcing.

common/src/main/java/org/opensearch/ml/common/MLModel.java
```diff
@@ -757,7 +757,9 @@ public static MLModel parse(XContentParser parser, String algorithmName) throws
                     } else if (FunctionName.TEXT_EMBEDDING.name().equals(algorithmName)) {
                         modelConfig = TextEmbeddingModelConfig.parse(parser);
                     } else if (FunctionName.REMOTE.name().equals(algorithmName)) {
-                        modelConfig = RemoteModelConfig.parse(parser);
+                        // Stored documents may predate the space_type requirement (issue #4999);
+                        // registration-time validation lives in MLRegisterModelInput / MLRegisterModelMetaInput.
+                        modelConfig = RemoteModelConfig.parse(parser, false);
```
ylwu-amzn: Two nits, both optional.

1. `parse(parser, false)` doesn't read as "stored document" without the comment carrying it. A named factory — `parseStoredDocument(parser)` delegating to a private `parse(parser, validate)` — would be self-documenting and let the comment go away. Your comment does mitigate it, so take it or leave it.

2. Worth one clause on why REMOTE is the only lenient branch: `TextEmbeddingModelConfig.parse` (line 758) and `BaseModelConfig.parse` (line 764) still construct-and-validate on this same stored-document path. That isn't a live break — for local `TEXT_EMBEDDING`, `embedding_dimension` and `framework_type` predate 3.1 as required fields — but the asymmetry now looks accidental, and a future reader could "harmonize" it in either direction. Something like "REMOTE is the only algorithm that gained a required config field after GA" would settle it.

common/src/test/java/org/opensearch/ml/common/MLModelTests.java
```diff
@@ -603,4 +606,32 @@ public void writeTo_ReadFrom_ProvisionedBy_OldVersion_IsNull() throws IOExceptio
...
+            + "\"model_type\":\"TEXT_EMBEDDING\","
+            + "\"embedding_dimension\":768,"
+            + "\"framework_type\":\"SENTENCE_TRANSFORMERS\""
+            + "}"
+            + "}";
+        try (
+            XContentParser parser = jsonXContent.createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, legacySource)
+        ) {
+            parser.nextToken();
+            MLModel parsed = MLModel.parse(parser, FunctionName.REMOTE.name());
```
ylwu-amzn: Confirming this test has teeth: I reverted `MLModel.java:762` back to `RemoteModelConfig.parse(parser)` and re-ran `MLModelTests` — this test fails (`23 tests completed, 1 failed`) and passes with the fix restored. Using the document shape from the issue report rather than a synthetic minimal case was the right call.

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
ylwu-amzn, line 79: Line 78 is a no-op — `super(...)` already calls `validateNoDuplicateKeys` unconditionally (`BaseModelConfig.java:85`), so `validate=false` doesn't bypass it. Verified: lenient parse of a doc with colliding `all_config`/`additional_config` keys still throws. Suggest deleting line 78.

Side effect: the Code-Diff-Analyzer finding above is wrong — only `validateTextEmbeddingConfig` is actually bypassed.

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
ylwu-amzn, line 84: Could you add a negative test for this overload? All the validation tests in `RemoteModelConfigTests` go through the builder, and `parse()` only covers a valid config — so flipping this to `parse(parser, false)` would keep the suite green while silently disabling registration validation. Asserting `parse(parser)` still throws on the legacy shape pins it.

jiapingzeng (approved):
LGTM!

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
Ystk-hsn: You're right, deleted. So only validateTextEmbeddingConfig is gated by the flag now, duplicate keys are　always checked in the superclass.

common/src/main/java/org/opensearch/ml/common/model/RemoteModelConfig.java
Ystk-hsn: Added parse_textEmbeddingMissingSpaceType for this, plus a lenient counterpart on the same input. Also double checked the new test goes red if the default ever flips to lenient.

ylwu-amzn (commented):
Both comments verified fixed, thanks.

One last small ask: `UpdateModelTransportAction.java:232-236` turns any `getModel` failure into a 404, which is why the reporter was told the model didn't exist. Keep 404 only when `mlModel == null` and propagate `e` otherwise — ~5 lines, existing tests stay green.

Ystk-hsn:
@ylwu-amzn Done. 404 is now returned only when the model is null, other failures propagate as-is. Added a test pinning that a getModel failure reaches the caller unmasked. thanks for the review

Ystk-hsn:
@ylwu-amzn Gentle ping, the last ask is addressed and waiting for another look.

mingshl:
@bzhangam this space_type is used in neural search, would you review this change

ylwu-amzn (approved):

bzhangam:
Wonder if we should decouple the parse and validateTextEmbeddingConfig and invoke validation explicitly on the write path to make the code more readable.

Besides also notice in update model path we implement the validation again to support validating with both the new config and existing config. But it kind makes the validation rule exist in two places. Would it be better to group them in one places? Maybe a new help function?

Not blocking. Just some thoughts on how to make the codes cleaner and easier to maintain. @Ystk-hsn @mingshl

Ystk-hsn:
@bzhangam Thanks for the thoughts. 
I kept validation in the constructor because the sibling config classes do the same (TextEmbeddingModelConfig, BaseModelConfig), and validate-by-default means a future write path can't silently skip the check. Decoupling makes sense to me if we do it for the whole family in one pass.
Agree on the update path. validateModelConfig in UpdateModelTransportAction repeats the same three rules and error strings, so grouping them into a shared helper sounds worth doing.

PR #4394, opened 2025-11-04
[FEATURE] Add an option to turn on and off the certificate validation of llm connectors

akolarkunnu:
### Description
Introduced a new connector parameter "skip_ssl_verification" to enable/disable ssl verification of llm connectors.

Tested with and without setting this parameter and I am able to connect a lll server by disabling ssl verification.

### Related Issues
Resolves #4371

### Check List
- [X] New functionality includes testing.
- [X] New functionality has been documented.
- [X] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [X] Public documentation issue/PR - will create it 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

## Summary by CodeRabbit

* **New Features**
  * Added support for skipping SSL certificate verification for remote connectors (AWS and HTTP JSON).

* **API Updates**
  * HTTP client factory now accepts an additional option to control SSL verification; connector interfaces expose a parameter key for this setting.

* **Tests**
  * Tests updated to cover the SSL verification toggle.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5003, opened 2026-09-07
Add Client Certificate Authentication Support for MCP Connectors #4971

### LINKED ISSUE #4971: [FEATURE] Add Client Certificate Authentication Support for MCP Connectors

 PR #4731 (issue #4515) adds client certificate / mutual TLS authentication for connectors, but only for the http protocol path
  (HttpJsonConnectorExecutor). MCP connectors protocols mcp_sse and mcp_streamable_http cannot use it, because they build their HTTP client through a completely different stack:

  - The MCP executors create an MCP Java SDK transport (HttpClientSseClientTransport / HttpClientStreamableHttpTransport), which wraps the JDK java.net.http.HttpClient.
  - They pass only connect/read timeouts (customizeClient) and headers derived from decrypted credentials (customizeRequest). They never call MLHttpClientFactory, so nothing TLS-related in client_config reaches the MCP transport.

  Consequences today:
  1. An MCP server that requires client certificates cannot be registered as an MCP connector at all, only header/token auth works.
  2. Setting mutual_tls_enabled: true (and keystore_type, cert credentials) on an mcp_sse / mcp_streamable_http connector is accepted without error and then silently ignored, no client certificate is presented and no validation failure surfaces.
  3. Related: skip_ssl_verification also appears to have no effect on the MCP transports, so the TLS-config gap is broader than mTLS alone.

 Wire the existing certificate machinery from #4731 into both MCP executors.

ishag4:
### Description

PR #4731 added mutual TLS to the `http` path only, so TLS settings in `client_config` were parsed for MCP connectors and then silently ignored: no client certificate was presented, and no error surfaced.

`MLSslContextFactory` bridges the gap. The AWS path takes `KeyManager[]`/`TrustManager[]` through its own provider hooks; the JDK client only accepts an assembled `SSLContext`. The factory reuses `CertificateProcessor#resolveMtls`, so no certificate parsing or validation is duplicated, and returns `null` when no TLS customization is requested.

Also fixes `connection_timeout` for MCP connectors: the SDK applies the `customizeClient` consumer eagerly and then overwrites `connectTimeout` with its own 10s default in `build()`, so the configured value never took effect. Setting it on the transport builder resolves this.

## Manual testing

Verified on a single-node Docker cluster (`opensearchstaging/opensearch:3.9.0`) with the plugin built from this branch installed, security enabled, over HTTPS.

- **MCP target:** a real MCP server behind an mTLS-terminating proxy configured with `ssl.CERT_REQUIRED`, logging every peer certificate it receives.
- **`http` target:** a live mTLS-protected LLM endpoint returning real chat completions.
- **Agent:** `conversational`, with the MCP connector supplied via `mcp_connectors`.

### TLS plumbing, before involving OpenSearch

| # | Check | Result |
|---|---|---|
| 1 | `curl` without a client certificate | Refused by the server at handshake |
| 2 | `curl` with the client certificate | `event: endpoint` — MCP stream established |
| 3 | Container → host reachability | Handshake refused rather than connection refused, confirming the container can route to the MCP endpoint |

### The feature this PR adds — `mcp_sse` over mutual TLS

| # | Scenario | Result |
|---|---|---|
| 4 | Agent executed against an mTLS MCP connector | Server logged `GET /sse  CLIENT_CERT: ... commonName=ml-commons-test-client` — the certificate was presented |
| 5 | Certificate used for the whole session, not just connect | One `GET /sse` plus a `POST /message` per MCP call, each showing the client identity |
| 6 | Full MCP protocol over mTLS | `initialize`, `tools/list` and `tools/call` all completed |
| 7 | Tool definitions reached the LLM | Agent named all 14 tools from the server; `prompt_tokens` ≈ 1300 |
| 8 | LLM selected and invoked an MCP tool | `{"action": "echo", "action_input": "banana"}`, correct result returned |
| 9 | MCP connector without mTLS | Created successfully — the non-mTLS path still accepts a connector |

Before this change nothing connected at all: the certificate log stayed empty and no tools reached the LLM.

### `http` protocol — the #4731 path, unregressed

| # | Scenario | Result |
|---|---|---|
| 10 | mTLS with `PEM` keystore | Connector created, real chat completion returned |
| 11 | mTLS with `PKCS12` keystore | Connector created, real chat completion returned |
| 12 | Connector without mTLS | Created successfully — the non-mTLS path still accepts a connector |

### Create-time validation (@ylwu-amzn's review)

All five were **accepted at create time** before this change.

| # | Scenario | Result |
|---|---|---|
| 13 | `mutual_tls_enabled` with no certificates | 400, names `client_cert_pem` / `client_key_pem` |
| 14 | `keystore_type: PKCS12` with no keystore | 400, names `client_cert_pkcs12` |
| 15 | `mutual_tls_enabled` + `skip_ssl_verification` | 400, explains why the combination is unsafe |
| 16 | `keystore_type: jks` | 400, `Unsupported keystore type` |
| 17 | Credential present but blank (`client_cert_pem` / `client_key_pem` set to `""`) | 400, names the field |

Each returns 400 rather than 500, and names the offending field. Rows 13 and 15 were also run against an `mcp_sse` connector; validation is protocol-independent, so the remaining rows exercise the same code path.

Scenario 17 came out of manual testing: an unset shell variable produced `"client_cert_pkcs12": ""`, which passed a presence-only check and then failed on every predict with `Failed to build mutual TLS managers`. Presence alone is no longer sufficient; the value must be non-blank.

### Update path

| # | Scenario | Result |
|---|---|---|
| 18 | Description-only change to a working mTLS connector | Accepted — credentials are stripped before update sees them, so presence cannot be judged there |
| 19 | Enabling mTLS without resending certificates | Accepted, for the same reason; still caught at execution as before |
| 20 | Adding `skip_ssl_verification` to an mTLS connector | 400 — closes the bypass around the create-time check |
| 21 | Update supplying incomplete certificates | 400, names the missing field |
| 22 | Unrelated field on a non-mTLS connector | Accepted |

### Gating

| # | Scenario | Result |
|---|---|---|
| 23 | Feature flag | Creation rejected until `plugins.ml_commons.mcp_connector_enabled` was set |
| 24 | Trusted-endpoint regex | Creation rejected until the endpoint regex matched |

### Not covered manually — unit tests only

- `mcp_streamable_http` — identical wiring to `mcp_sse`, which the manual runs used
- `skip_ssl_verification` on its own — covered by `MLSslContextFactoryTlsHandshakeTest`, which proves it bypasses both chain validation and hostname verification
- the `connection_timeout` fix — covered by asserting `connectTimeout(Duration.ofSeconds(30))` on the transport builder and its absence on the client builder

## Automated tests

24 new tests; 644 passing across `:opensearch-ml-common` and `:opensearch-ml-algorithms`.

- `MLSslContextFactoryTest` — config-to-`SSLContext` behaviour, including that a default config
  yields `null` so existing connectors are unaffected.
- `MLSslContextFactoryTlsHandshakeTest` — complete TLS handshakes driven through `SSLEngine` in
  memory (no sockets, so no ports or timeouts in CI), asserting the server actually receives the
  client certificate, and that `skip_ssl_verification` bypasses chain *and* hostname
  verification. Negative controls assert failures come from TLS validation rather than a stalled
  harness.
- Executor tests — the `SSLContext` reaches the JDK client builder for both mutual TLS and
  `skip_ssl_verification`, is left untouched by default, and misconfiguration fails loudly.

### Related Issues
Resolves #4971 

### Check List


mingshl:
is the ssl verification setting inherited from the connector setting? skip_ssl_verification / TrustAllX509TrustManager is an explicit MITM-enabling bypass. Right now it's guarded only by a per-connector flag + a warning log, which means anyone who can create a connector can disable TLS verification. Can we put this behind a cluster-level setting / feature flag (default off) so it can't be silently enabled in production? I asked earlier if we can adopt from the core level setting for ssl verification

ishag4:
> is the ssl verification setting inherited from the connector setting? skip_ssl_verification / TrustAllX509TrustManager is an explicit MITM-enabling bypass. Right now it's guarded only by a per-connector flag + a warning log, which means anyone who can create a connector can disable TLS verification. Can we put this behind a cluster-level setting / feature flag (default off) so it can't be silently enabled in production? I asked earlier if we can adopt from the core level setting for ssl verification

Hi @mingshl , 
Good catch! A few clarifications: 

**Is it inherited from the connector setting?** Yes, `ConnectorClientConfig.skipSslVerification`, set per connector.

**Scope:** `skip_ssl_verification` is not introduced here. Skipping certificate validation was added in #4394 ("Add an option to turn on and off the certificate validation of llm connectors", Feb 2026) via `TRUST_ALL_CERTIFICATES` in
`MLHttpClientFactory`, then moved into `client_config` as `skipSslVerification` by #4616. This PR extends the same existing capability to the MCP transports so `client_config` stops being silently ignored, the MITM exposure you describe already exists today for `http` connectors.

**Core setting:** I don't think we can adopt one. `plugins.security.ssl.*` governs OpenSearch's own inbound/transport TLS, not whether outbound connectors may skip verification. That would need a new ml-commons setting.

**Proposal:** I agree it should be gated, and there's already a precedent: `plugins.ml_commons.connector.private_ip_enabled` is a cluster-level, default-off gate on a comparable connector network capability, threaded through to the executors. A matching `plugins.ml_commons.connector.skip_ssl_verification_enabled` (default `false`) would fit
that design directly.

I'd prefer that as a follow-up PR rather than here, for two reasons: it needs to cover the `http` path too or the same flag behaves differently per protocol, and gating `http` is a breaking change for anyone relying on it today, so it warrants its own discussion and release note. Happy to open the issue!

Note: the mutual-TLS path already rejects the dangerous combination: `mutual_tls_enabled` + `skip_ssl_verification` together is refused outright (`CertificateProcessor#validateCertificateConfig`), so a client certificate can never be presented over an unverified connection.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
```diff
@@ -124,4 +140,96 @@ public void testUnimplementedMethods_ThrowUnsupportedOperationException() {
...
+            MockedStatic<McpClient> mockedClient = mockStatic(McpClient.class)
+        ) {
+            mockedTransport.when(() -> HttpClientSseClientTransport.builder(anyString())).thenReturn(transportBuilder);
+            mockedClient.when(() -> McpClient.sync(any(McpClientTransport.class))).thenReturn(builder);
+
+            new McpConnectorExecutor(mockConnector).getMcpToolSpecs();
+        }
+
+        verify(transportBuilder).connectTimeout(any());
+        verify(clientBuilder, never()).connectTimeout(any());
```
ylwu-amzn: **this guard can't fail.** It runs before `accept(clientBuilder)` on line 173, so `clientBuilder` still has zero interactions. Move the `never()` check below the `accept(...)`.

Also `any()` doesn't pin the value: `connectTimeout(Duration.ofSeconds(30))` would.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpStreamableHttpConnectorExecutorTest.java
```diff
@@ -127,4 +143,97 @@ public void testUnimplementedMethods_ThrowUnsupportedOperationException() {
...
+            MockedStatic<McpClient> mockedClient = mockStatic(McpClient.class)
+        ) {
+            mockedTransport.when(() -> HttpClientStreamableHttpTransport.builder(anyString())).thenReturn(transportBuilder);
+            mockedClient.when(() -> McpClient.sync(any(McpClientTransport.class))).thenReturn(builder);
+
+            new McpStreamableHttpConnectorExecutor(mockConnector).getMcpToolSpecs();
+        }
+
+        verify(transportBuilder).connectTimeout(any());
+        verify(clientBuilder, never()).connectTimeout(any());
```
ylwu-amzn: Same as the SSE test: the `never()` check precedes `accept(clientBuilder)` on line 175, so it can't fail. Move it after, and assert `Duration.ofSeconds(30)` rather than `any()`.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
```diff
@@ -124,4 +140,96 @@ public void testUnimplementedMethods_ThrowUnsupportedOperationException() {
...
+    }
+
+    @Test
+    public void getMcpToolSpecs_skipSslVerification_installsSslContext() {
+        when(mockConnector.getConnectorClientConfig())
+            .thenReturn(ConnectorClientConfig.builder().connectionTimeout(30).readTimeout(30).skipSslVerification(true).build());
+
+        HttpClient.Builder clientBuilder = captureClientCustomization();
+
+        verify(clientBuilder).sslContext(any(SSLContext.class));
```
ylwu-amzn: `any(SSLContext.class)` passes for either branch — swapping trust-all and mTLS in `MLSslContextFactory.create` would still pass this and the mTLS test. `MLSslContextFactoryTlsHandshakeTest` covers the semantics, just not the wiring.

common/src/test/java/org/opensearch/ml/common/httpclient/MLSslContextFactoryTest.java
```diff
@@ -0,0 +1,183 @@
...
+    private String loadCertificateFromFile(String filename) throws IOException {
+        try (InputStream in = getClass().getResourceAsStream("/certificates/" + filename)) {
+            assertNotNull("Missing test fixture on classpath: /certificates/" + filename, in);
+            return new String(in.readAllBytes(), StandardCharsets.UTF_8);
+        }
+    }
+
+    private void putValidPemCredentials() throws IOException {
+        credentials.put(CLIENT_CERT_PEM_FIELD, loadCertificateFromFile("test-client-cert.pem"));
+        credentials.put(CLIENT_KEY_PEM_FIELD, loadCertificateFromFile("test-client-key-pkcs8.pem"));
```
ylwu-amzn: The description lists PKCS12 as unit-test-covered, but every case here is `keystoreType("PEM")`. `CertificateProcessor`'s PKCS12 path (keystore + password) isn't exercised.

common/src/main/java/org/opensearch/ml/common/httpclient/MLSslContextFactory.java
```diff
@@ -0,0 +1,117 @@
...
+        return null;
+    }
+
+    private static SSLContext buildContext(KeyManager[] keyManagers, TrustManager[] trustManagers, String description) {
+        return doPrivileged(() -> {
+            try {
+                SSLContext sslContext = SSLContext.getInstance("TLS");
+                sslContext.init(keyManagers, trustManagers, null);
+                return sslContext;
+            } catch (NoSuchAlgorithmException | KeyManagementException e) {
```
ylwu-amzn: `MLHttpClientCacheManager.createHttpClient` maps `SecurityException` to `MLException("SSL security policy violation: ...")`. Here it escapes raw — and `create()` sits outside the executor's try — so a plugin-policy denial reaches the user unexplained.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutor.java
```diff
@@ -83,6 +89,10 @@ public List<MLToolSpec> getMcpToolSpecs() {
             ? connector.getParameters().get(SSE_ENDPOINT_FIELD)
             : MCP_DEFAULT_SSE_ENDPOINT;
         List<MLToolSpec> mcpToolSpecs = new ArrayList<>();
+
+        SSLContext sslContext = MLSslContextFactory
+            .create(super.getConnectorClientConfig(), connector.getDecryptedCredential(), certificateProcessor);
```
ylwu-amzn: Rebuilds the SSLContext, re-parsing the PEM, on every call; the AWS path caches via `MLHttpClientCacheManager`. Likely fine for tool listing — deliberate?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
ylwu-amzn, line 169: **This guard can't fail.** It runs before `accept(clientBuilder)` on line 173, so `clientBuilder` still has zero interactions. re-added `clientBuilder.connectTimeout(...)` inside `customizeClient` and this class still passed. Move the `never()` check below the `accept(...)`.

Also `any()` doesn't pin the value: `connectTimeout(Duration.ofSeconds(30))` would.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpStreamableHttpConnectorExecutorTest.java
ylwu-amzn, line 171: Same as the SSE test: the `never()` check precedes `accept(clientBuilder)` on line 175, so it can't fail. Move it after, and assert `Duration.ofSeconds(30)` rather than `any()`.

common/src/main/java/org/opensearch/ml/common/httpclient/MLSslContextFactory.java
```diff
@@ -0,0 +1,117 @@
...
+    public static SSLContext create(
+        ConnectorClientConfig config,
+        Map<String, String> decryptedCredentials,
+        CertificateProcessor certificateProcessor
+    ) {
+        if (config == null) {
+            return null;
+        }
+
+        CertificateProcessor.MtlsManagers mtlsManagers = certificateProcessor.resolveMtls(config, decryptedCredentials);
```
ylwu-amzn, line 55: nothing validates mTLS config at connector-create time (`validateCertificateConfig` is only reachable from `resolveMtls`, i.e. execution time). So an MCP connector with `mutual_tls_enabled: true` and missing certs works today — silently ignored — and hard-fails on every agent call after this.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
ylwu-amzn, line 191: `any(SSLContext.class)` passes for either branch — swapping trust-all and mTLS in `MLSslContextFactory.create` would still pass this and the mTLS test. `MLSslContextFactoryTlsHandshakeTest` covers the semantics, just not the wiring.

common/src/test/java/org/opensearch/ml/common/httpclient/MLSslContextFactoryTest.java
ylwu-amzn: The description lists PKCS12 as unit-test-covered, but every case here is `keystoreType("PEM")`. `CertificateProcessor`'s PKCS12 path (keystore + password) isn't exercised.

common/src/main/java/org/opensearch/ml/common/httpclient/MLSslContextFactory.java
ylwu-amzn: `MLHttpClientCacheManager.createHttpClient` maps `SecurityException` to `MLException("SSL security policy violation: ...")`. Here it escapes raw — and `create()` sits outside the executor's try — so a plugin-policy denial reaches the user unexplained.

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutor.java
ylwu-amzn, line 94: Runs per agent invocation (`MLChatAgentRunner`, `...V2`, `MLPlanExecuteAndReflect`), so PEM + RSA key parsing and `SSLContext.init` repeat on every user query; the AWS path caches via `MLHttpClientCacheManager`. Only a few ms against LLM round trips — deliberate?

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
ishag4: fixed

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
ishag4: fixed

common/src/main/java/org/opensearch/ml/common/httpclient/MLSslContextFactory.java
ishag4: Added the check

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpStreamableHttpConnectorExecutorTest.java
ishag4: fixed

ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutor.java
ishag4: Deliberate, yes. getMcpToolSpecs() constructs a fresh HttpClientSseClientTransport and McpSyncClient per call and then runs initialize() + listTools(), so every invocation already pays a new TLS handshake and two MCP round trips, caching just the SSLContext wouldn't avoid that. The AWS path caches an SdkAsyncHttpClient, a pooled Netty client that's much more expensive to rebuild, and its cache key has to encode ksType to stay correct across connector updates. Reproducing that machinery for a few ms against LLM latency didn't seem worth the invalidation risk. If profiling later says otherwise, the right fix is caching the MCP client itself rather than the context alone — happy to do that as a follow-up.

ishag4: Deliberate, yes. getMcpToolSpecs() constructs a fresh HttpClientSseClientTransport and McpSyncClient per call and then runs initialize() + listTools(), so every invocation already pays a new TLS handshake and two MCP round trips, caching just the SSLContext wouldn't avoid that. The client it belongs to is intentionally per-run: it's attached to the tool specs as a runtime resource so tools can call tools/call on the same session, and the agent runtime closes it via closeGracefully() when the run ends.

The AWS path caches an SdkAsyncHttpClient, a pooled Netty client, much more expensive to rebuild and its cache key has to encode ksType to stay correct across connector updates. Reproducing that machinery for a few ms against LLM latency didn't seem worth the invalidation risk. If profiling later says otherwise, the right fix is caching the MCP client itself rather than the context alone, and I'm happy to take that as a follow-up.

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
```diff
@@ -539,6 +541,119 @@ public void testExecuteWithBlockedDynamicHeaderThrowsException() {
...
+        updateConnectorTransportAction.doExecute(task, updateRequest, actionListener);
+
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener).onFailure(argumentCaptor.capture());
+        Exception failure = argumentCaptor.getValue();
+        assertTrue("Expected IllegalArgumentException, got: " + failure.getClass(), failure instanceof IllegalArgumentException);
+        assertTrue(failure.getMessage(), failure.getMessage().contains("client_key_pem"));
+    }
+
+    private void stubStoredConnector(ConnectorClientConfig clientConfig, Map<String, String> credential) {
```
ylwu-amzn: **Blocking — the three new tests never reach the validation.** This class stubs `validateConnectorAccess` per test, not in `setup()`, so here it returns Mockito's default `false` and `doExecute` bails with "You don't have permission to update the connector" before `validateMutualTlsConfig` runs.

`testUpdate_withIncompleteCredentials_isRejected` and `testUpdate_addingSkipSslVerification_isRejected` fail on this branch (all 4 build jobs are red); `withCredentialsStripped_isAccepted` passes vacuously.

Adding this line makes all three pass — the production logic is fine:

```java
doReturn(true).when(connectorAccessControlHelper).validateConnectorAccess(any(Client.class), any(Connector.class));
```

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -174,6 +182,31 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
...
+     * validating presence unconditionally would reject every update to a working mutual TLS
+     * connector.
+     */
+    private static void validateMutualTlsConfig(Connector connector, MLCreateConnectorInput updateContent) {
+        try {
+            Map<String, String> credential = updateContent.getCredential();
+            if (credential != null && !credential.isEmpty()) {
+                CERTIFICATE_PROCESSOR.validateCertificateConfig(connector.getConnectorClientConfig(), credential);
+            } else {
+                CERTIFICATE_PROCESSOR.validateMutualTlsSettings(connector.getConnectorClientConfig());
```
ylwu-amzn: **High-priority — mTLS can still be turned on with no certificates.** With no credentials in the request only the skip-SSL check runs, so `client_config: {mutual_tls_enabled: true}` alone is accepted on a connector with no stored certs — the same per-request hard failure the create-time check was added to prevent. Confirmed: the update completes with no error.

The transition is decidable from the request alone — if `updateContent.getConnectorClientConfig()` enables mTLS and `getCredential()` is empty, require credentials.

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
```diff
@@ -133,6 +137,10 @@ protected void doExecute(Task task, ActionRequest request, ActionListener<Update
                                     AbstractConnector.validateConnectorHeaders(headers, connector.getProtocol());
                                 }
                             }
+                            // Before encrypt(): reject an invalid mutual TLS configuration up front rather
+                            // than on every request the updated connector serves.
+                            validateMutualTlsConfig(connector, updateContent);
```
ylwu-amzn: `mutual_tls_enabled` and `skip_ssl_verification` both predate this PR, so connectors with both set can already exist. Validating on every update makes a description-only change to one return 400.

Gating this on `updateContent.getConnectorClientConfig() != null` keeps unrelated updates working and still closes the create-time bypass, since changing either flag requires sending the config.

plugin/src/test/java/org/opensearch/ml/action/connector/TransportCreateConnectorActionTests.java
```diff
@@ -995,4 +997,95 @@ private MLCreateConnectorInput createConnectorInputWithUrl(String url) {
...
+
+    /**
+     * Certificate validation is the only failure these tests care about; anything else is noise from
+     * the mocked indexing path. Asserting on the message rather than the exception type keeps this
+     * from silently passing if the type ever changes.
+     */
+    private void assertNoCertificateValidationFailure() {
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        // Mockito.atLeast, qualified: LuceneTestCase inherits an atLeast(int) that would win here.
+        verify(actionListener, Mockito.atLeast(0)).onFailure(argumentCaptor.capture());
```
ylwu-amzn: `atLeast(0)` plus a substring scan passes whenever the action fails for some other reason — that is how the missing permission stub in `UpdateConnectorTransportActionTests` stayed hidden. A `verify(actionListener).onResponse(any())` on the happy-path tests would catch it. Same pattern at `UpdateConnectorTransportActionTests:560`.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
```diff
@@ -124,4 +143,126 @@ public void testUnimplementedMethods_ThrowUnsupportedOperationException() {
...
+        // The exact instance the factory produced must be the one installed.
+        verify(clientBuilder).sslContext(same(expected));
+        Assert.assertEquals(Boolean.TRUE, capturedConfig.getSkipSslVerification());
+    }
+
+    @Test
+    public void getMcpToolSpecs_validMutualTlsConfig_installsSslContext() throws Exception {
+        Map<String, String> credentials = new HashMap<>();
+        credentials.put(CertificateProcessor.CLIENT_CERT_PEM_FIELD, readFixture("mtls-client-cert.pem"));
+        credentials.put(CertificateProcessor.CLIENT_KEY_PEM_FIELD, readFixture("mtls-client-key.pem"));
```
ylwu-amzn: `MLSslContextFactory.create` is stubbed with `any()` matchers now, so these fixtures are never parsed. Either capture the credentials argument and assert the cert fields reached the factory, or drop the fixture read.

common/src/main/java/org/opensearch/ml/common/connector/Connector.java
```diff
@@ -215,6 +215,9 @@ default void validateResolvedEndpoint(String resolvedUrl, List<String> urlRegexe
 
     Map<String, String> getDecryptedCredential();
 
+    /** The raw credential map, which may still be encrypted. Needed to validate config before encryption. */
+    Map<String, String> getCredential();
```
ylwu-amzn: Adding an abstract method to this published interface breaks any external implementor, and widens the public surface to the raw credential map. Likely fine in practice — flagging rather than objecting.

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
ylwu-amzn: **The three new tests never reach the validation.** `validateConnectorAccess` is stubbed per test rather than in `setup()`, and `stubStoredConnector` misses it, so it returns `false` and `doExecute` bails with "You don't have permission to update the connector" before `validateMutualTlsConfig` runs.

Two tests fail on this branch; `withCredentialsStripped_isAccepted` passes vacuously — a permission error contains none of the substrings it forbids. One line here fixes all three:

```java
doReturn(true).when(connectorAccessControlHelper).validateConnectorAccess(any(Client.class), any(Connector.class));
```

plugin/src/main/java/org/opensearch/ml/action/connector/UpdateConnectorTransportAction.java
ylwu-amzn: An update can still enable mTLS with no certificates: `connector.update()` replaces `connectorClientConfig` wholesale, so `{mutual_tls_enabled: true}` without `credential` reaches only the skip-SSL check and is accepted. Not a regression — `main` has no create-time check either — so a follow-up is fine.

Note that requiring credentials whenever the request enables mTLS would reject legitimate config-only updates, which must resend `mutual_tls_enabled: true` without certs. Capturing `connector.getConnectorClientConfig()` before `update()` lets you require them only on the false→true transition.

plugin/src/test/java/org/opensearch/ml/action/connector/TransportCreateConnectorActionTests.java
ylwu-amzn: `atLeast(0)` plus a substring scan passes whenever the action fails for any other reason — that's how the missing permission stub in `UpdateConnectorTransportActionTests` stayed hidden. `verify(actionListener).onResponse(any())` on the happy paths would catch it. Same pattern at `UpdateConnectorTransportActionTests:560`.

ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/remote/McpConnectorExecutorTest.java
ylwu-amzn: Nit: `MLSslContextFactory.create` is stubbed with `any()`, so these fixtures are never parsed and only the captured config is asserted. Either assert the cert fields reached the factory, or drop the read.

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
ylwu-amzn, line 611: **The three new tests never reach the validation.** `validateConnectorAccess` is stubbed per test rather than in `setup()`, and `stubStoredConnector` misses it, so it returns `false` and `doExecute` bails with "You don't have permission to update the connector" before `validateMutualTlsConfig` runs.

Two tests fail on this branch; `withCredentialsStripped_isAccepted` passes vacuously — a permission error contains none of the substrings it forbids. One line here fixes all three:

```java
doReturn(true).when(connectorAccessControlHelper).validateConnectorAccess(any(Client.class), any(Connector.class));
```

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
```diff
@@ -539,6 +541,119 @@ public void testExecuteWithBlockedDynamicHeaderThrowsException() {
...
+    public void testUpdate_mutualTlsConnector_withCredentialsStripped_isAccepted() {
+        stubStoredConnector(ConnectorClientConfig.builder().mutualTlsEnabled(true).keystoreType("PEM").build(), null);
+        when(updateRequest.getUpdateContent())
+            .thenReturn(MLCreateConnectorInput.builder().updateConnector(true).description("just a description change").build());
+        stubUpdateSucceeds();
+
+        updateConnectorTransportAction.doExecute(task, updateRequest, actionListener);
+
+        ArgumentCaptor<Exception> argumentCaptor = ArgumentCaptor.forClass(Exception.class);
+        verify(actionListener, Mockito.atLeast(0)).onFailure(argumentCaptor.capture());
```
ylwu-amzn: Even with the permission stub added, this test passes whether or not the update actually succeeds — `atLeast(0)` plus a substring scan only fails when the error happens to mention certificates. It's also what kept the missing stub from showing up here.

`stubUpdateSucceeds()` is already in place, so `verify(actionListener).onResponse(any())` would make it assert what its name claims.

ishag4: fixed

plugin/src/test/java/org/opensearch/ml/action/connector/UpdateConnectorTransportActionTests.java
ishag4: fixed

PR #5012, opened 2026-09-15
Fix models stuck in DEPLOYING when one model document fails to parse

Rishikesh1159:
### Description
  
  `MLSyncUpCron.refreshModelState()` had three related defects in how it reads model documents from `.plugins-ml-model`.
  
  **Symptom:** models can remain in `DEPLOYING` indefinitely with no obvious cause, and the state does not recover on its own. A single malformed document is enough to stall state
  reconciliation for *every* model on the cluster, repeating on each sync-up cycle.
  
  **1. One bad document stopped reconciliation for every model.** The entire search-hit loop was wrapped in a single `try/catch`, so as soon as any one document failed to parse,
  the loop aborted and `bulkUpdateModelState()` never ran. No model in that batch was refreshed. Because the offending document persists, this repeated on every cron cycle. Error
  handling is now per document: the offending model is logged with its id and skipped, and the remaining models are still refreshed.
  
  **2. `last_updated_time` was hard-cast to `Long`.** The field is mapped as `{"type": "date", "format": "strict_date_time||epoch_millis"}`, so an ISO-8601 string is a legal
  stored value, but `(Long) sourceAsMap.get(...)` throws `ClassCastException` on it. This was the most likely trigger for defect 1:
  
      java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Long
          at org.opensearch.ml.cluster.MLSyncUpCron.lambda$refreshModelState$21(MLSyncUpCron.java:347)
  
  It is now parsed leniently: epoch millis, a numeric string, or an ISO-8601 date-time via `OffsetDateTime.parse`. `OffsetDateTime` is used rather than `Instant` because
  `strict_date_time` permits a numeric UTC offset (for example `2020-01-01T00:00:00.000+05:30`), which `Instant.parse` rejects on JDK 11. An unparseable value is logged and
  treated as unknown rather than throwing.
  
  The two worker-count fields were read with `(int)` casts and are now read as `Number`, which also removes an NPE when the field is present but null. Since these fields are
  mapped as `integer`, a non-numeric value is malformed rather than an alternate representation, so it is logged rather than silently defaulting to 0, which would otherwise block
  the `DEPLOYED` transition.
  
  **3. The deploy grace window had no upper bound.** `getNewModelState()` treated a model as still within its deploy grace period whenever `lastUpdateTime +
  DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS > now`. A `last_updated_time` in the future makes that permanently true, so the model stays pinned in `DEPLOYING` and is never reconciled to
  `DEPLOY_FAILED`.
  
  The window is now bounded on both sides by the same grace constant:
  
      lastUpdateTime < now + DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS
          && lastUpdateTime + DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS > now
  
  A far-future value still expires, while a small forward skew is tolerated. That matters because the timestamp is written by the deploying node but compared against the cluster
  manager's clock, so a strict `<= now` check would mark a model that is legitimately mid-deploy as `DEPLOY_FAILED`.
  
  #### How to reproduce (before this change)
  
  On a cluster where `.plugins-ml-model` is writable, register a model, then set its state to `DEPLOYING` with zero worker nodes and:
  
  - `last_updated_time` in the future (for example `4102444800000`). The model stays `DEPLOYING` indefinitely instead of becoming `DEPLOY_FAILED`.
  - `last_updated_time` as `"2020-01-01T00:00:00.000Z"`. The cron logs `Failed to parse model search response` with a `ClassCastException` once per cycle, and unrelated healthy
  models in the same batch also stop being reconciled. Repairing that one document lets all affected models reconcile on the next cycle.
  
  #### Testing
  
  Six regression tests added to `MLSyncUpCronTests`:
  
  - **future epoch-millis timestamp** guards that a far-future value is reconciled to `DEPLOY_FAILED`
  - **unparseable timestamp** guards that it is treated as unknown and still reconciled
  - **malformed document alongside a healthy one** guards that the healthy model is still refreshed
  - **ISO-8601 with `Z` suffix, inside the grace window** guards that the value is actually parsed, not silently dropped
  - **ISO-8601 with `+05:30` offset, inside the grace window** guards that offset formats are parsed
  - **`last_updated_time` 5s ahead of now** guards that forward clock skew does not mark a deploying model failed
  
  The first three fail against the unpatched code. The last three are regression guards for the new behaviour, verified by deliberately breaking it: stubbing the ISO branch to
  return `null` fails both ISO tests, and reverting the grace condition to `<= now` fails the skew test. The in-grace tests deliberately assert that no update is issued, so a
  parser that returns `null` instead of throwing is caught.
  
  `MLSyncUpCronTests` is 22/22. The full `opensearch-ml-plugin` unit suite (3324 tests) passes, and `:opensearch-ml-plugin:check` is green including
  `jacocoTestCoverageVerification`.
  
  ### Related Issues
  
  N/A - no tracked issue.
  
  ### Check List
  
  For more information on following Developer Certificate of Origin and signing off your commits, please check
  [here](https://github.com/opensearch-project/ml-commons/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -447,6 +458,41 @@ private MLModelState getNewModelState(
...
     }
 
+    /**
+     * Reads {@code last_updated_time} from a model document.
+     *
+     * The field is mapped as {@code "type": "date", "format": "strict_date_time||epoch_millis"}, so a
+     * document may legitimately carry either epoch millis or an ISO-8601 string. Returns null when the
+     * value is absent or cannot be interpreted, which callers treat as "unknown".
+     */
+    private static Long parseLastUpdateTime(String modelId, Object raw) {
```
mingshl, line 472: Nice fix — the per-document isolation and the future-timestamp guard are both clearly needed. A few thoughts:

On the last_updated_time parsing:

The `ClassCastException` exists because the mapping declares "format": "strict_date_time||epoch_millis" but the read path assumed the value is always Long. The lenient read-side parsing here is the correct long-term fix — even if we normalized writes to epoch millis going forward, existing documents and the dual-format mapping would remain, so this code is necessary regardless.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -447,6 +458,41 @@ private MLModelState getNewModelState(
...
+        if (raw instanceof Number) {
+            return ((Number) raw).longValue();
+        }
+        if (raw instanceof String) {
+            String value = (String) raw;
+            try {
+                return Long.parseLong(value);
+            } catch (NumberFormatException ignored) {
+                try {
+                    return Instant.parse(value).toEpochMilli();
```
mingshl: Instant.parse() only handles the Z suffix, but strict_date_time also allows offset formats like +08:00. Consider OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME).toInstant().toEpochMilli() for full coverage — otherwise those values silently become null.

mingshl:
One more thing — potential link to known stuck-in-DEPLOYING issues:

We've seen clusters where models get stuck in DEPLOYING indefinitely with no clear root cause. Looking at this fix, there's a plausible chain: if even one document in .plugins-ml-model has last_updated_time as an ISO-8601 string (legal per the mapping), the ClassCastException would abort the entire refreshModelState loop — meaning bulkUpdateModelState() never runs for any model in that batch, not just the offending one. Since the bad document persists, this repeats every cron cycle, so affected models stay stuck forever.

Could you check whether this has been the cause? Specifically:
1. On a cluster with stuck DEPLOYING models, search the logs for ClassCastException + MLSyncUpCron — if it's been firing every cron cycle, that's the smoking gun.
2. Check .plugins-ml-model for any documents where last_updated_time is stored as a string rather than a long — those would be the trigger.

If this confirms it, it would be worth calling it out in the PR description and potentially filing a backport for active maintenance branches, since the stuck state is unrecoverable without this fix or manual intervention.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Rishikesh1159: Thanks. Agreed on the read-side point normalizing writes alone wouldn't be sufficient since the dual-format mapping and any existing documents remain, so the parser has to tolerate both regardless.

Rishikesh1159:
> One more thing — potential link to known stuck-in-DEPLOYING issues:
> 
> We've seen clusters where models get stuck in DEPLOYING indefinitely with no clear root cause. Looking at this fix, there's a plausible chain: if even one document in .plugins-ml-model has last_updated_time as an ISO-8601 string (legal per the mapping), the ClassCastException would abort the entire refreshModelState loop — meaning bulkUpdateModelState() never runs for any model in that batch, not just the offending one. Since the bad document persists, this repeats every cron cycle, so affected models stay stuck forever.
> 
> Could you check whether this has been the cause? Specifically:
> 
> On a cluster with stuck DEPLOYING models, search the logs for ClassCastException + MLSyncUpCron — if it's been firing every cron cycle, that's the smoking gun.
> Check .plugins-ml-model for any documents where last_updated_time is stored as a string rather than a long — those would be the trigger.
> If this confirms it, it would be worth calling it out in the PR description and potentially filing a backport for active maintenance branches, since the stuck state is unrecoverable without this fix or manual intervention.

Yeah, I hit exactly that. Put two docs in one batch — one with last_updated_time as an ISO
  string, one ordinary model stuck in DEPLOYING — and the healthy one never got reconciled
  either. The CCE repeated every cron cycle. Fixed the one bad doc and both cleared within a
  cycle. So it can definitely leave models stuck indefinitely, and it won't recover on its own.
  
  I couldn't tie it to a specific reported case though — I don't have log or index access on an
  affected cluster. If someone does, worth grepping for ClassCastException + MLSyncUpCron
  recurring on the sync-up interval, and checking .plugins-ml-model for last_updated_time
  stored as a string instead of a long.
  
  One thing that should narrow the search: this needs a write to .plugins-ml-model, which
  system-index protection blocks by default. So likely either that protection was off, or
  something external wrote the doc — reindex, snapshot restore, migration tooling.
  
  Will add the backport labels and call out the stuck-in-DEPLOYING symptom in the description.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Rishikesh1159: Good catch, changed. I tested this across JDKs before switching, and the behaviour splits:
  
                   2020-01-01T00:00:00.000+05:30
  JDK 21/17        Instant.parse OK      OffsetDateTime.parse OK
  JDK 11           Instant.parse THROWS  OffsetDateTime.parse OK
  
  ISO_INSTANT was relaxed in JDK 12, so Instant.parse actually does accept offsets on main.
  But 2.x supports JDK 11, where it throws so this matters for the backports and the change
  is worth making.
  
  Switched to OffsetDateTime.parse, which covers everything Instant.parse does plus offsets on
  every supported JDK, so no chained fallback is needed. Added a regression test using +05:30.
  Values with no offset at all still return null with a warning, which I think is correct since

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -337,48 +339,55 @@ void refreshModelState(Map<String, Set<String>> modelWorkerNodes, Map<String, Se
...
+                                int planningWorkerNodeCount = parseIntOrDefault(
+                                    sourceAsMap.get(MLModel.PLANNING_WORKER_NODE_COUNT_FIELD),
+                                    0
+                                );
+                                int currentWorkerNodeCountInIndex = parseIntOrDefault(
+                                    sourceAsMap.get(MLModel.CURRENT_WORKER_NODE_COUNT_FIELD),
+                                    0
+                                );
+                                boolean deployToAllNodes = sourceAsMap.containsKey(MLModel.DEPLOY_TO_ALL_NODES_FIELD)
+                                    && (boolean) sourceAsMap.get(MLModel.DEPLOY_TO_ALL_NODES_FIELD);
```
ylwu-amzn: `(boolean)` still NPEs on an explicit null, and the `(List<String>)` below still throws on a scalar. Contained by the new catch — just noting they didn't get hardened with the other three.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -337,48 +339,55 @@ void refreshModelState(Map<String, Set<String>> modelWorkerNodes, Map<String, Se
...
+                                    planningWorkerNodeCount,
+                                    currentWorkerNodeCountInIndex
+                                );
+                                if (mlModelState != null) {
+                                    newModelStates.put(modelId, mlModelState);
+                                }
+                            } catch (Exception e) {
+                                // Skip this model only. A single unparseable document must not prevent the
+                                // remaining models in this batch from being refreshed.
+                                log.error("Failed to parse model [{}] while refreshing model state, skipping it", modelId, e);
```
ylwu-amzn: The bad document persists by definition, so this logs a full stack trace every cron cycle forever. `log.warn` without the throwable?

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -413,11 +422,14 @@ private MLModelState getNewModelState(
...
-            && state != MLModelState.DEPLOY_FAILED
-            && !(state == MLModelState.DEPLOYING
-                && lastUpdateTime != null
-                && lastUpdateTime + DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS > Instant.now().toEpochMilli())) {
+        long now = Instant.now().toEpochMilli();
+        // A last update time in the future must not be treated as a live grace window, otherwise the model
+        // can never be reconciled out of DEPLOYING.
+        boolean withinDeployGracePeriod = state == MLModelState.DEPLOYING
+            && lastUpdateTime != null
+            && lastUpdateTime <= now
```
ylwu-amzn: `last_updated_time` comes from the deploy node's clock, `now` from the cluster manager's. Any forward skew makes this false, so a model mid-deploy (DEPLOYING, 0 workers, running deploy task) is marked DEPLOY_FAILED.

Suggest `lastUpdateTime < now + DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS` instead of `<= now`: a far-future value still expires, skew doesn't.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -447,6 +459,43 @@ private MLModelState getNewModelState(
...
+                try {
+                    // OffsetDateTime rather than Instant: strict_date_time permits a numeric offset
+                    // (for example 2020-01-01T00:00:00.000+05:30), which Instant.parse rejects on JDK 11.
+                    return OffsetDateTime.parse(value).toInstant().toEpochMilli();
+                } catch (DateTimeParseException ignored2) {
+                    // fall through to the warning below
+                }
+            }
+        }
+        log.warn("Ignoring unparseable {} [{}] on model [{}]", MLModel.LAST_UPDATED_TIME_FIELD, raw, modelId);
```
ylwu-amzn: `null` here means "no grace window", not "unknown" — a model genuinely mid-deploy with a bad timestamp gets DEPLOY_FAILED on the next tick. Fine if intended, but worth saying so in the description.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
```diff
@@ -447,6 +459,43 @@ private MLModelState getNewModelState(
...
+        return null;
+    }
+
+    /** Reads an integer counter that may be stored as any numeric type. */
+    private static int parseIntOrDefault(Object raw, int defaultValue) {
+        return raw instanceof Number ? ((Number) raw).intValue() : defaultValue;
```
ylwu-amzn: Silently returns 0 for a non-numeric value, unlike `parseLastUpdateTime`. A `planning_worker_node_count` of `"2"` becomes 0, which blocks the DEPLOYED transition at L442. Worth a `log.warn`.
```
private static int parseIntOrDefault(Object raw, int defaultValue) {
    if (raw instanceof Number) {
        return ((Number) raw).intValue();
    }
    if (raw != null) {
        log.warn("Ignoring non-numeric worker node count [{}]", raw);
    }
    return defaultValue;
}
```

plugin/src/test/java/org/opensearch/ml/cluster/MLSyncUpCronTests.java
```diff
@@ -428,6 +428,128 @@ public void testRefreshModelState_NotResetState_DeployingInGraceTime() {
...
+            ActionListener<SearchResponse> actionListener = invocation.getArgument(1);
+            actionListener
+                .onResponse(
+                    createSearchModelResponseWithRawLastUpdateTime(
+                        "modelId",
+                        "tenantId",
+                        MLModelState.DEPLOYING,
+                        2,
+                        null,
+                        "2020-01-01T00:00:00.000+05:30"
```
ylwu-amzn: This test passes with the offset bug still in. 2020 is outside the grace window either way, so DEPLOY_FAILED holds whether the string parsed or fell through to null — I broke the parser and all 21 tests stayed green. Same for the `Z` case at L465.

Use a timestamp inside the grace window and assert `verify(client, never()).bulk(...)`.

plugin/src/test/java/org/opensearch/ml/cluster/MLSyncUpCronTests.java
```diff
@@ -503,4 +644,52 @@ private SearchResponse createSearchModelResponse(
...
+    /**
+     * Builds a two-hit response where the first document carries an invalid model_state and the second is a
+     * well-formed model stuck in DEPLOYING outside its grace window.
+     */
+    private SearchResponse createSearchModelResponseWithInvalidFirstModel(String healthyModelId, long healthyLastUpdateTime)
+        throws IOException {
+        XContentBuilder broken = TestHelper.builder();
+        broken.startObject();
+        broken.field(CommonValue.TENANT_ID_FIELD, "tenantId");
+        broken.field(MLModel.MODEL_STATE_FIELD, "NOT_A_REAL_STATE");
```
ylwu-amzn: The search filters `model_state` to the 7 valid states, so this input can't reach the loop. An unknown `algorithm` hits the same catch via a reachable path.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Rishikesh1159: Good catch and you're right.  I incorrectly  introduced that. Fixed in latest commit 2b7c113c6.
  
  Ironically it's the same class of problem I dropped from the original attached patch (a
  zero-tolerance future check), and I reintroduced a variant of it with `<= now`. Took your
  suggestion as-is:
  
      lastUpdateTime < now + DEPLOY_MODEL_TASK_GRACE_TIME_IN_MS
  
  That bounds the window on both sides using the existing constant rather than inventing a
  separate tolerance: a far-future value still expires, but skew up to the grace period is
  absorbed. Added testRefreshModelState_ForwardClockSkew_StillWithinGracePeriod covering a
  timestamp 5s ahead I verified it fails if the condition is reverted to `<= now`.

plugin/src/main/java/org/opensearch/ml/cluster/MLSyncUpCron.java
Rishikesh1159: Agreed, fixed in 2b7c113c6. These fields are mapped as `integer`, so unlike
  last_updated_time a non-numeric value is genuinely malformed rather than an alternate
  representation logging and falling back is the right treatment, and silently returning 0
  blocking the DEPLOYED transition is exactly the kind of thing that should not be invisible.
  Added the warning with the model id, to match parseLastUpdateTime.

plugin/src/test/java/org/opensearch/ml/cluster/MLSyncUpCronTests.java
Rishikesh1159: You're right, and thanks for actually breaking it to check, that's a real hole. Both ISO
  tests used 2020 timestamps, which are outside the grace window whether the value parsed or
  fell through to null, so the DEPLOY_FAILED assertion held either way. They tested "doesn't
  throw", not "parses correctly".
  
  Reworked both in latest revision 2b7c113c6 to use a timestamp inside the grace window and assert
  verify(client, never()).bulk(...), so a parser returning null now fails the test. Verified
  the same way you did: with the ISO branch stubbed to return null, both tests fail
  (NeverWantedButInvoked). Previously that sabotage produced zero failures.
  
  Kept the year-2100 and unparseable cases asserting DEPLOY_FAILED, since for those the
  DEPLOY_FAILED outcome is the correct behaviour being checked.

ylwu-amzn (approved):

PR #5013, opened 2026-09-15
Rename resource sharing feature flag to the non-experimental key

DarshitChanpura:
### Description

Resource sharing graduates out of experimental in <https://github.com/opensearch-project/security/pull/6348>, which drops the `experimental.` segment from the feature-flag settings with **no deprecated fallback**:

| Before | After |
|---|---|
| `plugins.security.experimental.resource_sharing.enabled` | `plugins.security.resource_sharing.enabled` |
| `plugins.security.experimental.resource_sharing.protected_types` | `plugins.security.resource_sharing.protected_types` |

Updates the integTest cluster settings in `plugin/build.gradle` to the new keys.

Only the setting strings change. The `ConfigConstants` constant names (`OPENSEARCH_RESOURCE_SHARING_ENABLED`, `OPENSEARCH_RESOURCE_SHARING_PROTECTED_TYPES`) are unchanged, so no Java/Kotlin references are affected. The gradle `-Dresource_sharing.enabled` system property that gates the resource-sharing CI matrix is a separate toggle and is left as is.

No workflow in this repo sets `-Dresource_sharing.enabled=true`, so this block is unreachable in Actions: CI will neither fail on the old key nor catch a wrong one. The change still has to land in 3.9 together with security PR 6348 (merged), because a node handed the new key by a pre-3.9 security build fails to start on `unknown setting` — that just has to be verified locally rather than by CI here.

### Related Issues

Companion to <https://github.com/opensearch-project/security/pull/6348>. Related to <https://github.com/opensearch-project/security/issues/4500>

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).


mingshl (approved):

mingshl:
@jiapingzeng would you review this change?

plugin/build.gradle
```diff
@@ -413,8 +413,8 @@ def configureClusterPlugins(cluster, jobSchedZip, securityZip, securityEnabled)
                 node.setting "plugins.security.restapi.roles_enabled",                 "[\"all_access\", \"security_rest_api_access\"]"
                 node.setting "plugins.security.system_indices.enabled",                "true"
                 if (System.getProperty("resource_sharing.enabled") == "true") {
-                    node.setting "plugins.security.experimental.resource_sharing.enabled", "true"
-                    node.setting("plugins.security.experimental.resource_sharing.protected_types", "[\"ml-model-group\"]")
+                    node.setting "plugins.security.resource_sharing.enabled", "true"
```
ylwu-amzn, line 416: **High-priority — please spell out the backport intent.** The description says this needs to land in 3.8, but 3.8 is already GA (core `3.8.0` tag, 2026-07-31), and both the `3.8` and `3.7` branches still carry the old key. Backporting a no-fallback setting rename into a released minor means a 3.8.0 → 3.8.1 node with the old key in `opensearch.yml` won't start. Is that cost being accepted deliberately (experimental feature, few users expected), or should the rename be 3.9-only?

plugin/build.gradle
```diff
@@ -413,8 +413,8 @@ def configureClusterPlugins(cluster, jobSchedZip, securityZip, securityEnabled)
...
+                    node.setting("plugins.security.resource_sharing.protected_types", "[\"ml-model-group\"]")
```
ylwu-amzn, line 417: The migration notes in security#6348 already cover the unknown-setting and archived-setting cases; one ml-commons-side consequence worth adding there: once the archived persistent setting drops the feature back to `false`, `ModelAccessControlHelper.shouldUseResourceAuthz()` (:296) goes false and `validateModelGroupAccess` silently falls back to the backend-roles path — share grants stop being honored with no error. Was `SettingUpgrader` considered? It rewrites renamed keys on upgrade instead of archiving them, which would remove both cases.

plugin/build.gradle
```diff
@@ -413,8 +413,8 @@ def configureClusterPlugins(cluster, jobSchedZip, securityZip, securityEnabled)
                 node.setting "plugins.security.restapi.roles_enabled",                 "[\"all_access\", \"security_rest_api_access\"]"
                 node.setting "plugins.security.system_indices.enabled",                "true"
                 if (System.getProperty("resource_sharing.enabled") == "true") {
```
ylwu-amzn, line 415: FYI, no workflow here sets `-Dresource_sharing.enabled=true` (`grep -rn resource_sharing .github/` → no hits), so this block is unreachable in Actions: the "security integTest jobs here will fail to start a node" won't actually happen in CI, and a wrong key wouldn't be caught either. Merge timing probably needs a different signal.

DarshitChanpura: CI needs to have a dedicated Security enabled workflow. Other plugins follow this model. I can address this in fast-follow.

plugin/build.gradle
DarshitChanpura: that was an error in description. It should be 3.9. Updated now.

plugin/build.gradle
DarshitChanpura: Done now in https://github.com/opensearch-project/security/pull/6513 - the pre-graduation names become deprecated fallbacks plus an upgrader, so nothing needs changing before an upgrade and the silent revert to backend roles can't happen. Removal tracked for 4.0 in https://github.com/opensearch-project/security/issues/6514.

ylwu-amzn (approved):

Issue #5033, opened 2026-09-18
[BUG] Batch inference splitting silently stores wrong vectors in `knn_vector` fields, and several batch/queue config values are unvalidated or ignored


ylwu-amzn:
### What is the bug?

**Primary (HIGH — silent data corruption).** When a sub-batch comes back from the provider with a
different number of results than it had items, the **split** path concatenates whatever arrived and
returns HTTP 200. Because `combine()` concatenates per sub-batch, an anomaly that is harmless in an
unsplit call gets relocated into the **middle** of the combined list, displacing every subsequent
real embedding. The result is a wrong vector permanently stored in a `knn_vector` field, with no
error at any layer and nothing in the response or the log to indicate it.

The **queue** path added in #4996 guards exactly this case
(`plugin/src/main/java/org/opensearch/ml/batch/ModelBatchQueue.java:344-346`), so the check is
known-necessary and simply missing on the older split path
(`plugin/src/main/java/org/opensearch/ml/batch/TextDocsBatchableInput.java:68` `combine()`, reached
from `plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java:154`).

### How can one reproduce the bug?

**Count mismatch is accepted with HTTP 200 on the split path, rejected on the queue path**

Against a mock provider that always returns 3 embeddings regardless of input count:

```
model with batch_inference_config {"max_items_per_request":2}          # SPLIT path
POST /_plugins/_ml/_predict/text_embedding/<id> {"text_docs":["s","ss","sss","ssss","sssss","ssssss"]}
  -> HTTP 200, 6 docs in -> 9 embeddings out, first components [1,2,3,1,2,3,1,2,3]

same config plus {"queue":{"enabled":true,"flush_timeout_ms":100}}     # QUEUE path
  -> 500 illegal_state_exception "Model returned 3 results for a sub-batch of 2 items,
          so results cannot be routed back to their callers"           # correct
```

With a mock returning *fewer* results, 6 docs yield 3 embeddings, still HTTP 200.

**Confirmed corruption of an indexed `knn_vector` field**

Using a provider that returns an honest embedding per input **plus one extra trailing vector**
(component 0 = `999`) — which is what several real embedding APIs do when they skip or coalesce
inputs — and neural-search's `text_embedding` ingest processor over a single document with three text
fields of lengths 1/2/3:

```
WITH batch_inference_config {"max_items_per_request":2}      -> SPLIT
[...]

PR #5064, opened 2026-09-21
Rename batch_inference_config queue field to dynamic_batching and change Cluster settings

RexyZhang2001:
### Description

Moves the server-side batch inference feature onto the `dynamic_batching` vocabulary, simplifies the per-model queue lifecycle, and closes the remaining follow-up items from #5033. The feature is unreleased, so no backward compatibility is kept for the old names.

**Naming**
- Nested XContent field inside `batch_inference_config`: `queue` → `dynamic_batching`.
- Config class `BatchQueueConfig` → `DynamicBatchingConfig`; accessors `getDynamicBatching()` / `isDynamicBatchingEnabled()`; the unknown-field error message and javadoc name the `dynamic_batching` block.
- Queue mechanism classes `ModelBatchQueue` → `DynamicBatchingQueue`, `ModelBatchQueueManager` → `DynamicBatchingQueueManager` (history preserved via rename).
- Cluster settings:
  - `plugins.ml_commons.batch_queue.memory_fraction` → `plugins.ml_commons.dynamic_batching.memory.size`
  - `plugins.ml_commons.batch_queue.memory_floor` → `plugins.ml_commons.dynamic_batching.memory.size.min`
  - `plugins.ml_commons.batch_queue.memory_ceiling` → `plugins.ml_commons.dynamic_batching.memory.size.max`
  - `plugins.ml_commons.batch_queue.idle_ttl` → **removed**

**Queue lifecycle**
- Removed the periodic idle-queue TTL sweep. A per-model queue is removed as soon as it drains empty — including the too-large and rejected paths that never enqueue — and recreated on the next request. Removal runs through the map's per-key `compute`, so it serializes with admission and cannot strand a concurrent request.

**Follow-ups from #5033**
- **`PUT /_plugins/_ml/models/<id>` can now update `batch_inference_config`.** `MLUpdateModelInput` carries the field (version-gated on 3.9), the update writes it into the model doc, the change triggers a model-cache refresh, and `updateModelCache` now refreshes the cached model info so online predict picks up the new config instead of the value cached at deploy time.
- **Dynamic batching no longer coalesces requests with different caller algorithms.** The caller's algorithm (e.g. `TEXT_EMBEDDING` vs `SPARSE_ENCODING`) is normalized to `REMOTE` before dispatch; it is now captured once on `MLInput` (transport-only, version-gated on 3.9) and folded into the batch group key.
- **Unsplit remote embedding/sparse calls now validate result count.** A single call whose response does not contain exactly one result per input text fails, instead of returning HTTP 200 with misaligned results. Scoped to REMOTE `TEXT_EMBEDDING` / `SPARSE_ENCODING` / `SPARSE_TOKENIZE`; generic remote and streaming responses are unchanged.

The StreamInput/StreamOutput wire format for existing data is unchanged; the two new fields are gated on `VERSION_3_9_0`.

### Testing

Unit tests updated/added across the renamed field, config, settings, and mechanism classes, plus: empty-queue removal after flush and on the too-large/rejected paths; `batch_inference_config` parse/serialize/version-gating and update-triggers-cache-refresh; caller-algorithm capture-once and group-key separation; and unsplit result-count validation (short, long, config-present single call, generic-remote passthrough, streaming passthrough). Focused tests and Spotless pass.

### Related Issues
Closes the remaining items in #5033. Follow-up to #4996 / #4898.

### Check List


common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
```diff
@@ -151,7 +151,7 @@ public static BatchInferenceConfig parse(XContentParser parser, boolean rejectUn
                                 + ", "
                                 + MAX_BYTES_PER_REQUEST_FIELD
                                 + ", "
-                                + QUEUE_FIELD
+                                + DYNAMIC_BATCHING_FIELD
```
ylwu-amzn, line 154: `BatchQueueConfig.java:107` still hardcodes the old name in the nested block's error: `Unsupported field [flush_timeout] in the batch_inference_config queue block.` A user who typos inside `dynamic_batching` gets pointed at a "queue block" that no longer exists on the API. Suggest referencing `DYNAMIC_BATCHING_FIELD` there; the javadoc at `BatchQueueConfig.java:24` is stale for the same reason.

`BatchInferenceConfigTest:235` only asserts on `[flushTimeoutMs]`, so it passes either way.

common/src/main/java/org/opensearch/ml/common/model/BatchInferenceConfig.java
```diff
@@ -105,7 +105,7 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         builder.field(MAX_ITEMS_PER_REQUEST_FIELD, maxItemsPerRequest);
         builder.field(MAX_BYTES_PER_REQUEST_FIELD, maxBytesPerRequest);
         if (queue != null) {
-            builder.field(QUEUE_FIELD, queue);
+            builder.field(DYNAMIC_BATCHING_FIELD, queue);
```
ylwu-amzn: Intentional per the description — flagging once only: the contract now says `dynamic_batching` while `getQueue()`/`isQueueEnabled()`, public API of `common`, still say queue. If that's ever going to align, unreleased is the cheap moment.

common/src/test/java/org/opensearch/ml/common/model/BatchInferenceConfigTest.java
```diff
@@ -229,7 +229,7 @@ public void rejectsUnknownFieldOnTheApiInputPath() throws IOException {
 
     @Test
     public void rejectsUnknownFieldInsideQueueBlockOnTheApiInputPath() throws IOException {
```
ylwu-amzn: Nit, batched: test names still say "queue" — this one, `parse_RejectsUnknownFieldInBatchInferenceQueueConfig` (MLRegisterModelInputTest:310), and the `some_future_queue_field` key (MLModelTests:132).

common/src/main/java/org/opensearch/ml/common/model/BatchQueueConfig.java
ylwu-amzn: Type name not matching: `BatchQueueConfig`. Suggest `DynamicBatchingConfig`

RexyZhang2001: changed all `BatchQueueConfig` naming to `DynamicBatchingConfig` to be consistent

ylwu-amzn (approved):

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
```diff
@@ -129,12 +131,14 @@ void completeEnqueue(QueueEntry entry, EnqueueDecision decision) {
                         modelId,
                         entry.getRetainedByteSize(),
                         budget.getMaxBytes(),
-                        ML_COMMONS_BATCH_QUEUE_MEMORY_FRACTION.getKey(),
-                        ML_COMMONS_BATCH_QUEUE_MEMORY_CEILING.getKey()
+                        ML_COMMONS_DYNAMIC_BATCHING_MEMORY_SIZE.getKey(),
+                        ML_COMMONS_DYNAMIC_BATCHING_MEMORY_SIZE_MAX.getKey()
                     );
+                notifyIdle(); // nothing entered the queue; drop it if this request created an empty one
```
heemin32, line 144: Why call `notifyIdle()` here? Why this case is different from flush case?

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueueManager.java
```diff
@@ -152,16 +136,25 @@ private void notifyFailure(ActionListener<MLTaskResponse> listener, Exception fa
...
-        for (String modelId : queues.keySet()) {
-            queues.computeIfPresent(modelId, (id, queue) -> now - queue.getLastUsedNanos() > ttl && queue.isIdle() ? null : queue);
-        }
+    /**
+     * Removes a queue once it has drained empty. Runs under the map's per-key compute lock, which the enqueue
+     * path also holds, so a request that arrives first keeps the queue and a request that arrives after removal
+     * simply recreates one. The identity check leaves a replacement queue (installed on a config change) in place.
+     */
+    private void removeIfIdle(ModelBatchQueue queue) {
+        queues.computeIfPresent(queue.getModelId(), (id, existing) -> existing == queue && existing.isIdle() ? null : existing);
```
heemin32, line 145: What if new item is added after `existing.isIdle()` is executed but before setting it as null?

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
RexyZhang2001: The queue gets created and added to the map first, then `offer()` happens. 

Four `offer()` outcomes:
- SCHEDULE_TIMER – entry queued; the timer flushes it later → empties → `notifyIdle()`.
- FLUSH – entry queued; flushes right away → empties → `notifyIdle()`.
- **TOO_LARGE – returns before adding, so the entry never enters the queue.**
- **REJECTED – returns before adding, so the entry never enters the queue.**

For the first two senarios, flush eventually empties the queue and removes it. 
For TOO_LARGE/REJECTED, nothing is ever queued, so flush never runs — and if this was the model's first request, that empty queue just sits in the map with nothing to clean it up. The TTL sweep used to remove empty queues; but we now remove it, so we need `notifyIdle()` here to do it.

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueueManager.java
RexyZhang2001: This senario won't happen because of the `ConcurrentHashMap`'s per-key locking mechanism. 
Adding an item only happens via **enqueue**, which calls `offer()` inside `queues.compute(modelId, ...)`. **Removal** happens inside `computeIfPresent(modelId, ...)`. Both run under the same per-key lock, so for one model they can't overlap — an add can't slip in between `isIdle()` and returning null.

So only two orderings are possible, and neither loses a request:
1. enqueue first → item is in the queue → `isIdle()` sees it's non-empty → we keep it.
2. removal first → empty queue is removed → enqueue finds no queue in the map → creates a fresh one.

heemin32 (approved):

RexyZhang2001:
New commit fixes these remaining unaddressed issues from https://github.com/opensearch-project/ml-commons/issues/5033:   

1. PUT /_plugins/_ml/models/<id> silently discards batch_inference_config
2. The queue group key cannot distinguish text_embedding from sparse_encoding for the same remote model
3. A count-short response is pre-existing

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
```diff
@@ -157,49 +156,38 @@ private MLCommonsSettings() {}
             Setting.Property.Dynamic
         );
 
-    public static final Setting<Double> ML_COMMONS_BATCH_QUEUE_MEMORY_FRACTION = Setting
+    public static final Setting<Double> ML_COMMONS_DYNAMIC_BATCHING_MEMORY_SIZE = Setting
         .doubleSetting(
-            ML_PLUGIN_SETTING_PREFIX + "batch_queue.memory_fraction",
+            ML_PLUGIN_SETTING_PREFIX + "dynamic_batching.memory.size",
```
ylwu-amzn: `memory.size` is a heap fraction, not a size — default 0.01, max 0.1, and the field it feeds is named `memoryFraction`. But `.size.min`/`.size.max` are byte values (64mb/512mb), so `size: 0.01` sits next to `size.min: 64mb` and someone will try `memory.size: 100mb`. `memory.size` is also the prefix of `memory.size.min`, which nested YAML can't express (a key can't be both scalar and map).

Suggest `memory.fraction` + `memory.min`/`.max`.

Also: `release-notes/opensearch-ml-commons.release-notes-3.9.0.0.md:30` still references the `batch_queue` setting removed here.

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
```diff
@@ -245,7 +257,11 @@ private void dispatch(List<QueueEntry> batch) {
                 notifyFailure(entry, new IllegalStateException("Could not compute a batch group key for the predict request"));
                 continue;
             }
-            GroupKey groupId = new GroupKey(entry.getInput().getInputDataset().getInputDataType(), entry.getGroupKey());
+            GroupKey groupId = new GroupKey(
+                entry.getInput().getInputDataset().getInputDataType(),
+                entry.getInput().getCallerAlgorithm(),
```
ylwu-amzn: A pre-3.9 coordinator doesn't send `callerAlgorithm`, so it arrives null — and since `GroupKey` is a record, all nulls group together. During a rolling upgrade two requests with different caller algorithms still coalesce, which is the failure this change targets.

Worth treating a null `callerAlgorithm` as non-coalescable?

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
```diff
@@ -79,6 +80,11 @@ public class MLInput implements Input {
     // Input data to train model, run trained model to predict or run ML algorithms(no-model-based) directly.
     protected MLInputDataset inputDataset;
 
+    // The caller-supplied algorithm (e.g. TEXT_EMBEDDING vs SPARSE_ENCODING) captured before it is normalized to
+    // REMOTE for remote models. Transport-only, not part of the REST body; used by dynamic batching to avoid
+    // coalescing requests whose caller algorithm differs. Null unless the predict path set it.
+    protected FunctionName callerAlgorithm;
```
ylwu-amzn, line 86: Not a `@Builder` param, so `toBuilder().build()` silently drops it. Not reachable today — the only rebuild, `MLPredictTaskRunner:390`, is the SEARCH_QUERY branch, whose dataset has no `BatchableInput` handler — but it's a quiet trap. Add it to the builder, or note why it's excluded.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
```diff
@@ -72,22 +74,63 @@ public void execute(
...
+        int itemCount,
+        Predictable predictor,
+        TransportChannel channel,
+        ActionListener<MLTaskResponse> listener
+    ) {
+        ActionListener<MLTaskResponse> validatingListener = new ActionListener<>() {
+            @Override
+            public void onResponse(MLTaskResponse response) {
+                try {
+                    handler.distributeExactly(response.getOutput(), itemCount);
```
ylwu-amzn: Called only for validation with the result discarded, so every configured single-call predict allocates per-item wrappers it throws away. A count-only check would avoid it.

common/src/main/java/org/opensearch/ml/common/settings/MLCommonsSettings.java
RexyZhang2001: Updated the settings to `memory.fraction`, `memory.min`, and `memory.max`. The release note and related tests have also been updated.

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
RexyZhang2001: Requests without a reliable caller algorithm are now isolated by entry so they cannot coalesce, with tests covering both null and normalized REMOTE values.

common/src/main/java/org/opensearch/ml/common/input/MLInput.java
RexyZhang2001: Added callerAlgorithm to the builder constructor while retaining the existing three-argument constructor. A test now verifies that toBuilder() preserves it.

plugin/src/main/java/org/opensearch/ml/batch/BatchInferenceExecutor.java
RexyZhang2001: Added a count-only validation path for configured single-call prediction. Per-item outputs are now materialized only when they are needed for result routing.

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
```diff
@@ -245,7 +257,12 @@ private void dispatch(List<QueueEntry> batch) {
                 notifyFailure(entry, new IllegalStateException("Could not compute a batch group key for the predict request"));
                 continue;
             }
-            GroupKey groupId = new GroupKey(entry.getInput().getInputDataset().getInputDataType(), entry.getGroupKey());
+            FunctionName callerAlgorithm = entry.getInput().getCallerAlgorithm();
+            // A pre-3.9 coordinator cannot send callerAlgorithm. The receiving node may therefore see null, or
+            // REMOTE if the request was already normalized before transport. In either case, isolate the request:
+            // its original caller type is unknown, so coalescing it could mix text embedding with sparse encoding.
+            Object callerAlgorithmKey = callerAlgorithm != null && callerAlgorithm != FunctionName.REMOTE ? callerAlgorithm : entry;
```
ylwu-amzn: Isolating null is right, but folding `REMOTE` in with it disables coalescing for direct predict callers.

`POST /_plugins/_ml/models/<id>/_predict` has no algorithm param, so `RestMLPredictionAction:96` falls back to the model's function name — `REMOTE` — and `TransportPredictionTaskAction:134` captures that as `callerAlgorithm`. Every such request then gets its own group, so `dynamic_batching` coalesces nothing for direct callers with no indication why. Only neural-search (TEXT_EMBEDDING/SPARSE_ENCODING) still batches.

The mixed-version case I raised yields null, not `REMOTE` — a 3.8 stream just omits the field, and the null-guard on capture means a 3.9-captured value is always the caller's real algorithm. So the null check alone is sufficient; suggest dropping the `!= REMOTE` condition (and `requestsWithNormalizedRemoteCallerAlgorithmDoNotCoalesce` with it).

plugin/src/test/java/org/opensearch/ml/batch/TextDocsBatchableInputTests.java
```diff
@@ -187,6 +187,27 @@ public void distributeReturnsEmptyForNullModelOutputs() {
         assertTrue(handler.distribute(withNull).isEmpty());
     }
 
+    @Test
```
ylwu-amzn, line 190: Nit: worth also asserting `resultCount(o) == distribute(o).size()` — the javadoc removed here warned about exactly this drift between the two paths.

plugin/src/main/java/org/opensearch/ml/batch/DynamicBatchingQueue.java
RexyZhang2001: REMOTE callers now coalesce. The initial coordinator pass captures REMOTE, while a redispatched request without callerAlgorithm remains isolated for rolling-upgrade safety; tests cover both paths.

plugin/src/test/java/org/opensearch/ml/batch/TextDocsBatchableInputTests.java
RexyZhang2001: Added an assertion that resultCount(output) matches distribute(output).size() for an output containing multiple tensor groups.

ylwu-amzn (approved):
Approving for 3.9 — no blocking issues. Both fix claims verified, and `org.opensearch.ml.batch.*` / `action.prediction.*` / `processor.*` / `RestMLPredictionActionTests` pass locally (376 tests).

Correcting my earlier comment: a direct `POST /_plugins/_ml/models/<id>/_predict` on a remote model resolves the algorithm to `REMOTE`, and `MLInput.parse` routes that to `RemoteInferenceMLInput`, which always builds a `RemoteInferenceInputDataSet` — `BatchableInputRegistry` registers `TEXT_DOCS` only, so those requests never reach the grouping path. There was no lost coalescing, and the new `|| isDispatchTask()` clause can't change grouping for any request reachable today (deleting it fails only its own two tests). Non-blocking follow-up, post-3.9: `MLPredictionTaskRequest.builder()` leaves `dispatchTask` at `false`, so `MemoryProcessingService`, `MLModelTool` and the streaming `StreamingWrapper` path look redispatched on their own coordinating node — inert now, but it would silently disable batching for them if `RemoteInferenceInputDataSet` ever gets a `BatchableInput` handler.

PR #5068, opened 2026-09-21
fix: report 403 instead of hanging when connector execute is denied

### LINKED ISSUE #5067: Connector _execute hangs instead of returning 403 when the caller lacks access
### What is the bug?

`POST /_plugins/_ml/connectors/{connector_id}/_execute` never responds when the caller lacks access to the connector. The request hangs until the client times out, instead of returning 403.

In `ExecuteConnectorTransportAction.doExecute`, the access check has no `else`:

```java
// plugin/src/main/java/org/opensearch/ml/action/connector/ExecuteConnectorTransportAction.java:96-125
ActionListener<Connector> listener = ActionListener.wrap(connector -> {
    if (connectorAccessControlHelper.validateConnectorAccess(client, connector)) {
        ...
        connector.decrypt(finalConnectorAction, encryptor::decrypt, null, decryptSuccessfulListener);
    }
    // no else -> listener is never completed
}, e -> { ... });
```

The only `else` in the method (`:132-134`) belongs to the outer `doesMultiTenantIndexExist` check, not to the access check. So when `validateConnectorAccess` returns `false`, neither `actionListener.onResponse` nor `actionListener.onFailure` is ever called, and the transport action never completes.

Every other connector action reports denial explicitly — e.g. `UpdateConnectorTransportAction` fails with `"You don't have permission to update the connector, connector id: ..."`.

### How can one reproduce the bug?

1. Enable connector access control: `plugins.ml_commons.connector_access_control_enabled: true`
2. As user A, create a connector with `access_mode: private`
3. As user B (no backend role granting access), call:

```
POST /_plugins/_ml/connectors/<connector_id>/_execute
{ "parameters": { ... } }
```

The request hangs. Expected: 403.

### What is the expected behavior?

A 403 naming the connector id, consistent with the other connector APIs.

### Do you have any additional context?

**The existing test pins the current behaviour as if it were intended**, so a fix has to update it or it will fail:

```java
// plugin/src/test/java/org/opensearch/ml/action/connector/ExecuteConnectorTransportActionTests.java:181-183
// When access is denied, no action should be taken on the listener
verify(actionListener, times(0)).onResponse(any());
verify(actionListener, times(0)).onFailure(any());
```

That is worth flagging for whoever picks this up: the hang is asserted, not merely uncovered.

Impact is a hung request holding client resources, and a caller that cannot distinguish "denied" from "server stopped responding". It is not a data-integrity or authorization-bypass problem — access is correctly denied, it just is not reported.

Pre-existing and unrelated to MCP; noticed while reviewing #5063, which touches this file for a different reason (rejecting MCP connectors on `_execute`) but deliberately leaves this alone. Fix is roughly an `else` branch plus the test assertion above.

Verified on `main` at `eda32fbaa`.

ylwu-amzn:
### Description

`POST /_plugins/_ml/connectors/{connector_id}/_execute` never responded when the caller lacked access to the connector — the request hung until the client timed out instead of returning 403.

The access check had no `else`, so on denial the listener was simply never completed:

```java
if (connectorAccessControlHelper.validateConnectorAccess(client, connector)) {
    ...
    connector.decrypt(finalConnectorAction, encryptor::decrypt, null, decryptSuccessfulListener);
}
// nothing here -> actionListener is never called
```

The only `else` in that method belongs to the outer `doesMultiTenantIndexExist` check, which is what makes the gap easy to miss when reading the code.

Denial now fails with 403 and the connector id, following `GetConnectorTransportAction:114` ("You don't have permission to access this connector", `FORBIDDEN`) — the closest precedent for a read/use operation. The id is included because it is useful when several connectors are in play, matching `UpdateConnectorTransportAction`.

Access control itself was never broken: access was correctly denied, it just was not reported. So this is a hung request and an indistinguishable failure mode, not an authorization bypass.

### Issues Resolved

Resolves #5067

### Testing

`./gradlew :opensearch-ml-plugin:test` passes, `jacocoTestCoverageVerification` and `spotlessCheck` clean. `integTest` not run locally.

The existing `testExecute_AccessDenied` **asserted the old behaviour directly**:

```java
// When access is denied, no action should be taken on the listener
verify(actionListener, times(0)).onResponse(any());
verify(actionListener, times(0)).onFailure(any());
```

So the hang was pinned by a test rather than merely uncovered — any fix had to update it. It now asserts the reported failure (message and `RestStatus.FORBIDDEN`). I verified the updated assertion actually catches the bug by removing the new `else` branch and re-running: it fails with "Wanted but not invoked".

### Check List
- [ ] New functionality has been documented. — n/a: bug fix. This does change a response for a previously-hanging request; see below.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). — n/a: no request/response schema change.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). — n/a: no documented behaviour promised the old outcome, which was a timeout.

Behaviour change for reviewers: a denied `_execute` call goes from *no response* to **403**. Any client that currently relies on the timeout would see a prompt error instead, which is the intent.


Voyagerroc-Lab (commented):
Reviewed the denied-connector execution path and listener semantics. Completing the ActionListener with an explicit 403 prevents the prior indefinite request hang, keeps successful execution unchanged, and the regression test verifies both the message and RestStatus.FORBIDDEN while excluding onResponse. I found no blocking correctness issue.

Voyagerroc-Lab (commented):
Reviewed the denied-connector execution path and listener semantics. Completing the ActionListener with an explicit 403 prevents the prior indefinite request hang, keeps successful execution unchanged, and the regression test verifies both the message and RestStatus.FORBIDDEN while excluding onResponse. I found no blocking correctness issue.

jiapingzeng (approved):

ylwu-amzn:
Correction to this PR's description, for anyone reading it later (including for the 3.9 backport).

The description says the bug is reachable via `POST /_plugins/_ml/connectors/{connector_id}/_execute`. **That REST route does not exist.** `MLExecuteConnectorAction` is registered only as a transport action (`MachineLearningPlugin.java:602`), with no `RestHandler`, and its only production caller is `ConnectorTool.run()` (`ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/ConnectorTool.java:80`).

So the real path to the hang is:

`POST /_plugins/_ml/agents/{agent_id}/_execute` on an agent configured with a `ConnectorTool` → `ConnectorTool` → `MLExecuteConnectorAction` → the un-completed listener

which means the caller that hangs is server-side tool-invocation code with a `connector_id` fixed by whoever built the agent, not an arbitrary client supplying a path parameter. The exposure is narrower than the description implies: it needs an agent whose `ConnectorTool` points at a connector the *invoking* user cannot access.

The fix and the test are unaffected — the listener was genuinely never completed on denial, and it now returns 403. Only the reproduction path in the description was wrong.

Related: the same incorrect endpoint appears in #5067.

Two follow-ups worth filing separately, from a review of this change:
1. `UpdateConnectorTransportAction:188-193` returns 400 (`IllegalArgumentException`) for the same authorization failure, and `DeleteConnectorTransportAction:110` throws `MLValidationException`, which is not an `OpenSearchException` and therefore surfaces as **500**. Both should be 403, matching `GetConnectorTransportAction` and this change.
2. There is no integration coverage of a denied connector execution through the actual agent/`ConnectorTool` path. `RestConnectorToolIT` only covers the happy path, with connector access control disabled.

PR #5069, opened 2026-09-21
Warn in retention dry-run when the retention job is not scheduled

### LINKED ISSUE #5065: [BUG - Low] Retention dry-run reports pending deletions when the retention job is not scheduled, and gives no warning
### **What is the bug?**
`POST /_plugins/_ml/memory_containers/{id}/_retention/_dry_run` (and the cluster-wide route) exists to answer "what would the scheduled retention job delete?". It correctly reports an honest zero, with a warning, when retention is **disabled**:

`MemoryRetentionJobProcessor.java:2033-2043` (at merge commit `2252bd794da78c299cba839fe295c4e7219a2ef7`, from #4952):

```java
// Retention is disabled cluster-wide, so run() would not execute and nothing would be deleted. Skip the
// read-only passes and report an honest zero - the whole feature stays gated behind this flag.
if (!clusterService.getClusterSettings().get(ML_COMMONS_MEMORY_RETENTION_ENABLED)) {
    ctx.warnings.add(
        "retention is disabled cluster-wide (plugins.ml_commons.memory.retention_enabled=false);"
            + " the scheduled job will not run, so nothing would be deleted"
    );
    listener.onResponse(buildDryRunResult(ctx));
    return;
}
```

That mirrors the one gate on the real execution path (`MemoryRetentionJobProcessor.java:158-161`). **But `retention_enabled=true` is not sufficient for the job to run — the job document also has to exist and be scheduled.** The dry-run never checks that, so when retention is enabled but the job was never scheduled, the dry-run reports pending deletions for a job that will never execute, with no warning of any kind.

The counting itself is correct; the problem is that the dry-run's stated contract ("what the retention job *would* delete") is violated by the one precondition it does not verify.

### **How can one reproduce the bug?**
Reproduced on a single-node OpenSearch 3.9.0 cluster with `opensearch-ml` 3.9.0.0.

Prevent the jobs index from being created, so the retention job can never be scheduled. (In my case this happened *by itself* — see the note below — but it can be forced deterministically:)

```bash
# 1. Enable agentic memory + retention, and configure a default policy that will match data
PUT _cluster/settings
{"persistent":{
  "plugins.ml_commons.agentic_memory_enabled": true,
  "plugins.ml_commons.memory.retention_enabled": true,
  "plugins.ml_commons.memory.default_session_max_count": 1
}}

# 2. Create a container and 3 sessions (max_count=1 makes 2 of them deletable)
POST /_plugins/_ml/memory_containers/_create        {"name":"repro_mc"}
POST /_plugins/_ml/memory_containers/<id>/memories/sessions  {"summary":"s1"}
POST /_plugins/_ml/memory_containers/<id>/memories/sessions  {"summary":"s2"}
POST /_plugins/_ml/memory_containers/<id>/memories/sessions  {"summary":"s3"}

# 3. Dry-run
POST /_plugins/_ml/memory_containers/<id>/_retention/_dry_run
```

Observed response — two sessions reported as deletable, and the only warning is about the orphan baseline:

```json
{
  "container_id": "<id>",
  "policy_source": "default",
  "effective_policy": {"sessions": {"max_count": 1}, ...},
  "would_delete": {
    "sessions": {"total": 2, "by_reason": {"retention_days": 0, "max_count": 2}},
    "long_term": {"total": 0, "by_reason": {}},
    "history": {"total": 0, "by_reason": {}},
    "working_memory": {"total": 0, "by_reason": {"cascade": 0, "ttl": 0, "orphan": 0}}
  },
  "total_would_delete": 2,
  "pinned_would_skip": 0,
  "warnings": ["orphan sweep has not yet observed this container (no baseline); ..."]
}
```

Meanwhile the job does not exist and cannot be triggered:

```bash
GET _cat/indices/.plugins-ml-jobs
# -> index does not exist

POST /_plugins/_ml/memory_containers/_retention/_execute
# -> 500 {"error":{"type":"status_exception","reason":"Failed to trigger memory retention job"}}
```

Nothing would ever be deleted, but the dry-run says two documents would be.

### **What is the expected behavior?**
When retention is enabled but no scheduled retention job exists, the dry-run should say so, in the same place and the same way it already reports the disabled case. Something like:

> `retention is enabled but no scheduled retention job was found (.plugins-ml-jobs/MEMORY_RETENTION); the job will not run until it is scheduled, so nothing would be deleted`

Whether the counts are then still reported or short-circuited to zero (as the disabled branch does) is a judgement call — I'd lean toward **still reporting the counts but adding the warning**, since an operator debugging this wants to see both "the policy would match these 2 documents" and "but the job isn't scheduled."

### **Suggested fix**

Add an existence check for the job document alongside the existing `retention_enabled` check at `MemoryRetentionJobProcessor.java:2035`. The job is indexed at id `MLJobType.MEMORY_RETENTION.name()` in `CommonValue.ML_JOBS_INDEX` (`MLTaskManager.indexMemoryRetentionJob`, `:583-600`), so this is a single `GetRequest` against `.plugins-ml-jobs`. For the cluster-wide route the check is per-invocation, not per-container, so it can be hoisted and evaluated once.

OS 3.9

nathaliellenaa:
### Description
`retention_enabled=true` is necessary but not sufficient for the retention job to run — job-scheduler drives it from a document in `.plugins-ml-jobs`, and that registration can be deferred or fail silently (`indexJob` creates the index and writes the document as two steps, a failed write is only logged, and the listener's `startedMemoryRetentionJob` latch is set regardless). Retention is then enabled but inert, and the dry-run reported pending deletions for a job that would never run.

This resolves the job's actual state once per request in `TransportMemoryRetentionDryRunAction.doExecute` and passes the resulting advisory warning into `dryRunContainer`, so one lookup covers the cluster-wide route too and every container in a response sees the same answer. An absent jobs index is resolved from cluster state with no I/O; the job document is read only when the index exists. Unregistered and disabled are reported differently, since one needs the job scheduled and the other needs it re-enabled. Counts are still reported rather than short-circuited to zero, so an operator sees both that the policy matches documents and that the job will not act on them. Any failure to determine the state emits no warning and never fails a dry-run that would otherwise succeed.

### Testing

Seven unit tests added — two for `dryRunContainer` surfacing the warning while preserving counts, five for the resolver's state mapping (index missing, document missing, disabled, active, indeterminate). The index-missing case also asserts `verify(client, never()).get(...)`, pinning the no-I/O fast path. All existing tests in the affected packages pass unmodified; `spotlessCheck` clean.

Verified end-to-end on a clean cluster built from source:

| State | Result |
|---|---|
| document present, `enabled: true` | no warning (single **and** cluster-wide) |
| index present holding only `STATS_COLLECTOR`, retention document absent | `…no retention job is registered…` (both routes) |
| document present, `enabled: false` | `…the retention job…is disabled; it will not run until it is enabled` |
| jobs index deleted | `…no retention job is registered…` (no-I/O path) |
| job re-registered, `enabled: true` | warning clears (both routes) |
| nonexistent container | 404 unchanged |

### Related Issues
Resolves https://github.com/opensearch-project/ml-commons/issues/5065

### Check List


plugin/src/main/java/org/opensearch/ml/jobs/processors/MemoryRetentionJobProcessor.java
```diff
@@ -2515,6 +2515,20 @@ public void dryRunContainer(
                 listener.onResponse(buildDryRunResult(ctx));
                 return;
             }
+            // Retention being enabled only means the job is allowed to run; the job document the scheduler drives
+            // must also exist. Registration happens on a cluster state change and can be deferred or fail (for
+            // example when the jobs index cannot be created), leaving retention enabled but inert - and the dry-run
+            // then reports deletions for a job that will never execute. The counts are still reported rather than
+            // short-circuited to zero, so an operator sees both that the policy matches documents and that the job
+            // is missing.
+            if (!clusterService.state().metadata().hasIndex(ML_JOBS_INDEX)) {
```
ylwu-amzn: This misses the "registration failed" half of the state the comment describes. `MLTaskManager.indexJob:751` creates the index (`initMLJobsIndex`) and writes the job doc as two separate steps; a failure on the second is only logged (`:766`), while `MLCommonsClusterEventListener:200` has already set `startedMemoryRetentionJob = true` regardless of the async outcome, so there is no retry. The resulting state is index present, `MEMORY_RETENTION` doc absent — job inert, no warning. (`.plugins-ml-jobs` is also shared with `STATS_COLLECTOR`/`BATCH_TASK_UPDATE`, `MLTaskManager:728`/`:571`, but that isn't needed to hit this.)

So index-absent ⇒ no job doc is sound, but the converse isn't, and the comment reads doc-level ("the job document ... must also exist") while the check is index-level.

On the reason given for skipping the `GetRequest`: hoisting `new GetRequest(ML_JOBS_INDEX, MLJobType.MEMORY_RETENTION.name())` into `TransportMemoryRetentionDryRunAction.doExecute` and passing the boolean down is one GET per request, not per container — it covers the cluster-wide route too, which is what #5065 suggested. It needs a stashed thread context since this is a system index (`MLTaskManager.reconcileMemoryRetentionJob:671` is the pattern). If you'd rather keep the index-level check, worth saying in the comment that it's a proxy and what it misses.

ylwu-amzn: Index-absent ⇒ no job doc is sound, but the converse isn't, so the "registration failed" case stays silent: `MLTaskManager.indexJob:751` creates the index and writes the doc as two steps, a failed doc write is only logged (`:766`), and `MLCommonsClusterEventListener:200` latches `startedMemoryRetentionJob = true` regardless — index present, no `MEMORY_RETENTION` doc, no retry, no warning.

`new GetRequest(ML_JOBS_INDEX, MLJobType.MEMORY_RETENTION.name())` hoisted into `TransportMemoryRetentionDryRunAction.doExecute` is one GET per request and covers the cluster-wide route too; needs a stashed context like `MLTaskManager:671`.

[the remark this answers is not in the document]
nathaliellenaa: Yes — fixed. It now reads the job document at `.plugins-ml-jobs/MEMORY_RETENTION` and checks its `enabled` flag, so it covers three states instead of one:

| State | Before | Now |
|---|---|---|
| jobs index absent | warned | warned |
| index present, no `MEMORY_RETENTION` document | missed | warned |
| document present, `enabled: false` | missed | warned |

nathaliellenaa: Both done — the resolver now reads the `MEMORY_RETENTION` document rather than just the index, with distinct warnings for not-registered vs disabled, which covers the failed-write/latched case you describe (verified with the index present holding only `STATS_COLLECTOR`). The GET is hoisted into `doExecute` so it's one per request and covers the cluster-wide route, and it stashes the thread context via` runBefore(..., context::restore)`

ylwu-amzn (approved):

